PR #1, opened 2024-04-24
Fix compilation with GCC 12/13

da-gazzi:
See title - there were forgotten `#include`'s of C++ standard library headers which break compilation with recent versions of GCC.


olofj:
Thanks for your contribution!
I think GCC 13 also complains about the older fmt library in some cases, but these are obviously good changes.

I'll get this merged (we're still on our internal repo as primary so it won't show up as merged here, but it'll be your commits and ownership).

PR #4, opened 2024-05-24
Add API for fast 32-bit writes to static TLBs

joelsmithTT:
Existing code paths for performing such a write suffer from limitations: The "REG_TLB" path reconfigures a single 16M TLB and performs multiple hash table lookups (costing on the order of hundreds of nanoseconds each) prior to performing the write (on the order of tens of nanoseconds).  The non "REG_TLB" path can take two directions: use of an alternate 16M TLB (incurring the same performance penalty) or use of a statically-mapped TLB (if such a mapping exists).  These paths contain read-modify-write logic which may result in unexpected system behavior.

This change adds a mechanism that is patterned after the existing `get_fast_pcie_static_tlb_write_callable` API.  The intent is for callers requiring fast 32-bit write access to a specific core to create and maintain a write32 function using this new API.  The API is designed so that the resulting function can be called in a hot path.

### DIFF
--- a/device/tt_device.h
+++ b/device/tt_device.h
@@ -19,6 +19,8 @@
 #include "device/tt_cluster_descriptor_types.h"
 #include "device/tlb.h"
 
+#include "device/tt_io.hpp"
+
 using TLB_OFFSETS = tt::umd::tlb_offsets;
 using TLB_DATA = tt::umd::tlb_data;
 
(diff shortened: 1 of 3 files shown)


davorchap:
@joelsmithTT what's the timeline on merging this PR? 
This is critical for Metal run-time perf.

davorchap:
fyi @tt-asaigal

joelsmithTT:
See instead: https://github.com/tenstorrent/tt-umd/pull/6

PR #10, opened 2024-06-18
BH Static TLB changes for worker cores

pjanevskiTT:
Part of larger effort on Budabackend to bring up and test Blackhole.

This PR adds Blackhole specific functions for TLBs. It also removes assertions for usage of static TLBs for Blackhole, which were added during initial Blackhole bringup.

[Issue on BBE for core static TLBs](https://yyz-gitlab.local.tenstorrent.com/tenstorrent/budabackend/-/issues/2653)

### DIFF
--- a/device/tt_silicon_driver.cpp
+++ b/device/tt_silicon_driver.cpp
@@ -3063,8 +3063,8 @@ void *tt_SiliconDevice::channel_0_address(std::uint32_t offset, std::uint32_t de
     // Temporary hack for blackhole bringup.
     if (arch_name == tt::ARCH::BLACKHOLE) {
         // Use 195th 2MB TLB onwards.
-        // TLBs up to 195 are used as static or fallback TLBs.
-        const std::uint32_t TLB_CH0_START = 195;
+        // TLBs up to 184 are used as static or fallback TLBs.
+        const std::uint32_t TLB_CH0_START = 184;
         bar0_offset = offset - architecture_implementation->get_dram_channel_0_peer2peer_region_start()
                         + architecture_implementation->get_dynamic_tlb_2m_base() + (TLB_CH0_START * architecture_implementation->get_dynamic_tlb_2m_size());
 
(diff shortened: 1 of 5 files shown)


device/tt_silicon_driver.cpp
```diff
@@ -3063,8 +3059,21 @@ void *tt_SiliconDevice::channel_0_address(std::uint32_t offset, std::uint32_t de
...
+    // Temporary hack for blackhole bringup.
+    if (arch_name == tt::ARCH::BLACKHOLE) {
+
+        // Use 195th 2MB TLB onwards.
+        const std::uint32_t TLB_CH0_START = 195;
+        bar0_offset = offset - architecture_implementation->get_dram_channel_0_peer2peer_region_start()
+                        + architecture_implementation->get_dynamic_tlb_2m_base() + (TLB_CH0_START * architecture_implementation->get_dynamic_tlb_2m_size());
+
+    } else {
+        // This hard-codes that we use 16MB TLB #1 onwards for the mapping. See tt_SiliconDevice::init_pcie_tlb.
```
joelsmithTT, line 3073: Can you remove the duplicate comment at the top of the function?

Does `tt_SiliconDevice::init_pcie_tlb` still exist?

device/tt_silicon_driver.cpp
```diff
@@ -3063,8 +3059,21 @@ void *tt_SiliconDevice::channel_0_address(std::uint32_t offset, std::uint32_t de
...
     auto architecture_implementation = pci_device->hdev->get_architecture_implementation();
-    std::uint64_t bar0_offset = offset - architecture_implementation->get_dram_channel_0_peer2peer_region_start()
-                                + architecture_implementation->get_dynamic_tlb_16m_base() + architecture_implementation->get_dynamic_tlb_16m_size();
+    std::uint64_t bar0_offset;
+
+    // Temporary hack for blackhole bringup.
+    if (arch_name == tt::ARCH::BLACKHOLE) {
+
+        // Use 195th 2MB TLB onwards.
+        const std::uint32_t TLB_CH0_START = 195;
```
abhullar-tt, line 3068: is there a specific reason that the 195th 2MB TLB is the start?

pjanevskiTT: Added a comment, other TLBs before are used as either static of fallback

device/tt_silicon_driver.cpp
pjanevskiTT: Removed, also removed part of the comment for `init_pcie_tlb`, it doesn't exist

broskoTT (approved):

PR #11, opened 2024-06-18
Lower case attribute of l1, host, and eth interface param structs

abhullar-tt:
Upper case attribute names were causing compile errors in Metal when trying to include `noc/noc_parameters.h` header in a host/device common header

See: https://github.com/tenstorrent/tt-metal/pull/9499/files#diff-513f40ad654150a833a41aee778f08dcb66d60ee6b36e2c6a5b43fd49d52ff8e


joelsmithTT:
LGTM

PR #12, opened 2024-06-21
Blackhole 4GB tlbs

pjanevskiTT:
**Keeping this PR in draft until [*](https://github.com/tenstorrent/tt-umd/pull/10) is merged. Target branch is the branch on that PR to follow changes more easily, I will switch the target branch later on to main*

### Problem

Part of larger effort on Budabackend to bring up and test Blackhole

Blackhole has 32GB BAR4 space which is used for DRAM access (4GB per channel). 8 more TLBs are mapped in BAR0 after the TLBs for 2MB device access. Using this our access to DRAM should be faster because now we don't need to access it in chunks of 2MB or 16MB like in Grayskull and Wormhole

### Solution

This MR enables usage of BAR4, mapping BAR4 segment in runtime virtual memory, and makes sure that every access to DRAM goes through BAR4 segment. Access to DRAM in `write_device_memory` and `read_device_memory` should now happen only through the BAR4 space.

### Testing

Change is tested manually, by verifying that every access to DRAM is done through the BAR4 segment. [Commit that verifies the change](https://github.com/tenstorrent/tt-umd/commit/05eb06c82d89709f453343f6ae6a3112c01a639f) is on scratchpad branch, we probably don't want to merge this

### DIFF
--- a/device/blackhole_implementation.cpp
+++ b/device/blackhole_implementation.cpp
@@ -18,6 +18,18 @@ std::tuple<xy_pair, xy_pair> blackhole_implementation::multicast_workaround(xy_p
 }
 
 tlb_configuration blackhole_implementation::get_tlb_configuration(uint32_t tlb_index) const {
+
+    // If TLB index is in range for 4GB tlbs (8 TLBs after 202 TLBs for 2MB)
+    if (tlb_index >= blackhole::TLB_COUNT_2M && tlb_index < blackhole::TLB_COUNT_2M + blackhole::TLB_COUNT_4G) {
+        return tlb_configuration {
+            .size = blackhole::DYNAMIC_TLB_4G_SIZE,
+            .base = blackhole::DYNAMIC_TLB_4G_BASE,
+            .cfg_addr = blackhole::DYNAMIC_TLB_4G_CFG_ADDR,
+            .index_offset = tlb_index - blackhole::TLB_BASE_INDEX_4G,
+            .offset = blackhole::TLB_4G_OFFSET,
+        };
+    }
+    
     return tlb_configuration{
         .size = blackhole::DYNAMIC_TLB_2M_SIZE,
         .base = blackhole::DYNAMIC_TLB_2M_BASE,
@@ -27,13 +39,19 @@ tlb_configuration blackhole_implementation::get_tlb_configuration(uint32_t tlb_i
     };
 }
 
-std::optional<std::tuple<std::uint32_t, std::uint32_t>> blackhole_implementation::describe_tlb(
+std::optional<std::tuple<std::uint64_t, std::uint64_t>> blackhole_implementation::describe_tlb(
     std::int32_t tlb_index) const {
     std::uint32_t TLB_COUNT_2M = 202;
 
     std::uint32_t TLB_BASE_2M = 0;
     if (tlb_index < 0) {
         return std::nullopt;
+    }
+
+    if (tlb_index >= TLB_COUNT_2M && tlb_index < TLB_COUNT_2M + blackhole::TLB_COUNT_4G) {
+        auto tlb_offset = tlb_index - TLB_COUNT_2M;
+        auto size = blackhole::TLB_4G_SIZE;
+        return std::tuple(blackhole::TLB_BASE_4G + tlb_offset * size, size);
     }
 
     // Only have 2MB TLBs for now
--- a/device/tt_silicon_driver.cpp
+++ b/device/tt_silicon_driver.cpp
@@ -177,6 +177,9 @@ struct TTDeviceBase
     std::vector<DMAbuffer> dma_buffer_mappings;
 
     std::uint32_t read_checking_offset;
+
+    void* bar4_mapping = nullptr;
+    std::uint64_t bar4_mapping_size;
 };
 
 struct TTDevice : TTDeviceBase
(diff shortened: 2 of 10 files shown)


device/tt_silicon_driver.cpp
```diff
@@ -531,6 +537,28 @@ void TTDevice::do_open() {
...
+        // WC mapping
+        // this->bar4_mapping_size = bar2_wc_mapping.mapping_size;
+        // this->bar4_mapping = mmap(NULL, bar2_wc_mapping.mapping_size, PROT_READ | PROT_WRITE, MAP_SHARED, device_fd, bar2_wc_mapping.mapping_base);
+
+        // if (this->bar4_mapping == MAP_FAILED) {
+        //     throw std::runtime_error(std::string("BAR4 WC memory mapping failed for device ") + std::to_string(index) + ".");
+        // }
+
+        // UC mapping
+        this->bar4_mapping_size = bar2_uc_mapping.mapping_size;
```
joelsmithTT, line 555: This looks confusing.  If I understand the code, `bar2_uc_mapping` is misnamed and always corresponds to BAR4, even in WH code path.  `resource2_uc_mapping` or `bar4_uc_mapping` might be a better name?  What do you think?

device/tt_silicon_driver.cpp
```diff
@@ -531,6 +537,28 @@ void TTDevice::do_open() {
 
         this->system_reg_start_offset = (512 - 16) * 1024*1024;
         this->system_reg_offset_adjust = (512 - 32) * 1024*1024;
+    } else if(is_blackhole(device_info.out)) {
+
+        if (bar2_uc_mapping.mapping_id != TENSTORRENT_MAPPING_RESOURCE2_UC) {
+            throw std::runtime_error(std::string("Device ") + std::to_string(index) + " has no BAR4 UC mapping.");
+        }
+
+        // WC mapping
```
joelsmithTT, line 546: Why UC rather than WC?

device/tt_silicon_driver.cpp
```diff
@@ -2140,23 +2176,27 @@ void tt_SiliconDevice::write_device_memory(const void *mem_ptr, uint32_t size_in
...
+    std::optional<std::tuple<std::uint64_t, std::uint64_t>> tlb_data = std::nullopt;
     if(tlbs_init) {
         tlb_index = map_core_to_tlb(tt_xy_pair(target.x, target.y));
         tlb_data = dev->get_architecture_implementation()->describe_tlb(tlb_index);
     }
 
     if (tlb_data.has_value() && address_in_tlb_space(address, size_in_bytes, tlb_index, std::get<1>(tlb_data.value()), target.chip)) {
         auto [tlb_offset, tlb_size] = tlb_data.value();
-        write_block(dev, tlb_offset + address % tlb_size, size_in_bytes, buffer_addr, m_dma_buf_size);
+        if (dev->bar4_mapping != nullptr && tlb_size > BAR0_BH_SIZE) {
```
joelsmithTT, line 2187: There are 1 MiB, 2 MiB, 16 MiB, and 4 GiB TLB sizes supported by UMD.  What is the rationale behind comparing TLB size to BAR0 size?

device/blackhole_implementation.cpp
```diff
@@ -36,6 +48,12 @@ std::optional<std::tuple<std::uint32_t, std::uint32_t>> blackhole_implementation
         return std::nullopt;
     }
 
+    if (tlb_index >= TLB_COUNT_2M && tlb_index < TLB_COUNT_2M + blackhole::TLB_COUNT_4G) {
+        auto tlb_offset = tlb_index - TLB_COUNT_2M;
+        auto size = blackhole::TLB_4G_SIZE;
+        return std::tuple(blackhole::TLB_BASE_4G + tlb_offset * size, size);
+    }
+
     // Only have 2MB TLBs for now
```
joelsmithTT, line 57: This comment no longer seems to be true.

device/tt_silicon_driver.cpp
```diff
@@ -1057,7 +1089,12 @@ void write_block(TTDevice *dev, uint32_t byte_addr, uint32_t num_bytes, const ui
     record_access("write_block_b", byte_addr, num_bytes, false, true, true, false); // addr, size, turbo, write, block, endline
 
     void *reg_mapping;
-    if (dev->system_reg_mapping != nullptr && byte_addr >= dev->system_reg_start_offset) {
+    const uint64_t bar0_size = 512 * 1024 * 1024;
```
joelsmithTT, line 1092: `BAR0_BH_SIZE`?

device/tt_silicon_driver.cpp
joelsmithTT: I see what it is doing (add this constant to the address, which read/write block treats as a flag, of sorts).  This has the potential for confusion for two reasons:
* WH also has BAR4, but a different name (`system_reg_mapping`).  `dev->bar4_mapping` is BH specific, but this may not be obvious to someone reading the code.
* TLB size and BAR size are different things.

It makes sense from the standpoint that writes "beyond" BAR0 go to BAR4.

Not suggesting that you change anything here, but for your consideration:

 Would `if (dev->bar4_mapping != nullptr && tlb_size == blackhole::TLB_4G_SIZE) {` be clearer?  Is there opportunity for different/improved `write_block` and `read_block` functions?

joelsmithTT:
LGTM, left a few remarks for your consideration.

device/tt_silicon_driver.cpp
pjanevskiTT: I was not sure which region should be used so left both, I have investigated how we should do it and it should be in WC region, changed that

device/tt_silicon_driver.cpp
pjanevskiTT: I completely agree that comparison between TLB size and BAR0 size was not clear, problem in tt_silicon_driver in general is that arch specific stuff is not accessible in that file. I have defined TLB_4G_SIZE and changed the code how you suggested. Regarding write/read block, it could be possibly moved to arch specific files and then just exposed to tt silicon driver, because with different configs we want to choose TLB address in different way, as seen above

device/tt_silicon_driver.cpp
pjanevskiTT: Yeah I see what you are saying, I think the easiest way is to rename bar* variables to proper indexing. I can also add the comment saying that our resource -> bar mapping is (0,1,2) -> (0,2,4). I will see what is the best place to add the comment

device/tt_silicon_driver.cpp
```diff
@@ -177,6 +183,9 @@ struct TTDeviceBase
     std::vector<DMAbuffer> dma_buffer_mappings;
 
     std::uint32_t read_checking_offset;
+
+    void* bar4_mapping = nullptr;
```
broskoTT, line 187: Consider renaming to bar4_wc, and bar4_wc_size, to follow naming of other variables.

broskoTT, line 187: Add corresponding code to reset() and drop() functions

joelsmithTT (approved):

broskoTT (approved):

PR #13, opened 2024-07-02
Enabling Hugepage translation

broskoTT:
These are UMD changes needed to enable device writing to a host queue using hugepages.

I've created a PR on top of ongoing @pjanevskiTT 's changes, so that I can test all of the changes in one go, and so that we don't have to merge again a couple of days later.

Issue on BBE tracking this effort: https://yyz-gitlab.local.tenstorrent.com/tenstorrent/budabackend/-/issues/2658

### DIFF
--- a/device/tt_silicon_driver.cpp
+++ b/device/tt_silicon_driver.cpp
@@ -236,7 +236,8 @@ private:
 
     void reset() {
         if (arch == tt::ARCH::BLACKHOLE && bar2_uc != nullptr && bar2_uc != MAP_FAILED) {
-            // Disable ATU.
+            // Disable ATU index 0
+            // TODO: Implement disabling for all indexes, once more host channels are enabled.
             uint64_t iatu_index = 0;
             uint64_t iatu_base = UNROLL_ATU_OFFSET_BAR + iatu_index * 0x200;
             uint32_t region_ctrl_2 = 0 << 31; // REGION_EN = 0
@@ -1229,9 +1230,7 @@ void set_use_dma(bool msi, uint32_t dma_block_size_read_threshold_bytes, uint32_
 
 void write_regs(volatile uint32_t *dest, const uint32_t *src, uint32_t word_len) {
     while (word_len-- != 0) {
-        uint32_t temp;
-        memcpy(&temp, src++, sizeof(temp));
-        *dest++ = temp;
+        *dest++ = *src++;
     }
 }
 
@@ -2028,12 +2027,13 @@ void tt_SiliconDevice::initialize_pcie_devices() {
     }
 
     // If requires multi-channel or doesn't support mmio-p2p, init iatus without p2p.
-    if (m_num_host_mem_channels > 1 || arch_name != tt::ARCH::GRAYSKULL) {
+    if (m_num_host_mem_channels <= 1 && arch_name == tt::ARCH::GRAYSKULL) {
+        init_pcie_iatus();
+    } else {
+        // TODO: Implement support for multiple host channels on BLACKHOLE.
         log_assert(!(arch_name == tt::ARCH::BLACKHOLE && m_num_host_mem_channels > 1),
             "More channels are not yet supported for Blackhole");
         init_pcie_iatus_no_p2p();
-    } else {
-        init_pcie_iatus();
     }
 
     init_membars();
@@ -2338,6 +2338,7 @@ void tt_SiliconDevice::write_dma_buffer(
 
     void * user_scratchspace = nullptr;
     if(hugepage_mapping.at(src_device_id).at(channel)) {
+      log_assert(size <= HUGEPAGE_REGION_SIZE, "write_dma_buffer data has larger size {} than destination buffer {}", size, HUGEPAGE_REGION_SIZE);
       log_debug(LogSiliconDriver, "Using hugepage mapping at address {} offset {} chan {} size {}",
         hugepage_mapping.at(src_device_id).at(channel),
         (address & HUGEPAGE_MAP_MASK),
@@ -2346,6 +2347,7 @@ void tt_SiliconDevice::write_dma_buffer(
       user_scratchspace = static_cast<char*>(hugepage_mapping.at(src_device_id).at(channel)) + (address & HUGEPAGE_MAP_MASK);
     }
     else if(buf_mapping) {
+      log_assert(size <= DMA_BUF_REGION_SIZE, "write_dma_buffer data has larger size {} than destination buffer {}", size, DMA_BUF_REGION_SIZE);
       log_debug(LogSiliconDriver, "Using DMA Buffer at address {} offset {} size {}",
         buf_mapping,
         address,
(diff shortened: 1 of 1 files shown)


joelsmithTT:
Does `std::uint64_t tt_SiliconDevice::get_pcie_base_addr_from_device(uint32_t device_id) const` need to return something other than zero for this case?  Based on my reading of the ARC FW, I think it needs a `(4ULL << 58)`.  58 because upper six bits select the window, and 4 because windows 0-3 have `atu_bypass` set to 1 in the ARC codes.

device/tt_silicon_driver.cpp
```diff
@@ -1174,19 +1232,24 @@ void set_use_dma(bool msi, uint32_t dma_block_size_read_threshold_bytes, uint32_
     g_DMA_BLOCK_SIZE_WRITE_THRESHOLD_BYTES = dma_block_size_write_threshold_bytes;
 }
 
+void write_regs(volatile uint32_t *dest, const uint32_t *src, uint32_t word_len) {
+    while (word_len-- != 0) {
+        uint32_t temp;
```
joelsmithTT, line 1237: I see that this is refactored rather than new code.  But I wonder if
```
    while (word_len-- != 0) {
        *dest++ = *src++;
    }
```
is a clearer way to express it?

device/tt_silicon_driver.cpp
```diff
@@ -224,6 +232,14 @@ struct TTDevice : TTDeviceBase
     TTDevice() = default;
 
     void reset() {
+        if (arch == tt::ARCH::BLACKHOLE && bar2_uc != nullptr && bar2_uc != MAP_FAILED) {
+            // Disable ATU.
```
joelsmithTT, line 236: Does this disable all regions that were enabled or just index zero?

abhullar-tt:
> Does `std::uint64_t tt_SiliconDevice::get_pcie_base_addr_from_device(uint32_t device_id) const` need to return something other than zero for this case? Based on my reading of the ARC FW, I think it needs a `(4ULL << 58)`. 58 because upper six bits select the window, and 4 because windows 0-3 have `atu_bypass` set to 1 in the ARC codes.

`(4ULL << 58)` has been working for Metal for host <-> device data transfers

abhullar-tt:
do these changes need to KMD version `1.28-bh1-iommu`?

joelsmithTT:
> do these changes need to KMD version `1.28-bh1-iommu`?
No

device/tt_silicon_driver.cpp
joelsmithTT: Oh - I see that `initialize_pcie_devices` ensures only a single channel in the Blackhole case.  May be worth leaving a TODO here (or just implement it) so that when the code grows the ability to use multiple channels, all corresponding iATU regions are disabled here rather than just index 0.

device/tt_silicon_driver.cpp
```diff
@@ -1975,14 +2033,14 @@ void tt_SiliconDevice::initialize_pcie_devices() {
...
 
     // If requires multi-channel or doesn't support mmio-p2p, init iatus without p2p.
-    // MT Initial BH - skip device to host related init
-    if (arch_name != tt::ARCH::BLACKHOLE) {
-        if (m_num_host_mem_channels > 1 || arch_name != tt::ARCH::GRAYSKULL) {
-            init_pcie_iatus_no_p2p();
-        } else {
-            init_pcie_iatus();
-        }
+    if (m_num_host_mem_channels > 1 || arch_name != tt::ARCH::GRAYSKULL) {
```
joelsmithTT, line 2036: At first glance, it was not obvious to me which path this code takes in the case of BH and 1 memory channel.  I think there is an opportunity to improve the clarity here.

If I understand the code correctly, `init_pcie_iatus` is used only for Grayskull and a single host memory channel.  What do you think about,
```
if (m_num_host_mem_channels == 1 && arch_name == tt::ARCH::GRAYSKULL {
    init_pcie_iatus();
} else {
    ...
```

device/tt_silicon_driver.cpp
```diff
@@ -3025,11 +3098,40 @@ int tt_SiliconDevice::iatu_configure_peer_region (int logical_device_id, uint32_
     struct PCIdevice* pci_device = get_pci_device(logical_device_id);
     auto architecture_implementation = pci_device->hdev->get_architecture_implementation();
 
-    bar_write32(logical_device_id, architecture_implementation->get_arc_csm_mailbox_offset() + 0 * 4, region_id_to_use);
-    bar_write32(logical_device_id, architecture_implementation->get_arc_csm_mailbox_offset() + 1 * 4, dest_bar_lo);
-    bar_write32(logical_device_id, architecture_implementation->get_arc_csm_mailbox_offset() + 2 * 4, dest_bar_hi);
-    bar_write32(logical_device_id, architecture_implementation->get_arc_csm_mailbox_offset() + 3 * 4, region_size);
-    arc_msg(logical_device_id, 0xaa00 | architecture_implementation->get_arc_message_setup_iatu_for_peer_to_peer(), true, 0, 0);
+    // BR: ARC doesn't work yet on Blackhole, so programming ATU directly. Should be removed when arc starts working.
```
pjanevskiTT, line 3101: Maybe add TODO to this comment. It will be more clear for others looking into this that this needs to be fixed when ARC is enabled

device/tt_silicon_driver.cpp
```diff
@@ -3025,11 +3098,40 @@ int tt_SiliconDevice::iatu_configure_peer_region (int logical_device_id, uint32_
...
+    if (arch_name == tt::ARCH::BLACKHOLE) {
+        uint64_t base_addr = region_id_to_use*region_size;
```
pjanevskiTT, line 3103: nit: space between * and operands, same in the line below

device/tt_silicon_driver.cpp
```diff
@@ -3025,11 +3098,40 @@ int tt_SiliconDevice::iatu_configure_peer_region (int logical_device_id, uint32_
...
+        uint32_t region_ctrl_1 = 1 << 13; // INCREASE_REGION_SIZE = 1
+        uint32_t region_ctrl_2 = 1 << 31; // REGION_EN = 1
+        uint32_t region_ctrl_3 = 0;
+        uint32_t base_addr_lo = base_addr & 0xffffffff;
+        uint32_t base_addr_hi = (base_addr >> 32) & 0xffffffff;
+        uint32_t limit_address_lo = limit_address & 0xffffffff;
+        uint32_t limit_address_hi = (limit_address >> 32) & 0xffffffff;
+
+        uint64_t iatu_index = 0;
+        uint64_t iatu_base = UNROLL_ATU_OFFSET_BAR + iatu_index * 0x200;
```
pjanevskiTT, line 3116: What is 0x200 constant for?

pjanevskiTT (approved):
Looks good to me, I am giving approve just as a sign that I looked into the changes, didn't want to replicate some of the comments, when comments are resolved I feel good about merging this

broskoTT:
> Does `std::uint64_t tt_SiliconDevice::get_pcie_base_addr_from_device(uint32_t device_id) const` need to return something other than zero for this case? Based on my reading of the ARC FW, I think it needs a `(4ULL << 58)`. 58 because upper six bits select the window, and 4 because windows 0-3 have `atu_bypass` set to 1 in the ARC codes.

This is a good question. I'll edit this function, seems to me it does have to be updated.
However I don't see it being called anywhere in umd or budabackend (hence why I haven't already seen and edited it).

device/tt_silicon_driver.cpp
broskoTT: Don't want to spend much time on refactoring existing code at this point. But I will edit and test this.

device/tt_silicon_driver.cpp
broskoTT: I'd rather leave a comment. I'm new in this codepath, so any code change however small can take me significant time. Not sure that mapping more than one channel was used on wormhole, so I'd say this is not a priority.

device/tt_silicon_driver.cpp
broskoTT: I can't say I didn't have the same confusion at one point :) will edit this.

device/tt_silicon_driver.cpp
broskoTT: I don't want to leave a comment/constant guessing it. But from the code, I'd say that there are different registers for different ATU regions (i.e. you can setup multiple address range translations), and they are spaced out by this amount. This was copied from example python code.

joelsmithTT:
LGTM.  RE: `get_pcie_base_addr_from_device` - probably only Metal uses it then, but I imagine BBE has a similar constant somewhere.

PR #16, opened 2024-07-10
Force a single memory channel for BH, don't throw if more channels

broskoTT:
Currently we don't support more than a single channel for blackhole. Instead of hitting an assert when there is more than 1 hugepage setup on system, and throw, force using only a single channel, and ignore the rest.


pjanevskiTT (approved):

joelsmithTT (approved):
It looks fine, but why not add support for multiple channels?  A previous commit introduced the ability to program iATU from userspace.  What else is left to do?

PR #18, opened 2024-07-12
Add env var for AICLK get_clock

pjanevskiTT:
Add env var to set value which is going to be retreived from get_clock function for Blackhole. 

Lot of Blackhole chips used for testing at the moment are underclocked to 800MHz and ARC messages don't work at the moment so this is a problem when we want to actually read AICLK value. It is needed for testing Blackhole on Buda at the moment for Perf infrastructure

### DIFF
--- a/device/tt_silicon_driver.cpp
+++ b/device/tt_silicon_driver.cpp
@@ -2405,6 +2405,18 @@ void tt_SiliconDevice::set_pcie_power_state(tt_DevicePowerState state) {
 }
 
 int tt_SiliconDevice::get_clock(int logical_device_id) {
+
+    // TODO: remove this once ARC messages work.
+    // This is currently used only for testing and bringing up Blackhole on Buda.
+    if (arch_name == tt::ARCH::BLACKHOLE) {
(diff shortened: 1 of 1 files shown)


broskoTT (approved):

pjanevskiTT:
@joelsmithTT we need this for Buda, specifically Marko Bezulj needs it for Perf infra on Buda. He is waiting for permissions on UMD, he can comment here when this change is ok for him, and you can approve if you think this is ok. We are not aware of other use cases at the moment when this function is called but since it uses ARC probably we don't want to use it for Blackhole  anywhere at the moment

mbezuljTT:
> Add env var to set value which is going to be retreived from get_clock function for Blackhole.
> 
> Lot of Blackhole chips used for testing at the moment are underclocked to 800MHz and ARC messages don't work at the moment so this is a problem when we want to actually read AICLK value. It is needed for testing Blackhole on Buda at the moment for Perf infrastructure

Thanks for making this change @pjanevskiTT !

joelsmithTT (approved):

PR #19, opened 2024-07-16
Fix BH address translation

broskoTT:
The codepath hitting channel_address (former channel_0_address, but renamed to reduce confusion) is used by Buda when writing to these queues in DRAM.
Previous generations of chips didn't use static tlbs for channels other than 0. Now, BAR4 on BH allows us to map whole DRAM, hence a modification is needed to this function.

This change was integrated in BudaBackend and tested with accompanying BBE changes.

### DIFF
--- a/device/tt_silicon_driver.cpp
+++ b/device/tt_silicon_driver.cpp
@@ -3218,11 +3218,11 @@ void *tt_SiliconDevice::channel_address(std::uint32_t offset, const tt_cxy_pair&
     // Temporary hack for blackhole bringup.
     if (arch_name == tt::ARCH::BLACKHOLE) {
         // We use BAR4 segment for mapping for Blackhole.
-        log_assert(tlbs_init, "tlb not init.");
+        log_assert(tlbs_init, "TLBs were not initialized.");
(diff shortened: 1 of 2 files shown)


pjanevskiTT:
@abhullar-tt note that function name and declaration is changed, so just be careful if you are using this function in Metal

device/tt_silicon_driver.cpp
```diff
@@ -3209,17 +3209,21 @@ void tt_SiliconDevice::enable_local_ethernet_queue(const chip_id_t &device_id, i
     }
 }
 
-
-void *tt_SiliconDevice::channel_0_address(std::uint32_t offset, std::uint32_t device_id) const {
-    log_assert(ndesc->is_chip_mmio_capable(device_id), "Cannot call channel_0_address for non-MMIO device");
-    struct PCIdevice* pci_device = get_pci_device(device_id);
+void *tt_SiliconDevice::channel_address(std::uint32_t offset, const tt_cxy_pair& target) {
+    log_assert(ndesc->is_chip_mmio_capable(target.chip), "Cannot call channel_0_address for non-MMIO device");
```
pjanevskiTT, line 3213: Change to "Cannot call channel_address...

pjanevskiTT (approved):

PR #23, opened 2024-07-29
Remove device_data.hpp files and use <arch>_implementation.hpp instead

abhullar-tt:
It seems like only tests were using `<arch>/device_data.hpp` and the values were duplicated in `<arch>_implementation.hpp` so device_data.hpp headers have been removed.


joelsmithTT (approved):
LGTM.  Added @pjanevskiTT as I know he's been working on UMD lately too -- in case this has an effect on the Buda stack.

broskoTT (approved):

PR #25, opened 2024-07-31
TT-Metal remote VCS simulator setup

vtangTT:
Issue: https://yyz-gitlab.local.tenstorrent.com/tenstorrent/tt-metal/-/issues/1

Solution:
- instead of building VCS into metal, use NNG network api to send data in the form of Flatbuffers to a remote (in this case, the VCS simulator)
- the remote VCS will run the VCS simulator and send data back

VCS remote repo found here: https://yyz-gitlab.local.tenstorrent.com/bteng/metal-vcs

ToDo:

- [x] get rid of IO api using `std::vector<uint32_t>` instead of `void*`
- [x] change `TT_METAL_SIMULATOR_EN` to something more general -> changed to `TT_REMOTE_EXE`
- [x] randomize `NNG_SOCKET_NAME` to ensure it's unique (maybe future task)
- [x] squash commits


joelsmithTT (commented):
Overall this looks good.  I have one request: can you flip the implementation of the IO methods, such that the `void*` versions be the primary implementation, with the `std::vector` versions calling them.

Eventually I'd like the methods that accept `std::vector<uint32_t>` to be removed from UMD.  This PR increases the amount of effort it will take to remove these methods.  My rationale for wanting to remove them is as follows.

In question is
`write_to_device(std::vector<uint32_t> &vec, ...` vs  `write_to_device(const void* src, size_t size, ...` and similar methods:
1. Flexibility: first version is unnecessarily restrictive
2. Performance: first version can force a copy in various circumstances 
3. Generality: second version can accept anything, including uint32 in a vector
4. Unclear: in the vector-based `read_from_device`, it is unclear without looking at the implementation what will happen if vec is under or over-sized
5. Misuse potential --

This vec32 pattern has resulted in code like the following in tt-metal, which can break subtly if e.g. the structure changes.

```
std::vector<uint32_t> buffer;
buffer.resize(sizeof(some_struct) / 4);
some_struct *my_struct = reinterpret_cast<some_struct*>(&buffer[0]);
/* populate fields in my_struct */
SomeDeviceMethod(buffer);
```

Compare with,
```
some_struct my_struct{};
/* populate it */
BetterDeviceMethod(&my_struct, sizeof(my_struct))
```

device/device_api.h
```diff
@@ -7,4 +7,8 @@
 #pragma once
 #include "device/tt_device.h"
 #include "device/driver_atomics.h"
-#include "device/tt_emulation_device.h"
\ No newline at end of file
+#include "device/tt_emulation_device.h"
+
+#ifdef TT_METAL_SIMULATOR_EN
```
abhullar-tt, line 12: should this be named `TT_UMD_SIMULATOR_EN`?

device/simulation/tt_simulation_device.cpp
```diff
@@ -0,0 +1,261 @@
...
+#include "tt_simulation_device.h"
+#include "uv.h"
+
+tt_SimulationDevice::tt_SimulationDevice(const std::string &sdesc_path) : tt_device(sdesc_path){
+    // log_info(tt::LogEmulationDriver, "Instantiating simulation device");
+    soc_descriptor_per_chip.emplace(0, tt_SocDescriptor(sdesc_path));
+    std::set<chip_id_t> target_devices = {0};
+    
+    // Start VCS simulator in a separate process
+    TT_ASSERT(std::getenv("TT_METAL_VCS"), "TT_METAL_VCS not set, please provide path to the VCS binary");
```
abhullar-tt, line 26: same comment as above should the toggles be prefixed with `TT_METAL` or `TT_UMD`?

device/simulation/tt_simulation_device.cpp
```diff
@@ -0,0 +1,261 @@
...
+    uv_loop_t *loop = uv_default_loop();
+    uv_process_t child_p;
+    uv_process_options_t child_options = {0};
+
+    child_options.file = std::getenv("TT_METAL_VCS");
+    child_options.flags = UV_PROCESS_DETACHED;
+
+    int rv = uv_spawn(loop, &child_p, &child_options);
+    if (rv) {
+        log_info(tt::LogEmulationDriver, "Failed to spawn simulator process: {}", uv_strerror(rv));
```
abhullar-tt, line 36: I think we have a log_fatal

device/simulation/tt_simulation_device.cpp
```diff
@@ -0,0 +1,261 @@
...
+#include "common/logger.hpp"
+#include "common/assert.hpp"
+#include "device/driver_atomics.h"
+#include "device/tt_cluster_descriptor.h"
+
+#include "tt_simulation_device.h"
+#include "uv.h"
+
+tt_SimulationDevice::tt_SimulationDevice(const std::string &sdesc_path) : tt_device(sdesc_path){
+    // log_info(tt::LogEmulationDriver, "Instantiating simulation device");
```
abhullar-tt, line 21: should this be uncommented?

device/simulation/tt_simulation_device.cpp
```diff
@@ -0,0 +1,261 @@
...
+std::map<int,int> tt_SimulationDevice::get_clocks() {
+    return {{0, 0}};
+}
+
+void *tt_SimulationDevice::host_dma_address(std::uint64_t offset, chip_id_t src_device_id, uint16_t channel) const {
+    // might need to implement later
+    return nullptr;
+}
+
+std::uint64_t tt_SimulationDevice::get_pcie_base_addr_from_device() const {
```
abhullar-tt, line 205: we could probably commonize this with tt_SiliconDevice?

abhullar-tt (approved):
This is awesome! going to be so helpful to have this up and working :)

device/simulation/tt_simulation_device.cpp
vtangTT: Wdym by commonize? Right now, I just copy-pasted what's in the `tt_SilliconDevice` to here

device/simulation/tt_simulation_device.cpp
vtangTT: For this, I'll probably rename to something like "TT_REMOTE_EXE" to be more general, since this infra can be used for something like zebu in the future. 

The env var `TT_METAL_SIMULATOR_EN` will change to be not specefic to VCS, and I'm open to any suggestions :D .

device/simulation/tt_simulation_device.cpp
vtangTT: I just removed it and let the `TT_THROW` trigger instead if that works too

vtangTT:
> Overall this looks good. I have one request: can you flip the implementation of the IO methods, such that the `void*` versions be the primary implementation, with the `std::vector` versions calling them.
> 
> Eventually I'd like the methods that accept `std::vector<uint32_t>` to be removed from UMD. This PR increases the amount of effort it will take to remove these methods. My rationale for wanting to remove them is as follows.
> 
> In question is `write_to_device(std::vector<uint32_t> &vec, ...` vs `write_to_device(const void* src, size_t size, ...` and similar methods:
> 
> 1. Flexibility: first version is unnecessarily restrictive
> 2. Performance: first version can force a copy in various circumstances
> 3. Generality: second version can accept anything, including uint32 in a vector
> 4. Unclear: in the vector-based `read_from_device`, it is unclear without looking at the implementation what will happen if vec is under or over-sized
> 5. Misuse potential --
> 
> This vec32 pattern has resulted in code like the following in tt-metal, which can break subtly if e.g. the structure changes.
> 
> ```
> std::vector<uint32_t> buffer;
> buffer.resize(sizeof(some_struct) / 4);
> some_struct *my_struct = reinterpret_cast<some_struct*>(&buffer[0]);
> /* populate fields in my_struct */
> SomeDeviceMethod(buffer);
> ```
> 
> Compare with,
> 
> ```
> some_struct my_struct{};
> /* populate it */
> BetterDeviceMethod(&my_struct, sizeof(my_struct))
> ```

I just removed the `std::vector` implementations since metal doesn't use them anyways. The IO calls should all be passing data through `void *` now. 

Please take a look.

device/simulation/tt_simulation_host.cpp
```diff
@@ -0,0 +1,72 @@
...
+            std::this_thread::sleep_for(std::chrono::seconds(1));
+        }
+    } while (rv != 0);
+}
+
+void tt_SimulationHost::send_to_device(uint8_t *buf, size_t buf_size) {
+    int rv;
+    log_debug(tt::LogEmulationDriver, "Sending messsage to remote..");
+    
+    void *msg = nng_alloc(buf_size);
```
joelsmithTT, line 54: Curious if this alloc and copy is necessary?  (Can `buf` get passed into `nng_send` directly?)

device/simulation/tt_simulation_device_generated.h
```diff
@@ -0,0 +1,215 @@
...
+#include "flatbuffers/flatbuffers.h"
+
+// Ensure the included flatbuffers.h is the same version as when this file was
+// generated, otherwise it may not be compatible.
+static_assert(FLATBUFFERS_VERSION_MAJOR == 24 &&
+              FLATBUFFERS_VERSION_MINOR == 3 &&
+              FLATBUFFERS_VERSION_REVISION == 25,
+             "Non-compatible flatbuffers version included");
+
+struct tt_core;
```
joelsmithTT, line 16: Is there a way to namespace the autogenerated code to avoid the possibility of collisions?  I ask because `tt_core` is somewhat generic.

device/simulation/tt_simulation_device.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+    // Get read response
+    host.recv_from_device(&rd_resp, &rd_rsp_sz);
+
+    auto rd_resp_buf = GetDeviceRequestResponse(rd_resp);
+    if (addr != 0x40){
+        log_info(tt::LogEmulationDriver, "Device reading vec");
+        print_flatbuffer(rd_resp_buf); // 0x40 is host polling device, don't print since it'll spam
+    }
+    std::vector<uint32_t> rd_resp_vec(rd_resp_buf->data()->begin(), rd_resp_buf->data()->end());
```
joelsmithTT, line 149: Is there a way to do this without the allocation and copy into the vector?

joelsmithTT:
> I just removed the `std::vector` implementations since metal doesn't use them anyways. The IO calls should all be passing data through `void *` now.
> 
> Please take a look.
Thanks for doing that.

device/simulation/tt_simulation_device_generated.h
vtangTT: I don't think so, but I can change the name of the struct to something like `tt_simulation_core` and regenerate the flatbuffer schema

device/simulation/tt_simulation_host.cpp
vtangTT: I think doing so with the NNG provided alloc is safer since NNG will then manage the memory space and ensure compatibility with `nng_send/recv`. `nng_send` will also free the memory after sending the message. 

This should prevent any weird behaviour when sending data over the network

device/simulation/tt_simulation_device.cpp
vtangTT: `std::memcpy(mem_ptr, rd_resp_buf->data()->data(), rd_resp_buf->data()->size() * sizeof(uint32_t));` works I believe. If that's what you mean

device/simulation/tt_simulation_host.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include "common/assert.hpp"
+#include "tt_simulation_host.hpp"
+
+tt_SimulationHost::tt_SimulationHost() {
+    // Open socket and create dialer
+    nng_pair1_open(&host_socket);
+    std::string nng_socket_addr_str;
+    const char *ipc_socket_name = std::getenv("NNG_SOCKET_NAME");
+    if (!ipc_socket_name) {
+        nng_socket_addr_str = std::string(NNG_SOCKET_PREFIX) + std::string("nng_ipc_socket");
```
mlausinTT, line 21: Any way you can make this unique automatically? Unique by user for sure, but unique per test run would be ideal.

I see a potential issue if multiple users run same test, or same user runs multiple tests, it could end up with different host code targeting same simulator instance. 
As far as I can see, you assume here that user will always setup different NNG_SOCKET_NAME, and in multi user env, I fully expect that someone will setup that the same.

mlausinTT (approved):
Looks good to me!

joelsmithTT (approved):

device/simulation/tt_simulation_device.cpp
abhullar-tt: I suck at naming things so I leave that to you but I just figured we may not want to add `TT_METAL_...` inside of UMD?

device/simulation/tt_simulation_device.cpp
abhullar-tt: I guess if one changes we would need to rmr to update this as well, if they pull from a common function we don't have to do that

device/simulation/tt_simulation_device.cpp
vtangTT: that makes sense, I'll change the prefix to just `TT_*` or something similar

device/simulation/tt_simulation_device_generated.h
vtangTT: changed to `tt_vcs_core`

device/simulation/tt_simulation_host.cpp
vtangTT: Will now create unique socket addr that will be in the form of: 
`ipc:///tmp/<username>_<date/time>_<NNG_SOCKET_NAME>` 

where username is just `whoami` or `echo $USER` value on the machine.

eg:
`ipc:///tmp/vtang_08-02-16:00:34_deviceinit`

this should ensure nothing targets the same simulator instance as the date/time will be at seconds granularity

If `NNG_SOCKET_NAME` is not set, the default is `ipc:///tmp/<username>_<date/time>_nng_ipc`

device/device_api.h
vtangTT: removed

device/simulation/tt_simulation_device.cpp
vtangTT: changed to `TT_REMOTE_EXE`

device/simulation/tt_simulation_device.cpp
vtangTT: @joelsmithTT can I move this implementation as well these
```
tt_device_dram_address_params dram_address_params;
tt_device_l1_address_params l1_address_params;
tt_driver_host_address_params host_address_params;
tt_driver_eth_interface_params eth_interface_params;
```
into `tt_device.h` or should I leave this cleanup work for later?

joelsmithTT: Moving all to the base class would be an improvement.  It's up to you if you want to do it in this branch or in future work.  Maybe create an issue for it if you decide on the latter.

vtangTT: https://github.com/tenstorrent/tt-umd/issues/26

tests/simulation/test_simulation_device.cpp
```diff
@@ -0,0 +1,86 @@
+// SPDX-FileCopyrightText: (c) 2023 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include <random>
+#include "device_fixture.hpp"
+
+void generate_data(std::vector<uint32_t> &data, uint32_t size_in_bytes){
```
TT-billteng, line 8: why not have the function signature be 

`std::vector<uint32_t> data generate_data( uint32_t size_in_bytes )` 

?

tests/simulation/device_fixture.hpp
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    static void TearDownTestSuite() {
+        device->close_device();
+    }
+
+    static std::unique_ptr<tt_SimulationDevice> device;
+};
+
+std::unique_ptr<tt_SimulationDevice> SimulationDeviceFixture::device = nullptr;
```
TT-billteng, line 32: why is this a global?

device/simulation/tt_simulation_host.cpp
```diff
@@ -0,0 +1,79 @@
...
+    std::memcpy(msg, buf, buf_size);
+
+    rv = nng_send(host_socket, msg, buf_size, NNG_FLAG_ALLOC);
+    log_debug(tt::LogEmulationDriver, "Message sent.");
+    if (rv != 0) {
+        log_info(tt::LogEmulationDriver, "Failed to send message to remote: {}", nng_strerror(rv));
+    }
+}
+
+void tt_SimulationHost::recv_from_device(void **data_ptr, size_t *data_size) {
```
TT-billteng, line 71: why is `data_size` a pointer? this is odd

TT-billteng: Have you considered changing the function signature to:

`size_t tt_SimulationHost::recv_from_device(void **data_ptr)`

vtangTT: Passing a ptr since the nng api also takes a pointer `int nng_recv(nng_socket s, void *data, size_t *sizep, int flags);`

Should I just pass no ptr?

tests/simulation/test_simulation_device.cpp
vtangTT: RVO would prevent copying if it was `std::vector<uint32_t> wdata = generate_data(size_in_bytes)` right

tests/simulation/device_fixture.hpp
vtangTT: I made it global so that tests in that fixture will all run on one simulator instance.

CMakeLists.txt
```diff
@@ -6,49 +6,51 @@ set(CMAKE_CXX_STANDARD 17)
...
 endif()
 
+if(NOT CMAKE_BUILD_TYPE)
+    message(STATUS "Setting build type to 'Release' as none was specified.")
+    set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Release build is the default" FORCE)
+endif()
+
+message(STATUS "Build type: ${CMAKE_BUILD_TYPE}")
+set(CMAKE_CXX_FLAGS_RELEASE "-O3")
+set(CMAKE_CXX_FLAGS_DEBUG "-O0 -g -DDEBUG=DEBUG")
```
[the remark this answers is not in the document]
vtangTT: added

tests/simulation/test_simulation_device.cpp
vtangTT: Function signature changed to `std::vector<uint32_t> generate_data( uint32_t size_in_bytes )`

device/simulation/tt_simulation_host.cpp
vtangTT: Changed to `size_t tt_SimulationHost::recv_from_device(void **data_ptr)`

tests/simulation/test_simulation_device.cpp
TT-billteng: > RVO would prevent copying if it was `std::vector<uint32_t> wdata = generate_data(size_in_bytes)` right

yes it's a very simple function so RVO should kick in

tests/simulation/device_fixture.hpp
TT-billteng: is there a problem if each test owns its own sim instance? just perf?

TT-billteng (approved):
Great work!

tests/simulation/device_fixture.hpp
vtangTT: There's no problem, but wanted to test it like how multiple tests would run on same instance from metal level.

broskoTT:
This change broke BudaBackend build, since we don't use cmake in our build system. Please include us in the future on UMD PRs, while we're still supporting PyBuda

broskoTT:
> This change broke BudaBackend build, since we don't use cmake in our build system. Please include us in the future on UMD PRs, while we're still supporting PyBuda

@joelsmithTT @vtangTT This actually also broke the "make build" for UMD. Please update the readme on how UMD should be built going forward, or fix building without cmake.

vtangTT:
> > This change broke BudaBackend build, since we don't use cmake in our build system. Please include us in the future on UMD PRs, while we're still supporting PyBuda
> 
> @joelsmithTT @vtangTT This actually also broke the "make build" for UMD. Please update the readme on how UMD should be built going forward, or fix building without cmake.

Sorry about that, will fix ASAP.

PR #37, opened 2024-08-23
Add basic microbenchmark tests for device read/write

vtangTT:
Using nanobench: https://github.com/martinus/nanobench 

Example of results:
image

`elasped` column shows time taken for each read/write in seconds

### DIFF
--- a/tests/microbenchmark/test_rw_tensix.cpp
+++ b/tests/microbenchmark/test_rw_tensix.cpp
@@ -1,0 +1,94 @@
+
+#include <iostream>
+
+#include "nanobench.h"
+#include "device_fixture.hpp"
+
+std::uint32_t generate_random_address(std::uint32_t max, std::uint32_t min=0) {
+    ankerl::nanobench::Rng gen(80085);
+    std::uniform_int_distribution<> dis(min, max);  // between 0 and 1MB
+    return dis(gen);
+}
+
+TEST_F(uBenchmarkFixture, WriteAllCores32Bytes) {
+    std::vector<uint32_t> vector_to_write = {0, 1, 2, 3, 4, 5, 6, 7};
+    std::uint64_t address = l1_mem::address_map::DATA_BUFFER_SPACE_BASE;
+    std::uint64_t bad_address = 0x30000000;     // this address is not mapped, should trigger fallback write/read path
+
+    ankerl::nanobench::Bench bench_static;
+    ankerl::nanobench::Bench bench_dynamic;
+    for(auto& core : device->get_virtual_soc_descriptors().at(0).workers) {
+        std::stringstream wname;
+        wname << "Write to device core (" << core.x << ", " << core.y << ")";
+        // Write 32 bytes through static tlbs
+        bench_static.title("Write 32 bytes").unit("writes").minEpochIterations(50).output(nullptr).run(wname.str(), [&] {
+            device->write_to_device(vector_to_write, tt_cxy_pair(0, core), address, "SMALL_READ_WRITE_TLB");
+        });
+        // Write through "fallback/dynamic" tlb
+        bench_dynamic.title("Write 32 bytes fallback").unit("writes").minEpochIterations(50).output(nullptr).run(wname.str(), [&] {
+            device->write_to_device(vector_to_write, tt_cxy_pair(0, core), bad_address, "SMALL_READ_WRITE_TLB");
+        });
+        wname.clear();
+    }
+    bench_static.render(ankerl::nanobench::templates::csv(), results_csv);
+    bench_dynamic.render(ankerl::nanobench::templates::csv(), results_csv);
+}
+
+TEST_F(uBenchmarkFixture, ReadAllCores32Bytes){
+    std::vector<uint32_t> readback_vec = {};
+    std::uint64_t address = l1_mem::address_map::DATA_BUFFER_SPACE_BASE;
+    std::uint64_t bad_address = 0x30000000;     // this address is not mapped, should trigger fallback write/read path
+
+    ankerl::nanobench::Bench bench_static;
+    ankerl::nanobench::Bench bench_dynamic;
+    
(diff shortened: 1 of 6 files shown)


tests/microbenchmark/test_rw_tensix.cpp
```diff
@@ -0,0 +1,128 @@
...
+    std::unordered_map<std::string, std::int32_t> dynamic_tlb_config = {{"SMALL_READ_WRITE_TLB", 157}}; // Use both static and dynamic TLBs here
+    uint32_t num_host_mem_ch_per_mmio_device = 1;
+    std::shared_ptr<tt_SiliconDevice> device = std::make_shared<tt_SiliconDevice>(test_utils::GetAbsPath("tests/soc_descs/grayskull_10x12.yaml"), "", target_devices, num_host_mem_ch_per_mmio_device, dynamic_tlb_config, false, true);
+
+    for(int i = 0; i < target_devices.size(); i++) {
+        // Iterate over devices and only setup static TLBs for functional worker cores
+        auto& sdesc = device->get_virtual_soc_descriptors().at(i);
+        for(auto& core : sdesc.workers) {
+            // Statically mapping a 1MB TLB to this core, starting from address DATA_BUFFER_SPACE_BASE. 
+            device->configure_tlb(i, core, get_static_tlb_index(core), l1_mem::address_map::DATA_BUFFER_SPACE_BASE);
```
joelsmithTT, line 110: Not your change but I'll point it out anyway: `DATA_BUFFER_SPACE_BASE` and similar are all signed.  Even though it does not appear to be causing any problems, when copied (i.e. when someone adds a new constant) it can result in bugs like [this](https://yyz-gitlab.local.tenstorrent.com/tenstorrent/budabackend/-/issues/2604#note_320876).

tests/microbenchmark/test_rw_tensix.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+    auto get_static_tlb_index = [] (tt_xy_pair target) {
+        int flat_index = target.y * 10 + target.x;  // grid_size_x = 10 for GS/WH ????? something is wrong here
+        if (flat_index == 0) {
+            return -1;
+        }
+        return flat_index;
+    };
+    std::set<chip_id_t> target_devices = {0};
+    std::unordered_map<std::string, std::int32_t> dynamic_tlb_config = {{"SMALL_READ_WRITE_TLB", 157}}; // Use both static and dynamic TLBs here
```
joelsmithTT, line 101: If all of the (x,y) cores and addresses in the benchmarks have a TLB configured below (`device->configure_tlb...`) then the "fallback" path (aim TLB 157 at the core in question and read/write through its window) won't happen -- UMD will see that it has an existing mapping and use it.

If you want to benchmark the fallback path, I think you need both:
* Use an address that isn't covered by a static TLB window (so UMD takes the fallback path)
* "SMALL_READ_WRITE_TLB" passed in to the read/write methods (so UMD knows which TLB index to use)

Regarding address coverage: note that although the TLB config code below is using a non-zero address within the core (`l1_mem::address_map::DATA_BUFFER_SPACE_BASE` is 0x33000 for me -- building for WH), the TLB window itself is still aimed at the bottom of the Tensix address space (0x0) and access through the window is offset by UMD.  So UMD would use the same TLB window to access e.g. `DATA_BUFFER_SPACE_BASE-4`.

tests/microbenchmark/test_rw_tensix.cpp
```diff
@@ -0,0 +1,128 @@
...
+        rname.clear();
+    }
+    bench.render(ankerl::nanobench::templates::csv(), results_csv);
+}
+
+int main(){
+    std::ofstream results_csv("ubench_results.csv", std::ios_base::app);
+
+    auto get_static_tlb_index = [] (tt_xy_pair target) {
+        int flat_index = target.y * 10 + target.x;  // grid_size_x = 10 for GS/WH ????? something is wrong here
```
joelsmithTT, line 94: It would be really great if software that used UMD never had to worry about TLB indices.

tests/microbenchmark/test_rw_tensix.cpp
```diff
@@ -0,0 +1,128 @@
...
+        if (flat_index == 0) {
+            return -1;
+        }
+        return flat_index;
+    };
+    std::set<chip_id_t> target_devices = {0};
+    std::unordered_map<std::string, std::int32_t> dynamic_tlb_config = {{"SMALL_READ_WRITE_TLB", 157}}; // Use both static and dynamic TLBs here
+    uint32_t num_host_mem_ch_per_mmio_device = 1;
+    std::shared_ptr<tt_SiliconDevice> device = std::make_shared<tt_SiliconDevice>(test_utils::GetAbsPath("tests/soc_descs/grayskull_10x12.yaml"), "", target_devices, num_host_mem_ch_per_mmio_device, dynamic_tlb_config, false, true);
```
joelsmithTT, line 103: GS only for now?

Something to consider is if `ARCH_NAME` is set to something that isn't `grayskull`.  Longer term, getting rid of `ARCH_NAME` is a nice goal.

tests/microbenchmark/test_rw_tensix.cpp
vtangTT: Yep this makes sense and I have a note for it on my side. I have ubenchmarked the fallback tlb path and it is much slower than the static path. 

Right now, these are just basic tests but I'm planning on expanding them/making them more general in the future.

tests/microbenchmark/test_rw_tensix.cpp
vtangTT: Haha, so are you saying that everything in `l1_address_map.h` that is an `std::int32_t` should be a `std::uint32_t` ?

tests/microbenchmark/test_rw_tensix.cpp
vtangTT: test for fallback/dynamic read/write path should be added now

PR #42, opened 2024-09-03
Move atu disable to silicondevice destructor

broskoTT:
When I initially implemented ATU in https://github.com/tenstorrent/tt-umd/pull/13, I had the wrong impression that only a single TTDevice is allowed to be open. TTDevice are being opened multiple times when ran from Pybuda, furthermore tt_SiliconDevice can be opened multiple times.
I've now moved at which exact point is ATU disabled, to better reflect when it was even set up. 

This was tested with benchmark models on Pybuda.

This is a general fix for umd though, not Buda specific.

Buda still builds on main, hence the PR on main branch.

### DIFF
--- a/device/tt_silicon_driver.cpp
+++ b/device/tt_silicon_driver.cpp
@@ -232,15 +232,6 @@ private:
     TTDevice() = default;
 
     void reset() {
-        if (arch == tt::ARCH::BLACKHOLE && bar2_uc != nullptr && bar2_uc != MAP_FAILED) {
-            // Disable ATU index 0
-            // TODO: Implement disabling for all indexes, once more host channels are enabled.
-            uint64_t iatu_index = 0;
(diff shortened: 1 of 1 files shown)


abhullar-tt:
I know this PR was merged but do you think you can also push it to `metal-main`?

broskoTT:
> I know this PR was merged but do you think you can also push it to `metal-main`?

I can cherry pick the commit on that branch.
But isn't the idea to merge the metal-main into main anyway?

abhullar-tt:
> > I know this PR was merged but do you think you can also push it to `metal-main`?
> 
> I can cherry pick the commit on that branch. But isn't the idea to merge the metal-main into main anyway?

yeah I believe so, Im just not sure when this would happen

pjanevskiTT:
Maybe it is a good moment to make a specific BBE branch, and we can review metal-main branch if that is going to speed up the process. So we can make changes on BBE branch if those changes are Buda specific. If not, we will merge it to main in a standard way

PR #46, opened 2024-09-09
#12405: Refactor to use strong types

patrickroberts:
### Ticket
https://github.com/tenstorrent/tt-metal/issues/12405

### Problem description
Aliases of integral types are being used interchangeably with the integral types, allowing for erroneous implicit conversions.

### What's changed
Refactor to use enum class and make any intentional conversions explicit.

### DIFF
--- a/device/tt_cluster_descriptor_types.h
+++ b/device/tt_cluster_descriptor_types.h
@@ -4,24 +4,63 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-#pragma once 
+#pragma once
 
-#include <tuple>
+#include <fmt/format.h>
 
-using chip_id_t = int;
-using ethernet_channel_t = int;
-using eth_coord_t = std::tuple<int, int, int, int>;  // x, y, rack, shelf
+#include <functional>
+#include <type_traits>
 
-namespace std {
-template <>
-struct hash<eth_coord_t> {
-  std::size_t operator()(eth_coord_t const &c) const {
-    std::size_t seed = 0;
-    seed = std::hash<std::size_t>()(std::get<0>(c)) << 48 | 
-          std::hash<std::size_t>()(std::get<1>(c)) << 32 |
-          std::hash<std::size_t>()(std::get<2>(c)) << 16 |
-          std::hash<std::size_t>()(std::get<3>(c));
-    return seed;
-  }
+
+namespace tt::umd {
+
+enum class chip_id : int {
+    none = -1,
 };
-}
+
+enum class ethernet_channel : int {};
+
+struct eth_coord {
+    int x;
+    int y;
+    int rack;
+    int shelf;
+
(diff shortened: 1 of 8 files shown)


device/tt_cluster_descriptor_types.h
```diff
@@ -4,24 +4,63 @@
...
+    int y;
+    int rack;
+    int shelf;
+
+    // in C++20 this should be defined as:
+    // constexpr bool operator==(const eth_coord &other) const noexcept = default;
+    constexpr bool operator==(const eth_coord &other) const noexcept {
+        return (x == other.x and y == other.y and rack == other.rack and shelf == other.shelf);
+    }
+};
```
ayerofieiev-tt, line 34: Can establishing of a namespace for these structs come as another PR, where all code consistently use it from within that namespace and not global? 

This pattern is widespread in the codebase today and I highly discourage it

ayerofieiev-tt (commented):
overall lgtm 👍

device/tt_cluster_descriptor_types.h
patrickroberts: That's why the aliases are kept in place. I have not changed any existing usage of the aliases that were already there.

device/tt_cluster_descriptor_types.h
```diff
@@ -4,24 +4,63 @@
...
+
+}  // namespace tt::umd
+
+using chip_id_t = tt::umd::chip_id;
+using ethernet_channel_t = tt::umd::ethernet_channel;
+using eth_coord_t = tt::umd::eth_coord;
```
abhullar-tt, line 40: if this is only merged to `metal-main` then you can potentially get rid of these `*_t`?

device/tt_cluster_descriptor_types.h
```diff
@@ -4,24 +4,63 @@
...
-using chip_id_t = int;
-using ethernet_channel_t = int;
-using eth_coord_t = std::tuple<int, int, int, int>;  // x, y, rack, shelf
+#include <functional>
+#include <type_traits>
+
+
+namespace tt::umd {
+
+enum class chip_id : int {
```
broskoTT, line 17: Having this as an enum can be a bit confusing, since you're using it such that any int value can be assigned to it, and it doesn't necessarily have a corresponding enum constant. However, I do see the value in using an explicit name for a type to have a clearer requirement  and expectations over what is expected from an interface.

the eth_coord clearly benefits in readability from being defined as a struct, but whats the reasoning behind changing using chip_id_t to enum class?
You would then have define a constexpr int c_invalid_chip_id = -1;

broskoTT: Ah actually, I saw the attached tt-metal issue, this ensures explicit conversion. And enum class is a different thing than enum :) LGTM

PR #40, opened 2024-08-27
"New" TTDevice class

vtangTT:
This PR aims to separate/clean up tt_SiliconDevice by separating all functions that interact directly with the PCIe into the TTDevice class.

End goal: 
- something similar to `host_api.h` from metal, where the essential functions left in `tt_SiliconDevice` (`write_to_device`, `read_from_device`, tlb setup, etc.) will be exposed to TT-metal
- those functions will use TTDevice class to interact with the PCIe
- any helper functions will be moved to a separate util file

Next steps: 
- [ ] get rid of tt_SiliconDevice class and rename `tt_silicon_driver.cpp` and `tt_device.h` into something more appropriate (open to suggestions)
- [ ] split out helper functions to a diff file
- [ ] rename `pcie_device.cpp .hpp` into `tt_device.cpp .hpp`
- [ ] implement the necessary metal changes 

UMD post commit: https://github.com/tenstorrent/tt-metal/actions/runs/10636550474

PR #58, opened 2024-09-12
Add initial docs

### LINKED ISSUE #48: API documentation
tt-metal already has a tool which does this for their API.
Reuse the same one, and write documentation on our current API

pjanevskiTT:
Fixes #48 

Add initial setup on how we want to build docs

This PR is going to be in draft just as a discussion whether this is the way we want to build the docs. Doxygen is used and the format for the functions is same metal uses for host API, compute kernel API etc.

## TODOs

- [x]  Wait for #40 to know which API to document, is it still going to be `tt_device` class or `tt_umd.h` file -> Documenting tt_device
- [x]  Open follow up issue to add `sphinx` tools like in metal and build full page same as Metal docs -> Follow up #59 
- [x]  Add script to install requirements for doxygen


docs/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Requirements
+
+TODO
+
+## Build docs
+
+In order to build docs `TT_UMD_HOME` needs to be set to root of tt-umd project.
+
+After that you can run [`build_docs.sh`](build_docs.sh)
```
broskoTT, line 11: This is missing from the PR, right?

device/tt_device.h
```diff
@@ -219,9 +219,12 @@ class tt_device
...
     // Setup/Teardown Functions
     /**
-    * @brief Set L1 Address Map parameters used by UMD to communicate with the TT Device
-    * \param l1_address_params_ tt_device_l1_address_params encapsulating all the L1 parameters required by UMD
-    */ 
+     * Set L1 Address Map parameters used by UMD to communicate with the TT Device.
+     *
+     * | Argument            | Description                           | Type                        | Valid Range | Required |
+     * |---------------------|---------------------------------------|-----------------------------|-------------|----------|
+     * | l1_address_params_ | All the L1 parameters required by UMD. | tt_device_l1_address_params |             | Yes      |
```
broskoTT, line 226: nit: | mismatch :)

device/tt_device.h
```diff
@@ -55,7 +55,7 @@ struct tt_device_dram_address_params {
     std::uint32_t DRAM_BARRIER_BASE = 0;
```
broskoTT, line 55: tt_umd.hpp will be reverted (in that PR). You can safely document tt_device

docs/README.md
pjanevskiTT: Yeah there was gitignore line to ignore everythng that starts with build. I renamed it, changed README and pushed it

.gitignore
```diff
@@ -9,3 +9,4 @@ compile_commands.json
 .vscode/
```
broskoTT, line 9: While at it, please also change build* to build/

Doxyfile
```diff
@@ -0,0 +1,2781 @@
+# Doxyfile 1.9.6
```
broskoTT, line 1: Is this something that could also live in "docs" directory?

Doxyfile
```diff
@@ -0,0 +1,2781 @@
...
+# the logo to the output directory.
+
+PROJECT_LOGO           =
+
+# The OUTPUT_DIRECTORY tag is used to specify the (relative or absolute) path
+# into which the generated documentation will be written. If a relative path is
+# entered, it will be relative to the location where doxygen was started. If
+# left blank the current directory will be used.
+
+OUTPUT_DIRECTORY       = "docs/doxygen_build/"
```
broskoTT, line 71: Consider changing to build/docs.
Then also no addition to .gitignore

docs/README.md
```diff
@@ -0,0 +1,13 @@
+# TT-UMD Docs
+
+## Requirements
+
+TODO
```
broskoTT, line 5: Reminder to fill this up

docs/run_build_docs.sh
```diff
@@ -0,0 +1,9 @@
+if [[ -z "$TT_UMD_HOME" ]]; then
+  echo "Must provide TT_UMD_HOME in environment" 1>&2
+  exit 1
+fi
+
+echo "Building tt-umd docs..."
+
+cd $TT_UMD_HOME
```
broskoTT, line 8: consider changing so that users are left in the directory where they started from after the script. 
```
pushd $TT_UMD_HOME
doxygen Doxyfile
popd
```

device/tt_device.h
```diff
@@ -55,7 +55,7 @@ struct tt_device_dram_address_params {
...
 };
 /**
- * @brief Struct encapsulating all L1 Address Map parameters required by UMD.
+* Struct encapsulating all L1 Address Map parameters required by UMD.
```
broskoTT, line 58: This seems mismatched now, right? All the first stars should be in the same column. I think you need a space on this line, and on the */ line, but please check what is the standardized look of comments for doxygen.

Also in a couple of following comments

device/tt_device.h
```diff
@@ -232,170 +234,216 @@ class tt_device
...
 
     /**
-    * @brief Configure a TLB to point to a specific core and an address within that core. Should be done for Static TLBs
-    * \param logical_device_id Logical Device being targeted
-    * \param core The TLB will be programmed to point to this core
-    * \param tlb_index TLB id that will be programmed
-    * \param address All incoming transactions to the TLB will be routed to an address space starting at this parameter (after its aligned to the TLB size)
-    * \param ordering Ordering mode for the TLB. Can be Strict (ordered and blocking, since this waits for ack -> slow), Relaxed (ordered, but non blocking -> fast) or Posted (no ordering, non blocking -> fastest).
-    */ 
+     * Configure a TLB to point to a specific core and an address within that core. Should be done for Static TLBs.
```
broskoTT, line 259: Is this table some standard Doxygen way? Or is this how it is documented in tt_metal, and actually a preparation for sphinx?

I also wonder what would be the right way to have longer argument comments, this format caps them inconveniently.

broskoTT (approved):
Approved with suggestions

device/tt_device.h
joelsmithTT: Agree w/ @broskoTT.  The table format looks like it has a higher modification effort than e.g.
```
/**
 * @param p <some description> 
 * \param z <some description> (not sure if \ or @ is more Doxygen-y)
 */
```
and also seems like it will increase line length past 120 chars.

docs/run_build_docs.sh
joelsmithTT: I wonder if this can be integrated with the cmake build flow (opportunistically, so it only runs if Doxygen is available).

device/tt_device.h
pjanevskiTT: I was just following how tt-metal does it. I think it would be nice to have it in the same way, but don't really have strong opinions on it. Doxygen is ok with both formats that are specified, it is just a matter of what we choose. I will do whatever we agree on.

Just to point out, the format Joel specifed was being used, but it seemed very messy to me. It does give people more freedom to describe the param, but if too much text is needed to explain some param, maybe the API is broken or param is too complicated

As long as we put in the docs which format we want to follow I think we are good.
Maybe we can ask tt-metal folks as well if they want anything specific on the format

Doxyfile
pjanevskiTT: Following Metal on this as well, I can change it but maybe keeping it consistent across TT repos should be something to aim for

device/tt_device.h
pjanevskiTT: As per offline discussion, resolving comment, keeping compatibility with tt-metal for now

docs/run_build_docs.sh
pjanevskiTT: I think build system around docs should be setup when resolving issue #59, Sphinx is going to setup that for us. Right now we have better docs and a basic system to view it, which I think is enough as the initial effort for the docs

pjanevskiTT:
Per offline discussion, tt-umd is going to implement doxygen style. Since tt-metal has both table format and doxygen style comments, it can be changed to doxygen style in tt-metal to keep consistency. Multiple things in favour of doxygen style comments that were brought up
- Easier to edit
- clang ruins table format
- Broad audience more familiar with doxygen style

joelsmithTT (approved):
I did the doc build and am looking at the HTML output -- looks great!

Issue #55, opened 2024-09-10
Consistent error handling


broskoTT:
Per chat with Joel:

Consistent error handling.  UMD is using exceptions but I doubt anybody is catching them.  We also have a log_assert.  A pattern I've seen used successfully is to make macros for throwing an exception or terminating the program if it its a fatal error, so that the exact source location can be included in the exception what() or written to stderr.  UMD_THROW("That won't work") or UMD_FATAL("Grayskull is on fire").  On exceptions, if we're going to use them then I think we should be explicit in interface documentation what methods can throw and why.

vtangTT:
I feel like we should use the same error handling as metal does rn with `TT_FATAL`, `TT_ASSERT`, etc. 

There might be an opportunity to consolidate these utility functions as metal and umd both define their own rn in `common/assert.hpp` or `common/logger.hpp` and it gets confusing to tell which one people are including.

Maybe we can consider a third party utils repo that gets included in both metal and umd as a submodule so we don't have duplicate code?

broskoTT:
@vtangTT that makes sense. But the current issue is scoped only to make uniform error handling within this repo, I'd like to keep it like that for know.

I'd advise to create another issue for what you mentioned, although I'm not sure whether that should be tt-umd issue or tt-metal issue or something else. More of the problem is to actually create a new repo, then to move tt-metal to use it. I think it will be a minor task to make umd point to it as well.

pjanevskiTT:
I will close #139 since this issue has more details, @vtangTT can you take this over when you have time?

broskoTT:
Scope of this issue, figure out what should we use, which macro, should we stop using log_error or log_assert, etc.
Then make everything uniform in our repo.

aleksamarkovicTT:
Yes I agree. The only thing I'm not sure to use TT_THROW for is [out_of_range](https://github.com/tenstorrent/tt-umd/blob/bc14a3b5f3ae3b13092cfcdc1a596ee250d316ce/device/pcie/tlb_window.cpp#L139). What do you think?

pjanevskiTT:
Completely fine with it

pjanevskiTT:
I believe @aleksamarkovicTT solved this by UMD_THROW

PR #60, opened 2024-09-16
Replace snprintf and string concat with fmt::format

### LINKED ISSUE #54: Consistent string formatting
Per chat with Joel:

Consistent string formatting: we have fmt as a submodule.  I dislike fmt because it wrecks compile times, but everyone uses it.  There's a mix of string formatting techniques in UMD, both std::string::operator+ and snprintf.

pjanevskiTT:
Fixes #54

### Note

`hwloc` object has `snprintf` function for that specific type, which is builtin, so that is not changed in the PR

### DIFF
--- a/device/tt_cluster_descriptor.cpp
+++ b/device/tt_cluster_descriptor.cpp
@@ -12,6 +12,8 @@
 #include "common/logger.hpp"
 #include "yaml-cpp/yaml.h"
 
+#include "fmt/core.h"
+
 using namespace tt;
 bool tt_ClusterDescriptor::ethernet_core_has_active_ethernet_link(chip_id_t local_chip, ethernet_channel_t local_ethernet_channel) const {
     return this->ethernet_connections.find(local_chip) != this->ethernet_connections.end() &&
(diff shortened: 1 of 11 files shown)


device/tt_cluster_descriptor.cpp
```diff
@@ -297,7 +299,7 @@ std::unique_ptr<tt_ClusterDescriptor> tt_ClusterDescriptor::create_from_yaml(con
 
     std::ifstream fdesc(cluster_descriptor_file_path);
     if (fdesc.fail()) {
-        throw std::runtime_error("Error: cluster connectivity descriptor file " + cluster_descriptor_file_path + " does not exist!");
+        throw std::runtime_error(fmt::format("Error: cluster connectivity descriptor file {} does not exist!", cluster_descriptor_file_path));
```
joelsmithTT, line 302: I wonder if it makes sense for us to have a macro like `UMD_THROW` that throws a `std::exception`-derived UMD exception type and that applies formatting:

```
UMD_THROW("Cluster file {} doesn't exist!", path);
```

pjanevskiTT: Yeah it makes sense, we have an issue for that - https://github.com/tenstorrent/tt-umd/issues/55. It will be adressed in separate PR

joelsmithTT (approved):

Issue #63, opened 2024-09-17
Clarify interface for `get_register_address`


vtangTT:
As per comments from @joelsmithTT on https://github.com/tenstorrent/tt-umd/pull/40

I think the 'register' part of get_register_address is a bit misleading here.. if I understand the code correctly, it can end up writing to a register exposed in a PCIe BAR, or write through an inbound TLB window (to what might be either memory or a register).

It is probably worth making a distinction between these scenarios at the interface level.

Issue #62, opened 2024-09-17
Remove support for ancient KMD


vtangTT:
This change involves removing the need to open one device fd per host channel, see these functions/variables:
`device_fd_per_host_ch`
`open_hugepage_per_host_mem_ch`

Can check for a minimum KMD version somewhere to ensure nothing breaks.

Issue #61, opened 2024-09-17
Use `BAR0` for Wormhole instead of `system_reg_mapping`


vtangTT:
As per offline discussion with @joelsmithTT, `system_reg_mapping` is unneeded complication and we should be able to write directly through `BAR0` for wormhole

PR #64, opened 2024-09-17
"New" PCIDevice class 

vtangTT:
duplicate of https://github.com/tenstorrent/tt-umd/pull/40

Next steps/related issues:

- https://github.com/tenstorrent/tt-umd/issues/63
- https://github.com/tenstorrent/tt-umd/issues/62
- https://github.com/tenstorrent/tt-umd/issues/61 
- remove `tt_device` base class

CI: https://github.com/tenstorrent/tt-metal/actions/runs/10906490791
metal post-commit: https://github.com/tenstorrent/tt-metal/actions/runs/10946594194


device/pci_device.hpp
```diff
@@ -0,0 +1,102 @@
...
+static const uint32_t GS_WH_ARC_SCRATCH_6_OFFSET = 0x1FF30078;
+
+// See /vendor_ip/synopsys/052021/bh_pcie_ctl_gen5/export/configuration/DWC_pcie_ctl.h
+static const uint64_t UNROLL_ATU_OFFSET_BAR = 0x1200;
+
+// BAR0 size for Blackhole, used to determine whether write block should use BAR0 or BAR4
+const uint64_t BAR0_BH_SIZE = 512 * 1024 * 1024;
+
+// PCIe device IDs through ioctl
+static const uint16_t GS_PCIE_DEVICE_ID = 0xfaca;
```
vtangTT, line 27: @pjanevskiTT @broskoTT 
Constants for pcie device ids. Please let me know if these const are sufficient

broskoTT: It's fine

broskoTT:
As far as I see, you haven't changed the interface that tt-metal uses. What changes are needed on tt-metal in relation to this change?

vtangTT:
This PR is now workign with metal, I'm going to push unless there's any other problems @broskoTT

broskoTT (approved):
I think we'll end up organising stuff a bit different in the end, according to diagram https://docs.google.com/drawings/d/1-m1azdsBqMA0A6ATYRMfkhyeuOJuGCEI62N5a96LXj0/edit

But nevertheless, as Joel said, I think this is a step in the right direction in any case, so I'm happy for this to be merged. 
Thanks for verifying it still works with metal.

I see that you've changed CMakeLists to maintain the cmake flow, but not module.mk to maintain the "make" flow. I don't think it is a priority to fix this in this PR. I've created an issue to clarify the situation and delete redundant files/build system https://github.com/tenstorrent/tt-umd/issues/65

device/pcie/pci_device.cpp
```diff
@@ -0,0 +1,349 @@
...
+    if (bar4_wc != nullptr && byte_addr >= BAR0_BH_SIZE) {
+        byte_addr -= BAR0_BH_SIZE;
+        dest = reinterpret_cast<uint8_t *>(bar4_wc) + byte_addr;
+    }else {
+        dest = get_register_address<uint8_t>(byte_addr);
+    }
+
+    const void *src = reinterpret_cast<const void *>(buffer_addr);
+    memcpy(dest, src, num_bytes);
+// #ifndef DISABLE_ISSUE_3487_FIX
```
joelsmithTT, line 280: For WH, I think we need to handle this.
https://yyz-gitlab.local.tenstorrent.com/tenstorrent/syseng/-/issues/3487

device/pcie/pci_device.cpp
```diff
@@ -0,0 +1,349 @@
...
+        device_fd_per_host_ch.push_back(device_fd_for_host_mem);
+    }
+}
+
+tt::ARCH PCIDevice::get_arch() const {
+    return arch;
+}
+
+template<typename T>
+T* PCIDevice::get_register_address(std::uint32_t register_offset) {
```
joelsmithTT, line 253: `volatile T*`?

device/pcie/pci_device.cpp
```diff
@@ -0,0 +1,349 @@
...
+    // Right now, address can either be exposed PCIe BAR or a TLB window
```
joelsmithTT, line 254: TLB windows are in PCIe BAR0 (or BAR4, for BH)

device/tt_device.h
```diff
@@ -785,7 +786,7 @@ class tt_SiliconDevice: public tt_device
     std::set<chip_id_t> target_remote_chips = {};
     tt_SocDescriptor& get_soc_descriptor(chip_id_t chip_id);
     tt::ARCH arch_name;
-    std::map<chip_id_t, struct PCIdevice*> m_pci_device_map;    // Map of enabled pci devices
+    std::map<chip_id_t, std::unique_ptr<PCIDevice>> m_pci_device_map;    // Map of enabled pci devices
```
joelsmithTT, line 789: I wonder if the performance of `unordered_map` is any better here?

device/tt_silicon_driver.cpp
```diff
@@ -642,199 +233,93 @@ std::vector<chip_id_t> ttkmd_scan() {
...
+//     std::string busid = std::to_string(dev->pci_bus);
 
-int get_numa_node(TTDevice *dev) {
+//     dev->suspend_before_device_reset();
 
-    static const char pattern[] = "/sys/bus/pci/devices/%04x:%02x:%02x.%u/numa_node";
-    char buf[sizeof(pattern)];
-    std::snprintf(buf, sizeof(buf), pattern,
-    (unsigned int)dev->pci_domain, (unsigned int)dev->pci_bus, (unsigned int)dev->pci_device, (unsigned int)dev->pci_function);
+//     char *argv[3];
```
joelsmithTT, line 256: Suggestion, take it or leave it: for large blocks of code that we want to comment out but leave there for reference in the near future, suggest:

```
// leaving this in for now, TODO clean up
#if 0
<code>
#endif   
```

joelsmithTT (changes requested):
Can you take a look at https://yyz-gitlab.local.tenstorrent.com/tenstorrent/syseng/-/merge_requests/1149/diffs before merging and see if something equivalent needs to go in?

device/pcie/pci_device.cpp
vtangTT: The reason I didn't make this volatile is bc `write_block` and `read_block` are using it and then using resulting address in `memcpy`: 

```
src = get_register_address<uint8_t>(byte_addr);
memcpy(dest, src, num_bytes);
```

I don't think memcpy supports taking a volatile pointer, to do so I thikn we need a workaround like in `PCIDevice::read_regs` where they did:
```
uint32_t temp = *src++;
memcpy(dest++, &temp, sizeof(temp));```

Not sure if there's confusion where volatile is needed, not sure if its' needed for `write_block` or `read_block`?
Every other areas where `get_register_address` is used will alwyas assign the output to a volatile variable

device/pcie/pci_device.cpp
vtangTT: I'm unable to see the ticket right now. 

Regardless, I'm a bit confused about this compile option.. I assume it's arch specific to WH? 

Also looking at metal's cmake, the `DISABLE_ISSUE_3487_FIX` option hasn't been present since mid August it seems (`702bdf9`)

vtangTT: lol I'm blind, it's still in metal and only applied for GS.. the double negative is confusing so I will uncomment and change the #ifdef condition

device/pcie/pci_device.cpp
vtangTT: updated.

device/tt_device.h
vtangTT: I think it's being used basically as a look up table so performance for `unordered_map` is probably better. 

Changed.

device/tt_silicon_driver.cpp
vtangTT: Changed.

device/tt_device.h
broskoTT: I think that map is better for small containers (<1000), due to a large constant involved due to hashing. unordered_map generally has a bad standard implementation. But let's not think about it too much, either should be fast enough for now. And this PR is not about perf anyway

device/pcie/pci_device.cpp
joelsmithTT: The fix is for WH hardware.  It's likely that Metal is not doing anything to the chip that could trigger the issue.  Can you make sure that it's applied for the WH build?  We can revisit the conditional compilation removal later.

joelsmithTT (approved):

device/pcie/pci_device.cpp
vtangTT: Yep, I have verified it's only applied for the WH build.

PR #66, opened 2024-09-19
tt_mockup_device to be used for special profiling case with GraphCapture

mbezuljTT:
defining very basic tt_mockup_device that doesn't talk to any hardware or software. used to pass tt-metal initialization in order to run GraphCapture in the systems without cards, in order to profile memory usage with the offline compiler.

### DIFF
--- a/device/mockup/tt_mockup_device.hpp
+++ b/device/mockup/tt_mockup_device.hpp
@@ -1,0 +1,107 @@
+/*
+ * SPDX-FileCopyrightText: (c) 2023 Tenstorrent Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include <cstdint>
+#include <vector>
+
+#include "device/tt_cluster_descriptor.h"
+#include "device/tt_device.h"
+
+class tt_MockupDevice : public tt_device {
+   public:
+    tt_MockupDevice(const std::string& sdesc_path) : tt_device(sdesc_path) {
+        soc_descriptor_per_chip.emplace(0, tt_SocDescriptor(sdesc_path));
+        std::set<chip_id_t> target_devices = {0};
+    }
+    virtual ~tt_MockupDevice() {}
+
+    // Setup/Teardown Functions
+    virtual std::unordered_map<chip_id_t, tt_SocDescriptor>& get_virtual_soc_descriptors() override {
+        return soc_descriptor_per_chip;
+    }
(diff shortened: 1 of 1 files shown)


device/mockup/tt_mockup_device.hpp
```diff
@@ -0,0 +1,107 @@
...
+    virtual std::unordered_map<chip_id_t, tt_SocDescriptor>& get_virtual_soc_descriptors() override {
+        return soc_descriptor_per_chip;
+    }
+    virtual void set_device_l1_address_params(const tt_device_l1_address_params& l1_address_params_) override{};
+    virtual void set_device_dram_address_params(const tt_device_dram_address_params& dram_address_params_) override{};
+    virtual void set_driver_host_address_params(const tt_driver_host_address_params& host_address_params_) override{};
+    virtual void set_driver_eth_interface_params(
+        const tt_driver_eth_interface_params& eth_interface_params_) override{};
+    virtual void start_device(const tt_device_params& device_params) override {}
+    virtual void assert_risc_reset() override{};
```
pjanevskiTT, line 33: nit: change `override{};` to `override {}`, like for `start_device` function. Diff is semicolon (it is not needed), and space between `override` and `{}`

pjanevskiTT (approved):

device/mockup/tt_mockup_device.hpp
broskoTT: I hope that automatic linter will solve this in the future

broskoTT (approved):

device/mockup/tt_mockup_device.hpp
```diff
@@ -0,0 +1,107 @@
...
+        soc_descriptor_per_chip.emplace(0, tt_SocDescriptor(sdesc_path));
+        std::set<chip_id_t> target_devices = {0};
+    }
+    virtual ~tt_MockupDevice() {}
+
+    // Setup/Teardown Functions
+    virtual std::unordered_map<chip_id_t, tt_SocDescriptor>& get_virtual_soc_descriptors() override {
+        return soc_descriptor_per_chip;
+    }
+    virtual void set_device_l1_address_params(const tt_device_l1_address_params& l1_address_params_) override{};
```
broskoTT, line 27: nit: virtual is not needed when your implementation is final. So no "virtual" needed in this whole file.

device/mockup/tt_mockup_device.hpp
mbezuljTT: thanks - will do. 

btw. this was with automatic linter from tt-metal 😅

PR #69, opened 2024-09-23
Remove RD/WR device/sysmem API with vector argument

### LINKED ISSUE #67: Remove read/write methods on tt_device that accept std::vector<uint32_t>
There are read/write methods on the `tt_device` interface that accept `std::vector<uint32_t>`.  These could be removed (software can favor the equivalent methods that accept `void *` and `size_t`.

pjanevskiTT:
Fixes #67

Remove API for read and write to device/sysmem which has `std::vector ` as an argument. Use only functions that have `void*` as an argument. Change calls inside `tt_silicon_driver` which were using `std::vector` variant

TODO

- [x] Investigate whether tt-metal was using API with `std::vector` arguement. Change this if needed before bumping UMD -> tt-metal was just using `void*` interface

### DIFF
--- a/device/tt_device.h
+++ b/device/tt_device.h
@@ -366,20 +366,9 @@ class tt_device
         // Only implement this for Silicon Backend
         throw std::runtime_error("---- tt_device::write_to_device is not implemented\n");
     }
+
     virtual void broadcast_write_to_cluster(const void *mem_ptr, uint32_t size_in_bytes, uint64_t address, const std::set<chip_id_t>& chips_to_exclude,  std::set<uint32_t>& rows_to_exclude,  std::set<uint32_t>& columns_to_exclude, const std::string& fallback_tlb) {
         throw std::runtime_error("---- tt_device::broadcast_write_to_cluster is not implemented\n");
-    }
-
-    /**
-     * Write uint32_t vector to specified device, core and address (defined for Silicon).
-     *
-     * @param vec Data to write.
-     * @param core Chip and core being targeted.
-     * @param addr Address to write to.
-     * @param tlb_to_use Specifies fallback/dynamic TLB to use.
-     */
-    virtual void write_to_device(std::vector<uint32_t> &vec, tt_cxy_pair core, uint64_t addr, const std::string& tlb_to_use) {
-        throw std::runtime_error("---- tt_device::write_to_device is not implemented\n");
     }
 
     /**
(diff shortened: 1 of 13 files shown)


broskoTT:
What's the exact motivation for removing this? This seems just like another entrypoint, which can be convenient at some places (e.g. just pass a vector instead of vector.data() + vector.size())

device/tt_device.h
```diff
@@ -353,16 +353,6 @@ class tt_device
     virtual void broadcast_write_to_cluster(const void *mem_ptr, uint32_t size_in_bytes, uint64_t address, const std::set<chip_id_t>& chips_to_exclude,  std::set<uint32_t>& rows_to_exclude,  std::set<uint32_t>& columns_to_exclude, const std::string& fallback_tlb) {
         throw std::runtime_error("---- tt_device::broadcast_write_to_cluster is not implemented\n");
     }
-    /**
```
broskoTT, line 356: What's the exact motivation for removing this? This seems just like another entrypoint, which can be convenient at some places (e.g. just pass a vector instead of vector.data() + vector.size())

pjanevskiTT: tt-metal is not using the API that uses `std::vector`, but UMD is using the function inside `tt_silicon_driver.cpp`. However, whenever you use the one with vector, you can use the one with `void*`. Tests are using the variant with `std::vector`, which doesn't make a lot of sense if we really want to test the API that tt-metal uses.

broskoTT: Seems like this is adding ambiguity rather than clarity, so I agree. Thanks both for explanation.

broskoTT (approved):

pjanevskiTT:
@joelsmithTT take a look if you can/want, it is ready to merge

device/tt_silicon_driver.cpp
```diff
@@ -2243,7 +2243,8 @@ void tt_SiliconDevice::read_from_non_mmio_device(void* mem_ptr, tt_cxy_pair core
             } else {
                 // Read 4 byte aligned block from device/sysmem
                 if (use_dram) {
-                    read_from_sysmem(data_block, host_dram_block_addr, host_dram_channel, block_size, mmio_capable_chip_logical);
+                    size_buffer_to_capacity(data_block, block_size);
```
joelsmithTT, line 2246: Somewhat unrelated to your change, but pointing it out as something I noticed --

Is `data_block` necessary, or could sysmem/device memory be copied to from the source pointer (`mem_ptr`) directly?  I created https://github.com/tenstorrent/tt-umd/issues/90

joelsmithTT (approved):

device/tt_silicon_driver.cpp
pjanevskiTT: Thanks for pointing it out, @broskoTT and I are working on prototype of the new API on the branch https://github.com/tenstorrent/tt-umd/tree/brosko/new_api_experiment, so it will require rethinking remote API. That will be a good chance to take a look at this stuff

Issue #45, opened 2024-09-06
Work on adding CI to umd repo


broskoTT:
Nothing to add for know, it should be pretty clear. Guard the UMD with tests

vtangTT:
Just FYI, as a stopgap solution for metal, we are running most/all UMD tests on the metal CI. They run for every push to main and every post-commit

https://github.com/tenstorrent/tt-metal/actions/workflows/umd-unit-tests-wrapper.yaml

broskoTT:
Every push to tt-metal main, or umd main?
In any case we should have a CI here if we plan to ship this as a separate component/endpoint to our products.

vtangTT:
Every push to tt-metal main will run post-commit which includes umd tests. Example run from most recent main:

https://github.com/tenstorrent/tt-metal/actions/runs/10794191678/job/29937907086

broskoTT:
That seems unnecessary, since not every commit on tt-metal will update the UMD commit used. Sounds like you're running same tests on the same version of umd?
Also, this doesn't offer any guarantees for UMDs progress, i.e. this doesn't guarantee the current tip of the branch is functional.

vtangTT:
Yes, I agree it's not the best solution and like you said it only guarantees umd is not broken for the version metal points to. 

It was mainly implemented to have some sort of verification for UMD (since at that time we aren't able to get CI machines onto the UMD repo). Also on the off chance that metal exposes an edge-case bug in umd. 

It also allowed us to dispatch umd tests on CI through this pipeline: https://github.com/tenstorrent/tt-metal/actions/workflows/umd-unit-tests-wrapper.yaml

The tests themselves run fast enough such that they don't affect the total runtime of post-commit. If development is ramping up on UMD, CI should def be present in this repo, but temporarily it's possible to validate through metal CI if needed.

PR #70, opened 2024-09-23
Add CI for building and testing UMD

broskoTT:
Implements #45 

Adding CI jobs guarding UMD. This PR covers a couple of stuff:
- I did everything for both ubuntu 20 and 22
- There will be a job for building docker images and publishing them as artefacts of tenstorrent organization on github. This flow was already used, and you can see dockers here: https://github.com/orgs/tenstorrent/packages?q=tt-mlir&tab=packages&q=tt-umd
- I've created a "On PR" workflow, which will be mandatory check on checkins to main. It will make sure that umd_device lib can build
- I've created a "On PR - Optional" workflow, which won't be mandatory for now. This will build tests and run them on wormhole machines. Grayskull machines still don't work at the moment, and blackhole are not available. I'd be hesitant to put this as mandatory since our machines might not be stable. Once github-ci team switches to pools of machines, this should be better, and I'd switch it to mandatory.
- All of these workflows will be able to be ran manually, once they land in main, using "Actions" tab.
- I've switched to clang 17, but since this is not tied to CI change, this is in separate PR #74. I've tested that the dockers provided were able to build using clang17 as well.
- Fixed an issue (which shows on ubuntu22) with __FILE__ where calling create-ethernet-map from our tests doesn't work due to file resolving to absolute path.


.github/workflows/build-image.yml
```diff
@@ -0,0 +1,66 @@
...
+        uses: docker/build-push-action@v6
+        with:
+          context: .github
+          file: .github/${{ matrix.build.name }}.Dockerfile
+          push: true
+          build-args: |
+            GIT_SHA=${{ github.sha }}
+          tags: |
+            ${{ env.CI_IMAGE_NAME}}:${{ github.sha }}
+
```
vmilosevic, line 54: If you have just one image you can set both tags at once. This was split in another project because it was creating multiple images, and I wanted to make sure they could all be built before setting the latest tag.

```
tags: |
            ${{ env.CI_IMAGE_NAME}}:${{ github.sha }}
            ${{ env.CI_IMAGE_NAME}}:latest
```

.github/ubuntu-22.04.Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM ubuntu:22.04
```
vmilosevic, line 1: If your dependency list is the same you can have one Dockerfile and pass in ubuntu version as an arg.
It is also ok to separate them like this if you want to change dependencies depending on ubuntu verison

vmilosevic (approved):

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,54 @@
...
+      arch: ${{ matrix.test-group.arch }}
+
+  test-all:
+    secrets: inherit
+    needs: build-tests
+    strategy:
+      fail-fast: false
+      matrix:
+        test-group: [
+          # { arch: grayskull, card: e75 },
```
pjanevskiTT, line 44: Either add a comment why these lines are under comment or open the issue to add grayskull tests when grayskull machines are available

tests/test_utils/generate_cluster_desc.hpp
```diff
@@ -13,10 +13,19 @@
...
 
 inline std::string GetAbsPath(std::string path_){
-    std::filesystem::path umd_root_relative = std::filesystem::relative(std::filesystem::path(__FILE__).parent_path().parent_path().parent_path(), "../");
-    std::filesystem::path umd_root = std::filesystem::canonical(umd_root_relative);
-    std::filesystem::path abs_path = umd_root / path_;
-    return abs_path.string();
+    // Note that __FILE__ might be resolved at compile time to an absolute or relative address, depending on the compiler.
+    std::filesystem::path current_file_path = std::filesystem::path(__FILE__);
+    if (current_file_path.is_absolute()) {
+        std::filesystem::path umd_root_relative = std::filesystem::relative(std::filesystem::path(__FILE__).parent_path().parent_path().parent_path(), "../");
```
pjanevskiTT, line 19: nit: maybe you can move declarations of umd_root_relative, umd_root and abs_path before if, since all variables are used in all branches of if else. You can move abs_path.string() to the end of the function as well.

pjanevskiTT (approved):
Looks great, left few nit comments

.github/ubuntu-22.04.Dockerfile
broskoTT: I've created an .sh script which will hold common setup code. So I've left it in two separate Dockerfiles in case we need to add some setup specific for one or another version.

.github/workflows/on-push.yml
broskoTT: Added

PR #71, opened 2024-09-23
Create test-runner.yaml

broskoTT:
Related to #45 

@vmilosevic asked me to create this test runner workflow so that we can test newly added machines for CI.

### DIFF
--- a/.github/workflows/test-runner.yaml
+++ b/.github/workflows/test-runner.yaml
@@ -1,0 +1,103 @@
+name: Check runner
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+
+jobs:
+  
+  check-runners-host:
+    timeout-minutes: 10
+    strategy:
+      matrix:
+        build: [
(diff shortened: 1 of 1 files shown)


broskoTT:
This is the latest run: https://github.com/tenstorrent/tt-umd/actions/runs/11015038089
You can't manually schedule it since it is not present on main branch. I scheduled it by adding "push:", but then removed it in the latest commit.
As you can see some runners don't work, which we're trying to fix, but it should be a gate for this PR.

vtangTT:
LGTM, we could probably take a lot of the infra that metal already has in it's yaml files.

vtangTT (approved):

joelsmithTT (approved):

broskoTT:
> LGTM, we could probably take a lot of the infra that metal already has in it's yaml files.

I looked at them while writing #70 , along with current tt-mlir setup. For now, I don't think we need a lot of infra.

PR #74, opened 2024-09-24
Change compiler to clang

broskoTT:
Related to #45 
tt-metal uses clang compiler.
This configuration was copied from metal's CMakeLists.txt.
The irds that we're using for tt-metal already have this setup, so it shouldn't disrupt anyone.
I've already installed clang-17 in #70, so CI will still work with this switch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,8 +1,27 @@
 cmake_minimum_required(VERSION 3.16)
-project(umd_device)
+
+include(cmake/compilers.cmake)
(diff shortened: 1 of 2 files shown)


joelsmithTT (approved):

pjanevskiTT (approved):

broskoTT:
I'll check this in after #70 , to have the CI confirm this change works.

vtangTT (approved):

PR #75, opened 2024-09-24
Separate host arch specific code to directories

### LINKED ISSUE #56: Separate arch specific code to directories
Separate arch specific code into separate directories. This way we achieve two things
- Avoid conditional compilation (#51) -> we can have build target for each arch
- Code organization looks much better

For example, for something like https://github.com/tenstorrent/tt-umd/blob/main/device/driver_atomics.h, we should have something like

```
arch
|----- x86
|       |------ driver_atomics.cpp
|
|----- arm
|       |------ driver_atomics.cpp
```

Maybe driver atomics file is not the right granularity to do this, maybe we need to do it in smaller/bigger chunks, but that should be investigated through this issue as well.

pjanevskiTT:
Fixes #56 

`driver_atomics` files separated into specific directories for each possible host arch that we want to support for now

TODO

- [ ]  tt-metal post commit
- [ ]  Proper place for these files. Currently these are in `device/arch`. Do we need to move it elsewhere, for example root `host_arch` directory or something like that


CMakeLists.txt
```diff
@@ -78,6 +78,17 @@ include(${PROJECT_SOURCE_DIR}/cmake/dependencies.cmake)
 add_library(umd_common_directories INTERFACE)
 target_include_directories(umd_common_directories INTERFACE ${PROJECT_SOURCE_DIR} ${PROJECT_SOURCE_DIR}/device ${PROJECT_SOURCE_DIR}/third_party/fmt/include)
 
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "i386")
```
broskoTT, line 81: Per offline discussion, include .cpp sources by host arch instead of headers, that way interface to metal will be the same.

The drawback, though, will be that the compiler won't be able to remove function call in that case. Wondering if that might even be a problem... According to github copilot, this shouldn't be a problem.

It could be a perf hit though.
In anycase, these should be in a separate utils repo. Started an issue here: #78

pjanevskiTT: These function cannot be inline anymore since the implementation will move to .cpp files. Are we ok with this? 

If we want to keep it inline, we could just guard include headers with defines like 

```
#if defined (__x86_64__)
#include "x86/driver_atomics.h"
#endif

// empty file for device/driver_atomics.h
```

and move the implementation to separate .h files based on the architecture. This way we don't need to change cmake. Next step could be to add dependency on host arch to cmake (or investigate if there already is some dependency) in order to completely remove `device/driver_atomics.h`

joelsmithTT:
The conditional compilation for x86 vs arm vs riscv is fine as it's not something that the build system or user needs to worry about.  I am not convinced that the additional complexity this introduces (multiple files, additional build logic) is worth the benefit.

PR #87, opened 2024-09-27
#86: Remove ARCH_NAME dependency for libdevice.so

blozano-tt:
Closes #86 

Updated CI workflows to pass `-DTT_UMD_BUILD_TESTS=ON`

Issue #86, opened 2024-09-27
Remove ARCH_NAME dependency for libdevice.so target


blozano-tt:
Currently ARCH_NAME must be set to build libdevice.so

By using a cmake option to control building of the tests, we can conditionally check ARCH_NAME only when building the tests.

Issue #51, opened 2024-09-10
Remove conditional compilation


broskoTT:
Examine uses of conditional compilation (e.g. ARCH_NAME, ISSUE_123_DISABLE_FIX, etc.) and eliminate where possible.

broskoTT:
Conditional compilation for ARCH_NAME was removed by @tt-vjovanovic a while back. But it still remains in CMakeLists.
Make sure to remove it as part of this task.

joelsmithTT:
The `DISABLE_ISSUE_3487_FIX` define can be made a runtime thing.  Here's additional context for that issue:
https://yyz-gitlab.local.tenstorrent.com/tenstorrent/syseng/-/issues/3487
https://yyz-gitlab.local.tenstorrent.com/tenstorrent/syseng/-/merge_requests/1149/diffs

broskoTT:
Is this issue now complete? I do see some isolated cases of ifdefs, but looks like the two top offenders are now resolved? What do you think @joelsmithTT @pjanevskiTT

pjanevskiTT:
As for as umd_device target, I think it is complete. However, we still have tests. I am working on removing soc descriptor from tt_SiliconDevice constructor, after that we can have unified compilation for tests as well

joelsmithTT:
The code part is more or less complete (no more ARCH_NAME in source files) but as @pjanevskiTT points out there is still a bit more work to do for the tests.  (ARCH_NAME still shows up in tests/CMakeLists.txt and controls include paths).

blozano-tt:
Speaking of tests:

- I think we still need to get rid of ARCH_NAME ENV

I think we have a few options, but the easiest that is coming to mind is some CMake magic to generate 3x the number of executables. 

i.e. a custom add_executable

umd_add_exectuable(target_name arch_name)

Use the above in a for each loop.

cc: @afuller-TT

blozano-tt:
```
function(umd_add_executable target arch interface_lib)
    # Create the executable target
    add_executable(${target})

    # Link the interface library to the new target
    target_link_libraries(${target} PRIVATE ${interface_lib})

    # Set the compile definitions based on the architecture
    if(${arch} STREQUAL "ARCH1")
        target_compile_definitions(${target} PRIVATE DEFINE_ARCH1)
        target_include_directories(${target} PRIVATE ${CMAKE_SOURCE_DIR}/path/to/include_arch1)
    elseif(${arch} STREQUAL "ARCH2")
[...]

Issue #83, opened 2024-09-27
Update README.md


blozano-tt:
- What does UMD stand for?
- What does it provide its customers?
- Remove deprecated Makefile documentation

broskoTT:
Some of the answers to high level goals of UMD can be found here: https://docs.google.com/drawings/d/1-m1azdsBqMA0A6ATYRMfkhyeuOJuGCEI62N5a96LXj0/edit

PR #88, opened 2024-09-30
No ARCH_NAME in CI building umd_device

broskoTT:
A continuation to #87 , to reflect the current state in our CI.
Related to #86 and a part of #51 .

There is now enough difference between building umd_device and umd_tests to warrant separate files, since umd_tests still requires ARCH_NAME and umd_device does not. This way it looks clearer.


pjanevskiTT (approved):
It would be nice to update README accordingly along with these PRs that removed ARCH_NAME from the build. There is already an open issue about updating README - #83, but I think we should at least have build steps up to date

PR #92, opened 2024-10-01
[wip] Work on IO dir

pjanevskiTT:
No need to review, working on making the new API clear, PR is step in that direction

### DIFF
--- a/ideal_device/io/abstract_io.h
+++ b/ideal_device/io/abstract_io.h
@@ -12,16 +12,17 @@ namespace tt::umd {
 // This abstracts all readers and writers.
 // A user can hold this object to bypass the need to go through regular interfaces and 
 class AbstractIO {
-    public:
-    void write_u8(uint32_t address, uint8_t value);
-    void write_u32(uint32_t address, uint32_t value);
-    void write(uint32_t address, uint8_t* arr_ptr, uint32_t size);
-    void write(uint32_t address, uint32_t* arr_ptr, uint32_t size);
 
-    uint8_t read_u8(uint32_t address);
-    uint32_t read_u32(uint32_t address);
-    void read(uint32_t address, uint8_t* arr_ptr, uint32_t size);
-    void read(uint32_t address, uint32_t* arr_ptr, uint32_t size);
+public:
+    virtual void write_u8(uint32_t address, uint8_t value) = 0;
+    virtual void write_u32(uint32_t address, uint32_t value) = 0;
+    virtual void write(uint32_t address, uint8_t* arr_ptr, uint32_t size) = 0;
+    virtual void write(uint32_t address, uint32_t* arr_ptr, uint32_t size) = 0;
+
+    virtual read_u8(uint32_t address) = 0;
+    virtual uint32_t read_u32(uint32_t address) = 0;
+    virtual void read(uint32_t address, uint8_t* arr_ptr, uint32_t size) = 0;
+    virtual void read(uint32_t address, uint32_t* arr_ptr, uint32_t size) = 0;
 };
 
 }
--- a/ideal_device/io/tlb_manager.h
+++ b/ideal_device/io/tlb_manager.h
@@ -15,19 +15,22 @@ Questions:
 
 */
 
-#include "pci_device.h"
+#include "tt_device.h"
 #include "tlb_window.h"
 
 namespace tt::umd {
+
+typedef enum TLBType {
+    WC,
+    UC
+} TLBType;
 
 class TLBManager {
 
 public:
     virtual TLBWindow get_tlb_window(tt_xy_pair core) = 0;
     
-    virtual TLBWindow get_wc_tlb_window() = 0;
-
-    virtual TLBWindow get_uc_tlb_window() = 0;
+    virtual TLBWindow get_tlb_window(TLBType type) = 0;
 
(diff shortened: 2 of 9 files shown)


ideal_device/io/tlb_manager.h
```diff
@@ -0,0 +1,183 @@
...
+namespace tt::umd {
+
+class TLBManager {
+
+public:
+    virtual TLBWindow get_tlb_window(tt_xy_pair core) = 0;
+    
+    virtual TLBWindow get_wc_tlb_window() = 0;
+
+    virtual TLBWindow get_uc_tlb_window() = 0;
```
broskoTT, line 30: uc vs wc can be a parameter of the same function

ideal_device/io/tlb_manager.h
```diff
@@ -0,0 +1,183 @@
...
+
+// This should represent how we do things today. We have whole BAR0/4 space
+// and just map the TLB we need.
+class StandardTLBManager : public TLBManager {
+
+public:
+    StandardTLBManager(PCIDevice* pci_device) : pci_device(pci_device) {}
+
+    TLBWindow get_tlb_window(tt_xy_pair core) override {
+        uint32_t tlb_index = map_core_to_tlb_index(core);
```
broskoTT, line 60: So this is actually pci_device-> map_core_to_tlb_index
This is behind which call the architecture specific stuff is hidden (like what exact bar address is used to configure tlbs)?

ideal_device/io/tlb_window.h
```diff
@@ -0,0 +1,67 @@
+/*
+ * SPDX-FileCopyrightText: (c) 2023 Tenstorrent Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+
+namespace tt::umd {
+
+class TLBWindow {
```
broskoTT, line 10: This class seems like is redundant with the current interface.
It could be reduced to struct tlb_info {void ptr, int size}.

read and write implementation can be part of core_io.h

and write8/write16 read8/read16 and others (not _block functions) can be implemented in abstract_io directly (so that the specializations actually just implement write_block and read_block

ideal_device/io/tlb_manager.h
```diff
@@ -0,0 +1,183 @@
...
+        // We need to get DRAM core from DRAM channel.
+        return get_wc_tlb_window(tt_xy_pair(0, 0));
+    }
+
+    void release_dram_tlb_window(TLBWindow& tlb_window) override {
+        release_tlb_window(tlb_window);
+    }
+
+private:
+    PCIDevice* pci_device;
```
broskoTT, line 117: Not sure if intentional, due to my renamings. But the reference should be to ttdevice, right?
We could have TTDevice->get_tlb_manager(managertype) , and then use it to do management. Not sure if then the user would get the ownership as well (gets unique ptr), and that resulting manager has pointer to trdevice so it can use its interface which covers arch specific code

pjanevskiTT: Probably it should be `TTDevice`, because we can get `PCIDevice` from `TTDevice` if needed.  There is a question on how do we manage TLBs. If we don't program TLBs in advance, the question is where are we going to program the TLBs. For programming the TLB we need bar0 and bar4 pointers, which are exposed through PCIDevice. `TLBManager` seems like a good place to do it. Generally, `TLBManager` logic could go to `TTDevice`, but `TTDevice` has `PCIDevice` in it, so to me it looks the same. `TLBManager` was just used to encapsulate the logic for managing TLBs.

ideal_device/io/tlb_manager.h
pjanevskiTT: agree

ideal_device/io/tlb_manager.h
pjanevskiTT: Yes, makes sense to have the mapping in TTDevice or PCIDevice, wherever our soc descriptor is going to live

ideal_device/io/abstract_io.h
```diff
@@ -12,16 +12,17 @@ namespace tt::umd {
...
-    void write_u8(uint32_t address, uint8_t value);
-    void write_u32(uint32_t address, uint32_t value);
-    void write(uint32_t address, uint8_t* arr_ptr, uint32_t size);
-    void write(uint32_t address, uint32_t* arr_ptr, uint32_t size);
 
-    uint8_t read_u8(uint32_t address);
-    uint32_t read_u32(uint32_t address);
-    void read(uint32_t address, uint8_t* arr_ptr, uint32_t size);
-    void read(uint32_t address, uint32_t* arr_ptr, uint32_t size);
+public:
```
joelsmithTT, line 16: I suggest making addresses `uint64_t` and sizes `size_t`.  There are address spaces in Blackhole where > 32 bits are necessary.

```
    virtual void write(uint32_t address, uint8_t* arr_ptr, uint32_t size) = 0;
    virtual void write(uint32_t address, uint32_t* arr_ptr, uint32_t size) = 0;
```
I think this would be less ambiguous as a single method that takes a `const void*`.  From those two signatures, it is not clear if `size` means "number of elements" or "bytes".

broskoTT: Makes sense, void* is less ambiguous.

PR #113, opened 2024-10-03
Remove compile time check of DISABLE_ISSUE_FIX_3487 in favor of runtime check of arch

blozano-tt:
To aid in the goal of #51 

Change one compile time check to a runtime check.

I also thought it might be worth inlining the custom memcpy functions.

### DIFF
--- a/device/pcie/pci_device.cpp
+++ b/device/pcie/pci_device.cpp
@@ -5,6 +5,7 @@
  */
 
 #include <cstdint>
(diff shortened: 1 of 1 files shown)


joelsmithTT (approved):

broskoTT:
Note that inlining probably has no effect. Compiler likely knows better than the developers inlining effects to performance.
Thanks for this change though!

PR #114, opened 2024-10-03
Add a sysmem/DMA test for Wormhole

joelsmithTT:


tests/wormhole/test_silicon_driver_wh.cpp
```diff
@@ -88,6 +98,7 @@ TEST(SiliconDriverWH, CreateDestroy) {
     tt_device_params default_params;
     // Initialize the driver with a 1x1 descriptor and explictly do not perform harvesting
     for(int i = 0; i < 50; i++) {
+        std::cout << "Iteration " << i << std::endl;
```
pjanevskiTT, line 101: Remove print

tests/wormhole/test_silicon_driver_wh.cpp
```diff
@@ -644,3 +655,85 @@ TEST(SiliconDriverWH, VirtualCoordinateBroadcast) {
...
+ * This uses a small size for speed purposes.
+ *
+ * If/when we move to using IOMMU to map userspace memory for device access,
+ * the technique below is a straightforward way to test that hardware can access
+ * the buffer(s).
+ */
+TEST(SiliconDriverWH, SysmemTestWithPcie) {
+    auto target_devices = get_target_devices();
+
+    // Why is this required by the tt_SiliconDevice constructor?!
```
pjanevskiTT, line 684: Open the issue for this, even when we remove tt_SiliconDevice, it would be nice to have the issue to know what are the bad things about this design to avoid it

tests/wormhole/test_silicon_driver_wh.cpp
```diff
@@ -644,3 +655,85 @@ TEST(SiliconDriverWH, VirtualCoordinateBroadcast) {
...
+    tt_SiliconDevice device(test_utils::GetAbsPath("tests/soc_descs/wormhole_b0_8x10.yaml"),
+                            test_utils::GetClusterDescYAML(),
+                            target_devices,
+                            1,  // one "host memory channel", currently a 1G huge page
+                            dynamic_tlb_config,
+                            false, // skip driver allocs - no (don't skip)
+                            true,  // clean system resources - yes
+                            true); // perform harvesting - yes
+
+    // PCIe core is at (x=0, y=3) on Wormhole NOC0.
```
pjanevskiTT, line 698: Maybe you can immediately create `tt_xy_pair` that you create in function calls

tests/wormhole/test_silicon_driver_wh.cpp
```diff
@@ -644,3 +655,85 @@ TEST(SiliconDriverWH, VirtualCoordinateBroadcast) {
...
+    const size_t PCIE_X = 0;    // NOC0
+    const size_t PCIE_Y = 3;    // NOC0
+    const size_t test_size_bytes = 0x4000;
```
pjanevskiTT, line 701: Is there some reason why we are doing this with `0x4000` bytes? Whether it is random or specific for the test, it would be nice to have a comment here

pjanevskiTT (approved):
Test is very useful, looks great. Left a few nit comments

tests/wormhole/test_silicon_driver_wh.cpp
joelsmithTT: Created https://github.com/tenstorrent/tt-umd/issues/118

tests/wormhole/test_silicon_driver_wh.cpp
joelsmithTT: The technique in the test is slow.  0x4000 is somewhat arbitrary - enough that it demonstrates that the mechanism is working, but small enough that the test finishes in less than one second.  I added a comment.

PR #127, opened 2024-10-07
Fix #126

### LINKED ISSUE #126: "device ID" is overloaded / ambiguous
`get_num_tt_pci_devices_by_pci_device_id` in hwloc code does not get passed the same device id value consistently.

joelsmithTT:
* Remove redundant size variable in `init_hugepage` method.
* Do not `log_fatal` in a non-fatal situation.
* Instead of calling `get_num_tt_pci_devices_by_pci_device_id` incorrectly, don't call it at all (doesn't provide useful information in the error path).

### DIFF
--- a/device/cpuset_lib.cpp
+++ b/device/cpuset_lib.cpp
@@ -458,7 +458,6 @@ int tt_cpuset_allocator::_get_num_tt_pci_devices_by_pci_device_id(uint16_t devic
     if (m_num_tt_device_by_pci_device_id_map.find(device_id_revision) != m_num_tt_device_by_pci_device_id_map.end()) {
         return m_num_tt_device_by_pci_device_id_map.at(device_id_revision);
     } else {
-        log_fatal("Cannot find any TT device with PCI device_id: 0x{:x} and revision: {} in topology.", device_id, revision);
         return 0;
     }
 }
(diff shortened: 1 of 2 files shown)


device/cpuset_lib.cpp
```diff
@@ -458,7 +458,6 @@ int tt_cpuset_allocator::_get_num_tt_pci_devices_by_pci_device_id(uint16_t devic
     if (m_num_tt_device_by_pci_device_id_map.find(device_id_revision) != m_num_tt_device_by_pci_device_id_map.end()) {
         return m_num_tt_device_by_pci_device_id_map.at(device_id_revision);
     } else {
-        log_fatal("Cannot find any TT device with PCI device_id: 0x{:x} and revision: {} in topology.", device_id, revision);
```
broskoTT, line 461: maybe leave it as log_warn, since something did happen which was unintended (this function was called with a non existing device_id) ?

broskoTT (approved):

device/cpuset_lib.cpp
```diff
@@ -458,7 +458,6 @@ int tt_cpuset_allocator::_get_num_tt_pci_devices_by_pci_device_id(uint16_t devic
     if (m_num_tt_device_by_pci_device_id_map.find(device_id_revision) != m_num_tt_device_by_pci_device_id_map.end()) {
```
broskoTT, line 458: I don't see any functional changes in this PR, only cosmetic changes. How does this fix #126 , what am I missing?

device/cpuset_lib.cpp
joelsmithTT: That's reasonable, I'll turn it into a log_warn.

device/cpuset_lib.cpp
joelsmithTT: The call to `get_num_tt_pci_devices_by_pci_device_id` in the error path `if (mapping == MAP_FAILED)` is removed.  The `get_num_tt_pci_devices_by_pci_device_id` method was hitting a `log_fatal` because it was passed a bad parameter.

Issue #53, opened 2024-09-10
Consistent logging, with a proper logger module


broskoTT:
Per chat with Joel:

Consistent logging, with a proper logger module (there are a few different logging techniques in UMD, including a home-grown logger that looks like it came from BBE).  Something that gives us timestamps, and source location (file.cpp:123) would be nice.  I've used spdlog (fmt-based) and glog (stream style syntax) in the past but I don't have strong opinions about the library.  Something that's thread safe and fast, can log to stderr or to a file.

vtangTT:
Related to this, we should get rid of `LOG1` `LOG2` and similar stuff

blozano-tt:
I vote for `spdlog`

broskoTT:
> I vote for `spdlog`

Is this used by tt-metal as well? I'd like us to be consistent.

PR #130, opened 2024-10-08
Logging and constants cleanup

### LINKED ISSUE #112: Remove/revisit g_ constants and defines used for logging
Could be related to https://github.com/tenstorrent/tt-umd/issues/53

Revisit constants found at the beginning of tt_silicon_driver.cpp.
Could be just a rename from LOG1 LOG2 to log_debug or log_somethingelse

pjanevskiTT:
Fixes #112 #53 

Cleanup constants and logging modules inside `tt_silicon_driver.cpp`

- `LOG1` -> `log_debug`
- `LOG2` -> `log_trace`

Other prints were straight 1-1 mapping to logger prints (`log_warning`, `log_error`)

Removed unused constants

### DIFF
--- a/device/tt_silicon_driver.cpp
+++ b/device/tt_silicon_driver.cpp
@@ -54,12 +54,6 @@
 using namespace boost::interprocess;
 using namespace tt;
 
-int g_DEBUG_LEVEL; // /src/t6ifc/t6py/packages/tenstorrent/jlink/jtag_comm.cpp
-bool g_READ_CHECKING_ENABLED = true;
-
-// Print all buffers smaller than this number of bytes
-uint32_t g_NUM_BYTES_TO_PRINT = 8;
-
 // Workaround for tkmd < 1.21 use device_fd_per_host_ch[ch] instead of device_fd once per channel.
 const bool g_SINGLE_PIN_PAGE_PER_FD_WORKAROND = true;
 const uint32_t g_MAX_HOST_MEM_CHANNELS = 4;
@@ -70,7 +64,7 @@ const uint32_t HUGEPAGE_MAP_MASK = HUGEPAGE_REGION_SIZE - 1;
 static const uint32_t MSG_ERROR_REPLY = 0xFFFFFFFF;
 
 // Hardcode (but allow override) of path now, to support environments with other 1GB hugepage mounts not for runtime.
-const char* hugepage_dir_env = std::getenv("TT_BACKEND_HUGEPAGE_DIR");
+const char* hugepage_dir_env = std::getenv("TT_UMD_HUGEPAGE_DIR");
 std::string hugepage_dir = hugepage_dir_env ? hugepage_dir_env : "/dev/hugepages-1G";
 
 // TLB size for DRAM on blackhole - 4GB
@@ -275,7 +269,7 @@ bool auto_reset_board(PCIDevice *dev) {
 }
 
 void detect_ffffffff_read(PCIDevice *dev, std::uint32_t data_read = 0xffffffffu) {
-    if (g_READ_CHECKING_ENABLED && data_read == 0xffffffffu && is_hardware_hung(dev)) {
+    if (data_read == 0xffffffffu && is_hardware_hung(dev)) {
         std::uint32_t scratch_data = *register_address<std::uint32_t>(dev, dev->read_checking_offset);
 
         if (auto_reset_board(dev)) {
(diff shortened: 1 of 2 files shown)


joelsmithTT (approved):
Looks reasonable.  Here are a couple ideas for future work in this area.

* I'd rather type `UMD_INFO("...` or `UMD_WARN("...` than `log_debug(LogSiliconDriver, "...`.  Whatever `LogSiliconDriver` represents (module name?) I don't think is useful for a project the size of UMD.
* Timestamp and exact source location (`tt_silicon_driver.cpp:123`) would go a long way in making UMD logging more useful.  It becomes much easier to trace the flow of a failing program if you have line numbers compared to looking at the locations where `tt_SiliconDevice::method_name` is logged.

broskoTT:
> * I'd rather type `UMD_INFO("...` or `UMD_WARN("...` than `log_debug(LogSiliconDriver, "...`.  Whatever `LogSiliconDriver` represents (module name?) I don't think is useful for a project the size of UMD.

But this way we're using the same logging infra as the other part of our stack? I'd vote for uniformity, rather than each component having their own logging.

device/tt_silicon_driver.cpp
```diff
@@ -90,7 +64,7 @@ const uint32_t HUGEPAGE_MAP_MASK = HUGEPAGE_REGION_SIZE - 1;
 static const uint32_t MSG_ERROR_REPLY = 0xFFFFFFFF;
 
 // Hardcode (but allow override) of path now, to support environments with other 1GB hugepage mounts not for runtime.
-const char* hugepage_dir_env = std::getenv("TT_BACKEND_HUGEPAGE_DIR");
+const char* hugepage_dir_env = std::getenv("TT_UMD_HUGEPAGE_DIR");
```
broskoTT, line 67: If you're changing this, you need to make sure that you're not breaking some pipeline/setup somewhere. For instance, maybe the cloud team relies on setting TT_BACKEND_HUGEPAGE_DIR env?

device/tt_silicon_driver.cpp
```diff
@@ -307,19 +281,17 @@ void detect_ffffffff_read(PCIDevice *dev, std::uint32_t data_read = 0xffffffffu)
 }
 
 inline void record_access (const char* where, uint32_t addr, uint32_t size, bool turbo, bool write, bool block, bool endline) {
-    LOG2 ("%s PCI_ACCESS %s 0x%8x  %8d bytes %s %s%s", where, write ? "WR" : "RD", addr, size, turbo ? "TU" : "  ", block ? "BLK" : "   ", endline ? "\n" : "" );
+    log_trace(LogSiliconDriver, "{} PCI_ACCESS {} 0x{:x}  {} bytes {} {}{}", where, write ? "WR" : "RD", addr, size, turbo ? "TU" : "  ", block ? "BLK" : "   ", endline ? "\n" : "" );
 }
 
 inline void print_buffer (const void* buffer_addr, uint32_t len_bytes = 16, bool endline = true) {
     // Prints each byte in a buffer
-    if (g_DEBUG_LEVEL > 1) {
```
broskoTT, line 315: Was this always enabled previously? Was it always defined as larger than 1? 
Or is the same check hidden behind log_* calls?

broskoTT: Ah I see now bellow it was passed through std::getenv("TT_PCI_LOG_LEVEL");
So now we'll use LOGGER_LEVEL=something instead of it, which looks fine.

device/tt_silicon_driver.cpp
```diff
@@ -1149,13 +1112,13 @@ void tt_SiliconDevice::write_device_memory(const void *mem_ptr, uint32_t size_in
             address += transfer_size;
             buffer_addr += transfer_size;
         }
-        // LOG1 ("Write done Dynamic TLB with pid=%ld\n", (long)getpid());
+        // log_debug(LogSiliconDriver, "Write done Dynamic TLB with pid={}", (long)getpid());
```
broskoTT, line 1115: You can uncomment this if it looks useful

device/tt_silicon_driver.cpp
```diff
@@ -1192,7 +1155,7 @@ void tt_SiliconDevice::read_device_memory(void *mem_ptr, tt_cxy_pair target, std
             address += transfer_size;
             buffer_addr += transfer_size;
         }
-        // LOG1 ("Read done Dynamic TLB with pid=%ld\n", (long)getpid());
+        // log_debug(LogSiliconDriver, "Read done Dynamic TLB with pid={}", (long)getpid());
```
broskoTT, line 1158: You can uncomment this if it looks useful

broskoTT (approved):
Nice!

device/tt_silicon_driver.cpp
pjanevskiTT: Reverted this, created the issue to unify naming of all env vars https://github.com/tenstorrent/tt-umd/issues/146

Issue #118, opened 2024-10-04
Rethink tt_SiliconDevice construction


joelsmithTT:
`tt_SiliconDevice` construction is clumsy:
* TLB windows should be an implementation detail to the greatest extent possible.  Do not require the caller to specify the relationship between TLB index and strings used to name them.
* Hugepages, number of host memory channels should be an implementation detail.  What UMD users really care about is memory that is device-accessible.
* Flags controlling whether to do allocations and cleanup at construction indicate a broken design (see https://github.com/tenstorrent/tt-umd/issues/72)
* `tt_SiliconDevice` isn't necessarily usable after it is constructed: depending on what the caller is trying to do, there are other methods that need to be invoked.

Part of the problem is that `tt_SiliconDevice` is not a device abstraction, but has grown into a [god object](https://en.wikipedia.org/wiki/God_object).  At the _device_ level, a SOC descriptor makes sense, but a topology descriptor does not.  So some of these issues will go away as the design is stratified to be more hierarchical.

broskoTT:
During the experimental cleanups I did in the code, I found that it looks much easier to do tt_SiliconDevice -> tt::umd::Cluster thansition, and then lower the stuff which is chip specific to Chip. So "At the device level, a SOC descriptor makes sense, but a topology descriptor does not" -> topology descriptor should be left in this class, and soc will be moved to new class. This is just a technical comment.

A comment on the constructor: The constructor of tt_SiliconDevice should have 0 mandatory arguments. The constructor of future tt::umd::Chip should have one argument, device_id.
Soc and topology descriptors should be fetched from umd, not given to.

broskoTT:
> * `tt_SiliconDevice` isn't necessarily usable after it is constructed: depending on what the caller is trying to do, there are other methods that need to be invoked.

It is okay that something else has to be done as well. For example start/stop the device, and maybe some custom configuration prior to starting the device (maybe some custom tlb maps, this is up for discussion anyways, or setting some address parameters), which could even be modified in the same object after stopping/starting the device again. But I do agree that if there is a set of functions that are expected to be called each time after constructor, that should be part of constructor.

joelsmithTT:
>  I found that it looks much easier to do tt_SiliconDevice -> tt::umd::Cluster thansition, and then lower the stuff which is chip specific to Chip

This is a good insight.

> But I do agree that if there is a set of functions that are expected to be called each time after constructor, that should be part of constructor.

[...]

PR #131, opened 2024-10-08
Remove dynamic tlb config from constructor

pjanevskiTT:
Remove `dynamic_tlb_config` map from `tt_SiliconDevice` constructor.

Part of the larger effort #118. This is just removing one of the problems that we are facing with tt_SiliconDevice constructor.

### DIFF
--- a/device/architecture_implementation.h
+++ b/device/architecture_implementation.h
@@ -52,6 +52,7 @@ class architecture_implementation {
     virtual uint32_t get_grid_size_x() const = 0;
     virtual uint32_t get_grid_size_y() const = 0;
     virtual uint32_t get_tlb_cfg_reg_size_bytes() const = 0;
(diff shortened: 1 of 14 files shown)


broskoTT:
> Might be worth creating a Metal branch, updating its UMD submodule, fixing up whatever breaks (somewhere Metal must be constructing a `tt_SiliconDevice`) and then running Metal CI.

The ideal way would be making two-phase changes for each API change that we make. That means, implement the new api while keeping old functional. Then change tt-metal usage to new api. Then delete the old api here. 
That way we don't block other improvements in UMD from being merged into metal. However, this only solves problems if we have a high throughput of code changes in UMD, but since we don't, this might be fine.

All this said, please do create an experimental tt-metal branch and verify everything will remain functional after this change before merging this.

pjanevskiTT:
Build with tt-metal, merging...

PR #132, opened 2024-10-08
Disable SiliconDriverWH.HarvestingRuntime test

joelsmithTT:
It's flaky on UMD CI.


broskoTT:
I actually preferred fixing this rather than disabling... It didn't seem flaky to me, it seemed as not passing.
With this test enabled we could see immediately if this was fixed on a new machine.
Although, now the red status on new PRs will indicate something is likely wrong in the PR, there is benefit in keeping our CI green.

Issue #134, opened 2024-10-09
Make sure PCIDevice::numa_node is correct


joelsmithTT:
It might not be getting initialized properly.

PR #141, opened 2024-10-09
Fix NUMA node detection

joelsmithTT:
https://github.com/tenstorrent/tt-umd/issues/134

1. Fix implicit assumption that every accelerator is of the same type (GS/WH/BH)
2. Remove some use of PCIDevice::logical_id from tt_silicon_driver.cpp
3. Add NUMA node detection to PCIDevice class
4. Add a PCIDevice class test harness (+ test for NUMA node detection)

### DIFF
--- a/device/pcie/pci_device.cpp
+++ b/device/pcie/pci_device.cpp
@@ -156,6 +156,24 @@ inline void memcpy_from_device(void *dest, const void *src, std::size_t num_byte
 // --------------------------------------------------------------------------------------------------------------
 // --------------------------------------------------------------------------------------------------------------
 
+/* static */ std::vector<int> PCIDevice::enumerate_devices() {
+    std::vector<int> device_ids;
+    std::string path = "/dev/tenstorrent/";
+
+    for (const auto& entry : std::filesystem::directory_iterator(path)) {
+        std::string filename = entry.path().filename().string();
+
+        // TODO: this will skip any device that has a non-numeric name, which
+        // is probably what we want longer-term (i.e. a UUID or something).
+        if (std::all_of(filename.begin(), filename.end(), ::isdigit)) {
+            device_ids.push_back(std::stoi(filename));
+        }
+    }
+
+    std::sort(device_ids.begin(), device_ids.end());
+    return device_ids;
+}
+
 PCIDevice::PCIDevice(int device_id, int logical_device_id) {
     // TODO: use C++ constructor to do everything
     // TODO: make public member vars const
--- a/device/pcie/pci_device.cpp
+++ b/device/pcie/pci_device.cpp
@@ -23,6 +23,38 @@
 #include "common/assert.hpp"
 #include "common/logger.hpp"
 
+static PciDeviceInfo read_device_info(int fd)
+{
+    tenstorrent_get_device_info info{};
+    info.in.output_size_bytes = sizeof(info.out);
+
+    if (ioctl(fd, TENSTORRENT_IOCTL_GET_DEVICE_INFO, &info) < 0) {
+        TT_THROW("TENSTORRENT_IOCTL_GET_DEVICE_INFO failed");
+    }
+
+    uint16_t bus = info.out.bus_dev_fn >> 8;
+    uint16_t dev = (info.out.bus_dev_fn >> 3) & 0x1F;
+    uint16_t fn = info.out.bus_dev_fn & 0x07;
+
+    return PciDeviceInfo{info.out.vendor_id, info.out.device_id, info.out.pci_domain, bus, dev, fn};
+}
+
+static int determine_numa_node(int fd)
+{
+    const auto device_info = read_device_info(fd);
+    const auto sysfs_path = fmt::format("/sys/bus/pci/devices/{:04x}:{:02x}:{:02x}.{}/numa_node",
+                                        device_info.pci_domain, device_info.pci_bus,
+                                        device_info.pci_device, device_info.pci_function);
+
+    std::ifstream numa_file(sysfs_path);
+    int numa_node = -1;
+    if (numa_file >> numa_node) {
+        return numa_node;
+    }
+    return -1;
+}
+
+
 tt::ARCH detect_arch(int device_id){
     std::uint32_t pcie_device_id = get_pcie_info(device_id, "pcie_device_id");
     std::uint32_t pcie_revision_id = get_pcie_info(device_id, "revision");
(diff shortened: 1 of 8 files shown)


tests/pcie/test_pcie_device.cpp
```diff
@@ -0,0 +1,52 @@
+
```
broskoTT, line 1: Thanks for the test! 
Can we make this run on CI? Is this currently being bundled in a separate binary during build?

tests/pcie/test_pcie_device.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include <gtest/gtest.h>
+
+#include <algorithm>
+#include <filesystem>
+#include <string>
+#include <vector>
+
+#include "device/pcie/pci_device.hpp"
+
+static std::vector<int> simple_pcie_device_enumeration()
```
broskoTT, line 11: This is essentially what ttkmd_scan function does, right?
I'd really like for us to avoid duplication in some way.
Maybe both should be merged into a static PCIDevice::enumerate_devices ?

device/tt_silicon_driver.cpp
```diff
@@ -1369,28 +1369,33 @@ void tt_SiliconDevice::set_fallback_tlb_ordering_mode(const std::string& fallbac
 }
 
 // TT<->TT P2P support removed in favor of increased Host memory.
+// TODO: this is in the wrong place, it should be in the PCIDevice.
```
broskoTT, line 1372: I plan to move around stuff like this from tt_SiliconDevice to PCIDevice and newly formed TTDevice.
Just fyi

device/pcie/pci_device.hpp
```diff
@@ -39,15 +50,16 @@ class PCIDevice {
...
     void open_hugepage_per_host_mem_ch(uint32_t num_host_mem_channels);
-    bool reset_board();
     tt::umd::architecture_implementation* get_architecture_implementation() const { return architecture_implementation.get(); }
 
-    int device_id;
-    int logical_id;
+    PciDeviceInfo info;
+
+    int device_id;  // N in /dev/tenstorrent/N
+    int logical_id; // TODO: does not belong in here
```
broskoTT, line 58: Is there a difference between these two? I mean, will they evaluate to different values? On a system with two n300, we'll have /dev/tenstorrent/0 and /dev/tenstorrent/1, and then the remote chips will land as 2 and 3, but is it ever possible that logical and pci ids are not the same?

device/pcie/pci_device.cpp
```diff
@@ -124,18 +156,26 @@ inline void memcpy_from_device(void *dest, const void *src, std::size_t num_byte
...
     this->device_id = device_id;
     this->logical_id = logical_device_id;
     setup_device();
+
+    this->info = read_device_info(device_fd);
+
 }
 
-PCIDevice::~PCIDevice(){
+PCIDevice::~PCIDevice() {
```
broskoTT, line 171: Fyi, clang format will fix these issues when I enable it

device/pcie/pci_device.cpp
```diff
@@ -23,6 +23,38 @@
 #include "common/assert.hpp"
 #include "common/logger.hpp"
 
+static PciDeviceInfo read_device_info(int fd)
```
broskoTT, line 26: There is a similar function in get_pcie_device_info and get_pcie_info in utils.hpp, can you just merge this with that, we definitely don't need two ways to call TENSTORRENT_IOCTL_GET_DEVICE_INFO?
Also, I don't like the implementation being in .hpp file. And all the utils in utils.hpp are pci related. Maybe we could just move all of that inside pci_device as private methods (which would make the merge easier then)?

tests/pcie/test_pcie_device.cpp
```diff
@@ -0,0 +1,52 @@
...
+    // 1. all of them are -1 (not a NUMA system)
+    // 2. all of them are >= 0 (NUMA system)
+    // 3. empty vector (no devices enumerated)
+
+    if (!nodes.empty()) {
+        bool all_negative_one = std::all_of(nodes.begin(), nodes.end(), [](int node) { return node == -1; });
+        bool all_non_negative = std::all_of(nodes.begin(), nodes.end(), [](int node) { return node >= 0; });
+
+        EXPECT_TRUE(all_negative_one || all_non_negative)
+            << "NUMA nodes should either all be -1 (non-NUMA system) or all be non-negative (NUMA system)";
```
broskoTT, line 48: If there is ever any error here, printing out what was the the actual observed result would be beneficial to understand what is happening

device/pcie/pci_device.cpp
```diff
@@ -23,6 +23,38 @@
...
+    uint16_t bus = info.out.bus_dev_fn >> 8;
+    uint16_t dev = (info.out.bus_dev_fn >> 3) & 0x1F;
+    uint16_t fn = info.out.bus_dev_fn & 0x07;
+
+    return PciDeviceInfo{info.out.vendor_id, info.out.device_id, info.out.pci_domain, bus, dev, fn};
+}
+
+static int determine_numa_node(int fd)
+{
+    const auto device_info = read_device_info(fd);
```
broskoTT, line 44: You will already have this in this->info ?

device/pcie/utils.hpp
```diff
@@ -47,7 +47,7 @@ tenstorrent_get_device_info get_pcie_device_info(int device_fd) {
 std::uint32_t get_pcie_info(int device_id, const std::string &info_needed) {
     // Get PCIe device info through iotcl
     int device_fd = find_device(device_id);
-    static const tenstorrent_get_device_info device_info = get_pcie_device_info(device_fd);
+    auto device_info = get_pcie_device_info(device_fd);
```
broskoTT, line 50: If you choose to move this to pci_device as part of this PR (it's okay if you don't, we can improve it later), then this is unnecessary since you already have it in "this->info"

broskoTT (approved):
I've left a lot of comments, feel free to leave some for later if easier (but just write to the issue that they're not addressed).

device/pcie/pci_device.hpp
joelsmithTT: I looked closer at the behavior of `create-ethernet-map` and I think under situations most normal users would encounter, they will be the same.  However this is not guaranteed (I can manipulate my system into a state where those two ID values are different).

tests/pcie/test_pcie_device.cpp
joelsmithTT: It is a separate binary, since it can run independent of the architecture.  I've added to `run-tests.yml` - please take a look.

device/pcie/pci_device.cpp
joelsmithTT: I noticed utils.hpp after writing this code.  I will consolidate it into the PCIDevice class.

device/pcie/utils.hpp
joelsmithTT: I created https://github.com/tenstorrent/tt-umd/issues/147 to remove this utils.hpp

tests/pcie/test_pcie_device.cpp
joelsmithTT: Good point - added.

tests/pcie/test_pcie_device.cpp
joelsmithTT: Looks like I also needed to add the test to the CMake target `umd_tests` for CI to pick it up.

device/pcie/pci_device.cpp
joelsmithTT: I think I will sort this out in #147

device/pcie/pci_device.cpp
joelsmithTT: #147 - will take care of that here

joelsmithTT:
@broskoTT thanks I appreciate the input.  I'm going to defer some of the the PCIDevice code consolidation/cleanup to #147.

device/pcie/pci_device.hpp
broskoTT: I actually just had this case.
VM gives you /dev/tenstorrent/1, which will resolve to logical_device_id 0 in cluster descriptor, but obviously PCI id will be 1

joelsmithTT: I did not consider the VM or the Docker container case.  Was this with VM on cloud, or IRD?

PR #144, opened 2024-10-10
Cleanup leftover PCIDevice methods from tt_silicon_driver

broskoTT:
A couple of methods were commented out with detecting hanged device disabled after #64 
This PR reintroduces it back, and moves all this PCIDevice related functions to that class.

### DIFF
--- a/device/pcie/pci_device.cpp
+++ b/device/pcie/pci_device.cpp
@@ -23,8 +23,6 @@
 #include "common/assert.hpp"
 #include "common/logger.hpp"
 
-constexpr unsigned int c_hang_read_value = 0xffffffffu;
-
 static PciDeviceInfo read_device_info(int fd)
 {
     tenstorrent_get_device_info info{};
--- a/device/pcie/pci_device.cpp
+++ b/device/pcie/pci_device.cpp
@@ -512,11 +512,3 @@ void PCIDevice::detect_ffffffff_read(std::uint32_t data_read) {
         throw std::runtime_error("Read 0xffffffff from PCIE: you should reset the board.");
     }
 }
-
-void PCIDevice::resume_after_device_reset() {
-    setup_device();
-}
-
-void PCIDevice::suspend_before_device_reset() {
-    close_device();
-}
--- a/device/pcie/pci_device.cpp
+++ b/device/pcie/pci_device.cpp
@@ -12,6 +12,8 @@
 #include <sys/ioctl.h> // for ioctl
 #include <sys/mman.h>  // for mmap, munmap
 #include <linux/pci.h> // for PCI_SLOT, PCI_FUNC
+#include <spawn.h> // for posix_spawn
+#include <wait.h> // for waitpid
 
 #include "pci_device.hpp"
 #include "utils.hpp"
@@ -448,6 +450,10 @@ void PCIDevice::read_block(uint64_t byte_addr, uint64_t num_bytes, uint8_t* buff
     } else {
         memcpy(dest, src, num_bytes);
     }
+
+    if (num_bytes >= sizeof(std::uint32_t)) {
+        detect_ffffffff_read(*reinterpret_cast<std::uint32_t*>(dest));
+    }
 }
 
 // This is only needed for the BH workaround in iatu_configure_peer_region since no arc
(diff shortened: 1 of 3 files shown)


vtangTT:
Originally commented these out as they didn't seem necessary and were in the critical read path. Are there plans to refactor or improve this functionality in the future?

broskoTT:
> Originally commented these out as they didn't seem necessary and were in the critical read path. Are there plans to refactor or improve this functionality in the future?

But this does detect a hang during read operation? Wouldn't commenting these out return wrong data?

device/pcie/pci_device.cpp
```diff
@@ -493,3 +499,83 @@ void PCIDevice::write_tlb_reg(uint32_t byte_addr, std::uint64_t value_lower, std
...
+
+
+bool PCIDevice::is_hardware_hung() {
+    volatile const void *addr = reinterpret_cast<const char *>(bar0_uc) + (get_architecture_implementation()->get_arc_reset_scratch_offset() + 6 * 4) - bar0_uc_offset;
+    std::uint32_t scratch_data = *reinterpret_cast<const volatile std::uint32_t*>(addr);
+
+    return (scratch_data == 0xffffffffu);
+}
+
+bool PCIDevice::reset_by_sysfs() {
```
joelsmithTT, line 511: As far as I can tell, this is dead code.

device/pcie/pci_device.cpp
```diff
@@ -493,3 +499,83 @@ void PCIDevice::write_tlb_reg(uint32_t byte_addr, std::uint64_t value_lower, std
...
+    reset_device.in.flags = 0;
+
+    if (ioctl(device_fd, TENSTORRENT_IOCTL_RESET_DEVICE, &reset_device) == -1) {
+        return false;
+    }
+
+    return (reset_device.out.result == 0);
+}
+
+bool PCIDevice::auto_reset_board() {
```
joelsmithTT, line 559: I don't think UMD should be in the business of auto resetting anything.  UMD is not in a position to restore any the hardware to a known good state, so at best this is a convenience for the developer (won't have to run tt-smi).  At worst, it will complicate debugging efforts if user hangs a chip.

For what it's worth, I've never seen this work the way one would expect it to from reading the code.

pjanevskiTT:
> But this does detect a hang during read operation? Wouldn't commenting these out return wrong data?

I found this useful few times while testing multiple things, for example on Blackhole this helped us debug few PCI hangs we had, so I would say we still need to keep this around. About that being on critical perf path, I agree, 99% of the time we are not going to hit reading 0xFF but we still check. I don't think we should remove this, we can easily remove it later on, I am not really sure what is the actual perf benefit of removing this. If we can measure it and then have an accurate numbers, that could be the argument to remove it.

Maybe we could expose, based on Release/debug build, safe debug read function, which is going to check for 0xFF always, and release read function which is not going to check that

device/pcie/pci_device.cpp
```diff
@@ -493,3 +499,83 @@ void PCIDevice::write_tlb_reg(uint32_t byte_addr, std::uint64_t value_lower, std
...
+}
+
+void PCIDevice::detect_ffffffff_read(std::uint32_t data_read) {
+    if (data_read == 0xffffffffu && is_hardware_hung()) {
+        std::uint32_t scratch_data = *get_register_address<std::uint32_t>(read_checking_offset);
+
+        if (auto_reset_board()) {
+            throw std::runtime_error("Read 0xffffffff from PCIE: auto-reset succeeded.");
+        } else {
+            throw std::runtime_error("Read 0xffffffff from PCIE: you should reset the board.");
```
joelsmithTT, line 570: What should the application do if it catches a `std::runtime_error`?  Unless it parses the `what()` string, the only reasonable thing to do is to terminate the program.

Either branch here is a fatal error.

joelsmithTT (approved):
Detecting a hung chip is fine, but I am not convinced that:
* the reset code actually works
* the reset code is something UMD should be doing

device/pcie/pci_device.cpp
broskoTT: I think that is exactly the purpose of runtime_error. To tell the user this is fatal. 
Yes, both branches return runtime_error, as normal execution cannot continue after hang is detected. It is obvious from this code that the intention for auto reset feature was just to conveniently reset the card if it can, not try to continue normal execution.

device/pcie/pci_device.cpp
broskoTT: Makes sense to me. I'll remove the auto reset feature, we can bring it back if the need arises. I'll leave the hang detection though.

device/pcie/pci_device.cpp
```diff
@@ -493,3 +497,26 @@ void PCIDevice::write_tlb_reg(uint32_t byte_addr, std::uint64_t value_lower, std
     }
     tt_driver_atomics::mfence(); // Otherwise subsequent WC loads move earlier than the above UC store to the TLB register.
 }
+
+bool PCIDevice::is_hardware_hung() {
+    volatile const void *addr = reinterpret_cast<const char *>(bar0_uc) + (get_architecture_implementation()->get_arc_reset_scratch_offset() + 6 * 4) - bar0_uc_offset;
```
pjanevskiTT, line 502: nit: space between char and *

broskoTT:
I've removed the auto reset code as @joelsmithTT had doubts on whether it was even functional (I also don't remember ever seeing it succeed).

@vtangTT and @pjanevskiTT raised concerns on the impact of detecting if the hardware is hung on reads, I've added this issue to further investigate that #156

device/pcie/pci_device.cpp
```diff
@@ -493,3 +497,26 @@ void PCIDevice::write_tlb_reg(uint32_t byte_addr, std::uint64_t value_lower, std
...
+    std::uint32_t scratch_data = *reinterpret_cast<const volatile std::uint32_t*>(addr);
+
+    return (scratch_data == 0xffffffffu);
```
pjanevskiTT, line 505: Maybe we can put `0xffffffffu` to some const var or define, since we are using it in multiple places

pjanevskiTT (approved):
nice, few nit comments

device/pcie/pci_device.cpp
broskoTT: I'm sorry, I specifically don't want to fix comments like since adding clang formatting to our repo will fix all of these automatically :) Will work on adding that soon.

device/pcie/pci_device.cpp
broskoTT: Added c_hang_read_value and also renamed detect_ffffffff_read to detect_hang_read

Issue #98, opened 2024-10-02
TTDevice class


broskoTT:
According to https://docs.google.com/drawings/d/1-m1azdsBqMA0A6ATYRMfkhyeuOJuGCEI62N5a96LXj0/edit?usp=sharing
Make TTDevice abstract class which holds unique_ptr<PCIDevice>. This is then specialized per Arch , rather than holding architecture_implementation per arch. Also move functions from PCIDevice for what makes sense to be in TTDevice.
One potential implementation can be found here https://github.com/tenstorrent/tt-umd/blob/brosko/new_api_experiment/new_device/tt_device.h

broskoTT:
Further work is to move more tt_device related stuff to tt_device, both from pci_device and cluster.cpp

broskoTT:
Initial push for this class was done. Further movement of functions might be captured in another task.

PR #145, opened 2024-10-10
Rename architecture_implementation to TTDevice

broskoTT:
This is just a first step related to #98 
For more information consult https://docs.google.com/drawings/d/1-m1azdsBqMA0A6ATYRMfkhyeuOJuGCEI62N5a96LXj0/edit

This change just renames and moves files, without any functional changes -> for ease of review.
The following change will implement further changes to set TTDevice own PCIDevice, rather than other way around.

Not sure though if I should wait for original tt_device.h to be renamed first to something else (cluster.h), not to add confusion with this rename.


device/tt_device/tt_device.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include "device/tt_device/grayskull_tt_device.h"
+#include "device/tt_device/wormhole_tt_device.h"
+
+namespace tt::umd {
+
+std::unique_ptr<TTDevice> TTDevice::create(architecture architecture) {
+    switch (architecture) {
+        case architecture::blackhole: return std::make_unique<BlackholeTTDevice>();
+        case architecture::grayskull: return std::make_unique<GrayskullTTDevice>();
+        case architecture::wormhole:
```
pjanevskiTT, line 17: not sure whether we should still support wormhole (without b0), I think we can deprecate that arch

pjanevskiTT (approved):
Nice

device/tt_device/tt_device.cpp
broskoTT: I don't want to make the decision to drop this for now. It's not a big cost to support it. When we choose to drop it, we will easily remove it.

device/pcie/pci_device.cpp
```diff
@@ -197,7 +196,7 @@ void PCIDevice::setup_device() {
     this->pcie_device_id = get_pcie_info(this->device_id, "pcie_device_id");
     this->pcie_revision_id = get_pcie_info(this->device_id, "revision");
     this->arch = detect_arch(pcie_device_id, pcie_revision_id);
-    this->architecture_implementation = tt::umd::architecture_implementation::create(static_cast<tt::umd::architecture>(arch));
+    this->architecture_implementation = tt::umd::TTDevice::create(static_cast<tt::umd::architecture>(arch));
```
joelsmithTT, line 199: `this->tt_device`?

device/pcie/pci_device.hpp
```diff
@@ -55,7 +55,8 @@ class PCIDevice {
     void write_tlb_reg(uint32_t byte_addr, std::uint64_t value_lower, std::uint64_t value_upper, std::uint32_t tlb_cfg_reg_size);
 
     void open_hugepage_per_host_mem_ch(uint32_t num_host_mem_channels);
-    tt::umd::architecture_implementation* get_architecture_implementation() const { return architecture_implementation.get(); }
+    bool reset_board();
+    tt::umd::TTDevice* get_architecture_implementation() const { return architecture_implementation.get(); }
```
joelsmithTT, line 59: Does `get_architecture_implementation` still make sense?

device/tt_device.h
```diff
@@ -585,7 +585,7 @@ class tt_device
     std::unordered_map<chip_id_t, tt_SocDescriptor> soc_descriptor_per_chip = {};
 };
 
-#include "device/architecture_implementation.h"
+#include "device/tt_device/tt_device.h"
```
joelsmithTT, line 588: `device/tt_device.h` and `device/tt_device/tt_device.h` is confusing.

joelsmithTT (approved):

device/tt_device.h
broskoTT: Yes, exactly. I've left a comment in the PR because of that, to wait for tt_device rename to cluster.h first maybe. Even tt_device and TTDevice is confusing. @vtangTT removed this class previously specifically to remove confusion.
I'll leave this comment open until this is resolved.

device/pcie/pci_device.cpp
broskoTT: This will disappear with the next actual refactor, this was just a rename of the type so the following refactor is clearer. PCIDevice shouldn't hold TTDevice, but vice-versa, according to the redesign diagram.

device/pcie/pci_device.hpp
broskoTT: This will disappear with the next actual refactor, this was just a rename of the type so the following refactor is clearer. PCIDevice shouldn't hold TTDevice, but vice-versa, according to the redesign diagram.

broskoTT:
Deprecated by #329

PR #148, opened 2024-10-11
Remove soc desc from constructor

pjanevskiTT:
Opening a PR as a draft for reviewing the removal of passing soc descriptor to `tt_SiliconDevice` constructor. 

### Notes for reviewing

- Method `get_silicon_drivers` is implemented in cluster descriptor, as that is what tt-metal's `tt_cluster` has. Maybe it is not the right class, but when we impelement our own cluster we should move it there. It was just done according to current design
- A lot of BH and GS tests are failing, I ported just WH tests to confirm that everything is working
- Cherry picked the commit for removing `dynamic_tlb_config` from constructor, I can still merge it as a separate PR
- Try to focus on the concept of removing soc desc, I guess there are some nit details as well, I will fix those in a proper PR


device/tt_silicon_driver.cpp
```diff
@@ -645,6 +659,19 @@ tt_SiliconDevice::tt_SiliconDevice(const std::string &sdesc_path, const std::str
...
+tt_SiliconDevice::tt_SiliconDevice(const std::string& sdesc_path, const std::string &ndesc_path, const std::set<chip_id_t> &target_devices, 
+                                   const uint32_t &num_host_mem_ch_per_mmio_device, const bool skip_driver_allocs,
+                                   const bool clean_system_resources, bool perform_harvesting, std::unordered_map<chip_id_t, uint32_t> simulated_harvesting_masks) : tt_device() {
+    construct_tt_silicon_device(sdesc_path, ndesc_path, target_devices, num_host_mem_ch_per_mmio_device, skip_driver_allocs, clean_system_resources, perform_harvesting, simulated_harvesting_masks);
+}
+
+tt_SiliconDevice::tt_SiliconDevice(tt::ARCH arch, const std::string &ndesc_path, const std::set<chip_id_t> &target_devices, 
+                                   const uint32_t &num_host_mem_ch_per_mmio_device, const bool skip_driver_allocs,
+                                   const bool clean_system_resources, bool perform_harvesting, std::unordered_map<chip_id_t, uint32_t> simulated_harvesting_masks) : tt_device() {
+    std::string sdesc_path = get_sdesc_path(arch);
```
broskoTT, line 671: arch should not be passed, but inferred from target_devices, right?

joelsmithTT: It would be ideal that `arch` isn't passed in.  An edge case is where the system has different cards in it (e.g. a WH and a GS) although it is reasonable if we don't want to support that case.

Something this eliminates is the ability of the caller to provide a custom SOC descriptor, i.e. a 1x1 or 2x2 grid.

device/tt_cluster_descriptor.cpp
```diff
@@ -640,3 +652,40 @@ BoardType tt_ClusterDescriptor::get_board_type(chip_id_t chip_id) const {
   BoardType board_type = this->chip_board_type.at(chip_id);
   return board_type;
 }
+
+std::unordered_map<chip_id_t, std::unique_ptr<tt_SiliconDevice>> tt_ClusterDescriptor::get_silicon_drivers() {
```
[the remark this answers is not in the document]
joelsmithTT: Metal instantiates one `tt_SiliconDevice` per PCIe device.  I think (?) UMD could be used with just a single instance of it.

https://github.com/tenstorrent/tt-metal/blob/021137c47a58841b6dfe9233a70ea449177bae51/tt_metal/llrt/tt_cluster.cpp#L211
https://github.com/tenstorrent/tt-metal/blob/021137c47a58841b6dfe9233a70ea449177bae51/tt_metal/llrt/tt_cluster.cpp#L295

Agree with @broskoTT, for what `tt_SiliconDevice` is doing a more apt name would be Cluster.  As a Cluster object, what we currently call `tt_SiliconDevice` accepting SOC/Cluster descriptors makes sense.  Part of the design problem we have is that `tt_SiliconDevice` grew from a (PCI) device abstraction to a cluster abstraction.

Passing in the SOC descriptor can be eliminated, but only if we are removing the ability for the user to provide their own descriptor at runtime.  That may be OK (users who need such a thing could add a new ARCH) but it's not clear to me if that's the right way to go.

device/tt_silicon_driver.cpp
broskoTT: If a system has different cards, you reference which device_id (as reported by tt_ClusterDescriptor) you want to open, and you still don't pass arch or soc_desc

device/tt_cluster_descriptor.cpp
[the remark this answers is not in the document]
broskoTT: After https://github.com/tenstorrent/tt-metal/pull/13949 this will be true. After that I'll rename to Cluster

pjanevskiTT:
Closing it to avoid a lot of conflicts, PR that addresses this properly is going to be out soon

Issue #117, opened 2024-10-04
Improve device enumeration code


joelsmithTT:
tt_silicon_driver.cpp has `std::vector<chip_id_t> ttkmd_scan()` (exposed as `static std::vector<chip_id_t> tt_SiliconDevice::detect_available_device_ids();`).

cpuset_lib.cpp has its own PCIe device enumeration code, which has caused me problems in the past in mutli-card systems with one or more cards unbound from the tenstorrent kernel driver.

Finally, `create-ethernet-map` binary does device enumeration both locally (PCIe) and across the topology of Ethernet-connected chips.

Some considerations:
* Ideal: a single interface for PCIe device enumeration, single interface for topology discovery.
* Get rid of `create-ethernet-map`: https://github.com/tenstorrent/tt-umd/issues/73
* Long term, can we skip the topology information going through a YAML phase and instantiate the right Chip instances directly?
* Get rid of ARCH_NAME when building unit tests.  
* Enumerate PCIe devices so that caller can know some details (definitely architecture, possibly board type) without having to fully map the device BARs.  Sysfs has the information needed.

broskoTT:
> * Get rid of ARCH_NAME when building unit tests.

This will be covered through some other issue.



> * Ideal: a single interface for PCIe device enumeration, single interface for topology discovery.
I do agree device_id discovery should be in single place.
As is also described (briefly) on https://docs.google.com/drawings/d/1-m1azdsBqMA0A6ATYRMfkhyeuOJuGCEI62N5a96LXj0/edit
The idea is for this to be behind static functions tt::umd::ClusterDescriptor::GetClusterDescriptor() and tt::umd::ClusterDescriptor::GetAllChips() and cluster_descriptor->get_mmio_chips()

pjanevskiTT:
Closing this just because it is already broken up into multiple smaller issues

Issue #104, opened 2024-10-03
register_address untangle


broskoTT:
The function register_address is clearly aarch specific, but this is hidden under some ifelses on bar spaces.
Refactor these functions (in TTDevice/PCIDevice), make them clearly arch specific and easier to read.

pjanevskiTT:
Duplicate of #63, closing this

PR #149, opened 2024-10-11
PCIDevice class cleanup

joelsmithTT:
* Rehome code in pcie/utils.hpp to the PCIDevice class
* Attempt to clean up ambiguity RE: what "device id" actually means
* Start enforcing an abstractional boundary at the PCI device level


device/pcie/pci_device.cpp
```diff
@@ -23,6 +23,47 @@
...
 #include "common/logger.hpp"
 
+static const uint16_t GS_PCIE_DEVICE_ID = 0xfaca;
+static const uint16_t WH_PCIE_DEVICE_ID = 0x401e;
+static const uint16_t BH_PCIE_DEVICE_ID = 0xb140;
+
+// TODO: we'll have to rethink this when KMD takes control of the inbound PCIe
+// TLB windows and there is no longer a pre-defined WC/UC split.
+static const uint32_t GS_BAR0_WC_MAPPING_SIZE = (156<<20) + (10<<21) + (18<<24);
+static const uint32_t BH_BAR0_WC_MAPPING_SIZE = 188<<21; // Defines the address for WC region. addresses 0 to BH_BAR0_WC_MAPPING_SIZE are in WC, above that are UC
```
pjanevskiTT, line 33: I know this is not your addition, but I think it would be nice to be consistent with comments of the variables. Keep the comments above the variable rather than in the same line. Like I said, change it if you want, if not maybe we can do it for eveyrthing in one PR

device/pcie/pci_device.cpp
```diff
@@ -174,59 +207,38 @@ inline void memcpy_from_device(void *dest, const void *src, std::size_t num_byte
...
-
-void PCIDevice::setup_device() {
-    this->device_fd = find_device(this->device_id);
-    this->numa_node = determine_numa_node(this->device_fd);
-    this->pcie_device_id = get_pcie_info(this->device_id, "pcie_device_id");
-    this->pcie_revision_id = get_pcie_info(this->device_id, "revision");
-    this->arch = detect_arch(pcie_device_id, pcie_revision_id);
-    this->architecture_implementation = tt::umd::architecture_implementation::create(static_cast<tt::umd::architecture>(arch));
-
+PCIDevice::PCIDevice(int device_number, int logical_device_id)
```
broskoTT, line 210: I'm not sure if this is where we want to land, changing this from setup_device to the constructor... We can leave it like this for now, but I did have some other thoughts:

I thought maybe we could have "two modes" of opening PCIDevice. When you create a class you'd just be able to read what arch it is, and some other PCI parameters. Then, when calling setup_device, you actually open the device for usage and you get BARs available.

So this would map to constructor not calling any KMD function, and setup_device calling TENSTORRENT_IOCTL_QUERY_MAPPINGS.

These thoughts were during my rewriting experiment, maybe this makes it clearer: https://github.com/tenstorrent/tt-umd/blob/pjanevski/new_api_experiment/ideal_device/pci/pci_device.h#L37

The only argument I have for doing it this way, is that than some other host application (while another is doing whatever with this device) can get arch type or even soc descriptor without mapping all the mappings, just by opening PCIDevice class and reading the arch (so without calling setup_device). While working on Blackhole Buda issues, I ran across some Buda runtime code calling get_arch_type multiple times, which resulted in PCIDevice being opened and closed multiple times (which created some bugs for me). Still not sure if it is a good argument though.

device/pcie/pci_device.cpp
```diff
@@ -39,26 +80,21 @@ static PciDeviceInfo read_device_info(int fd)
...
-    int numa_node = -1;
-    if (numa_file >> numa_node) {
-        return numa_node;
+static tt::ARCH detect_arch(uint32_t pcie_device_id, uint32_t pcie_revision_id) {
+    if (pcie_device_id == GS_PCIE_DEVICE_ID){
+        return tt::ARCH::GRAYSKULL;
+    } else if (pcie_device_id == WH_PCIE_DEVICE_ID && pcie_revision_id == 0x01){
+        return tt::ARCH::WORMHOLE_B0;
+    } else if (pcie_device_id == WH_PCIE_DEVICE_ID){
+        // TODO: did we ship any of these?  I've never seen one.  Can we stop
```
pjanevskiTT, line 89: I left similar comment to one of @broskoTT PRs. I think deprecating Wormhole is totally fine, I think we don't have a system in the company that is running anything on these cards

device/tt_silicon_driver.cpp
```diff
@@ -313,9 +311,23 @@ namespace {
...
+    bool multicast,
+    std::unordered_map<chip_id_t, std::unordered_map<tt_xy_pair, tt_xy_pair>>& harvested_coord_translation,
+    std::uint64_t ordering)
+{
     auto architecture_implementation = dev->get_architecture_implementation();
+
+    // TODO(Joel): the PCIDevice should not really be carring this around - this
+    // is the last place that extracts it from the PCIDevice.  Since KMD will
+    // eventually take over the TLB programming, this can get removed later on.
+    auto logical_id = dev->get_logical_id();
```
broskoTT, line 329: Yes, I actually started removing logical_id, but found some other refactors (like moving set_dynamic_tlb) to have more sense to be moved first.
logical_id should be deleted from PCIDevice. I planed to do it, if you don't overtake me in doing it :)

device/pcie/pci_device.cpp
```diff
@@ -297,63 +309,64 @@ void PCIDevice::setup_device() {
...
         uint64_t iatu_base = UNROLL_ATU_OFFSET_BAR + iatu_index * 0x200;
         uint32_t region_ctrl_2 = 0 << 31; // REGION_EN = 0
-        write_regs(reinterpret_cast<std::uint32_t*>(static_cast<uint8_t*>(bar2_uc) + iatu_base + 0x04), &region_ctrl_2, 1);
+        write_regs(reinterpret_cast<uint32_t*>(static_cast<uint8_t*>(bar2_uc) + iatu_base + 0x04), &region_ctrl_2, 1);
     }
 
-    if (device_fd != -1) {
-        ::close(device_fd);
-    }
+    close(fd);
```
broskoTT, line 369: Are you sure this won't throw if fd is -1? Or is the code written in such way that this should be impossible?

device/tt_silicon_driver.cpp
```diff
@@ -54,8 +54,6 @@
 using namespace boost::interprocess;
 using namespace tt;
 
-// Workaround for tkmd < 1.21 use device_fd_per_host_ch[ch] instead of device_fd once per channel.
-const bool g_SINGLE_PIN_PAGE_PER_FD_WORKAROND = true;
```
broskoTT, line 58: So this is not needed anymore, we've advanced kmd?

device/pcie/pci_device.hpp
```diff
@@ -34,84 +34,144 @@ struct PciDeviceInfo
...
 public:
     /**
-     * Return a list of integers corresponding to character devices in /dev/tenstorrent/
+     * @return a list of integers corresponding to character devices in /dev/tenstorrent/
      */
     static std::vector<int> enumerate_devices();
 
-    PCIDevice(int device_id, int logical_device_id);
+    /**
+     * PCI device constructor.
```
pjanevskiTT, line 54: Great docs. You can maybe add these functions to doxygen docs list, I think these are undocumented at the moment

pjanevskiTT (approved):
nice, few nit comments

device/tt_silicon_driver.cpp
```diff
@@ -1475,7 +1497,7 @@ bool tt_SiliconDevice::init_hugepage(chip_id_t device_id) {
         pin_pages.in.virtual_address = reinterpret_cast<std::uintptr_t>(mapping);
         pin_pages.in.size = hugepage_size;
 
-        auto &fd = g_SINGLE_PIN_PAGE_PER_FD_WORKAROND ? dev->device_fd_per_host_ch[ch] : dev->device_fd;
+        auto fd = dev->get_fd();
 
         if (ioctl(fd, TENSTORRENT_IOCTL_PIN_PAGES, &pin_pages) == -1) {
```
broskoTT, line 1502: I plan to move this to PCIDevice, so we encapsulate all ioctls there, and reduce code in future tt::umd::Chip. Just FYI, no action on this comment.

device/pcie/pci_device.cpp
```diff
@@ -297,63 +309,64 @@ void PCIDevice::setup_device() {
...
 }
 
-void PCIDevice::close_device() {
+PCIDevice::~PCIDevice() {
     if (arch == tt::ARCH::BLACKHOLE && bar2_uc != nullptr && bar2_uc != MAP_FAILED) {
         // Disable ATU index 0
         // TODO: Implement disabling for all indexes, once more host channels are enabled.
+
+        // This is not going to happen if the application crashes, so if it's
+        // essential for correctness then it needs to move to the driver.
```
broskoTT, line 362: Yeah, if we deleted this code, everything would still work. I was just trying to be diligent, but this code is not necessary (since you setup the ATU each time you start using the device). 
Just FYI, no action on this comment.

device/pcie/pci_device.hpp
```diff
@@ -5,23 +5,23 @@
...
 #include "device/architecture_implementation.h"
 
-static uint32_t GS_BAR0_WC_MAPPING_SIZE = (156<<20) + (10<<21) + (18<<24);
-static uint32_t BH_BAR0_WC_MAPPING_SIZE = 188<<21; // Defines the address for WC region. addresses 0 to BH_BAR0_WC_MAPPING_SIZE are in WC, above that are UC
-
-static const uint32_t BH_NOC_NODE_ID_OFFSET = 0x1FD04044;
-static const uint32_t GS_WH_ARC_SCRATCH_6_OFFSET = 0x1FF30078;
-
+// TODO: this is used up in tt_silicon_driver.cpp but that logic ought to be
+// lowered into the PCIDevice class since it is specific to PCIe cards.
```
broskoTT, line 17: FYI on the redesign.
My idea is that PCIDevice should be rather thin, and the new TTDevice class would hold this.
As your logic was here, so was mine: If it looks fairly low level, it should go out of Chip and Cluster classes (which tt_SiliconDevice ultimately goes towards).
And another logic was, try to put everything arch related (which includes this) into TTDevice abstraction.

No action on this comment.

device/pcie/pci_device.hpp
```diff
@@ -34,84 +34,144 @@ struct PciDeviceInfo
 };
 
 class PCIDevice {
+    const std::string device_path;  // Path to character device: /dev/tenstorrent/N
+    const int device_num;           // N in /dev/tenstorrent/N
```
broskoTT, line 38: I'm not convinced device_num is a more clarifying name than device_id.
Is maybe pci_device_id better?

device/pcie/pci_device.hpp
```diff
@@ -34,84 +34,144 @@ struct PciDeviceInfo
...
+    const int logical_id;           // Unique identifier for each device in entire network topology
+    const int fd;                   // Character device file descriptor
```
broskoTT, line 40: I really don't have a problem with longer and more descriptive variable names, now that we're out of the dark age and don't have such limits on variable names and code sizes, etc.
Can we name this pci_device_file_desc?

device/pcie/pci_device.hpp
```diff
@@ -34,84 +34,144 @@ struct PciDeviceInfo
...
+    // interface assumes the caller knows what they are doing - but it's unclear
+    // how to use this interface correctly without knowing details of the chip
+    // and its state.
+    // TODO: build a proper abstraction for IO.  At this level, that is access
+    // to registers in BAR0 (although possibly the right abstraction is to add
+    // methods that perform specific operations as opposed to generic register
+    // read/write methods) and access to segments of BAR0/4 that are mapped to
+    // NOC endpoints.  Probably worth waiting for the KMD to start owning the
+    // resource management aspect of these PCIe->NOC mappings (the "TLBs")
+    // before doing too much work here...
```
broskoTT, line 129: FYI on the redesign.
As I mentioned in another comment, this is arch specific, and will be raised to TTDevice. For the part of "use this interface correctly without knowing details of the chip and its state", yes, that will be hidden probably behind tt::umd::Core or tt::umd::AbstractIO

No action on this comment.

device/pcie/pci_device.hpp
```diff
@@ -34,84 +34,144 @@ struct PciDeviceInfo
...
+
+    // TODO: this didn't survive a refactor.  It was part of some logic to
+    // detect a hung chip and reset it.  Detecting a hung chip is a good idea,
+    // but it should probably be a fatal error.  It is unclear if the recovery
+    // code was functional -- the "reset by sysfs" path certainly didn't look
+    // like it was.  Let's leave recovery up to the user + tt-smi for now, since
+    // UMD is in no position to actually be able to recover the hardware state
+    // on behalf of the application.
+    //
+    // But we should put back a if (hardware_is_hung()) TT_FATAL("it's dead");
```
broskoTT, line 170: Yeah this is exactly what #144 does. I've removed auto_reset code from there.

device/pcie/pci_device.hpp
```diff
@@ -34,84 +34,144 @@ struct PciDeviceInfo
...
-    std::size_t bar2_uc_size;
+    size_t bar2_uc_size;
 
     void *bar4_wc = nullptr;
-    std::uint64_t bar4_wc_size;
+    uint64_t bar4_wc_size;
 
+    // TODO: let's get rid of this unless we need to run UMD on WH systems with
+    // shrunk BAR0.  If we don't (and we shouldn't), then we can just use BAR0
+    // and simplify the code.
```
broskoTT, line 156: Right, I was thinking the same, this system_reg_mapping thing is a bit confusing. We will end up refactoring it. First it needs to be raised to TTDevice class since it is arch specific.

No action on this comment.

broskoTT (commented):
Thanks, this looks better now!

device/pcie/pci_device.cpp
joelsmithTT: Fixed

device/pcie/pci_device.cpp
joelsmithTT: I'll leave it alone for now, since it's unrelated to this PR.  Created https://github.com/tenstorrent/tt-umd/issues/158, I will remove the comment.

device/pcie/pci_device.hpp
joelsmithTT: Fixed - although this was a little hard to find so I updated `docs/README.md`

device/pcie/pci_device.cpp
joelsmithTT: The rationale for relying on constructor/destructor rather than separate methods is to provide a stronger guarantee that PCIDevice instances are in a good state.

I haven't carefully considered the problem you describe, and although I intend to leave this change as is for now I am not opposed to refactoring it back in the future if it ends up being the wrong thing.

However, I think improved PCI device enumeration code can address the issue.  (Full topology device enumeration -- what create-ethernet-map does today -- is a different story).  I updated https://github.com/tenstorrent/tt-umd/issues/117 - I agree that we should not be mapping the BAR(s) simply to discover what architecture a chip is.

device/tt_silicon_driver.cpp
joelsmithTT: That comment is actually a lie - I wrote it before I encountered another instance (`tt_SiliconDevice::broadcast_pcie_tensix_risc_reset` needs the logical id also).  Your observation about other refactors is consistent with mine - too big a blast radius to delete logical_id from PCIDevice in this PR.  I fixed the comment.

device/pcie/pci_device.hpp
joelsmithTT: A PCI device ID is a 16-bit identifier that uniquely identifies a PCI device model, it's part of the PCI spec.  For Grayskull it is `0xfaca`, Wormhole `0x401e`.  I was trying to avoid having "device ID" in the name because of how overloaded this term is in our system.

device/pcie/pci_device.hpp
joelsmithTT: Sure, I renamed it.

device/pcie/pci_device.hpp
joelsmithTT: Ok cool - I will remove this comment when I rebase on main.

device/pcie/pci_device.hpp
joelsmithTT: I just confirmed with syseng that future systems will not have a shrunken BAR0, even with large numbers of chips.

device/pcie/pci_device.hpp
broskoTT: I agree device_id is much more ambiguous than anything. 
What about pci_device_num? Adding pci_ prefix makes it clearer, at least for me.

joelsmithTT: Renamed to `pci_device_num`

device/pcie/pci_device.hpp
broskoTT: #104

device/tt_arch_types.h
```diff
@@ -10,6 +10,9 @@
 #include "device/architecture.h"
 
 namespace tt {
+
+// TODO: why do we have ARCH and architecture?  This is a mess.  Can we have just one?
+// Can we get rid of the entries that (for all practical purposes) do not exist?
```
broskoTT, line 15: Yes, just so we don't forget, #161

broskoTT (approved):
Much better now!

device/tt_arch_types.h
joelsmithTT: Thanks for adding that - I've been in the habit of leaving these TODOs as I am working on the code with the intent of creating actual issues for them, but some I forget about!

PR #153, opened 2024-10-14
First API test

broskoTT:
A plan is to have a set of API tests:
- To showcase UMD usage to clients
- Which is arch agnostic
- Which showcases (like this simplest one) how cumbersome is the current minimal UMD setup.

A first PR for #154

### DIFF
--- a/tests/api/test_cluster.cpp
+++ b/tests/api/test_cluster.cpp
@@ -37,7 +37,7 @@ TEST(ApiTest, OpenAllChips) {
         int physical_device_id = pci_device_ids[0];
         // TODO: remove logical_device_id
         PCIDevice pci_device (physical_device_id, 0);
-        tt::ARCH device_arch = pci_device.get_arch();
+        device_arch = pci_device.get_arch();
     }
 
     // TODO: Make this test work on a host system without any tt devices.
--- a/tests/api/test_cluster.cpp
+++ b/tests/api/test_cluster.cpp
@@ -24,10 +24,29 @@ using Cluster = tt_SiliconDevice;
 // This test should be one line only.
 TEST(ApiTest, OpenAllChips) {
 
+    // TODO: This should not be needed. And could be part of the cluster descriptor probably.
+    // Note that cluster descriptor holds logical ids of chips.
+    // Which are different than physical PCI ids, which are /dev/tenstorrent/N ones.
+    // You have to see if physical PCIe is GS before constructing a cluster descriptor.
+    std::vector<int> pci_device_ids = PCIDevice::enumerate_devices();
+    std::set<int> pci_device_ids_set(pci_device_ids.begin(), pci_device_ids.end());
+
+    // TODO: This should be removed from the API, the driver itself should do it.
+    std::string soc_path;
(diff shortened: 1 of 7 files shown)


tests/api/test_cluster.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+// This test should be one line only.
+TEST(ApiTest, OpenAllChips) {
+
+    // TODO: This should not be needed. And could be part of the cluster descriptor probably.
+    // Note that cluster descriptor holds logical ids of chips.
+    // Which are different than physical PCI ids, which are /dev/tenstorrent/N ones.
+    // You have to see if physical PCIe is GS before constructing a cluster descriptor.
+    std::vector<int> pci_device_ids = PCIDevice::enumerate_devices();
+    std::set<int> pci_device_ids_set;
```
joelsmithTT, line 32: nit: You can avoid the loop like, `std::set<int> s{ pci_device_ids.being(), pci_device_ids.end() }`
Although maybe having the cluster accept a vector rather than a set makes more sense?

tests/api/test_cluster.cpp
```diff
@@ -0,0 +1,76 @@
...
+    for (int pci_device_id : pci_device_ids) {
+        pci_device_ids_set.insert(pci_device_id);
+    }
+
+    // TODO: This should be removed from the API, the driver itself should do it.
+    std::string soc_path;
+    int physical_device_id = pci_device_ids[0];
```
joelsmithTT, line 39: This is undefined if `pci_device_ids` is empty, maybe an early exit + a warning ("no chips detected") in that case?

joelsmithTT (approved):

tests/api/test_cluster.cpp
broskoTT: Will change it.
I won't change the API as part of this PR (changing it to accept a vector), I wrote this test mainly to just to document our current API usage.

tests/api/test_cluster.cpp
broskoTT: Left another TODO in test to make sure Cluster() doesn't throw on a system without TT device.

tests/api/test_cluster.cpp
```diff
@@ -0,0 +1,79 @@
...
+    std::vector<int> pci_device_ids = PCIDevice::enumerate_devices();
+    std::set<int> pci_device_ids_set (pci_device_ids.begin(), pci_device_ids.end());
+
+    tt::ARCH device_arch = tt::ARCH::GRAYSKULL;
+    if (!pci_device_ids.empty()) {
+        // TODO: This should be removed from the API, the driver itself should do it.
+        int physical_device_id = pci_device_ids[0];
+        // TODO: remove logical_device_id
+        PCIDevice pci_device (physical_device_id, 0);
+        tt::ARCH device_arch = pci_device.get_arch();
```
pjanevskiTT, line 40: Not sure I understand, why are we redefining device_arch in this scope?

broskoTT: It's an error, good catch!

Issue #142, opened 2024-10-09
Lower PCIDevice-specific operations to PCIDevice class


joelsmithTT:
The `tt_SiliconDevice` class contains data structures that are indexed by the so-called "logical" id, yet some of these structures represent things that conceptually do not exist for remote chips.

From this standpoint, these structures would make more sense if they were indexed by "device id" (not to be confused with PCIe device id).  However, a better design would move these concepts out of the `tt_SiliconDevice` class and into `PCIDevice`.

broskoTT:
I'm going through some rewrite... here are the items that make sense to be moved:

- bar io: bar_write32 and bar_read32
- iatu specific code: iatu_configure_peer_region and init_pcie_iatus and code from destructor
- hugepages: lots of code, init_hugepage, also destruction. Also moves out the communication with the kernel driver.
- arc_msg: There is also remote_arc_msg. Some part should live in some lower part of the stack, and some part has to be part of the Chip abstraction.

pjanevskiTT:
@broskoTT situation here?

broskoTT:
> [@broskoTT](https://github.com/broskoTT) situation here?

In progress.
#253 resolved hugepages part.
The other 3 items still remain.

PR #155, opened 2024-10-15
Move PCIDevice functions to PCIDevice class

broskoTT:
Some of the functions which clearly accept PCIDevice pointer were located in tt_silicon_driver, but they obviously should be part of the PCIDevice class.
Those same functions are also arch specific, so they'll probably be moved to new class TTDevice according to https://docs.google.com/drawings/d/1-m1azdsBqMA0A6ATYRMfkhyeuOJuGCEI62N5a96LXj0/edit

I've also changed two functions, broadcast_pcie_tensix_risc_reset and get_power_state_arc_msg to accept chip_id rather than PCIDevice pointer. These were private members, so no API change.

Related to #142

### DIFF
--- a/device/pcie/pci_device.cpp
+++ b/device/pcie/pci_device.cpp
@@ -524,7 +524,7 @@ dynamic_tlb PCIDevice::set_dynamic_tlb(unsigned int tlb_index, tt_xy_pair start,
     auto translated_start_coords = harvested_coord_translation.at(logical_id).at(start);
     auto translated_end_coords = harvested_coord_translation.at(logical_id).at(end);
     uint32_t tlb_address    = address / tlb_config.size;
-    uint32_t local_offset   = address % tlb_config.size;
+    uint32_t local_address   = address % tlb_config.size;
     uint64_t tlb_base       = tlb_config.base + (tlb_config.size * tlb_config.index_offset);
     uint32_t tlb_cfg_reg    = tlb_config.cfg_addr + (TLB_CFG_REG_SIZE_BYTES * tlb_config.index_offset);
 
@@ -543,10 +543,9 @@ dynamic_tlb PCIDevice::set_dynamic_tlb(unsigned int tlb_index, tt_xy_pair start,
     }.apply_offset(tlb_config.offset);
 
     log_debug(LogSiliconDriver, "set_dynamic_tlb() with tlb_index: {} tlb_index_offset: {} dynamic_tlb_size: {}MB tlb_base: 0x{:x} tlb_cfg_reg: 0x{:x}", tlb_index, tlb_config.index_offset, tlb_config.size/(1024*1024), tlb_base, tlb_cfg_reg);
-    // write_regs(dev -> hdev, tlb_cfg_reg, 2, &tlb_data);
     write_tlb_reg(tlb_cfg_reg, tlb_data.first, tlb_data.second, TLB_CFG_REG_SIZE_BYTES);
 
-    return { tlb_base + local_offset, tlb_config.size - local_offset };
+    return { tlb_base + local_address, tlb_config.size - local_address };
 }
 
 dynamic_tlb PCIDevice::set_dynamic_tlb(unsigned int tlb_index, tt_xy_pair target, std::uint64_t address, std::unordered_map<chip_id_t, std::unordered_map<tt_xy_pair, tt_xy_pair>>& harvested_coord_translation, std::uint64_t ordering) {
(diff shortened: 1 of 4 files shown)


pjanevskiTT (approved):

device/pcie/pci_device.cpp
```diff
@@ -493,3 +493,53 @@ void PCIDevice::write_tlb_reg(uint32_t byte_addr, std::uint64_t value_lower, std
...
+        std::tie(start, end) = architecture_implementation->multicast_workaround(start, end);
+    }
+
+    log_trace(LogSiliconDriver, "set_dynamic_tlb with arguments: tlb_index = {}, start = ({}, {}), end = ({}, {}), address = 0x{:x}, multicast = {}, ordering = {}",
+         tlb_index, start.x, start.y, end.x, end.y, address, multicast, (int)ordering);
+
+    tt::umd::tlb_configuration tlb_config = architecture_implementation->get_tlb_configuration(tlb_index);
+    std::uint32_t TLB_CFG_REG_SIZE_BYTES = architecture_implementation->get_tlb_cfg_reg_size_bytes();
+    auto translated_start_coords = harvested_coord_translation.at(logical_id).at(start);
+    auto translated_end_coords = harvested_coord_translation.at(logical_id).at(end);
```
joelsmithTT, line 511: I don't have anything actionable to suggest here, just some thoughts - 

Passing this in seems clumsy.  I don't understand all of the coordinate stuff, but if this structure is going to live inside the PCIDevice, it doesn't need to be keyed off of logical id anymore (each PCIDevice can have its own).  Also I think the Metal guys would be happy if this were built out of array lookups rather than hash table.

device/pcie/pci_device.cpp
```diff
@@ -493,3 +493,53 @@ void PCIDevice::write_tlb_reg(uint32_t byte_addr, std::uint64_t value_lower, std
...
+    uint32_t tlb_address    = address / tlb_config.size;
+    uint32_t local_offset   = address % tlb_config.size;
+    uint64_t tlb_base       = tlb_config.base + (tlb_config.size * tlb_config.index_offset);
+    uint32_t tlb_cfg_reg    = tlb_config.cfg_addr + (TLB_CFG_REG_SIZE_BYTES * tlb_config.index_offset);
+
+    std::pair<std::uint64_t, std::uint64_t> tlb_data = tt::umd::tlb_data {
+        .local_offset = tlb_address,
```
joelsmithTT, line 518: I know this isn't your fault, but this code is confusing.  There's a local variable named `local_offset`, but the struct member named `local_offset` is assigned something different.

device/pcie/pci_device.cpp
```diff
@@ -493,3 +493,53 @@ void PCIDevice::write_tlb_reg(uint32_t byte_addr, std::uint64_t value_lower, std
...
+        .mcast = multicast,
+        .ordering = ordering,
+        // TODO #2715: hack for Blackhole A0, will potentially be fixed in B0.
+        // Using the same static vc for reads and writes through TLBs can hang the card. It doesn't even have to be the same TLB.
+        // Dynamic vc should not have this issue. There might be a perf impact with using dynamic vc.
+        .static_vc = (get_arch() == tt::ARCH::BLACKHOLE) ? false : true,
+    }.apply_offset(tlb_config.offset);
+
+    log_debug(LogSiliconDriver, "set_dynamic_tlb() with tlb_index: {} tlb_index_offset: {} dynamic_tlb_size: {}MB tlb_base: 0x{:x} tlb_cfg_reg: 0x{:x}", tlb_index, tlb_config.index_offset, tlb_config.size/(1024*1024), tlb_base, tlb_cfg_reg);
+    // write_regs(dev -> hdev, tlb_cfg_reg, 2, &tlb_data);
```
joelsmithTT, line 532: nit: dead code

joelsmithTT (approved):
A step in the right direction.  Left a few comments for your consideration.

Longer term, KMD is going to assume control over some of this stuff for better resource management (see https://tenstorrent.atlassian.net/wiki/spaces/syseng/pages/537100642/PCIe+TLBs -- it looks like KMD will handle arch-specific differences).

device/pcie/pci_device.cpp
broskoTT: I'll fix this when removing logical_id from pci_device. We should at least be passing harvested_coord for this chip only. But also probably not even that (that will be left for even further refactor)

device/pcie/pci_device.cpp
broskoTT: Changed to local_address to remove this confusion, couldn't think of anything better at the moment.

Issue #157, opened 2024-10-15
tt_SiliconDriver support multiple cards


broskoTT:
Related to redesign https://docs.google.com/drawings/d/1-m1azdsBqMA0A6ATYRMfkhyeuOJuGCEI62N5a96LXj0/edit

Following are the blockers to allow trivial transition in tt-metal from using several tt_SiliconDevice to using a single one to manage the whole cluster (tt_SiliconDevice will be renamed to tt::umd::Cluster). I tried doing the rewrite and have stumbled upon a couple of API issues:

- wait_for_non_mmio_flush doesn't accept chip_id
- assert_risc_reset -> there should be a variant accepting chip_id. Also not sure if the current variant works for more than one mmio chip.
  - This includes broadcast_tensix_risc_reset_to_cluster, broadcast_write_to_cluster and ethernet_broadcast_write
- start_device -> check if it works for multiple mmio chips
- devices_grouped_by_assoc_mmio_device_ -> add support in tt_ClusterDescriptor
- nice to have: check for a chip if it is a remote chip: https://github.com/tenstorrent/tt-metal/blob/main/tt_metal/llrt/tt_cluster.cpp#L458

broskoTT:
5 is already present with is_chip_mmio_capable

broskoTT:
Probably related to 4, get_closest_mmio_chip seems to assume a single connected cluster of chips

broskoTT:
Related to 4: We don't support multiple cluster of cards currently, marked explicit by #165 .
However, tt_metal still needs this map which it builds itself. I've added the support directly to tt_clusterDesc in #167
1 will be fixed by #166

Issue #154, opened 2024-10-14
API tests


broskoTT:
A plan is to have a set of API tests:

- To showcase UMD usage to clients
- Which is arch agnostic
- Which showcases (like this simplest one) how cumbersome is the current minimal UMD setup.

A lot of stuff should be taken from existing tests.
At this point it seems harder to rewrite existing tests rather than starting a new one.

The tests should also somewhat follow code examples shown in this diagram: https://docs.google.com/drawings/d/1-m1azdsBqMA0A6ATYRMfkhyeuOJuGCEI62N5a96LXj0/edit

pjanevskiTT:
Closing this since we have initial API tests, following tests need to be added together with new features...

PR #159, opened 2024-10-15
API test - simple IO

broskoTT:
This change adds minimal setup to read/write including remote chips.
I've tested this on a setup with 3 N300 cards.

Motivation was that I wanted to change wait_for_non_mmio_flush to include chip_id. Then I wanted to write a test. Then it didn't make sense writing a test for this if basic IO is not setup.

This test adds to the collection of API tests which we should work on reducing. Related to #154 . Related to #157 

Minor changes:
- get_soc_descriptor had two different definitions, consolidated those.
- added is_chip_remote for convenience


device/tt_cluster_descriptor.h
```diff
@@ -77,6 +77,7 @@ class tt_ClusterDescriptor {
   std::vector<std::tuple<ethernet_channel_t, ethernet_channel_t>> get_directly_connected_ethernet_channels_between_chips(const chip_id_t &first, const chip_id_t &second) const;
   
   bool is_chip_mmio_capable(const chip_id_t &chip_id) const;
+  bool is_chip_remote(const chip_id_t &chip_id) const;
```
joelsmithTT, line 80: nit: consider passing by value - const reference can sometimes imply the object is large or expensive to copy.  (I do see that you're following the pattern of the method above...)

device/tt_device.cpp
```diff
@@ -27,4 +27,6 @@ tt_device::tt_device(const std::string& sdesc_path) : soc_descriptor_per_chip({}
 tt_device::~tt_device() {
 }
 
-const tt_SocDescriptor *tt_device::get_soc_descriptor(chip_id_t chip) const { return &soc_descriptor_per_chip.at(chip); }
+tt_SocDescriptor& tt_device::get_soc_descriptor(chip_id_t chip_id){
```
joelsmithTT, line 30: Can this be `const tt_SocDescriptor&`?  No `const` here implies that the caller can mutate it.  (I don't know if mutating it is necessary... ideally the SOC doesn't change?)

tests/api/test_cluster.cpp
```diff
@@ -75,5 +114,88 @@ TEST(ApiTest, OpenAllChips) {
...
+
+        if (cluster_desc->is_chip_remote(chip_id) && soc_desc.arch != tt::ARCH::WORMHOLE_B0) {
+            std::cout << "Skipping remote chip " << chip_id << " because it is not a wormhole_b0 chip." << std::endl;
+            continue;
+        }
+
+        std::cout << "Reading from chip " << chip_id << " core " << any_core.str() << std::endl;
+
+        std::vector<uint8_t> readback_data(data_size, 0);
+        umd_cluster->read_from_device(readback_data.data(), any_core_global, 0, data_size, "LARGE_WRITE_TLB");
```
joelsmithTT, line 197: I don't think this is a problem, but it might be confusing to someone unfamiliar with the fallback TLB situation: `read_from_device` and `LARGE_WRITE_TLB` are incongruous.

joelsmithTT (approved):

PR #162, opened 2024-10-16
Add cmake install

blozano-tt:
At some point...

We should be able to run cmake --install

The library should be in an export set, but doing that currently causes error due to poor separation of public and private dependencies / include paths.

This is a stepping stone to find_package(umd).

### DIFF
--- a/device/CMakeLists.txt
+++ b/device/CMakeLists.txt
@@ -20,8 +20,7 @@ target_link_libraries(umd_device
     PUBLIC yaml-cpp::yaml-cpp umd_common_directories nng uv compiler_flags
     PRIVATE hwloc rt Boost::interprocess fmt
 )
-target_include_directories(umd_device
-    PRIVATE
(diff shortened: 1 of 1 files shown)


device/CMakeLists.txt
```diff
@@ -30,3 +30,17 @@ set_target_properties(umd_device PROPERTIES
     LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib
     POSITION_INDEPENDENT_CODE ON
 )
+
```
broskoTT, line 33: This should also be used by build-tests.yml and then consumed by run-tests.yml?
I had some problems while trying to make it work the first time...

I tried copying the same config as https://github.com/tenstorrent/tt-mlir/blob/main/.github/workflows/build-and-test.yml

blozano-tt:
These changes have been incorporated into #192

PR #165, opened 2024-10-16
ClusterDescriptor API test - not supporting multiple clusters

broskoTT:
A continuation on working on #154 
The first test showcases the basic usage of cluster descriptor.

The second test tests whether ClusterDescriptor can manage multiple clusters of cards, cluster meaning connected chips (through ethernet). It actually shows explicitly that this is not currently supported. So this test fails on a system with multiple unconnected clusters of cards (simple 2xN150 setup would suffice). We don't have a machine like that in UMD pool, so the test passes on CI.

### DIFF
--- a/tests/api/test_cluster_descriptor.cpp
+++ b/tests/api/test_cluster_descriptor.cpp
@@ -1,25 +1,15 @@
 
 #include <gtest/gtest.h>
-#include "fmt/xchar.h"
 
-#include <algorithm>
-#include <filesystem>
 #include <string>
 #include <vector>
+#include <unordered_map>
 
 #include "tests/test_utils/generate_cluster_desc.hpp"
 
-// TODO: change to tt_cluster
-#include "device/tt_device.h"
+#include "device/pcie/pci_device.hpp"
 #include "device/tt_cluster_descriptor.h"
 
-// TODO: do proper renaming.
-using Cluster = tt_SiliconDevice;
-
-// These tests are intended to be run with the same code on all kinds of systems:
-// E75, E150, E300
-// N150. N300
-// Galaxy
 
 std::unique_ptr<tt_ClusterDescriptor> get_cluster_desc() {
 
--- a/tests/api/test_cluster_descriptor.cpp
+++ b/tests/api/test_cluster_descriptor.cpp
@@ -61,11 +61,20 @@ TEST(ApiClusterDescriptorTest, BasicFunctionality) {
         return;
     }
 
-    auto all_chips = cluster_desc->get_all_chips();
-    auto harvesting_for_chips = cluster_desc->get_harvesting_info();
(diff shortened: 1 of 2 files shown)


tests/api/test_cluster_descriptor.cpp
```diff
@@ -0,0 +1,140 @@
...
+    for (auto [chip, _]: local_chips_to_pci_device_id) {
+        local_chips.insert(chip);
+    }
+    std::unordered_set<chip_id_t> remote_chips;
+    for (auto chip : all_chips) {
+        if (local_chips.find(chip) == local_chips.end()) {
+            remote_chips.insert(chip);
+        }
+    }
+}
```
joelsmithTT, line 68: Does this test need any kind of check, or is it just intended to exercise the code?

joelsmithTT (approved):

tests/api/test_cluster_descriptor.cpp
broskoTT: Well it checks that the code doesn't throw :)
My intention for these api tests is for them to be more focused on offering usage examples and documenting API rather than testing functionality.

tests/api/test_cluster_descriptor.cpp
```diff
@@ -0,0 +1,140 @@
...
+            return sets.size();
+        }
+
+    private:
+        std::unordered_map<int, int> parent;
+};
+
+// This tests fails on a machine with multiple cards.
+// It works as long as all the devices that are discoverable are connected through ethernet.
+// Our ClusterDescriptor doesn't have a notion of multiple unconnected clusters of cards.
```
abhullar-tt, line 108: I thought this does work when we have multiple N300s in a system for e.g.

broskoTT: This is what I asked about a couple of days ago on the umd_improvement chat if you remember.
Imagine first N300 resolves to chips 0, 1 and the second one resolves to 2, 3.
create-eth-map would give eth coordinates of 0,0,0,0 to 0 and 2, and 1,0,0,0 to 1 and 3.

When calling get_closest_mmio, such as this call in tt_metal: https://github.com/tenstorrent/tt-metal/blob/main/tt_metal/llrt/tt_cluster.cpp#L180 , which fills up devices_grouped_by_assoc_mmio_device_ used later, it will return wrong chip, since it will only look at eth_coords.
For instance, it will return either 0 for both 1 and 3, or 2 for both 1 and 3 (depending on the order of collection).

The tricky thing is that if you have some sequential test going over all remote chips and doing something, it might actually look like it is working, but you'd communicate twice with one of the remote chips actually. And I don't think if we even have a test which would start an AI workload over two N300 cards?

Also note that all of this is not try if those two N300 are connected through an ethernet cable, they would have different eth_coords in that case.

PR #168, opened 2024-10-16
Fix hugepage bug

pjanevskiTT:
Fix hugepage bug introduced by https://github.com/tenstorrent/tt-umd/commit/5293e508ade90e758b386068770a882667a535f0 

[Code line](https://github.com/tenstorrent/tt-umd/commit/5293e508ade90e758b386068770a882667a535f0#diff-24a7adb016c3614e7d9564ae1a2016fd04509e26ea5eb565b8945af92e2a4fb9R1450)

TODO

- [x] tt-metal post commit fix

PR #169, opened 2024-10-16
Fix hugetlbfs filename generation

joelsmithTT:


joelsmithTT:
Didn't see https://github.com/tenstorrent/tt-umd/pull/168

PR #170, opened 2024-10-16
Fix a WH unit test

joelsmithTT:
tt_SiliconDevice instances can exist in an invalid state.  This test was relying on hardware state configured by previous tests.  It would fail if executed on its own.

### DIFF
--- a/tests/wormhole/test_silicon_driver_wh.cpp
+++ b/tests/wormhole/test_silicon_driver_wh.cpp
@@ -698,6 +698,9 @@ TEST(SiliconDriverWH, SysmemTestWithPcie) {
                             true,  // clean system resources - yes
                             true); // perform harvesting - yes
 
+    set_params_for_remote_txn(device);
+    device.start_device(tt_device_params{});  // no special parameters
+
     // PCIe core is at (x=0, y=3) on Wormhole NOC0.
     const size_t PCIE_X = 0;    // NOC0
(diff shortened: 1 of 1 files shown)


joelsmithTT:
I had some remarks on Slack about this one, but writing here for posterity -- 

Without the call to `start_device()`, PCIe address translation is not configured.  In this case, `tt_SiliconDevice::write_to_device` and `tt_SiliconDevice::read_from_device` with a NOC target of the PCIe core still appear to succeed, but in this situation the hardware is accessing incorrect addresses on the system bus.  This should not be possible!

With IOMMU enabled, the device would be prevented from accessing the wrong addresses.

This is a good example of how the `tt_SiliconDevice` API is easy to use incorrectly... even for someone who's tripped over this `start_device()` thing before.  I know better and still wrote this bug!

broskoTT:
Please consider adding maybe a log_warning that would be based on a flag which is set when the device is started, and checked when some functions are used which should not be used prior to "start_device" call

joelsmithTT:
> based on a flag which is set when the device is started, and checked when some functions are used

There isn't a completely safe way to do this unless all IO is blocked until that method is called, which defeats its original purpose.  If I understand correctly it was so legacy software could instantiate multiple `tt_SiliconDevice` objects in different processes.  I don't know if the current consumers of UMD have any expectation that that still works.

The situation is indicative of a design problem that resulted from delegating hardware resource management to transient userspace processes.

PR #173, opened 2024-10-16
Check KMD version prior to querying device info

### LINKED ISSUE #133: Add KMD version check
At device enumeration or device instantiation, determine the version of the kernel driver and log it.  Later on this path can be used to enforce a minimum required KMD version for UMD to operate.

joelsmithTT:


device/pcie/pci_device.cpp
```diff
@@ -37,6 +37,38 @@ static const uint32_t BH_BAR0_WC_MAPPING_SIZE = 188<<21;
...
+
+        if (std::getline(iss, token, '.')) {
+            semver.minor = std::stoul(token);
+
+            if (std::getline(iss, token, '.')) {
+                semver.patch = std::stoul(token);
+            }
+        }
+    }
+
```
broskoTT, line 68: Should we print a warning if parsing the version failed?

device/pcie/pci_device.cpp
```diff
@@ -37,6 +37,38 @@ static const uint32_t BH_BAR0_WC_MAPPING_SIZE = 188<<21;
...
 static const uint32_t GS_WH_ARC_SCRATCH_6_OFFSET = 0x1FF30078;
 
+static constexpr semver_t MINIMUM_SUPPORTED_KMD_VERSION{1, 27, 1};
+
+static semver_t read_kmd_version() {
+    static const std::string path = "/sys/module/tenstorrent/version";
+    std::ifstream file(path);
+
+    if (!file.is_open()) {
+       TT_THROW("Failed to open: {}; could not determine tenstorrent driver version", path);
```
broskoTT, line 47: use log_assert.
And maybe switch to a warning?

broskoTT: On second though, please DO switch this to a warning.
If for anycase something is different on some system, this will block any kind of workload and be a pain.

device/pcie/pci_device.hpp
```diff
@@ -43,11 +43,51 @@ struct PciDeviceInfo
...
+    bool operator<=(const semver_t& other) const {
+        return !(other < *this);
+    }
+
+    bool operator>=(const semver_t& other) const {
+        return !(*this < other);
+    }
+
+    std::string to_string() const {
+        return std::to_string(major) + "." + std::to_string(minor) + "." + std::to_string(patch);
```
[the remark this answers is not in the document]
joelsmithTT: Fixed

device/pcie/pci_device.cpp
joelsmithTT: I can make it return {0,0,0} here which will make it fail the check.  Are you suggesting that the check `if (kmd_version < MINIMUM_SUPPORTED_KMD_VERSION)` should be a warning?

If a system is missing `/sys/module/tenstorrent/version` then it simply does not have the driver installed.

broskoTT: Yes, I'm saying both should be a warning, sorry for not being precise. Or if you make it return {0,0,0} here, than just the if (kmd_version < MINIMUM_SUPPORTED_KMD_VERSION) should throw a warning.
This has a potential to affect all existing systems (CI machine, Cloud machines, customer machines...), which might be functional today, but will make them non-functional due to this check.
There might be something custom on some system where this check might fail even though the driver is completely functional.
I'm not confident for the couple of us, without wider consensus, to make these hard constraints on our systems.
Logging a warning is always good though, since it could point in the right direction if something is failing.

joelsmithTT: I rewrote most of this, please take another look.
For now the KMD version is simply logged.  Later on when UMD actually needs to impose a minimum version check, we'll have the infrastructure in place.

device/CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ set(UMD_DEVICE_SRCS
     wormhole/wormhole_coordinate_manager.cpp
     pcie/pci_device.cpp
     hugepage.cpp
+    ../common/semver.cpp
```
broskoTT, line 24: Should we instead compile a common lib, and then make device relying on common lib? @blozano-tt

device/pcie/pci_device.cpp
```diff
@@ -45,6 +46,22 @@ const uint32_t HUGEPAGE_REGION_SIZE = 1 << 30; // 1GB
 
 using namespace tt;
 using namespace tt::umd;
+using tt::umd::semver_t;
```
broskoTT, line 49: This should not be needed

device/pcie/pci_device.hpp
```diff
@@ -56,6 +59,9 @@ struct PciDeviceInfo
     // onto this struct as methods?  e.g. current_link_width etc.
 };
 
+// Do we want to put everything into this file into tt::umd namespace?
+using tt::umd::semver_t;
```
broskoTT, line 63: This seems like an overkill
In any case, yes, we do want everything to leave inside tt::umd eventually.

device/pcie/pci_device.cpp
```diff
@@ -45,6 +46,22 @@ const uint32_t HUGEPAGE_REGION_SIZE = 1 << 30; // 1GB
...
+
+static semver_t read_kmd_version() {
```
broskoTT, line 51: Can you make this a private function of PCIDevice instead?

joelsmithTT: yes, fixed

device/pcie/pci_device.cpp
joelsmithTT: Removed

device/CMakeLists.txt
joelsmithTT: I can do that in this PR, but my thought is that we have opportunity for a broader cleanup.. today we have code in:

common/
src/ 
device/ <- most stuff lives here

I'm not sure what the best way to structure things is.  What do you think?  I am happy to create a ticket and defer this, equally happy to add a libumdcommon (or some such) for now.. let me know.

device/pcie/pci_device.hpp
joelsmithTT: Can you clarify?

joelsmithTT: P.S. comment has a typo, it should read "Do we want to put everything _in_ this file into tt::umd namespace?"

.clang-format
```diff
@@ -15,12 +19,17 @@ AlignAfterOpenBracket: AlwaysBreak
...
 # Disallow single statements after if/else/for/while/do without curly braces.
 InsertBraces: true
 
 # Separate definition blocks, including classes, structs, enums, and functions.
 SeparateDefinitionBlocks: Always
+
+# Line up : and , in ctor init list
+AllowShortFunctionsOnASingleLine: false
+BreakConstructorInitializers: BeforeComma
+ConstructorInitializerAllOnOneLineOrOnePerLine: false
```
joelsmithTT, line 32: This reflects a personal preference on my part, if folks don't like it I can back it out.

The other changes in .clang-format I think we want...

.clang-format
```diff
@@ -2,6 +2,10 @@ BasedOnStyle: Google
 ColumnLimit: 120
 IndentWidth: 4
 
+DerivePointerAlignment: false
```
broskoTT, line 5: Hmmm.. I don't like changing these repo wide clang parameters as part of other PRs, but would like it to be a separate PR which also updates all the affected files so we can see the changes. Can you move it out of this PR?
Also, please comment on ALL options so that they're clear on what they do. For instance, I'm currently not sure on DerivePointerAlignment and what does AllowShortFunctionsOnASingleLine exactly do.

device/CMakeLists.txt
broskoTT: ... or we could keep common as header only, and move those small implementations that you have in .cpp file to a .h file as inline functions. Not sure what is the best way to keep things organised. Would like @blozano-tt 's insight here

.clang-format
joelsmithTT: I'll introduce .clang-format changes separately.

device/CMakeLists.txt
joelsmithTT: My rationale for putting it in .cpp was to keep the fmt include out of the header.

device/pcie/pci_device.cpp
broskoTT: If there is a minimum version needed for anything now, I don't mind printing a warning if version is not at least some predefined minimum version constant.

device/CMakeLists.txt
joelsmithTT: I put it all into the header.  Please take another look.

device/pcie/pci_device.cpp
```diff
@@ -258,7 +258,10 @@ PCIDevice::PCIDevice(int pci_device_number, int logical_device_id) :
     numa_node(read_sysfs<int>(info, "numa_node")),
     revision(read_sysfs<int>(info, "revision")),
     arch(detect_arch(info.device_id, revision)),
-    architecture_implementation(tt::umd::architecture_implementation::create(arch)) {
+    architecture_implementation(tt::umd::architecture_implementation::create(arch)),
+    kmd_version(read_kmd_version()) {
+    log_info(LogSiliconDriver, "Opened device {}; KMD version: {}", pci_device_num, kmd_version.to_string());
```
broskoTT, line 263: nit: add PCI to message: Opened PCI device 

I think it would make less confusion (given that this is log info) for end users.

broskoTT (approved):
Thanks for addressing many comments I've left

device/pcie/pci_device.cpp
joelsmithTT: Added

Issue #136, opened 2024-10-09
Add explicit GCC support 


broskoTT:
Previously, I've switched from GCC to Clang to match tt-metal's toolchain in #74 
@joelsmithTT raised a concern that there is no reason for us to force the user to use clang while gcc is available on the ubuntu images we support. Also switching to Clang breaks gdb debugging which is also useful.

The consensus on the chat is to support GCC as the default compiler, but leave the option for building with Clang, through Cmake options. Tasks related to this issue:
- Build by default using gcc compiler, use a cmake flag for clang CMAKE_CXX_COMPILER / CMAKE_C_COMPILER
  - Double check if the flags can be matched with tt-metal's flags for consistency
- Change CI jobs to build using both (if we're officially supporting both)
- Change README accordingly

broskoTT:
Double check that sudo apt install commands from Readme actually work

PR #174, opened 2024-10-16
Allow GCC 9.4

joelsmithTT:
This change allows UMD to be built and debugged with the distribution-provided tools under Ubuntu 20.04.  This requires CMAKE_C_COMPILER and CMAKE_CXX_COMPILER environment variables to be set to gcc and g++, respectively.  Without them, the build will still default to using a third-party compiler which the user must have installed.

### DIFF
--- a/cmake/compilers.cmake
+++ b/cmake/compilers.cmake
@@ -20,7 +20,7 @@ function(CHECK_COMPILERS)
         endif()
     elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
         if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS "12.0.0")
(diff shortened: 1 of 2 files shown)


joelsmithTT:
Related: https://github.com/tenstorrent/tt-umd/issues/136 although this doesn't default to GCC.

cmake/compilers.cmake
```diff
@@ -20,7 +20,7 @@ function(CHECK_COMPILERS)
         endif()
```
broskoTT, line 20: Can you also document the procedure of building with GCC through readme?

pjanevskiTT (approved):

cmake/compilers.cmake
joelsmithTT: Done

Issue #171, opened 2024-10-16
Fix Metal build


joelsmithTT:
Metal not building against UMD main -- 
https://github.com/tenstorrent/tt-metal/actions/runs/11368114498/job/31622428205

PR #175, opened 2024-10-16
Fix Metal build: put detect_arch() back

joelsmithTT:
This family of functions was removed during a refactor.  Metal is relying on one of them.  This change reintroduces it.

Fixes https://github.com/tenstorrent/tt-umd/issues/171

### DIFF
--- a/device/pcie/pci_device.cpp
+++ b/device/pcie/pci_device.cpp
@@ -190,6 +190,17 @@ inline void memcpy_from_device(void *dest, const void *src, std::size_t num_byte
     }
 }
 
+tt::ARCH PciDeviceInfo::get_arch() const {
+    if (this->device_id == GS_PCIE_DEVICE_ID){
+        return tt::ARCH::GRAYSKULL;
+    } else if (this->device_id == WH_PCIE_DEVICE_ID) {
+        return tt::ARCH::WORMHOLE_B0;
+    } else if (this->device_id == WH_PCIE_DEVICE_ID){
+        return tt::ARCH::BLACKHOLE;
+    }
+    return tt::ARCH::Invalid;
+}
+
 
 /* static */ std::vector<int> PCIDevice::enumerate_devices() {
     std::vector<int> device_ids;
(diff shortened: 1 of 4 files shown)


device/pcie/pci_device.cpp
```diff
@@ -190,6 +190,17 @@ inline void memcpy_from_device(void *dest, const void *src, std::size_t num_byte
     }
 }
 
+tt::ARCH PciDeviceInfo::get_arch() const {
+    if (this->device_id == GS_PCIE_DEVICE_ID){
```
broskoTT, line 194: But we do already have this->arch?

device/tt_silicon_driver.cpp
```diff
@@ -66,6 +66,27 @@ std::string hugepage_dir = hugepage_dir_env ? hugepage_dir_env : "/dev/hugepages
...
+// looks like Metal is doing this because we don't provide any other way... When
+// we are further along in our refactoring efforts and `tt_device` is more of a
+// Cluster abstraction, we should provide Metal with interfaces for:
+//      1. Checking that all chips are of the same architecture (we may not care
+//         about this, but the application might).
+//      2. Getting the architecture of a specific chip.
+// Until then... I'm putting this function back so that Metal will still build
+// next time someone bumps its UMD submodule version.
+tt::ARCH detect_arch(int pci_device_num) {
+    const auto devices_info = PCIDevice::enumerate_devices_info();
```
broskoTT, line 80: Why not just return PCIDevice(pci_device_num)->get_arch(); ?

device/pcie/pci_device.cpp
joelsmithTT: I'm thinking of consolidating these kinds of attributes into the `PciDeviceInfo` struct.

By `this` I assume you are referring to a `PCIDevice` instance.  In this case there isn't one.

device/tt_silicon_driver.cpp
joelsmithTT: It's related to this: https://github.com/tenstorrent/tt-umd/pull/149#discussion_r1801466759

I think doing that should be harmless, but creating/destroying a `PCIDevice` on the same line maps/unmaps the BAR(s) unnecessarily.

broskoTT (approved):
Let's fix this for tt_metal, then we can change it afterwards if there's something to change.

PR #178, opened 2024-10-17
Fix Metal build: put detect_arch() back v2

### LINKED ISSUE #171: Fix Metal build

broskoTT:
I've rushed into merging Joel's change #175
Seems like there were two detect_arch() calls which were needed. 
This time I've manually verified that tt_metal builds with UMD on this branch.

I've also added an API test to document this usage, which should be changed, probably to tt::umd::ClusterDescriptor::get_arch(chip_id) or something similar

Fixes #171

### DIFF
--- a/tests/api/test_cluster_descriptor.cpp
+++ b/tests/api/test_cluster_descriptor.cpp
@@ -10,7 +10,7 @@
 #include "device/pcie/pci_device.hpp"
 #include "device/tt_cluster_descriptor.h"
 
-// Needed for detect_arch, remove when it is part of cluster descriptor.
+// TODO: Needed for detect_arch, remove when it is part of cluster descriptor.
 #include "device/tt_device.h"
 
 
--- a/tests/api/test_cluster_descriptor.cpp
+++ b/tests/api/test_cluster_descriptor.cpp
@@ -59,6 +59,9 @@ TEST(ApiTest, DetectArch) {
         // TODO: This should be the only available API, previous call should be routed to this one to get any arch.
         tt::ARCH arch2 = detect_arch(PCIDevice::enumerate_devices()[0]);
         EXPECT_NE(arch2, tt::ARCH::Invalid);
+
+        // In our current setup, we expect all arch to be the same.
+        EXPECT_EQ(arch, arch2);
     }
 }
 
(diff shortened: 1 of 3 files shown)


tests/api/test_cluster_descriptor.cpp
```diff
@@ -10,6 +10,9 @@
 #include "device/pcie/pci_device.hpp"
 #include "device/tt_cluster_descriptor.h"
 
+// Needed for detect_arch, remove when it is part of cluster descriptor.
```
pjanevskiTT, line 13: add a TODO to remove, otherwise someone won't see it probably

tests/api/test_cluster_descriptor.cpp
```diff
@@ -43,6 +46,22 @@ std::unique_ptr<tt_ClusterDescriptor> get_cluster_desc() {
...
+
+    // Expect it to be invalid if no devices are found.
+    if (PCIDevice::enumerate_devices().empty()) {
+        EXPECT_EQ(arch, tt::ARCH::Invalid);
+    } else {
+        EXPECT_NE(arch, tt::ARCH::Invalid);
+
+        // TODO: This should be the only available API, previous call should be routed to this one to get any arch.
+        tt::ARCH arch2 = detect_arch(PCIDevice::enumerate_devices()[0]);
+        EXPECT_NE(arch2, tt::ARCH::Invalid);
```
pjanevskiTT, line 61: If I am reading the code correctly, we can also add line 
`EXPECT_EQ(arch, arch2)`

pjanevskiTT (approved):
nice, nit comments

tests/api/test_cluster_descriptor.cpp
broskoTT: While that will be true in our current setup/code, that is wrong from purely the API perspective. 
The detect_arch() returns ANY arch found in the cluster, and with another call I specified a specific PCI device for which I want arch for.
Which is the reason to avoid the first call (without arguments) altogether probably.

pjanevskiTT: > Which is the reason to avoid the first call (without arguments) altogether probably.

I think this is the main point. But I still think it would still be nice to check that `detect_arch()` returns at least one of the arch that you get when you iterate ove PCI devices

broskoTT: I've added the initially suggested check. This will get removed anyway, so hence my indifference.

Issue #68, opened 2024-09-21
Improve perf on TLB path


broskoTT:
Investigate perf drawback on a path of writing to device.
@joelsmithTT created a proposal for TLBWindows which should improve the perf here, although there is already tt::Writer added to umd which essentially does this, but there is still room for improved design: https://docs.google.com/document/d/1uvvLeV1-wT9masV8T2yQpX8xe-1bh5uFIlM-N4dh3S8/edit#heading=h.x1vn32gnpys7

I think that write_* interfaces could also be improved, there are many unordered_maps which degrade perf on each write. Metal folks also mentioned this issue here: https://docs.google.com/presentation/d/1kR7BswtHLS9J6dda2illQ_FV4IRJ2I28/edit#slide=id.p1

pjanevskiTT:
I think this will be mainly fixed by #2455

pjanevskiTT:
closed by #2455 , UMD doesn't do overlapping transactions with memcpy any more

PR #179, opened 2024-10-17
TLB map setup per chip

broskoTT:
Currently tt_SiliconDevice accepts one map for everything. But that goes against our effort to separate cluster vs chip responsibilities, related to #157 
This change includes:
- Adding chip_id to setup_core_to_tlb_map. There should be a map per mmio chip
- Started a chip api tests file
- Added an api example/test on how TLB setup functions at the moment. This also does some minor configuration testing, whether it throws or not.
- Minor cosmetic changes to other API tests

Contributes to #154 since it adds more api tests.
This change will require tt_metal changes.

- Breaks existing usages of setup_core_to_tlb_map
- tt_metal corresponding PR: https://github.com/tenstorrent/tt-metal/pull/13949
- tt_debuda change: Not used

### DIFF
--- a/device/tt_device.h
+++ b/device/tt_device.h
@@ -282,9 +282,10 @@ class tt_device
     /**
      * Give UMD a 1:1 function mapping a core to its appropriate static TLB (currently only support a single TLB per core).
      *
+     * @param logical_device_id MMIO chip being targeted.
      * @param mapping_function Function which maps core to TLB index.
      */
-    virtual void setup_core_to_tlb_map(std::function<std::int32_t(tt_xy_pair)> mapping_function) {
+    virtual void setup_core_to_tlb_map(const chip_id_t logical_device_id, std::function<std::int32_t(tt_xy_pair)> mapping_function) {
         throw std::runtime_error("---- tt_device::setup_core_to_tlb_map is not implemented\n");
     }
 
@@ -633,7 +634,7 @@ class tt_SiliconDevice: public tt_device
     virtual void set_driver_eth_interface_params(const tt_driver_eth_interface_params& eth_interface_params_);
     virtual void configure_tlb(chip_id_t logical_device_id, tt_xy_pair core, std::int32_t tlb_index, std::int32_t address, uint64_t ordering = TLB_DATA::Posted);
     virtual void set_fallback_tlb_ordering_mode(const std::string& fallback_tlb, uint64_t ordering = TLB_DATA::Posted);
-    virtual void setup_core_to_tlb_map(std::function<std::int32_t(tt_xy_pair)> mapping_function);
+    virtual void setup_core_to_tlb_map(const chip_id_t chip_id, std::function<std::int32_t(tt_xy_pair)> mapping_function);
     virtual void configure_active_ethernet_cores_for_mmio_device(chip_id_t mmio_chip, const std::unordered_set<tt_xy_pair>& active_eth_cores_per_chip);
     virtual void start_device(const tt_device_params &device_params);
     virtual void assert_risc_reset();
@@ -661,7 +662,7 @@ class tt_SiliconDevice: public tt_device
     /**
      * If the tlbs are initialized, returns a tuple with the TLB base address and its size
      */
-    std::optional<std::tuple<uint32_t, uint32_t>> get_tlb_data_from_target(const tt_xy_pair& target);
+    std::optional<std::tuple<uint32_t, uint32_t>> get_tlb_data_from_target(const tt_cxy_pair& target);
     /**
      * This API allows you to write directly to device memory that is addressable by a static TLB
      */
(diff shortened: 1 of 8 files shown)


joelsmithTT:
I will give this PR a more thorough review, but my initial reaction is that the relationship between TLB index and (x,y) location is something UMD should hide.

UMD has to know about the TLB windows, there's no way around that.  But the application does not.  It shouldn't matter what TLB is mapped where as long as UMD/KMD manage these resources effectively.

broskoTT:
> I will give this PR a more thorough review, but my initial reaction is that the relationship between TLB index and (x,y) location is something UMD should hide.
> 
> UMD has to know about the TLB windows, there's no way around that. But the application does not. It shouldn't matter what TLB is mapped where as long as UMD/KMD manage these resources effectively.

I agree, but bear in mind that this is not what this PR is about. This is not a final API, it is just pointing in the right direction, that TLB setup is chip specific. We might end up removing it altogether in the end

device/tt_device.h
```diff
@@ -623,7 +623,7 @@ class tt_SiliconDevice: public tt_device
     virtual void set_driver_eth_interface_params(const tt_driver_eth_interface_params& eth_interface_params_);
     virtual void configure_tlb(chip_id_t logical_device_id, tt_xy_pair core, std::int32_t tlb_index, std::int32_t address, uint64_t ordering = TLB_DATA::Posted);
     virtual void set_fallback_tlb_ordering_mode(const std::string& fallback_tlb, uint64_t ordering = TLB_DATA::Posted);
-    virtual void setup_core_to_tlb_map(std::function<std::int32_t(tt_xy_pair)> mapping_function);
+    virtual void setup_core_to_tlb_map(const chip_id_t chip_id, std::function<std::int32_t(tt_xy_pair)> mapping_function);
```
joelsmithTT, line 626: `logical_device_id`?

device/tt_device.h
```diff
@@ -801,7 +802,8 @@ class tt_SiliconDevice: public tt_device
     std::map<chip_id_t, std::unordered_map<std::int32_t, std::int32_t>> tlb_config_map = {};
     std::set<chip_id_t> all_target_mmio_devices;
     std::unordered_map<chip_id_t, std::vector<uint32_t>> host_channel_size;
-    std::function<std::int32_t(tt_xy_pair)> map_core_to_tlb;
+    std::unordered_map<chip_id_t, std::function<std::int32_t(tt_xy_pair)>> map_core_to_tlb_per_chip = {};
+    std::unordered_map<chip_id_t, bool> tlbs_init_per_chip = {};
```
joelsmithTT, line 806: These are keyed by logical id, which can refer to a remote chip.  But the notion of inbound PCIe TLB has no meaning for a remote chip.

device/tt_silicon_driver.cpp
```diff
@@ -932,8 +932,8 @@ void tt_SiliconDevice::write_device_memory(const void *mem_ptr, uint32_t size_in
 
     std::int32_t tlb_index = 0;
     std::optional<std::tuple<std::uint64_t, std::uint64_t>> tlb_data = std::nullopt;
-    if(tlbs_init) {
-        tlb_index = map_core_to_tlb(tt_xy_pair(target.x, target.y));
+    if(tlbs_init_per_chip[target.chip]) {
+        tlb_index = map_core_to_tlb_per_chip[target.chip](tt_xy_pair(target.x, target.y));
```
joelsmithTT, line 936: This is going to impose a performance penalty.

joelsmithTT (approved):
I understand the intent behind the change, but it expands an existing pattern that I regard as a core UMD design flaw: the pervasive use of hash tables for per-chip information lookup.  This pattern - which exists in UMD independent of this change - undermines performance and has led to workarounds (`get_fast_pcie_static_tlb_write_callable` and `tt::Writer`).

As an interim change along a pathway to a better API, I have no objection.

broskoTT:
> - undermines performance and has led to workarounds (`get_fast_pcie_static_tlb_write_callable` and `tt::Writer`).

I completely agree. But it is fine for me to have a more flexible API for slower access, and a performant option (such as tt::Writer, or as you suggested TLBWindow, which I ended up suggesting be AbstractIO in the end).

On the other hand I do want to investigate maps that we use a bit. Theoretically, it shouldn't be expensive to use a hash map. It is known to have a bad implementation in stdlib though, there are much better options. I also suspect that just changing everything to map and set would improve perf (they should be more performant for very small containers, which are ours). Hash maps are not slow by design, and I'd very much like to get to the bottom of where the perf issue really is.

device/tt_device.h
broskoTT: Yes, there will be only local chip ids here. Do you mean I should make this more explicit?

device/tt_silicon_driver.cpp
broskoTT: I know, but I'm also ready to argue it's a small one. Do you suggest something to be changed here or is this just a general comment?

joelsmithTT: Just a general comment.  The Metal guys did some measurements and found that UMD's write path takes an order of magnitude longer than the actual write to the device, mainly on account of all of the map lookups.  I don't have any numbers handy, but from memory I want to say the write itself is ~30 nanoseconds and the map lookups put this code path into the low hundreds of nanoseconds.  Obviously this depends on the system...

In principle a hash table lookup is fast.  But relative to a 4 byte write across PCIe, they start to add up.  There's a lot of them.

Not suggesting you change anything.  Just pointing it out, as this area of UMD has received some scrutiny from its consumers in the past.

device/tt_device.h
joelsmithTT: It's more a general remark.  This isn't the only place where a container looks like it contains per-chip data, but in reality only contains per-PCIe-chip data.  That is probably self-evident to people who have worked on the code for a while, but not so obvious to someone that doesn't have the context.

If you can think of something to clarify, feel free to put it in.

But I am pointing this situation mainly because it'd be nice to eliminate this pattern in the long run.

broskoTT: Yes, I'm thinking about that. To my current understanding, only IO is on the critical perf path, and we do plan to provide "custom" solution for that.

pjanevskiTT (approved):

device/tt_silicon_driver.cpp
broskoTT: Will be addressed when the time comes as part of #68

device/tt_device.h
broskoTT: Added a comment for now. Maybe there will be a better way to explicitly state this in the future.

PR #180, opened 2024-10-17
Versim setup and VCS fixes

vtangTT:
Changes for Versim setup:
- modify `tt_SimulationDevice` start and close device
- update the flatbuffer used to communicate with the backend

Changes for VCS fixes:
- forgot to modify `write_to_device` signature when I did my initial UMD cleanup
- revert `test_simulation_device` to not use `test::utils::read_from_device` since it's using simulation not silicon device


broskoTT:
Do we have any kind of CI that tests simulation device?

vtangTT:
> Do we have any kind of CI that tests simulation device?

There could be for Versim, however it's not possible for VCS since only some machines have the licenses

device/simulation/tt_simulation_device_generated.h
```diff
@@ -21,36 +21,33 @@ struct DeviceRequestResponseBuilder;
 enum DEVICE_COMMAND : int8_t {
```
pjanevskiTT, line 21: This file seems to be generated from flatbuffer file `tt_simulation_device.fbs `. Do we even want to push the generated file to remote? I would assume that if you can always generate it from fbs we would just want to push the fbs to the repo

tests/simulation/test_simulation_device.cpp
```diff
@@ -38,7 +38,7 @@ TEST_P(LoopbackAllCoresParam, LoopbackSingleTensix){
     tt_cxy_pair core = {0, GetParam()};
 
     device->write_to_device(wdata.data(), wdata.size()*sizeof(uint32_t), core, 0x100, "");
-    test_utils::read_data_from_device(*device, rdata, core, 0x100, wdata.size()*sizeof(uint32_t), "");
+    device->read_from_device(rdata.data(), core, 0x100, wdata.size()*sizeof(uint32_t), "");
```
pjanevskiTT, line 41: You are reading to `rdata`, but the size of the read is expressed as `wdata.size()`. I assume these are the same size but it would be more readable if we would have `rdata.size()`.

tests/simulation/test_simulation_device.cpp
```diff
@@ -50,7 +50,7 @@ bool loopback_stress_size(std::unique_ptr<tt_SimulationDevice> &device, tt_xy_pa
     std::vector<uint32_t> rdata(wdata.size(), 0);
 
     device->write_to_device(wdata.data(), wdata.size()*sizeof(uint32_t), tt_cxy_pair{0, core}, addr, "");
-    test_utils::read_data_from_device(*device, rdata, tt_cxy_pair{0, core}, addr, wdata.size()*sizeof(uint32_t), "");
+    device->read_from_device(rdata.data(), tt_cxy_pair{0, core}, addr, wdata.size()*sizeof(uint32_t), "");
```
pjanevskiTT, line 53: same as above

tests/simulation/test_simulation_device.cpp
```diff
@@ -80,8 +80,8 @@ TEST_F(SimulationDeviceFixture, LoopbackTwoTensix){
     device->write_to_device(wdata1.data(), wdata1.size()*sizeof(uint32_t),  core1, 0x100, "");
     device->write_to_device(wdata2.data(), wdata2.size()*sizeof(uint32_t), core2, 0x100, "");
 
-    test_utils::read_data_from_device(*device, rdata1, core1, 0x100, wdata1.size()*sizeof(uint32_t), "");
-    test_utils::read_data_from_device(*device, rdata2, core2, 0x100, wdata2.size()*sizeof(uint32_t), "");
+    device->read_from_device(rdata1.data(), core1, 0x100, wdata1.size()*sizeof(uint32_t), "");
```
pjanevskiTT, line 83: same as above

tests/simulation/test_simulation_device.cpp
```diff
@@ -80,8 +80,8 @@ TEST_F(SimulationDeviceFixture, LoopbackTwoTensix){
...
+    device->read_from_device(rdata2.data(), core2, 0x100, wdata2.size()*sizeof(uint32_t), "");
```
pjanevskiTT, line 84: same as above

pjanevskiTT:
Left a few nit comments. As a general comment, I am not sure whether we have some kind of docs (at least I am not aware of it) on how to setup and run our simulation devices. I think this would be very useful. You don't have to add it in this PR, but you can maybe open the issue to add simulation docs and slowly build it when you have time. If there are already some docs on it, just link it somehow through main README of the repo

tests/simulation/test_simulation_device.cpp
vtangTT: Yep they are the same size as these are just simple loopback tests.

vtangTT:
> Do we have any kind of CI that tests simulation device?

No. It is possible to set one up to test Versim, but not possible for VCS as I don't think we have access to the machines that have VCS licenses on CI

vtangTT:
> Left a few nit comments. As a general comment, I am not sure whether we have some kind of docs (at least I am not aware of it) on how to setup and run our simulation devices. I think this would be very useful. You don't have to add it in this PR, but you can maybe open the issue to add simulation docs and slowly build it when you have time. If there are already some docs on it, just link it somehow through main README of the repo

Yes there is a backend gitlab repo that contains instructions on how to run the simulator. I will link those somewhere in our readme.

tests/simulation/test_simulation_device.cpp
vtangTT: Changed to use `rdata.size()`

tests/simulation/test_simulation_device.cpp
vtangTT: fixed ty

tests/simulation/test_simulation_device.cpp
vtangTT: fixed ty

tests/simulation/test_simulation_device.cpp
vtangTT: fixed ty

device/simulation/tt_simulation_device_generated.h
vtangTT: The fbs is also pushed in the same repo. The reason I have the generated file pushed is because it gets rid of another step needed to setup the simulator. 

The flatbuffer structure itself shouldn't be changing often, so I don't expect this file to get updated often by anyone but me.

README.md
```diff
@@ -73,6 +73,11 @@ add_subdirectory(<path to umd>)
 apt install ./umd-dev-x.y.z-Linux.deb 
 ```
 
+## Simulator Integration
```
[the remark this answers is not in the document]
pjanevskiTT: Great thanks. We might want to move Tenstorrent internal docs out of main README, but you don't have to do it here. I added a comment in #206, so we will tackle that in the future. Thanks for the docs again, it looks super useful

device/simulation/tt_simulation_device_generated.h
vtangTT: The build flow will generate the `flatc` executable and then it needs to be manually run on the `.fbs` file to generate the header file. 

Actually, this could probably be made into a custom cmake target so someone can target it right after. Will look further into this.

vtangTT: I will do this in a separate PR as it solely involves the build flow.

https://github.com/tenstorrent/tt-umd/pull/225

pjanevskiTT (approved):

PR #183, opened 2024-10-18
get_pcie_base_addr_from_device per chip

broskoTT:
This attribute is tied to a chip. Although it could be argued that currently we only support a cluster of chips of same architecture, therefore this would be the same for all chips in a single cluster.
Still, this makes sense and further unblocks supporting multiple architectures in a single cluster/driver. It will also allow it to be lowered to TTDevice class which will is arch specific (currently architecture_implementation).

Related to #157 , contributes to #154 .

- Breaks existing usages of get_pcie_base_addr_from_device
- tt_metal corresponding PR: https://github.com/tenstorrent/tt-metal/pull/13949
- tt_debuda change: Not used

### DIFF
--- a/device/mockup/tt_mockup_device.hpp
+++ b/device/mockup/tt_mockup_device.hpp
@@ -87,7 +87,7 @@ class tt_MockupDevice : public tt_device {
     void* host_dma_address(std::uint64_t offset, chip_id_t src_device_id, uint16_t channel) const override {
         return nullptr;
     }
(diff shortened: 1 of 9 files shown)


joelsmithTT (approved):

abhullar-tt (approved):

broskoTT:
Related https://github.com/tenstorrent/tt-metal/pull/13949 got approved, so now merging this PR.

PR #186, opened 2024-10-18
Clang-format in CI

broskoTT:
From this PR, our CI will start enforcing clang-formatted code.
There is a .github/clang_format_repo.sh script available, which can also be used offline to format your files.

I've taken this script and part of .clang-format from budabackend, where I did previous effort on enabling clang. For more info, you can see https://yyz-gitlab.local.tenstorrent.com/tenstorrent/budabackend/-/issues/2660

See this for the difference between the old .clang-format and the new one. Note that most of the rules in the old one are default GoogleStyle rules. https://yyz-gitlab.local.tenstorrent.com/tenstorrent/budabackend/-/issues/2660#difference-between-clang-rules

The only difference from the agreed .clang-format from budabackend is whether braces are split into newline or not. I know that this can become a heated argument in this company :)

I intentionally disabled formatting for most of the code to reduce impact of this PR. Will introduce formatted code gradually in the following PRs. I only enabled it in couple of files in common/ directory.

These are the changes that would be implemented at the moment: https://github.com/tenstorrent/tt-umd/commit/304f102721518c8f593bd7f02a1cbca8e0c7a22c

TBD: Add instructions in README how to add this as a post commit check.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,147 +1,26 @@
----
-Language:        Cpp
-# BasedOnStyle:  Google
-AccessModifierOffset: -1
+BasedOnStyle: Google
+ColumnLimit: 120
+IndentWidth: 4
+
+# This will make access modifiers (public/protected/private) sit on the same indentation as `class` keyword
+AccessModifierOffset: -4
+
+# Arguments, parameters and construction initializer are broken as following:
+# - Try to fit everything into single line (controlled by ColumnLimit).
+# - If it doesn't fit, break immediately after open bracket (in case of arguments and parameters)
+#   or after colon in case of constructor initializers.
+# - Try to fit everything else into the second line.
+# - If it doesn't fit in second line, then each argument, parameter or initializer will sit in its own line.
 AlignAfterOpenBracket: AlwaysBreak
-AlignConsecutiveAssignments: false
-AlignConsecutiveDeclarations: false
-AlignEscapedNewlines: Left
-AlignOperands:   true
-AlignTrailingComments: true
-AllowShortBlocksOnASingleLine: false
-AllowShortCaseLabelsOnASingleLine: true
-AllowShortFunctionsOnASingleLine: All
-AllowShortLoopsOnASingleLine: true
-AlwaysBreakAfterDefinitionReturnType: None
-AlwaysBreakAfterReturnType: None
-AlwaysBreakBeforeMultilineStrings: true
-AlwaysBreakTemplateDeclarations: Yes
 BinPackArguments: false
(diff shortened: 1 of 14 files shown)


joelsmithTT (commented):
I am opposed to CI rejecting a change if it doesn't meet some aesthetic requirement.

broskoTT:
> I am opposed to CI rejecting a change if it doesn't meet some aesthetic requirement.

Without a CI requirement, there is no mechanism to force unified format of code. Which helps with readability.

Why are you opposed? You can setup everything locally so that it happens automatically, and you wouldn't have a need to do anything manually. I'll add this info to the README. Think of it as code syntax, you have to follow it, and it is good that it is standardised and that it's there because it improves readability. Usually it resolves some ambiguity but a lot of times it is also just aesthetic. These are just some additional requirement with (almost) no additional work to comply with them.

joelsmithTT:
An example...

https://github.com/tenstorrent/tt-umd/commit/304f102721518c8f593bd7f02a1cbca8e0c7a22c#diff-683d784e9eaa806f7993e168215b7e3dfbfe2a8973bb9c9fd1003107db298ffeL17

Screenshot 2024-10-18 at 2 25 16 PM

In this case, I think it makes readability worse.  Unclear to me if the clang-format is intelligent enough to leave such code alone.

UMD's formatting is a mess.  We can and should use clang-format to clean it up.  But we can do that without CI enforcing an empty diff on clang-format output.

I know this sort of stuff is important to lots of developers, so I am not going to obstruct any attempt to impose this.  Just don't think it's necessary.

blozano-tt:
I usually see clang-format ran using pre-commit.

Using that method would save us from writing a custom script.

broskoTT:
> In this case, I think it makes readability worse.

@joelsmithTT This is an argument over whether formatting is fine or not. I'd encourage everybody to use // clang-format off and // clang-format on, whenever such formatting is undesirable. And we should find all such cases through code review which will format all the code.

However, this is not an argument not to force some agreed upon formatting on PRs, so I still don't see why shouldn't we do it and how would it not make our lives easier.



> I usually see clang-format ran using pre-commit.

I'd be very interested to see this, I didn't know if it was possible. I thought that you have to do something manually to enforce commit-related hooks in your local repository? I asked this question also on your other cmake PR

broskoTT:
Deprecated by #238 which adds instructions, #227 which formats common folder, #203 which runs pre-commit in CI and adds default .clang-format files from this PR and adds clang-formatter in pre-commit config.

PR #188, opened 2024-10-18
CMake Overhaul

blozano-tt:
# Summary of Changes
- Project name changed from umd_device to umd (who knows you could have other targets some day
- umd_device target renamed to device (alias target umd_device is there for legacy support, but umd::device preferred)
- libdevice's dependencies have been organized between private and public
- device target is now installed and exported
- All header files are designated part of install
- added support for find_package
- added feature to generate debian development package installer
- added toolchain file for building with clang (Currently used by default)
- If compiler is Clang, default to using LLVM libc++ instead of libstdc++
- General reorganization (tucked away some things into cmake/ folder
- Add pre-commit hook for formatting CMake files
- Due to rewrite of core cmake files, I applied the formatter in this same PR to the files that were touched.
- Update README.md

# Adaptation needed in tt-metal
https://github.com/tenstorrent/tt-metal/pull/14045


README.md
```diff
@@ -45,7 +43,7 @@ If your project has CMake support, simply add this repo as a subdirectory:
 ```
 add_subdirectory(<path to umd>)
 ```
-You can then use `libdevice.so` by linking against the `umd_device` target wheverever is needed.
+You can then use `libdevice.so` by linking against the `umd::device` target wheverever is needed.
 ```
-target_link_libraries(tt_metal PUBLIC umd_device)
+target_link_libraries(tt_metal PUBLIC umd::device)
```
afuller-TT, line 48: I'm guessing most consumers should be pulling it in as a private dep, and be conscious about making it public.

```suggestion
target_link_libraries(tt_metal PRIVATE umd::device)
```

cmake/compilers.cmake
```diff
@@ -25,22 +36,31 @@ function(CHECK_COMPILERS)
...
+    message(STATUS "Using specifed C++ compiler: ${CMAKE_CXX_COMPILER}")
+    message(STATUS "Using specifed C compiler: ${CMAKE_C_COMPILER}")
+else()
+    message(STATUS "No C or C++ compiler specified, defaulting to Clang-17")
+    FIND_AND_SET_CLANG17()
+endif()
+
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    message(
+        STATUS
```
afuller-TT, line 62: If we're saying it's a warning, let's issue a warning :)

```suggestion
        WARNING
```

device/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
+include(GNUInstallDirs)
```
afuller-TT, line 1: Why not in the root CMake so any other (future) libs can use defaults, too?

device/CMakeLists.txt
```diff
@@ -15,18 +16,50 @@ set(UMD_DEVICE_SRCS
...
+    device
+    PUBLIC
+        $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
+        $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}>
+        $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}/device>
+)
+
+# flatbuffers is public - exposed to tt_metal by tt_simulation_device_generated.h
+# nng is public - exposed to tt_metal by tt_simulation_host.hpp
```
afuller-TT, line 34: Yay!  At least we're honest and transparent about it now!  \o/

device/CMakeLists.txt
```diff
@@ -15,18 +16,50 @@ set(UMD_DEVICE_SRCS
...
+target_link_libraries(
+    device
+    PUBLIC
+        nng
+        flatbuffers
```
afuller-TT, line 39:
```suggestion
        FlatBuffers::FlatBuffers
```

blozano-tt: https://github.com/google/flatbuffers/blob/807adb73b20f0afd1cfe92c576339ac8d21592a7/CMakeLists.txt#L716-L722

Google did not do a good job ^

FlatBuffers::FlatBuffers is not a real target, its a couple include directories, and causes problems when we need to export our `libdevice.so` as it is not a part of any export set.

The only way I could get this working was to directly link the target flatbuffers.

I also had to ensure we build it.

We were trying to use flatbuffers as a header only library for some reason, which doesn't look to be officially supported by flatbuffer CMake infra.

device/CMakeLists.txt
blozano-tt, line 1:
```suggestion
```

device/CMakeLists.txt
afuller-TT: They have a FlatBuffers target AND a flatbuffers target?  srsy?  🙈

.github/workflows/run-tests.yml
```diff
@@ -92,6 +92,11 @@ jobs:
 
     - name: Run arch-specific UMD unit tests
       run: |
+        pwd
+        export LD_LIBRARY_PATH=$(pwd)/build/lib:$LD_LIBRARY_PATH
```
blozano-tt, line 96: @afuller-TT @broskoTT I don't understand why suddenly I need this hack

But looks like with this I can get the tests running.

afuller-TT: That suggests we're trying to use something that we haven't told CMake that we're using.  Which library is failing to be found?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,5 @@
+repos:
+- repo: https://github.com/BlankSpruce/gersemi
+  rev: 0.16.2
+  hooks:
+  - id: gersemi
```
broskoTT, line 5: Where does this get executed? This won't configure my local git repository, right?

README.md
```diff
@@ -40,12 +38,38 @@ export CMAKE_C_COMPILER=/usr/bin/gcc
...
+
+## From Source (CMake)
+You can link `libdevice.so` by linking against the `umd::device` target.
+
+### Using CPM Package Manager
+```
+CPMAddPackage(
+  NAME umd
+  GITHUB_REPOSITORY tenstorrent/tt-umd
+  GIT_TAG blozano/cmake_magic_0
```
pjanevskiTT, line 59: Leaving comment just as a note that we probably don't want to keep GIT_TAG to this branch when we merge it

.pre-commit-config.yaml
blozano-tt: This will run on local machine if you enable it.

It would get enabled if you do
```
pre-commit install
(modify a CMakeLists.txt file)
git add that_file
pre-commit run
```

I think it can also run when you create the commit.
If you don't want the changes it suggests, you can do git commit --no-verify

pre-commit hooks are becoming quite standard in most repos these days.

README.md
blozano-tt: Good point. I will change it now.

blozano-tt, line 59:
```suggestion
  GIT_TAG v0.1.0
```

blozano-tt:
Closing in favor of #192

Issue #161, opened 2024-10-15
Deduplicate Architecture Enum


broskoTT:
We have tt::ARCH and tt::umd::architecture.
This is unnecessary.

Issue #158, opened 2024-10-15
Remove unsupported architectures


joelsmithTT:
A0 Wormhole and Jawbridge are referenced in the code.  Consider removing these.

pjanevskiTT:
With this change `WORMHOLE_B0` will be changed to `WORMHOLE`

PR #189, opened 2024-10-21
ARCH code cleanup

pjanevskiTT:
Fixes https://github.com/tenstorrent/tt-umd/issues/161 and https://github.com/tenstorrent/tt-umd/issues/158.
Cleanup code related to arch of TT cards. Remove duplicated enums and eliminate architectures that we don't support

Small fix for [Blackhole PCI id](https://github.com/tenstorrent/tt-umd/pull/189/files#diff-0f8e378b7962e2d493add39786fc5c9fdd4f795f13a812171c425ba46faa224cR90) included

### tt-metal build 

branch with successful build https://github.com/tenstorrent/tt-metal/tree/pjanevski/bump_umd_arch

### TODO

- [ ] Do we want to remove `wormhole_b0` annotation since we don't really support A0? If yes, just have Wormhole instead of Wormhole_b0


pjanevskiTT:
As per offline discussion with @broskoTT , I will split this into multiple PRs since some things would be nice to check in faster than others (it also follows PR per issue)

PR #196, opened 2024-10-22
Fix BH unit test

pjanevskiTT:
FIx BH unit test caused by removing tlb mapping from `tt_SiliconDevice` constructor. 
BH tests were using different strings. We should strive towards removing this as fast as possible.


broskoTT (approved):
I confirm the tests now pass on BH machine

Issue #102, opened 2024-10-03
Harvesting in SoCDescriptor


broskoTT:
Related to https://docs.google.com/drawings/d/1-m1azdsBqMA0A6ATYRMfkhyeuOJuGCEI62N5a96LXj0/edit?usp=sharing

SoCDescriptor holds anything related to chip's placement, locations , etc. Harvesting logic should be moved to it.
SoCDescriptor's constructor should accept Arch and harvesting info.
Try to remove harvesting related functions from tt_SiliconDriver. 
Ones which are still needed should be accessed through tt_SiliconDriver::get_soc_descriptor()

Issue #100, opened 2024-10-03
SoCDescriptor


broskoTT:
Related to https://docs.google.com/drawings/d/1-m1azdsBqMA0A6ATYRMfkhyeuOJuGCEI62N5a96LXj0/edit?usp=sharing
This should not be something which is given to UMD, rather provided by UMD.
Remove passing socDescriptor to constructor of tt_SiliconDriver.
Add default SoCDescriptors (from tt-metal repo) to UMD repo. Construct one of these based on arch type reported by TTDevice/PCIDevice class.

See if some methods from metal_SocDescriptor make sense to be moved here.

broskoTT:
Per chat on the main channel, make sure to leave the variant with passing a custom soc descriptor in case it is needed for some custom testing

broskoTT:
@pjanevskiTT per chat from today we've had. If you're enthusiastic, you can try to embed default soc_descriptor files while building, using "file" and "configure_file" commands from cmake to embed the contents of soc_desciptors directly into c++ code. Then the actual binary won't need files in the right place alongside it.

pjanevskiTT:
https://github.com/tenstorrent/tt-umd/issues/314 separate issue for paths to soc descriptors, this one was mostly about removal of soc descriptor from Cluster constructor, so closing it since it was resolved by #277

PR #202, opened 2024-10-22
Soc descriptor harvesting

pjanevskiTT:
Fix https://github.com/tenstorrent/tt-umd/issues/102, part of a general effort around soc descriptor https://github.com/tenstorrent/tt-umd/issues/100

Goal of the PR is to move harvesting logic to soc descriptor, with clear interface for coordinate translation

### Finished TODOs

- [x] Derive logical, virtual and physical coordinates from tt_xy_pair
- [x] Make soc descriptor tests arch independent as much as possible
- [x] Cleanup soc desc code for harvesting on y or x based on the arch
- [x] Revert commented tests
- [x] Move open UMD doc about harvesting to UMD repo
- [x] Add tests for grayskull
- [x] Full interface to convert between cordinate system

### First class TODOs (will finish in this PR)

- [x] Blackhole to translated coordinates fix
- [x] cpp nits

### UMD client changes

No changes needed for bumping

### TODOs with follow up issues (won't fix in this PR)

- [ ] Cleanup coordinate system docs - issue #217 
- [ ] Cover non-worker cores (DRAM, eth, arc) - issue https://github.com/tenstorrent/tt-umd/issues/219
- [ ] Compile time polymorphism for coordinate types - issue https://github.com/tenstorrent/tt-umd/issues/221
- [ ] Document CoordinateManager classes - issue #216 
- [ ] Confirm that `std::map` for mapping physical coordinates doesn't have big perf penalty -issue https://github.com/tenstorrent/tt-umd/issues/222
- [ ] Provide tt_SiliconDevice API with all types of coordinates - issue https://github.com/tenstorrent/tt-umd/issues/220


tests/blackhole/test_soc_descriptor.cpp
```diff
@@ -0,0 +1,137 @@
...
+// For the full grid of logical coordinates we expect that there are no duplicates of physical coordinates.
+// For the reverse mapping back of physical to logical coordinates we expect that same logical coordinates are returned as from original mapping.
+TEST(SocDescriptorBH, SocDescriptorLogicalPhysicalMapping) {
+    tt_SocDescriptor soc_desc = tt_SocDescriptor(test_utils::GetAbsPath("tests/soc_descs/blackhole_140_arch_no_eth.yaml"), 5);
+
+    std::map<tt_logical_coords, tt_physical_coords> logical_to_physical;
+    std::set<tt_physical_coords> physical_coords_set;
+    tt_xy_pair worker_grid_size = soc_desc.worker_grid_size;
+
+    std::size_t num_harvested_x = 2;
```
abhullar-tt, line 81: could you add a test that varies which two physical columns are harvested?

device/tt_soc_descriptor.h
```diff
@@ -123,24 +123,35 @@ class tt_SocDescriptor {
...
     int dst_size_alignment;
-    int packer_version;
     int worker_l1_size;
     int eth_l1_size;
     bool noc_translation_id_enabled;
     uint64_t dram_bank_size;
 
+    // Vectors used to translate logical coordinates to physical and translated coordinates.
+    std::map<std::size_t, std::size_t> physical_y_to_logical_y;
+    std::map<std::size_t, std::size_t> physical_x_to_logical_x;
```
abhullar-tt, line 134: do you think the maps could be changed to vectors for faster look ups?

abhullar-tt:
is the idea for tt_xy_pair to have an additional field to denote coordinate type (logical, virtual, physical, etc?)

pjanevskiTT:
> is the idea for tt_xy_pair to have an additional field to denote coordinate type (logical, virtual, physical, etc?)

I am probably going to derive each type of coordinates from tt_xy_pair, so instead of

```cpp
using tt_logical_coords = tt::umd::tt_xy_pair;

```
we are going to have derive `tt_logical_coords` from `tt_xy_pair`, so you still have only x and y as fields, but you can use coordinates as types, you will know which coordinates you are using just based on the type of the variable. Also, we want to be sure to not use the API in a way it is not intended, we can provide API for each type of coordinates. Does that make more sense?

device/tt_xy_pair.h
```diff
@@ -12,3 +12,15 @@
...
+using tt_chip_physical_coords = tt::umd::cxy_pair;
+
+using tt_logical_coords = tt::umd::xy_pair;
+using tt_chip_logical_coords = tt::umd::cxy_pair;
+
+using tt_virtual_coords = tt::umd::xy_pair;
+using tt_chip_virtual_coords = tt::umd::cxy_pair;
+
+using tt_translated_coords = tt::umd::xy_pair;
+using tt_chip_translated_coords = tt::umd::cxy_pair;
```
joelsmithTT, line 26: Can you add documentation that explains what these mean?

joelsmithTT (commented):
I do not have an accurate mental model for the coordinate systems.. I suspect I am not the only person with this problem.  I think that clear definitions of what each of the coordinate systems represent (perhaps with some examples) could go a long way for helping developers build an accurate understanding of coordinates and harvesting.

abhullar-tt:
> > is the idea for tt_xy_pair to have an additional field to denote coordinate type (logical, virtual, physical, etc?)
> 
> I am probably going to derive each type of coordinates from tt_xy_pair, so instead of
> 
> ```c++
> using tt_logical_coords = tt::umd::tt_xy_pair;
> ```
> 
> we are going to have derive `tt_logical_coords` from `tt_xy_pair`, so you still have only x and y as fields, but you can use coordinates as types, you will know which coordinates you are using just based on the type of the variable. Also, we want to be sure to not use the API in a way it is not intended, we can provide API for each type of coordinates. Does that make more sense?

yes that makes sense, thanks!

device/tt_xy_pair.h
pjanevskiTT: So I have copied over docs from open-umd. I will go through it to refactor it a bit maybe, but in general all coordinate systems are described https://github.com/tenstorrent/tt-umd/blob/pjanevski/soc_desc_harvesting/docs/coordinate_systems.md

device/tt_soc_descriptor.h
pjanevskiTT: In theory I could, but physical coordinates are not continuous in general, like logical. That would make some vector elements unused, which is not a big problem but I think with map it is more clear that physical coordinates are not continuous. I get that you are considered for perf, but I think this is not going to have a big impact. Map is going to be small and it is using integers as keys so it shouldn't be a problem.
However, I want to confirm this so I could write a small test on a separate branch with both vector and map to confirm that perf penalty is not big / doesn't exist

abhullar-tt: I think that should be okay, you are right in that there are gaps and handling that would be awkward.

device/coordinate_manager.h
```diff
@@ -0,0 +1,76 @@
...
+class CoordinateManager {
+
+public:
+    CoordinateManager(const tt_xy_pair& worker_grid_size, const std::vector<tt_xy_pair>& workers, std::size_t harvesting_mask)
+        : worker_grid_size(worker_grid_size), workers(workers), harvesting_mask(harvesting_mask) {}
+
+    virtual void perform_harvesting(std::size_t harvesting_mask);
+
+    virtual tt_physical_coords logical_to_physical_coords(tt_logical_coords logical_coords);
+    virtual tt_translated_coords logical_to_translated_coords(tt_logical_coords logical_coords);
```
broskoTT, line 29: Consider naming this to_translated_coords(tt_logical_coords logical_coords); , e.g. without the "from" coordinates. Same for all functions. The to_translated_coords would then have 3 overrides for example.

Consider, and if you think this is better then feel free to leave it as it is.

docs/coordinate_systems.md
```diff
@@ -0,0 +1,361 @@
...
+* As part of product definition, translation tables are enabled on Wormhole devices regardless of their grid size
+
+# Coordinates System Definitions
+## Physical Coordinates - Any Arch
+These are the NOC coordinates that the hardware understands, there are two distinct variations for NOC0 and NOC1.  In hardware, each node is given an ID (which is different for each NOC) which can be used to identify this node.  In the SOC descriptor, physical coordinates are specified for NOC0. 
+
+### Important Notes 
+* The coordinates are **NOT** contiguous for tensix cores -- In the example shown, ethernet corresponds to physical coords `[*-6]`:  this row is not on in the grid of worker cores
+* The coordinates are statically assigned to each "node" regardless of harvesting
+* `grayskull` are listed in `[y-x]` while `wormhole_b0` are listed in `[x-y]`
```
broskoTT, line 45: Wow is this really true? We parse them from yaml differently?

tests/api/test_soc_descriptor_gs.cpp
```diff
@@ -0,0 +1,149 @@
...
+        // Expect that reverse mapping of physical coordinates gives the same logical coordinates
+        // using which we got the physical coordinates.
+        EXPECT_EQ(it.first, logical_coords);
+    }
+}
+
+// Test logical to virtual coordinate translation.
+// For the full grid of logical coordinates we expect that there are no duplicates of virtual coordinates.
+// For the reverse mapping back of virtual to logical coordinates we expect that same logical coordinates are returned as from original mapping.
+TEST(SocDescriptor, SocDescriptorGSLogicalVirtualMapping) {
```
broskoTT, line 119: There is a lot of common code between these tests.

Do you think it might be clearer to have everything in a single test_soc_descriptor.cpp file?
Then, for example, this test would have a foreach over three different files, but the rest of the code remains almost the same.

For other tests you might need some constant maps maybe (which map from file to expected values), but I still think it could be cleaner. For example first_worker_core_physical_location in  SocDescriptorTopLeftCore and max_num_harvested_x in tests with harvesting... 

You could also make it explicit that grayskull doesn't support harvesting (EXPECT_THROW), supports default harvesting mask 0 only.

tests/api/test_soc_descriptor_gs.cpp
```diff
@@ -0,0 +1,149 @@
+#include "gtest/gtest.h"
```
broskoTT, line 1: Would also love to see a test (which would be arch agnostic, it would run differently on each chip but pass) in api_chip which just creates a driver, and fetches soc_desc and then does something with some coords and calls some driver API (maybe simple write_device).
But you probably need the other PR done which removes requirement for soc_desc in constructor, right? Maybe this test would then belong in that PR.

device/blackhole/blackhole_coordinate_manager.cpp
```diff
@@ -0,0 +1,14 @@
+#include "blackhole_coordinate_manager.h"
+
+std::set<std::size_t> BlackholeCoordinateManager::get_x_coordinates_to_harvest(std::size_t harvesting_mask) {
```
patrickroberts, line 3: Can we use [`std::bit_set`](https://en.cppreference.com/w/cpp/utility/bitset) for this?

device/coordinate_manager.cpp
```diff
@@ -0,0 +1,184 @@
...
+    if (arch == tt::ARCH::GRAYSKULL) {
+        return new GrayskullCoordinateManager(worker_grid_size, workers, harvesting_mask);
+    } else if (arch == tt::ARCH::WORMHOLE_B0) {
+        return new WormholeCoordinateManager(worker_grid_size, workers, harvesting_mask);
+    } else if (arch == tt::ARCH::BLACKHOLE) {
+        return new BlackholeCoordinateManager(worker_grid_size, workers, harvesting_mask);
+    } else {
+        throw std::runtime_error("Invalid architecture for coordinate manager");
+    }
+}
```
patrickroberts, line 184: Consider returning a `std::unique_ptr<CoordinateManager>`?

patrickroberts: Also
```cpp
switch (arch) {
    case tt::ARCH::GRAYSKULL:
        return ...
    ...
}

throw ...
```
might be less error-prone. I believe there's a clang-tidy rule to check that all enumerated cases are handled if the enum ever grows.

device/tt_soc_descriptor.h
```diff
@@ -162,16 +163,38 @@ class tt_SocDescriptor {
...
+    tt_translated_coords virtual_to_translated_coords(tt_virtual_coords virtual_coords);
+
+    tt_logical_coords translated_to_logical_coords(tt_translated_coords translated_coords);
+    tt_physical_coords translated_to_physical_coords(tt_translated_coords translated_coords);
+    tt_virtual_coords translated_to_virtual_coords(tt_translated_coords translated_coords);
+
+    void perform_harvesting(std::size_t harvesting_mask);
+
+private:
+    CoordinateManager* coordinate_manager = nullptr;
```
patrickroberts, line 196: I don't see where this is `delete`'d, which is why I suggested `std::unique_ptr<CoordinateManager>`

tests/test_utils/soc_desc_test_utils.hpp
```diff
@@ -0,0 +1,21 @@
...
+
+static std::size_t get_num_harvested(std::size_t harvesting_mask) {
+    std::size_t num_harvested = 0;
+    while (harvesting_mask > 0) {
+        if (harvesting_mask & 1) {
+            num_harvested++;
+        }
+        harvesting_mask >>= 1;
+    }
+    return num_harvested;
```
patrickroberts, line 18:
```suggestion
    const std::bit_set<sizeof(harvesting_mask) * CHAR_BIT> harvesting_bits{harvesting_mask};
    return harvesting_bits.count();
```

tests/api/test_soc_descriptor_gs.cpp
pjanevskiTT: Completely makes sense, thought the same thing. I will write it when I complete removal of soc descriptor

pjanevskiTT: Follow up https://github.com/tenstorrent/tt-umd/issues/212

device/coordinate_manager.h
abhullar-tt: I like that suggestion. It would reduce the callers conditional checks based on the src coordinate type

pjanevskiTT: done

tests/api/test_soc_descriptor_gs.cpp
pjanevskiTT: I thought about this when writing these tests. I don't really have problems with a lot of common code in tests. I think it would be much easier to debug when in the future some tests fail, since one test is going to be only for one arch. I can try to write some common functions which are arch agnostic so I can have as much code as possible in one common place, with that I agree. Also, I think it is cleaner to have separate files because you are actually testing different features (row harvestign, column harvesting) and you can write that in the comments to make the test clearer. Let me know how you feel about this

pjanevskiTT:
I know this PR is becoming a bit hard to review with a lot of code, so I will separate TODOs left into first class that I will solve in this PR and then merge it, and into second class tasks that I will open issues for and solve it all as separate PRs. I tried thinking about making this PR smaller as well, but I think this is a good base for harvesting so we can build on top of it. Uplifting this is also going to be seamless since it doesn't change or removes the API, it just adds the functions, so I can make the API function for tt-metal in the following commits.

@broskoTT @joelsmithTT let me know if this is ok for you

tests/test_utils/soc_desc_test_utils.hpp
pjanevskiTT: @joelsmithTT any reasoning behind thumbs down for this? @patrickroberts is bit_set more performant here?

joelsmithTT: It's less readable.  The loop is obvious.  If brevity is important, you can say `return __builtin_popcountll(bits); // counts the 1 bits`

device/blackhole/blackhole_coordinate_manager.cpp
joelsmithTT: What problem does that solve?

joelsmithTT (approved):
If you want to get this in and refine later, that is good with me.

But I maintain the view that it is too difficult to establish a correct understanding of the coordinate schemes from the information available in the repo.  Furthermore, I am suspicious that the coordinate situation is overly complex and that there is opportunity for simplification.

tests/soc_descs/grayskull_10x12.yaml
```diff
@@ -16,20 +16,18 @@ eth:
...
+    1-1,  2-1,  3-1,  4-1,  5-1,  6-1,  7-1,  8-1,  9-1, 10-1, 11-1, 12-1,
+    1-2,  2-2,  3-2,  4-2,  5-2,  6-2,  7-2,  8-2,  9-2, 10-2, 11-2, 12-2,
+    1-3,  2-3,  3-3,  4-3,  5-3,  6-3,  7-3,  8-3,  9-3, 10-3, 11-3, 12-3,
+    1-4,  2-4,  3-4,  4-4,  5-4,  6-4,  7-4,  8-4,  9-4, 10-4, 11-4, 12-4,
+    1-5,  2-5,  3-5,  4-5,  5-5,  6-5,  7-5,  8-5,  9-5, 10-5, 11-5, 12-5,
+    1-7,  2-7,  3-7,  4-7,  5-7,  6-7,  7-7,  8-7,  9-7, 10-7, 11-7, 12-7,
+    1-8,  2-8,  3-8,  4-8,  5-8,  6-8,  7-8,  8-8,  9-8, 10-8, 11-8, 12-8,
+    1-9,  2-9,  3-9,  4-9,  5-9,  6-9,  7-9,  8-9,  9-9, 10-9, 11-9, 12-9,
+    1-10, 2-10, 3-10, 4-10, 5-10, 6-10, 7-10, 8-10, 9-10, 10-10, 11-10, 12-10,
+    1-11, 2-11, 3-11, 4-11, 5-11, 6-11, 7-11, 8-11, 9-11, 10-11, 11-11, 12-11
```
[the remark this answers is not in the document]
pjanevskiTT: If I align the columns, is it ok to keep the change? This was changed in order to have same representation for gs, wh and bh soc descriptors

tests/test_utils/soc_desc_test_utils.hpp
pjanevskiTT: Went with `__builtin_popcount`

device/blackhole/blackhole_coordinate_manager.cpp
pjanevskiTT: Went with `__builtin_popcount` for `std::size_t`

docs/coordinate_systems.md
pjanevskiTT: Removed the docs, going to address it in a separate PR (#223) when I rewrite it properly. I will resolve this comment there

tests/soc_descs/grayskull_10x12.yaml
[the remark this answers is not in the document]
joelsmithTT: Sure

device/tt_soc_descriptor.cpp
```diff
@@ -166,7 +165,127 @@ void tt_SocDescriptor::load_core_descriptors_from_device_descriptor(YAML::Node &
...
+tt_physical_coords tt_SocDescriptor::logical_to_physical_coords(tt_logical_coords logical_coords) {
+    // log_assert(logical_coords.x < logical_x_to_physical_x.size());
+    // log_assert(logical_coords.y < logical_y_to_physical_y.size());
+    return tt_physical_coords(logical_x_to_physical_x[logical_coords.x], logical_y_to_physical_y[logical_coords.y]);
+}
+
+// TODO(pjanevski): is it enough just to add 18 to the logical coordinates
+// in order to get the translated coordinates?
+tt_translated_coords tt_SocDescriptor::logical_to_translated_coords(tt_logical_coords logical_coords) {
+    static const std::size_t translated_offset = 18;
```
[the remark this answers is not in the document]
pjanevskiTT: Implemented in c97a22cb387486604bd9577c0f5c7a4ac396792d

broskoTT (approved):

PR #203, opened 2024-10-23
Add pre-commit in CI

blozano-tt:
It runs pre-commit checks on the files that have been changed (not the whole repo).

I setup the workflow to generate a PR targeting the PR branch, with the required fixes to satisfy the linters/formatters.

Developers can review that PR, and accept/reject/modify the proposed changes.

In this pre-commit PR, I added a dummy change to a CMake file, and the CMake formatter made a change.

The workflow generated this PR: https://github.com/tenstorrent/tt-umd/pull/205

What do you think?

This way we can add all of our linters/formatters/tidy-ers directly in the pre-commit-config.yaml

I was thinking it would be cool if it could comment or notify people about that PR, but that’s just icing.

### DIFF
--- a/.github/workflows/on-pr.yml
+++ b/.github/workflows/on-pr.yml
@@ -31,10 +31,25 @@ jobs:
 
       - name: Run Pre-commit and Fix Issues
         uses: pre-commit/action@v3.0.1
+        with:
+          extra_args: ['--from-ref', 'origin/main', '--to-ref', 'HEAD']
+        continue-on-error: true
+
+      - name: Check for Modified Files and Set Flag
+        id: git_diff
+        run: |
+          git diff --exit-code || echo "::set-output name=changes-detected::true"
 
       - name: Create Pull Request with Fixes
+        if: steps.git_diff.outputs.changes-detected == 'true'
         uses: peter-evans/create-pull-request@v5
         with:
           commit-message: "Fix pre-commit issues"
           title: "Propose pre-commit fixes"
           body: "This PR contains proposed fixes from pre-commit hooks."
+
+      - name: Fail the Job if Changes Were Made
+        if: steps.git_diff.outputs.changes-detected == 'true'
+        run: |
+          echo "Pre-commit made changes. Marking job as failed."
+          exit 1  # Fail the job explicitly if changes were made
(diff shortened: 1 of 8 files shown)


.github/workflows/on-pr.yml
```diff
@@ -16,3 +16,49 @@ jobs:
...
+      - name: Run Pre-commit and Fix Issues
+        uses: pre-commit/action@v3.0.1
+        with:
+          extra_args: "--from-ref origin/main --to-ref HEAD"
+        continue-on-error: false
+
+      # Declarations from here irrelevant
+      # We are failing PRs on pre-commit error
+      # Nothing below this comment will have impact
+      # Leaving this incase we want this feature later
```
broskoTT, line 44: I don't get it. We will run all these steps in case the pre-commit passes? Why?

.github/workflows/on-pr.yml
```diff
@@ -16,3 +16,49 @@ jobs:
...
+        with:
+          fetch-depth: 0  # Fetch all history so 'origin/main' is available
+          ref: ${{ github.head_ref }}
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.11
+
+      - name: Run Pre-commit and Fix Issues
```
broskoTT, line 35: I see this in logs:

gersemi..............................................(no files to check)Skipped
clang-format.........................................(no files to check)Skipped

Can you try to submit a wrongly formatted cmake and/or cpp file, and see if the PR will be blocked?

.github/workflows/on-pr.yml
blozano-tt: No, but I just deleted the noise to make it more clear what happens.

.github/workflows/on-pr.yml
blozano-tt: Already did this earlier in this pr history

blozano-tt: It will work, I guarantee

blozano-tt: Just added change to C++

broskoTT: Ah, looks like it checks only touched cpp files.
Thanks for checking!

broskoTT (approved):

PR #210, opened 2024-10-24
Update tests/CMakeLists.txt

blozano-tt:
We intend to use the header only version of fmt for consistency within this project, and across tt-metal side


broskoTT:
Do you want to add the public "uv" change to this PR as well?

broskoTT (approved):

PR #214, opened 2024-10-26
Remove DEBUG define in CMakeLists.txt

blozano-tt:
This define is not consumed by UMD code.
This was useless.
Remove noise, and possible duplicate definition.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,11 +56,6 @@ endif()
 
 message(STATUS "UMD build type: ${CMAKE_BUILD_TYPE}")
 
-add_compile_definitions(
-    $<$<CONFIG:Debug>:DEBUG>
-    $<$<CONFIG:RelWithDebInfo>:DEBUG>
-)
-
 include(dependencies)
 
 add_subdirectory(device)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,13 +6,13 @@ set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED True)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON) # This also impacts dependencies brought in through CPM
 
-if (DEFINED ENV{CMAKE_C_COMPILER} AND DEFINED ENV{CMAKE_CXX_COMPILER})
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -56,11 +56,6 @@ endif()
 
 message(STATUS "UMD build type: ${CMAKE_BUILD_TYPE}")
 
-add_compile_definitions(
```
broskoTT, line 59: Does this affect the debug build, i.e. would I still be able to build for debugging? I previously used this flag I think: `-DCMAKE_BUILD_TYPE=Debug`

CMakeLists.txt
```diff
@@ -6,13 +6,13 @@ set(CMAKE_CXX_STANDARD 17)
...
 set(CMAKE_POSITION_INDEPENDENT_CODE ON) # This also impacts dependencies brought in through CPM
 
-if (DEFINED ENV{CMAKE_C_COMPILER} AND DEFINED ENV{CMAKE_CXX_COMPILER})
+if(DEFINED ENV{CMAKE_C_COMPILER} AND DEFINED ENV{CMAKE_CXX_COMPILER})
     message(STATUS "Setting C and C++ compiler from environment variables")
     set(CMAKE_C_COMPILER $ENV{CMAKE_C_COMPILER})
     set(CMAKE_CXX_COMPILER $ENV{CMAKE_CXX_COMPILER})
 endif()
 
-if (CMAKE_CXX_COMPILER AND CMAKE_C_COMPILER)
```
broskoTT, line 15: Just wondering why this wasn't picked up previously with our cmake formatter...

broskoTT (approved):

CMakeLists.txt
broskoTT: Or will the formatter check only when we touch the files for the first time?

blozano-tt: If a poorly formatted file made it in to the repo, it will only be corrected the next time it is touched.

I must have not had pre-commit turned on in CI when I last checked in CMakeLists.txt

CMakeLists.txt
blozano-tt: No it has no impact on debug build. This PR just removes an unused #define.

This looks like it was mistakenly copied from tt-metal at some point.

Issue #187, opened 2024-10-18
tt_simulation_host does not abstract away implementation details


blozano-tt:
We are currently exposing our implementation details to consumers.

This header file is included in a `tt_metal` file.

nng is a public dependency, because we don't separate interface and implementation.

Can we make this more opaque? @vtangTT 

@broskoTT are you already planning for this?

https://github.com/tenstorrent/tt-umd/blob/f7b1ce0f6ed3101bdd4f1367145378a59996f07b/device/simulation/tt_simulation_host.hpp#L7-L8

vtangTT:
Will look into this next week.

broskoTT:
I didn't think about it until recently (the whole simulation class). Will bear this in mind.
@blozano-tt  Did you come to a consensus with the metal folks on layout? Should it be include/public_headers.h and src/private_headers. and src/private_impl.cpp and src/public_impl.cpp?

vtangTT:
https://github.com/tenstorrent/tt-umd/pull/224 closes this issue.

PR #224, opened 2024-10-28
Hide simulation device/host impl details

vtangTT:
Issue: https://github.com/tenstorrent/tt-umd/issues/187

Changes include: 
- Forward declare `nng_socket` and `nng_dialer`, changing them into pointers and propagating this change to not expose `nng.h` in `tt_simulation_host.hpp`
- remove `create_flatbuffer` and `print_flatbuffer` as a member function in `tt_simulation_device` so now no need to expose flatbuffers in `tt_simulation_device.hpp`
- hijacked a quick simulation test fix onto this PR

### DIFF
--- a/device/simulation/tt_simulation_host.hpp
+++ b/device/simulation/tt_simulation_host.hpp
@@ -3,13 +3,14 @@
 // SPDX-License-Identifier: Apache-2.0
 #include <vector>
 #include <cstdint>
-
-#include <nng/nng.h>
-#include <nng/protocol/pair1/pair.h>
+#include <memory>
 
 #include "device/tt_xy_pair.h"
 
 #define NNG_SOCKET_PREFIX "ipc:///tmp/"
+
+typedef struct nng_socket_s nng_socket;
+typedef struct nng_dialer_s nng_dialer;
 
 class tt_SimulationHost {
 public:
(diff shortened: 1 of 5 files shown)


device/simulation/tt_simulation_host.hpp
```diff
@@ -20,6 +20,6 @@ class tt_SimulationHost {
     void send_to_device(uint8_t *buf, size_t buf_size);
     size_t recv_from_device(void **data_ptr);
 private:
-    nng_socket host_socket;
-    nng_dialer host_dialer;
+    nng_socket *host_socket;
+    nng_dialer *host_dialer;
```
blozano-tt, line 24: Can you use `std::unique_ptr<nng_socket>` to avoid having to worry about memory management?

blozano-tt (approved):
LGTM! Thank you!

device/simulation/tt_simulation_host.hpp
vtangTT: 🫡

device/simulation/tt_simulation_host.hpp
```diff
@@ -4,13 +4,13 @@
 #include <vector>
 #include <cstdint>
```
blozano-tt, line 5:
```suggestion
#include <memory>
```

include-what-you-use for safety :-)

device/simulation/tt_simulation_host.hpp
```diff
@@ -4,13 +4,13 @@
...
```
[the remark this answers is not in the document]
vtangTT: ty 🙏

PR #225, opened 2024-10-29
Generate fbs header file through cmake during build flow

vtangTT:
As discussed here: https://github.com/tenstorrent/tt-umd/pull/180#discussion_r1808311946

Old setup involved building the `flatc` executable and then manually running it to generate the header file. 

Now it is integrated into the cmake build flow and `tt_simulation_device_generated.h` will be generated as a dependency of `device` target.


device/CMakeLists.txt
```diff
@@ -24,6 +28,8 @@ add_library(device SHARED ${UMD_DEVICE_SRCS})
 add_library(${PROJECT_NAME}::device ALIAS device)
 add_library(${PROJECT_NAME}_device ALIAS device) # For legacy I guess
 
+add_dependencies(device flatbuffer_generated_header)
```
vtangTT, line 31: Not sure if this is something I should be doing? But I'm also unsure how to get this to work otherwise..  :(

.gitignore
```diff
@@ -9,3 +9,6 @@ compile_commands.json
 .vscode/
 *.log
 *.csv
+
+# Generated flatbuffer header for simulation device
+device/simulation/tt_simulation_device_generated.h
```
afuller-TT, line 14: Normally all generated files (binary and text) live in the build directory.  Is there something that forces us to generate a file into our source tree?

vtangTT: I just have it generating into the source tree right now, but it should definitely go into build directory now that I think of it. Will change.

Not sure if this will affect anything else, as I will now have to `target_include_directories` for `${CMAKE_BINARY_DIR}`? And it seems like it's installing all headers into a certain place?

device/CMakeLists.txt
afuller-TT: Try this:

```
target_sources(device
  PRIVATE
    path/to/generated/header.h
)
```
If the header is part of the API, then use `PUBLIC` instead.
With CMake this old you may still need to keep the `add_custom_target`, or we may be able to drop it.  I'm not certain.

.gitignore
afuller-TT: That's correct, we (and if the header is public, then others, too) will need to include the destination inside the bindir in the search path.  This is normal and expected.  And one of several reasons why we should be cognizant of which headers are public vs. which are private.

device/CMakeLists.txt
vtangTT: Works tyty 🙏

.gitignore
vtangTT: Will be private, waiting for https://github.com/tenstorrent/tt-umd/pull/224

vtangTT:
This PR will be ready to go out after https://github.com/tenstorrent/tt-umd/pull/224 that privatizes nng and fbs headers

.gitignore
pjanevskiTT, line 14: nit comment here. Maybe it would be nicer to have `device/simulation/generated/` directory, so we can gitignore whole dir. This would be bulletproof for future generated files. Not requesting this to be changed, you can consider it if you want. If not, I am totally fine with that as well

pjanevskiTT (approved):
Great changes. Approving mostly for build flow, I think this is much cleaner now. You can wait on other approvals regarding detailed cmake changes

.gitignore
vtangTT: Woops, I changed this to generate into the build folder, so this .gitignore change is not needed anymore. Will change.

vtangTT: fixed

afuller-TT (approved):
Lovely!

Issue #47, opened 2024-09-10
Linter


broskoTT:
Add clang, or some other linter support.
Take a look at how metal does it.
In any case, make sure to be compatible with tt-metal.

broskoTT:
We can use "yamllint", and it is already present in tt-metal's image.
"clang-format" is also present.
Figure out if there's a way to also do cmake-format.

broskoTT:
Also make sure to check this https://yyz-gitlab.local.tenstorrent.com/tenstorrent/budabackend/-/issues/2660
Information from there can be used to reduce .clang-format, while maintaining the same rules. It's probably good to keep the same .clang-format rules as tt-metal for consistency.

pjanevskiTT:
@broskoTT can we close this?

PR #227, opened 2024-10-30
Linting sweep with gersemi

vtangTT:
Running gersemi over entire repo as it hasn't been done yet. No functional changes, just linting


broskoTT:
Thanks for doing this, and appologies for the hassle!

broskoTT (approved):

broskoTT:
Related to #47

.gersemirc
```diff
@@ -1,6 +1,6 @@
 # yaml-language-server: $schema=https://raw.githubusercontent.com/BlankSpruce/gersemi/master/gersemi/configuration.schema.json
 
-definitions: [cmake]
+definitions: [cmake, cmake/stubs]
 indent: 4
 line_length: 80
```
joelsmithTT, line 5: Should we make this 120?  It's not merged yet, but https://github.com/tenstorrent/tt-umd/pull/186 has 120 for C/C++.

joelsmithTT (commented):
The changes to logger.hpp etc look fine but I am a bit puzzled why a CMake linter is touching C++?

afuller-TT (approved):
Thanks for doing this, @vtangTT .

Looks like you ran all of pre-commit against the whole repo; ie: rather than `pre-commit run gersemi -a` you ran `pre-commit run -a`.  In that case you've swept both CMake formatting with gersemi, and also C++ formatting with clang-format.  Could you update the title accordingly?

.gersemirc
vtangTT: I have no opinion on this.. @afuller-TT @blozano-tt does it matter?

afuller-TT: I put 120 in `tt-metal` https://github.com/tenstorrent/tt-metal/blob/main/.gersemirc#L5

And @blozano-tt put 80 here just 'cause that's what the default is.  Go for it!

vtangTT: changed to 120

broskoTT (approved):

Issue #229, opened 2024-10-30
Cannot detect multiple devices on 2x p150 machine


abhullar-tt:
yyzo-bh-01 is 2x p150 but running `tt_SiliconDriver::detect_available_device_ids()` only returns one device. Not sure if this is a UMD issue or something wrong with the machine.

broskoTT:
I'm detecting 2 devices.
`ls /dev/tenstorrent` also returns 2 devices
This was mu reservation command for ird, did you indeed take 2 cards?
`ird reserve --docker-image ghcr.io/tenstorrent/tt-metal/tt-metalium/ubuntu-20.04-amd64:v0.52.0-rc4-dev --timeout max blackhole --machine yyzo-bh-01 --num-pci 2`

Note that create-ethernet-map still doesn't work for blackhole, so somewhere in the code we might just assume 1 card in the cluster. But enumerate_device, which just lists /dev/tenstorrent seems to be working.

Can you double check, and if it still doesn't work please provide steps for reproducing the issue

abhullar-tt:
Ah user error, I didn't add `--num-pci 2` last night when I tried this. But I confirmed that create-eth-map does not work for BH

broskoTT:
Will be supported after #235

PR #230, opened 2024-10-30
Fix BH tests

broskoTT:
We still don't have CI for BH, so we easily forget to test if all tests (including new API tests) still work for Blackhole.
I was testing something unrelated on a blackhole machine and found these disrepancies.
Related to #154 and #229


tests/api/test_chip.cpp
```diff
@@ -44,12 +44,16 @@ inline std::unique_ptr<tt_ClusterDescriptor> get_cluster_desc() {
...
-    // TODO: remove getting manually cluster descriptor from yaml.
-    std::string yaml_path = test_utils::GetClusterDescYAML();
+    // TODO: Remove different branch for different archs
     std::unique_ptr<tt_ClusterDescriptor> cluster_desc;
     if (device_arch == tt::ARCH::GRAYSKULL) {
         cluster_desc = tt_ClusterDescriptor::create_for_grayskull_cluster(pci_device_ids_set, pci_device_ids);
+    } else if (device_arch == tt::ARCH::BLACKHOLE) {
+        std::string yaml_path = test_utils::GetAbsPath("blackhole_1chip_cluster.yaml");
+        cluster_desc = tt_ClusterDescriptor::create_from_yaml(yaml_path);
     } else {
```
pjanevskiTT, line 54: would explicit 

```cpp
if (device_arch == tt::ARCH::WORMHOLE_B0)
```

be more readable here? Same for all other changes

pjanevskiTT (approved):

tests/blackhole/test_silicon_driver_bh.cpp
```diff
@@ -229,10 +229,9 @@ TEST(SiliconDriverBH, UnalignedStaticTLB_RW) {
                 // Statically mapping a 1MB TLB to this core, starting from address NCRISC_FIRMWARE_BASE.  
                 device.configure_tlb(i, core, get_static_tlb_index_callback(core), l1_mem::address_map::NCRISC_FIRMWARE_BASE);
             }
+            device.setup_core_to_tlb_map(i, get_static_tlb_index_callback);
         } 
     }
-
-    device.setup_core_to_tlb_map(get_static_tlb_index_callback);
```
joelsmithTT, line 235: How did this compile?

joelsmithTT (approved):
does create-ethernet-map not work for BH?

tests/blackhole/test_silicon_driver_bh.cpp
broskoTT: It didn't compile before, we don't check for blackhole :) Actually I can add a build for blackhole, even though we don't have blackhole cards to run on.

tests/api/test_chip.cpp
broskoTT: Well actually not. The else branch which is following is not "something specific for wormhole", rather it is a "standard" way of doing things. Grayskull and Blackhole have something specific done to them. 
To be specific, in one of the following PRs (specifically when cem starts working for BH) I'd expect to remove the blackhole branch. Maybe in some of the following changes to remove Grayskull branch as well.

PR #185, opened 2024-10-18
Checking for spdx header in CI

broskoTT:
This was copied from tt-mlir: https://github.com/tenstorrent/tt-mlir/blob/main/.github/workflows/spdx.yml

This job check whether we forgot this header somewhere.

PR #231, opened 2024-10-30
spdx headers

broskoTT:
Related to #47 
Adding pre-commit check for SPDX headers. 
Deprecates #185 

### API Changes
No API changes introduced


.check-copyright-config.yaml
```diff
@@ -0,0 +1,44 @@
+DEFAULT:
```
broskoTT, line 1: Will switch branch to point to main once the other PR is checked in

.check-copyright-config.yaml
```diff
@@ -0,0 +1,44 @@
...
+    # SPDX-License-Identifier: {license}
+
+  # comment lines matching:
+  # SPDX-FileCopyrightText: year[-year] Espressif Systems
+  # or
+  # SPDX-FileContributor: year[-year] Espressif Systems
+  # are replaced with this template prefixed with the correct comment notation (# or // or *) and SPDX- notation
+  espressif_copyright: '{years} Tenstorrent Inc.'
+
+# # You can create your own rules for files or group of files
```
pjanevskiTT, line 28: nit: this is under comment twice

device/tt_io.hpp
```diff
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: (c) 2024 Tenstorrent Inc.
+// SPDX-FileCopyrightText: (c) 2023 Tenstorrent Inc.
```
joelsmithTT, line 1: Why? This file didn't exist in 2023.

joelsmithTT (approved):

blozano-tt (approved):

device/tt_io.hpp
broskoTT: I don't know what is our policy on this. Just added 2023 everywhere.
 Should we update these each year? Or are they supposed to hold the year the file was first created?

joelsmithTT: I think it's supposed to be the latter but I'm not sure.  Probably doesn't matter

PR #233, opened 2024-10-30
fix tt_mockup_device

mbezuljTT:
minimal set of changes to get tt_mockup_device compile again

### DIFF
--- a/device/mockup/tt_mockup_device.hpp
+++ b/device/mockup/tt_mockup_device.hpp
@@ -21,7 +21,7 @@ class tt_MockupDevice : public tt_device {
     virtual ~tt_MockupDevice() {}
 
     // Setup/Teardown Functions
-    std::unordered_map<chip_id_t, tt_SocDescriptor>& get_virtual_soc_descriptors() override {
+    virtual std::unordered_map<chip_id_t, tt_SocDescriptor>& get_virtual_soc_descriptors() override {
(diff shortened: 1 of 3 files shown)


device/tt_silicon_driver.cpp
```diff
@@ -2538,7 +2538,7 @@ void tt_SiliconDevice::set_membar_flag(const chip_id_t chip, const std::unordere
                     cores_synced.insert(core);
                 }
                 else {
-                    log_info(LogSiliconDriver, "Waiting for core {} to recieve mem bar flag {} in function", core.str(), barrier_value);
+                    log_trace(LogSiliconDriver, "Waiting for core {} to recieve mem bar flag {} in function", core.str(), barrier_value);
```
broskoTT, line 2541: This broke tt_metal CI, we already fixed it :)

device/mockup/tt_mockup_device.hpp
```diff
@@ -21,7 +21,7 @@ class tt_MockupDevice : public tt_device {
     virtual ~tt_MockupDevice() {}
```
broskoTT, line 21: Can we add building mockup device to CI? How do you build it? Can you maybe add a mockup test to tests/api?

mbezuljTT: added very simple test.

PR #234, opened 2024-10-30
Deprecate set_driver_host_address_params

blozano-tt:
Deprecate unnecessary API by introducing default initialization for the effected struct, and adding a namespace for the involved constants.

I made the new namespace inline so that existing code wouldn't break either in tests or in tt_metal.

### DIFF
--- a/device/tt_device.h
+++ b/device/tt_device.h
@@ -243,6 +243,7 @@ class tt_device
      *
      * @param host_address_params_ All the Host Address space parameters required by UMD.
      */ 
+    [[deprecated("Using unnecessary function.")]]
     virtual void set_driver_host_address_params(const tt_driver_host_address_params& host_address_params_) {
         throw std::runtime_error("---- tt_device::set_driver_host_address_params is not implemented\n");
     }
--- a/device/tt_silicon_driver.cpp
+++ b/device/tt_silicon_driver.cpp
@@ -48,6 +48,10 @@
 #include "tt_device.h"
 #include "ioctl.h"
 
+#include "src/firmware/riscv/grayskull/host_mem_address_map.h"
+#include "src/firmware/riscv/wormhole/host_mem_address_map.h"
+#include "src/firmware/riscv/blackhole/host_mem_address_map.h"
+
 using namespace boost::interprocess;
 using namespace tt;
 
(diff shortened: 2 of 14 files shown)


device/tt_silicon_driver.cpp
```diff
@@ -531,6 +535,22 @@ tt_SiliconDevice::tt_SiliconDevice(const std::string &sdesc_path, const std::str
             }
         }
     }
+
+    // Default initialize host_address_params based on detected arch
+    switch(arch_name) {
+        case tt::ARCH::GRAYSKULL:
+            host_address_params = {grayskull::host_mem::address_map::ETH_ROUTING_BLOCK_SIZE, grayskull::host_mem::address_map::ETH_ROUTING_BUFFERS_START};
+            break;
+
```
joelsmithTT, line 544: This may give the reader the impression that the notions of ETH block size and ETH routing buffer have meaning for Grayskull.  They do not.

blozano-tt: Would a comment to that effect suffice?
I can't really hardcode it to 0x0, 0x0 because we stuff an actual value for ETH_ROUTING_BLOCK_SIZE in grayskull.

device/tt_device.h
```diff
@@ -243,6 +243,7 @@ class tt_device
      *
      * @param host_address_params_ All the Host Address space parameters required by UMD.
      */ 
+    [[deprecated("Using unnecessary function.")]]
     virtual void set_driver_host_address_params(const tt_driver_host_address_params& host_address_params_) {
```
joelsmithTT, line 247: There is some nuance here that isn't captured but probably should be, irrespective of this change.

UMD gives the application full control over the sysmem.  By calling this method, the application is acknowledging that it knows that part of sysmem is 'off limits' because UMD and ETH firmware are using it for remote IO.

I agree that the parameters being set here can and should be UMD implementation detail.  However, I don't see any documentation that describes the effect this method has or the implications for how the application uses sysmem.

device/tt_silicon_driver.cpp
joelsmithTT: Yes, a comment is fine.

device/tt_silicon_driver.cpp
```diff
@@ -531,6 +535,22 @@ tt_SiliconDevice::tt_SiliconDevice(const std::string &sdesc_path, const std::str
             }
         }
     }
+
+    // Default initialize host_address_params based on detected arch
+    switch(arch_name) {
+        case tt::ARCH::GRAYSKULL:
```
blozano-tt, line 541:
```suggestion
        // ETH_ROUTING_BLOCK_SIZE and ETH_ROUTING_BUFFER_START have no meaning for grayskull
        case tt::ARCH::GRAYSKULL:
```

device/tt_silicon_driver.cpp
```diff
@@ -48,6 +48,10 @@
 #include "tt_device.h"
 #include "ioctl.h"
 
+#include "src/firmware/riscv/grayskull/host_mem_address_map.h"
+#include "src/firmware/riscv/wormhole/host_mem_address_map.h"
+#include "src/firmware/riscv/blackhole/host_mem_address_map.h"
```
broskoTT, line 53: We should really move all of this to TTDevice/architecture_implementation

blozano-tt: Move the include files you mean?

broskoTT: I meant, architecture_implementation should hold something like "get_host_address_params", then each specialization per arch should return it's constants.
The code in tt_SiliconDevice would then be just calling AnyPCIDevice->get_architecture_implementation()->get_host_address_params.

Then you don't need these includes here

src/firmware/riscv/blackhole/host_mem_address_map.h
```diff
@@ -5,6 +5,9 @@
 #include <cstdint>
 #include <stdint.h>
 
+// Remove inline ASAP
```
broskoTT, line 8: "inline" is there so you don't have to change other tests? E.g. constants can be used as both blackhole::host_mem::address_map::CONST and host_mem::address_map::CONST ?

broskoTT (approved):
Feel free not to address comments now, but create an issue to address them later. If it would block you.

src/firmware/riscv/blackhole/host_mem_address_map.h
blozano-tt: Correct, don't have to immediately change tests or tt_metal.
I was thinking it would make it easier to remove this API if we first deprecate, because we have so much trouble moving UMD forward in metal.

device/tt_silicon_driver.cpp
blozano-tt: I took your advice, and ran into a nasty circular header file dependency lol!

However, with the fix, I guess the code base is now better than ever!

PR #240, opened 2024-10-31
A CMake target for firmware headers

afuller-TT:
Create a target to encapsulate firmware
    
 Bonus: this pushes ARCH_NAME down closer to the source where it can be abstracted away more easily.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -7,7 +7,8 @@ add_library(test_common INTERFACE)
 target_link_libraries(
     test_common
     INTERFACE
-        umd_device
+        umd::device
+        umd::Firmware
         gtest_main
         gtest
         pthread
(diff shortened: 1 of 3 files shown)


device/CMakeLists.txt
```diff
@@ -38,12 +37,12 @@ target_link_libraries(
     PUBLIC
         nng
         flatbuffers
+        uv
```
blozano-tt, line 40: I thought I had already made this public

afuller-TT: This PR has 2 "bonus" commits.  Only the last one is my change.  If the 2 former commits have been merged to `main` and cherry-pick'd back to this branch, then this branch should merge to main so that Git sees in its commit graph that those commits are part of shared history.

afuller-TT: Actually that probably won't work if UMD follows the same rewrite-all-commit-IDs-on-merge that tt-metal does 🥴

blozano-tt:
I’m confused about the arch name changes.

do we need to adjust what is in tests/CMakeLists.txt ?

afuller-TT:
> I’m confused about the arch name changes.
> 
> do we need to adjust what is in tests/CMakeLists.txt ?

re: tests/CMake -- yes that should just link from what i see.  I'll add that.
re: confused -- what specifically?

device/CMakeLists.txt
broskoTT: Yes, please rebase to the latest main without the first two commits

tests/CMakeLists.txt
```diff
@@ -10,43 +10,19 @@ add_library(test_common INTERFACE)
...
     INTERFACE
         ${PROJECT_SOURCE_DIR}
         ${CMAKE_CURRENT_SOURCE_DIR}
 )
-if($ENV{ARCH_NAME} STREQUAL "grayskull")
-    message(STATUS "UMD: Building for Grayskull")
-    target_include_directories(
-        test_common
-        INTERFACE
-            ${PROJECT_SOURCE_DIR}/device/grayskull
```
broskoTT, line 30: Where did this (and for other archs) include go?
I see where the other one, src/firmware/riscv, landed...

broskoTT (approved):

tests/CMakeLists.txt
afuller-TT: I took it out and nothing broke 🤷  so I figured it was dead code.

device/CMakeLists.txt
afuller-TT: Done.

tests/CMakeLists.txt
blozano-tt: Maybe those files are directly included and the include paths aren’t necessary. Or they could be dead code. I’ll take a close look in a bit.

afuller-TT: Should I hold off merging this PR?  Or okay to merge since everything is green?  (Or is there something to manually trigger?)

blozano-tt (approved):
LGTM!

Issue #99, opened 2024-10-03
ClusterDescriptor


broskoTT:
Related to https://docs.google.com/drawings/d/1-m1azdsBqMA0A6ATYRMfkhyeuOJuGCEI62N5a96LXj0/edit?usp=sharing
This should not be something which is given to UMD, rather provided by UMD.
Change the API so that ClusterDescriptor can itself run create-ethernet-map and create itself. This is used by tt_SiliconDriver (look at it as Cluster class in the diagram). Replace the constructor to call ClusterDescriptor rather than get it as an argument.

broskoTT:
Remove test_utils::GetClusterDescYAML as part of this effort (or reuse it's code)

broskoTT:
I've completed the first part of not providing CEM externally. @pjanevskiTT is currently working on the other part which removes it from tt_SiliconDriver constructor.

broskoTT:
A follow up from #277 is now to reduce ClusterDescriptor's construction not to surface yaml at all.
After this task I think we can close this issue which tracked initial push of pruning cluster descriptor

PR #245, opened 2024-11-01
create-ethernet-map lib instead of bin

broskoTT:
Related to #99 

This change will simplify our build system changes by a lot. No additional artefacts needed to be shipped with libdevice, and no thinking about relative/absolute path once the lib is built.

@TTDRosen provided this library.

Changes:
- Add lib to git, add it to build system.
- Remove CEM binary from git
- Add cluster yaml creation to tt_ClusterDescriptor
- Use this interface in tests, remove deprecated GetClusterDescYAML

The following PR will remove cluster yaml altogether. Both from tt_ClusterDescriptor interface and from tt_SiliconDevice. Didn't want to overwhelm this PR.

This change requires client changes. Breaks existing references to create-ethernet-map binary
- [x] tt_metal corresponding PR: https://github.com/tenstorrent/tt-metal/pull/14718
- tt_debuda has its own CEM so it will continue to build without any changes.


joelsmithTT:
Can you or @TTDRosen add instructions for generating this new binary and perhaps consider maintaining ARM64 support in UMD (which this change breaks), and possibly adding RISC-V support?  There's community interest in the latter and I have hardware setups for both.  (I did N150 home trials with an Ampere system and although I never merged the [metal support](https://github.com/tenstorrent/tt-metal/compare/main...joelsmith/aarch64-build) it would be nice if this kept working)

TTDRosen:
I am not sure of the appropriate place to put this (someone please advise) but to reduce our bus factor.

From luwen root
`cargo build -p create-ethernet-map --release --lib`

joelsmithTT:
Great, thanks @TTDRosen.  This change is fine with me provided that `cargo` command plus a link to Luwen is placed in a comment (maybe in the CMakeLists.txt?) so anyone with a non-X86 platform can build and swap out that binary.

device/CMakeLists.txt
```diff
@@ -53,6 +53,7 @@ target_link_libraries(
         Boost::interprocess
         fmt::fmt-header-only
         yaml-cpp::yaml-cpp
+        ${CMAKE_CURRENT_SOURCE_DIR}/libs/libcreate_ethernet_map.a
```
blozano-tt, line 56: This scares me.
UMD could be built with g++/clang various versions, apparently on different architecture CPUs.
Isn't this asking for trouble?

cc: @afuller-TT 

I take it this library came from rust code due to other comments.

broskoTT: @TTDRosen 
Can we use the command you provided
`cargo build -p create-ethernet-map --release --lib`
for building arm64 and risc-v libs?

TTDRosen: Yeah, I've tested with riscv and aarch64. I just don't like distributing without sanity checking on real hardware (instead of qemu). But it's extremely trivial to cross compile as well. 

But also packages like [corrosion](https://github.com/corrosion-rs/corrosion) seem to make it pretty easy to just build the library when building umd (if we don't mind adding Rust as a dependency)

joelsmithTT: UMD's build situation has dramatically increased in complexity over the past few months, and my preference would be for it to get simpler rather than even more complicated by depending on the Rust toolchain.

My rationale is that although every incremental increase in complexity has been introduced to solve a problem, it represents additional (and occasionally unwelcome) friction for people who are using the codebase in a way that is different than the developers.  I can give specific and detailed examples, but it's not really relevant here...

My other line of reasoning is that having this create-ethernet-map logic in a separate binary isn't actually necessary: there's no reason we couldn't implement this logic in C++, in UMD, and eliminate the complexity inherent in carrying around the binaries and figuring out how to call them (or now, build/link against them).

The non-X86 platform is a bit of an edge case, so for now I think it is sufficient to document what is going on such that users on ARM/RISC-V have an extra step to get going but aren't blocked.

blozano-tt: @TTDRosen when you have time can you please point me to the source code for this library?
I am just curious how large it is, and if porting to C++ would be easy/medium/hard.

joelsmithTT: The code is at https://github.com/tenstorrent/luwen/blob/main/src/bin/luwen-cem.rs

UMD has a circular dependency that needs fixing:
1. Output from create-ethernet-map is needed to instantiate a tt_SiliconDevice
2. create-ethernet-map does remote chip IO operations
3. In UMD, remote chip IO operations require a tt_SiliconDevice

UMD design here is wrong and needs to be refactored.  UMD would benefit from a cleaner remote read/write implementation too.

broskoTT: > UMD design here is wrong

While I absolutely agree that this is an issue -> It is not our current focus.
I'm focusing on fixing our externally facing API. For that, it doesn't matter if this is a linked lib, built in UMD using rust toolchain, or code written in C++ UMD. I does matter a lot whether there's a binary that needs to ship with libdevice or not. So to list the options and which problems it solves:

1. (What I propose by this PR) Use static lib. Simple change. Removes the need to ship cem binary with the lib. Don't need to think about absolute/relative paths when libdevice within itself calls cem. 
2. Add Rust compilation step. Would need to change cmake, CI images, and adds additional dependancy. (In addition to 1.) No changes to external API (the end client doesn't see any changes). It removes the need to update lib file in UMD repo when CEM updates. More clearly points to the source code.
3. Write what CEM does in UMD in C++. Significant time to implement this change. (In addition to 1.) No changes to external API (the end client doesn't see any changes). Removes external dependancy on a lib written in another language. Probably overall improves UMD design. Code in our control vs external dependancy.

I'd probably like more if we did 3. instead of 2., but we don't have to delve deeper into that, since I don't want to do anything in addition to 1. at the moment. I can create an issue to track this and capture our current conversation on it.

To restate, I'm focusing on changes which refactor API, and purposely not focusing on changes which have no effect to the usage of API. I don't want to say that the former is more important, but it is important to narrow the focus to deliver relatively quickly.

Re: Different architectures.
I don't like the idea of degrading without reason, so let's support it if it is easy.
@TTDRosen Can you give me some instructions on how to build these lib for arm64 and riscv? Or if you can find the time to directly provide built libs, that would help a lot.

TTDRosen: Please keep in mind that I don't have a real aarch64 system to test on, and emulated systems don't do a good job of showing the processor specific issues that lead to most of the problems I've hit on aarch64. But you can grab the lib [here](https://drive.google.com/file/d/1IITsuN_ZaGzof9adloiAmuSpiyi-d1Wf/view?usp=sharing) 

It's as simple as running `cross build --target aarch64-unknown-linux-gnnu --release --lib -p create-ethernet-map`

device/libs/libcreate_ethernet_map.a
afuller-TT, line 1: A 22MB binary in VCS?  😢 

This right here is a big incentive to jump to Option (3) sooner rather than later.

broskoTT: Added them as LFS instead, thanks for the comment :)

device/CMakeLists.txt
broskoTT: Added aarch64 lib, added readme for building these libs.
We should probably add a CI job for arm build if we do want to support it.

pjanevskiTT (approved):

joelsmithTT (approved):

PR #246, opened 2024-11-01
Fix bug building UMD as submodule

blozano-tt:
Include path is given to device library as PROJECT_BINARY_DIR

Function uses CMAKE_BINARY_DIR.

This causes problems when building UMD as a submodule.


afuller-TT (approved):
SHIP IT!

PR #258, opened 2024-11-04
Remove redundant copy of noc_parameters.h

blozano-tt:
The root of truth for `noc_parameters.h` is maintained in the main branch of tt-metal.

Instead of sharding the file across two repos, we can download the current version at CMake configure time.

In the future we can move to a submodule if we can reach agreement on that.

However, this is a perfectly golden solution for now.

This PR deletes the local "copy" of `noc/noc_parameters.h`


tests/api/test_cluster.cpp
```diff
@@ -23,7 +23,7 @@
 #include "src/firmware/riscv/wormhole/eth_l1_address_map.h"
 #include "src/firmware/riscv/wormhole/host_mem_address_map.h"
 #include "src/firmware/riscv/wormhole/l1_address_map.h"
-#include "src/firmware/riscv/wormhole/noc/noc_parameters.h"
+#include "noc/noc_parameters.h"
```
afuller-TT, line 26: How does `umd` resolve this file when its the top-level?  Isn't this breaking their build?

And, uh, this file was only used by tests and not by the actual code?  🤔

blozano-tt: Right, only used for unit tests ... cause they need to simulate the behavior of tt-metal driving their APIs.

No it doesn't break the build, because of the new include path I introduced.

This test probably should not have been hardcoded for wormhole.

afuller-TT: Willing to hold my nose and get this in quickly so we can make forward progress.  Also knowing that the UMD interface is being looked at elsewhere, so this shouldn't sit around for toooo long 🤞

tests/CMakeLists.txt
```diff
@@ -21,6 +21,36 @@ target_include_directories(
...
+)
+foreach(ARCH IN LISTS ARCHS)
+    set(HW_NOC_PARAMETERS_URL
+        "https://raw.githubusercontent.com/tenstorrent/tt-metal/refs/heads/main/tt_metal/hw/inc/${ARCH}/noc/noc_parameters.h"
+    )
+    set(DESTINATION_DIR ${CMAKE_CURRENT_BINARY_DIR}/${ARCH}/noc)
+    file(MAKE_DIRECTORY ${DESTINATION_DIR})
+    get_filename_component(fileName ${HW_NOC_PARAMETERS_URL} NAME)
+    file(
+        DOWNLOAD
```
afuller-TT, line 38: I don't like the idea of introducing a download in consumers.  When it's in tt-metal, can we reference the file from the same source tree (ugly, I know) and guard this with "are we top-level?".

Both options are ugly, but one has a performance cost and a new avenue for failures.

blozano-tt: I really don't want to do the "are we top level thing".

It disrespects the independence of the two repos. Its also somewhat cyclic to do that.

It could prevent us from using CPM in the future.

afuller-TT: How do you feel about "only download if the file doesn't exist"?  It'll mean hiccups when needing to consume a newer version of the file, but at least we won't be inflicting papercuts every time CMake reconfigures.

blozano-tt: sure... on it

afuller-TT (approved):
LGTM, thanks!

Issue #106, opened 2024-10-03
Revisit device_params and driver_params


broskoTT:
See out of these four which are hardware specific and not compiler specific (what cannot be changed vs what can)
    virtual void set_device_l1_address_params
    virtual void set_device_dram_address_params
    virtual void set_driver_host_address_params
    virtual void set_driver_eth_interface_params
    
My guess is many of these can be reduced or removed.

blozano-tt:
My PR #234 addresses one of these.

blozano-tt:
https://github.com/tenstorrent/tt-umd/pull/260

PR #261, opened 2024-11-05
Optimize eth_interface

blozano-tt:
### Ticket
https://github.com/tenstorrent/tt-umd/issues/106

### Description
Deprecate set_driver_eth_interface_params API by self initializing the parameters.

### List of the changes
- Split eth_interface_params struct into noc_params and eth_interface_params
- Deprecate api to set eth_interface_params (We can delete it later)
- Default initialize eth_interface_params
- Update tests

### API Changes
- [ ] https://github.com/tenstorrent/tt-metal/pull/14779
- [ ] https://github.com/tenstorrent/tt-debuda/pull/164


blozano-tt:
@broskoTT @abhullar-tt 

I would appreciate a second pair of eyes on this item:

tt-metal was never updated to pass `CMD_ORDERED` or `CMD_BROADCAST`, so they were always default 0.

Now I have made tt-umd self initialize eth_interface params, which means I also default init those two fields of the struct to whatever is set in eth_interface.h.
https://github.com/tenstorrent/tt-umd/blob/f27b7c7fe1ae61f138c46c0f9dca8ed960ae794c/device/wormhole/wormhole_implementation.cpp#L121-L122

I am assuming this is the right thing to do, but could you confirm?

Are those fields meant to be some kind of dynamic control of UMD's behavior? or are they meant to be a constant device specific initialization?

joelsmithTT (approved):
Nice.  Looks like `set_driver_noc_params` could eventually get the same treatment?

Grayskull has no ethernet, and I think there is a way of factoring the code so that `grayskull/eth_interface.h` isn't needed and

```
tt_driver_eth_interface_params grayskull_implementation::get_eth_interface_params() const {
    throw std::runtime_error("Grayskull has no Ethernet, fix your program"); 
}
```
... although that is a more invasive change.  If left as-is please consider leaving a comment at/near `grayskull_implementation::get_eth_interface_params()` or the zero constants that explains the situation ("Grayskull has no Ethernet, so return zero constants as placeholders" or some such).  It might be self-evident to some developers, but it could be helpful information to others who are building their mental model of how this system works.

device/tt_device.h
```diff
@@ -238,6 +241,15 @@ class tt_device
         throw std::runtime_error("---- tt_device::set_device_dram_address_params is not implemented\n");
     }
 
+    /**
+     * Set NOC parameters used by UMD to communicate with the TT Device (used for remote transactions).
+     *
+     * @param noc_params_ All the NOC parameters required by UMD.
+     */
+    virtual void set_driver_noc_params(const tt_driver_noc_params& noc_params_) {
```
broskoTT, line 249: Isn't this hardware specific? We can't set custom noc parameters, it won't work, right? To be clear, I'd expect that noc_params are not needed at all.

broskoTT:
According to https://github.com/tenstorrent/tt-umd/pull/268 , please abide by the new rules of committing API breaking changes. Add this to your PR:

This PR has API changes:

device/tt_device.h
blozano-tt: @joelsmithTT also mentioned he doesn't think we should need those parameters, but on the surface they appear used:

https://github.com/tenstorrent/tt-umd/blob/dfbd4bf648d5561e627c4b876c248cc6feae75aa/device/tt_silicon_driver.cpp#L1376-L1388

Any guidance on how to remove them?

I would love to not need this new API.

joelsmithTT: Those parameters are used only in this `get_sys_addr` function -- as far as I can tell (?).  I think the simplest thing to do is define them as constants in `get_sys_addr`.  36 and 6.  If Blackhole and Wormhole ethernet firmwares diverge and they need to be different, we will have bigger problems to address than this method.

That the application is passing these in seems like unnecessary complication.

blozano-tt: @joelsmithTT thanks for the guidance!
I made the changes, and stripped out the new API.

device/tt_device.h
```diff
@@ -23,6 +23,9 @@
 #include "pcie/pci_device.hpp"
 #include "fmt/core.h"
 
+constexpr std::uint32_t NOC_ADDR_LOCAL_BITS = 36; // source: noc_parameters.h, common for WH && BH
```
pjanevskiTT, line 26: Why are we not defining these constants in a separate file, same as ethernet constants. Then we could have same things in constructor for NOC params

```cpp
noc_params = architecture_implementation->get_noc_params();
```

It is not that I am not ok with this change at the moment, but in the future we will surely move this around and if it is separated into a file then we could keep it around while redesigning everything but keep param logic there. Let me know if this makes sense

blozano-tt: I took path of least resistance. What you are saying makes sense though.

blozano-tt: I just did a quick refactor.

pjanevskiTT (approved):

PR #262, opened 2024-11-05
Enable clang for tests/ folder

broskoTT:

Removed tests/.clang-format which disabled formatting in this folder, and ran `pre-commit run --all-files`

### DIFF
--- a/tests/.clang-format
+++ b/tests/.clang-format
@@ -1,2 +1,0 @@
-DisableFormat: true
-SortIncludes: false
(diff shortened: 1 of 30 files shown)


joelsmithTT (approved):
Might be easier to drop all changes except removing tests/.clang-format and rebase?

broskoTT:
> Might be easier to drop all changes except removing tests/.clang-format and rebase?

Yes, it should be easy to rebase

PR #263, opened 2024-11-05
Enable clang for src/ folder

broskoTT:

Removed src/.clang-format which disabled formatting in this folder, and ran `pre-commit run --all-files`

### DIFF
--- a/src/.clang-format
+++ b/src/.clang-format
@@ -1,2 +1,0 @@
-DisableFormat: true
-SortIncludes: false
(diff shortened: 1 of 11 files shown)


joelsmithTT (approved):
These files show up in other projects and occasionally are edited in one place but not the other, just pointing out that effectively comparing the same file between repos will require formatting the other one.

broskoTT:
> These files show up in other projects and occasionally are edited in one place but not the other, just pointing out that effectively comparing the same file between repos will require formatting the other one.

Yes, I'm not sure about this one. We could just leave it ignored for formatting changes. Not like we do any edits to these files where standardised formatting would help.

broskoTT:
Leaving this folder as is, we won't format it for now. It is also not being modified.

PR #264, opened 2024-11-05
Enable clang in device folder

broskoTT:

Removed device/.clang-format which disabled formatting in this folder, and ran `pre-commit run --all-files`

### DIFF
--- a/device/.clang-format
+++ b/device/.clang-format
@@ -1,2 +1,0 @@
-DisableFormat: true
-SortIncludes: false
--- a/device/architecture_implementation.cpp
+++ b/device/architecture_implementation.cpp
@@ -12,10 +12,14 @@ namespace tt::umd {
 
 std::unique_ptr<architecture_implementation> architecture_implementation::create(tt::ARCH architecture) {
     switch (architecture) {
-        case tt::ARCH::BLACKHOLE: return std::make_unique<blackhole_implementation>();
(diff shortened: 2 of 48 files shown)


joelsmithTT (approved):
I think ioctl.h should be excluded, not sure how easy that is to do.

Also suggest adding DerivePointerAlignment: false, PointerAlignment: Left, ReferenceAlignment: Left (see https://tenstorrent.slack.com/archives/C07L16Z59AP/p1731099457956649?thread_ts=1731092455.581859&cid=C07L16Z59AP for discussion).

But that can always be done later

broskoTT:
> I think ioctl.h should be excluded, not sure how easy that is to do.
> 

Why though, what exactly do you find problematic? 
I think we should add //clang-format off and //clang-format on everywhere where we want to keep formatting, we can add it to whole file if that makes sense. That would make it more explicit than ignoring the file...

joelsmithTT:
> Why [exclude ioctl.h]?

It's not UMD's file -- it's copied from KMD.  We have to update it whenever KMD augments its interface [(example](https://github.com/tenstorrent/tt-kmd/pull/33)) and UMD wants to use the new feature.

I suppose its fine to reformat it... diffs between UMD's formatted versions will be consistent.  Just not diffs between UMD's version and KMD's version.  Not that big of a deal.

broskoTT:
This PR is deprecated by https://github.com/tenstorrent/tt-umd/pull/312
I've left ioctl.h in original state there.

PR #265, opened 2024-11-05
Rename tt_SiliconDevice to tt::umd::Cluster

### LINKED ISSUE #108: Rename tt_SiliconDriver to tt::umd::Cluster
Related to https://docs.google.com/drawings/d/1-m1azdsBqMA0A6ATYRMfkhyeuOJuGCEI62N5a96LXj0/edit?usp=sharing

Prior to further improvements, renames and separations: rename what is being used as a cluster of chips.

Also double check what is now left in tt_silicon_driver.cpp file (which will now be renamed to cluster.cpp), and remove stuff not related to Cluster

broskoTT:
Fixes #108 
For bigger picture, see https://docs.google.com/drawings/d/1-m1azdsBqMA0A6ATYRMfkhyeuOJuGCEI62N5a96LXj0/edit

Changes:
- Renamed tt_SiliconDevice to Cluster
- Added Cluster to tt::umd namespace.
- Renamed tt_device.h to cluster.h, and tt_silicon_driver.cpp to cluster.cpp
- Moved a few trivial functions from tt_device.cpp to cluster.cpp.

Note that tt_device will be removed, according to the diagram. Currently, simulation device and mockup device are specializations of current tt::umd::Cluster. When we create tt::umd::Chip abstraction, they will be moved correspondingly, and this tt_device abstraction will be removed.

### API Changes
This PR has API changes:
- [x] (If breaking change) tt_metal approved PR pointing to this branch: https://github.com/tenstorrent/tt-metal/pull/14752
- [x] (If breaking change) tt_debuda approved PR pointing to this branch: https://github.com/tenstorrent/tt-debuda/pull/158


joelsmithTT:
> Currently, simulation device and mockup device are specializations of current tt::umd::Cluster. When we create tt::umd::Chip abstraction, they will be moved correspondingly, and this tt_device abstraction will be removed.

Can you elaborate a bit on this?  Simulation & Mockup both inherit `tt_cluster`, as does `tt::umd::Cluster` (formerly `tt_SiliconDevice`).  Are you saying that a single `tt::umd::Cluster` will handle all cases (mockup, simulation, silicon) depending on the `tt::umd::Chip` subclasses that are instantiated at runtime?

Do you think there is value in `tt_device` -> `ICluster` or `AbstractCluster` (i.e. maintain an abstract Cluster interface for testing purposes)?

broskoTT:
@joelsmithTT 
For the first part, yes you got it right.

For the second part we can change it if we chose to implement those tests. It does make sense, we should consider it.

In any case, according to the diagram, the Cluster class will be publicly facing class, which will have ClusterInternal holding implementation (the pimpl design pattern). The ClusterInternal would be a good candidate for this.

joelsmithTT (approved):

Issue #243, opened 2024-11-01
Core coordinates prototype


pjanevskiTT:

pjanevskiTT:
Closed by #266

Issue #228, opened 2024-10-30
Rename WORKER to TENSIX in CoreType


abhullar-tt:
This is not a high priority

broskoTT:
Makes sense.

pjanevskiTT:
TENSIX enum was added to CoreType, there will be a separate PR for removing WORKER and renaming everything to TENSIX

broskoTT:
CoreType::WORKER is present at many many places in tt_metal...

pjanevskiTT:
Requires a lot of work in tt-metal, not going to work on it anytime soon, closing...

Issue #212, opened 2024-10-25
[Harvesting] Harvesting tests


pjanevskiTT:
In addition to initial harvesting tests from soc descriptor add more tests

Wait for the removal of soc desc from tt_SiliconDevice constructor to merge, as well as https://github.com/tenstorrent/tt-umd/pull/202

pjanevskiTT:
resolved by #266

PR #223, opened 2024-10-28
Coordinate systems docs

pjanevskiTT:
Write a doc about coordinate systems. Base for this document is open-umd document which explains all coordinate systems we have

### Note for reviews 

Feel free to suggest any naming conventions for these coordinate systems. Naming from the initial doc is going to be followed for now, just because it seems that this is mostly accepted across different tenstorrent products (syseng tools, tt-metal, tt-umd)

PR #266, opened 2024-11-05
Core coordinates prototype

### LINKED ISSUE #219: [Harvesting] Cover non-tensix cores
Follow up from #202 

Coordinate translation is implemented for worker (tensix) cores. Implement proper conversions (where possible) for DRAM, ETH, ARC cores

pjanevskiTT:
Implement core coordinates translation

This PR is going to close #219 #243 #228 #212 

As discussed, we are going to have `CoordSystem` enum, which is going to be used to determine the appropriate coordiante system of the coordinates that are used in the API

```cpp
enum class CoordSystem {
    LOGICAL,
    PHYSICAL,
    VIRTUAL,
    TRANSLATED,
};
```

For now, we support these 4 coordinate systems. For explanations on each coordinate system there is a pending PR to cleanup the docs properly, but understanding these coordinate systems should not be prerequisite to review this PR. It is just needed to understand that each core coordinate can be in multiple coordinate systems.

## Note for reviews

Since we have `CoordinateManager` class here, in order to better understand the use of that class it might be useful to skim the PR #202. In general it is mostly used to support the translation between coordinate systems.

The most useful parts of the PR to review are (in my opinion)

1. For tt-metal people - [test_core_coordinate.cpp file](https://github.com/tenstorrent/tt-umd/blob/pjanevski/core_coordinates/tests/api/test_core_coordinates.cpp) - these tests should represent how tt-metal (and other clients) should use coordinate types
2. For tt-umd people - everything else :) 
3. Not in 1 or 2 - then probably same as tt-metal people, since this would be API change for all other clients

## V1 coordinates

For V1 prototype, we have additional enum representing core type 

```cpp
enum class CoreType {
  ARC,
  DRAM,
  ETH,
  PCIE,
  TENSIX,
  ROUTER_ONLY,
};
```

which is used in the main struct

```cpp
 struct CoreCoord_V1 : public tt_xy_pair {
    CoreType core_type;
    CoordSystem coord_system;
 };
```

When using this, it is needed to set both the core type and coordinate system, in order for API to work properly.

## V2 coordinates (not going to be implemented, it was used for discussion)

For V2 prototype, we would have one struct for each core type that is represented using enum in V1. In PR, structs are implemented for Tensix and DRAM cores, just to see the effects of having multiple core types

```cpp
struct CoreCoord_V2 : public tt_xy_pair {
    CoordSystem coord_system;
};

struct TensixCoreCoord_V2 : public CoreCoord_V2 {};
 
struct DramCoreCoord_V2 : public CoreCoord_V2 {};
```

When using this it is needed to set just the coordinate system, core type comes from struct type.

## API changes

API is provided inside tt_SiliconDevice class for both the V1 and V2 core coordinates. Focus was on providing API for reads/writes as the example. I think looking this from PR is the best way to explain it. 

Main point to make here is that API is the same for all coordinate systems, which was one of the points that tt-metal folks made in the meeting. UMD can figure out how to properly program the TLBs and write to device for any coordinate system. In the PR we rely on translating everything to physical (NOC) coordinates

## Pros/cons

All pros/cons are my view, feel free to comment on everything

### V1

Pros
- coordinate translation logic is one place
- less API functions

Cons
- One extra step for figuring out core type on each API call
- readability
- core type can be concluded at runtime

### V2

Pros
- readability
- core type is clear from the type

Cons
- API blows up, need to provide functions for each core type

### Questions for reviewers

- Do we want to be able to get worker/DRAM cores in all coordinate systems from Soc descriptor?

### TODOs

- [ ] Think about CoreCoordRange, is that representation simply a set of CoreCoord
- [x] Finish translation for Blackhole translated coordinates (DRAM, ARC, PCIE)
- [ ] Implement L2CPU cores if needed ?
- [ ] Test custom soc descriptor coordinate manager
- [x] Rewrite tests to use CoordinateManager instead of tt_SocDescriptor


device/tt_silicon_driver.cpp
```diff
@@ -2957,3 +2959,64 @@ tt_version tt_SiliconDevice::get_ethernet_fw_version() const {
...
+void tt_SiliconDevice::read_from_device(void* mem_ptr, chip_id_t chip, CoreCoord_V1 core_coord, uint64_t addr, uint32_t size, const std::string& fallback_tlb) {
+    CoreCoord_V1 physical_core_coord = coordinate_manager->to_physical(core_coord);
+
+
+}
+
+// v2 functions
+
+// tensix core coord
+void tt_SiliconDevice::write_to_device(const void *mem_ptr, uint32_t size_in_bytes, chip_id_t chip, TensixCoreCoord_V2 core_coord, uint64_t addr) {
```
broskoTT, line 2979: In any case we will probably end up with write_to_device functions per coreType. 
I do like how V1 will hide this though, probably by having a switch case statement pointing to those functions.

device/coordinate_manager.h
```diff
@@ -45,6 +46,17 @@ class CoordinateManager {
 
     CoordinateManager(CoordinateManager& other) = default;
```
broskoTT, line 47: Since this is mostly translating CoreCoord_V1, and not providing additional functionality, maybe rename to CoreCoordTranslator, or something alike

device/tt_soc_descriptor.h
```diff
@@ -20,6 +20,7 @@
 #include "device/tt_arch_types.h"
 
 #include "device/coordinate_manager.h"
+#include "device/tt_core_coordinates.h"
```
broskoTT, line 23: We will probably have this discussion with metal folks, but my guess is they'd request from us to enrich SoCDescriptor class with getters which return vectors or maps of CoreCoords

device/tt_core_coordinates.h
```diff
@@ -0,0 +1,57 @@
...
+};
+
+// ************************************************************************************************
+// V1
+
+/*
+ * CoreType is an enum class that represents all types of cores
+ * present on the Tenstorrent chip.
+ */
+enum class CoreType {
```
pgkeller, line 31: need to be sure this is uint8 (1 byte)

device/tt_core_coordinates.h
```diff
@@ -0,0 +1,57 @@
...
+  ARC,
+  DRAM,
+  ETH,
```
pgkeller, line 34: I think we want to expose the concept of both active and idle eth from the bottom of the stack

device/tt_core_coordinates.h
```diff
@@ -0,0 +1,57 @@
...
+  PCIE,
+  TENSIX,
+  ROUTER_ONLY,
+};
+
+ struct CoreCoord_V1 : public tt_xy_pair {
```
pgkeller, line 40: I prefer V1.  I was thinking we could use derived types to construct the base type: TensixCoreCoord() sets core_type to Tensix, then keep the CoreType internal only

device/tt_silicon_driver.cpp
pjanevskiTT: Yes we can discuss about it in more detail later on, but one function can do the job as well. If we translate coordinates to physical, we can just program the TLB and do the write, it will work as well

device/tt_core_coordinates.h
pjanevskiTT: If I understand, `TensixCoreCoord` would me tt-metal abstraction? There would still be just CoreCoord_V1 in umd?

pgkeller: yes.  is an API consideration for metal

device/tt_core_coordinates.h
pgkeller: yes, active is "has links" and inactive is "no links".  active is running the "base firmware" to handle host read/write tunneling, "inactive" does not.  we run workloads on both types, but different types of workloads.

device/coordinate_manager.h
```diff
@@ -45,6 +46,17 @@ class CoordinateManager {
...
 
+    // v1 functions
+    // We need only one function per coordinate system for all core types.
+    virtual CoreCoord_V1 to_physical(const CoreCoord_V1 core_coords);
+
+    // v2 functions
+    // We need as many functions as there are core types for each coordinate system.
```
abhullar-tt, line 54: these could be templated to avoid many functions

device/tt_soc_descriptor.cpp
```diff
@@ -126,7 +126,7 @@ void tt_SocDescriptor::load_core_descriptors_from_device_descriptor(YAML::Node &
     for (const auto &core_string : worker_cores) {
         CoreDescriptor core_descriptor;
         core_descriptor.coord = format_node(core_string);
-        core_descriptor.type = CoreType::WORKER;
+        core_descriptor.type = CoreType::TENSIX;
```
abhullar-tt, line 129: thanks for the rename :)

broskoTT:
@pjanevskiTT feel free to just delete the V2 coords from this PR and continue working on this one, rather than creating a new PR. This would nicely keep all the conversation threads at the same place...

device/tt_soc_descriptor.h
tt-asaigal: I think users should at least know what coordinate system a SOC descriptor uses, and what core type each endpoint corresponds to. Today in TT-Metal, its restricted to Physical, but in future, we may need to introduce translated coordinate based SOC descriptors. Tracking coordinate systems in different layers of the stack may lead to repeated code and confusion.
Having said that, the direct user of UMD APIs in Metal is `tt_cluster`, which knows the SOC descriptor variant that gets loaded. This PR already introduces APIs that allow switching between coordinate systems. I think just having APIs from the coordinate manager should be good enough. We likely won't need to heavily augment the `tt_SocDescriptor` class.

tt-vjovanovic:
In `debuda` (soon to be called `tt-lens`), we are using "exported" soc descriptor back in yaml format. It would be great if I could choose coordinates to be returned with `tt_device::get_soc_descriptor`...

tt-vjovanovic:
In JTAG scenario when there is no working PCI device attached to PC, we would still like to use `CoordinateManager` - just make constructor public...

device/coordinate_manager.h
```diff
@@ -45,6 +46,17 @@ class CoordinateManager {
```
broskoTT, line 46: Just talked offline with @tt-vjovanovic 
Let's consider changing get_coordinate_manager to pass either ARCH + harvesting, or just a yaml soc descriptor. We wouldn't pass worker_grid_size nor worker_grid_size in both options.

pjanevskiTT:
@broskoTT @abhullar-tt @pgkeller @tt-vjovanovic @tt-asaigal The PR is taking final shape in terms of how translation is going to be implemented. One more thing I need to do is confirm all translated coordinates on Blackhole and implement that, but that will just require overriding few functions, in a way that has already been done. Can you all review parts of the PR that you are interested in, I will continue resolving the original comment all of you have left or I will open issues if those comments are not specific for translations, rather for API/other changes

device/tt_soc_descriptor.h
pjanevskiTT: I was thinking about providing the API for getting the cores in any coordinate system from the same soc descriptor. It would be something like

```
tt_SocDescritpor soc_desc = tt_SocDescriptor(...);
std::vector<CoreCoord> cores = soc_desc.get_cores(CoreType::TENSIX, CoordSystem::LOGICAL);
```
But the soc descriptor would still be created from physical (noc coordinates)

device/tt_core_coordinates.h
pjanevskiTT: Follow up issue for getting active ethernet cores https://github.com/tenstorrent/tt-umd/issues/344

device/coordinate_manager.h
```diff
@@ -6,74 +6,161 @@
...
+    std::vector<size_t> logical_y_to_physical_y;
+    std::vector<size_t> logical_x_to_physical_x;
+
+    std::vector<size_t> logical_y_to_virtual_y;
+    std::vector<size_t> logical_x_to_virtual_x;
+
+    std::map<size_t, size_t> virtual_y_to_logical_y;
+    std::map<size_t, size_t> virtual_x_to_logical_x;
+
+    std::map<tt_xy_pair, CoreCoord> tensix_logical_to_translated;
```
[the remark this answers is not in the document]
pjanevskiTT: As we talked offline, I like the proposal for cache. Follow up issue to redesign that part, since I would like to push this PR to unblock clients (metal and lens) for the API work. 

On the second comment, I don't have strong opinions on this. I think yours and @abhullar-tt (and other metal folks) opinion is more important here, since it is more API design question. I can always forward the calls to already existing functions to_logical, to_physical, etc...

device/coordinate_manager.h
```diff
@@ -6,74 +6,161 @@
...
+    void clear_tensix_harvesting_structures();
+    void clear_dram_harvesting_structures();
 
-    CoordinateManager(CoordinateManager& other) = default;
+    virtual void fill_tensix_logical_to_translated();
+    virtual void fill_eth_logical_to_translated();
+    virtual void fill_pcie_logical_to_translated();
 
-    virtual ~CoordinateManager() {}
+    std::map<size_t, size_t> physical_y_to_logical_y;
```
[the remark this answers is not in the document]
pjanevskiTT: No, will delete it

device/tt_soc_descriptor.h
```diff
@@ -167,31 +157,19 @@ class tt_SocDescriptor {
         coordinate_manager.reset(new CoordinateManager(*other.coordinate_manager));
```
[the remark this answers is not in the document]
pjanevskiTT: This is a mess, the problem is that we don't really even want to copy SocDescriptor in my opinion, it is just a case at the moment because of some legacy copying in Cluster. This will be deleted. Until then I can change it to shared pointer

device/tt_soc_descriptor.h
```diff
@@ -167,31 +157,19 @@ class tt_SocDescriptor {
...
-
-    tt_logical_coords to_logical_coords(tt_virtual_coords virtual_coords);
-    tt_physical_coords to_physical_coords(tt_virtual_coords virtual_coords);
-    tt_translated_coords to_translated_coords(tt_virtual_coords virtual_coords);
-
-    tt_logical_coords to_logical_coords(tt_translated_coords translated_coords);
-    tt_physical_coords to_physical_coords(tt_translated_coords translated_coords);
-    tt_virtual_coords to_virtual_coords(tt_translated_coords translated_coords);
+    // CoreCoord conversions.
+    CoreCoord to_physical(const CoreCoord core_coord);
```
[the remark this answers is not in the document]
pjanevskiTT: Similar comment to the one above, look at https://github.com/tenstorrent/tt-umd/pull/266#discussion_r1860384858

device/coordinate_manager.h
```diff
@@ -6,74 +6,161 @@
...
+#include <iostream>
 #include <map>
 #include <set>
 #include <vector>
 
 #include "device/tt_arch_types.h"
-#include "device/tt_xy_pair.h"
+#include "tt_core_coordinates.h"
 
 class CoordinateManager {
```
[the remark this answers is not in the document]
pjanevskiTT: Not silly, coordinates have been a pain point to a lot of people, hopefully this effort will eliminate that. I have pending PR for docs as well #223. You can take a look there as well. I will put the comment for the docs in the code as well

device/coordinate_manager.cpp
```diff
@@ -6,184 +6,450 @@
...
+    }
 }
 
-tt_logical_coords CoordinateManager::to_logical_coords(tt_virtual_coords virtual_coords) {
-    return tt_logical_coords(virtual_x_to_logical_x[virtual_coords.x], virtual_y_to_logical_y[virtual_coords.y]);
-}
+CoreCoord CoordinateManager::to_physical(const CoreCoord core_coord) {
+    switch (core_coord.coord_system) {
+        case CoordSystem::PHYSICAL:
+            return core_coord;
```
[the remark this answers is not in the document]
pjanevskiTT: Not yet, will add it, maybe as a follow up issue

device/coordinate_manager.h
```diff
@@ -6,74 +6,161 @@
...
 
-    virtual tt_logical_coords to_logical_coords(tt_physical_coords physical_coords);
-    virtual tt_virtual_coords to_virtual_coords(tt_physical_coords physical_coords);
-    virtual tt_translated_coords to_translated_coords(tt_physical_coords physical_coords);
+    CoordinateManager(CoordinateManager& other) = default;
 
-    virtual tt_logical_coords to_logical_coords(tt_virtual_coords virtual_coords);
-    virtual tt_physical_coords to_physical_coords(tt_virtual_coords virtual_coords);
-    virtual tt_translated_coords to_translated_coords(tt_virtual_coords virtual_coords);
+    CoreCoord to_physical(const CoreCoord core_coord);
```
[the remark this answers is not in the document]
pjanevskiTT: Don't understand the question really. If you think what if the coordinate specified is not valid for given coordinate system, there will be an exception there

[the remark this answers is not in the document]
tt-vjovanovic: User of tt-lens says, read 128 bytes from noc0(1,1). My question is, do I need to check which core is on noc0(1,1) and provide that information to UMD so I don't get exception or that information won't be checked for coordinates that don't require `core_type`?

device/coordinate_manager.h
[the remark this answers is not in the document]
tt-vjovanovic: From API point of view, scaling with adding additional entry to `enum` is easier than adding more functions... From user point of view, it is going to be explicit either way (either in function name or function parameter).

device/coordinate_manager.h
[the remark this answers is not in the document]
pjanevskiTT: Yes, makes sense, I have code for that which is integrating core coords in Cluster, but it is not ready for PR yet. That is not possible through Coordinate Manager at the moment, meaning you need to specify coordinate type, but nice catch. I t. Again, it won't be in this PR, it will be in one of the following

device/coordinate_manager.h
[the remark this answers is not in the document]
abhullar-tt: I like the proposal to have `CoreCoord to(CoreCoord input, CoordSystem output) const;`

Under the hood it would be good to use some level of caching just to have fast translation.

device/coordinate_manager.cpp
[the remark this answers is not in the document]
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/347

device/coordinate_manager.h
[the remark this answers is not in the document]
pjanevskiTT: #348

device/coordinate_manager.h
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/349

device/blackhole/blackhole_coordinate_manager.cpp
```diff
@@ -1,29 +1,115 @@
...
+                CoreCoord(tensix_core.x, tensix_core.y, CoreType::TENSIX, CoordSystem::VIRTUAL);
+            tensix_virtual_to_logical[tensix_core] = CoreCoord(x, y, CoreType::TENSIX, CoordSystem::LOGICAL);
+        }
+    }
+
+    BlackholeCoordinateManager::fill_tensix_logical_to_translated();
+}
+
+void BlackholeCoordinateManager::fill_tensix_logical_to_translated() {
+    const size_t num_harvested_x = __builtin_popcount(CoordinateManager::tensix_harvesting_mask);
```
broskoTT, line 45: This will make the code not cross-compiler friendly. Please write your own implementation (I don't think there's a solution in stdlib) instead, as a private member of coordinate_manager for example.

device/coordinate_manager.cpp
```diff
@@ -6,184 +6,462 @@
...
         }
-        virtual_x_to_logical_x[*physical_x_it] = x;
-        physical_x_it++;
     }
 }
 
-#include "blackhole/blackhole_coordinate_manager.h"
-#include "grayskull/grayskull_coordinate_manager.h"
-#include "wormhole/wormhole_coordinate_manager.h"
+#include "umd/device/blackhole_coordinate_manager.h"
```
broskoTT, line 350: Move to the top of the file.

device/coordinate_manager.cpp
```diff
@@ -6,184 +6,462 @@
...
+                tt::umd::wormhole::ETH_GRID_SIZE,
+                tt::umd::wormhole::ETH_CORES,
+                tt::umd::wormhole::ARC_GRID_SIZE,
+                tt::umd::wormhole::ARC_CORES,
+                tt::umd::wormhole::PCIE_GRID_SIZE,
+                tt::umd::wormhole::PCIE_CORES);
+        case tt::ARCH::BLACKHOLE:
+            return get_coordinate_manager(
+                arch,
+                tt::umd::blackhole::TENSIX_GRID_SIZE,
```
broskoTT, line 393: You seem to pass something from there which is constant and obviously is just need to be known by specific implementation.

Can you:
- Remove the other (with longer signature) get_coordinate_manager
- Make BlackholeCoordinateManager accept only two arguments, harvesting masks
- Make GS and WH managers accept only one argument, tensix harvesting mask
- Move the "#include "umd/device/blackhole_implementation.h"" to blackhole_coordinate_manager.cpp, and then call the base constructor there with these same constants? Same for wh and gs

device/coordinate_manager.cpp
```diff
@@ -6,184 +6,462 @@
...
+        case CoreType::TENSIX:
+            return tensix_logical_to_virtual;
+        case CoreType::DRAM:
+            return dram_logical_to_virtual;
+        case CoreType::ACTIVE_ETH:
+        case CoreType::IDLE_ETH:
+        case CoreType::ETH:
+            return eth_logical_to_virtual;
+        case CoreType::ARC:
+            return arc_logical_to_translated;
```
broskoTT, line 43: to_virtual?

device/coordinate_manager.cpp
```diff
@@ -6,184 +6,462 @@
...
+        case CoreType::PCIE:
+            return pcie_logical_to_translated;
```
broskoTT, line 45: to_virtual?

device/coordinate_manager.cpp
```diff
@@ -6,184 +6,462 @@
...
+        for (size_t y = 0; y < pcie_grid_size.y; y++) {
+            const tt_xy_pair pcie_core = pcie_cores[x * pcie_grid_size.y + y];
+            pcie_logical_to_virtual[{x, y}] = CoreCoord(pcie_core.x, pcie_core.y, CoreType::PCIE, CoordSystem::VIRTUAL);
+            pcie_virtual_to_logical[pcie_core] = CoreCoord(x, y, CoreType::PCIE, CoordSystem::LOGICAL);
+
+            pcie_logical_to_physical[{x, y}] =
+                CoreCoord(pcie_core.x, pcie_core.y, CoreType::PCIE, CoordSystem::PHYSICAL);
+            pcie_physical_to_logical[pcie_core] = CoreCoord(x, y, CoreType::PCIE, CoordSystem::LOGICAL);
+
+            pcie_logical_to_translated[{x, y}] =
```
broskoTT, line 315: You should be calling fill_pcie_logical_to_translated() here instead, right?

device/coordinate_manager.cpp
```diff
@@ -6,184 +6,462 @@
...
+            dram_logical_to_virtual[{x, y}] = CoreCoord(dram_core.x, dram_core.y, CoreType::DRAM, CoordSystem::VIRTUAL);
+            dram_virtual_to_logical[dram_core] = CoreCoord(x, y, CoreType::DRAM, CoordSystem::LOGICAL);
 
-    std::size_t grid_size_x = worker_grid_size.x;
-    std::size_t grid_size_y = worker_grid_size.y;
+            dram_logical_to_physical[{x, y}] =
+                CoreCoord(dram_core.x, dram_core.y, CoreType::DRAM, CoordSystem::PHYSICAL);
+            dram_physical_to_logical[dram_core] = CoreCoord(x, y, CoreType::DRAM, CoordSystem::LOGICAL);
+        }
+    }
```
broskoTT, line 269: You never fill translated dram coords?

device/blackhole/blackhole_coordinate_manager.cpp
```diff
@@ -1,29 +1,115 @@
...
+        if (!(dram_harvesting_mask & (1 << x))) {
+            for (size_t y = 0; y < dram_grid_size.y; y++) {
+                const tt_xy_pair& dram_core = CoordinateManager::dram_cores[x * dram_grid_size.y + y];
+                CoordinateManager::dram_logical_to_physical[{logical_x, y}] =
+                    CoreCoord(dram_core.x, dram_core.y, CoreType::DRAM, CoordSystem::PHYSICAL);
+                CoordinateManager::dram_physical_to_logical[dram_core] =
+                    CoreCoord(logical_x, y, CoreType::DRAM, CoordSystem::LOGICAL);
+            }
+            logical_x++;
         }
```
broskoTT, line 89: Left in base coord_manager file, but also here so you don't forget this place as well:
You never fill translated dram coords?

tests/api/test_core_coord_translation_wh.cpp
```diff
@@ -0,0 +1,279 @@
...
+    std::shared_ptr<CoordinateManager> coordinate_manager =
+        CoordinateManager::get_coordinate_manager(tt::ARCH::WORMHOLE_B0);
+    const tt_xy_pair eth_grid_size = tt::umd::wormhole::ETH_GRID_SIZE;
+
+    for (size_t x = 0; x < eth_grid_size.x; x++) {
+        for (size_t y = 0; y < eth_grid_size.y; y++) {
+            const CoreCoord eth_logical = CoreCoord(x, y, CoreType::ETH, CoordSystem::LOGICAL);
+            const CoreCoord eth_translated = coordinate_manager->to(eth_logical, CoordSystem::TRANSLATED);
+
+            EXPECT_EQ(eth_translated.x, x + 18);
```
broskoTT, line 231: Can we defined these constants 18 and 16 somewhere in coord_manager/soc descriptor?

tests/api/test_core_coord_translation_bh.cpp
```diff
@@ -0,0 +1,419 @@
...
+    }
+}
+
+// Test ethernet coordinate translation.
+TEST(CoordinateManager, CoordinateManagerBlackholeETHTranslation) {
+    std::shared_ptr<CoordinateManager> coordinate_manager =
+        CoordinateManager::get_coordinate_manager(tt::ARCH::BLACKHOLE);
+    const tt_xy_pair eth_grid_size = tt::umd::blackhole::ETH_GRID_SIZE;
+
+    const size_t eth_translated_coordinate_start_x = 20;
```
broskoTT, line 402: Can we define these constants 20 and 25 somewhere in the coord_manager or soc descriptor?

broskoTT (approved):
Awesome that we will have this!
Sorry for leaving 32 comments, hopefully you'll also like the result in the end...

device/api/umd/device/blackhole_implementation.h
```diff
@@ -58,13 +58,43 @@ enum class arc_message_type {
 };
 
 // DEVICE_DATA
-static constexpr std::array<xy_pair, 24> DRAM_LOCATIONS = {
+const static tt_xy_pair TENSIX_GRID_SIZE = {14, 10};
+const static std::vector<tt_xy_pair> TENSIX_CORES = {{
+    {1, 2},   {2, 2},   {3, 2},   {4, 2},   {5, 2},   {6, 2},   {7, 2},   {10, 2},  {11, 2}, {12, 2}, {13, 2}, {14, 2},
```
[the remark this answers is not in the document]
pjanevskiTT: Yes this is a step towards removing yaml files for standard chips.  We will work towards having all the constants in header files

device/api/umd/device/tt_core_coordinates.h
```diff
@@ -0,0 +1,61 @@
...
+    DRAM,
+    ACTIVE_ETH,
+    IDLE_ETH,
+    PCIE,
+    TENSIX,
+    ROUTER_ONLY,
+    // TODO: this keeps compatibility with existing code in SocDescriptor
+    // but it won't be needed later on
+    HARVESTED,
+    ETH,
```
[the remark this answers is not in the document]
pjanevskiTT: I want to check in this without changes in tt-metal, and then start removing/cleaning up stuff one by one with uplifting to tt-metal and tt-lens.

device/api/umd/device/coordinate_manager.h
```diff
@@ -11,69 +11,143 @@
...
+    virtual void translate_eth_coords();
+    virtual void translate_arc_coords();
+    virtual void translate_pcie_coords();
 
-    static std::unique_ptr<CoordinateManager> get_coordinate_manager(
-        tt::ARCH arch,
-        const tt_xy_pair& worker_grid_size,
-        const std::vector<tt_xy_pair>& workers,
-        std::size_t harvesting_mask);
+    void clear_tensix_harvesting_structures();
```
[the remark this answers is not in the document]
pjanevskiTT: That was when functions were public, will remove it

tests/api/test_core_coord_translation_wh.cpp
pjanevskiTT: Moved it to implementation files

tests/api/test_core_coord_translation_bh.cpp
pjanevskiTT: same as above

device/coordinate_manager.cpp
pjanevskiTT: FIxed

device/blackhole/blackhole_coordinate_manager.cpp
pjanevskiTT: Fixed

device/api/umd/device/coordinate_manager.h
```diff
@@ -11,69 +11,143 @@
...
+    std::map<tt_xy_pair, tt::umd::CoreCoord> pcie_logical_to_virtual;
+    std::map<tt_xy_pair, tt::umd::CoreCoord> pcie_logical_to_physical;
+
+    std::map<tt_xy_pair, tt::umd::CoreCoord> pcie_physical_to_logical;
+    std::map<tt_xy_pair, tt::umd::CoreCoord> pcie_virtual_to_logical;
+    std::map<tt_xy_pair, tt::umd::CoreCoord> pcie_translated_to_logical;
 
-    virtual std::set<std::size_t> get_x_coordinates_to_harvest(std::size_t harvesting_mask);
-    virtual std::set<std::size_t> get_y_coordinates_to_harvest(std::size_t harvesting_mask);
+    std::map<tt_xy_pair, tt::umd::CoreCoord>& get_logical_to_translated(CoreType core_type);
```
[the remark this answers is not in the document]
pjanevskiTT: As per offline discussion, I want to check this in to unblock clients on the API side, I will fix internals of the class as part of the issue https://github.com/tenstorrent/tt-umd/issues/355

device/coordinate_manager.cpp
```diff
@@ -6,184 +6,462 @@
...
-
-tt_physical_coords CoordinateManager::to_physical_coords(tt_logical_coords logical_coords) {
-    return tt_physical_coords(logical_x_to_physical_x[logical_coords.x], logical_y_to_physical_y[logical_coords.y]);
+#include <stdexcept>
+
+using namespace tt::umd;
+
+std::map<tt_xy_pair, CoreCoord>& CoordinateManager::get_logical_to_translated(CoreType core_type) {
+    switch (core_type) {
+        case CoreType::TENSIX:
```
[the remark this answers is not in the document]
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/355

device/coordinate_manager.cpp
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/355

device/coordinate_manager.cpp
```diff
@@ -6,184 +6,462 @@
...
+    return logical_mapping[{core_coord.x, core_coord.y}];
 }
 
-tt_translated_coords CoordinateManager::to_translated_coords(tt_virtual_coords virtual_coords) {
-    return to_translated_coords(to_logical_coords(virtual_coords));
+CoreCoord CoordinateManager::to_logical(const CoreCoord core_coord) {
+    switch (core_coord.coord_system) {
+        case CoordSystem::LOGICAL:
+            return core_coord;
+        case CoordSystem::PHYSICAL: {
```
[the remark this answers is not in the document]
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/355

PR #268, opened 2024-11-05
Pull request template

broskoTT:
### Ticket
No issue for this change

### Description
After discussion with @pjanevskiTT on how to handle API breaking UMD PRs, we agreed to try with this approach. The point is that there should be minimal time between merging UMD PR and client (tt_metal and tt_debuda) PRs. This should then not block other improvements/PRs.

### List of the changes
- Added pull_request_template file.

### API Changes

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,4 +1,4 @@
-### Ticket
+### Issue
 (Link to Github issue(s))
 
 ### Description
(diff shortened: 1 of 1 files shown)


joelsmithTT (approved):
Sounds good to me

abhullar-tt (approved):

.github/pull_request_template.md
```diff
@@ -0,0 +1,16 @@
+### Ticket
```
pjanevskiTT, line 1: Issue instead of ticket maybe. Not a strong opinion, it just may be more compatible with wording we use in general. You can decide

pjanevskiTT (approved):

PR #280, opened 2024-11-07
Fix parsing cluster descriptor

### LINKED ISSUE #241: create-eth-map for Blackhole not working
Introduced by #235, create-eth-map binary is not working fully as expected. Chips section of the cluster descriptor is empty, it doesn't show any chips. The binary still works for Wormhole (Grayskull not tested)

I am not really sure how this is built from source, but I know tt-debuda has a binary in their own repo, which I have tested and it produced valid results. This binary is for Blackhole only, so we couldn't reuse it as a general binary, but maybe we could replicate their way of using these binaries

For reference, I am attaching three files (txt files to be able to upload it here)

1. cluster descriptor that binary from tt-debuda produces - 
[cluster_debuda.txt](https://github.com/user-attachments/files/17598278/cluster_debuda.txt)
2. cluster descriptor that we have used as a hardcoded for blackhole so far - https://github.com/tenstorrent/tt-umd/blob/main/blackhole_1chip_cluster.yaml
3. cluster descriptor that updated binary in tt-umd produces - 
[cluster_umd.txt](https://github.com/user-attachments/files/17598279/cluster_umd.txt)

pjanevskiTT:
### Issue

Fix #241 

### Description

Fix parsing of cluster descriptor yaml when there is no chips section (Blachkole and Grayskull path)

### List of the changes

- Fix parsing of cluster descriptor without chips section
- Add warning for unknown board type
- Remove `create_for_grayskull_cluster`
- Fix all tests to use `create-ethernet-map`
- Remove default fallback cluster descriptor for Blackhole

### Testing

Unit testing

### API Changes

No API changes

tt_debuda PR which removes CEM: https://github.com/tenstorrent/tt-debuda/pull/165


device/hugepage.h
```diff
@@ -27,4 +27,4 @@ std::string find_hugepage_dir(std::size_t pagesize);
 // Today we assume there's only one pipeline running within the system.
 // One hugepage per device such that each device gets unique memory.
 int open_hugepage_file(const std::string &dir, chip_id_t physical_device_id, uint16_t channel);
-}
\ No newline at end of file
+}
```
broskoTT, line 30: I think that one tool tries to enforce no newlines at the end of the file, and another does the opposite.
If you continue seeing this let me know so I can investigate. No action here

device/tt_cluster_descriptor.cpp
```diff
@@ -338,40 +338,6 @@ std::unique_ptr<tt_ClusterDescriptor> tt_ClusterDescriptor::create_from_yaml(con
     return desc;
 }
 
-std::unique_ptr<tt_ClusterDescriptor> tt_ClusterDescriptor::create_for_grayskull_cluster(
```
broskoTT, line 341: Yeah, unfortunately this is an API change, will need an update in tt_metal. Probably in tt_debuda as well. Please update the main PR comment.

broskoTT: You can ignore tt_debuda though, I'll update it (I don't think that they call this function actually). I broke it when I removed CEM, I was mistaken that they don't use it.

device/tt_cluster_descriptor.cpp
```diff
@@ -523,15 +489,20 @@ void tt_ClusterDescriptor::load_ethernet_connections_from_connectivity_descripto
...
+    for (YAML::const_iterator node = yaml["arch"].begin(); node != yaml["arch"].end(); ++node) {
         chip_id_t chip_id = node->first.as<int>();
-        std::vector<int> chip_rack_coords = node->second.as<std::vector<int>>();
-        log_assert(chip_rack_coords.size() == 4, "Galaxy (x, y, rack, shelf) coords must be size 4");
-        eth_coord_t chip_location{
-            chip_rack_coords.at(0), chip_rack_coords.at(1), chip_rack_coords.at(2), chip_rack_coords.at(3)};
-        
-        desc.chip_locations.insert({chip_id, chip_location});
-        desc.coords_to_chip_ids[std::get<2>(chip_location)][std::get<3>(chip_location)][std::get<1>(chip_location)][std::get<0>(chip_location)] = chip_id;
+        // Not all archs (e.g. BH) have chip coordinates
```
broskoTT, line 497: nit: e.g. BH during bringup and GS

device/tt_cluster_descriptor.h
```diff
@@ -36,6 +36,7 @@ class tt_ClusterDescriptor {
 
   protected:
 
+  bool chips_have_coordinates;
```
broskoTT, line 39: Or, we could fill chip_locations if "chips" is defined (which we do right now), but also change the check of chips_have_coordinates to !chip_locations.empty()

broskoTT (approved):
really happy for the changes in tests/api and tt_silicon_driver

device/tt_cluster_descriptor.cpp
abhullar-tt: I can help out with the metal required changes for this

device/tt_cluster_descriptor.cpp
abhullar-tt: Correct me if I am wrong @TTDRosen  but I don't think BH will have chip coordinates even after bring-up

device/tt_cluster_descriptor.cpp
```diff
@@ -569,6 +540,9 @@ void tt_ClusterDescriptor::load_chips_from_connectivity_descriptor(YAML::Node &y
             } else if (chip_board_type.second == "GALAXY") {
                 board_type = BoardType::GALAXY;
             } else {
+                if (chip_board_type.second != "e150" && chip_board_type.second != "p150A") {
```
abhullar-tt, line 543: why can't we add E150 and P150A to BoardType enums?

device/tt_cluster_descriptor.cpp
broskoTT: Ah actually, there is a flow in tt_metal
```
if (this->target_type_ == TargetDevice::Simulator) {
this->cluster_desc_ =
            tt_ClusterDescriptor::create_for_grayskull_cluster(logical_mmio_device_ids, physical_mmio_device_ids);
}
```

Let's add this to our API tests, so we keep track of it.

Not sure what would be the right solution here, maybe convert this function into a ::create_dummy_cluster(set<chip_id_t>), which would create simplest clusterdescriptor with single chip?

@abhullar-tt what's your take on this?

Re: Changes in tt_metal, if you have the time, sure, but we can also do it (you can just help with the review). If you choose to do it, we're trying to have a process around it where tt_metal PR gets approved before UMD one is merged, so that you can quickly merge the tt_metal counterpart after tt_umd one is merged, therefore we don't block each other. So a PR in tt_metal should be created pointing to this branch, and then switched to UMD main before merging.

abhullar-tt: yeah I think it would be to good to have an api `create_cluster_for_simulation`

broskoTT: As discussed offline with @pjanevskiTT , you can leave the create_for_grayskull_cluster for now in the code (this will mean that it will not break build for tt_metal), and in one of the following PRs, we can effectively rename and restructure it to cover only Simulation/Mockup cluster.

pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/286

pjanevskiTT:
Blackhole tests pass as well localy, we don't have it on CI

abhullar-tt (approved):

PR #305, opened 2024-11-15
Build ubench in CI

joelsmithTT:
### Issue
N/A

### Description
tests/microbenchmark wasn't getting built by CI

### List of the changes
modify CI yaml to build it with the unit tests

### Testing
Builds locally, CI test pending

### API Changes


tests/CMakeLists.txt
```diff
@@ -83,4 +81,5 @@ add_custom_target(
         simulation_tests
         test_pcie_device
         api_tests
+        ubench
```
broskoTT, line 84: This does seem like a separate target, not a suite of unit tests that we want to run ocassionaly, but only when we want to run benchmarks, right?

Consider not adding this to umd_tests, but rather adding building ubench target (as part of already existing build command, or as a new command) in .github/workflows/build-tests.yml

joelsmithTT: will do

broskoTT (approved):

PR #306, opened 2024-11-15
Support multiple unconnected clusters

### LINKED ISSUE #226: Cluster doesn't support multiple unconnected clusters
A follow up from #165 
Due to the way get_closest_mmio_capable_chip works, and due to the way create-ethernet-map works, it doesn't work correctly for multiple unconnected clusters.
Copying some details from the chat:

It only uses eth_coord_t data for each chip to determine this. It looks to me like this code assumes that all the chips are connected in some way, and doesn't handle two separate clusters of chips (like three N300 cards that I have). So the code currently returns the same chip for any remote chip (so it will return an unconnected local chip sometimes).

Daniel Rosen: But imo we don’t need to worry too much about these corner cases, I think that tt-fabric is looking to be ready in December-ish (if not sooner) and that’ll completely change the cem and coordinate requirements

broskoTT:
### Issue
Fixes tenstorrent/tt-umd#226 
Unblocks systems with multiple blackhole cards requested by @abhullar-tt 
Makes situation in https://github.com/tenstorrent/tt-metal/issues/15101 better.

### Description
I'm not fully certain if this was the only required change needed to support multiple unconnected clusters. There was a clear issue due to way eth coords were handled, which is fixed. Not sure if there is some other assumption somewhere else in the driver.

### List of the changes
- Changed eth_coord_t to a struct. Made changes throughout the code accordingly. This somewhat deprecates tenstorrent/tt-umd#46 
- Added cluster_id to eth_coord struct, which designates which connected graph component a chip belongs to.
- Filling up cluster_id as part of construction.
- Split filling ethernet coordinates and galaxy shelf related structures.
- If cluster_id doesn't match, the get_eth_distance function will return limits::max().

### Testing
Uncommented previously skipped test, which was failing for one of the example cluster descriptors.

### API Changes

### DIFF
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -46,8 +46,8 @@ function(fetch_dependencies)
     # boost::interprocess
     ############################################################################################################################
     include(${PROJECT_SOURCE_DIR}/cmake/fetch_boost.cmake)
+    fetch_boost_library(container_hash)
     fetch_boost_library(interprocess)
-    fetch_boost_library(functional)
 
     ############################################################################################################################
     # Nanomsg
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -47,6 +47,7 @@ function(fetch_dependencies)
     ############################################################################################################################
     include(${PROJECT_SOURCE_DIR}/cmake/fetch_boost.cmake)
     fetch_boost_library(interprocess)
+    fetch_boost_library(functional)
 
     ############################################################################################################################
     # Nanomsg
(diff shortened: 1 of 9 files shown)


joelsmithTT (approved):
If I have two N300 chips in my system that are not connected to each other, do I need to use `tt-topology` to give them different coordinates before I can use them both in a single `tt::umd::Cluster` or is this change sufficient?

Thank you for fixing `eth_coord_t`, this is a huge improvement in readability.

broskoTT:
> If I have two N300 chips in my system that are not connected to each other, do I need to use `tt-topology` to give them different coordinates before I can use them both in a single `tt::umd::Cluster` or is this change sufficient?
> 
> Thank you for fixing `eth_coord_t`, this is a huge improvement in readability.

I have no idea what tt-topology is, so I guess the answer is that this change is sufficient. See tests/api/cluster_descriptor_examples/wormhole_2xN300_unconnected.yaml for reference

PR #310, opened 2024-11-18
create_for_grayskull_cluster -> create_mock_cluster

broskoTT:
### Issue
Contributes to #99 , and a continuation of #280 

### Description
This function is now only used for creating a cluster descriptor for simulation or mockup environment. Changing accordingly.

### List of the changes
- Renamed create_for_grayskull_cluster to create_mock_cluster
- Changed arguments and functionality so that this function does not accept physical device ids at all.
- Accepting std::vector instead of std::set for logical device ids

### Testing
Code builds.

### API Changes
This PR has API changes:
- [x] tt_metal approved PR pointing to this branch: https://github.com/tenstorrent/tt-metal/pull/15301
- tt_debuda doesn't use this


vtangTT:
Just FYI that I just uncovered last night that simulator was broken on the metal level due to changes made in this https://github.com/tenstorrent/tt-metal/pull/14945.

Is there a branch with related metal changes? I want to see if this fixes the simulator problem.

pjanevskiTT:
Could you try and bisect the commits to see which one breaks the simulator? Not really sure there is a branch with related metal changes. So far uplift strategy was to uplift everything that changes the API plus Almeet uplifts some things that are useful for metal when there is some change inside UMD. If you can pinpoint the commit maybe we can debug it together

vtangTT:
I have a small change to fix for simulator. Is it ok if I push to your branch @broskoTT

device/tt_cluster_descriptor.cpp
```diff
@@ -437,44 +437,31 @@ std::unique_ptr<tt_ClusterDescriptor> tt_ClusterDescriptor::create_from_yaml(
...
+        eth_coord_t chip_location{0, logical_id, 0, 0, 0};
         desc->chip_locations.insert({logical_id, chip_location});
-        desc->coords_to_chip_ids[chip_location.rack][chip_location.shelf][chip_location.y][chip_location.x] =
-            logical_id;
-        log_debug(
-            tt::LogSiliconDriver, "{} - adding logical: {} => physical: {}", __FUNCTION__, logical_id, physical_id);
+        desc->coords_to_chip_ids[chip_location.rack][chip_location.shelf][chip_location.y][chip_location.x] = logical_id;
+        log_debug(tt::LogSiliconDriver, "{} - adding logical: {}", __FUNCTION__, logical_id);
+        desc->chip_board_type.insert({logical_id, board_type});
+        desc->chips_with_mmio.insert({logical_id, logical_id});
```
broskoTT, line 464: @vtangTT is mmio required by simulator?
Would we ever want not to have mmio chips in this mock cluster descriptor, or we're always fine with having mmio?
@mbezuljTT Is this whole create_mock_cluster in line with your expectations for mock class?

vtangTT: mmio is not required by the simulator however, the `chip_with_mmio` map has to be populated or else the metal cluster will segfault

Issue #270, opened 2024-11-06
Remove conditional compilation from tests


broskoTT:
We've addressed conditional compilation as part of #51 
This is now a follow up with lower priority.

Currently (as can be seen from readme and CI) our tests are built using ARCH_NAME. This should be removed to further improve conditional compilation situation with tt_metal and tt_umd

PR #322, opened 2024-11-21
Place the public API in a special place with a canonical way of accessing them

afuller-TT:
### Issue
Solving part of #324

### Description
It's currently unclear which files consist of UMD's public API and which are private headers.
Additionally, consumers like tt-metal are reaching around with arbitrary paths to fetch headers.  They shouldn't.
There should be one place for the public API and one way for consumers to access it, and private stuff should remain private.

### List of the changes
* Place all UMD Device's public API under `api/umd/device/`
* Expose `api/` as the include dir, so all consumers must use `umd/device/foo.h` to include API headers
    * Even internal to UMD::Device, though this is subjective.
* Do not expose any other directory publicly
    * Inside Device we may include other headers, but external consumers cannot.

### Testing
Corresponding PR in tt-metal to consume with the new paths: 

### API Changes

- [x] (If breaking change) tt_metal approved PR pointing to this branch: https://github.com/tenstorrent/tt-metal/pull/15322
- [x] (If breaking change) tt_debuda approved PR pointing to this branch: https://github.com/tenstorrent/tt-lens/pull/193


common/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+add_library(umd_common INTERFACE)
```
broskoTT, line 1: @joelsmithTT FYI, related to #290 , you can rebase after this change, and it will be easier for the new logging library

common/utils.hpp
```diff
@@ -18,10 +18,10 @@ std::string get_abs_path(std::string path) {
     std::filesystem::path current_file_path = std::filesystem::path(__FILE__);
     std::filesystem::path umd_root;
     if (current_file_path.is_absolute()) {
-        umd_root = current_file_path.parent_path().parent_path();
+        umd_root = current_file_path.parent_path().parent_path().parent_path();
```
broskoTT, line 21: What would be your suggestion on how to resolve this? I don't like having "get_abs_path".

I would imagine we can "install" .yaml files , by adding them explicitly somewhere in CMake config? Then these files would be consumed relatively (to the lib path, if this is even possible? ) just by referencing the name of the file without any dirs, if they end up sitting next to the lib somewhere in /build folder?

I'd prefer if we didn't solve this in this PR, lets leave it for another one. We can put this info in an issue if there's relevant info to be put.

device/CMakeLists.txt
```diff
@@ -2,41 +2,41 @@ set(POSITION_INDEPENDENT_CODE ON)
...
+        wormhole/wormhole_coordinate_manager.cpp
+        wormhole/wormhole_implementation.cpp
+        xy_pair.cpp
+        ${FBS_GENERATED_HEADER}
+)
 
 target_include_directories(
     device
     PUBLIC
-        $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
```
broskoTT, line 35: Does this mean that the root of the repo will not be in the include dir anymore? nice

src/CMakeLists.txt
```diff
@@ -7,4 +7,5 @@ target_include_directories(
         $<$<STREQUAL:$ENV{ARCH_NAME},wormhole_b0>:${CMAKE_CURRENT_SOURCE_DIR}/firmware/riscv/wormhole>
         $<$<STREQUAL:$ENV{ARCH_NAME},grayskull>:${CMAKE_CURRENT_SOURCE_DIR}/firmware/riscv/grayskull>
         $<$<STREQUAL:$ENV{ARCH_NAME},blackhole>:${CMAKE_CURRENT_SOURCE_DIR}/firmware/riscv/blackhole>
+        firmware/riscv
```
broskoTT, line 10: shouldn't this be ${CMAKE_CURRENT_SOURCE_DIR}/firmware/riscv ?

And second question, why is this needed, aren't these headers already included with previous 3 lines? Or are previous 3 lines not needed now?
Do you maybe have a problem that in umd::device which is not conditionally compiled based on ARCH_NAME, you need to "#include blackhole/something.h", but umd::tests which is conditionally compiled based on ARCH_NAME will have "#include something.h" ?
It is probably not that simple to just change all "#include something.h" to ""#include arch/something.h"? Probably if there is some common file in all tests which should include based on arch...

Again, please just trying to clarify, no action for this PR, let's just leave info in the issue if there is anything additional to add https://github.com/tenstorrent/tt-umd/issues/270

device/blackhole/blackhole_implementation.cpp
```diff
@@ -2,11 +2,11 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-#include "blackhole_implementation.h"
+#include "umd/device/blackhole_implementation.h"
 
-#include "device/cluster.h"
-#include "src/firmware/riscv/blackhole/eth_interface.h"
-#include "src/firmware/riscv/blackhole/host_mem_address_map.h"
+#include "blackhole/eth_interface.h"
```
broskoTT, line 7: This is src/firmware/risv/blackhole/eth_interface.h ?

But the src/firmware/riscv is not exposed publicly, right?

tests/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ target_include_directories(
     INTERFACE
         ${PROJECT_SOURCE_DIR}
         ${CMAKE_CURRENT_SOURCE_DIR}
+        "$<TARGET_PROPERTY:umd::device,INCLUDE_DIRECTORIES>"
```
broskoTT, line 22: So the line above in target_link_libraries which references umd::device, won't include umd::device's include_directories automatically?

broskoTT (approved):
Awesome effort, thanks!
I've left some comments, but they're for my clarification, none of them call to changes in this PR.

common/utils.hpp
afuller-TT: A few ideas, depending on how/when/where the tests are invoked.

Typically a unit test is invoked from the build tree without any install step.  Unit tests typically aren't "installed" and shouldn't require that extra step because they you can't just run it form the IDE while developing.

For this type, and if the data is in a single place, then simply setting [WORKING_DIRECTORY](https://cmake.org/cmake/help/latest/command/add_test.html) in CTest would mean running it via CTest or the IDE would have a known `pwd` regardless of where the binary exists.

Another option is to pass the data dir as a #define in the build.  Then the binary can be run even from the CLI and know the path to reach to.

Neither of those solutions allows moving the data to a new root (eg: on a different machine).  If doing that, then you'll want to go through the install step and rearrange files, at that point choose a convention like all tests are together in one dir, and input data is in a subdir named `data`.  Then the code loads from "where am I right now? + data/"

To support running the test locally for development pre-install and also on other machines post-install, then the data lookup might need to try both locations since they differ.

device/CMakeLists.txt
afuller-TT: It means exactly that!

src/CMakeLists.txt
afuller-TT: The implementations are always compiled:
https://github.com/tenstorrent/tt-umd/blob/main/device/CMakeLists.txt#L16-L18

And each one needs to know its piece:
https://github.com/tenstorrent/tt-umd/blob/main/device/grayskull/grayskull_implementation.cpp#L8-L9

The lines above seem to be abstracting away _which_ arch it is.  The effort to remove ARCH_NAME envvar would need to change such instances.

device/blackhole/blackhole_implementation.cpp
afuller-TT: Define 'public' :)

`src/firmware/riscv` is public from [firmware](https://github.com/tenstorrent/tt-umd/blob/afuller/clarify-api/src/CMakeLists.txt#L6)
Firmware is private to [device](https://github.com/tenstorrent/tt-umd/blob/afuller/clarify-api/device/CMakeLists.txt#L53).

So like a Foo that has a private member variable Bar, Bar's public interface is not exposed through Foo.  But anyone who uses Bar directly can see it.
umd::device gets to see `src/firmware/riscv`, but anyone who uses umd::device will NOT see `src/firmware/riscv`.

tests/CMakeLists.txt
afuller-TT: When `target_link_libraries` connects up A and B, the `PUBLIC` stuff is brought across.  So all of umd::device's public include directories (and any public compile definitions, dependencies, etc).  But none of its private stuff.

HOWEVER, for tests we have legitimate claim to snoop on the UUT's private bits such as header files.  So here we're sniffing out its `INCLUDE_DIRECTORIES` which are all dirs listed as either `PUBLIC` or `PRIVATE` (only `INTERFACE` ones won't show up here, as those are exported but not used internally).

The mapping is:
[interface, public] -> INTERFACE_INCLUDE_DIRECTORIES
[public, private] -> INCLUDE_DIRECTORIES

common/CMakeLists.txt
joelsmithTT: ok, I will wait for this to go in, then rebase & merge #290.

device/api/umd/device/cluster.h
```diff
@@ -13,14 +13,14 @@
 #include <unordered_set>
```
joelsmithTT, line 13: You don't need to change anything and this remark isn't specific to this file...

device/cluster.h -> device/api/umd/device/cluster.h

Can we declutter this?  Get rid of "device"?  Now it's in there twice

broskoTT: Yes, that will happen. Also it might make sense to remove "device" altogether, since "umd" itself is "device"
https://github.com/tenstorrent/tt-umd/issues/324

blozano-tt: related: I've never liked the library name "libdevice.so" ... could there be a more generic name? However, it may be too late to fix.

vote: `libtt_umd.so`

device/cluster.cpp
```diff
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: (c) 2023 Tenstorrent Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
-#include "cluster.h"
+#include "umd/device/cluster.h"
```
joelsmithTT, line 4: Why is `umd/device` not part of the build include path?

afuller-TT: Because we haven't discussed what internal access should look like.  This PR was demonstrating external access and putting all the public headers in a segregated place to distinguish external vs internal.  That aspect is more straightforward / objective.

Internal access is more subjective.

On the one hand, accessing the same way as an external consumer makes it obvious when we're using a public API -- does that matter?  For some projects maybe that's useful information and for others it's inconsequential.
It also means that moving a file from internal to external doesn't require changing the includes.  In a project highly in flux, perhaps that's a useful convenience.  But if structing include ordering to indicate proximity, then it'll need to be updated anyway.  And in a more well established project hoisting a module externally is very very rare in the grand scheme of things.

On the other hand, accessing like any other internal header separates on a different concern: I'm conveying that this is a header within this project, but I'm not affected nor do I care whether said header is ALSO available externally or not.  Not my concern.

I believe @blozano-tt prefers the former.  I prefer the latter.  But ultimately neither of us own UMD, so it's you and @broskoTT who should make a call re: what you prefer internal for your project.  I can help guide on the CMake change for whichever you prefer.

PR #323, opened 2024-11-22
Make nng private and static

blozano-tt:
### Issue
NA

### Description
libnng.so is no longer a public dependency as of this PR: https://github.com/tenstorrent/tt-umd/pull/224
Additionally, given that it is only consumed here it would be advantageous to statically link it.
Why?
So we don't have to:
- Install it
- Package it in a debian
- Bundle it in a ttnn python wheel ... etc ... 

### List of the changes
Build nng as static
Make it private linkage

### Testing
Hoping CI is good enough

### API Changes

### DIFF
--- a/device/CMakeLists.txt
+++ b/device/CMakeLists.txt
@@ -41,15 +41,14 @@ target_include_directories(
 )
 
 # flatbuffers is public - exposed to tt_metal by tt_simulation_device_generated.h
-# nng is public - exposed to tt_metal by tt_simulation_host.hpp
 target_link_libraries(
     device
     PUBLIC
-        nng
         flatbuffers
         uv
     PRIVATE
         hwloc
+        nng
         rt
         Boost::interprocess
         fmt::fmt-header-only
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -56,7 +56,6 @@ function(fetch_dependencies)
         GITHUB_REPOSITORY nanomsg/nng
         GIT_TAG v1.8.0
         OPTIONS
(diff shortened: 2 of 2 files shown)


device/CMakeLists.txt
```diff
@@ -41,16 +41,15 @@ target_include_directories(
 )
 
 # flatbuffers is public - exposed to tt_metal by tt_simulation_device_generated.h
-# nng is public - exposed to tt_metal by tt_simulation_host.hpp
```
broskoTT, line 44: @vtangTT  Does this have some implication for simulator usage in tt_metal?

joelsmithTT (approved):

device/CMakeLists.txt
vtangTT: yep should be fixed I forgot to add comment

device/CMakeLists.txt
```diff
@@ -41,16 +41,15 @@ target_include_directories(
...
 target_link_libraries(
     device
     PUBLIC
-        nng
         flatbuffers
-        uv
     PRIVATE
```
afuller-TT, line 48: ❤️

afuller-TT (approved):

blozano-tt:
> LGTM, especially since all the checks are green. Not sure if I'm missing some impact of this change

I am doing a quick test integration on metal side

blozano-tt:
> > LGTM, especially since all the checks are green. Not sure if I'm missing some impact of this change
> 
> I am doing a quick test integration on metal side

https://github.com/tenstorrent/tt-metal/actions/runs/11978804872

Things are looking good ... just one flaky test failure...

Issue #326, opened 2024-11-24
Don't assume numa_node is available in sysfs


joelsmithTT:
Just ran into this on a dev system:
```
2024-11-24 21:34:22.810 | INFO     | SiliconDriver   - Detected 1 PCI device : [0]
unknown file: Failure
C++ exception with description "TT_THROW @ /home/ubuntu/git/tt-umd/device/pcie/pci_device.cpp:59: tt::exception
info:
Failed reading sysfs attribute: {}
/sys/bus/pci/devices/0000:01:00.0/numa_node
backtrace:
```

PR #327, opened 2024-11-24
Add default value parameter to read_sysfs

joelsmithTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/326

### Description
For some reason, a kernel I booted on a RISC-V dev board didn't have the `numa_node` entry in sysfs -- this commit fixes the assumption that this file will always exist.

### List of the changes
Adds function for reading from sysfs with a default return value if the read fails.

### Testing
Manual

### API Changes

### DIFF
--- a/device/pcie/pci_device.cpp
+++ b/device/pcie/pci_device.cpp
@@ -79,6 +79,15 @@ static T read_sysfs(const PciDeviceInfo &device_info, const std::string &attribu
     return value;
 }
 
+template <typename T>
+T read_sysfs(const PciDeviceInfo &device_info, const std::string &attribute_name, const T &default_value) {
+    try {
+        return read_sysfs<T>(device_info, attribute_name);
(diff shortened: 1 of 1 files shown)


device/pcie/pci_device.cpp
```diff
@@ -255,7 +264,7 @@ PCIDevice::PCIDevice(int pci_device_number, int logical_device_id) :
     logical_id(logical_device_id),
     pci_device_file_desc(open(device_path.c_str(), O_RDWR | O_CLOEXEC)),
     info(read_device_info(pci_device_file_desc)),
-    numa_node(read_sysfs<int>(info, "numa_node")),
+    numa_node(read_sysfs<int>(info, "numa_node", -1)),
```
broskoTT, line 267: Seeing this line different than others can be confusing. Please leave a comment, you can just copy paste from description.

Issue #281, opened 2024-11-07
Non REG_TLB address concatenated to 32bit


broskoTT:
As reported by Luka and @tt-vjovanovic from tt-debuda team, reads/writes to regions of virtual addresses which sit above 32bits won't work. Specifically, they used this path for ARC.

Original message:
0x880030060
0x880030064
0x880030068
....
0x810000000
0x880000000
First three are scratch registers.
Last two are CSM memory and ROM
size = 1
core = 0,10
fallback_tlb = "LARGE_WRITE_TLB"
![Image](https://github.com/user-attachments/assets/dfe27943-9ada-4c9d-bcbe-4d6b5cf6fa13)

joelsmithTT:
Should be fixed in https://github.com/tenstorrent/tt-umd/pull/328 FYI @tt-vjovanovic

Issue #303, opened 2024-11-15
Cluster::configure_tlb is wrong


joelsmithTT:
`void Cluster::configure_tlb(chip_id_t logical_device_id, tt_xy_pair core, std::int32_t tlb_index, std::int32_t address, uint64_t ordering)` 
 The `int32_t address` is incorrect.  It should be uint64_t.

joelsmithTT:
Fixed in https://github.com/tenstorrent/tt-umd/pull/328

PR #328, opened 2024-11-24
Fix address types

joelsmithTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/303

### Description
Using signed integer for address is wrong and can cause bugs due to sign extension since UMD represents addresses as uint64_t internally.  32 bits is unnecessarily restrictive (there are address spaces in the chips that reach beyond 0x7fffffff that I might want to configure a static window for).

### List of the changes
* Use uint64_t instead of int32_t in `configure_tlb` method
* Convert sizes/constants in some header files from int32_t to uint32_t
* Remove unnecessary `std::` prefixing

### Testing
CI

### API Changes
This PR has API changes, but it shouldn't break anything.

### DIFF
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -171,11 +171,14 @@ struct tt_4_byte_aligned_buffer {
 namespace tt::umd {
 
 bool Cluster::address_in_tlb_space(
-    uint32_t address, uint32_t size_in_bytes, int32_t tlb_index, uint64_t tlb_size, std::uint32_t chip) {
-    return (
-        (tlb_config_map.at(chip).find(tlb_index) != tlb_config_map.at(chip).end()) &&
-        address >= tlb_config_map.at(chip).at(tlb_index) &&
-        (address + size_in_bytes <= tlb_config_map.at(chip).at(tlb_index) + tlb_size));
+    uint64_t address, uint32_t size_in_bytes, int32_t tlb_index, uint64_t tlb_size, std::uint32_t chip) {
+    const auto& tlb_map = tlb_config_map.at(chip);
+    const auto it = tlb_map.find(tlb_index);
+    if (it != tlb_map.end()) {
+        auto mapped_address = it->second;
+        return address >= mapped_address && (address + size_in_bytes <= mapped_address + tlb_size);
+    }
+    return false;
 }
 
 std::unordered_map<chip_id_t, tt_SocDescriptor>& Cluster::get_virtual_soc_descriptors() {
@@ -1113,7 +1116,7 @@ void Cluster::write_device_memory(
     const void* mem_ptr,
     uint32_t size_in_bytes,
     tt_cxy_pair target,
-    std::uint32_t address,
+    uint64_t address,
     const std::string& fallback_tlb) {
     PCIDevice* dev = get_pci_device(target.chip);
     const uint8_t* buffer_addr = static_cast<const uint8_t*>(mem_ptr);
@@ -1164,13 +1167,7 @@ void Cluster::write_device_memory(
 }
 
 void Cluster::read_device_memory(
-    void* mem_ptr,
-    tt_cxy_pair target,
-    std::uint32_t address,
-    std::uint32_t size_in_bytes,
-    const std::string& fallback_tlb) {
-    // Assume that mem_ptr has been allocated adequate memory on host when this function is called. Otherwise, this
-    // function will cause a segfault.
+    void* mem_ptr, tt_cxy_pair target, uint64_t address, uint32_t size_in_bytes, const std::string& fallback_tlb) {
     log_debug(
         LogSiliconDriver,
         "Cluster::read_device_memory to chip:{} {}-{} at 0x{:x} size_in_bytes: {}",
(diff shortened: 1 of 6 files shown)


src/firmware/riscv/wormhole/l1_address_map.h
```diff
@@ -18,120 +18,120 @@ struct mailbox_type {
...
+  static constexpr uint32_t BRISC_PERF_BUF_BASE_ADDR = PERF_RISC_MAILBOX_SIZE + PERF_UNUSED_SIZE + PERF_RISC_MAILBOX_ADDR;
+  static constexpr uint32_t MATH_PERF_BUF_BASE_ADDR = BRISC_PERF_BUF_BASE_ADDR + BRISC_PERF_BUF_SIZE;
+  static constexpr uint32_t UNPACK_PACK_PERF_BUF_BASE_ADDR = MATH_PERF_BUF_BASE_ADDR + MATH_PERF_BUF_SIZE;
+  static constexpr uint32_t PERF_NUM_THREADS = 5;
+
+  static constexpr uint32_t PERF_QUEUE_PTRS = PERF_QUEUE_HEADER_ADDR;
+  static constexpr uint32_t PERF_THREAD_HEADER = PERF_QUEUE_PTRS + 8;
+  static constexpr uint32_t PERF_WR_PTR_COPY = PERF_QUEUE_PTRS + 12;
+
+  static constexpr uint32_t WALL_CLOCK_L = 0xFFB121F0;
```
broskoTT, line 133: How did dhis work previously? This would result in a negative value, right?

broskoTT:
Slightly related to #281

broskoTT (approved):

src/firmware/riscv/wormhole/l1_address_map.h
joelsmithTT: UMD used to get away with this brokenness because it would do something like,

```
void b(uint32_t i) { /* fixed: truncated back to 32 bits */ }
void a(uint64_t i) { /* oops: now it's sign extended */ b(i); }

int32_t WALL_CLOCK_L = 0xFFB121F0; /* yes, this is negative */
a(WALL_CLOCK_L);
```
We had some bugs when the equivalent of `b(uint32_t)` was changed to uint64.

joelsmithTT: I think this was fixed by changing Metal's version of the address maps to be unsigned.

joelsmithTT:
> Slightly related to #281

I might as well fix this too.

device/cluster.cpp
```diff
@@ -1164,13 +1167,7 @@ void Cluster::write_device_memory(
 }
 
 void Cluster::read_device_memory(
-    void* mem_ptr,
-    tt_cxy_pair target,
-    std::uint32_t address,
-    std::uint32_t size_in_bytes,
-    const std::string& fallback_tlb) {
-    // Assume that mem_ptr has been allocated adequate memory on host when this function is called. Otherwise, this
-    // function will cause a segfault.
```
joelsmithTT, line 1173: I removed this because it's not true.  If there is inadequate memory where that points, you are lucky if your program segfaults.

device/cluster.cpp
```diff
@@ -171,11 +171,14 @@ struct tt_4_byte_aligned_buffer {
...
 
 bool Cluster::address_in_tlb_space(
-    uint32_t address, uint32_t size_in_bytes, int32_t tlb_index, uint64_t tlb_size, std::uint32_t chip) {
-    return (
-        (tlb_config_map.at(chip).find(tlb_index) != tlb_config_map.at(chip).end()) &&
-        address >= tlb_config_map.at(chip).at(tlb_index) &&
-        (address + size_in_bytes <= tlb_config_map.at(chip).at(tlb_index) + tlb_size));
+    uint64_t address, uint32_t size_in_bytes, int32_t tlb_index, uint64_t tlb_size, std::uint32_t chip) {
+    const auto& tlb_map = tlb_config_map.at(chip);
+    const auto it = tlb_map.find(tlb_index);
```
joelsmithTT, line 176: I rewrote this so that it is more explicit about the number of lookups that need to be performed: 2.

joelsmithTT:
This should now fix #281 - @broskoTT please take another look at your convenience.

device/cluster.h
```diff
@@ -278,8 +278,8 @@ class tt_device {
     virtual void configure_tlb(
         chip_id_t logical_device_id,
         tt_xy_pair core,
-        std::int32_t tlb_index,
```
pjanevskiTT, line 281: Do we generally want to work towards removing std prefix everywhere? Asking in general, if that is what we want to follow I am ok with that. I see a lot of inconsistent usage of this

pjanevskiTT (approved):
Nice, left just one comment for general discussion

device/cluster.h
joelsmithTT: Yes, in the spirit of aligning with tt-metal's style let us favor integer types without the prefix.

pjanevskiTT: Makes sense

device/cluster.cpp
```diff
@@ -171,11 +171,14 @@ struct tt_4_byte_aligned_buffer {
 namespace tt::umd {
```
broskoTT, line 171: Please wait until UMD main is healthy before completing this PR...

Issue #137, opened 2024-10-09
Remove notion of logical id from PCIDevice class


joelsmithTT:
It doesn't belong there.

broskoTT:
I've left a comment in the PR as well, but want to make this clear here as well:
Is device ID guaranteed to be the same as logical ID ?

Having a well thought and consistent naming should be enough.

soc_id -> since we detect soc_descriptor based on it anyway
dev_id -> it is in /dev folder
chip_id -> it corresponds to one chip

I don't even think that these are the best names, just suggested some naming scheme.

joelsmithTT:
> Is device ID guaranteed to be the same as logical ID ?

Not guaranteed - I can get my system into a state where they are not.

Most of the time it is probably a reasonable assumption that with N PCIe chips and M total chips, 0 <= dev_id < N and 0 <= chip_id < M, with the bottom N elements of both equal and referring to the same chip.

But this behavior is not guaranteed by KMD and `create-ethernet-map`.  There is a more systemic issue here in that UMD relies on an undocumented tool to produce undocumented output which UMD parses to know what chip is where.

pjanevskiTT:
@broskoTT @joelsmithTT what is the state of this?

joelsmithTT:
Still needs work.  There are APIs where it isn't clear what the right type is.

broskoTT:
This was somewhat fixed in #149 . Looks like it might still need work.

broskoTT:
This has been somewhat fine for some time, maybe another pass through code to make sure that all ids have any kind of prefix, which points to the type of id.

pjanevskiTT:
I think this is going to become interesting if we need to expose unique chip IDs to, for example, tt-distributed. Not sure if they would like to combine two Clusters (probably yes), so for internal usage we need to find some kind of unique chip ID. I think for Blackhole this can be sorted as board_id + asic location, I know for Wormhole this is a bit messy at the moment. We can discuss this in a separate issue maybe as well.

broskoTT:
I don't think there is anything left to do on this particular issue.

PR #330, opened 2024-11-25
Remove logical_id from pci_device

broskoTT:
### Issue
Related to #137

### Description
logical_id should not exist in PCIDevice. Logical id is the notion of the current runtime. PCIDevice only cares about physical card connected to the

### List of the changes
- Remove logical_id from PCIDevice constructor
- Pass harvested info only for this device, not whole map.

### Testing
Code builds

### API Changes


joelsmithTT (approved):
Glad to see this going away.

Issue #254, opened 2024-11-04
Simplify existing hugepage implementation


joelsmithTT:
* Remove code complexity determining the right number of host channels.  Just use what is passed in at construction, on a per-device basis.  (User passes in 2: use 2 1G page _per device_).  If this fails due to inadequate hugepage count, terminate with an actionable error message.
* Remove cpuset_lib code (currently used for channel quantity determination and NUMA migration)
* Hop to desired NUMA node _prior_ to hugepage allocation
* Remove hugetlbfs; access hugepages via `MAP_HUGETLB`, `MAP_HUGE_1GB`.  This means users will lose the ability to share a hugepage across multiple processes or inspect the files in /dev/hugepage-1G.  Although this represents the loss of a debugging feature, it will be lost when we move off of hugepages entirely.

broskoTT:
Closed since parent is closed.

joelsmithTT:
This never made it in, these efforts remain blocked by Metal CI.

Issue #146, opened 2024-10-10
Prefix all env vars with TT_UMD


pjanevskiTT:
Prefix all env vars that can be used in the cde with `TT_UMD`, similar to how tt-metal env vars are prefixed with `TT_METAL`. This issue requires investigating which clients use these env vars in order to not break anything. For example, look at the comment https://github.com/tenstorrent/tt-umd/pull/130#discussion_r1793119760

joelsmithTT:
I would suggest also to document them all in a single place, so that it is easy to understand how to change the runtime behavior of a UMD-based application without grepping the code for getenv.

joelsmithTT:
TT_BACKEND_HUGEPAGE_DIR is deleted

PR #332, opened 2024-11-26
Simplify hugepage code, Chapter I

joelsmithTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/254

### Description
Part 1 of an effort to simplify the hugepage scheme for shared memory between device and host.  Initial work on this also removed hugetlbfs aspects of the setup, but I will save that for part 2.

### List of the changes
* Remove cpuset_lib code; stop doing fancy things to determine host memory channel quantity - just use what the application passes in.  If there are not enough hugepages available, UMD will fail.
* Use libnuma to prefer hugepage allocations on the NUMA node associated with the accelerator device.
* Stop passing desired number of huge pages to `Cluster` as `const uint32 &`.
* Remove env var controlling what directory the hugetlbfs is mounted in, as a step towards not requiring this at all.

### Testing
CI

### API Changes
This PR has API changes, but they should not break anything.  Metal already requires libnuma.
If moving to libnuma from hwloc is a problem, there's a [commit](https://github.com/tenstorrent/tt-umd/commit/a93ff55db18ce792aca977756a6b51fc8d6362b3) that uses hwloc instead that we can use.


device/pcie/pci_device.cpp
```diff
@@ -687,6 +687,14 @@ tt::umd::architecture_implementation *PCIDevice::get_architecture_implementation
 bool PCIDevice::init_hugepage(uint32_t num_host_mem_channels) {
     const size_t hugepage_size = HUGEPAGE_REGION_SIZE;
 
+    if (numa_node > numa_max_node()) {
+        log_warning(LogSiliconDriver, "numa_node: {} is greater than numa_max_node: {}.", numa_node, numa_max_node());
+    }
+
+    // Prefer allocations on the NUMA node associated with the device.
+    numa_set_preferred(numa_node);
```
broskoTT, line 695: Can you please create a tt_metal branch which consumes this UMD commit, and then run tt_metal CI? We don't have machines on our CI that are affected by the change in this PR.

You could run these:
    "all-post-commit-workflows.yaml"
    "t3000-unit-tests.yaml"
    "tg-unit-tests.yaml"
    "tgg-unit-tests.yaml"

device/pcie/pci_device.cpp
```diff
@@ -748,19 +756,6 @@ bool PCIDevice::init_hugepage(uint32_t num_host_mem_channels) {
             continue;
         }
 
-        // Beter performance if hugepage just allocated (populate flag to prevent lazy alloc) is migrated to same
-        // numanode as TT device.
-        if (!tt::cpuset::tt_cpuset_allocator::bind_area_to_memory_nodeset(physical_device_id, mapping, hugepage_size)) {
```
broskoTT, line 753: So the above numa_set_preferred should offer similar/same functionality as this old code?

broskoTT (approved):

device/pcie/pci_device.cpp
joelsmithTT: Yes, the main behavioral difference is the code will now attempt to allocate the page on the desired node (rather than allocate it and then attempt to migrate it).

device/pcie/pci_device.cpp
joelsmithTT: This exposed deficiencies in Metal CI, which is running tests that request more hugepages than are actually provisioned.

device/hugepage.cpp
```diff
@@ -8,15 +8,13 @@
...
 
-#include "cpuset_lib.hpp"
-#include "logger.hpp"
+#include <regex>
 
-const uint32_t g_MAX_HOST_MEM_CHANNELS = 4;
+#include "logger.hpp"
 
-// Hardcode (but allow override) of path now, to support environments with other 1GB hugepage mounts not for runtime.
-const char* hugepage_dir_env = std::getenv("TT_BACKEND_HUGEPAGE_DIR");
```
pjanevskiTT, line 18: I tried deleting this at some point as well and @broskoTT had useful comment that maybe some UMD client relies on this env var. From TT_BACKEND prefix I assume that it might be budabackend so it should not be the problem. I created an issue to investigate usage of env vars #146, so if you end up deleting this just write down there that this one is deleted

device/api/umd/device/cluster.h
```diff
@@ -669,7 +669,7 @@ class Cluster : public tt_device {
     Cluster(
         const std::string& sdesc_path,
         const std::set<chip_id_t>& target_devices,
-        const uint32_t& num_host_mem_ch_per_mmio_device = 1,
```
pjanevskiTT, line 672: nit: are you removing reference here just because it is a basic type?

pjanevskiTT (approved):
Looks good, left few nit comments

device/api/umd/device/cluster.h
joelsmithTT: Yes -- there's no need for this to be passed by reference.

device/hugepage.cpp
joelsmithTT: I'll update the issue after this gets merged.

README.md
```diff
@@ -5,7 +5,7 @@ Usermode Driver for Tenstorrent AI Accelerators
 ## Dependencies
 Required Ubuntu dependencies:
 ```
-sudo apt install -y libhwloc-dev cmake ninja-build
+sudo apt install -y cmake ninja-build libnuma-dev
```
broskoTT, line 8: You do have to add this to build_image.yaml, and run Build Docker workflow on this branch to build another image which will have this lib, right?

device/CMakeLists.txt
```diff
@@ -64,7 +63,7 @@ target_link_libraries(
     PRIVATE
         umd::Common
         umd::Firmware
-        hwloc
+        numa
```
broskoTT, line 66: @blozano-tt does having this lib in PRIVATE section mean this will get statically linked, and thus not needed when using libdevice, just when building?

blozano-tt: No, PRIVATE means that -lnuma linker flag will not be transitively passed to consumers of libdevice.so

It makes sense for a dependency to be private when it is not part of the public API interface, but is instead a private implementation detail. In other words, UMD needs to link libnuma, but UMD consumers don’t have to directly link it.

broskoTT: Right, so this solves a problem which you mentioned on that chat thread, or I understood that wrongly. libnuma is not needed when you want to USE umd, but it is needed when you want to build it.

blozano-tt: > libnuma is not needed when you want to USE umd

Not entirely true. It is not needed during link stage of compilation (between tt_metal and tt_umd). It is needed at runtime, because it is dynamically linked by UMD.

blozano-tt: My explanation may not have been great. 

I may have made it confusing between:
image

And, when the loader `ld` dynamically loads / links the library ??

Sorry, I never really polished my semantics here.

joelsmithTT: libnuma is needed at runtime.  `docker_install_common.sh` installs it.  Metal is installing it into their environments somehow, because their [device_pool.cpp uses it](https://github.com/tenstorrent/tt-metal/blame/59cf190b73c17c7c0ae110f27affbf89b5933124/tt_metal/impl/device/device_pool.cpp#L7)

device/pcie/pci_device.cpp
broskoTT: So this PR is waiting on resolving of https://github.com/tenstorrent/tt-metal/issues/15675 ?

joelsmithTT: Yes

joelsmithTT:
This has gone stale and remains blocked, so I'm abandoning it.  If the blocking issue https://github.com/tenstorrent/tt-metal/issues/15675 gets resolved, I will re-trace my steps and open a new PR.

PR #337, opened 2024-11-26
Change simulation logging to debug level

vtangTT:
### Issue
(Link to Github issue(s))

### Description
Before, when running with simulator this would spam the output because we would continuously poll the device at the same address.

### List of the changes
Changed to debug log level.

### Testing
This is a logging change and does not change functionality.

### API Changes


abhullar-tt (approved):
Im okay with this but do you think we should have some progress indicator that aren't just log_debug so we know the sim hasn't hung?

vtangTT:
> Im okay with this but do you think we should have some progress indicator that aren't just log_debug so we know the sim hasn't hung?

Hmm, yea in terms of CI for simulators, I'll have timeouts set for now if anything hangs. I'm not sure how to get a progress indicator though :(

Issue #257, opened 2024-11-04
Explore an IOMMU based scheme that mimics sysmem API for GS/WH


joelsmithTT:
In UMD, "sysmem" refers to 1G hugepages used as shared buffers between device and host.  The APIs for system include tt_SiliconDevice constructor (specifies number of "channels") and accessor methods (read/write/get pointer to buffer) that are channel based.

While preserving the API, write an alternate implementation that does not huge hugepages, but relies on the KMD ability to map physically noncontiguous buffers to a contiguous IO virtual address space seen by the device.

The challenges at the system level:
* WH hardware is limited to accessing an address space of 0x0-0xffff'ffff without iATU programming
* KMD IOVA allocation does not seem to support complete coverage of this address space
* With iATU we can work around this.  However, iATU is limited to 16 regions ~with a constraint that the IOVA must be naturally aligned with the region size.~ Edit - don't think this is accurate.
* Unlike VFIO, KMD API provides no userspace control over the IOVA ~- so some over-allocation may be necessary to achieve alignment~ Edit - see above
* ~KMD lacks an API to unmap~ Edit - this has been introduced

Challenge at the user driver level:
* Maintain the illusion of one to four channels, even if there are up to 16 sub allocations
* At runtime, determine whether to do sysmem via IOMMU or sysmem via hugepage

nsmithtt:
`dma_alloc_coherent` or `dma_pool_create` might be useful in getting aligned memory?

broskoTT:
I think that Joel already did a great job of mimicking current hugepage setup with iommu setup, we can close this one, it does not cover work currently being done in this field

PR #338, opened 2024-11-26
Add IOMMU support

joelsmithTT:
This implementation is not going to work very well for large allocations.  The problem is detailed in a comment in the code.  This is something we will have to address before we can deploy (i.e. enable system IOMMU in our infra).

### Issue
https://github.com/tenstorrent/tt-umd/issues/257

### Description
Adds support for using a normal (i.e. not backed by 1G page) buffer for sysmem.  This requires a system with IOMMU enabled and not in passthrough mode, and also a recent (>= 1.29.0) KMD.

### List of the changes
* Adds IOMMU detection logic to pci_device.cpp
* KMD version check if IOMMU is enabled
* Adds API for initializing per-device sysmem without hugepages
* Defaults to not using hugepages if the system IOMMU is enabled

### Testing
Manual testing on my development machine.

### API Changes

### DIFF
--- a/device/pcie/pci_device.cpp
+++ b/device/pcie/pci_device.cpp
@@ -88,6 +88,15 @@ T read_sysfs(const PciDeviceInfo &device_info, const std::string &attribute_name
     }
 }
 
+static bool detect_iommu(const PciDeviceInfo &device_info) {
+    try {
+        auto iommu_type = read_sysfs<std::string>(device_info, "iommu_group/type");
+        return iommu_type.substr(0, 3) == "DMA";  // DMA or DMA-FQ
+    } catch (...) {
+        return false;
+    }
+}
+
 static PciDeviceInfo read_device_info(int fd) {
     tenstorrent_get_device_info info{};
     info.in.output_size_bytes = sizeof(info.out);
@@ -258,6 +267,8 @@ tt::ARCH PciDeviceInfo::get_arch() const {
     return infos;
 }
 
+static const semver_t kmd_ver_for_iommu = semver_t(1, 29, 0);
+
 PCIDevice::PCIDevice(int pci_device_number) :
     device_path(fmt::format("/dev/tenstorrent/{}", pci_device_number)),
     pci_device_num(pci_device_number),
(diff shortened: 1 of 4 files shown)


alewycky-tenstorrent (approved):

device/pcie/pci_device.hpp
```diff
@@ -73,6 +74,7 @@ class PCIDevice {
     const int revision;              // PCI revision value from sysfs
     const tt::ARCH arch;             // e.g. Grayskull, Wormhole, Blackhole
     const semver_t kmd_version;      // KMD version
+    const bool behind_iommu;         // Whether the system is protected from this device by an IOMMU
```
broskoTT, line 77: behind_iommu -> iommu_enabled ?

device/pcie/pci_device.cpp
```diff
@@ -812,6 +869,55 @@ hugepage_mapping PCIDevice::get_hugepage_mapping(int channel) const {
...
+
+    // With IOMMU, this will probably fail on you if you're mapping something
+    // large.  The situation today is that the kernel driver uses a 32-bit DMA
+    // address mask, so all DMA allocations and mappings show up in the IOVA
+    // range of 0x0 to 0xffff'ffff.  According to syseng, we can get up to 3GB
+    // on Intel, 3.75GB on AMD, but this requires multiple mappings with small
+    // chunks, down to 2MB.  It's possible to make such non-contiguous mappings
+    // appear both virtually contiguous (to the application) and physically
+    // contiguous (to the NOC, using iATU), but it's not clear that this is
+    // worth the effort...  the scheme this is intended to replace supports up
```
broskoTT, line 896: Are you saying that the iommu solution doesn't offer 4GB space?
Are you also saying that we have a hard requirement to offer 4GB of space?

device/pcie/pci_device.cpp
```diff
@@ -812,6 +869,55 @@ hugepage_mapping PCIDevice::get_hugepage_mapping(int channel) const {
...
+    // to 4GB which is what application developers want.
+    //
+    // What can we do here?
```
broskoTT, line 899: So we're not sure that we'd want to use IOMMU at all, at this point?

device/pcie/pci_device.cpp
joelsmithTT: 1. IOMMU solution does offer > 4GB of space, but to use that much address space we need to lift an addressing constraint imposed by the driver.
2. I don't know.  My opinion is that any application can be made to work with less, but the request from the Metal team is to support 4GB because that's what they have with the non-IOMMU scheme.

device/pcie/pci_device.hpp
joelsmithTT: sure

device/pcie/pci_device.cpp
joelsmithTT: I think we do want to.  When we ship Blackhole, I think IOMMU should be supported.  I am in favor of making it a requirement for Blackhole.

For WH/GS it is less clear.  Users have a desire to move away from hugepages, but less of a desire to rewrite code that depends on bad UMD APIs.  UMD can maintain these APIs for the no-hugepage case, but to make them work identically (i.e. support 4GB) we need to lift the addressing constraint.

device/pcie/pci_device.cpp
```diff
@@ -802,6 +828,37 @@ bool PCIDevice::init_hugepage(uint32_t num_host_mem_channels) {
     return success;
 }
 
+bool PCIDevice::init_nohugepage(size_t size) {
```
pjanevskiTT, line 831: `init_iommu` instead of `init_nohugepage`?

device/pcie/pci_device.hpp
```diff
@@ -152,6 +154,11 @@ class PCIDevice {
      */
     tt::ARCH get_arch() const { return arch; }
 
+    /**
+     * @return whether the system is protected from this device by an IOMMU
+     */
+    bool is_behind_iommu() const { return behind_iommu; }
```
pjanevskiTT, line 160: same as Bojan's comment for naming, just for function -> `is_iommu_enabled`?

pjanevskiTT (approved):

device/pcie/pci_device.hpp
joelsmithTT: renamed

joelsmithTT:
RE: README - I created https://github.com/tenstorrent/tt-umd/issues/370 to track further work on this, which includes documenting it.

PR #340, opened 2024-11-27
Build ubench in CI (#305)

broskoTT:
### Issue
Original PR was #305 and then reverted in #335

### Description
tests/microbenchmark wasn't getting built by CI

### List of the changes
modify CI yaml to build it with the unit tests

### Testing
Builds locally, CI test pending

### API Changes


broskoTT:
@joelsmithTT I repeated the same PR as original one, and now you can see the failing runs for building tt_metal (which is the reason why I reverted the original one)
Please fix this before this goes in.

pjanevskiTT:
@joelsmithTT how do we run these benchmarks, can you add docs in the main README as part of this PR?

Issue #248, opened 2024-11-04
Create Chip abstraction under Cluster


broskoTT:
Now that tt_SiliconDevice has moved to tt::umd::Cluster, we can work forward towards the diagram https://docs.google.com/drawings/d/1-m1azdsBqMA0A6ATYRMfkhyeuOJuGCEI62N5a96LXj0/edit

Create proper Chip abstraction which takes away most of functionalities under the Cluster.

broskoTT:
I'll create several PRs to come to resolution of this:
- Create Chip abstraction underneath.
- Move Simulation and Mockup to extend Chip rather than Cluster.
  - tt_metal changes
- Delete tt_device
  - tt_metal changes
- Delete unnecessary Cluster api's, move them to Cluster->get_chip(id)->something().
  - tt_metal_changes

broskoTT:
I've continued working on this abstraction. We don't want to bog everything again in one class, so I'll be adding a couple of helper classes.

I tried to capture some ideas at the bottom of https://docs.google.com/drawings/d/1-m1azdsBqMA0A6ATYRMfkhyeuOJuGCEI62N5a96LXj0/edit

I'll be adding SysmemManager
I'll be adding something like a UMDMutex, or LockingManager (haven't decided on the naming yet), this is captured by #418 as well
There were some misalignments between @pjanevskiTT and me on the current TLBManager, and we realised that actually there should be two classes, TLBAllocator, and StaticTLBManager, which are depicted on the above diagram.

broskoTT:
After many many PRs, time to close this one as complete.

PR #351, opened 2024-11-28
Initial Chip class

broskoTT:
### Issue
First PR towards #248 

### Description
First in set of changes to introduce Chip abstraction class according to the plan https://docs.google.com/drawings/d/1-m1azdsBqMA0A6ATYRMfkhyeuOJuGCEI62N5a96LXj0/edit
This change also required minor rethinking of some parts of how Cluster is constructed (this took much time to figure out).

### List of the changes
- Added Chip class, and derived Local, Remote, and Mock classes. Local will be created for mmio chips, Remote for non-mmio. Mock chip is now only a placeholder, to show the Cluster constructor API changes, but it should merge with existing MockupDevice. Chip is an abstract class.
- Chip classes for now only hold soc descriptors and nothing else. Further PRs will gradually move out logic from Cluster to Chip.
- get_soc_descriptor and soc_descriptor_per_chip were moved around to conform to the new reality
- This was reverted after offline discussion: (Cluster constructor which takes soc descriptor was changed so that it takes tt_SocDescriptor object rather than yaml file. This is because the object also holds harvesting information.)
- Added another Cluster constructor which takes a set of chips. This is an experimental one, and users should know what they're doing and they can break the code easily.
- Previously mentioned new constructor was used to offer Cluster::create_mock_cluster which is a helper function to build you a cluster with mock chip.
- Wrote a test which showcases different constructor usages, which was a leftover from #277 

### Testing
Existing tests should prove no functional changes in this PR.

### API Changes
This PR has no API changes.

### DIFF
--- a/device/api/umd/device/chip/chip.h
+++ b/device/api/umd/device/chip/chip.h
@@ -14,17 +14,15 @@ namespace tt::umd {
 // An abstract class that represents a chip.
 class Chip {
 public:
-    Chip(chip_id_t chip_id, tt_SocDescriptor soc_descriptor);
+    Chip(tt_SocDescriptor soc_descriptor);
 
     virtual ~Chip() = default;
 
-    chip_id_t get_chip_id() const;
     tt_SocDescriptor& get_soc_descriptor();
 
     virtual bool is_mmio_capable() const = 0;
 
 private:
-    chip_id_t chip_id_;
     tt_SocDescriptor soc_descriptor_;
 };
 
--- a/device/api/umd/device/chip/chip.h
+++ b/device/api/umd/device/chip/chip.h
@@ -21,6 +21,8 @@ public:
     chip_id_t get_chip_id() const;
     tt_SocDescriptor& get_soc_descriptor();
 
+    virtual bool is_mmio_capable() const = 0;
+
 private:
     chip_id_t chip_id_;
     tt_SocDescriptor soc_descriptor_;
--- a/device/api/umd/device/chip/chip.h
+++ b/device/api/umd/device/chip/chip.h
@@ -1,0 +1,29 @@
+/*
+ * SPDX-FileCopyrightText: (c) 2024 Tenstorrent Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include "umd/device/tt_cluster_descriptor_types.h"
+#include "umd/device/tt_soc_descriptor.h"
+
+namespace tt::umd {
+
+// An abstract class that represents a chip.
+class Chip {
+public:
+    Chip(chip_id_t chip_id, tt_SocDescriptor soc_descriptor);
+
+    virtual ~Chip() = default;
+
+    chip_id_t get_chip_id() const;
+    tt_SocDescriptor& get_soc_descriptor();
+
+private:
+    chip_id_t chip_id_;
+    tt_SocDescriptor soc_descriptor_;
+};
+
+}  // namespace tt::umd
--- a/device/api/umd/device/chip/chip.h
+++ b/device/api/umd/device/chip/chip.h
@@ -6,8 +6,8 @@
 
 #pragma once
 
-#include "umd/device/tt_cluster_descriptor_types.h"
 #include "umd/device/tt_soc_descriptor.h"
+#include "umd/device/types/cluster_descriptor_types.h"
 
 namespace tt::umd {
 
--- a/device/api/umd/device/cluster.h
+++ b/device/api/umd/device/cluster.h
@@ -479,7 +479,8 @@ public:
      * @param skip_driver_allocs
      * @param clean_system_resource Specifies if host state from previous runs needs to be cleaned up.
      * @param perform_harvesting Allow the driver to modify the SOC descriptors per chip.
-     * @param simulated_harvesting_masks
+     * @param simulated_harvesting_masks Manually specify additional harvesting masks for the devices in the cluster.
+     * The ones defined by the devices itself have to be used, they will be merged with the ones passed here.
      */
     Cluster(
         const uint32_t& num_host_mem_ch_per_mmio_device = 1,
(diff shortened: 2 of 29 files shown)


device/api/umd/device/cluster.h
```diff
@@ -698,13 +695,30 @@ class Cluster : public tt_device {
...
+
+    // /**
+    //  * Helper function to construct a chip from info found in cluster descriptor.
+    //  * This allows users to construct a Chip instance from a logical chip id which is a cluster descriptor concept.
+    //  *
+    //  * @param logical_device_id Logical device id of the chip to create, as found in the cluster descriptor.
+    //  * @param cluster_desc Cluster descriptor to use. If not provided, it will be constructed. If many chips are
+    //  * constructed
+    //  */
+    // static std::unique_ptr<Chip> construct_chip_from_cluster(
```
broskoTT, line 719: Remove or surface

tests/api/test_cluster.cpp
```diff
@@ -56,6 +58,70 @@ void setup_wormhole_remote(Cluster* umd_cluster) {
...
+    tt_SocDescriptor soc_desc = tt_SocDescriptor(sdesc_path, harvesting_mask);
+    umd_cluster = std::make_unique<Cluster>(soc_desc, target_devices);
+    umd_cluster = nullptr;
+
+    // 4. Constructor for creating a cluster with mock chip.
+    umd_cluster = Cluster::create_mock_cluster();
+    umd_cluster = nullptr;
+
+    // // ?????
+    // // 5.1. First chip will be created using helper function from the cluster
```
broskoTT, line 94: Remove

device/cluster.cpp
```diff
@@ -578,43 +604,43 @@ Cluster::Cluster(
 }
 
 Cluster::Cluster(
-    const std::string& sdesc_path,
-    const std::set<chip_id_t>& target_devices,
+    std::unordered_map<chip_id_t, std::unique_ptr<Chip>>& chips,
```
broskoTT, line 607: @pjanevskiTT I couldn't resist offering this constructor, once I figured out that it enables me to do the create_mock_cluster thing without the hacked private constructor.
The static create_mock_cluster function can now be viewed just as a helper function to call the constructor with the right arguments for you.
I intentionally left out this constructor from the api tests, because that should not be a recommended way to do things for our users (but I still like that we leave that flexibility available).

pjanevskiTT: > I intentionally left out this constructor from the api tests, because that should not be a recommended way to do things for our users (but I still like that we leave that flexibility available).

broskoTT: I'd like to keep more flexibility until we have a better grasp over what do our customers needs. I envision this might unblock some experimentation without our intevention.

tests/api/test_cluster.cpp
```diff
@@ -56,6 +58,39 @@ void setup_wormhole_remote(Cluster* umd_cluster) {
 // This test should be one line only.
 TEST(ApiClusterTest, OpenAllChips) { std::unique_ptr<Cluster> umd_cluster = get_cluster(); }
 
+TEST(ApiClusterTest, DifferentConstructors) {
+    std::vector<int> pci_device_ids = PCIDevice::enumerate_devices();
+    // TODO: Make this test work on a host system without any tt devices.
+    if (pci_device_ids.empty()) {
+        GTEST_SKIP() << "No chips present on the system. Skipping test.";
+    }
```
pjanevskiTT, line 66: I like this guard of stopping run of silicon tests without silicon, can we somehow add it to all tests that we are running?

device/cluster.cpp
```diff
@@ -483,46 +489,49 @@ void Cluster::construct_cluster(
...
+
+std::unique_ptr<Chip> Cluster::construct_chip_from_cluster(chip_id_t chip_id, tt_ClusterDescriptor* cluster_desc) {
+    tt::ARCH arch = cluster_desc->get_arch(chip_id);
+    std::string soc_desc_path = tt_SocDescriptor::get_soc_descriptor_path(arch);
+    uint32_t harvesting_info = cluster_desc->get_harvesting_info().at(chip_id);
+    tt_SocDescriptor soc_desc = tt_SocDescriptor(soc_desc_path, harvesting_info);
+    return construct_chip_from_cluster(chip_id, cluster_desc, soc_desc);
+}
+
+void Cluster::add_chip(chip_id_t chip_id, std::unique_ptr<Chip> chip) {
```
pjanevskiTT, line 509: I see this is a private function, so we should use it in a proper way always, client can't misuse it, but maybe we could add a check whether chip_id is already in some of this maps. Feel free to ignore this change if you want

pjanevskiTT (approved):
Nice changes, this is in sync with my mental model of how Chip should fit in our cluster class as well. So you can resolve the comment, a lot of it quite small changes. So feel free to check it in after resolving comments

device/cluster.cpp
pjanevskiTT: agree

device/api/umd/device/chip/chip.h
```diff
@@ -0,0 +1,29 @@
...
+    Chip(tt_SocDescriptor soc_descriptor);
+
+    virtual ~Chip() = default;
+
+    tt_SocDescriptor& get_soc_descriptor();
+
+    virtual bool is_mmio_capable() const = 0;
+
+private:
+    tt_SocDescriptor soc_descriptor_;
```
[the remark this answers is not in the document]
broskoTT: I have no idea. It will be an easy fix nevertheless.
I plan to propose something and then to figure out a way to come to an agreement, and then implement the agreed changes in the whole codebase.

device/cluster.cpp
broskoTT: Probably a good futureproof safeguard, will add it.

device/api/umd/device/tt_simulation_device.h
```diff
@@ -63,6 +65,9 @@ class tt_SimulationDevice : public tt_device {
     virtual std::uint32_t get_host_channel_size(std::uint32_t device_id, std::uint32_t channel);
     virtual std::uint32_t get_numa_node_for_pcie_device(std::uint32_t device_id);
 
+protected:
+    std::unordered_map<chip_id_t, tt_SocDescriptor> soc_descriptor_per_chip = {};
```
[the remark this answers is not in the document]
broskoTT: I've revisited this member and related methods, and reverted some changes. This map is needed only due to get_virtual_soc_descriptors, which is used by metal_soc_descriptor, which will be removed when your corecoord changes go through.

device/api/umd/device/chip/chip.h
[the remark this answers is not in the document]
pjanevskiTT: makes sense

device/api/umd/device/cluster.h
```diff
@@ -698,15 +703,42 @@ class Cluster : public tt_device {
      * @param simulated_harvesting_masks
      */
     Cluster(
+        tt_SocDescriptor soc_desc,
```
broskoTT, line 706: Figure out what happens with harvesting now. Previously harvesting was passed with cluster descriptor. Then we moved to using default cluster descriptor. But now, that the object is passed, who does the harvesting?

broskoTT: Synced offline with Pavle. 
Will wait for his changes which remove harvesting from cluster.cpp class and start using tt_socdescriptor's harvesting

broskoTT: After more discussion offline, I've reverted this change. We don't want to allow the user to have an option to pass wrong harvesting info.

Issue #103, opened 2024-10-03
tt::umd namespace


broskoTT:
Everything in umd should be inside tt::umd namespace, so when used it is clear from which component it came from.

pjanevskiTT:
@broskoTT can you take this?

PR #357, opened 2024-11-30
Refactor types 

broskoTT:
### Issue
Related to #103 

### Description
Reformat ARCH into tt::umd::Arch
While at it, I looked at other types, and realised that cluster_types are a thing. 
And now we have 3 or 4 different type files, so this warrants a dedicated folder for these small types.

### List of the changes
- Initially renamed tt::ARCH to tt::umd::Arch. Then figured out through comments that it actually makes more sense to leave it inside tt:: namespace. Then figured out the change from ARCH -> Arch is unnecessary, and doesn't justify the amount of changes needed.
- Renamed file from tt_arch_type.h to types/arch.h
- Renamed tt_cluster_descriptor_types.h to types/cluster_descriptor_types.h
- Moved structs from cluster.h to types/cluster_types.h
- Renamed xy_pair.h to types/xy_pair.h
- Marked tt_xy_pair.h to be deleted.
- Consolidated multiple arch_to_string and string_to_arch methods into a single source of truth, which should be fine for all applications

Following changes will be 
1. adding everything to tt::umd namespace
2. removing tt_ prefix from all classes.

### Testing
Code builds in existing CI tests.

### API Changes
This PR has API changes:
- [x] tt_metal approved PR pointing to this branch: https://github.com/tenstorrent/tt-metal/pull/15753
- [x] tt_debuda approved PR pointing to this branch:https://github.com/tenstorrent/tt-lens/pull/203


device/api/umd/device/types/arch.h
```diff
@@ -0,0 +1,62 @@
...
+ * Enums for different architectures.
+ */
+enum class Arch {
+    GRAYSKULL = 1,
+    WORMHOLE_B0 = 2,
+    BLACKHOLE = 3,
+    Invalid = 0xFF,
+};
+
+static inline tt::umd::Arch arch_from_str(const std::string &arch_str) {
```
broskoTT, line 30: After https://github.com/tenstorrent/tt-umd/pull/345, consolidate with this function

device/api/umd/device/architecture_implementation.h
```diff
@@ -25,7 +25,7 @@ class architecture_implementation {
 public:
     virtual ~architecture_implementation() = default;
 
-    virtual tt::ARCH get_architecture() const = 0;
+    virtual tt::umd::Arch get_architecture() const = 0;
```
pjanevskiTT, line 28: Not really sure about this one, I think Arch may be something that could potentially go into common repo, when there is something like that. Not sure whether we should put it in umd namespace. I guess this is one of the things that are, let's say, layer below UMD so basic types that we are building umd on top of. Maybe I would like to keep this in tt namespace so we can more easily distinguish things that could go to common repo.

device/api/umd/device/types/arch.h
```diff
@@ -0,0 +1,62 @@
...
+    } else if ((arch_str_lower == "wormhole") || (arch_str_lower == "wormhole_b0")) {
+        return tt::umd::Arch::WORMHOLE_B0;
+    } else if (arch_str_lower == "blackhole") {
+        return tt::umd::Arch::BLACKHOLE;
+    } else {
+        return tt::umd::Arch::Invalid;
+    }
+}
+
+static inline std::string arch_to_str(const tt::umd::Arch arch) {
```
pjanevskiTT, line 44: Like to see that these functions are in one place. No action

pjanevskiTT (commented):
Like the changes, let me know what you think about namespace for ARCH

tests/api/test_mockup_device.cpp
```diff
@@ -24,42 +24,25 @@ std::string get_env_arch_name() {
...
-    }
-    if (arch_str == "Invalid" || arch_str == "INVALID") {
-        return tt::ARCH::Invalid;
-    }
-
-    throw std::runtime_error(arch_str + " is not recognized as tt::ARCH.");
-}
-
-std::string get_soc_descriptor_file(tt::ARCH arch) {
+std::string get_soc_descriptor_file(tt::umd::Arch arch) {
```
broskoTT, line 27: Remove this

device/api/umd/device/architecture_implementation.h
broskoTT: Makes sense, I'll revert to tt::.
I somewhat agree on the common argument, but I wouldn't like to fuel the fire there, and would like to keep it here in the umd folder outside common. When we consider adding common utils repo, we can revisit this.

pjanevskiTT: Makes sene. I think we should still put as many things into umd namespace, but yeah seems like we are on the same page here

pjanevskiTT (approved):
approved after discussion about tt::ARCH

PR #359, opened 2024-12-01
TTDevice init

broskoTT:
### Issue
Related to #98 

### Description
Initial TTDevice class. It holds PCIDevice and arch implementation.
Gradually, I'd like to move all branches on arch type to be moved from cluster and pci_device to tt_device, and that class to be the only one in the stack which offers different implementation for each arch. According to: https://docs.google.com/drawings/d/1-m1azdsBqMA0A6ATYRMfkhyeuOJuGCEI62N5a96LXj0/edit

### List of the changes
- Create TTDevice class.
- Created arch specific TTDevice classes, which currently don't hold much implementation.
- architecture_implementation is moved to TTDevice, and ~half of PCIDevice is moved. PCIDevice should hold only non-arch specific code, except getting the arch itself. There were only mild, compile related changes in the functions moved from PCIDevice to TTDevice.
- cluster.cpp and tests changed accordingly.
- read_checking_offset moved to architecture_implementation
- Blackhole destructor specific code moved to BlackholeTTDevice from PCIDevice.

### Testing
Existing CI tests should be enough.

### API Changes
But I scheduled post commit tests just to be sure.
- [x] All post-commit tests : https://github.com/tenstorrent/tt-metal/actions/runs/12156329972
- [x] Blackhole post-commit tests : https://github.com/tenstorrent/tt-metal/actions/runs/12156332357

### DIFF
--- a/device/api/umd/device/cluster.h
+++ b/device/api/umd/device/cluster.h
@@ -819,7 +819,7 @@ public:
     virtual std::uint32_t get_numa_node_for_pcie_device(std::uint32_t device_id);
     virtual tt_version get_ethernet_fw_version() const;
 
-    TTDevice* get_tt_device(int device_id) const;
+    TTDevice* get_tt_device(chip_id_t device_id) const;
 
     // Destructor
     virtual ~Cluster();
@@ -973,7 +973,9 @@ private:
     std::set<chip_id_t> target_devices_in_cluster = {};
     std::set<chip_id_t> target_remote_chips = {};
     tt::ARCH arch_name;
-    std::unordered_map<chip_id_t, std::unique_ptr<TTDevice>> m_tt_device_map;  // Map of enabled tt devices
+
+    // Map of enabled tt devices
+    std::unordered_map<chip_id_t, std::unique_ptr<TTDevice>> m_tt_device_map;
     std::shared_ptr<tt_ClusterDescriptor> cluster_desc;
 
     // remote eth transfer setup
--- a/device/api/umd/device/cluster.h
+++ b/device/api/umd/device/cluster.h
@@ -17,9 +17,9 @@
 #include "tt_silicon_driver_common.hpp"
 #include "tt_soc_descriptor.h"
 #include "tt_xy_pair.h"
-#include "umd/device/pci_device.hpp"
 #include "umd/device/tlb.h"
 #include "umd/device/tt_cluster_descriptor_types.h"
+#include "umd/device/tt_device/tt_device.h"
 #include "umd/device/tt_io.hpp"
 
 using TLB_DATA = tt::umd::tlb_data;
@@ -818,8 +818,8 @@ public:
     virtual std::uint32_t get_host_channel_size(std::uint32_t device_id, std::uint32_t channel);
     virtual std::uint32_t get_numa_node_for_pcie_device(std::uint32_t device_id);
     virtual tt_version get_ethernet_fw_version() const;
-    // TODO: This should be accessible through public API, probably to be moved to tt_device.
-    PCIDevice* get_pci_device(int device_id) const;
+
+    TTDevice* get_tt_device(int device_id) const;
 
     // Destructor
     virtual ~Cluster();
@@ -973,8 +973,7 @@ private:
     std::set<chip_id_t> target_devices_in_cluster = {};
     std::set<chip_id_t> target_remote_chips = {};
     tt::ARCH arch_name;
-    std::unordered_map<chip_id_t, std::unique_ptr<PCIDevice>> m_pci_device_map;  // Map of enabled pci devices
-    int m_num_pci_devices;  // Number of pci devices in system (enabled or disabled)
+    std::unordered_map<chip_id_t, std::unique_ptr<TTDevice>> m_tt_device_map;  // Map of enabled tt devices
     std::shared_ptr<tt_ClusterDescriptor> cluster_desc;
 
     // remote eth transfer setup
--- a/device/api/umd/device/pci_device.hpp
+++ b/device/api/umd/device/pci_device.hpp
@@ -19,27 +19,9 @@
 #include "umd/device/tt_cluster_descriptor_types.h"
 #include "umd/device/tt_xy_pair.h"
 
-// TODO: this is used up in cluster.cpp but that logic ought to be
-// lowered into the PCIDevice class since it is specific to PCIe cards.
-// See /vendor_ip/synopsys/052021/bh_pcie_ctl_gen5/export/configuration/DWC_pcie_ctl.h
-static const uint64_t UNROLL_ATU_OFFSET_BAR = 0x1200;
-
-// TODO: this is a bit of a hack... something to revisit when we formalize an
-// abstraction for IO.
-// BAR0 size for Blackhole, used to determine whether write block should use BAR0 or BAR4
-static const uint64_t BAR0_BH_SIZE = 512 * 1024 * 1024;
-
-constexpr unsigned int c_hang_read_value = 0xffffffffu;
-
 namespace tt::umd {
-class architecture_implementation;
 struct semver_t;
 }  // namespace tt::umd
-
-struct dynamic_tlb {
-    uint64_t bar_offset;      // Offset that address is mapped to, within the PCI BAR.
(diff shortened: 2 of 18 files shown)


device/api/umd/device/cluster.h
```diff
@@ -978,8 +978,7 @@ class Cluster : public tt_device {
     std::set<chip_id_t> target_devices_in_cluster = {};
     std::set<chip_id_t> target_remote_chips = {};
     tt::ARCH arch_name;
-    std::unordered_map<chip_id_t, std::unique_ptr<PCIDevice>> m_pci_device_map;  // Map of enabled pci devices
-    int m_num_pci_devices;  // Number of pci devices in system (enabled or disabled)
+    std::unordered_map<chip_id_t, std::unique_ptr<TTDevice>> m_tt_device_map;  // Map of enabled tt devices
```
pjanevskiTT, line 981: Can you move the comment above the code line?

device/api/umd/device/pci_device.hpp
```diff
@@ -219,12 +154,24 @@ class PCIDevice {
 
     uint32_t read_checking_offset;
 
-private:
-    bool is_hardware_hung();
-
     template <typename T>
-    T *get_register_address(uint32_t register_offset);
+    T *get_register_address(uint32_t register_offset) {
```
pjanevskiTT, line 158: why was implementation moved to .hpp file?

device/cluster.cpp
```diff
@@ -1432,43 +1429,47 @@ void Cluster::init_pcie_iatus() {
...
         return ret_val;
     } else if (arch_name == tt::ARCH::BLACKHOLE) {
         // MT Inital BH - Try to enable this, but double check "regval == 33"
-        // int ret_val = 0;
-        // PCIDevice *dev = m_pci_device_map.begin()->second->hdev;
 
-        // uint32_t mapped_reg = dev->set_dynamic_tlb(m_pci_device_map.begin()->second,
-        // dev->get_architecture_implementation()->get_reg_tlb(), tt_xy_pair(1, 0), 0xffb20108,
-        // harvested_coord_translation.at(logical_device_id)).bar_offset;
+        // uint32_t mapped_reg = tt_device
```
pjanevskiTT, line 1463: off topic for this PR, but do we know what is the exact issue with this commented code. I see Milos's comment above, but not really sure what is the problem just by reading it. Can we at least track this with the issue as it is

pjanevskiTT (approved):
Nice, left a few nit comments

pjanevskiTT:
So even if this PR is not changing the API to the clients, I would like to see tt-metal post commit pipeline pass with all this changes. I guess our tests are testing the correctness well enough (apart from Galaxy), but I would like to make sure that perf is not worse with all of this layering. And metal has a lot of tests that would confirm this

device/api/umd/device/cluster.h
```diff
@@ -823,8 +823,8 @@ class Cluster : public tt_device {
     virtual std::uint32_t get_host_channel_size(std::uint32_t device_id, std::uint32_t channel);
     virtual std::uint32_t get_numa_node_for_pcie_device(std::uint32_t device_id);
     virtual tt_version get_ethernet_fw_version() const;
-    // TODO: This should be accessible through public API, probably to be moved to tt_device.
-    PCIDevice* get_pci_device(int device_id) const;
+
+    TTDevice* get_tt_device(int device_id) const;
```
joelsmithTT, line 827: TTDevice abstraction only exists for PCIe-attached chips, correct?  Should this be clear from the interface?

device/cluster.cpp
```diff
@@ -1751,11 +1753,11 @@ void* Cluster::host_dma_address(std::uint64_t offset, chip_id_t src_device_id, u
...
 
 // Wrapper for throwing more helpful exception when not-enabled pci intf is accessed.
-inline PCIDevice* Cluster::get_pci_device(int device_id) const {
-    if (!m_pci_device_map.count(device_id)) {
+inline TTDevice* Cluster::get_tt_device(int device_id) const {
+    if (!m_tt_device_map.count(device_id)) {
         throw std::runtime_error(fmt::format("device_id: {} attempted to be accessed, but is not enabled.", device_id));
     }
-    return m_pci_device_map.at(device_id).get();
+    return m_tt_device_map.at(device_id).get();
```
joelsmithTT, line 1760: Since the `m_tt_device_map` is a `std::unordered_map<chip_id_t, std::unique_ptr<TTDevice>>`, does it make sense for this method to accept a `chip_id_t` instead of an `int`?

joelsmithTT (commented):
> PCIDevice should hold only non-arch specific code, except getting the arch itself

The conventions for how BAR0 and BAR4 are mapped are architecture-specific.  Is the plan for TTDevice implementations to manage these mappings?

If KMD starts enforcing proper resource management for inbound TLB windows and the firmware messaging interface, is the TTDevice abstraction still necessary?

device/api/umd/device/pci_device.hpp
broskoTT: This is a mess and we might refactor it anyway.
But this is a template function, it has to be implemented in header file. We got away with it before because it was private, and thus only used in pci_device.cpp

device/api/umd/device/cluster.h
broskoTT: Yes, it will be. Look at https://github.com/tenstorrent/tt-umd/pull/351
The LocalChip (and not remotechip) will eventually hold this pointer, so it will be clear.

device/tt_device/tt_device.cpp
```diff
@@ -0,0 +1,313 @@
...
+            .x_end = static_cast<uint64_t>(translated_end_coords.x),
+            .y_end = static_cast<uint64_t>(translated_end_coords.y),
+            .x_start = static_cast<uint64_t>(translated_start_coords.x),
+            .y_start = static_cast<uint64_t>(translated_start_coords.y),
+            .mcast = multicast,
+            .ordering = ordering,
+            // TODO #2715: hack for Blackhole A0, will potentially be fixed in B0.
+            // Using the same static vc for reads and writes through TLBs can hang the card. It doesn't even have to be
+            // the same TLB. Dynamic vc should not have this issue. There might be a perf impact with using dynamic vc.
+            .static_vc = (arch == tt::ARCH::BLACKHOLE) ? false : true,
```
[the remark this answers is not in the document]
broskoTT: Yes, I don't want to flag all of them. All of "if arch == tt::ARCH::somearch" in pci_device and cluster should go to tt_device. And then further all such if's should be refactored into virutal functions implemented by gs wh bh.

This is just an initial commit, I intentionally wanted to change as less stuff as possible.

device/cluster.cpp
broskoTT: https://github.com/tenstorrent/tt-umd/issues/367

broskoTT:
> So even if this PR is not changing the API to the clients, I would like to see tt-metal post commit pipeline pass with all this changes. I guess our tests are testing the correctness well enough (apart from Galaxy), but I would like to make sure that perf is not worse with all of this layering. And metal has a lot of tests that would confirm this

No worries, I've left the part to populate tt_metal and tt_lens PRs.
There are API changes, soc descriptor is now an object not a file path :)

broskoTT:
> So even if this PR is not changing the API to the clients, I would like to see tt-metal post commit pipeline pass with all this changes. I guess our tests are testing the correctness well enough (apart from Galaxy), but I would like to make sure that perf is not worse with all of this layering. And metal has a lot of tests that would confirm this

Scheduled.



> > PCIDevice should hold only non-arch specific code, except getting the arch itself
> 
> The conventions for how BAR0 and BAR4 are mapped are architecture-specific. Is the plan for TTDevice implementations to manage these mappings?
> 
> If KMD starts enforcing proper resource management for inbound TLB windows and the firmware messaging interface, is the TTDevice abstraction still necessary?

For BARs, I'll have to think about it more. The mapping itself no. But how they are used, e.g. which one is used for system_register access, etc, possibly yes.
If KMD implements that, we should remove it from TTDevice. If TTDevice becomes small enough such that it doesn't make sense anymore, we'll remove it.

joelsmithTT (approved):

Issue #364, opened 2024-12-03
LargeAddressTlb failing


broskoTT:
This test has been failing intermittently 
https://github.com/tenstorrent/tt-metal/actions/runs/12134589477/job/33832069926
https://github.com/tenstorrent/tt-metal/actions/runs/12131095838/job/33822630072

Issue #362, opened 2024-12-03
SpecifyIllegalDeassertRiscResetOnCore failing


broskoTT:
For some time this tests seems to be failing intermittently. This happens on both UMD CI and tt_metal CI.
Failing runs examples: 
https://github.com/tenstorrent/tt-umd/actions/runs/11889008268/job/33124886089?pr=310
https://github.com/tenstorrent/tt-umd/actions/runs/11966959645/job/33363528724?pr=323

This was also seen when trying to add more similar tests in #308  

This is a failure of SiliconDriverWH.LargeAddressTlb which could also be related:
https://github.com/tenstorrent/tt-metal/actions/runs/12134589477/job/33832069926
https://github.com/tenstorrent/tt-metal/actions/runs/12131095838/job/33822630072

@abhullar-tt @joelsmithTT

broskoTT:
Culprit is likely running a bogus firmware.
We've removed all deasserts/asserts from UMD tests, should be added in https://github.com/tenstorrent/tt-umd/issues/422

PR #366, opened 2024-12-03
Fix SiliconDriverWH.LargeAddressTlb

joelsmithTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/364

### Description
Firmware may change the low 16 bits of the scratch register used for this test.

### List of the changes
Mask off the low 16 bits of the values read back from hardware before comparing.

### Testing
CI

### API Changes

### DIFF
--- a/tests/wormhole/test_silicon_driver_wh.cpp
+++ b/tests/wormhole/test_silicon_driver_wh.cpp
@@ -981,6 +981,11 @@ TEST(SiliconDriverWH, LargeAddressTlb) {
     // Read the scratch register via a different TLB, different code path:
     cluster.read_from_device(&value2, ARC_CORE, addr, sizeof(uint32_t), "REG_TLB");
 
+    // Mask off lower 16 bits; FW changes these dynamically:
+    value0 &= 0xffff0000;
+    value1 &= 0xffff0000;
+    value2 &= 0xffff0000;
+
     // Check that the values are the same:
     EXPECT_EQ(value1, value0);
     EXPECT_EQ(value2, value0);


tests/wormhole/test_silicon_driver_wh.cpp
```diff
@@ -981,6 +981,11 @@ TEST(SiliconDriverWH, LargeAddressTlb) {
     // Read the scratch register via a different TLB, different code path:
     cluster.read_from_device(&value2, ARC_CORE, addr, sizeof(uint32_t), "REG_TLB");
 
+    // Mask off lower 16 bits; FW changes these dynamically:
```
broskoTT, line 984: Could it be the same cause as for https://github.com/tenstorrent/tt-umd/issues/362 , since that one also has lower 16 bits changing?

tests/wormhole/test_silicon_driver_wh.cpp
```diff
@@ -981,6 +981,11 @@ TEST(SiliconDriverWH, LargeAddressTlb) {
...
+    value0 &= 0xffff0000;
```
broskoTT, line 985: Can you elaborate (at least here in github comment), why is this expected? The chips are reset before running tests right? When they are reset firmware starts executing and I guess it comes to some stable state? Do you think that this exact register might be used as a scratch register and oscillating between values, which this write/read operation interferes with?

tests/wormhole/test_silicon_driver_wh.cpp
joelsmithTT: Don't think so - this register is controlled by ARC FW.
That problem looks like Tensix reset in a weird state.

tests/wormhole/test_silicon_driver_wh.cpp
joelsmithTT: I asked Andrew about the behavior of this register, I will reproduce my question and his answer:

Q: Can you explain the behavior of the register at 0x1ff30060 in Wormhole BAR0?  I have a unit test that reads it, and also reads it via TLB window to ARC.  In CI, there's intermittent failures because the BAR0 read doesn't match the TLB window read.  0xc0de0042 vs 0xc0de004d.  I don't want to mask off the low 16 bits without understanding why...

A: That's scratch 0, the "post code". The ARC FW writes status info there and it changes. A valid ARC FW post code has 0xc0de in the upper 16 bits. The low 16 bits may indicate where we are in the boot process or what the FW did last (e.g. handle a message). I'd say that 0x1 and 0x30-0x4F are normal codes.


This test wants to ensure that we are looking at the same thing regardless of whether we're using the REG_TLB path (dynamically assigned TLB window) or via a statically assigned TLB window.  I added this test while fixing addressing bugs that unnecessarily limited static TLB assignment and REG_TLB path to 32 bit addresses.  Since the NOC address of this register is > 32 bits, and because it is visible in BAR0, it is a good candidate for this sanity check.

broskoTT: Does this mean that boot process advances exactly at the moment when you're reading the value in three different ways? It should eventually settle on a value, right?

Also, whatever is the answer, can we just read some other address without masking?

broskoTT (approved):

PR #369, opened 2024-12-04
Harvesting mask NOC layout

pjanevskiTT:
### Issue
/ 

### Description

Harvesting mask targets tensix rows based on the physical chip layout. So far I have assumed it is NOC layout. This PR is fixing that.

### List of the changes

- Impement shuffling of harvesting mask bits to be be in sync with NOC layout in all cordinate managers 

### Testing

Unit tests.

### API Changes

No API changes

### DIFF
--- a/device/coordinate_manager.cpp
+++ b/device/coordinate_manager.cpp
@@ -28,6 +28,7 @@ CoordinateManager::CoordinateManager(
     tensix_grid_size(tensix_grid_size),
     tensix_cores(tensix_cores),
     tensix_harvesting_mask(tensix_harvesting_mask),
+    physical_layout_tensix_harvesting_mask(tensix_harvesting_mask),
     dram_grid_size(dram_grid_size),
     dram_cores(dram_cores),
     dram_harvesting_mask(dram_harvesting_mask),
@@ -277,6 +278,10 @@ void CoordinateManager::fill_arc_physical_translated_mapping() {
     }
 }
 
+size_t CoordinateManager::get_tensix_harvesting_mask() const { return physical_layout_tensix_harvesting_mask; }
+
+size_t CoordinateManager::get_dram_harvesting_mask() const { return dram_harvesting_mask; }
+
 void CoordinateManager::assert_create_coordinate_manager(
     const tt::ARCH arch, const size_t tensix_harvesting_mask, const size_t dram_harvesting_mask) {
     log_assert(
@@ -285,6 +290,26 @@ void CoordinateManager::assert_create_coordinate_manager(
     if (arch == tt::ARCH::BLACKHOLE) {
         log_assert(get_num_harvested(dram_harvesting_mask) <= 1, "Only one DRAM bank can be harvested on Blackhole");
     }
+}
+
+void CoordinateManager::shuffle_tensix_harvesting_mask(const std::vector<uint32_t>& harvesting_locations) {
+    std::vector<uint32_t> sorted_harvesting_locations = harvesting_locations;
+    std::sort(sorted_harvesting_locations.begin(), sorted_harvesting_locations.end());
+    size_t new_harvesting_mask = 0;
+    uint32_t pos = 0;
+    while (tensix_harvesting_mask > 0) {
+        if (tensix_harvesting_mask & 1) {
+            uint32_t sorted_position =
+                std::find(
+                    sorted_harvesting_locations.begin(), sorted_harvesting_locations.end(), harvesting_locations[pos]) -
+                sorted_harvesting_locations.begin();
+            new_harvesting_mask |= (1 << sorted_position);
+        }
+        tensix_harvesting_mask >>= 1;
+        pos++;
+    }
+
+    tensix_harvesting_mask = new_harvesting_mask;
 }
 
 std::shared_ptr<CoordinateManager> CoordinateManager::create_coordinate_manager(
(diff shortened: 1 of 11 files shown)


device/coordinate_manager.cpp
```diff
@@ -405,6 +405,26 @@ void CoordinateManager::assert_create_coordinate_manager(
...
+    std::sort(sorted_harvesting_locations.begin(), sorted_harvesting_locations.end());
+    size_t new_harvesting_mask = 0;
+    uint32_t pos = 0;
+    while (tensix_harvesting_mask > 0) {
+        if (tensix_harvesting_mask & 1) {
+            uint32_t sorted_position =
+                std::find(
+                    sorted_harvesting_locations.begin(), sorted_harvesting_locations.end(), harvesting_locations[pos]) -
+                sorted_harvesting_locations.begin();
+            new_harvesting_mask |= (1 << sorted_position);
```
broskoTT, line 419: So tensix_harvesting_mask won't be the original harvesting_mask passed.
Can you comment on the field tensix_harvesting_mask in header file about it not being passed harvesting mask, but rather "normalized", or "noc" harvesting mask, whereas passed one is expected to be a a physical one? Maybe add that comment to the harvesting parameter of the constructor.

broskoTT: The resulting harvesting mask is in logical_coordinates right? 
for blackhole for example, each successive bit would correspond to logical coord 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13
which translates to noc0 real coordinate: 1, 2, 3, 4, 5, 6, 7, 10, 11, 12, 13, 14, 15, 16

device/coordinate_manager.cpp
```diff
@@ -405,6 +405,26 @@ void CoordinateManager::assert_create_coordinate_manager(
     }
 }
 
+void CoordinateManager::shuffle_tensix_harvesting_mask(const std::vector<uint32_t>& harvesting_locations) {
+    std::vector<uint32_t> sorted_harvesting_locations = harvesting_locations;
```
broskoTT, line 409: How about 
```
std::vector<std::pair<uint32_t, uint32_t>> sorted_harvesting_locations;
for (int i = 0; i < harvesting_locations.size(); i ++) {
  sorted_harvesting_locations.push_back({harvesting_locations[i], i});
}
// Check if lambda is required.
std::sort(sorted_harvesting_locations.begin(), sorted_harvesting_locations.end(), [](auto &left, auto &right) {
    return left.first < right.first;
});
```
Then below you can write just this instead of having to search again:
```
uint32_t sorted_position = sorted_harvesting_location[pos].second;
```

device/coordinate_manager.cpp
```diff
@@ -405,6 +405,26 @@ void CoordinateManager::assert_create_coordinate_manager(
...
+    std::sort(sorted_harvesting_locations.begin(), sorted_harvesting_locations.end());
+    size_t new_harvesting_mask = 0;
+    uint32_t pos = 0;
+    while (tensix_harvesting_mask > 0) {
```
broskoTT, line 413: Consider:
```
for (uint32_t pos = 0; pos < harvesting_locations.size() && tensix_harvesting_mask > 0; pos ++, tensix_harvesting_mask >>= 1)
```

tests/api/test_core_coord_translation_gs.cpp
```diff
@@ -52,8 +52,10 @@ TEST(CoordinateManager, CoordinateManagerGrayskullTopLeftCore) {
 // We expect that the top left core will have virtual and physical coordinates (1, 1) and (1, 2) for
 // the logical coordinates if the first row is harvested.
 TEST(CoordinateManager, CoordinateManagerGrayskullTopLeftCoreHarvesting) {
+    // This is targeting first row of Tensix cores on NOC layout.
+    const size_t harvesting_mask = (1 << 8);
```
broskoTT, line 56: Do you think it might make sense to have inverse of HARVESTING_NOC_LOCATIONS? For blackhole that would be:
`static const std::vector<uint32_t> LOGICAL_HARVESTING_LAYOUT = {0, 2, 4, 6, 8, 10, 12, 13, 11, 9, 7, 5, 3, 1};`

Then
`harvesting_mask = (1 << LOGICAL_HARVESTING_LAYOUT[0])`
and this is the same for all arch

broskoTT (approved):

device/coordinate_manager.cpp
broskoTT: I just saw get_harvested_noc_rows, it is a similar function, but the harvesting returned is in physical coordinates, meaning you don't have to sort here. Whereas your returned harvesting is in logical coordinates, right? Consider what would be better here

device/coordinate_manager.cpp
pjanevskiTT: yes like that, I will add a comment

device/coordinate_manager.cpp
pjanevskiTT: I think the code you suggested is not doing the same things as the original one, but maybe you referenced that through the other two comments. As you mentioned, new harvesting mask should be in logical coordinates lets say. But what is missing here is maybe being able to retrieve the original harvesting mask, I will add that. I think it is much easier to compute everything coordinate related if we shuffle the mask this way in the beginning

device/coordinate_manager.cpp
pjanevskiTT: I think while is a bit more readable here because we are not going to do some calculation for every `pos` index, it depends a bit more on the harvesting mask. Not strong opionions on this however, it is probably equally readable

tests/api/test_core_coord_translation_gs.cpp
pjanevskiTT: makes sense, will do it

pjanevskiTT: I am jos not sure whether we would like to expose this through the *_implementation files. Thoughts?

Issue #355, opened 2024-11-29
Coordinate Manager internal redesign


pjanevskiTT:
Look into cleaner/more performant implementation for Coordinate manager

Comments from #266 that can be useful

https://github.com/tenstorrent/tt-umd/pull/266#discussion_r1861831772

https://github.com/tenstorrent/tt-umd/pull/266#discussion_r1861839022

https://github.com/tenstorrent/tt-umd/pull/266#discussion_r1861963518

https://github.com/tenstorrent/tt-umd/pull/266#discussion_r1861965875

https://github.com/tenstorrent/tt-umd/pull/266#discussion_r1861985294

PR #374, opened 2024-12-06
Coordinate manager maps redesign

pjanevskiTT:
### Issue

Some parts of https://github.com/tenstorrent/tt-umd/issues/355

### Description

Redesign maps of the Coordinate Manager class. Instead of per core type maps, we use one map now for all core types. Virtual to physical to translated mappings are now implemented for harvested coordinates as well. We may want to use those cores sometimes so in order to support proper reading/writing to these cores. Now all translations are going throug the physical (NOC0) coordinate system, since soc descriptors are in physical coordinate system, and each core has those coordinates.

### List of the changes

- Remove per core type map
- Add one map for all cores
- Implement translation for harvested tensix cores
- Implement translation for harvested dram cores

### Testing

Existing unit tests, added more

### API Changes

No API changes

### DIFF
--- a/device/api/umd/device/coordinate_manager.h
+++ b/device/api/umd/device/coordinate_manager.h
@@ -63,6 +63,8 @@ protected:
         const tt_xy_pair& pcie_grid_size,
         const std::vector<tt_xy_pair>& pcie_cores);
 
+    void initialize();
+
     virtual void translate_tensix_coords();
     virtual void translate_dram_coords();
     virtual void translate_eth_coords();
@@ -136,3 +138,5 @@ protected:
     const tt_xy_pair pcie_grid_size;
     const std::vector<tt_xy_pair>& pcie_cores;
 };
+
+// friend
--- a/device/api/umd/device/coordinate_manager.h
+++ b/device/api/umd/device/coordinate_manager.h
@@ -16,20 +16,6 @@
 
 class CoordinateManager {
 public:
-    CoordinateManager(
-        const tt_xy_pair& tensix_grid_size,
-        const std::vector<tt_xy_pair>& tensix_cores,
-        const size_t tensix_harvesting_mask,
-        const tt_xy_pair& dram_grid_size,
-        const std::vector<tt_xy_pair>& dram_cores,
-        const size_t dram_harvesting_mask,
-        const tt_xy_pair& eth_grid_size,
-        const std::vector<tt_xy_pair>& eth_cores,
-        const tt_xy_pair& arc_grid_size,
-        const std::vector<tt_xy_pair>& arc_cores,
-        const tt_xy_pair& pcie_grid_size,
-        const std::vector<tt_xy_pair>& pcie_cores);
-
     static std::shared_ptr<CoordinateManager> create_coordinate_manager(
         tt::ARCH arch,
         const tt_xy_pair& tensix_grid_size,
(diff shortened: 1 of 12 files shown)


pjanevskiTT:
Thanks @broskoTT @tt-vjovanovic for useful tips on previous PR, I will implement other suggestions later on, with this we have baseline to support API changes to Cluster

device/coordinate_manager.cpp
```diff
@@ -38,179 +38,53 @@ CoordinateManager::CoordinateManager(
...
-}
-
 CoreCoord CoordinateManager::to_physical(const CoreCoord core_coord) {
     switch (core_coord.coord_system) {
         case CoordSystem::PHYSICAL:
             return core_coord;
         case CoordSystem::VIRTUAL:
         case CoordSystem::TRANSLATED:
-            return to_physical(to_logical(core_coord));
+        case CoordSystem::LOGICAL: {
```
nhuang-tt, line 47: I recommend adding a default case to the switch statements to throw an error otherwise if a new enum value is added it could become very hard to remember and find where it's missing

device/coordinate_manager.cpp
```diff
@@ -38,179 +38,53 @@ CoordinateManager::CoordinateManager(
...
         case CoordSystem::VIRTUAL:
-            return to_translated(to_logical(core_coord));
+            return to_translated(to_physical(core_coord));
         case CoordSystem::TRANSLATED:
             return core_coord;
     }
-
-    // Coord system is surely logical.
-    auto& logical_mapping = get_logical_to_translated(core_coord.core_type);
-    return logical_mapping[{core_coord.x, core_coord.y}];
```
tt-vjovanovic, line 213: This comment refers to `CoreCoord CoordinateManager::to(const CoreCoord core_coord, const CoordSystem coord_system)` function....

I think you should write it like this:
```
CoreCoord CoordinateManager::to(const CoreCoord core_coord, const CoordSystem coord_system) {
    if (core_coord.coord_system == CoordSystem::PHYSICAL) {
        return from_physical_map.at({{core_coord.x, core_coord.y}, coord_system});
    } else {
        return to(to_physical_map.at(core_coord), coord_system);
        // You can also replace with next line if you don't like recursion:
        // return from_physical_map.at({to_physical_map.at(core_coord), coord_system});
    }
}
```
This way, you always check correctness of input coordinate... Again, if you want better error handling, you can use `std::map::find` instead of `std::map::at`...

tt-vjovanovic (changes requested):

device/coordinate_manager.cpp
```diff
@@ -38,179 +38,53 @@ CoordinateManager::CoordinateManager(
...
+        }
     }
-
-    // Coord system is surely logical.
-    auto& logical_mapping = get_logical_to_physical(core_coord.core_type);
-    return logical_mapping[{core_coord.x, core_coord.y}];
 }
 
 CoreCoord CoordinateManager::to_virtual(const CoreCoord core_coord) {
     switch (core_coord.coord_system) {
```
broskoTT, line 55: Related to what Nigel and Vuk commented on the other function.
If you add an identity map to to_physcal and from_physical, so that it knows how to handle PHYSICAL coords as well, then that fuction (to_physical) will reduce to 2 lines of code. 
But also, in that case, all of these to_virtual, to_logical, to_translated) functions would be simpler, since you wouldn't need to specially handle the case for CoordSystem::PHYSICAL.

And then, even further, we can say that it is fine to "do" the translation even though if same CoreSystem was passed (so, without optimization in case coord system was the same as requested), then all of these functions would also reduce to one lines of code. For example:
```
CoreCoord CoordinateManager::to_virtual(const CoreCoord core_coord) {
  return from_physical_map[{to_physical_map[core_coord], CoordSystem::VIRTUAL}]
}
```

broskoTT (approved):

device/coordinate_manager.cpp
pjanevskiTT: Added identity mapping, so no need for to_virtual, to_physical... functions, everything can happen in to function

device/coordinate_manager.cpp
pjanevskiTT: removed to_physical, everything goes through to function now

device/blackhole/blackhole_coordinate_manager.cpp
```diff
@@ -41,7 +41,7 @@ BlackholeCoordinateManager::BlackholeCoordinateManager(
 }
 
 void BlackholeCoordinateManager::translate_tensix_coords() {
-    size_t num_harvested_x = __builtin_popcount(tensix_harvesting_mask);
+    size_t num_harvested_x = CoordinateManager::get_num_harvested(tensix_harvesting_mask);
```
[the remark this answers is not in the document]
pjanevskiTT: Let me know if this is more readable 9cecd977540a21a375b50eafc8d5f14985e97e82

device/api/umd/device/coordinate_manager.h
```diff
@@ -118,53 +118,8 @@ class CoordinateManager {
...
-
-    std::map<tt_xy_pair, tt::umd::CoreCoord>& get_logical_to_translated(CoreType core_type);
-    std::map<tt_xy_pair, tt::umd::CoreCoord>& get_logical_to_virtual(CoreType core_type);
-    std::map<tt_xy_pair, tt::umd::CoreCoord>& get_logical_to_physical(CoreType core_type);
-
-    std::map<tt_xy_pair, tt::umd::CoreCoord>& get_physical_to_logical(CoreType core_type);
-    std::map<tt_xy_pair, tt::umd::CoreCoord>& get_virtual_to_logical(CoreType core_type);
-    std::map<tt_xy_pair, tt::umd::CoreCoord>& get_translated_to_logical(CoreType core_type);
+    std::map<tt::umd::CoreCoord, tt_xy_pair> to_physical_map;
+    std::map<std::pair<tt_xy_pair, CoordSystem>, tt::umd::CoreCoord> from_physical_map;
```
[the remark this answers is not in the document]
pjanevskiTT: per offline discussion, leaving this as it is

Issue #376, opened 2024-12-07
TLB data verification code is wrong


joelsmithTT:
`bool tlb_data::check(const tlb_offsets &offset) const` is incorrect.  It says,

```
 local_offset > ((1 << (offset.x_end - offset.local_offset)) - 1) |
```
in an apparent attempt to ensure that the address being programmed actually fits.  For Blackhole, offset.x_end is 43.  This code is undefined and is incorrectly rejecting a valid TLB configuration on me.

[ As a bit of a digression, it is puzzling why this is referred to as "local_offset"... since the window is aligned at a 2 megabyte boundary, the notion of a local offset makes more sense when performing IO through the window (if your target address is not a multiple of 2 megabytes, you will need a local offset into the window).  What the code refers to as "local offset" is actually the upper bits of the address.  This is unnecessarily confusing. ]

PR #379, opened 2024-12-08
Fix undefined behavior bug

joelsmithTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/376

### Description
The literal `1` has type `int`.  The left shift operation is undefined when the result can not be represented in the left operand type.

### List of the changes
* Use a 64 bit 1.

### Testing
N/A

### API Changes
N/A

### DIFF
--- a/device/tlb.cpp
+++ b/device/tlb.cpp
@@ -7,15 +7,16 @@
 namespace tt::umd {
 
 bool tlb_data::check(const tlb_offsets &offset) const {
-    return local_offset > ((1 << (offset.x_end - offset.local_offset)) - 1) |
-           x_end > ((1 << (offset.y_end - offset.x_end)) - 1) | y_end > ((1 << (offset.x_start - offset.y_end)) - 1) |
-           x_start > ((1 << (offset.y_start - offset.x_start)) - 1) |
-           y_start > ((1 << (offset.noc_sel - offset.y_start)) - 1) |
-           noc_sel > ((1 << (offset.mcast - offset.noc_sel)) - 1) |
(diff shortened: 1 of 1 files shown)


device/tlb.cpp
```diff
@@ -7,15 +7,16 @@
...
-    return local_offset > ((1 << (offset.x_end - offset.local_offset)) - 1) |
-           x_end > ((1 << (offset.y_end - offset.x_end)) - 1) | y_end > ((1 << (offset.x_start - offset.y_end)) - 1) |
-           x_start > ((1 << (offset.y_start - offset.x_start)) - 1) |
-           y_start > ((1 << (offset.noc_sel - offset.y_start)) - 1) |
-           noc_sel > ((1 << (offset.mcast - offset.noc_sel)) - 1) |
-           mcast > ((1 << (offset.ordering - offset.mcast)) - 1) |
-           ordering > ((1 << (offset.linked - offset.ordering)) - 1) |
-           linked > ((1 << (offset.static_vc - offset.linked)) - 1) |
-           static_vc > ((1 << (offset.static_vc_end - offset.static_vc)) - 1);
+    return local_offset > ((1ULL << (offset.x_end - offset.local_offset)) - 1) |
```
broskoTT, line 10: Is it possible to easily cover this by any existing/new test?

broskoTT (approved):

device/tlb.cpp
joelsmithTT: Yes although I'm reluctant to spend time writing such tests because I don't think tlb.cpp and the architecture_implementation business is a good way to do this, plus all of this belongs in the kernel driver anyway.

PR #382, opened 2024-12-09
Add Wall Werror

broskoTT:
### Issue
No issue
Reported from tt_metal that they're seeing some warnings

### Description
Guard against warnings when building. Ignore unused variables/functions not to spam with warnings, and these don't do much damage anyway.

### List of the changes
- Wall to show more warnings
- Wno-unused-variable, Wno-unused-function, Wno-unused-but-set-variable not to spam for unused
- Werror not to allow warnings passing our CI
- Did some changes in our repo which were throwing warnings

### Testing
Current CI tests are fine

### API Changes

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,6 +53,15 @@ message(STATUS "UMD build type: ${CMAKE_BUILD_TYPE}")
 
 include(dependencies)
 
+# Enable all warnings and treat them as errors
+add_compile_options(
+    -Wall
+    -Werror
+    -Wno-unused-variable
+    -Wno-unused-function
+    -Wno-unused-but-set-variable
+)
+
 add_subdirectory(common)
 add_subdirectory(device)
 add_subdirectory(src)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,12 +54,15 @@ message(STATUS "UMD build type: ${CMAKE_BUILD_TYPE}")
 include(dependencies)
 
 # Enable all warnings and treat them as errors
+# Ignore unused variables and functions since there is no harm.
+# TODO: Re-enable sign-compare
 add_compile_options(
     -Wall
     -Werror
     -Wno-unused-variable
     -Wno-unused-function
     -Wno-unused-but-set-variable
+    -Wno-sign-compare
 )
 
 add_subdirectory(common)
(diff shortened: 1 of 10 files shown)


device/coordinate_manager.cpp
```diff
@@ -455,6 +455,8 @@ std::shared_ptr<CoordinateManager> CoordinateManager::create_coordinate_manager(
                 tt::umd::blackhole::ARC_CORES,
                 tt::umd::blackhole::PCIE_GRID_SIZE,
                 tt::umd::blackhole::PCIE_CORES);
+        case tt::ARCH::Invalid:
```
blozano-tt, line 458: Ok, for now, but what about getting rid of this enum value altogether in the future?

broskoTT: Makes sense but I see it is currently used at various places in metal. I should check if in one of those cases we don't want to throw, etc...

CMakeLists.txt
```diff
@@ -53,6 +53,15 @@ message(STATUS "UMD build type: ${CMAKE_BUILD_TYPE}")
...
 include(dependencies)
 
+# Enable all warnings and treat them as errors
+add_compile_options(
+    -Wall
+    -Werror
+    -Wno-unused-variable
+    -Wno-unused-function
+    -Wno-unused-but-set-variable
+)
```
blozano-tt, line 63: Are these flags the same between clang and gcc?

blozano-tt (approved):

CMakeLists.txt
broskoTT: I've added GCC CI build in #384, so that one should hopefully be enough to verify this.

tests/api/test_cluster_descriptor.cpp
```diff
@@ -21,7 +21,7 @@ TEST(ApiClusterDescriptorTest, DetectArch) {
         EXPECT_THROW(tt_ClusterDescriptor::detect_arch(0), std::runtime_error);
     } else {
         tt::ARCH arch = tt_ClusterDescriptor::detect_arch(0);
-        EXPECT_NE(arch, tt::ARCH::Invalid);
+        EXPECT_TRUE(arch != tt::ARCH::Invalid);
```
broskoTT, line 24: Move operator<< inside namespace tt, and remove this change.

PR #389, opened 2024-12-10
Fix core locations

pjanevskiTT:
### Issue
/

### Description

Fix some core locations that were incorrect in implementanion h files

### List of the changes

- Fix PCIE core for wormhole
- Fix ARC core for wormhole
- Format DRAM and tensix cores

### Testing

ci

### API Changes

no api changes


device/api/umd/device/wormhole_implementation.h
```diff
@@ -103,49 +103,44 @@ enum class arc_message_type {
...
+     {5, 5}, {5, 6},  {5, 7}}};
+// clang-format on
 
 // TODO: DRAM locations should be deleted. We keep it for compatibility with
 // the existing code in clients which rely on DRAM_LOCATIONS.
 static const std::vector<tt_xy_pair> DRAM_LOCATIONS = DRAM_CORES;
 
 static const tt_xy_pair ARC_GRID_SIZE = {1, 1};
-static const std::vector<tt_xy_pair> ARC_CORES = {{0, 2}};
+static const std::vector<tt_xy_pair> ARC_CORES = {{0, 10}};
```
broskoTT, line 139: I'm afraid there might be more issues like this.
Why was this not caught by tests? I guess tests were written so that this passes :) Is there some other codepath which we know is working which we can test against? Or is just the set of constants used by the clients (tt-lens for example) just a different set of constants?

broskoTT (approved):

device/api/umd/device/wormhole_implementation.h
pjanevskiTT: Just a different set of constants. I went through the git blame for this one. The wrong constant in UMD were there for a long time, and soc descriptor coordinates (which were the one everyone was using) were different from the constants in implementation files. No one was using implementation constants until Coordinate manager, and it was properly working for the translation of the wrong constants, it was expecting these. The bug came when @tt-vjovanovic was using coordinate manager created with just arch, and was expecting to translate soc descriptor coordinates (which in theory should be the same). Since no one was using implementation coordinates, I think we are good. 

I checked other values as well

Issue #372, opened 2024-12-05
Figure out how Simulation/Mock device fits into the API redesign


broskoTT:
It should probably be a specialization of the new Chip class.

PR #396, opened 2024-12-11
[Chip] Move TTDevice to local chip

broskoTT:
### Issue
Related to #351 

### Description
Another PR in a push to define Chip class properly.
This PR moves tt_device under Chip, specifically under LocalChips.

### List of the changes
- Add TTDevice to LocalChip
- Renamed all_chips, target_chips and remote_chips to have more consistent naming.
- MockChip disabled at the moment, it will start working by itself at some point, when enough stuff is moved to Chip class.

### Testing
Existing tests

### API Changes

### DIFF
--- a/device/api/umd/device/cluster.h
+++ b/device/api/umd/device/cluster.h
@@ -790,7 +790,7 @@ private:
     std::vector<tt::ARCH> archs_in_cluster = {};
     std::set<chip_id_t> all_chip_ids_ = {};
     std::set<chip_id_t> remote_chip_ids_ = {};
-    std::set<chip_id_t> mmio_chip_ids_ = {};
+    std::set<chip_id_t> local_chip_ids_ = {};
     std::unordered_map<chip_id_t, std::unique_ptr<Chip>> chips_;
     tt::ARCH arch_name;
 
--- a/device/api/umd/device/cluster.h
+++ b/device/api/umd/device/cluster.h
@@ -312,20 +312,6 @@ public:
     }
 
     /**
-     * Get the total number of chips in the cluster based on the network descriptor.
-     */
-    virtual int get_number_of_chips_in_cluster() {
-        throw std::runtime_error("---- tt_device::get_number_of_chips_in_cluster is not implemented\n");
-    }
-
-    /**
-     * Get the logical ids for all chips in the cluster
-     */
-    virtual std::unordered_set<chip_id_t> get_all_chips_in_cluster() {
-        throw std::runtime_error("---- tt_device::get_all_chips_in_cluster is not implemented\n");
-    }
-
-    /**
      * Get cluster descriptor object being used in UMD instance.
      */
     virtual tt_ClusterDescriptor* get_cluster_description() {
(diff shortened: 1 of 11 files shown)


device/api/umd/device/cluster.h
```diff
@@ -830,14 +815,12 @@ class Cluster : public tt_device {
     tt_driver_noc_params noc_params;
     tt_driver_eth_interface_params eth_interface_params;
     std::vector<tt::ARCH> archs_in_cluster = {};
-    std::set<chip_id_t> target_devices_in_cluster = {};
-    std::set<chip_id_t> target_remote_chips = {};
-    std::set<chip_id_t> all_target_mmio_devices = {};
+    std::set<chip_id_t> all_chip_ids_ = {};
+    std::set<chip_id_t> remote_chip_ids_ = {};
+    std::set<chip_id_t> mmio_chip_ids_ = {};
```
pjanevskiTT, line 820: why not call this local chip straight away?

device/chip/local_chip.cpp
```diff
@@ -6,9 +6,14 @@
 
 #include "umd/device/chip/local_chip.h"
 
+#include "umd/device/tt_device/tt_device.h"
+
 namespace tt::umd {
 
-LocalChip::LocalChip(tt_SocDescriptor soc_descriptor) : Chip(soc_descriptor) {}
+LocalChip::LocalChip(tt_SocDescriptor soc_descriptor, int pci_device_num) :
+    Chip(soc_descriptor), tt_device_(TTDevice::create(pci_device_num)) {}
```
pjanevskiTT, line 14: nit: `pci_device_num` might sound like total number of pci devices. Consider renaming this to something like `pci_device_id`

device/api/umd/device/cluster.h
```diff
@@ -648,13 +634,12 @@ class Cluster : public tt_device {
...
     virtual std::unordered_map<chip_id_t, uint32_t> get_harvesting_masks_for_soc_descriptors();
     virtual void translate_to_noc_table_coords(chip_id_t device_id, std::size_t& r, std::size_t& c);
-    virtual int get_number_of_chips_in_cluster();
-    virtual std::unordered_set<chip_id_t> get_all_chips_in_cluster();
     virtual tt_ClusterDescriptor* get_cluster_description();
     static int detect_number_of_chips();
     static std::vector<chip_id_t> detect_available_device_ids();
     virtual std::set<chip_id_t> get_target_mmio_device_ids();
     virtual std::set<chip_id_t> get_target_remote_device_ids();
+    virtual std::set<chip_id_t> get_target_device_ids();
```
pjanevskiTT, line 642: nit: maybe rename to `get_chips_ids`

device/api/umd/device/cluster.h
```diff
@@ -648,13 +634,12 @@ class Cluster : public tt_device {
     virtual bool using_harvested_soc_descriptors();
     virtual std::unordered_map<chip_id_t, uint32_t> get_harvesting_masks_for_soc_descriptors();
     virtual void translate_to_noc_table_coords(chip_id_t device_id, std::size_t& r, std::size_t& c);
-    virtual int get_number_of_chips_in_cluster();
```
pjanevskiTT, line 651: Is this function not being used or we just don't want to support this interface since client can always call 
```
int num = get_target_device_ids().size();
```

tests/api/test_cluster.cpp
```diff
@@ -84,15 +83,18 @@ TEST(ApiClusterTest, DifferentConstructors) {
     umd_cluster = std::make_unique<Cluster>(sdesc_path, target_devices);
     umd_cluster = nullptr;
 
-    // 4. Constructor for creating a cluster with mock chip.
-    umd_cluster = Cluster::create_mock_cluster();
-    umd_cluster = nullptr;
+    // TODO: This doesn't work at the moment.
```
pjanevskiTT, line 86: open an issue as a reminder to reenable this

pjanevskiTT (approved):
left few nit comments, mostly naming, otherwise looks good

device/api/umd/device/cluster.h
broskoTT: we are using name `device` throughout the whole class to actually refer to a chip. I don't want to change this now, we might change all of these at some point.

And `target` is there because this refers to only the chips chosen in the constructor to be used. As opposed to `detect_available_device_ids` which returns all chips available on system. I also don't want to change other _target_ functions, because that would lead to an API change which I'd have to address in tt_metal, and this PR doesn't have an API change for now. Leaving others with "target" in name would then be another inconsistency.

We can consider changing `get_target_mmio_device_ids` `get_target_remote_device_ids` `get_target_device_ids` and `detect_available_device_ids`, but I'd leave it out of this PR for now.

device/api/umd/device/cluster.h
broskoTT: Both.
Also not `get_target_device_ids().size()` but rather `detect_available_device_ids().size()`, see previous comment :) It adds to confusion I guess.

tests/api/test_cluster.cpp
broskoTT: It already has to be done as part of https://github.com/tenstorrent/tt-umd/issues/372

Issue #220, opened 2024-10-28
[Harvesting] Provide Cluster API for CoreCoord


pjanevskiTT:
Provide `Cluster` (read/write) API for all types of coordinates (logical, physical, virtual)

pjanevskiTT:
Closed by #401

PR #401, opened 2024-12-16
Implement Cluster CoreCoord API

pjanevskiTT:
### Issue

https://github.com/tenstorrent/tt-umd/issues/220)

### Description

Provide Cluster API with CoreCoord. Every function that was expecting virtual coordinates has the same function just with CoreCoord structure. Old API is still there just to not break tt-metal and tt-lens at the moment. When the clients have switched to the new API we can delete the old one. In the meantime, internals of Cluster can be redesigned in parallel with client work on switching to new API. So this should be a two phase change. This is the first one where the new API is added, and when clients have switched to the new API second phase will be deleting the old API 

### List of the changes

- Separate Cluster API into 3 groups (something we want to keep, to delete and new API)
- Add API with CoreCoord
- Change API tests to use the new API

### Testing

UMD unit tests

### API Changes

Old API is still there, so no API changes at the moment

### DIFF
--- a/device/api/umd/device/cluster.h
+++ b/device/api/umd/device/cluster.h
@@ -9,6 +9,7 @@
 #include <cstdint>
 #include <memory>
 #include <set>
+#include <stdexcept>
 #include <string>
 #include <unordered_set>
 #include <vector>
@@ -73,6 +74,15 @@ public:
         throw std::runtime_error("---- tt_device::configure_tlb is not implemented\n");
     }
 
+    virtual void configure_tlb(
+        chip_id_t logical_device_id,
+        tt::umd::CoreCoord core,
+        int32_t tlb_index,
+        uint64_t address,
+        uint64_t ordering = TLB_DATA::Relaxed) {
+        throw std::runtime_error("---- tt_device::configure_tlb is not implemented\n");
+    }
+
     /**
      * Set ordering mode for dynamic/fallback TLBs (passed into driver constructor).
      *
@@ -94,6 +104,12 @@ public:
      */
     virtual void configure_active_ethernet_cores_for_mmio_device(
         chip_id_t mmio_chip, const std::unordered_set<tt_xy_pair>& active_eth_cores_per_chip) {
+        throw std::runtime_error(
+            "---- tt_device::configure_active_ethernet_cores_for_mmio_device is not implemented\n");
+    }
+
+    virtual void configure_active_ethernet_cores_for_mmio_device(
+        const std::unordered_set<tt::umd::CoreCoord>& active_eth_cores_per_chip, chip_id_t mmio_chip) {
         throw std::runtime_error(
             "---- tt_device::configure_active_ethernet_cores_for_mmio_device is not implemented\n");
     }
@@ -125,6 +141,13 @@ public:
         throw std::runtime_error("---- tt_device::deassert_risc_reset_at_core is not implemented\n");
     }
 
+    virtual void deassert_risc_reset_at_core(
+        const chip_id_t chip,
+        const tt::umd::CoreCoord core,
+        const TensixSoftResetOptions& soft_resets = TENSIX_DEASSERT_SOFT_RESET) {
+        throw std::runtime_error("---- tt_device::deassert_risc_reset_at_core is not implemented\n");
+    }
+
     /**
      * Broadcast assert soft Tensix Reset to the entire device.
      */
@@ -138,6 +161,10 @@ public:
      * @param core Chip and core being targeted.
      */
     virtual void assert_risc_reset_at_core(tt_cxy_pair core) {
+        throw std::runtime_error("---- tt_device::assert_risc_reset_at_core is not implemented\n");
+    }
+
+    virtual void assert_risc_reset_at_core(const chip_id_t chip, const tt::umd::CoreCoord core) {
         throw std::runtime_error("---- tt_device::assert_risc_reset_at_core is not implemented\n");
     }
 
@@ -180,6 +207,16 @@ public:
         throw std::runtime_error("---- tt_device::write_to_device is not implemented\n");
     }
 
+    virtual void write_to_device(
+        const void* mem_ptr,
+        uint32_t size_in_bytes,
+        chip_id_t chip,
+        tt::umd::CoreCoord core,
+        uint64_t addr,
+        const std::string& tlb_to_use) {
+        throw std::runtime_error("---- tt_device::write_to_device is not implemented\n");
+    }
+
     virtual void broadcast_write_to_cluster(
         const void* mem_ptr,
         uint32_t size_in_bytes,
@@ -206,6 +243,16 @@ public:
         throw std::runtime_error("---- tt_device::read_from_device is not implemented\n");
     }
 
+    virtual void read_from_device(
+        void* mem_ptr,
+        chip_id_t chip,
+        tt::umd::CoreCoord core,
+        uint64_t addr,
+        uint32_t size,
+        const std::string& fallback_tlb) {
+        throw std::runtime_error("---- tt_device::read_from_device is not implemented\n");
+    }
+
     /**
      * Write uint32_t vector to specified address and channel on host (defined for Silicon).
      *
@@ -229,6 +276,11 @@ public:
         throw std::runtime_error("---- tt_device::l1_membar is not implemented\n");
     }
 
+    virtual void l1_membar(
+        const chip_id_t chip, const std::unordered_set<tt::umd::CoreCoord>& cores, const std::string& fallback_tlb) {
+        throw std::runtime_error("---- tt_device::l1_membar is not implemented\n");
+    }
+
     virtual void dram_membar(
         const chip_id_t chip, const std::string& fallback_tlb, const std::unordered_set<uint32_t>& channels = {}) {
         throw std::runtime_error("---- tt_device::dram_membar is not implemented\n");
@@ -236,6 +288,11 @@ public:
 
     virtual void dram_membar(
         const chip_id_t chip, const std::string& fallback_tlb, const std::unordered_set<tt_xy_pair>& cores = {}) {
+        throw std::runtime_error("---- tt_device::dram_membar is not implemented\n");
+    }
+
+    virtual void dram_membar(
+        const chip_id_t chip, const std::unordered_set<tt::umd::CoreCoord>& cores, const std::string& fallback_tlb) {
         throw std::runtime_error("---- tt_device::dram_membar is not implemented\n");
     }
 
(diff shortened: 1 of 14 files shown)


pjanevskiTT:
@pgkeller

device/api/umd/device/cluster.h
```diff
@@ -666,18 +729,45 @@ class Cluster : public tt_device {
...
+        void* mem_ptr,
+        chip_id_t chip,
+        tt::umd::CoreCoord core,
+        uint64_t addr,
+        uint32_t size,
+        const std::string& fallback_tlb);
+    std::optional<std::tuple<uint32_t, uint32_t>> get_tlb_data_from_target(
+        const chip_id_t chip, const tt::umd::CoreCoord core);
+    tt::Writer get_static_tlb_writer(const chip_id_t chip, const tt::umd::CoreCoord target);
+    virtual tt::umd::CoreCoord to(
```
pgkeller, line 763: I don't see the base class so not clear to me - why do we need a virtual fn to do the coordinate conversion?
Also, "to" is pretty terse.

pgkeller (commented):
looks good overall to me, just a couple questions about "to".
please be sure to get Almeet's review

device/api/umd/device/cluster.h
```diff
@@ -452,6 +499,11 @@ class tt_device {
     virtual const tt_SocDescriptor& get_soc_descriptor(chip_id_t chip_id) const {
         return soc_descriptor_per_chip.at(chip_id);
     }
+    virtual tt::umd::CoreCoord to(
```
broskoTT, line 502: rename to something more descriptive (when calling this function, you don't reference return type).
Also since this is in cluster, the "to" might be confusing as it is not referring to "cluster".
Maybe "translate_coord_to"

device/api/umd/device/cluster.h
```diff
@@ -452,6 +499,11 @@ class tt_device {
...
+        const chip_id_t chip, const tt::umd::CoreCoord core_coord, const CoordSystem coord_system) {
+        throw std::runtime_error("---- tt_device::to is not implemented\n");
+        return tt::umd::CoreCoord();
```
broskoTT, line 505: Not needed after throw statement.

device/cluster.cpp
```diff
@@ -1000,6 +1012,15 @@ void Cluster::deassert_risc_reset_at_core(tt_cxy_pair core, const TensixSoftRese
     }
 }
 
+void Cluster::deassert_risc_reset_at_core(
+    const chip_id_t chip, const CoreCoord core, const TensixSoftResetOptions& soft_resets) {
+    tt_cxy_pair virtual_core;
+    const CoreCoord virtual_coord = to(chip, core, CoordSystem::VIRTUAL);
+    virtual_core.x = virtual_coord.x;
+    virtual_core.y = virtual_coord.y;
```
broskoTT, line 1020: Consider adding "get_xy_pair" to CoreCoord class, which would eliminate the need to do this. And then change everywhere in this class.

broskoTT (approved):
Looking forward to metal team using the new api...

device/api/umd/device/cluster.h
broskoTT: I've left the same comment about "to".
Regarding virtual, you can see we have "virtual" all around. We should do a sweep and remove all of it from this Cluster class. I'm waiting on enough API changes going through so we can finally remove the tt_device abstract class, and then we'll probably fix this one as well.
You're right that this should not have `virtual`

device/api/umd/device/cluster.h
```diff
@@ -452,6 +499,11 @@ class tt_device {
     virtual const tt_SocDescriptor& get_soc_descriptor(chip_id_t chip_id) const {
         return soc_descriptor_per_chip.at(chip_id);
     }
+    virtual tt::umd::CoreCoord to(
+        const chip_id_t chip, const tt::umd::CoreCoord core_coord, const CoordSystem coord_system) {
```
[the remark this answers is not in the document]
abhullar-tt: I agree, it gets confusing to have multiple entry points doing same thing. Unless the soc desc becomes a read only descriptor, in this case I think it makes sense to have all apis in Cluster

device/cluster.cpp
```diff
@@ -498,8 +499,8 @@ std::unique_ptr<Chip> Cluster::construct_chip_from_cluster(chip_id_t chip_id, tt
     // true.
     // TODO: This should be changed, harvesting should be done in tt_socdescriptor's constructor and not as part of
     // cluster class.
-    // uint32_t harvesting_info = cluster_desc->get_harvesting_info().at(chip_id);
-    tt_SocDescriptor soc_desc = tt_SocDescriptor(soc_desc_path /*, harvesting_info*/);
+    uint32_t tensix_harvesting_mask = cluster_desc->get_harvesting_info().at(chip_id);
```
abhullar-tt, line 502: note, we will eventually have dram and eth harvesting info for BH

device/cluster.cpp
```diff
@@ -678,6 +679,17 @@ void Cluster::configure_active_ethernet_cores_for_mmio_device(
     non_mmio_transfer_cores_customized = true;
 }
 
+void Cluster::configure_active_ethernet_cores_for_mmio_device(
```
abhullar-tt, line 682: how are active vs. idle eth being derived in UMD today? I am putting together a document that will describe how this is being done for BH.

device/api/umd/device/cluster.h
```diff
@@ -97,6 +98,15 @@ class tt_device {
         throw std::runtime_error("---- tt_device::configure_tlb is not implemented\n");
     }
 
+    virtual void configure_tlb(
+        chip_id_t logical_device_id,
+        tt::umd::CoreCoord core,
+        int32_t tlb_index,
+        uint64_t address,
+        uint64_t ordering = TLB_DATA::Relaxed) {
+        throw std::runtime_error("---- tt_device::configure_tlb is not implemented\n");
```
tt-vjovanovic, line 107: Consider using pure virtual functions instead of runtime throw. It is easier to find places during compile time instead having 100% code coverage with tests...

device/cluster.cpp
tt-vjovanovic: Isn't `create-ethernet-map` returning list of connections between ETH cores across devices?

device/cluster.cpp
pjanevskiTT: Yes, we already have tests with these, putting this into cluster path after we have arc messages for blackhole implemented

tests/api/test_cluster.cpp
```diff
@@ -229,40 +205,88 @@ TEST(ApiClusterTest, SimpleIOSpecificChips) {
...
-        }
+        const CoreCoord any_core = soc_desc.get_cores(CoreType::TENSIX)[0];
 
         std::cout << "Reading from chip " << chip_id << " core " << any_core.str() << std::endl;
 
         std::vector<uint8_t> readback_data(data_size, 0);
-        umd_cluster->read_from_device(readback_data.data(), any_core_global, 0, data_size, "LARGE_READ_TLB");
+        umd_cluster->read_from_device(readback_data.data(), chip_id, any_core, 0, data_size, "LARGE_READ_TLB");
+
+        // umd_cluster->wait_for_non_mmio_flush(chip_id);
```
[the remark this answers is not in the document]
pjanevskiTT: by mistake, fixed it

device/cluster.cpp
pjanevskiTT: It is, but this was just adding the same API that is public in Cluster to be the same just with CoreCoord. If this API ends up being not used, we will remove it. As far as idle vs active, since we have a meeting tomorrow for it, we will discuss it there

device/cluster.cpp
pjanevskiTT: This is tt_cxy_pair, so it needs chip as well which we don't have in CoreCoord. Idea is to use (chip, CoreCoord) instead of tt_cxy_pair

device/api/umd/device/tt_core_coordinates.h
```diff
@@ -83,3 +83,8 @@ struct CoreCoord : public tt_xy_pair {
 };
 
 }  // namespace tt::umd
+
```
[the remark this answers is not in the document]
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/411

device/cluster.cpp
broskoTT: Ah, it's the opposite case.
In that case you can do deassert_risc_reset_at_core({chip , to(chip, core, CoordSystem::VIRTUAL)}, soft_resets);

pjanevskiTT: to is deleted from Cluster, there was a comment to have translation only on Soc desc level. And it would return CoreCoord, but the old API expects tt_cxy_pair. I would keep this, we are going to dismantle this by removing old API

device/cluster.cpp
```diff
@@ -2932,6 +2973,16 @@ void Cluster::l1_membar(
     }
 }
 
+void Cluster::l1_membar(
+    const chip_id_t chip, const std::unordered_set<tt::umd::CoreCoord>& cores, const std::string& fallback_tlb) {
```
tt-vjovanovic, line 2977: Is there a way to use `std::ranges::input_range` instead of `std::unordered_set<>`?

device/cluster.cpp
```diff
@@ -503,8 +504,8 @@ std::unique_ptr<Chip> Cluster::construct_chip_from_cluster(chip_id_t chip_id, tt
     // true.
     // TODO: This should be changed, harvesting should be done in tt_socdescriptor's constructor and not as part of
     // cluster class.
-    // uint32_t harvesting_info = cluster_desc->get_harvesting_info().at(chip_id);
-    tt_SocDescriptor soc_desc = tt_SocDescriptor(soc_desc_path /*, harvesting_info*/);
+    uint32_t tensix_harvesting_mask = cluster_desc->get_harvesting_info().at(chip_id);
+    tt_SocDescriptor soc_desc = tt_SocDescriptor(soc_desc_path, tensix_harvesting_mask /*, harvesting_info*/);
```
tt-vjovanovic, line 508: Add this change:
```
        uint32_t tensix_harvesting_mask = cluster_desc->get_harvesting_info().at(chip_id);
        tt_SocDescriptor soc_desc = tt_SocDescriptor(sdesc_path, tensix_harvesting_mask);
```
to this constructor:
```
Cluster::Cluster(
    const std::string& sdesc_path,
    const std::set<chip_id_t>& target_devices,
    const uint32_t& num_host_mem_ch_per_mmio_device,
    const bool skip_driver_allocs,
    const bool clean_system_resources,
    bool perform_harvesting,
    std::unordered_map<chip_id_t, uint32_t> simulated_harvesting_masks)
```

device/cluster.cpp
pjanevskiTT: I can investigate this, not as part of this PR, this is just for API. Follow up issue https://github.com/tenstorrent/tt-umd/issues/427

tests/api/test_cluster.cpp
```diff
@@ -223,38 +196,80 @@ TEST(ApiClusterTest, SimpleIOSpecificChips) {
...
+                cluster->write_to_device(
+                    vector_to_write.data(),
+                    vector_to_write.size() * sizeof(std::uint32_t),
+                    chip,
+                    core,
+                    address,
+                    "SMALL_READ_WRITE_TLB");
+
+                // Barrier to ensure that all writes over ethernet were commited
+                cluster->wait_for_non_mmio_flush();
```
broskoTT, line 256: is l1_membar() call also necessary here in case of local chip?

broskoTT (approved):

Issue #44, opened 2024-09-04
TLB setup


pjanevskiTT:
## Problem

Currently we don't setup the TLBs inside UMD, just provide the API for compiler runtimes to be able to do it. It looks like we could move this logic to UMD, since we already have layout of TLBs and mapping of TLBs that all of compiler runtimes are following (Buda and Metal), which come mostly from syseng guide. It seems that there is no much room to map TLBs differently, since the layout we are using seems to be working fine and there is no clear more optimal mapping (for example 4GB TLBs on BH can't really be used for something else, 2MB TLBs are also pretty fixed)

It also leads to code duplicates like in tlb_config files in [Buda](https://yyz-gitlab.local.tenstorrent.com/tenstorrent/budabackend/-/blob/master/loader/tlb_config.cpp) and [Metal](https://github.com/tenstorrent/tt-metal/blob/main/tt_metal/llrt/tlb_config.cpp). If someone wants to write some simple tests using only UMD and some runtime code it also needs to setup TLBs manually outside of UMD, which can be a bit tiring.

## Proposed solution

Basically it would be nice to move the TLB setup inside UMD, something similar to `tlb_config.cpp` files in compilers. We can still do few things in order to keep flexibility:

- keep the API to setup the TLBs manually
- provide an option to not do the setup inside UMD 

For example we could have something like 

```

// functions inside some class like tt_SiliconDevice
void setup_tlbs();

void setup_hugepages();

void setup_device_communication() {
   
   setup_tlbs();

   setup_hugepages();
}
```

So we can allow runtime to call `setup_device_communication` and UMD is going to setup the TLBs (and everything else) or runtime can do it like in `tlb_config.cpp` through already [existing API](https://github.com/tenstorrent/tt-umd/blob/main/device/tt_silicon_driver.cpp#L2542) if it wants to do something different.

The proposed API is not 100% exact since we probably first need to make more high level plan of how UMD code is going to be organized, but I think the point is clear

joelsmithTT:
Thanks for raising this and proposing a solution.  Moving static TLB assignment into UMD could indeed reduce code duplication and reduce the burden of writing standalone tests.  However, I'd like to explore some additional considerations that might impact this approach.

1. How does static assignment address the challenge of chips having more addressable memory than what can be covered by the windows?  It seems dynamic assignment is necessary.
2. How can we account for scenarios where multiple applications need to access the device simultaneously?  For instance, gathering telemetry across a chip topology while an ML workload is active.
[...]

PR #403, opened 2024-12-16
Remove setup_core_to_tlb_map

broskoTT:
### Issue
Related to #44 but also on a path of #351 

### Description
This function in API is a surplus. All the info is already available to the Cluster class through configure_tlb calls.
This is initial PR which changes the API. The following PR won't change the API but will restructure functions internally to introduce TLBManager class.

### List of the changes
- Removed setup_core_to_tlb_map. Fill up the same internal structure through configure_tlb api
- tlbs_init_per_chip is also unnecessary
- Extracted the logic for checking whether static tlb is setup

### Testing
Covered through existing tests.

### API Changes
This PR has API changes:
- [x] tt_metal approved PR pointing to this branch: https://github.com/tenstorrent/tt-metal/pull/16048
- [X] tt_lens doesn't use this API call.

### DIFF
--- a/device/api/umd/device/cluster.h
+++ b/device/api/umd/device/cluster.h
@@ -84,18 +84,6 @@ public:
     }
 
     /**
-     * Give UMD a 1:1 function mapping a core to its appropriate static TLB (currently only support a single TLB per
-     * core).
-     *
-     * @param logical_device_id MMIO chip being targeted.
-     * @param mapping_function Function which maps core to TLB index.
-     */
-    virtual void setup_core_to_tlb_map(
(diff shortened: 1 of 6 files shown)


pjanevskiTT (approved):
Nice. One thing, which probably wasn't working even before your change, is calling configure_tlb twice for different cores and the same index, so some of our maps will have two cores pointing to the same index, which obviously is not correct. That is why there was a requirement that mapping function is 1:1 I guess. Per offline discussion, open an issue on this and maybe leave a comment somewhere to fix this. This problem will go away with some kind of TLB manager, but it would be nice just to be aware of this problem right now

PR #415, opened 2024-12-18
Reduce CI load

broskoTT:
### Issue
No issue

### Description
We are running same tests on both versions of ubuntu.
There is no need to think that tests would fall on one of the version and not the other, and we never saw this case. I still intentionally left building on ubuntu 22 though.
I chose ubuntu-20 since this is the same one as used in tt-metal.

### List of the changes
- Changed build-tests and run-tests to accept ubuntu version argument
- Added required ubuntu versions in build_and_run tests. Build on both 20 and 22, run tests only on 20.
- I'll also add these tests as required for PRs after this PR.

### Testing
CI tests on this PR

### API Changes

### DIFF
--- a/.github/workflows/build-and-run-all-tests.yml
+++ b/.github/workflows/build-and-run-all-tests.yml
@@ -19,9 +19,14 @@ jobs:
           {arch: wormhole_b0},
           {arch: blackhole},
         ]
+        ubuntu-version: [
+          'ubuntu-22.04',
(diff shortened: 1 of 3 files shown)


pjanevskiTT (approved):
Nice. I thought I saw the tests pass on 20 and fail on 22, but just searched some previous runs, those were some flaky tests, those would fail on 20 as well if run multiple times.

Issue #417, opened 2024-12-18
TLBManager class


broskoTT:
While working on Chip class #248 but also related to having internal default tlb setup #44 it seems natural to extract this logic into a separate class.
This will help in the future if we want to change this manager to communicate with the KMD implementation of TLB resource manager.

PR #419, opened 2024-12-18
init

broskoTT:
### Issue
Related to #417 

### Description
Seems like we've had a redundant get_tlb_configuration and describe_tlb functions. I also noted that sometimes we downcasted 64bit address to 32bit, so this should solve those problems if they were even present.

### List of the changes
- Removed describe_tlb everywhere
- Changed describe_tlb calls with get_tlb_configuration
- API now returns tlb_configuration type.

### Testing
Existing CI tests

### API Changes
This PR has API changes:
- [ ] tt_metal approved PR pointing to this branch: link
- [ ] tt_lens doesn't use this API call.

### DIFF
--- a/device/api/umd/device/cluster.h
+++ b/device/api/umd/device/cluster.h
@@ -562,7 +562,7 @@ public:
     /**
      * If the tlbs are initialized, returns a tuple with the TLB base address and its size
      */
-    std::optional<std::tuple<uint32_t, uint32_t>> get_tlb_data_from_target(const tt_cxy_pair& target);
+    tlb_configuration get_tlb_data_from_target(const tt_cxy_pair& target);
     /**
      * This API allows you to write directly to device memory that is addressable by a static TLB
      */
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -172,10 +172,9 @@ bool Cluster::is_tlb_mapped(tt_cxy_pair target, uint64_t address, uint32_t size_
     auto* dev = get_tt_device(target.chip);
 
     int32_t tlb_index = map_core_to_tlb_per_chip.at(target.chip).at(tt_xy_pair(target.x, target.y));
-    auto tlb_description = dev->get_architecture_implementation()->describe_tlb(tlb_index);
+    auto tlb_description = dev->get_architecture_implementation()->get_tlb_configuration(tlb_index);
 
-    return tlb_description.has_value() &&
-           address_in_tlb_space(address, size_in_bytes, tlb_index, std::get<1>(tlb_description.value()), target.chip);
+    return address_in_tlb_space(address, size_in_bytes, tlb_index, tlb_description.size, target.chip);
 }
 
 void Cluster::initialize_interprocess_mutexes(int pci_interface_id, bool cleanup_mutexes_in_shm) {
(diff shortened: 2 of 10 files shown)


device/api/umd/device/cluster.h
```diff
@@ -562,7 +562,8 @@ class Cluster : public tt_device {
     /**
      * If the tlbs are initialized, returns a tuple with the TLB base address and its size
      */
-    std::optional<std::tuple<uint32_t, uint32_t>> get_tlb_data_from_target(const tt_cxy_pair& target);
+    [[deprecated]] std::optional<std::tuple<uint32_t, uint32_t>> get_tlb_data_from_target(const tt_cxy_pair& target);
+    tlb_configuration get_tlb_configuration(const tt_cxy_pair& target);
```
pjanevskiTT, line 566: add docs for the new function

device/cluster.cpp
```diff
@@ -172,10 +172,9 @@ bool Cluster::is_tlb_mapped(tt_cxy_pair target, uint64_t address, uint32_t size_
     auto* dev = get_tt_device(target.chip);
 
     int32_t tlb_index = map_core_to_tlb_per_chip.at(target.chip).at(tt_xy_pair(target.x, target.y));
-    auto tlb_description = dev->get_architecture_implementation()->describe_tlb(tlb_index);
+    auto tlb_description = dev->get_architecture_implementation()->get_tlb_configuration(tlb_index);
```
pjanevskiTT, line 175: I think replacing auto with exact type here (tlb_configuration) would be more readable

device/cluster.cpp
```diff
@@ -1075,12 +1074,10 @@ tt::Writer Cluster::get_static_tlb_writer(tt_cxy_pair target) {
     }
 
     auto tlb_index = map_core_to_tlb_per_chip.at(target.chip).at(tt_xy_pair(target.x, target.y));
-    auto tlb_data = dev->get_architecture_implementation()->describe_tlb(tlb_index);
+    auto tlb_description = dev->get_architecture_implementation()->get_tlb_configuration(tlb_index);
```
pjanevskiTT, line 1077: same as above

device/cluster.cpp
```diff
@@ -1103,15 +1100,17 @@ void Cluster::write_device_memory(
         small_access);
 
     if (is_tlb_mapped(target, address, size_in_bytes)) {
-        auto tlb_description = dev->get_architecture_implementation()->describe_tlb(
+        auto tlb_description = dev->get_architecture_implementation()->get_tlb_configuration(
```
pjanevskiTT, line 1103: same as above

device/cluster.cpp
```diff
@@ -1151,17 +1150,23 @@ void Cluster::read_device_memory(
     log_debug(LogSiliconDriver, "  tlb_index: {}, tlb_data.has_value(): {}", tlb_index, tlb_data.has_value());
 
     if (is_tlb_mapped(target, address, size_in_bytes)) {
-        auto tlb_description = dev->get_architecture_implementation()->describe_tlb(
+        auto tlb_description = dev->get_architecture_implementation()->get_tlb_configuration(
```
pjanevskiTT, line 1153: same as above

device/cluster.cpp
```diff
@@ -1354,7 +1362,7 @@ void Cluster::configure_tlb(
 
     TTDevice* tt_device = get_tt_device(logical_device_id);
     tt_device->set_dynamic_tlb(tlb_index, core, address, harvested_coord_translation.at(logical_device_id), ordering);
-    auto tlb_size = std::get<1>(tt_device->get_architecture_implementation()->describe_tlb(tlb_index).value());
+    auto tlb_size = tt_device->get_architecture_implementation()->get_tlb_configuration(tlb_index).size;
```
pjanevskiTT, line 1365: same for auto, just here we probably need size type

pjanevskiTT (approved):
Nice changes. Left a few nit comments. In general, I think we should start writing tests for TLB configs, to be sure that we hit the address that we want or index if we care for the index. Not sure what is the right path here since maybe we don't want to expose BAR0 space, but we can think about it, maybe you can just open the follow up issue to have this on the radar

PR #420, opened 2024-12-18
TLB Manager

broskoTT:
### Issue
Related to #417 

### Description
Introduction into TLB manager class. This class should hold maps between tlb indexes and cores, and should hand out tlb_indicies. This might not be completelly true after this PR, which was more focused on getting things out from cluster.cpp file. But not too much stuff.

### List of the changes
- Created TLBManager class which lives inside TTDevice
- Added get_tlb_manager to TTDevice
- Moved dynamic_tlb_configs and dynamic_tlb_config_ordering maps.
- Moved configure_tlb main logic with maps that hold these mapped tlbs
- Moved all related functions from cluster.cpp
- These moved maps/functions are still used throughout cluster.cpp, but this will be refactored further, so that the tlbmanager itself checks and hands out the writer.

### Testing
Existing CI tests should cover this class.
Will try to add additional TLBManager focused tests.

### API Changes


pjanevskiTT (commented):
I like how the PR moves TLB logic from cluster downwards (Chip, TTDevice), but I think it is not in the right place in the layering. We can discuss this offline and get back to moving this if needed

device/cluster.cpp
```diff
@@ -1712,14 +1612,22 @@ void* Cluster::host_dma_address(std::uint64_t offset, chip_id_t src_device_id, u
...
     log_assert(chips_.find(device_id) != chips_.end(), "Device id {} not found in cluster.", device_id);
     auto tt_device = chips_.at(device_id)->get_tt_device();
     log_assert(tt_device != nullptr, "TTDevice not found for device: {}", device_id);
     return tt_device;
 }
 
+// Wrapper for throwing a more helpful exception when trying to access non pci enabled interface.
+inline TLBManager* Cluster::get_tlb_manager(chip_id_t device_id) const {
+    log_assert(chips_.find(device_id) != chips_.end(), "Device id {} not found in cluster.", device_id);
+    auto tlb_manager = chips_.at(device_id)->get_tlb_manager();
```
joelsmithTT, line 1626: Not sure if this will get optimized to a single lookup.  If it's in a perf path, consider,
```
auto it = chips_.find(device_id); // once
log_assert(...
auto* manager = it->second->get_tlb_manager()
```

device/chip/tlb_manager.h
```diff
@@ -0,0 +1,46 @@
+/*
+ * SPDX-FileCopyrightText: (c) 2024 Tenstorrent Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
```
joelsmithTT, line 5: Don't have to do it now, but it might be worth adding a bit of documentation about what is TLB, how they are used, why we have them etc.

device/chip/tlb_manager.h
```diff
@@ -0,0 +1,46 @@
...
+
+namespace tt::umd {
+
+class TTDevice;
+
+class TLBManager {
+public:
+    TLBManager(TTDevice* tt_device);
+
+    void configure_tlb(tt_xy_pair core, int32_t tlb_index, uint64_t address, uint64_t ordering);
```
joelsmithTT, line 23: How does the caller know what a valid tlb_index is?

device/chip/tlb_manager.h
```diff
@@ -0,0 +1,46 @@
...
+
+    void set_dynamic_tlb_config(std::string fallback_tlb_name, int32_t tlb_index);
+    void set_dynamic_tlb_config_ordering(std::string fallback_tlb_name, uint64_t ordering);
```
joelsmithTT, line 26: Valid fallback_tlb_names?

device/chip/tlb_manager.h
pjanevskiTT: We have some internal docs that we could convert to md files. I think it would be nice to do it in parallel with designing this TLB manager class. Not saying that Bojan has to do it, I have played around with TLBs as well earlier...https://github.com/tenstorrent/tt-umd/issues/429

device/chip/tlb_manager.h
broskoTT: A very good question, i have to think about it

device/api/umd/device/chip/local_chip.h
```diff
@@ -9,15 +9,20 @@
...
     LocalChip(tt_SocDescriptor soc_descriptor, int pci_device_id);
 
     TTDevice* get_tt_device() override;
+    TLBManager* get_tlb_manager() override;
 
     bool is_mmio_capable() const override;
 
 private:
     std::unique_ptr<TTDevice> tt_device_;
+    std::unique_ptr<TLBManager> tlb_manager_;
```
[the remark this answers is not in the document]
broskoTT: I think I see TTDevice as something which is just there just so you don't have to read documentation to know how to use our card. And TTDevice should be something which is valid for tooling potentially as well, not just AI workload.

Also, please don't look at TLBManagement more of a Chip thing rather than a TTDevice thing. If the client of the TTDevice wants to have some kind of tlb helper, the client would create a TLBManager (which uses core TTDevice functions underneath) to program TLBs, and that exactly is Chip. Chip is a client of TTDevice. TLBManager should not be seen as something which is part of Chip.

I think you're suggesting that TLBManagement should have what it needs for programming TLBS, and then actually TTDevice would call TLB management, and not the other way around?

In any case, let's try to answer some questions to pinpoint the design. Let's pretend TLBManager doesn't exist, and we want to decide between TTDevice and Chip API:
- Which of these options is the desired API for TTDevice:
  - TTDevice offering write, accepting core. No tlb_index, no fallback_tlb. There is a flag in TTDevice constructor which sets up the default map of static tlbs. There is also a call to set_dynamic_tlb which allows user to setup custom TLB, and it will be saved to internal TTDevice map.
  - TTDevice offering write to tlb index. Doesn't care if you set up the tlb properly or not.
- In the regular simple user example, with the default TLB setup, which class has the bool flag "setup_default_tlbs" in its constructor, and then the constructor does something complex?
- Which class is the owner of core_to_tlb map?
- How does the scenario look like for a power user, which doesn't use Chip class? Power user meaning they want to program their own TLBs. Maybe they have some custom workload, and don't need static TLBs for cores how we handle them by default, maybe they setup most of TLBs as multicasts.

I'm obviously more aligned with Chip owning TLBmanager. I don't see at the moment any other way to separate cleanly into two structures, lower one which just provides syntactic sugar for configuring the device, and doesn't provide any security or robustness, and a higher one which should not allow you to break the device in any way.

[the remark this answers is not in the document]
broskoTT: @pjanevskiTT and me discussed this offline.
I misunderstood the suggestion. The suggestion was only that TTDevice should own its TLBManager, rather than TLBManager having to be created on its own. I agree with this.
I misunderstood that TLBManager should be "underneath" TTDevice, so that TTDevice's API is in front of TLBManager.

device/chip/tlb_manager.cpp
```diff
@@ -0,0 +1,106 @@
...
+        LogSiliconDriver,
+        "Configuring TLB for chip: {} core: {} tlb_index: {} address: {} ordering: {}",
+        logical_device_id,
+        core.str(),
+        tlb_index,
+        address,
+        ordering);
+    log_assert(tlb_config_map_.find(tlb_index) == tlb_config_map_.end(), "TLB index already configured {}", tlb_index);
+
+    tt_device_->set_dynamic_tlb(tlb_index, core, address, ordering);
```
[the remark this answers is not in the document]
broskoTT: I'll leave set_dynamic_tlb for now in tt_device. I see that as a device specific (as opposed to workload specific), and I think having that API is useful in TTDevice.

device/api/umd/device/chip/chip.h
```diff
@@ -23,6 +24,7 @@ class Chip {
     tt_SocDescriptor& get_soc_descriptor();
 
     virtual TTDevice* get_tt_device();
+    virtual TLBManager* get_tlb_manager();
```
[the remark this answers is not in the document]
broskoTT: Removed

device/chip/tlb_manager.h
pjanevskiTT: It doesn't, that is one of the flaws of how UMD manages TLBs today. That is why we would like to transition to different API, something along the lines of

```
TLBWindow* get_tlb_wc();
TLBWindow* get_tlb_uc();
TLBWindow* get_any_tlb();
```

Note: TLBWindow is pointing to the struct that Joel suggested in some of the issues earlier.

then the client doesn't need to worry about index. We could however add a check baed on architecture implementation

broskoTT: Wrote a test and added a couple of utility functions to architecture_implementation.
Ultimatelly, the workload user (using chip and cluster) should not see tlb_index in their APIs, will work on removing that. tlb_index should be surfaced only in the lower levels.

Please take a look at new file test_tlb_manager.cpp

device/chip/tlb_manager.h
broskoTT: What is the comment? The function does check whether this is a valid fallback tlb.
Or is the comment around naming?
This should be converted to an enum obviously.

device/cluster.cpp
broskoTT: Applied this to get_tt_device. The get_tlb_manager function has changed

device/chip/tlb_manager.h
broskoTT: Captured in an issue. @pjanevskiTT please fill out the mentioned references in the issue.

tests/api/test_tlb_manager.cpp
```diff
@@ -0,0 +1,82 @@
...
+            tlb_index += tlb_2m_base_and_count.first;
+        }
+
+        return tlb_index;
+    };
+
+    std::int32_t c_zero_address = 0;
+
+    for (tt_xy_pair core : soc_desc.workers) {
+        tlb_manager->configure_tlb(core, get_static_tlb_index(core), c_zero_address, tlb_data::Relaxed);
```
pjanevskiTT, line 65: can you use the API introduced in #401. Let me know if this functions doesn't have the CoreCoord version

tests/api/test_tlb_manager.cpp
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    // So now that we have configured TLBs we can use it to interface with the TTDevice.
+    auto any_worker_core = soc_desc.workers[0];
```
pjanevskiTT, line 69: same as above, can you use 

`soc_desc.get_cores(CoreType::TENSIX)[0]`

pjanevskiTT (approved):
just a few API comments, changes look great now

device/chip/tlb_manager.h
broskoTT: Let me know if there's additional comment here.

PR #424, opened 2024-12-19
Removing LFS from workflows

warthog9:
Since we are removing LFS usage from the main setup, this should also include removing it from the workflows just in case it causes issues.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,0 @@
-device/libs/**/*.a filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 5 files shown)


joelsmithTT (approved):

joelsmithTT:
Probably https://github.com/tenstorrent/tt-umd/pull/423 should go in first?

warthog9:
yeah, or just merge this into your series

broskoTT (approved):

PR #436, opened 2024-12-25
[Chip] Move *_params to chip 

broskoTT:
### Issue
Related to #248 

### Description
Move all *_params structures to chip.
They will be currently moved to Chip, but after all the functions which use them are moved to Local/Remote chips, they will naturally fall in where they need to be, probably in LocalChip.

### List of the changes
- Moved all *_params to chip class
- Related changes in cluster.cpp
- Setting default parameters is now done in Chip class.

### Testing
CI, no additional testing.

### API Changes


pjanevskiTT (approved):
per offline discussion, these params can be set on per chip basis, so it should stay in chip

Issue #439, opened 2024-12-26
Remove Cluster' version of harvesting


broskoTT:
While working on #248
Harvesting should be removed from Cluster class, now that it exists in tt_socdescriptor. This old functionality should be carefully removed since it has a lot of code/structures, but all of it should be already reimplemented through coord_manager and tt_soc_descriptor.

Issue #435, opened 2024-12-23
Software harvesting


pjanevskiTT:
Implement support for software harvesting, meaning that clients can specify which rows should be "harvested". Use cases

- Test different harvested rows, to simulate what would happen on harvested chip
- Custom soc descriptors are going to use this as well

Questions

- Should harvesting masks be adjusted for NOC layout. Meaning each bit [0, N-1] would be for some of the N rows of tensix cores, dram banks, eth cores...

### Proposal

Have simple `SoftwareHarvesting` struct which should be comined with silicon harvesting masks

```cpp
struct SoftwareHarvesting {
    uint8_t tensix_harvesting_mask;
    uint8_t dram_harvesting_mask;
    uint8_t eth_harvesting_mask;
}
```

Then have constructors for cluster/chip/soc descriptor which take maps of these structures for each chip in cluster.

Something similar to `simulated_harvesting_masks` at the moment

pjanevskiTT:
Closed by #456

PR #440, opened 2024-12-26
[Harvesting] simulated_harvesting_masks passed to tt_SocDescriptor

broskoTT:
### Issue
Related to #439 

### Description
First of the changes preparing removal of harvesting logic inside Cluster.
perform_harvesting, and simulated_harvesting_masks arguments are now added to affect tt_SocDescriptor.
Had to remove logic converting harvesting_mask from physical layout coords to logical coords out of the CoordinateManager constructor.

### List of the changes
- perform_harvesting now affects harvesting in tt_SocDescriptor
- simulated_harvesting_masks are now used for harvesting in tt_SocDescriptor
- get_tensix_harvesting_mask is a helper function which does both of these
- shuffle_tensix_harvesting_mask is now a static public function, which should be called prior to calling CoordinateManager
- CoordinateManager now accepts harvesting in logical coords.
- Due to deprecated get_harvesting_masks_for_soc_descriptors, there is an additional shuffle_tensix_harvesting_mask_to_noc0_coords which does the required translation.
- tt_SocDescriptor removed copy constructor, since it is already a default one.
- Other functions changed accordingly to pass this arguments

### Testing
- Changed call to get_harvesting_masks_for_soc_descriptors, so it is tested with the new flow, through harvesting provided by socdescriptor class
- Added HarvestingShuffle tests in all arch tests

### API Changes

### DIFF
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -524,8 +524,6 @@ Cluster::Cluster(
         add_chip(
             chip_id,
             construct_chip_from_cluster(chip_id, cluster_desc.get(), perform_harvesting, simulated_harvesting_masks));
-        // log_info(LogSiliconDriver, "Added chip {} to cluster with harvesting {}.",
-        // chips_.at(chip_id)->get_soc_descriptor().tensix_harvesting_mask);
     }
 
     // TODO: work on removing this member altogether. Currently assumes all have the same arch.
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -501,6 +501,14 @@ uint32_t Cluster::get_tensix_harvesting_mask(
             chip_id,
             tensix_harvesting_mask);
     }
+    log_debug(
+        LogSiliconDriver,
+        "Harvesting mask for chip {} is {} (physical layout: {}, logical: {}, simulated harvesting mask: {}).",
+        chip_id,
+        tensix_harvesting_mask | simulated_harvesting_mask,
+        tensix_harvesting_mask_physical_layout,
+        tensix_harvesting_mask,
+        simulated_harvesting_mask);
     return tensix_harvesting_mask | simulated_harvesting_mask;
 }
 
@@ -516,6 +524,8 @@ Cluster::Cluster(
         add_chip(
             chip_id,
             construct_chip_from_cluster(chip_id, cluster_desc.get(), perform_harvesting, simulated_harvesting_masks));
+        // log_info(LogSiliconDriver, "Added chip {} to cluster with harvesting {}.",
+        // chips_.at(chip_id)->get_soc_descriptor().tensix_harvesting_mask);
     }
 
     // TODO: work on removing this member altogether. Currently assumes all have the same arch.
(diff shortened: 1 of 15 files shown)


device/cluster.cpp
```diff
@@ -478,6 +479,30 @@ void Cluster::add_chip(chip_id_t chip_id, std::unique_ptr<Chip> chip) {
...
+                                             : 0;
+    if (simulated_harvesting_mask != 0) {
+        log_info(
+            LogSiliconDriver,
+            "Adding simulated harvesting mask {} for chip {} which has real harvesting mask {}.",
+            simulated_harvesting_mask,
+            chip_id,
+            tensix_harvesting_mask);
+    }
+    return tensix_harvesting_mask | simulated_harvesting_mask;
```
pjanevskiTT, line 503: I think we can't do this in this way. `tensix_harvesting_mask` is going to be retrieved from cluster descriptor, which is going to be based on the actual physical layout of the chip, and `simulated_harvesting_mask` is going to be in either logical (from 0 to N-1) or NOC0 (bit b is for the row with y=b on NOC). Either way masking this way won't work. I think there are 2 solutions here
- Make sure that simulated harvesting mask is passed based on physical chip layout. This probably won't be the solution since it is difficult to use
- Shuffle tensix_harvesting_mask to be in logical/NOC0 layout (figure out what is the layout of simulated mask) and the do the logical OR. I had similar change in #369, maybe we can do this shuffling out of Coordinate manager and soc descriptor
Let me know what you think

device/cluster.cpp
```diff
@@ -478,6 +479,30 @@ void Cluster::add_chip(chip_id_t chip_id, std::unique_ptr<Chip> chip) {
...
+}
+
 Cluster::Cluster(
```
pjanevskiTT, line 506: nit: since we would more often like to do some simulated harvesting, #435, maybe it would be nice to move simulated_harvesting_masks to be before all other parameters. Just maybe, it came across my mind

device/cluster.cpp
```diff
@@ -452,15 +452,16 @@ std::unique_ptr<Chip> Cluster::construct_chip_from_cluster(
     }
 }
 
-std::unique_ptr<Chip> Cluster::construct_chip_from_cluster(chip_id_t chip_id, tt_ClusterDescriptor* cluster_desc) {
+std::unique_ptr<Chip> Cluster::construct_chip_from_cluster(
+    chip_id_t chip_id,
+    tt_ClusterDescriptor* cluster_desc,
+    bool perform_harvesting,
```
pjanevskiTT, line 458: I think it would just be nice to document what are the expected use cases for `perform_harvesting = false`. One I see is running custom soc descriptor on some different silicon, but there custom soc descriptor could hit harvested rows. Not sure whether that is something we should take care of. No need to do anything about it in this PR, but maybe open an issue so we document this later

device/cluster.cpp
broskoTT: Makes sense but I don't want to introduce API change only because of this.

device/cluster.cpp
broskoTT: Yeah I'm currently not sure of the expected use cases

device/cluster.cpp
broskoTT: I took this out of the class. The code is in CoordinateManager class, but it is a static function which is supposed to be called by the clients, currently called in Cluster. We discussed offline around this a bit. I wanted to avoid having two harvesting arguments in CoordinateManager which were used differently. This way. this api is simpler.

pjanevskiTT (approved):

Issue #343, opened 2024-11-27
[CoreCoord] Soc descriptor different coordinate systems


pjanevskiTT:
Good starting point for this feature https://github.com/tenstorrent/tt-umd/pull/266#discussion_r1833015987

We want SocDescriptor to be able to return any core type in any coordinate system. SocDescirptor would still be created from yaml/cpp code that has physical (noc) coordinates

Issue #356, opened 2024-11-30
Coding style


broskoTT:
Make a brief document on the coding style: 
- Namings of classes, functions, private members, constants, structs, local variables, etc...
- File structure
- Usage of namesspace
- Potentially some dos and donts

broskoTT:
+ whether header files are .h or .hpp

broskoTT:
+ How are arguments passed to a function. Whether it is a reference or pointer if it will be modified. Whether the primitive types are also passed as const references. 
+ reference google style guide

@blozano-tt @afuller-TT , just wondering if there is something already in place which answers these questions for tt_metal.

blozano-tt:
> * How are arguments passed to a function. Whether it is a reference or pointer if it will be modified. Whether the primitive types are also passed as const references.
> * reference google style guide
> 
> [@blozano-tt](https://github.com/blozano-tt) [@afuller-TT](https://github.com/afuller-TT) , just wondering if there is something already in place which answers these questions for tt_metal.

I'm not aware of anything that focuses on style, just some best practices.

Metal uses a mix of .h and .hpp for C++ header files.

There are some documents here that dictate some general best practices, like "don't use bool function arguments.

https://github.com/tenstorrent/tt-metal/tree/main/contributing

broskoTT:
- using `auto` keyword

blozano-tt:
> * using `auto` keyword

I love auto

broskoTT:
- mutable keyword (when changing some field in a const reference to an object.

blozano-tt:
> * mutable keyword (when changing some field in a const reference to an object.

This always felt like a hack to me. Is there a legit reason to do this?

broskoTT:
Just to be clear, the points I raise in this issue is just the points that I want to be addressed with coding style document, I'm still not suggesting anything yet.

For mutable: when you want a field to be modifiable in a const object. Yes, no legit reason to use it.
Actually, just googled it. One potentially legit usage is for internal mutex, so that you can call a .read() const method which need some synchronization.

When you create yourself a "const hell" by having a lot of consts wherever it was possible, this might be the only option sometimes.

afuller-TT:
> > * mutable keyword (when changing some field in a const reference to an object.
> 
> This always felt like a hack to me. Is there a legit reason to do this?

`std::mutex`.  A getter should be const, but the mutex must be non-const to guard against a setting joining the party.

[...]

PR #442, opened 2024-12-26
Soc descriptor API for all coordinate systems

pjanevskiTT:
### Issue

#343 

### Description

API to get cores in any coordinate systems

### List of the changes

- Add coordinate system as parameter for getting cores

### Testing

CI, added tests

### API Changes
/

### DIFF
--- a/device/api/umd/device/tt_soc_descriptor.h
+++ b/device/api/umd/device/tt_soc_descriptor.h
@@ -125,9 +125,9 @@ private:
     // is not needed anymore. Soc descriptor and coordinate manager should be
     // created once per chip.
     std::shared_ptr<CoordinateManager> coordinate_manager = nullptr;
-    mutable std::map<CoreType, std::vector<tt::umd::CoreCoord>> cores_map;
+    std::map<CoreType, std::vector<tt::umd::CoreCoord>> cores_map;
     std::map<CoreType, tt_xy_pair> grid_size_map;
-    mutable std::map<CoreType, std::vector<tt::umd::CoreCoord>> harvested_cores_map;
+    std::map<CoreType, std::vector<tt::umd::CoreCoord>> harvested_cores_map;
     std::map<CoreType, tt_xy_pair> harvested_grid_size_map;
 
     // DRAM cores are kept in additional vector struct since one DRAM bank
--- a/device/api/umd/device/tt_soc_descriptor.h
+++ b/device/api/umd/device/tt_soc_descriptor.h
@@ -60,8 +60,10 @@ public:
 
     static std::string get_soc_descriptor_path(tt::ARCH arch);
 
-    std::vector<tt::umd::CoreCoord> get_cores(const CoreType core_type) const;
-    std::vector<tt::umd::CoreCoord> get_harvested_cores(const CoreType core_type) const;
+    std::vector<tt::umd::CoreCoord> get_cores(
+        const CoreType core_type, const CoordSystem coord_system = CoordSystem::PHYSICAL) const;
+    std::vector<tt::umd::CoreCoord> get_harvested_cores(
+        const CoreType core_type, const CoordSystem coord_system = CoordSystem::PHYSICAL) const;
     tt_xy_pair get_grid_size(const CoreType core_type) const;
     tt_xy_pair get_harvested_grid_size(const CoreType core_type) const;
 
@@ -116,14 +118,16 @@ private:
     void get_cores_and_grid_size_from_coordinate_manager();
 
     static tt_xy_pair calculate_grid_size(const std::vector<tt_xy_pair> &cores);
+    std::vector<tt::umd::CoreCoord> translate_coordinates(
+        const std::vector<tt::umd::CoreCoord> &physical_cores, const CoordSystem coord_system) const;
 
     // TODO: change this to unique pointer as soon as copying of tt_SocDescriptor
     // is not needed anymore. Soc descriptor and coordinate manager should be
     // created once per chip.
     std::shared_ptr<CoordinateManager> coordinate_manager = nullptr;
-    std::map<CoreType, std::vector<tt::umd::CoreCoord>> cores_map;
+    mutable std::map<CoreType, std::vector<tt::umd::CoreCoord>> cores_map;
     std::map<CoreType, tt_xy_pair> grid_size_map;
-    std::map<CoreType, std::vector<tt::umd::CoreCoord>> harvested_cores_map;
+    mutable std::map<CoreType, std::vector<tt::umd::CoreCoord>> harvested_cores_map;
     std::map<CoreType, tt_xy_pair> harvested_grid_size_map;
 
     // DRAM cores are kept in additional vector struct since one DRAM bank
--- a/device/tt_soc_descriptor.cpp
+++ b/device/tt_soc_descriptor.cpp
@@ -14,6 +14,7 @@
 
 #include "api/umd/device/tt_soc_descriptor.h"
 #include "fmt/core.h"
+#include "logger.hpp"
 #include "utils.hpp"
 #include "yaml-cpp/yaml.h"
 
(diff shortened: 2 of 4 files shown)


device/tt_soc_descriptor.cpp
```diff
@@ -336,18 +348,33 @@ void tt_SocDescriptor::get_cores_and_grid_size_from_coordinate_manager() {
...
 }
 
-std::vector<tt::umd::CoreCoord> tt_SocDescriptor::get_harvested_cores(const CoreType core_type) const {
-    if (harvested_cores_map.find(core_type) == harvested_cores_map.end()) {
-        return {};
+std::vector<tt::umd::CoreCoord> tt_SocDescriptor::get_harvested_cores(
+    const CoreType core_type, const CoordSystem coord_system) const {
+    log_assert(coord_system != CoordSystem::LOGICAL, "Harvested cores are not supported for logical coordinates");
+    if (harvested_cores_map.find({core_type, coord_system}) == harvested_cores_map.end()) {
+        translate_coordinates(harvested_cores_map, core_type, coord_system);
```
tt-vjovanovic, line 375: Note that this call is not thread safe... Lazy cache needs to be locked (both read and write) for thread safety....

device/tt_soc_descriptor.cpp
```diff
@@ -336,18 +348,33 @@ void tt_SocDescriptor::get_cores_and_grid_size_from_coordinate_manager() {
...
+    std::vector<CoreCoord> translated_cores;
+    for (const auto &core : cores) {
+        translated_cores.push_back(translate_coord_to(core, coord_system));
+    }
+    core_coord_map.insert({{core_type, coord_system}, translated_cores});
+}
+
+std::vector<tt::umd::CoreCoord> tt_SocDescriptor::get_cores(
+    const CoreType core_type, const CoordSystem coord_system) const {
+    if (cores_map.find({core_type, coord_system}) == cores_map.end()) {
```
tt-vjovanovic, line 365: You are always doing to std::map queries + potential write. You can reduce if you store result of `find()` and if it is not `end()`, return its value.

device/api/umd/device/tt_soc_descriptor.h
```diff
@@ -149,14 +151,18 @@ class tt_SocDescriptor {
...
+        std::map<std::pair<CoreType, CoordSystem>, std::vector<tt::umd::CoreCoord>> &core_coord_map,
+        const CoreType core_type,
+        const CoordSystem coord_system) const;
 
     // TODO: change this to unique pointer as soon as copying of tt_SocDescriptor
     // is not needed anymore. Soc descriptor and coordinate manager should be
     // created once per chip.
     std::shared_ptr<CoordinateManager> coordinate_manager = nullptr;
-    std::map<CoreType, std::vector<tt::umd::CoreCoord>> cores_map;
+    mutable std::map<std::pair<CoreType, CoordSystem>, std::vector<tt::umd::CoreCoord>> cores_map;
```
tt-vjovanovic, line 163: Do we actually need complication with lazy buffer and "double" buffering in soc descriptor, since we have translation map in coordinate_manager already?

device/tt_soc_descriptor.cpp
pjanevskiTT: Fixed, should be doing minimal number of queries now

device/api/umd/device/tt_soc_descriptor.h
pjanevskiTT: Removed lazy buffering

device/tt_soc_descriptor.cpp
pjanevskiTT: Lazy buffering is removed, now everything is read only

device/api/umd/device/tt_soc_descriptor.h
```diff
@@ -149,14 +151,16 @@ class tt_SocDescriptor {
...
     static tt_xy_pair calculate_grid_size(const std::vector<tt_xy_pair> &cores);
+    std::vector<tt::umd::CoreCoord> translate_coordinates(
+        const std::vector<tt::umd::CoreCoord> &physical_cores, const CoordSystem coord_system) const;
 
     // TODO: change this to unique pointer as soon as copying of tt_SocDescriptor
     // is not needed anymore. Soc descriptor and coordinate manager should be
     // created once per chip.
     std::shared_ptr<CoordinateManager> coordinate_manager = nullptr;
-    std::map<CoreType, std::vector<tt::umd::CoreCoord>> cores_map;
+    mutable std::map<CoreType, std::vector<tt::umd::CoreCoord>> cores_map;
```
broskoTT, line 161: You added this because this is being used in some const function somewhere? We should ban this modifier.
If it is too much work to do the necessary changes, feel free to leave it for some later PR. But we should avoid this keyword.

broskoTT: Added topic to https://github.com/tenstorrent/tt-umd/issues/356

tests/api/test_soc_descriptor.cpp
```diff
@@ -342,3 +342,118 @@ TEST(SocDescriptor, CustomSocDescriptor) {
 
     EXPECT_EQ(soc_desc.get_num_dram_channels(), 1);
 }
+
+bool core_coord_vectors_equal(const std::vector<CoreCoord>& vec1, const std::vector<CoreCoord>& vec2) {
```
broskoTT, line 346: Can `operator=` be used?

tests/api/test_soc_descriptor.cpp
```diff
@@ -342,3 +342,118 @@ TEST(SocDescriptor, CustomSocDescriptor) {
...
+    tt_SocDescriptor soc_desc(test_utils::GetAbsPath("tests/soc_descs/grayskull_10x12.yaml"));
+
+    std::vector<CoreCoord> cores_physical = soc_desc.get_cores(CoreType::TENSIX, CoordSystem::PHYSICAL);
+
+    std::vector<CoreCoord> virtual_from_physical;
+    std::vector<CoreCoord> logical_from_physical;
+    std::vector<CoreCoord> translated_from_physical;
+
+    for (const CoreCoord& core : cores_physical) {
+        virtual_from_physical.push_back(soc_desc.translate_coord_to(core, CoordSystem::VIRTUAL));
```
broskoTT, line 369: You've essentially reimplemented the get_cores completely here.

Maybe check some features of results of `get_cores` some other way, like the expected size, the CoordSystem of the returned value. Or use specific (hardcoded) expected values.

broskoTT (approved):
Nice

device/tt_soc_descriptor.cpp
```diff
@@ -314,6 +315,12 @@ void tt_SocDescriptor::get_cores_and_grid_size_from_coordinate_manager() {
     harvested_cores_map.insert({CoreType::ETH, coordinate_manager->get_harvested_cores(CoreType::ETH)});
     harvested_grid_size_map.insert({CoreType::ETH, coordinate_manager->get_harvested_grid_size(CoreType::ETH)});
 
+    harvested_cores_map.insert({CoreType::ARC, coordinate_manager->get_harvested_cores(CoreType::ARC)});
```
broskoTT, line 318: So for our physical coord system, if we look at core which is a harvested tensix, will it have CoreType::TENSIX or CoreType::HARVESTED? What is ::HARVESTED used for exactly?

device/api/umd/device/tt_soc_descriptor.h
pjanevskiTT: It was used because we were updating it in in const function, forgot to remove it

device/tt_soc_descriptor.cpp
pjanevskiTT: It will have TENSIX, HARVESTED should become deprecated, didn't have time to remove it, but it is not used in anything CoordinateManager related

PR #443, opened 2024-12-26
[Harvesting] noc_translation_enabled added to tt_SocDescriptor

broskoTT:
### Issue
Related to #439 

### Description
The last missing feature of Coordinate Manager is noc_translation enabled feature. If this is disabled, translated coords will be same as physical coords.

### List of the changes
- Add bool noc_translation_enabled to CoordinateManager constructor
- Added default versions of translate-physical mapping, which do identity mapping.
- Added a test to test if this flag works
- The translation flag is now required, so added it explicitly to tests where needed.

### Testing
Added another test, existing CI tests.

### API Changes

### DIFF
--- a/device/api/umd/device/coordinate_manager.h
+++ b/device/api/umd/device/coordinate_manager.h
@@ -188,10 +188,10 @@ protected:
     std::map<tt::umd::CoreCoord, tt_xy_pair> to_physical_map;
     std::map<std::pair<tt_xy_pair, CoordSystem>, tt::umd::CoreCoord> from_physical_map;
 
-    // Whether NOC translation is enabled.
-    // Effectivelly, when this is enabled, we can interface NOC with Virtual coordinates (on wormhole, translated by
-    // some constant). If not, Physical should be used. This will be implemented through Translated coordinates, so the
-    // user can always use those to interface with NOC.
+    // Whether NOC translation is enabled on chip.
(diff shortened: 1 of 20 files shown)


device/api/umd/device/coordinate_manager.h
```diff
@@ -149,19 +162,25 @@ class CoordinateManager {
...
      * should override this method.
      */
-    virtual void fill_arc_physical_translated_mapping();
+    virtual void fill_arc_physical_translated_mapping() = 0;
 
     std::map<tt::umd::CoreCoord, tt_xy_pair> to_physical_map;
     std::map<std::pair<tt_xy_pair, CoordSystem>, tt::umd::CoreCoord> from_physical_map;
 
+    // Whether NOC translation is enabled.
+    // Effectivelly, when this is enabled, we can interface NOC with Virtual coordinates (on wormhole, translated by
```
pjanevskiTT, line 179: I would avoid mentioning virtual coordinates, since that is something that we came up with, lets say. We shouldn't use those coordinates for NOC endpoints, we should use either Physical (NOC0) or translated. Virtual are on API level at the moment

pjanevskiTT (approved):
nice, like the changes

device/api/umd/device/coordinate_manager.h
broskoTT: But Virtual is used all around coordinate_manager?

PR #445, opened 2024-12-29
[Harvesting] Changed harvested_coord_translation to CoordinateManager translation

broskoTT:
### Issue
Related to #439 

### Description
Almost the last step which allows harvesting code to be removed from cluster

### List of the changes
Remove harvested_coord_translation in favour of core coordinates api.

### Testing
Existing tests

### API Changes

### DIFF
--- a/device/api/umd/device/cluster.h
+++ b/device/api/umd/device/cluster.h
@@ -1195,7 +1195,7 @@ private:
         std::unordered_map<chip_id_t, HarvestingMasks> simulated_harvesting_masks);
     // Helper functions for translating chip coordinates.
     tt_xy_pair translate_to_api_coords(const chip_id_t chip, const tt::umd::CoreCoord core_coord) const;
-    // Most of the old APIs accept virtual coordinates, but we communicate with the device through translate
+    // Most of the old APIs accept virtual coordinates, but we communicate with the device through translated
     // coordinates. This is an internal helper function, until we switch the API to accept translated coordinates.
     tt_xy_pair translate_chip_coord_virtual_to_translated(const chip_id_t chip_id, const tt_xy_pair core) const;
 
--- a/device/api/umd/device/cluster.h
+++ b/device/api/umd/device/cluster.h
@@ -1193,8 +1193,11 @@ private:
         const bool clean_system_resources,
         bool perform_harvesting,
         std::unordered_map<chip_id_t, HarvestingMasks> simulated_harvesting_masks);
-    // Helper function for translating chip coordinates.
+    // Helper functions for translating chip coordinates.
     tt_xy_pair translate_to_api_coords(const chip_id_t chip, const tt::umd::CoreCoord core_coord) const;
+    // Most of the old APIs accept virtual coordinates, but we communicate with the device through translate
+    // coordinates. This is an internal helper function, until we switch the API to accept translated coordinates.
+    tt_xy_pair translate_chip_coord_virtual_to_translated(const chip_id_t chip_id, const tt_xy_pair core) const;
 
     // State variables
     std::vector<tt::ARCH> archs_in_cluster = {};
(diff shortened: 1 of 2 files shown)


device/cluster.cpp
```diff
@@ -971,10 +972,12 @@ void Cluster::broadcast_pcie_tensix_risc_reset(chip_id_t chip_id, const TensixSo
     auto [soft_reset_reg, _] = tt_device->set_dynamic_tlb_broadcast(
         architecture_implementation->get_reg_tlb(),
         architecture_implementation->get_tensix_soft_reset_addr(),
-        harvested_coord_translation.at(chip_id).at(tt_xy_pair(0, 0)),
-        harvested_coord_translation.at(chip_id).at(tt_xy_pair(
-            architecture_implementation->get_grid_size_x() - 1,
-            architecture_implementation->get_grid_size_y() - 1 - num_rows_harvested.at(chip_id))),
+        translate_chip_coord_virtual_to_translated(chip_id, tt_xy_pair(0, 0)),
```
pjanevskiTT, line 975: I think this field in TLB config is not used unless we are doing multicast, so there is no point in translating coord 0,0. Even worse, I think CoordinateManager may throw if that coordinate doesn't represent any meaningful location. I think it would be better to just leave `tt_xy_pair(0,0)` or even better remove the need to specify that field of the config unless we are using multicast

device/cluster.cpp
```diff
@@ -1440,7 +1444,7 @@ int Cluster::test_setup_interface() {
         uint32_t mapped_reg = tt_device
                                   ->set_dynamic_tlb(
                                       tt_device->get_architecture_implementation()->get_reg_tlb(),
-                                      harvested_coord_translation.at(chip_id).at(tt_xy_pair(0, 0)),
+                                      translate_chip_coord_virtual_to_translated(chip_id, tt_xy_pair(0, 0)),
```
pjanevskiTT, line 1447: same as above

pjanevskiTT (commented):
Let me know if the comment below makes sense, apart from that the changes look good

device/cluster.cpp
broskoTT: This function is called "broadcast_pcie_tensix_risc_reset", this is indeed a multicast.

broskoTT: I'll make a separate PR for this.

device/cluster.cpp
broskoTT: This is exactly what coordinates were used previously for test. I'll leave it at that.

device/api/umd/device/cluster.h
```diff
@@ -1193,8 +1193,11 @@ class Cluster : public tt_device {
         const bool clean_system_resources,
         bool perform_harvesting,
         std::unordered_map<chip_id_t, HarvestingMasks> simulated_harvesting_masks);
-    // Helper function for translating chip coordinates.
+    // Helper functions for translating chip coordinates.
     tt_xy_pair translate_to_api_coords(const chip_id_t chip, const tt::umd::CoreCoord core_coord) const;
+    // Most of the old APIs accept virtual coordinates, but we communicate with the device through translate
```
pjanevskiTT, line 1198: nit: typo `translate` instead of `translated`

device/cluster.cpp
```diff
@@ -971,7 +971,8 @@ std::unordered_map<tt_xy_pair, tt_xy_pair> Cluster::create_harvested_coord_trans
 }
 
 void Cluster::translate_to_noc_table_coords(chip_id_t device_id, std::size_t& r, std::size_t& c) {
```
pjanevskiTT, line 973: does this function accept r and c as virtual coordinates?

pjanevskiTT (approved):

device/cluster.cpp
broskoTT: Yes. Will go away pretty soon, don't worry.

PR #451, opened 2025-01-03
[Harvesting] Move tt_SocDescriptor accessors to new API inside UMD

broskoTT:
### Issue
PR towards #439 

### Description
There are currently workers inside tt_socDescriptor, and then another version inside its coordinate manager. Same for other types of cores. 
This PR moves as much as it can to the new API. The final touch will be done when finally removing completely the cluster's version of harvesting.

### List of the changes
- Made all tt_socDescriptor intenal structures private (Reverted)
- Moved all code which used those to the new get_cores or get_harvested_cores api.

### Testing
Existing CI test should be fine.

### API Changes
I've left the accessors public, even though I've removed the usage at most places in UMD repo. This will allow us to remove this subsequently in tt_metal, and only then move it to private here.

### DIFF
--- a/device/api/umd/device/tt_soc_descriptor.h
+++ b/device/api/umd/device/tt_soc_descriptor.h
@@ -84,21 +84,11 @@ public:
     bool is_ethernet_core(const tt_xy_pair &core) const;
 
     tt::ARCH arch;
-    tt_xy_pair grid_size;
-    tt_xy_pair worker_grid_size;
     std::unordered_map<tt_xy_pair, CoreDescriptor> cores;
-    std::vector<tt_xy_pair> arc_cores;
-    std::vector<tt_xy_pair> workers;
-    std::vector<tt_xy_pair> harvested_workers;
-    std::vector<tt_xy_pair> pcie_cores;
     std::unordered_map<int, int> worker_log_to_routing_x;
     std::unordered_map<int, int> worker_log_to_routing_y;
     std::unordered_map<int, int> routing_x_to_worker_x;
     std::unordered_map<int, int> routing_y_to_worker_y;
-    std::vector<std::vector<tt_xy_pair>> dram_cores;                             // per channel list of dram cores
-    std::unordered_map<tt_xy_pair, std::tuple<int, int>> dram_core_channel_map;  // map dram core to chan/subchan
-    std::vector<tt_xy_pair> ethernet_cores;                                      // ethernet cores (index == channel id)
-    std::unordered_map<tt_xy_pair, int> ethernet_core_channel_map;
     std::vector<std::size_t> trisc_sizes;  // Most of software stack assumes same trisc size for whole chip..
     std::string device_descriptor_file_path = std::string("");
 
@@ -111,6 +101,11 @@ public:
     bool noc_translation_id_enabled;
     uint64_t dram_bank_size;
     uint32_t tensix_harvesting_mask;
+
+    // TODO: Move to private when remove_worker_row_from_descriptor is removed.
+    tt_xy_pair worker_grid_size;
+    std::vector<tt_xy_pair> workers;
+    std::vector<tt_xy_pair> harvested_workers;
 
 private:
     void create_coordinate_manager(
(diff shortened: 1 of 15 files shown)


device/cluster.cpp
```diff
@@ -652,9 +654,12 @@ void Cluster::populate_cores() {
...
         auto& soc_desc = chip->get_soc_descriptor();
         workers_per_chip.insert(
-            {chip_id, std::unordered_set<tt_xy_pair>(soc_desc.workers.begin(), soc_desc.workers.end())});
+            {chip_id,
+             std::unordered_set<tt_xy_pair>(
+                 soc_desc.get_cores(CoreType::TENSIX).begin(), soc_desc.get_cores(CoreType::TENSIX).end())});
         if (count == 0) {
-            eth_cores = std::unordered_set<tt_xy_pair>(soc_desc.ethernet_cores.begin(), soc_desc.ethernet_cores.end());
+            eth_cores = std::unordered_set<tt_xy_pair>(
+                soc_desc.get_cores(CoreType::ETH).begin(), soc_desc.get_cores(CoreType::ETH).end());
```
pjanevskiTT, line 662: nit: calling get_cores twice here, maybe call it just once before and store it in a struct. However, this is not on critical perf path

device/cluster.cpp
```diff
@@ -720,10 +726,11 @@ void Cluster::remove_worker_row_from_descriptor(
 }
 
 void Cluster::harvest_rows_in_soc_descriptor(tt::ARCH arch, tt_SocDescriptor& sdesc, uint32_t harvested_rows) {
-    std::uint32_t max_row_to_remove =
-        (*std::max_element((sdesc.workers).begin(), (sdesc.workers).end(), [](const auto& a, const auto& b) {
-            return a.y < b.y;
-        })).y;
+    std::uint32_t max_row_to_remove = (*std::max_element(
+                                           (sdesc.get_cores(CoreType::TENSIX)).begin(),
```
pjanevskiTT, line 730: same as above

device/cluster.cpp
```diff
@@ -944,12 +951,13 @@ void Cluster::deassert_risc_reset_at_core(tt_cxy_pair core, const TensixSoftRese
...
-            get_soc_descriptor(target_device).workers.begin(), get_soc_descriptor(target_device).workers.end(), core) !=
-                get_soc_descriptor(target_device).workers.end() ||
+            get_soc_descriptor(target_device).get_cores(CoreType::TENSIX).begin(),
+            get_soc_descriptor(target_device).get_cores(CoreType::TENSIX).end(),
+            core) != get_soc_descriptor(target_device).get_cores(CoreType::TENSIX).end() ||
             std::find(
-                get_soc_descriptor(target_device).ethernet_cores.begin(),
-                get_soc_descriptor(target_device).ethernet_cores.end(),
-                core) != get_soc_descriptor(target_device).ethernet_cores.end(),
+                get_soc_descriptor(target_device).get_cores(CoreType::ETH).begin(),
```
pjanevskiTT, line 958: same as above

device/api/umd/device/tt_soc_descriptor.h
```diff
@@ -163,6 +124,16 @@ class tt_SocDescriptor {
     // has multiple NOC endpoints, so some UMD clients prefer vector of vectors returned.
     std::vector<std::vector<tt::umd::CoreCoord>> dram_cores_core_coord;
     std::vector<std::vector<tt::umd::CoreCoord>> harvested_dram_cores_core_coord;
+
+    // Internal structures filled from soc descriptor yaml
+    tt_xy_pair grid_size;
+    std::vector<tt_xy_pair> arc_cores;
```
pjanevskiTT, line 130: why do we need to keep this structures, can't we just get it through `get_cores` API

broskoTT: These are now moved to being internal, they aren't public. Public users would indeed use .get_cores.
If you're talking about internal usage, we can use .get_cores, but not until we initialize coord manager.
load_core_descriptors_from_device_descriptor fills these structures, which are then used in other functions.

I could reformat code and functions, so that everything is done in the same function with local variables, or that these variables are returned and passed through some functions instead being "implicitly" passed as private members, but I'd rather not.

Let me know what you think.

pjanevskiTT: I am totally ok with keeping this structures for now if some code depends on it, the questions was more whether get_cores could completely replace these structures. If it can't or if it's too big of an effort it is ok to keep it as it is

pjanevskiTT (approved):

broskoTT:
Done in other PRs

Issue #94, opened 2024-10-01
[Blackhole] ARC messages and telemetry


pjanevskiTT:
ARC is not enabled on Blackhole at the moment. When we were bringing up new Blackhole features in UMD, there was a system limitation for us to enable ARC

Now everything is set up and we should implement ARC support for Blackhole

abhullar-tt:
Just wanted to point out that this will likely be required to extract harvesting info (fw for this expected post holidays)

pjanevskiTT:
Closed by #493

PR #452, opened 2025-01-08
Implement BH ARC message queue

pjanevskiTT:
### Issue

#94 - shouldn't close it, second part of the issue is getting harvesting info

### Description

Impement ARC messages for Blachkole. The design has changed from Wormhole, we now have circular queues for messages. Integrating this into Cluster API will happen in some of the following PRs, didn't want to mix this with this PR.

### List of the changes

- Implement Blackhole ARC message queue
- Add test for sending ARC message

### Testing

Simple test to send 100 TEST messages

### API Changes
/

### DIFF
--- a/device/api/umd/device/blackhole_arc_message_queue.h
+++ b/device/api/umd/device/blackhole_arc_message_queue.h
@@ -51,8 +51,6 @@ private:
 
     void write_words(uint32_t* data, size_t num_words, size_t offset);
 
-    void create_request(uint32_t* request, ArcMessageType message_type, uint32_t* data, size_t num_words);
-
     void trigger_fw_int();
 
     const uint64_t base_address;
--- a/device/api/umd/device/blackhole_arc_message_queue.h
+++ b/device/api/umd/device/blackhole_arc_message_queue.h
@@ -30,6 +30,8 @@ class Cluster;
  *
  * The read and write pointers are double-wrapping, meaning that they wrap at twice queue size. The number of occupied
  * entries in a queue is (wptr – rptr) % (2*size).
+ *
+ * Usage of this class is not thread-safe, this synchronization is going to be implemented, similar to Wormhole.
  */
 class BlackholeArcMessageQueue {
 private:
@@ -55,7 +57,7 @@ public:
      */
     uint32_t send_message(const ArcMessageType message_type, uint16_t arg0 = 0, uint16_t arg1 = 0);
 
-    static std::shared_ptr<BlackholeArcMessageQueue> get_blackhole_arc_message_queue(
+    static std::unique_ptr<BlackholeArcMessageQueue> get_blackhole_arc_message_queue(
         Cluster* cluster, const chip_id_t chip, const size_t queue_index);
 
 private:
--- a/device/api/umd/device/blackhole_arc_message_queue.h
+++ b/device/api/umd/device/blackhole_arc_message_queue.h
@@ -16,6 +16,21 @@ namespace tt::umd {
 
 class Cluster;
 
+/* On Blackhole there are few ARC message queues that can be used to communicate with ARC FW.
+ * ARC message queues are simple circular queues. There are read/write pointers both for requests and responses.
+ * Reading from SCRATCH_RAM[11] gives the address of the ARC message queuse descriptor.
+ *
+ * | Purpose                  | Start bit offset | Length in bits |
+ * ----------------------------------------------------------------
+ * ARC address of first queue |                0 |             32 |
+ * # of entries in each queue |               32 |              8 |
+ * # of queues                |               40 |              8 |
+ *
+ * Each queue starts with a header followed by the request queue then the response queue.
+ *
+ * The read and write pointers are double-wrapping, meaning that they wrap at twice queue size. The number of occupied
+ * entries in a queue is (wptr – rptr) % (2*size).
+ */
 class BlackholeArcMessageQueue {
 private:
     // Header length and entry length in words.
@@ -35,6 +50,9 @@ public:
         const uint64_t size,
         const CoreCoord arc_core);
 
+    /*
+     * Send ARC message. The call of send_message is blocking, timeout is to be implemented.
+     */
     uint32_t send_message(const ArcMessageType message_type, uint16_t arg0 = 0, uint16_t arg1 = 0);
 
     static std::shared_ptr<BlackholeArcMessageQueue> get_blackhole_arc_message_queue(
--- a/device/api/umd/device/blackhole_arc_message_queue.h
+++ b/device/api/umd/device/blackhole_arc_message_queue.h
@@ -1,0 +1,65 @@
+/*
+ * SPDX-FileCopyrightText: (c) 2024 Tenstorrent Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+
+#include "umd/device/blackhole_implementation.h"
+#include "umd/device/tt_core_coordinates.h"
+#include "umd/device/types/blackhole_arc.h"
+#include "umd/device/types/cluster_descriptor_types.h"
+
+using namespace tt::umd::blackhole;
+
+namespace tt::umd {
+
+class Cluster;
+
+class BlackholeArcMessageQueue {
+private:
+    // Header length and entry length in words.
+    static constexpr uint8_t header_len = 8;
+    static constexpr uint8_t entry_len = 8;
+
+    static constexpr uint8_t request_wptr_offset = 0;
+    static constexpr uint8_t response_rptr_offset = 1;
+    static constexpr uint8_t request_rptr_offset = 4;
+    static constexpr uint8_t response_wptr_offset = 5;
+
+public:
+    BlackholeArcMessageQueue(
+        Cluster* cluster,
+        const chip_id_t chip,
+        const uint64_t base_address,
+        const uint64_t size,
+        const CoreCoord arc_core);
+
+    uint32_t send_message(const ArcMessageType message_type, uint16_t arg0 = 0, uint16_t arg1 = 0);
+
+    static std::shared_ptr<BlackholeArcMessageQueue> get_blackhole_arc_message_queue(
+        Cluster* cluster, const chip_id_t chip, const size_t queue_index);
+
+private:
+    void push_request(std::array<uint32_t, BlackholeArcMessageQueue::entry_len>& request);
+
+    std::array<uint32_t, entry_len> pop_response();
+
+    void read_words(uint32_t* data, size_t num_words, size_t offset);
+
+    uint32_t read_word(size_t offset);
+
+    void write_words(uint32_t* data, size_t num_words, size_t offset);
+
+    void create_request(uint32_t* request, ArcMessageType message_type, uint32_t* data, size_t num_words);
+
+    void trigger_fw_int();
+
+    const uint64_t base_address;
+    const uint64_t size;
+    Cluster* cluster;
+    const chip_id_t chip;
+    const CoreCoord arc_core;
+};
+
+}  // namespace tt::umd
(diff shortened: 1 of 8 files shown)


device/api/umd/device/blackhole_arc_message_queue.h
```diff
@@ -0,0 +1,65 @@
...
+#include "umd/device/types/blackhole_arc.h"
+#include "umd/device/types/cluster_descriptor_types.h"
+
+using namespace tt::umd::blackhole;
+
+namespace tt::umd {
+
+class Cluster;
+
+class BlackholeArcMessageQueue {
```
broskoTT, line 19: Can you reference in the code comment some spec that this code is following?

device/api/umd/device/blackhole_implementation.h
```diff
@@ -188,6 +188,18 @@ static constexpr uint32_t MSG_TYPE_SETUP_IATU_FOR_PEER_TO_PEER = 0x97;
...
+// Register from which address of the ARC queue control block is read.
+constexpr uint64_t SCRATCH_RAM_11 = 0x8003042C;
+
+// ARC message queue header and entry size in bytes.
+constexpr uint32_t ARC_MSG_QUEUE_HEADER_SIZE = 32;
+constexpr uint32_t ARC_QUEUE_ENTRY_SIZE = 32;
+
+// ARC firmware interrupt address and value to write in order
+// to make an interrupt request.
+constexpr uint32_t ARC_FW_INT_ADDR = 2147680512;
```
broskoTT, line 200: Would these two look better in hex?

device/api/umd/device/cluster.h
```diff
@@ -18,6 +18,7 @@
 #include "tt_silicon_driver_common.hpp"
 #include "tt_soc_descriptor.h"
 #include "tt_xy_pair.h"
+#include "umd/device/blackhole_arc_message_queue.h"
```
broskoTT, line 21: This is the only line changed in this file. 
Is this include really needed here?

device/api/umd/device/types/blackhole_arc.h
```diff
@@ -0,0 +1,59 @@
...
+#pragma once
+
+#include <cstdint>
+
+namespace tt::umd {
+
+namespace blackhole {
+
+// Note, this only includes message IDs that have actually be implemented in CMFW
+enum class ArcMessageType : uint8_t {
```
broskoTT, line 16: How come we don't have a similar enum for wh and gs?

device/api/umd/device/cluster.h
broskoTT: You're missing the part in this PR which is integrating these changes into the rest of UMD (which is fine).
But I'm wondering where will BlackholeArcMessageQueue end up being used, in which class? 
My main concern is whether this will indeed be surfaced through API somewhere. I guess there won't be a need for it? And in that case the header file should not be in the api folder. Most of our headers are in the api folder, due to not ideal design, but we should try to keep things private where possible, and expose only the true interface.

If BlackholeArcMessageQueue will only be used internally in, for example, local_chip.cpp or cluster_descriptor.cpp, then put blackhole_arc_message_queue.h  and blackhole_arc.h in device/ folder rather than device/api/umd/device. You can keep the same structure, e.g. blackhole_arc would go to device/types/

device/blackhole/blackhole_arc_message_queue.cpp
```diff
@@ -0,0 +1,144 @@
...
+#include "umd/device/blackhole_arc_message_queue.h"
+
+#include "umd/device/cluster.h"
+
+using namespace tt::umd;
+
+namespace tt::umd {
+
+BlackholeArcMessageQueue::BlackholeArcMessageQueue(
+    Cluster* cluster,
```
broskoTT, line 15: This will transition to Chip once write/read_from_device are moved to Chip, right?

broskoTT: Yeah, or even better down the path maybe even ends up being Core*, since this is only communicating with ARC CORE...

device/api/umd/device/blackhole_arc_message_queue.h
```diff
@@ -0,0 +1,65 @@
...
+
+public:
+    BlackholeArcMessageQueue(
+        Cluster* cluster,
+        const chip_id_t chip,
+        const uint64_t base_address,
+        const uint64_t size,
+        const CoreCoord arc_core);
+
+    uint32_t send_message(const ArcMessageType message_type, uint16_t arg0 = 0, uint16_t arg1 = 0);
```
broskoTT, line 38: Comment here that this call is blocking without timeout (if I read the code right, pop_response will block).

device/api/umd/device/blackhole_arc_message_queue.h
```diff
@@ -0,0 +1,65 @@
...
+
+    static std::shared_ptr<BlackholeArcMessageQueue> get_blackhole_arc_message_queue(
```
broskoTT, line 40: unique_ptr?

Also comment that this class is not thread safe (If I read it right it is not)

device/blackhole/blackhole_arc_message_queue.cpp
```diff
@@ -0,0 +1,144 @@
...
+}
+
+void BlackholeArcMessageQueue::write_words(uint32_t* data, size_t num_words, size_t offset) {
+    cluster->write_to_device(
+        data,
+        num_words * sizeof(uint32_t),
+        chip,
+        arc_core,
+        base_address + offset * sizeof(uint32_t),
+        "LARGE_WRITE_TLB");
```
broskoTT, line 40: are LARGE_READ_TLB and LARGE_WRITE_TLB meant for these kind of transfers? Shouldn't REG_TLB be used?

device/blackhole/blackhole_arc_message_queue.cpp
```diff
@@ -0,0 +1,144 @@
...
+    cluster->write_to_device(&ARC_FW_INT_VAL, sizeof(uint32_t), chip, arc_core, ARC_FW_INT_ADDR, "LARGE_WRITE_TLB");
+}
+
+void BlackholeArcMessageQueue::push_request(std::array<uint32_t, BlackholeArcMessageQueue::entry_len>& request) {
+    cluster->l1_membar(chip, "LARGE_READ_TLB");
+    uint32_t request_queue_wptr = read_word(request_wptr_offset);
+
+    while (true) {
+        uint32_t request_queue_rptr = read_word(request_rptr_offset);
+        if (abs((int)request_queue_rptr - (int)request_queue_wptr) % (2 * size) != size) {
```
broskoTT, line 60: Explain this math in a comment, or add a named variable which would help with this.

device/blackhole/blackhole_arc_message_queue.cpp
```diff
@@ -0,0 +1,144 @@
...
+        arc_core,
+        base_address + offset * sizeof(uint32_t),
+        "LARGE_WRITE_TLB");
+}
+
+void BlackholeArcMessageQueue::create_request(
+    uint32_t* request, ArcMessageType message_type, uint32_t* data, size_t num_words) {
+    request[0] = (uint32_t)message_type;
+    memcpy(request + 1, data, num_words * sizeof(uint32_t));
+    memset(request + 1 + num_words, 0, (BlackholeArcMessageQueue::entry_len - (1 + num_words)) * sizeof(uint32_t));
```
broskoTT, line 47: Can you put some of these formulas from these two lines in dedicated named variables which would help understand this math

device/blackhole/blackhole_arc_message_queue.cpp
```diff
@@ -0,0 +1,144 @@
...
+        if (abs((int)request_queue_rptr - (int)request_queue_wptr) % (2 * size) != size) {
+            break;
+        }
+    }
+
+    // Offset in words.
+    uint32_t request_entry_offset = header_len + (request_queue_wptr % size) * BlackholeArcMessageQueue::entry_len;
+    write_words(request.data(), BlackholeArcMessageQueue::entry_len, request_entry_offset);
+
+    request_queue_wptr = (request_queue_wptr + 1) % (2 * size);
```
broskoTT, line 69: Why + 1) % (2 * size)

also below

device/blackhole/blackhole_arc_message_queue.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+    std::array<uint32_t, BlackholeArcMessageQueue::entry_len> request = {(uint32_t)message_type, arg, 0, 0, 0, 0, 0, 0};
+
+    push_request(request);
+
+    std::array<uint32_t, BlackholeArcMessageQueue::entry_len> response = pop_response();
+
+    uint32_t status = response[0] & 0xFF;
+
+    if (status < 240) {
```
broskoTT, line 110: Is the exact value important, do we want to log it?

device/blackhole/blackhole_arc_message_queue.cpp
```diff
@@ -0,0 +1,144 @@
+/*
+ * SPDX-FileCopyrightText: (c) 2024 Tenstorrent Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#include "umd/device/blackhole_arc_message_queue.h"
+
+#include "umd/device/cluster.h"
+
+using namespace tt::umd;
```
broskoTT, line 10: Not needed if everything below is wrapped in tt::umd namespace anyway, right?

tests/blackhole/test_arc_messages_bh.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+    std::unique_ptr<Cluster> cluster = get_cluster();
+
+    std::shared_ptr<BlackholeArcMessageQueue> blackhole_arc_msg_queue =
+        BlackholeArcMessageQueue::get_blackhole_arc_message_queue(
+            cluster.get(), 0, BlackholeArcMessageQueueIndex::APPLICATION);
+
+    for (int i = 0; i < num_loops; i++) {
+        uint32_t response = blackhole_arc_msg_queue->send_message(ArcMessageType::TEST);
+        if (response != 0) {
```
broskoTT, line 34: ASSERT_NE(response, 0);

broskoTT (approved):
Glad to see this working!

device/blackhole/blackhole_arc_message_queue.cpp
pjanevskiTT: exactly

device/api/umd/device/types/blackhole_arc.h
pjanevskiTT: I guess just nobody put wh and gs message codes in an enum :). Not sure whether all arc messages are same for all arch, I can investigate and create something similar for gs and bh

device/blackhole/blackhole_arc_message_queue.cpp
pjanevskiTT: queue pointers wrap at 2*size, there is an explanation on top of BlachkoleArcMessageQueue class

device/blackhole/blackhole_arc_message_queue.cpp
pjanevskiTT: Related to wrapping of queues on double the size

device/blackhole/blackhole_arc_message_queue.cpp
pjanevskiTT: No, this is CSM memory so something like L1 for tensix cores, we should use non-reg tlbs for this. On blackhole I think we have dedicated TLBs for ARC cores, but this is not yet implemented

device/api/umd/device/cluster.h
pjanevskiTT: I will put this as part of the issue where I will integrate everything inside cluster, since at the moment this should be used separately. When I integrate it I will remove it from api folder

PR #456, opened 2025-01-13
HarvestingMasks struct

pjanevskiTT:
### Issue

#435 

### Description

Add struct for software harvesting which should be used to simulate all 3 types of harvesting (Tensix, DRAM, ETH)

### List of the changes

- Add struct for software harvesting
- Integrate it into Cluster

### Testing
CI

### API Changes
/

### DIFF
--- a/device/api/umd/device/blackhole_coordinate_manager.h
+++ b/device/api/umd/device/blackhole_coordinate_manager.h
@@ -13,14 +13,13 @@ class BlackholeCoordinateManager : public CoordinateManager {
 public:
     BlackholeCoordinateManager(
         const bool noc_translation_enabled,
+
+        tt::umd::HarvestingMasks harvesting_masks,
         const tt_xy_pair& tensix_grid_size,
         const std::vector<tt_xy_pair>& tensix_cores,
-        const size_t tensix_harvesting_mask,
         const tt_xy_pair& dram_grid_size,
         const std::vector<tt_xy_pair>& dram_cores,
-        const size_t dram_harvesting_mask,
         const std::vector<tt_xy_pair>& eth_cores,
-        const size_t eth_harvesting_mask,
         const tt_xy_pair& arc_grid_size,
         const std::vector<tt_xy_pair>& arc_cores,
         const tt_xy_pair& pcie_grid_size,
--- a/device/api/umd/device/blackhole_coordinate_manager.h
+++ b/device/api/umd/device/blackhole_coordinate_manager.h
@@ -13,7 +13,6 @@ class BlackholeCoordinateManager : public CoordinateManager {
 public:
     BlackholeCoordinateManager(
         const bool noc_translation_enabled,
-
         tt::umd::HarvestingMasks harvesting_masks,
         const tt_xy_pair& tensix_grid_size,
         const std::vector<tt_xy_pair>& tensix_cores,
--- a/device/api/umd/device/cluster.h
+++ b/device/api/umd/device/cluster.h
@@ -16,8 +16,6 @@
 
 #include "fmt/core.h"
 #include "tt_silicon_driver_common.hpp"
-#include "tt_soc_descriptor.h"
-#include "tt_xy_pair.h"
 #include "umd/device/blackhole_arc_message_queue.h"
 #include "umd/device/chip/chip.h"
 #include "umd/device/tt_device/tt_device.h"
@@ -25,7 +23,6 @@
 #include "umd/device/types/arch.h"
 #include "umd/device/types/cluster_descriptor_types.h"
 #include "umd/device/types/cluster_types.h"
-#include "umd/device/types/harvesting.h"
 #include "umd/device/types/tlb.h"
 
 using TLB_DATA = tt::umd::tlb_data;
@@ -500,7 +497,7 @@ public:
         const bool skip_driver_allocs = false,
         const bool clean_system_resources = false,
         bool perform_harvesting = true,
-        std::unordered_map<chip_id_t, SoftwareHarvesting> simulated_harvesting_masks = {});
+        std::unordered_map<chip_id_t, HarvestingMasks> simulated_harvesting_masks = {});
 
     /**
      * Cluster constructor.
@@ -520,7 +517,7 @@ public:
         const bool skip_driver_allocs = false,
         const bool clean_system_resources = false,
         bool perform_harvesting = true,
-        std::unordered_map<chip_id_t, SoftwareHarvesting> simulated_harvesting_masks = {});
+        std::unordered_map<chip_id_t, HarvestingMasks> simulated_harvesting_masks = {});
 
     /**
      * Cluster constructor.
@@ -544,7 +541,7 @@ public:
         const bool skip_driver_allocs = false,
         const bool clean_system_resources = false,
         bool perform_harvesting = true,
-        std::unordered_map<chip_id_t, SoftwareHarvesting> simulated_harvesting_masks = {});
+        std::unordered_map<chip_id_t, HarvestingMasks> simulated_harvesting_masks = {});
 
     /**
      * Cluster constructor.
@@ -566,7 +563,7 @@ public:
         const bool skip_driver_allocs = false,
         const bool clean_system_resources = false,
         bool perform_harvesting = true,
-        std::unordered_map<chip_id_t, SoftwareHarvesting> simulated_harvesting_masks = {});
+        std::unordered_map<chip_id_t, HarvestingMasks> simulated_harvesting_masks = {});
 
     /**
      * Cluster constructor which creates a cluster with Mock chips.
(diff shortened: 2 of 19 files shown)


device/api/umd/device/types/harvesting.h
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+#pragma once
+
+#include <cstdint>
+
+namespace tt::umd {
+
+struct SoftwareHarvesting {
+    uint32_t tensix_harvesting_mask;
```
broskoTT, line 14: The objects of type SoftwareHarvesting are likely to be called something_harvesting. 
So accessing all of these would be ..._harvesting.tensix_harvesting_mask.
The _harvesting_ in names of all of these members looks a bit like a surplus?

device/cluster.cpp
```diff
@@ -440,19 +442,25 @@ std::unique_ptr<Chip> Cluster::construct_chip_from_cluster(
...
     uint32_t tensix_harvesting_mask =
         get_tensix_harvesting_mask(chip_id, cluster_desc, perform_harvesting, simulated_harvesting_masks);
+    size_t dram_harvesting_mask = get_dram_harvesting_mask(chip_id, perform_harvesting, simulated_harvesting_masks);
+    size_t eth_harvesting_mask = get_eth_harvesting_mask(chip_id, perform_harvesting, simulated_harvesting_masks);
     tt_SocDescriptor soc_desc = tt_SocDescriptor(
-        soc_desc_path, cluster_desc->get_noc_translation_table_en().at(chip_id), tensix_harvesting_mask);
+        soc_desc_path,
+        cluster_desc->get_noc_translation_table_en().at(chip_id),
+        tensix_harvesting_mask,
+        dram_harvesting_mask,
```
broskoTT, line 454: Do you want to change the constructor of socdescriptor to accept SoftwareHarvesting? Or some of the next PRs?

device/cluster.cpp
```diff
@@ -440,19 +442,25 @@ std::unique_ptr<Chip> Cluster::construct_chip_from_cluster(
     chip_id_t chip_id,
     tt_ClusterDescriptor* cluster_desc,
     bool perform_harvesting,
-    std::unordered_map<chip_id_t, uint32_t>& simulated_harvesting_masks) {
+    std::unordered_map<chip_id_t, SoftwareHarvesting>& simulated_harvesting_masks) {
     uint32_t tensix_harvesting_mask =
         get_tensix_harvesting_mask(chip_id, cluster_desc, perform_harvesting, simulated_harvesting_masks);
+    size_t dram_harvesting_mask = get_dram_harvesting_mask(chip_id, perform_harvesting, simulated_harvesting_masks);
+    size_t eth_harvesting_mask = get_eth_harvesting_mask(chip_id, perform_harvesting, simulated_harvesting_masks);
```
broskoTT, line 449: I don't get it.  Why not just modify get_tensix_harvesting_mask into get_harvesting_masks() ? Why separate out these functions? They will always be used together when fetched right?

broskoTT (approved):

device/cluster.cpp
pjanevskiTT: you mean together with silicon masks? Arguments of the constructor right now are masks from silicon combined with sw masks

device/api/umd/device/types/harvesting.h
pjanevskiTT: Not sure that we should rely on variable naming. I think var names are pretty clear right now. I can add the docs just for the expected format, don't think we should change the naming

device/cluster.cpp
pjanevskiTT: Yes, but what would be the return type there? I can use the struct that I have created but then it should probably be called HarvestingMasks (without software). It was mostly made for sw people if they want to simulate the harvesting, we should never really use it if we only fetch things from silicon

abhullar-tt:
will the software harvesting + silicon harvesting masks be combined or is the idea for users to query both and combine them?

pjanevskiTT:
Well we need to apply silicon harvesting masks always in order to not hang. Software harvesting is combined at the moment with silicon masks. I am happy to change that anyway you want, this is probably never going to be used by UMD, it is mostly for clients like tt-lens and tt-metal. If you have any idea how to implement this with different API let me know

abhullar-tt:
> Well we need to apply silicon harvesting masks always in order to not hang. Software harvesting is combined at the moment with silicon masks. I am happy to change that anyway you want, this is probably never going to be used by UMD, it is mostly for clients like tt-lens and tt-metal. If you have any idea how to implement this with different API let me know

Internally if UMD stores software/silicon masks separately that is okay but would prefer if they are exposed together just as `HarvestingMask`

abhullar-tt:
> Sure, is logical representation of harvesting masks good for you? That means if bit 0 is set, 0 logical row of tensix is harvested, bank 0 and eth channel 0?

yeah that is okay, its different than today (I think we get physical rows)

device/cluster.cpp
broskoTT: Ah yes, I misunderstood as it can be some from this and the other comment.
We will be passing around these three together. I don't see why we wouldn't have HarvestingMasks struct? Whats the difference if you pass true hardware reported harvesting or software reported one?

broskoTT:
Client should not have an option to not harvest truly harvested hw rows.
I think today we do combine whatever is passed as simulated harvesting mask with the one reported in cluster descriptor, and I think it should stay that way.

device/cluster.cpp
pjanevskiTT: Yes I will change this in this PR

pjanevskiTT:
I put all harvesting masks into a struct, so now those can be expose together as Almeet suggested. Patched the tests as well

device/api/umd/device/blackhole_coordinate_manager.h
```diff
@@ -13,14 +13,13 @@ class BlackholeCoordinateManager : public CoordinateManager {
 public:
     BlackholeCoordinateManager(
         const bool noc_translation_enabled,
+
```
broskoTT, line 16: How does this empty line pass our clang?

device/blackhole/blackhole_coordinate_manager.cpp
```diff
@@ -43,28 +39,30 @@ BlackholeCoordinateManager::BlackholeCoordinateManager(
...
 }
 
 void BlackholeCoordinateManager::translate_tensix_coords() {
-    if (CoordinateManager::get_num_harvested(tensix_harvesting_mask) > tensix_grid_size.x) {
-        tensix_harvesting_mask = 0;
+    if (CoordinateManager::get_num_harvested(harvesting_masks.tensix_harvesting_mask) > tensix_grid_size.x) {
+        harvesting_masks.tensix_harvesting_mask = 0;
     }
-    size_t num_harvested_x = CoordinateManager::get_num_harvested(tensix_harvesting_mask);
+    size_t num_harvested_x = CoordinateManager::get_num_harvested(harvesting_masks.tensix_harvesting_mask);
```
broskoTT, line 58: You see that you now have `harvesting_masks.something_harvesting_mask` everywhere :)
Just a thought to reconsider again to name members "tensix, dram, eth", without the harvesting_mask suffix, for more concise code, but feel free to leave it like this.

tests/api/test_core_coord_translation_wh.cpp
```diff
@@ -207,7 +207,7 @@ TEST(CoordinateManager, CoordinateManagerWormholePhysicalTranslatedHarvestedMapp
     const size_t harvesting_mask = (1 << 0) | (1 << 1);
     const size_t num_harvested = CoordinateManager::get_num_harvested(harvesting_mask);
     std::shared_ptr<CoordinateManager> coordinate_manager =
-        CoordinateManager::create_coordinate_manager(tt::ARCH::WORMHOLE_B0, true, harvesting_mask);
+        CoordinateManager::create_coordinate_manager(tt::ARCH::WORMHOLE_B0, true, {harvesting_mask});
```
broskoTT, line 210: Please consider naming these explicitly tensix_harvesting_mask. Same when eth or dram is used, consider adding prefix. Or create a separete HarvestingMasks structure outside of this function call. Same for other 20 - 30 calls in tests

device/api/umd/device/cluster.h
```diff
@@ -499,7 +497,7 @@ class Cluster : public tt_device {
         const bool skip_driver_allocs = false,
         const bool clean_system_resources = false,
         bool perform_harvesting = true,
-        std::unordered_map<chip_id_t, uint32_t> simulated_harvesting_masks = {});
+        std::unordered_map<chip_id_t, HarvestingMasks> simulated_harvesting_masks = {});
```
broskoTT, line 500: Not sure if this breaks compilation for tt-metal, but if it does (should be able to see from the ci job) please add a PR for that.

broskoTT (approved):
looking good!

device/api/umd/device/cluster.h
pjanevskiTT: It is not breaking metal build. This is something that metal is not using in standard code path, more for testing, so they probably don't have this in the repo

device/blackhole/blackhole_coordinate_manager.cpp
pjanevskiTT: Keeping it as it is for now, maybe we can change it with some external feedback later on

PR #457, opened 2025-01-14
Implement ETH channel id mapping

pjanevskiTT:
### Issue
/

### Description

Viewing eth cores as grid doesn't make much sense, we have channel -> core mapping that is used much more.

### List of the changes

- Implement logical (0, channel) to other coordinate systems mappings
- Change ETH cores list to match this
- Rename blackhole soc descriptors to type1/2 instead of local/remote

### Testing
Existing tests

### API Changes
/


device/coordinate_manager.cpp
```diff
@@ -448,7 +443,7 @@ tt_xy_pair CoordinateManager::get_tensix_grid_size() const {
 
 tt_xy_pair CoordinateManager::get_dram_grid_size() const { return dram_grid_size; }
 
-tt_xy_pair CoordinateManager::get_eth_grid_size() const { return eth_grid_size; }
+tt_xy_pair CoordinateManager::get_eth_grid_size() const { return {1, num_eth_channels}; }
```
broskoTT, line 446: Do we want to change this api as well, to get_num_eth_channels?

PR #460, opened 2025-01-17
Fix flatc issue on cmake 3.22

blozano-tt:
### Description
CMake 3.22 can't find flatc

### List of the changes
- Switch from functional based dependencies to declarative
- Generate header along with dependency build phase
- Define dependency between device target and generated header

### Testing

### API Changes


blozano-tt:
@afuller-TT @broskoTT 
We discussed this PR.

While we do like moving the dependencies to a `add_subdirectory` scope ... we don't need to move the generation around.

The error that I saw was not caused by cmake version, but by switching to using a toolchain file. Very strange dependency if you ask me, but it is what it is.

Closing this PR for now, we can circle back and clean up the dependency definition in the future.

PR #466, opened 2025-01-20
Fix macro collision

joelsmithTT:
### Issue
N/A

### Description
sys/sysmacros.h defines major() and minor() macros.  These may collide with semver_t initialization list.

### List of the changes
Rewrite the semver_t constructor to avoid having `major(...)` and `minor(...)`.

### Testing
CI

### API Changes

### DIFF
--- a/device/api/umd/device/semver.hpp
+++ b/device/api/umd/device/semver.hpp
@@ -21,11 +21,15 @@ namespace tt::umd {
 
 class semver_t {
 public:
-    const uint64_t major;
-    const uint64_t minor;
-    const uint64_t patch;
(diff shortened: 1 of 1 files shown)


blozano-tt:
@joelsmithTT I’m okay with any solution, but just curious why you preferred the rename vs undefining the macros? They seem like dangerous names for macros in my view.

blozano-tt (approved):

joelsmithTT:
RE: undefining the macros - I didn't want to fight with glibc - I think avoiding the conflict entirely is better than working around it.

PR #468, opened 2025-01-22
Additional logical coord tests

broskoTT:
### Issue
/

### Description
There was a fix for something a while back, and this test would've helped find this issue. I think it was related to the ordering of the cores in the yamls.

### List of the changes
- Additional tests which verify some logical coordinates are pointing to the expected ones.

### Testing
CI

### API Changes


tests/api/test_core_coord_translation_bh.cpp
```diff
@@ -51,6 +50,50 @@ TEST(CoordinateManager, CoordinateManagerBlackholeTopLeftCore) {
...
+    // This is targeting first column of Tensix cores on NOC layout.
+    const size_t tensix_harvesting_mask = (1 << 0);
+    std::shared_ptr<CoordinateManager> coordinate_manager =
+        CoordinateManager::create_coordinate_manager(tt::ARCH::BLACKHOLE, true, {tensix_harvesting_mask});
+
+    tt_xy_pair tensix_grid_size = coordinate_manager->get_grid_size(CoreType::TENSIX);
+    EXPECT_EQ(tensix_grid_size.x, 13);
+    EXPECT_EQ(tensix_grid_size.y, 10);
+    size_t max_x = tensix_grid_size.x - 1;
+    CoreCoord logical_coords = CoreCoord(max_x, 0, CoreType::TENSIX, CoordSystem::LOGICAL);
```
pjanevskiTT, line 66: nit: looks like max_x is initialized in the line above and then used only here. Maybe put directly tensix_grid_size.x - 1 into the x coordinate of CoreCoord. Same for all other tests and max_x/y variables

pjanevskiTT (approved):
nice

Issue #476, opened 2025-01-23
Harvesting masks - 1 is enabled on Blackhole


pjanevskiTT:
So far our harvesting mask followed the pattern

- bit is equal to 0 -> tensix row/bank/eth channel is enabled
- bit is equal to 1 -> tensix row/bank/eth channel is disabled

Blackhole changes this pattern. Easy solution for this is to either reverse old harvesting masks or reverse Blackhole harvesting masks when we read those from silicon. Probably closer to invert the blackhole harvesting masks at the moment

pjanevskiTT:
Closed by #494

Issue #459, opened 2025-01-15
Blackhole topology discovery


pjanevskiTT:
create-ethernet-map is getting deprecated for Blackhole and there is a new way to create cluster descriptor. Design explained in [ethernet for blackhole](https://tenstorrent.sharepoint.com/:w:/r/sites/Jasmina/_layouts/15/Doc.aspx?sourcedoc=%7B940DD52D-30F9-442A-B24A-7944A7D8031D%7D&file=Ethernet%20for%20Metal.docx&action=default&mobileredirect=true)

On Blackhole all products are going to have only MMIO chips but we still to need to recognize the topology, basically what create-ethernet-map was doing

broskoTT:
I added sub-issues to have the same parent issue as this one.
Now, this task as an individual task was actually done.

PR #477, opened 2025-01-23
Blackhole basic topology discovery

pjanevskiTT:
### Issue

Multiple issues, some of these should be closed and some are master issues. Sub-issues are going to be solved later

#459 - This is master issue. This PR implements basic BH topology discovery. Mulitple subissues are going to be solved according to priority.
#476 - this PR should close it
#94 - messages are moving to KMD, and the telemetry reading is implemented in this PR, so we can close this issue after this PR as well

### Description

Implement basic blackhole topology discovery. It implements discovery over ETH channels for all chips that are connected to PCIE of the host where UMD is running. There will be more issues added to #459 regarding topology discovery

### List of the changes

- Implement BH telemetry reading
- Implement reading of ETH structures for blackhole chip2chip connectivity
- create cluster descriptor from chip2chip info
- integrate blackhole cluster descriptor creation into cluster code path

### Testing

This should be transparent for tests. All tests should pass on CI

### API Changes
/

### Additional notes

Functions for creating tt_cluster_descriptor should probably move away from tt_cluster_descriptor class. I am going to work towards a topology service that is going to capture everything related to topology. It would be nice to not change this in this PR. tt_cluster_descriptor API is not mature enough as well to support everything here, so I just wanted to create some basic logic for topology discovery

### DIFF
--- a/device/api/umd/device/types/blackhole_eth.h
+++ b/device/api/umd/device/types/blackhole_eth.h
@@ -13,30 +13,6 @@ namespace blackhole {
 static constexpr uint32_t NUM_SERDES_LANES = 8;
 
 typedef enum {
-    LOOPBACK_NONE,
-    LOOPBACK_NEAR_END_MAC,
-    LOOPBACK_NEAR_END_FIFO,
-    LOOPBACK_NEAR_END_PMA,
-    LOOPBACK_FAR_END_FIFO,
-    LOOPBACK_SERDES_NEP,
-    LOOPBACK_SERDES_NES_PREDRIVER,
-} loopback_mode_e;
-
-typedef enum {
-    AW_MANUAL_EQ,
-    AW_ANLT_MODE,
-    AW_LT_MODE,
-} link_train_mode_e;
-
-typedef enum {
-    ORION,
-    P100,
-    P150,
-    P300,
-    UBB,
-} pcb_type_e;
-
-typedef enum {
     LINK_TRAIN_TRAINING,
     LINK_TRAIN_SKIP,
     LINK_TRAIN_PASS,
@@ -56,61 +32,6 @@ typedef enum {
     PORT_DOWN,
     PORT_UNUSED,
 } port_status_e;
-
-typedef enum {
-    NRZ_1p25,
-    NRZ_10p3125,
-    NRZ_25p78125,
-    NRZ_26p5625,
-    NRZ_53p125,
-    PAM4_53p125,
-    PAM4_106p25,
-    NUM_RATES,
-} serdes_rate_e;
-
-typedef enum {
-    AW_128W = 7,
-    AW_64W = 6,
-    AW_40W = 5,
-    AW_32W = 4,
-    AW_20W = 3,
-    AW_16W = 2,
-    AW_10W = 1,
-} serdes_width_e;
-
-typedef enum {
-    AW_TIMER,
-    AW_DWELL,
-} serdes_bist_mode_e;
-
-typedef enum {
-    AW_PRBS7,
-    AW_PRBS9,
-    AW_PRBS11,
-    AW_PRBS13,
-    AW_PRBS15,
-    AW_PRBS23,
-    AW_PRBS31,
-    AW_QPRBS13,
-    AW_JP03A,
-    AW_JP03B,
-    AW_LINEARITY_PATTERN,
-    AW_USER_DEFINED_PATTERN,
-    AW_FULL_RATE_CLOCK,
-    AW_HALF_RATE_CLOCK,
-    AW_QUARTER_RATE_CLOCK,
-    AW_PATT_32_1S_32_0S,
-    AW_BIST_PATTERN_MAX,
-} serdes_bist_pattern_e;
-
-typedef enum {
-    AW_FULL_EQ,
-    AW_EVAL_ONLY,
-    AW_INIT_EVAL,
-    AW_CLEAR_EVAL,
-    AW_FULL_EQ_FOM,
-    AW_EVAL_ONLY_FOM,
-} serdes_eq_type_e;
 
 typedef struct {
     uint32_t patch : 8;
--- a/device/api/umd/device/types/blackhole_eth.h
+++ b/device/api/umd/device/types/blackhole_eth.h
@@ -1,0 +1,236 @@
+/*
+ * SPDX-FileCopyrightText: (c) 2025 Tenstorrent Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+namespace tt::umd {
+
+namespace blackhole {
+
+static constexpr uint32_t NUM_SERDES_LANES = 8;
+
+typedef enum {
+    LOOPBACK_NONE,
+    LOOPBACK_NEAR_END_MAC,
+    LOOPBACK_NEAR_END_FIFO,
+    LOOPBACK_NEAR_END_PMA,
+    LOOPBACK_FAR_END_FIFO,
+    LOOPBACK_SERDES_NEP,
+    LOOPBACK_SERDES_NES_PREDRIVER,
+} loopback_mode_e;
+
+typedef enum {
+    AW_MANUAL_EQ,
+    AW_ANLT_MODE,
+    AW_LT_MODE,
+} link_train_mode_e;
+
+typedef enum {
+    ORION,
+    P100,
+    P150,
+    P300,
+    UBB,
+} pcb_type_e;
+
+typedef enum {
+    LINK_TRAIN_TRAINING,
+    LINK_TRAIN_SKIP,
+    LINK_TRAIN_PASS,
+    LINK_TRAIN_INT_LB,
+    LINK_TRAIN_EXT_LB,
+    LINK_TRAIN_TIMEOUT_MANUAL_EQ,
+    LINK_TRAIN_TIMEOUT_ANLT,
+    LINK_TRAIN_TIMEOUT_CDR_LOCK,
+    LINK_TRAIN_TIMEOUT_BIST_LOCK,
+    LINK_TRAIN_TIMEOUT_LINK_UP,
+    LINK_TRAIN_TIMEOUT_CHIP_INFO,
+} link_train_status_e;
+
+typedef enum {
+    PORT_UNKNOWN,
+    PORT_UP,
+    PORT_DOWN,
+    PORT_UNUSED,
+} port_status_e;
+
+typedef enum {
+    NRZ_1p25,
+    NRZ_10p3125,
+    NRZ_25p78125,
+    NRZ_26p5625,
+    NRZ_53p125,
+    PAM4_53p125,
+    PAM4_106p25,
+    NUM_RATES,
+} serdes_rate_e;
+
+typedef enum {
+    AW_128W = 7,
+    AW_64W = 6,
+    AW_40W = 5,
+    AW_32W = 4,
+    AW_20W = 3,
+    AW_16W = 2,
+    AW_10W = 1,
+} serdes_width_e;
+
+typedef enum {
+    AW_TIMER,
+    AW_DWELL,
+} serdes_bist_mode_e;
+
+typedef enum {
+    AW_PRBS7,
+    AW_PRBS9,
+    AW_PRBS11,
+    AW_PRBS13,
+    AW_PRBS15,
+    AW_PRBS23,
+    AW_PRBS31,
+    AW_QPRBS13,
+    AW_JP03A,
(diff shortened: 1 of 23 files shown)


pjanevskiTT:
@abhullar-tt tagging you for visibility, if this makes your work on tt-metal easier. No need to review it if you don't want it. API should stay the same

pjanevskiTT:
Blackhole unit tests passing

device/tt_cluster_descriptor.cpp
```diff
@@ -894,3 +901,173 @@ const std::unordered_map<chip_id_t, std::unordered_set<chip_id_t>> &
...
+                log_debug(LogSiliconDriver, "Eth core ({}, {}) on chip {} is active", eth_core.x, eth_core.y, chip_id);
+                // active eth core
+                const chip_info_t &local_info = boot_results.local_info;
+                const chip_info_t &remote_info = boot_results.remote_info;
+
+                chip_id_t local_chip_id = desc->get_chip_id(local_info);
+                chip_id_t remote_chip_id = desc->get_chip_id(remote_info);
+
+                // Adding a connection only one way, the other chip should add it another way.
+                desc->ethernet_connections[local_chip_id][local_info.eth_id] = {remote_chip_id, remote_info.eth_id};
```
abhullar-tt, line 1049: also the eth id are not the same as logical eth channels, see mapping from eth id to noc0 x coordinates [here](https://tenstorrent.sharepoint.com/:x:/r/sites/SOC/Blackhole%20Documents/Blackhole%20-%20NOC%20Co-ordinates.xlsx?d=w449397eff6fc48abaed13762398c30dd&csf=1&web=1&e=bKt38y)


also:
```
static constexpr std::array<uint32_t, 14> eth_id_to_noc0_x_mapping =  {1, 16, 2, 15, 3, 14, 4, 13, 5, 12, 6, 11, 7, 10};
```

abhullar-tt: on WH do we get the logical eth channel?

device/tt_cluster_descriptor.cpp
```diff
@@ -894,3 +901,173 @@ const std::unordered_map<chip_id_t, std::unordered_set<chip_id_t>> &
...
+
+            } else if (boot_results.eth_status.port_status == port_status_e::PORT_DOWN) {
```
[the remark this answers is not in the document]
pjanevskiTT: Yes it can be tagged as active ethernet core, no real difference to UMD. We can provide the API to query the status of the link

device/tt_cluster_descriptor.cpp
pjanevskiTT: > on WH do we get the logical eth channel?

Not sure, there is no info on how to convert logical_eth_id to eth_id so I would guess we get same things as eth_id

abhullar-tt: yes we should use `eth_id` but just wanted to highlight that this isn't the same as noc0 physical eth x. on my branh in Metal `abhullar/bh-multichip` I do something like:
```
static constexpr std::array<uint32_t, 14> eth_id_to_noc0_x_mapping =  {1, 16, 2, 15, 3, 14, 4, 13, 5, 12, 6, 11, 7, 10};
CoreCoord physical_eth_core(eth_id_to_noc0_x_mapping[eth_id], eth_noc0_y); // where eth_id is read out from the eth core and eth_noc0_y is hardcoded to 1
CoreCoord remote_logical_eth = soc_desc.get_logical_ethernet_core_from_physical(physical_eth_core); // use metal soc descriptor to convert physical to logical
```

abhullar-tt: my question is what the eth id that cluster descriptor apis accept/return would be representing. would it be whatever is read out from device and its up to client to understand this conversion?

pjanevskiTT: well I think we should return whatever is read from the device. When it comes to getting actual core, UMD's CoreCoord supports the conversion from `eth_id` to physical core.

```cpp
CoreCoord eth_logical = CoreCoord(0, eth_channel, CoordSystem::LOGICAL, CoreType::ETH);
CoreCoord eth_physical = soc_desc.translate_coord_to(eth_logical, CoordSystem::PHYSICAL);
```
Not sure what is `remote_logical_eth` in the code that you pasted?

device/api/umd/device/types/blackhole_eth.h
```diff
@@ -0,0 +1,157 @@
...
+
+#pragma once
+
+namespace tt::umd {
+
+namespace blackhole {
+
+static constexpr uint32_t NUM_SERDES_LANES = 8;
+
+typedef enum {
```
broskoTT, line 15: Are these structs copied from somewhere in ARC FW? Can you reference the location in a comment?

device/api/umd/device/types/blackhole_telemetry.h
```diff
@@ -0,0 +1,54 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+namespace tt::umd {
+
+namespace blackhole {
+
+constexpr uint8_t TAG_BOARD_ID_HIGH = 1;
```
broskoTT, line 13: Same here, link the source of these constants if plausible

device/cluster.cpp
```diff
@@ -508,18 +509,48 @@ uint32_t Cluster::get_tensix_harvesting_mask(
     return tensix_harvesting_mask | simulated_harvesting_mask;
 }
 
+void Cluster::create_cluster_descriptor() {
```
broskoTT, line 512: I would not add this code here.

How are our clients who don't want the full Cluster, just the ClusterDescriptor going to interface? Why not move all of the contents from there to tt_ClusterDescriptor::create.
Cluster.cpp's code would end up being unchanged

device/tt_cluster_descriptor.cpp
```diff
@@ -894,3 +901,173 @@ const std::unordered_map<chip_id_t, std::unordered_set<chip_id_t>> &
...
+std::unique_ptr<tt_ClusterDescriptor> tt_ClusterDescriptor::create_cluster_descriptor(
+    const std::unordered_map<chip_id_t, std::unique_ptr<Chip>> &chips) {
+    std::unique_ptr<tt_ClusterDescriptor> desc = std::unique_ptr<tt_ClusterDescriptor>(new tt_ClusterDescriptor());
+
+    const auto tlb_index = tt::umd::blackhole::MEM_LARGE_READ_TLB;
+
+    for (auto &it : chips) {
+        const chip_id_t chip_id = it.first;
+        const std::unique_ptr<Chip> &chip = it.second;
+        desc->board_type_asic_location_to_chip_id.insert(
```
broskoTT, line 1007: Hmmm....
This is too entangled.

So in Cluster, you call tt_ClusterDescriptor::get_cluster_chip_info to obtain ChipInfo, than wrap it up in Chip classes, then you pass those Chip classes back to tt_ClusterDescriptor. Then, tt_ClusterDescriptor fetches the ChipInfo (which it itself provided actually) to obtain some info and fetch another struct within itself?

In any case, when you remove code from cluster.cpp let's see how it turns out then.

I also suspect having ClusterDescriptor know about Chip is maybe not necessary? Everything can be achieved through TTDevice?

device/tt_device/tt_device.cpp
```diff
@@ -207,6 +208,20 @@ void TTDevice::read_block(uint64_t byte_addr, uint64_t num_bytes, uint8_t *buffe
     }
 }
 
+void TTDevice::read_from_device(
+    void *mem_ptr, tt_xy_pair core, uint64_t addr, uint32_t size, const uint32_t tlb_index) {
+    uint8_t *buffer_addr = static_cast<uint8_t *>(mem_ptr);
+    while (size > 0) {
+        auto [mapped_address, tlb_size] = set_dynamic_tlb(tlb_index, core, addr, tt::umd::tlb_data::Strict);
```
broskoTT, line 215: I think I'd really like to keep the problem of managing tlb's to the client of this class. The client should themselves program a tlb.

If this is left like this, then we quickly come to a question on what happens with multiple threads, then somebody has to report here (or ttdevice itself has to manage this) which tlbs are used, so that this function call doesn't break that, etc...

device/tt_cluster_descriptor.cpp
```diff
@@ -894,3 +901,173 @@ const std::unordered_map<chip_id_t, std::unordered_set<chip_id_t>> &
...
+    const auto tlb_index = tt::umd::blackhole::MEM_LARGE_READ_TLB;
+
+    tt_xy_pair arc_core = tt::umd::blackhole::ARC_CORES[0];
+    std::vector<ChipInfo> chip_info_vec;
+    for (auto &tt_device : tt_devices) {
+        uint32_t arc_boot_status;
+        do {
+            tt_device->read_from_device(
+                &arc_boot_status, arc_core, tt::umd::blackhole::SCRATCH_RAM_2, sizeof(uint32_t), tlb_index);
+        } while ((arc_boot_status & 0x7) != 0x5);
```
broskoTT, line 917: Can we have enums for these values?

device/tt_cluster_descriptor.cpp
```diff
@@ -894,3 +901,173 @@ const std::unordered_map<chip_id_t, std::unordered_set<chip_id_t>> &
...
+            } else if (tag_val == tt::umd::blackhole::TAG_ENABLED_ETH) {
+                // DRAM harvesting mask on Blackhole has 1 for enabled dram banks, so mask
+                // needs to be complemented to fit previous harvesting masks (1 for harvested).
+                dram_harvesting_mask = ~(telemetry_val & ((1 << 8) - 1));
+            } else if (tag_val == tt::umd::blackhole::TAG_ENABLED_GDDR) {
+                // ETH harvesting mask on Blackhole has 1 for enabled eth channels, so mask
+                // needs to be complemented to fit previous harvesting masks (1 for harvested).
+                eth_harvesting_mask = ~(telemetry_val & ((1 << 13) - 1));
+            }
+        }
```
broskoTT, line 970: The code here, reading multiple times from ARC, seems complex enough to maybe warrant a separate class?
The input to the class would be a mempointer to use with TTDevice when reading (this would practically be base address of tlb_index programmed to arc core)

device/tt_cluster_descriptor.cpp
```diff
@@ -894,3 +901,173 @@ const std::unordered_map<chip_id_t, std::unordered_set<chip_id_t>> &
...
+            tt_device->read_from_device(&tag_val, arc_core, tag_table_index, sizeof(uint16_t), tlb_index);
+            uint16_t offset_val;
+            tt_device->read_from_device(
+                &offset_val, arc_core, tag_table_index + sizeof(uint16_t), sizeof(uint16_t), tlb_index);
+
+            uint32_t telemetry_val;
+            tt_device->read_from_device(
+                &telemetry_val, arc_core, telemetry + offset_val * sizeof(uint32_t), sizeof(uint32_t), tlb_index);
+
+            tag_table_index += sizeof(uint32_t);
```
broskoTT, line 951: The previous 7 reads are hard to follow. Can this be restructured so that the reads are clearly read from OBJ_LOCATION into an OBJ.

For instance, I'm not sure what exists in telemetry_table, maybe that should be a struct, and then instead of reading  telemetry_table + sizeof(uint32_t), sizeof(uint32_t), we read the whole struct.

device/tt_cluster_descriptor.cpp
```diff
@@ -894,3 +901,173 @@ const std::unordered_map<chip_id_t, std::unordered_set<chip_id_t>> &
...
+        uint32_t entry_count;
+        tt_device->read_from_device(
+            &entry_count, arc_core, telemetry_table + sizeof(uint32_t), sizeof(uint32_t), tlb_index);
+
+        uint32_t telemetry;
+        tt_device->read_from_device(
+            &telemetry, arc_core, tt::umd::blackhole::SCRATCH_RAM_12, sizeof(uint32_t), tlb_index);
+
+        uint32_t count = entry_count;
+        uint32_t tag_table_index = telemetry_table + 2 * sizeof(uint32_t);
```
broskoTT, line 939: 2 * sizeof(uint32_t) is a magic number

device/tt_cluster_descriptor.cpp
broskoTT: Also some separation in at least two or three functions would make things clearer, e.g. process_telemetry_value(), telemetry_value_in_queue(), fetch_telemetry_value().

pjanevskiTT:
Closing this since it is split into multiple smaller PRs

Issue #486, opened 2025-01-27
Disable ethernet broadcast for all systems with MMIO access to all chips


aliuTT:

aliuTT:
```
- // MT: Initial BH - Disable dependency to ethernet firmware
-    if (arch_name == tt::ARCH::BLACKHOLE) {
+    // Disable dependency to ethernet firmware for all BH devices and WH devices with all chips having MMIO (e.g. UBB Galaxy)
+    if (cluster_desc->get_chips_with_mmio().size() == cluster_desc->get_all_chips().size()) {

```

PR #487, opened 2025-01-28
Disable ethernet broadcast for all systems with MMIO access to all chips

aliuTT:
### Issue
[(Link to Github issue(s))](https://github.com/tenstorrent/tt-umd/issues/486)

### Description
Ethernet broadcast was getting used for all-mmio system like UBB. This is causing race conditions when opening and closing drivers, since the reset signals are getting broadcasted through ethernet FW but the done check is through MMIO.

### List of the changes
Disable ethernet broadcast for systems with num chips equal to num mmio chips.

### Testing
Tests in Metal on UBB Galaxy

### API Changes

### DIFF
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -305,8 +305,9 @@ void Cluster::construct_cluster(
 
     create_device(local_chip_ids_, num_host_mem_ch_per_mmio_device, skip_driver_allocs, clean_system_resources);
 
-    // MT: Initial BH - Disable dependency to ethernet firmware
-    if (arch_name == tt::ARCH::BLACKHOLE) {
+    // Disable dependency to ethernet firmware for all BH devices and WH devices with all chips having MMIO (e.g. UBB
+    // Galaxy), do not disable for N150, was seeing some issues in CI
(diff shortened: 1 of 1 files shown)


pjanevskiTT (approved):
thanks for the fix

device/cluster.cpp
```diff
@@ -298,8 +298,9 @@ void Cluster::construct_cluster(
 
     create_device(local_chip_ids_, num_host_mem_ch_per_mmio_device, skip_driver_allocs, clean_system_resources);
 
-    // MT: Initial BH - Disable dependency to ethernet firmware
-    if (arch_name == tt::ARCH::BLACKHOLE) {
+    // Disable dependency to ethernet firmware for all BH devices and WH devices with all chips having MMIO (e.g. UBB
+    // Galaxy)
+    if (cluster_desc->get_chips_with_mmio().size() == cluster_desc->get_all_chips().size()) {
```
broskoTT, line 303: Actually, to be precise, not all chips present in the cluster_desc are necessarily used, the Cluster might've been constructed with less chips.
This might be a better comparison:
`if (remote_chip_ids_.empty())`

broskoTT (approved):
Thanks for the fix!

device/cluster.cpp
aliuTT: I can make the change

PR #488, opened 2025-01-28
Bump spdlog to unblock us from upgrading fmt

afuller-TT:
### Issue
One step towards https://github.com/tenstorrent/tt-metal/issues/7915

### Description
spdlog uses fmt, but the version we're using uses a function from fmt that is no longer there in the current release of fmt.
spdlog has since added support for newer fmt.  Pull in that update.

### List of the changes
Bump spdlog to the commit that works with current fmt.  Pointed to the hash itself since spdlog hasn't cut a 1.x release since last Nov.

### Testing
Built TT-Metalium locally with this change.

### API Changes

### DIFF
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -103,6 +103,11 @@ function(fetch_dependencies)
     ####################################################################################################################
     # spdlog
     ####################################################################################################################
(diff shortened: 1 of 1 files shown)


joelsmithTT (approved):
Thanks -- reminds me that I ought to actually integrate this.

PR #490, opened 2025-01-28
Use components to group installed pieces

afuller-TT:
### Issue
One step towards https://github.com/tenstorrent/tt-metal/issues/7915

### Description
Best practice is to put installed pieces into components.  And for them to be named uniquely to the project.  That way any consuming project can choose where to redirect each piece.
Any install items without a component go into "Unspecified".  And once two things land in the same component they're lumped together.  Separate components may be placed together by choice.

### List of the changes
* Created a FILE_SET for the public API to be installed
* Removed the INSTALL_INTERFACE directives; these are populated based on where the FILE_SET is told to be installed to.
  * This may be done by the consuming project, if they wish, and the include dir will be adjusted accordingly at install time
* Distinguished between files needed at runtime vs. those required only for SDK development.

### Testing
Built and packaged in TT-Metalium (on a hacked up branch; TT-Metalium has lots of work to do, too).

### API Changes

### DIFF
--- a/device/CMakeLists.txt
+++ b/device/CMakeLists.txt
@@ -53,12 +53,59 @@ target_sources(
         ${FBS_GENERATED_HEADER}
 )
 
+if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.23)
+    target_sources(
+        device
+        PUBLIC
+            FILE_SET api
+            TYPE HEADERS
+            BASE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/api
+            FILES
+                api/umd/device/architecture_implementation.h
+                api/umd/device/blackhole_arc_message_queue.h
+                api/umd/device/blackhole_coordinate_manager.h
+                api/umd/device/blackhole_implementation.h
+                api/umd/device/chip/chip.h
+                api/umd/device/chip/local_chip.h
+                api/umd/device/chip/mock_chip.h
+                api/umd/device/chip/remote_chip.h
+                api/umd/device/cluster.h
+                api/umd/device/coordinate_manager.h
+                api/umd/device/device_api_metal.h
+                api/umd/device/driver_atomics.h
+                api/umd/device/grayskull_coordinate_manager.h
+                api/umd/device/grayskull_implementation.h
+                api/umd/device/hugepage.h
+                api/umd/device/pci_device.hpp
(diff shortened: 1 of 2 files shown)


device/CMakeLists.txt
```diff
@@ -53,12 +53,59 @@ target_sources(
...
+                api/umd/device/wormhole_implementation.h
+    )
+endif()
+
 target_include_directories(
     device
     PUBLIC
         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/api>
-        $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}>
-        $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}/device>
```
blozano-tt, line 61: How is this going away?

blozano-tt (approved):

device/CMakeLists.txt
afuller-TT: One way to handle headers is to specify a path here in `target_include_directories` with the genex to have it only appear when it's moved into the install location.  Then another step must ensure that the headers are placed in said location.  One such way is with `PUBLIC_HEADERS` and the other is just`install(FILE...` or `install(DIRECTORY...`

In v3.23 [FILE_SET](https://cmake.org/cmake/help/latest/command/target_sources.html#file-sets) was introduced.  A target can define a FILE_SET and later `install(... FILE_SET ...)` says where to put it.  Since a FILE_SET is associated with a target, then when installing the target+file_set, CMake will tweak the target according to where the FILE_SET was instructed to be placed.

The FILE_SET approach has a couple benefits:
* It has a better separation of "what" vs "where".  The FILE_SET is the what, and the where is handled by the `install...` directives, not leaked into `target_include_directories`.
* A SSoT for the installed headers.  The include path will follow along with where the headers land and doesn't have its own source of truth that could diverge.
* FILE_SET maintains a directory hierarchy, which PUBLIC_HEADERS do not.

They have one significant drawback:
* Must use CMake v3.23+, which was released in July of '22.  Too late for Ubuntu 22.04 to have by default.

blozano-tt: Wow what a flex.

You are dwarfing my Ubuntu 18.04 era understanding of the world.

PR #493, opened 2025-01-29
Implement blackhole telemetry

pjanevskiTT:
### Issue

#94 - this should close the issue

### Description
Tearing apart things for work towards Blackhole topology. 
Implement telemetry reading for Blackhole. RW are going through TTDevice, could be switched to go through Chip in the future, although this is not something that should be on a perf path. Telemetry reading is not same for Wormhole. We can generalize this class and derive ARCH specific things in the future if we ever implement Wormhole telemetry from source.

### List of the changes
- Create BlackholeArcTelemetryReader that provides the API for reading of blackhole telemetry
- Add test

### Testing
Added test for BH telemetry. Existing CI

### API Changes
/

### DIFF
--- a/device/api/umd/device/types/blackhole_telemetry.h
+++ b/device/api/umd/device/types/blackhole_telemetry.h
@@ -1,0 +1,70 @@
+/*
+ * SPDX-FileCopyrightText: (c) 2025 Tenstorrent Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+namespace tt::umd {
+
+namespace blackhole {
+
+constexpr uint8_t NUMBER_TELEMETRY_TAGS = 38;
+
+constexpr uint8_t TAG_BOARD_ID_HIGH = 1;
+constexpr uint8_t TAG_BOARD_ID_LOW = 2;
+constexpr uint8_t TAG_ASIC_ID = 3;
+constexpr uint8_t TAG_HARVESTING_STATE = 4;
+constexpr uint8_t TAG_UPDATE_TELEM_SPEED = 5;
+constexpr uint8_t TAG_VCORE = 6;
+constexpr uint8_t TAG_TDP = 7;
+constexpr uint8_t TAG_TDC = 8;
+constexpr uint8_t TAG_VDD_LIMITS = 9;
+constexpr uint8_t TAG_THM_LIMITS = 10;
+constexpr uint8_t TAG_ASIC_TEMPERATURE = 11;
+constexpr uint8_t TAG_VREG_TEMPERATURE = 12;
+constexpr uint8_t TAG_BOARD_TEMPERATURE = 13;
+constexpr uint8_t TAG_AICLK = 14;
+constexpr uint8_t TAG_AXICLK = 15;
+constexpr uint8_t TAG_ARCCLK = 16;
+constexpr uint8_t TAG_L2CPUCLK0 = 17;
+constexpr uint8_t TAG_L2CPUCLK1 = 18;
+constexpr uint8_t TAG_L2CPUCLK2 = 19;
+constexpr uint8_t TAG_L2CPUCLK3 = 20;
+constexpr uint8_t TAG_ETH_LIVE_STATUS = 21;
+constexpr uint8_t TAG_DDR_STATUS = 22;
+constexpr uint8_t TAG_DDR_SPEED = 23;
+constexpr uint8_t TAG_ETH_FW_VERSION = 24;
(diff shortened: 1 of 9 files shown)


device/api/umd/device/types/blackhole_telemetry.h
```diff
@@ -0,0 +1,70 @@
...
+constexpr uint8_t TAG_PCIE_USAGE = 38;
+
+struct telemetry_entry {
+    uint16_t tag;
+    uint16_t offset;
+};
+
+// telemetry_table struct is going to be represented as a BlackholeArctelemetryReader class.
+// It is not possible to statically create this struct becase number of entries is not know at compile time.
+// struct telemetry_table {
```
broskoTT, line 61: Can we define this struct? The third and fourth element can be vectors, initially empty. 
Of course, you still won't be able just to read the struct fully, you'd still have to read element by element. But I still think it we'd organize vars like that instead of directly as members of BlackholeArcTelemetryReader

device/blackhole/blackhole_arc_telemetry_reader.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+    if (!telemetry_entry_available[telemetry_tag - 1]) {
+        throw std::runtime_error(fmt::format(
+            "Telemetry entry {} not available. You can use is_entry_available() to check if the entry is available.",
+            telemetry_tag));
+    }
+
+    const uint32_t offset = telemetry_offset[telemetry_tag - 1];
+    uint32_t telemetry_val;
+    tt_device->read_from_device(
```
broskoTT, line 78: Is the telemetry something that is constantly flowing, and then read_entry should just read the last known state?
Should we have some background thread polling the device occassionally?

tests/blackhole/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 set(UNIT_TESTS_BH_SRCS
     test_cluster_bh.cpp
     test_arc_messages_bh.cpp
+    test_arc_telemetry_bh.cpp
```
broskoTT, line 4: Thanks for separating this, it is much clearer now.

device/api/umd/device/types/blackhole_telemetry.h
broskoTT: Oh okay so actually BlackholeArcTelemetryReader is that exact struct itself.
I find it confusing that some info is here and some is in device/api/umd/device/blackhole_arc_telemetry_reader.h.
Can you do either of these:
- Revive this struct, make the arrays vectors, so they can be resized while initializing bh arc telemetry reader. The BlackholeArcTelemetryReader would then just hold the instance of this struct.
- Remove this comment altogether, move telemetry_entry to blackhole_arc_telemetry_reader.h, and add comments to private fields explaining how do these values sit exactly in ARC memory (kind of what this comments does implicitly).

device/blackhole/blackhole_arc_telemetry_reader.cpp
```diff
@@ -0,0 +1,93 @@
...
+        telemetry_data.data(),
+        BlackholeArcTelemetryReader::arc_core,
+        telemetry_values_addr,
+        entry_count * sizeof(uint32_t));
+
+    for (const blackhole::telemetry_entry& tag_entry : telemetry_tag_entries) {
+        const uint16_t tag_val = tag_entry.tag;
+        const uint16_t offset_val = tag_entry.offset;
+
+        telemetry_values[tag_val - 1] = telemetry_data[offset_val];
```
broskoTT, line 59: This logic is a bit complex (it is not obvious to me what happens here after ~30secs), can you add an explanation above this for loop, on what is referenced where, in general how are this values compiled?

broskoTT (approved):
(Y)

device/blackhole/blackhole_arc_telemetry_reader.cpp
pjanevskiTT: Sure

device/blackhole/blackhole_arc_telemetry_reader.cpp
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/507

device/api/umd/device/types/blackhole_telemetry.h
pjanevskiTT: I don't want to have this struct with vectors in BlackoleArcTelemetryReader because it makes search for certain entry more difficult then with static arrays. I will move this struct and add comments to BlackholeArcTelemtryReader class

PR #495, opened 2025-01-30
Address some CPM problems

blozano-tt:
### Description

- Align CPM Version with that in tt-metal
- Don't depend on an environment variable for CPM_SOURCE_CACHE, we can ingest a CMake Cache Var.

The eventual goal here, is to make it so that we can use a centralized location (Not inside the git tree) where we can cache the downloaded / built CPM dependencies.

### Testing
CI will show us the way.

### API Changes


cmake/CPM.cmake
```diff
@@ -2,16 +2,13 @@
...
 
-if(CPM_SOURCE_CACHE)
-    set(CPM_DOWNLOAD_LOCATION "${CPM_SOURCE_CACHE}/cpm/CPM_${CPM_DOWNLOAD_VERSION}.cmake")
-elseif(DEFINED ENV{CPM_SOURCE_CACHE})
-    set(CPM_DOWNLOAD_LOCATION "$ENV{CPM_SOURCE_CACHE}/cpm/CPM_${CPM_DOWNLOAD_VERSION}.cmake")
-else()
-    set(CPM_DOWNLOAD_LOCATION "${PROJECT_BINARY_DIR}/cmake/CPM_${CPM_DOWNLOAD_VERSION}.cmake")
-endif()
+# Always Require the CMake option, but provide default
+set(CPM_SOURCE_CACHE "${CMAKE_SOURCE_DIR}/.cpmcache" CACHE STRING "Path to CPM source cache" FORCE)
```
afuller-TT, line 9: I thought we wanted to _stop_ doing this, not force it.

blozano-tt: ChatGPT told me that a user can override this with `-DCPM_SOURCE_CACHE=/valhalla`

afuller-TT: They can if you change it to this:

```suggestion
set(CPM_SOURCE_CACHE "${CMAKE_SOURCE_DIR}/.cpmcache" CACHE STRING "Path to CPM source cache")
```

cmake/CPM.cmake
```diff
@@ -2,16 +2,13 @@
 #
```
broskoTT, line 2: Seems like the build fails on tt-metal...

blozano-tt: The problem is that both repos need to move forward to the better way together.
However, @afuller-TT has thought of a way that we can do it stepwise.

afuller-TT: https://github.com/tenstorrent/tt-umd/pull/498

Just need to wait for the next time TT-UMD comes in to TT-Metalium for the next phase of the transition.

broskoTT: Yeah please don't merge this PR if tt-metal build fails, if you don't have tt-metal PR ready as well, since you'll block further UMD API development.

blozano-tt:
@broskoTT @afuller-TT 
I am closing this PR as we are decomposing it into smaller steps/PRs. Part of this pr is already merged. 

@broskoTT when is next UMD bump in tt-metal?

broskoTT:
We're occasionally bumping UMD as part of uplifting new API changes... So I'd say this is done a couple of times per month in general

broskoTT:
@blozano-tt also you can do it anytime you want, we don't have any special procedure, or cadence, etc. Usually the person who needs something from the new UMD bumps it

broskoTT (approved):

Issue #485, opened 2025-01-27
BOS UMD spec


broskoTT:
A push to write comments and provide additional info if needed on UMDs API.
Figure out which entry points are needed from tt-metal's perspective.

PR #496, opened 2025-01-30
Docs for UMD implementation guide

broskoTT:
### Issue
#485 

### List of the changes
- Added SOC descriptor guide
- Refined tt_device explanation
- Added custom device UMD guide

### Testing
/

### API Changes


ygiladTT:
DRAM controller port - we need to specify that in TT silicon , DDR controllers are placed on specific grid coordiantes within our NoC . in a customer's SoC , DDR may not be on the grid, however the interface to external memory would be ( NoC2AXI in the Tensix IP case)

- I don't understand, how would you then address this DDR? Is this something outside NoC addresses?

-> They will address NoC2AXI and they would have mapping on the AXI side to DDR.

 what is "traTic" nad  "diTerent" ?

- Sorry, didn't see, this was actually copied from pdf, so some stuff didn't copy well, will fix
-> thanks!

 remove refernece to Buda
- Actually did, but I missed a couple. Will fix
-> Thanks!

There could be several other components on the Tensix IP NoC : Router only, Dispatch engines, Tensix , NoC2AXI

- I'm not aware of all of these, we don't support them currently with our tt_SocDescriptor class. Which line does this reference? Other types of cores are mentined on line 31.

-> in tt-metal you need to speocify which cores are going to be acting as dispatch cores ,and router only , :
router_only (there are examples in the blackhole/wormhole soc descriptors)
  [
   0-2, 0-4, 0-8, 0-9
  ]
 dispatch_cores:
        [[0, 0], [0, 1], [0, 2], [0, 3], [0, 4], [0, 5], [0, 6], [0, 7], [0, 8], [0, 9], [0, 10], [0, 11], [0, 12], [0, 13]]

 please explain how to section nOC2AXI into channels/banks , i.e the differnce between [0-1, 1-1] to [0-1] [1-1] , and other variants that the user can play with

- You mean for dram_cores? I can explain this and other fields better.
-> yes, an explanation on how to set and partition the access to DRAM ( how to set channels, how to set banks, etc..)
 of course, the customer would have to put NOC2AXI items in those coordiantes
 
for L1 Memory size: can you please list the options per architecture ?

- What do you mean, it is clearly defined in each of the soc yamls (grayskull, wh, bh yamls all have a value there)?
-> right, if we can copy paste the possible values, into the guide, or at least link to examples (blackhole device yaml example, etc...). that would also help

  for architecture names : could you please list the legal values?
- will fix
-> Thanks!

changes that are needed for get_platform_architecture() and tt_cluster.cpp are not clear

- I really don't know how to be clearer. For get_platform_architecture, I wrote that a suggestion is for it just to be reimplemented to return any hardcoded architecture.For tt_cluser, I'll add a link, but I have no idea how the constructor will look like. It depends on that.

- -> yes, but the customer does not know what 'architecture' values are legal . please link to an example if not explicitly write the possible values in comments

 The hyper link in this line is merged into one hyper link

- Yeah this was intentional, both functions are one right above the other, is this really that important to separate?
-> not important, more cosmetic...
 
- see that there is an 'offset' vairable
Ah right good catch. I forgot that tt_metal actually has their own additional fields in the soc descriptor. Will explain that.
-> Awesome! thanks ! 

how to fetch the soc_descriptor in the constructor

- You mean how to create tt_SocDescriptor class in the constructor?
-> yes, it seems to be a one line reference..

same for getting and setting the address parameters for barriers

- You mean that set_membar_flag can be called? I can reference that they can look at current implementation...
->sure that would be helpful 

I think we need to add a description of what TLBs are

- This is one of the things that will soon go away probably, metal doesn't need to call this right now. I think it is complex to explain it properly. Does this customer even have TLBs ? I'm not sure which parts of our design they have and which they don't...

-> No customers don't have TLBs, best to just explain that this should be ignored by non TT silicon  and tt-metal 

 SOFT_RESET register address might change, please remove reference to specific register address, instead instruct to look at the tensixx docs for specific register address

- Yeah I have no idea where to find that, will try to figure out.
-> Sorry I think that this could be found in the device specific tensix.h file (in tt-metal), but also in the docs that we provide when we ship the IP , IP customers should have this from us one way or another

 I think that there are typos in device.h

- will fix
-> Thanks!

- we have 2 implementations
Well... this is in the middle of two phase API change. Today only one has to be implemented. In the following weeks we'll switch to the other implementation. Then only the other has to be implemented, and the first one will be removed. Not sure what is the right call here. Did you already freeze tt-metal commit which they'll build upon?
-> not sure what version of tt-metal would be used, certainly it would be a tt-metal release (not simply 'main') that is compatible with your new UMD 

- since there are functions that pertain to 'harvesting'
But they won't have harvesting? They don't need to know it? And we're planning on removing those functions altogether.
 can you show a default implementation
 -> They don't know what harvesting is , so they don't know if they have it or not... I figured that UMD would be the first time that they would encounter this term and so it needs to be explained a little . BOS already knows what it is , so we can keep it as is for now.
 
What do you mean, there is already some simple implementation in cluster.cpp, right? That is one big example?
can you say something like,  'check the reference implementation in device.cpp. ?

ygiladTT:
SOFT_RESET register address might change
Do you mean with new generation, or specifically for customer? How is this not fixed?

-> This might change or might not change, but registers might move around as HW adds / removes features.
They would provide a c header file with the updated address which gets propagated to tt-metal ,and also , we would provide the new address map via IP documentation

docs/SOC_DESCRIPTOR.md
```diff
@@ -0,0 +1,39 @@
...
+A guide on writing up a custom SoC descriptor:
+1. Set the proper grid size with X/Y coordinates (including NoC2AXI coordinates).
+2. Leave pcie/eth/arc/harvested_workers/routers_only empty, as these are SoC specific
+resources configured in a very specific way in our silicon devices.
+3. Define the grid of Tensix Cores in the functional_workers section.
+4. In the DRAM section, set the coordinates of your NoC2AXI components.
+5. Set the dram_bank_size to the desired size to be allocated on the device, where the software stack
+can store input command/data queues, output command/data queues, epoch data/command binaries
+queues, etc.
+6. The memory size of a singular tensix is defined in worker_l1_size.
```
broskoTT, line 38: Copying previous conversation:

for L1 Memory size: can you please list the options per architecture ?

> What do you mean, it is clearly defined in each of the soc yamls (grayskull, wh, bh yamls all have a value there)?
> > right, if we can copy paste the possible values, into the guide, or at least link to examples (blackhole device yaml example, etc...). that would also help

broskoTT: @ygiladTT I already added a folder with example SoC descirptors. It looks pretty clear for me that one can look inside for examples on any of the fields, do you agree?

device/api/umd/device/cluster.h
```diff
@@ -198,8 +230,8 @@ class tt_device {
     }
 
     /**
-     * Send a soft assert reset signal to a single tensix core.
-     *
+     * Send a BRISC soft assert reset signal to a single tensix core.
+     * It writes to TENSIX register 0xFFB121B0 to assert the soft reset.
```
broskoTT, line 234: Copying previous conversation:

SOFT_RESET register address might change, please remove reference to specific register address, instead instruct to look at the tensixx docs for specific register address

> Yeah I have no idea where to find that, will try to figure out.
> > Sorry I think that this could be found in the device specific tensix.h file (in tt-metal), but also in the docs that we provide when we ship the IP , IP customers should have this from us one way or another

broskoTT: @ygiladTT Let's figure this out, as far as I see this is a constant value.
All values/addresses which are constant, or constant per architecture are defined in *_implementation.h files, such as device/api/umd/device/wormhole_implementation.h

docs/CUSTOM_MODIFICATION_GUIDE.md
```diff
@@ -0,0 +1,57 @@
...
+with the class derived from tt_device. For instructions on what each of the functions represents and how to implement them, please see [cluster.h](../device/api/umd/device/cluster.h).
+
+#### soc_descriptor in tt_metal
+
+Soc descriptors used from tt_metal have additional fields which hold additional dram parameters related to through which ports and with which offsets it is accessed. You can just leave it with default values, [here's an example](https://github.com/tenstorrent/tt-metal/blob/783d35aea7df03f9927934ec7b9a640a8032d371/tt_metal/soc_descriptors/blackhole_simulation_1x2_arch.yaml#L14C1-L21C8).
+
+#### config_tlb
+
+[Configuring static TLBs](https://github.com/tenstorrent/tt-metal/blob/main/tt_metal/llrt/tt_cluster.cpp#L295) should be skipped in tt_metal.
+
```
broskoTT, line 48: core_descriptor.yaml, how it should be changed

docs/SOC_DESCRIPTOR.md
```diff
@@ -0,0 +1,39 @@
...
+changed to adapt the software to work on your Tensix IP configuration.
+
+## SoC Descriptor explanation
+
+The SoC descriptor file tells UMD what resources it could provide for a given workload on a particular SoC. Example SoC descriptors
+can be found in [soc_descs folder](../tests/soc_descs/).
+
+A guide on writing up a custom SoC descriptor:
+1. Set the proper grid size with X/Y coordinates (including NoC2AXI coordinates).
+2. Leave pcie/eth/arc/harvested_workers/routers_only empty, as these are SoC specific
```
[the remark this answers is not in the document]
ygiladTT: will talk to the runtime team to get this documented properly

device/api/umd/device/cluster.h
broskoTT: will rewrite this not to include hardcoded address

ygiladTT: Yuval will resolve this

docs/SOC_DESCRIPTOR.md
broskoTT: agreed during offline discussion

docs/SOC_DESCRIPTOR.md
```diff
@@ -0,0 +1,39 @@
...
+The Tenstorrent (TT) software stack was intended to be used with a variety of Tenstorrent silicon devices, e.g.,
+Grayskull™, Wormhole™, etc. All of these devices have a particular, grid-like SoC architecture that the software
+was tailored towards.
+
+(TBD: insert image)
+
+More specifically, here are some (but not all) of that assumptions that were considered when developing our
+software stacks:
+1. The SoC is built in a grid configuration, where each coordinate could be associated with a particular SoC
+function, e.g., coordinate 0,0 is a Tensix Compute Core, coordinate 7,3 is a DRAM controller port, etc.
```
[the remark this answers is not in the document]
broskoTT: I explained DRAM channels subchannels slightly better on line 32

docs/CUSTOM_MODIFICATION_GUIDE.md
broskoTT: Added some minor description of the core_descriptor yaml structure

docs/SOC_DESCRIPTOR.md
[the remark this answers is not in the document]
ygiladTT: since this is not a part of UMD, we will ommit this from this effort

ygiladTT (approved):
looks good

ygiladTT (approved):
looks good

Issue #480, opened 2025-01-23
[Topology] Wait for ETH training to finish


pjanevskiTT:
When discovering topology, we should wait for ETH training to finish before reading any additional information

pjanevskiTT:
Closed by #497

PR #497, opened 2025-01-31
Blackhole topology

pjanevskiTT:
### Issue

#459 
#480 

### Description

Implement basic blackhole topology discovery. It implements discovery over ETH channels for all chips that are connected to PCIE of the host where UMD is running. There will be more issues added to https://github.com/tenstorrent/tt-umd/issues/459 regarding topology discovery

### List of the changes

- Add topology discovery of chips in source
- Wait for ARC cores of the chips to boot
- Wait for ETH to finish training
- 

### Testing

- [x] Blackhole tests 

### API Changes
/

### TODO

- [x] Address comments from #459 
- [x] Follow up issue for code improvements https://github.com/tenstorrent/tt-umd/issues/521

### DIFF
--- a/device/api/umd/device/cluster.h
+++ b/device/api/umd/device/cluster.h
@@ -1010,6 +1010,7 @@ public:
     virtual void dram_membar(
         const chip_id_t chip, const std::unordered_set<tt::umd::CoreCoord>& cores, const std::string& fallback_tlb);
 
+    static std::unique_ptr<tt_ClusterDescriptor> create_cluster_descriptor();
     // Destructor
     virtual ~Cluster();
 
@@ -1198,6 +1199,9 @@ private:
     // Most of the old APIs accept virtual coordinates, but we communicate with the device through translated
     // coordinates. This is an internal helper function, until we switch the API to accept translated coordinates.
     tt_xy_pair translate_chip_coord_virtual_to_translated(const chip_id_t chip_id, const tt_xy_pair core) const;
+
+    static std::unique_ptr<tt_ClusterDescriptor> create_cluster_descriptor(
+        const std::unordered_map<chip_id_t, std::unique_ptr<tt::umd::Chip>>& chips);
 
     // State variables
     std::vector<tt::ARCH> archs_in_cluster = {};
--- a/device/chip/local_chip.cpp
+++ b/device/chip/local_chip.cpp
@@ -6,13 +6,38 @@
 
 #include "umd/device/chip/local_chip.h"
 
+#include "logger.hpp"
 #include "umd/device/tt_device/tlb_manager.h"
 #include "umd/device/tt_device/tt_device.h"
+#include "umd/device/types/blackhole_eth.h"
 
 namespace tt::umd {
 
 LocalChip::LocalChip(tt_SocDescriptor soc_descriptor, int pci_device_id) :
     Chip(soc_descriptor), tt_device_(TTDevice::create(pci_device_id)) {
+    initialize_local_chip();
+}
+
+LocalChip::LocalChip(std::unique_ptr<TTDevice> tt_device) :
+    Chip(
+        tt_device->get_chip_info(),
+        tt_SocDescriptor(
+            tt_SocDescriptor::get_soc_descriptor_path(
+                tt_device->get_arch(),
+                tt_device->get_chip_info().board_type,
+                tt_device->get_chip_info().chip_uid.asic_location),
+            tt_device->get_chip_info().noc_translation_enabled,
+            tt_device->get_chip_info().harvesting_masks)),
+    tt_device_(std::move(tt_device)) {
+    initialize_local_chip();
+}
+
+void LocalChip::initialize_local_chip() {
+    initialize_tlb_manager();
+    wait_chip_to_be_ready();
+}
+
+void LocalChip::initialize_tlb_manager() {
     auto tlb_manager = tt_device_->get_tlb_manager();
     // Setup default dynamic tlbs.
     tlb_manager->set_dynamic_tlb_config(
@@ -27,5 +52,50 @@ LocalChip::LocalChip(tt_SocDescriptor soc_descriptor, int pci_device_id) :
 TTDevice* LocalChip::get_tt_device() { return tt_device_.get(); }
 
 bool LocalChip::is_mmio_capable() const { return true; }
+
+void LocalChip::wait_eth_cores_training(const uint32_t timeout_per_core_ms) {
+    if (get_tt_device()->get_arch() != tt::ARCH::BLACKHOLE) {
+        return;
+    }
+
+    const std::vector<CoreCoord> eth_cores = get_soc_descriptor().get_cores(CoreType::ETH);
+    for (const CoreCoord& eth_core : eth_cores) {
+        TTDevice* tt_device = get_tt_device();
+
+        const tt_xy_pair eth_core_pair = {eth_core.x, eth_core.y};
+
+        uint32_t postcode;
+        auto start = std::chrono::system_clock::now();
+        while (true) {
+            tt_device->read_from_device(&postcode, eth_core_pair, blackhole::BOOT_RESULTS_ADDR, sizeof(postcode));
+
+            if (postcode == blackhole::POSTCODE_ETH_INIT_PASS) {
+                return;
+            }
+
+            if (postcode == blackhole::POSTCODE_ETH_INIT_FAIL) {
+                // TODO: Exception should be thrown here. ETH connections are very flaky
+                // on Blackhole right now. When this is fixed we can throw the exception here.
+                // Since we are not going to do any remote IO at the moment it is fine to just log the error.
+                log_error("Eth core ({}, {}) failed to initialize", eth_core_pair.x, eth_core_pair.y);
+                return;
+            }
+
+            auto end = std::chrono::system_clock::now();
+            auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
+            if (duration.count() > timeout_per_core_ms) {
+                // TODO: Exception should be thrown here. ETH connections are very flaky
+                // on Blackhole right now. When this is fixed we can throw the exception here.
+                // Since we are not going to do any remote IO at the moment it is fine to just log the error.
+                log_error(
+                    "Timed out after waiting {} seconds for eth core ({}, {}) to initialize",
+                    timeout_per_core_ms,
+                    eth_core_pair.x,
+                    eth_core_pair.y);
+                return;
+            }
+        }
+    }
+}
 
 }  // namespace tt::umd
--- a/device/chip/local_chip.cpp
+++ b/device/chip/local_chip.cpp
@@ -53,24 +53,24 @@ TTDevice* LocalChip::get_tt_device() { return tt_device_.get(); }
 
 bool LocalChip::is_mmio_capable() const { return true; }
 
-void LocalChip::wait_eth_cores_training(const uint32_t timeout_per_core_ms) {
+void LocalChip::wait_eth_cores_training(const uint32_t timeout_ms) {
     if (get_tt_device()->get_arch() != tt::ARCH::BLACKHOLE) {
         return;
     }
 
     const std::vector<CoreCoord> eth_cores = get_soc_descriptor().get_cores(CoreType::ETH);
+    TTDevice* tt_device = get_tt_device();
+    auto start = std::chrono::system_clock::now();
     for (const CoreCoord& eth_core : eth_cores) {
-        TTDevice* tt_device = get_tt_device();
-
         const tt_xy_pair eth_core_pair = {eth_core.x, eth_core.y};
 
         uint32_t postcode;
-        auto start = std::chrono::system_clock::now();
+
         while (true) {
             tt_device->read_from_device(&postcode, eth_core_pair, blackhole::BOOT_RESULTS_ADDR, sizeof(postcode));
 
             if (postcode == blackhole::POSTCODE_ETH_INIT_PASS) {
-                return;
+                break;
             }
 
             if (postcode == blackhole::POSTCODE_ETH_INIT_FAIL) {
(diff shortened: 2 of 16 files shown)


device/cluster.cpp
```diff
@@ -3431,4 +3432,123 @@ tt_xy_pair Cluster::translate_to_api_coords(const chip_id_t chip, const tt::umd:
...
+    if (pci_device_info.begin()->second.get_arch() == tt::ARCH::BLACKHOLE) {
+        std::vector<int> pci_device_ids = PCIDevice::enumerate_devices();
+
+        std::vector<std::unique_ptr<TTDevice>> tt_devices;
+        for (auto& device_id : pci_device_ids) {
+            std::unique_ptr<TTDevice> tt_device = TTDevice::create(device_id);
+            tt_devices.push_back(std::move(tt_device));
+        }
+
+        std::vector<ChipInfo> chip_info_vec = Cluster::get_cluster_chip_info(tt_devices);
```
broskoTT, line 3463: I don't understand why do you need three loops, or why do you need get_cluster_chip_info function.
To me it looks like the code would be simpler with a single for loop. Something like:

```
std::unordered_map<chip_id_t, std::unique_ptr<Chip>> chips;
for (auto& device_id : pci_device_ids) {
  std::unique_ptr<TTDevice> tt_device = TTDevice::create(device_id);
  tt_device->wait_arc_core_start(arc_core);
  ChipInfo chip_info = tt_device->get_chip_info();
  chips.emplace(chip_id, std::move(tt_device), chip_info));
}
```

device/cluster.cpp
```diff
@@ -3431,4 +3432,123 @@ tt_xy_pair Cluster::translate_to_api_coords(const chip_id_t chip, const tt::umd:
...
+
+        std::unordered_map<chip_id_t, std::unique_ptr<Chip>> chips;
```
broskoTT, line 3465: Why is not all of this code part of chip class? Chip class can fetch chip_info from its underlying tt_device if needed (currently, if blackhole) ?
That would also remove additional constructors in Chip and LocalChip classes

broskoTT: To be specific here's how get_cluster_descriptor would look like in that case:
```
std::unique_ptr<tt_ClusterDescriptor> Cluster::get_cluster_descriptor() {
  std::map<int, PciDeviceInfo> pci_device_info = PCIDevice::enumerate_devices_info();
    if (pci_device_info.begin()->second.get_arch() == tt::ARCH::BLACKHOLE) {
        std::unordered_map<chip_id_t, std::unique_ptr<Chip>> chips;
        for (auto& device_id : pci_device_ids) {
          chips.emplace(chip_id, soc_desc, device_id)); // chip created in place in the map, with proper soc_desc
        }
        return Cluster::create_cluster_descriptor(chips);
      }
...
```

device/tt_cluster_descriptor.cpp
```diff
@@ -894,3 +894,5 @@ const std::unordered_map<chip_id_t, std::unordered_set<chip_id_t>> &
 tt_ClusterDescriptor::get_chips_grouped_by_closest_mmio() const {
     return chips_grouped_by_closest_mmio;
 }
+
+chip_id_t tt_ClusterDescriptor::get_chip_id(const ChipUID &chip_uid) const { return chip_uid_to_chip_id.at(chip_uid); }
```
broskoTT, line 898: This is added interactivelly by reading chip information in tt::Cluster. But we'd also like populating this field from yaml as well?

I imagine the scenario for tt-fabric folks might be to fetch cluster desc and serialize it somewhere and then pass it on the next run to create Cluster from it?

device/cluster.cpp
```diff
@@ -3431,4 +3432,123 @@ tt_xy_pair Cluster::translate_to_api_coords(const chip_id_t chip, const tt::umd:
...
+    for (auto& tt_device : tt_devices) {
+        tt_device->wait_arc_core_start(arc_core);
+        chip_info_vec.push_back(tt_device->get_chip_info());
+    }
+
+    return chip_info_vec;
+}
+
+std::unique_ptr<tt_ClusterDescriptor> Cluster::create_cluster_descriptor(
+    const std::unordered_map<chip_id_t, std::unique_ptr<tt::umd::Chip>>& chips) {
```
broskoTT, line 3497: Are we really able to create_cluster_descriptor from a collection of Chips, or from a collection of LocalChips? Think about what is going to be needed in the future

device/cluster.cpp
```diff
@@ -3448,4 +3449,123 @@ tt_xy_pair Cluster::translate_chip_coord_virtual_to_translated(const chip_id_t c
...
+        desc->harvesting_masks.insert({chip_id, chip->get_chip_info().harvesting_masks.tensix_harvesting_mask});
+
+        const std::vector<CoreCoord> eth_cores = chip->get_soc_descriptor().get_cores(CoreType::ETH);
+
+        for (size_t eth_channel = 0; eth_channel < eth_cores.size(); eth_channel++) {
+            const CoreCoord& eth_core = eth_cores[eth_channel];
+            TTDevice* tt_device = chip->get_tt_device();
+            boot_results_t boot_results;
+
+            tt_device->read_from_device(
```
broskoTT, line 3529: Everything bellow up to the end of the for body loop can be a function "get_eth_channel_status" right on the TTDevice.

Not sure if it would need some arguments though, looks like not. blackhole::BOOT_RESULTS_ADDR would be hardcoded in BlackholeTTDevice instead of here. You'll need to alter the debug messages a bit though. Actually it would accept just the eth_core

Then the code here would look something like this:
```
port_status_e status = tt_device->get_eth_channel_status();
if (status == port_status_e::PORT_UP) {
  desc->ethernet_connections[local_chip_id][local_info.eth_id] = {remote_chip_id, remote_info.eth_id};
} else {
  log_debug (LogSiliconDriver, "Port on eth core ({}, {}) on chip {} is not up {}", eth_core.x, eth_core.y, chip_id, status);
}
```

device/cluster.cpp
```diff
@@ -3448,4 +3449,123 @@ tt_xy_pair Cluster::translate_chip_coord_virtual_to_translated(const chip_id_t c
...
+
+        desc->all_chips.insert(chip_id);
+        desc->chip_arch.insert({chip_id, chip->get_tt_device()->get_arch()});
+
+        desc->chips_with_mmio.insert({chip_id, chip->get_tt_device()->get_pci_device()->get_device_num()});
+
+        desc->chip_board_type.insert({chip_id, chip->get_chip_info().board_type});
+
+        desc->noc_translation_enabled.insert({chip_id, chip->get_chip_info().noc_translation_enabled});
+        desc->harvesting_masks.insert({chip_id, chip->get_chip_info().harvesting_masks.tensix_harvesting_mask});
```
broskoTT, line 3520: Consider creating a function on tt_ClusterDescriptor which would accept all of these values from the right hand side, and put them in the right maps. Also if you accept the suggestion bellow you might add remote_chip_id info as well.

```
tt_ClusterDescriptor::add_chip_to_collection(int chip_id, tt::ARCH arch, int mmio = -1, BoardType board, bool translation, int harvesting, std::map<int, std::pair<int,int>> eth_channel_to_remote_mapping)
```

broskoTT (commented):
Thanks for separating out the PRs, it is much easier to focus like this. Let me know if you have uncertainties about my comments, hope they'd lead to more readable code and cleaner API

device/tt_device/blackhole_tt_device.cpp
```diff
@@ -111,4 +111,24 @@ ChipInfo BlackholeTTDevice::get_chip_info() {
     return chip_info;
 }
 
+void BlackholeTTDevice::wait_arc_core_start(const tt_xy_pair arc_core, const uint32_t timeout_ms) {
+    auto start = std::chrono::system_clock::now();
+    uint32_t arc_boot_status;
+    while (true) {
+        read_from_device(&arc_boot_status, arc_core, tt::umd::blackhole::SCRATCH_RAM_2, sizeof(arc_boot_status));
```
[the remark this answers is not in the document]
pjanevskiTT: Not sure how fast ARC core boots, I would leave this as it is, investigate how fast ARC core boots. If it is slow, I can add timeouts, but from running this I think ARC core boots very fast

device/tt_cluster_descriptor.cpp
pjanevskiTT: Yup, ChipUID is currently not provided for Wormhole devices, at least we don't construct it. When we rewrite bigger chunk of create-ethernet-map this will become available for all chip architectures. I think we should keep it and then populate it for all architectures when it becomes available

device/cluster.cpp
pjanevskiTT: It is now

device/cluster.cpp
pjanevskiTT: I would keep Chips here. In general our future configs are all going to have only LocalChips for one UMD, but it would be nice to generalize it for old WH architectures so sometimes in the future RemoteChip could be in the map as well

device/cluster.cpp
pjanevskiTT: I will post this in an issue to think about it when refining cluster descriptor API

pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/510

device/cluster.cpp
broskoTT: You're right about boot_results_t struct.
But I'd still argue that the details of reading this info is part of tt_device rather than Chip (in the future).
So I'd move this read_from_device to tt_device, and here we would have 
```
boot_results_t boot_results = tt_device->read_eth_channel_info(eth_core);
```

device/chip/local_chip.cpp
```diff
@@ -28,4 +53,49 @@ TTDevice* LocalChip::get_tt_device() { return tt_device_.get(); }
...
+    const std::vector<CoreCoord> eth_cores = get_soc_descriptor().get_cores(CoreType::ETH);
+    for (const CoreCoord& eth_core : eth_cores) {
+        TTDevice* tt_device = get_tt_device();
+
+        const tt_xy_pair eth_core_pair = {eth_core.x, eth_core.y};
+
+        uint32_t postcode;
+        auto start = std::chrono::system_clock::now();
+        while (true) {
+            tt_device->read_from_device(&postcode, eth_core_pair, blackhole::BOOT_RESULTS_ADDR, sizeof(postcode));
```
broskoTT, line 70: Similar comment to the other one. 
Let's have a function for this, and not bother "chip" class with read details:
```
uint32_t postcode = tt_device->read_eth_boot_postcode();
```

Or consider even pushing the whole logic with timeout
```
bool success = tt_device->wait_for_ethcore_to_be_ready(timeout)
```

device/chip/local_chip.cpp
```diff
@@ -28,4 +53,49 @@ TTDevice* LocalChip::get_tt_device() { return tt_device_.get(); }
...
+                // TODO: Exception should be thrown here. ETH connections are very flaky
+                // on Blackhole right now. When this is fixed we can throw the exception here.
+                // Since we are not going to do any remote IO at the moment it is fine to just log the error.
+                log_error("Eth core ({}, {}) failed to initialize", eth_core_pair.x, eth_core_pair.y);
+                return;
+            }
+
+            auto end = std::chrono::system_clock::now();
+            auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
+            if (duration.count() > timeout_per_core_ms) {
```
broskoTT, line 86: So the first core will have this value of miliseconds to train, but every next will have more? E.g. in the worst case the 5th core would have 5* timeout_per_core_ms to train?

I'd find it a bit cleaner if we had a timeout for the whole train process. The suggestion is to put `start` initialization out of the for loop, and change the variable name to `timeout_ms`

device/tt_device/blackhole_tt_device.cpp
```diff
@@ -80,4 +83,52 @@ void BlackholeTTDevice::configure_iatu_region(size_t region, uint64_t base, uint
...
+
+    return chip_info;
+}
+
+void BlackholeTTDevice::wait_arc_core_start(const tt_xy_pair arc_core, const uint32_t timeout_ms) {
+    auto start = std::chrono::system_clock::now();
+    uint32_t arc_boot_status;
+    while (true) {
+        read_from_device(&arc_boot_status, arc_core, tt::umd::blackhole::SCRATCH_RAM_2, sizeof(arc_boot_status));
+
```
broskoTT, line 119: Same comment as the other places, consider creating a function in tt_device called "read_arc_boot_status" or "wait_for_arc_to_boot".

My idea in general for tt_device, is for it to hold as much  magic hardware and firmware numbers as possible. It should offer many simple descriptive functions.

broskoTT (approved):
Thanks looks much better now!

pjanevskiTT:
Don't merge this until it is confirmed how noc_translation info should be read

device/tt_device/blackhole_tt_device.cpp
```diff
@@ -100,15 +100,45 @@ ChipInfo BlackholeTTDevice::get_chip_info() {
 
     // For now, NOC translation is disabled on all Blackhole boards.
     // TODO: read this information when it becomes available.
-    chip_info.noc_translation_enabled = false;
+
+    const uint64_t niu_cfg_addr = 0x1000A0000 + 0x100;
```
pjanevskiTT, line 104: Don't merge this until it is confirmed how noc_translation info should be read

device/tt_device/blackhole_tt_device.cpp
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/521

device/chip/local_chip.cpp
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/521

device/tt_device/blackhole_tt_device.cpp
```diff
@@ -80,4 +83,52 @@ void BlackholeTTDevice::configure_iatu_region(size_t region, uint64_t base, uint
...
+    // It is expected that these entries are always available.
+    chip_info.chip_uid.board_id = ((uint64_t)telemetry->read_entry(blackhole::TAG_BOARD_ID_HIGH) << 32) |
+                                  (telemetry->read_entry(blackhole::TAG_BOARD_ID_LOW));
+
+    chip_info.board_type = get_board_type_from_board_id(chip_info.chip_uid.board_id);
+
+    return chip_info;
+}
+
+void BlackholeTTDevice::wait_arc_core_start(const tt_xy_pair arc_core, const uint32_t timeout_ms) {
```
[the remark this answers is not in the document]
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/521

device/cluster.cpp
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/521

device/chip/local_chip.cpp
pjanevskiTT: Changed it to timeout for whole training process

device/tt_device/blackhole_tt_device.cpp
pjanevskiTT: Proper way to do this

`PCIE BAR, it's at 0x1FD04100 for NOC0, 0x1FD14100 for NOC1`

broskoTT: feel free to merge it anyway, and just change the logic in the next PR...

device/api/umd/device/cluster.h
```diff
@@ -1010,6 +1010,7 @@ class Cluster : public tt_device {
     virtual void dram_membar(
         const chip_id_t chip, const std::unordered_set<tt::umd::CoreCoord>& cores, const std::string& fallback_tlb);
 
+    static std::unique_ptr<tt_ClusterDescriptor> create_cluster_descriptor();
```
broskoTT, line 1013: Sorry to raise this now, but can we update in api examples cluster_descriptor_test.cpp the current suggested way to create this object?

device/tt_device/blackhole_tt_device.cpp
pjanevskiTT: Changed it, reading it through PCI bar should be the only proper way to do it

device/api/umd/device/cluster.h
pjanevskiTT: I'll add it as a follow up to this PR

Issue #431, opened 2024-12-20
Metal SW requests for UMD cluster descriptor


aliuTT:
WIP
Specs from Almeet, aligned with syseng as well: [ethernet for Metal](https://tenstorrent.sharepoint.com/:w:/r/sites/Jasmina/_layouts/15/Doc.aspx?sourcedoc=%7B940DD52D-30F9-442A-B24A-7944A7D8031D%7D&file=Ethernet%20for%20Metal.docx&action=default&mobileredirect=true), 

List of APIs that Metal SW will need from the cluster descriptor. I've structured the APIs into a table, because in general it looks like features for BH ethernet FW will not be ported to WH (unless we make a request and can get new headcount for syseng ).

Question is `ethernet_channel_t` a logical 1D ethernet id? If so then I think the `get_ethernet_connections()` could return an array of arrays where the first index is device id and second is logical eth id 

| API                                             | Functionality | Wormhole    | Blackhole |
| ----------------------------- | ------- | -----------| ---------|
| `std::unordered_map<chip_id_t, std::unordered_map<ethernet_channel_t, std::tuple<chip_id_t, ethernet_channel_t>>> get_ethernet_connections() const;` |  For each eth on a given chip, return the remote chip and eth it is connected to (if we turn this into array of arrays then the tuple could be optional to depict idle eths)  |      |   |
|`const std::unordered_map<chip_id_t, chip_id_t> get_chips_with_mmio() const;` |   if UMD supports non-zero mmio id, this can be deprecated   | |  |
| `bool ethernet_core_has_active_ethernet_link(chip_id_t local_chip, ethernet_channel_t local_ethernet_channel) const;` |  specifies whether a given eth is active. If the eth is active but it failed training would it still return `true`? If this queries whether link is successfully trained then this should throw if `local_ethernet_channel` is not an active eth  | |  |
|`std::tuple<chip_id_t, ethernet_channel_t> get_chip_and_channel_of_remote_ethernet_core(chip_id_t local_chip, ethernet_channel_t local_ethernet_channel) const;` |  queries map returned by `get_ethernet_connections`  | |  |

abhullar-tt:
I think we need to have one API return a collection of active ethernets (includes links that are down and successfully trained) and then another API to check if the link is active

pjanevskiTT:
[...]

PR #499, opened 2025-02-03
Serialize cluster into yaml

pjanevskiTT:
### Issue

Solving part of #431 

### Description

Add a function to serialize cluster to the yaml (reverse of what we were doing so far). Add flag to all Cluster constructors to create mock chips inside Cluster if user wants that. Expose this (de)serializing through the API as mentioned in the issue above. Output yaml look like (note the styling difference than original, but in YAML world it is the same)

```
arch:
  1: wormhole_b0
  0: wormhole_b0
ethernet_connections:
  -
    - chip: 1
      chan: 1
    - chip: 0
      chan: 9
  -
    - chip: 1
      chan: 0
    - chip: 0
      chan: 8
chips_with_mmio:
  - 0: 0
harvesting:
  1:
    noc_translation: true
    harvest_mask: 513
  0:
    noc_translation: true
    harvest_mask: 129
boardtype:
  1: n300
  0: n300 
```

### List of the changes

- Add serialize function to Cluster (public)
- Add serialize function cluste descriptor (not public)
- Add flag to all Cluster constructors to create mock chips
- Add a test

### Testing
CI + new test

### API Changes

Added create_mock_chips to Cluster constuctors, but changes not needed in UMD clients.

### DIFF
--- a/device/api/umd/device/cluster.h
+++ b/device/api/umd/device/cluster.h
@@ -690,7 +690,7 @@ public:
      * Simplest form, creates a cluster of all available devices on the system.
      *
      * @param num_host_mem_ch_per_mmio_device Requested number of host channels (hugepages).
-     * @param skip_driver_allocs
+     * @param create_mock_chips Create mock chips for the devices in the cluster descriptor.
      * @param clean_system_resource Specifies if host state from previous runs needs to be cleaned up.
      * @param perform_harvesting Allow the driver to modify the SOC descriptors per chip.
      * @param simulated_harvesting_masks Manually specify additional harvesting masks for the devices in the cluster.
@@ -699,11 +699,10 @@ public:
      */
     Cluster(
         const uint32_t& num_host_mem_ch_per_mmio_device = 1,
-        const bool skip_driver_allocs = false,
+        const bool create_mock_chips = false,
         const bool clean_system_resources = false,
         bool perform_harvesting = true,
-        std::unordered_map<chip_id_t, HarvestingMasks> simulated_harvesting_masks = {},
-        const bool create_mock_chips = false);
+        std::unordered_map<chip_id_t, HarvestingMasks> simulated_harvesting_masks = {});
 
     /**
      * Cluster constructor.
@@ -711,21 +710,19 @@ public:
      *
      * @param target_devices Devices to target.
      * @param num_host_mem_ch_per_mmio_device Requested number of host channels (hugepages).
-     * @param skip_driver_allocs
+     * @param create_mock_chips Create mock chips for the devices in the cluster descriptor.
      * @param clean_system_resource Specifies if host state from previous runs needs to be cleaned up.
      * @param perform_harvesting Allow the driver to modify the SOC descriptors per chip.
      * @param simulated_harvesting_masks Manually specify additional harvesting masks for the devices in the cluster.
      * The ones defined by the devices itself have to be used, they will be merged with the ones passed here.
-     * @param create_mock_chips Create mock chips for the devices in the cluster descriptor.
      */
     Cluster(
         const std::set<chip_id_t>& target_devices,
         const uint32_t& num_host_mem_ch_per_mmio_device = 1,
-        const bool skip_driver_allocs = false,
+        const bool create_mock_chips = false,
         const bool clean_system_resources = false,
         bool perform_harvesting = true,
-        std::unordered_map<chip_id_t, HarvestingMasks> simulated_harvesting_masks = {},
-        const bool create_mock_chips = false);
+        std::unordered_map<chip_id_t, HarvestingMasks> simulated_harvesting_masks = {});
 
     /**
      * Cluster constructor.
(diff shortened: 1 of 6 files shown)


device/tt_cluster_descriptor.cpp
```diff
@@ -896,3 +903,75 @@ tt_ClusterDescriptor::get_chips_grouped_by_closest_mmio() const {
...
+    std::filesystem::path temp_path = std::filesystem::temp_directory_path();
+    std::string cluster_path_dir_template = temp_path / "umd_XXXXXX";
+    std::filesystem::path cluster_path_dir = mkdtemp(cluster_path_dir_template.data());
+    std::filesystem::path cluster_path = cluster_path_dir / "cluster_descriptor.yaml";
+    std::ofstream file(cluster_path);
+    file << out.c_str();
+    file.close();
+
+    return cluster_path;
+}
```
cfjchu, line 977: can we separate out serialization from file I/O? 

Something like:

```
std::string tt_ClusterDescriptor::serialize() const {
    YAML::Emitter out;
    ...
    return out.c_str();
}

std::filesystem::path tt_ClusterDescriptor::serialize_to_file() const {
    auto serialized = this->serialize();
    ...
    // write to file and return path

}
```

tests/api/test_cluster.cpp
```diff
@@ -279,3 +272,10 @@ TEST(ClusterAPI, DynamicTLB_RW) {
     }
     cluster->close_device();
 }
+
+TEST(ClusterAPI, TestClusterSerialize) {
+    std::string cluster_path = tt::umd::Cluster::serialize();
+    std::unordered_map<chip_id_t, HarvestingMasks> simulated_harvesting_masks = {};
+    std::unique_ptr<Cluster> cluster = std::make_unique<Cluster>(
+        tt_ClusterDescriptor::create_from_yaml(cluster_path), 1, false, false, true, simulated_harvesting_masks, true);
```
cfjchu, line 280: @pjanevskiTT Thanks! For the creation from yaml path, is the important flag here the `create_mock_chips` boolean? Do you have more information on the mock chips? 

This path:

```
    std::unique_ptr<Cluster> cluster = std::make_unique<Cluster>(
        tt_ClusterDescriptor::create_from_yaml(cluster_path), 1, false, false, true, simulated_harvesting_masks, true);
```
can be used on a machine without a silicon device?

device/api/umd/device/cluster.h
```diff
@@ -758,34 +760,17 @@ class Cluster : public tt_device {
...
-     * TODO: Could fail if logical_ids not match the ones in cluster descriptor, while Cluster still uses cluster
-     * descriptor.
-     *
-     * @param chips Map of logical device ids to Chip instances.
-     * @param num_host_mem_ch_per_mmio_device Requested number of host channels (hugepages).
-     * @param skip_driver_allocs
-     * @param clean_system_resource Specifies if host state from previous runs needs to be cleaned up.
-     * @param perform_harvesting Allow the driver to modify the SOC descriptors per chip.
-     * @param simulated_harvesting_masks
-     */
```
broskoTT, line 776: Leave some comment on the function

tests/api/test_cluster.cpp
```diff
@@ -70,13 +70,6 @@ TEST(ApiClusterTest, DifferentConstructors) {
     std::string sdesc_path = tt_SocDescriptor::get_soc_descriptor_path(device_arch, BoardType::UNKNOWN);
     umd_cluster = std::make_unique<Cluster>(sdesc_path, target_devices);
     umd_cluster = nullptr;
-
-    // TODO: This doesn't work at the moment.
-    // It will start working when we move enough stuff to the chips. At the moment this was disabled, it was mostly due
-    // to harvesting info.
-    // // 4. Constructor for creating a cluster with mock chip.
-    // umd_cluster = Cluster::create_mock_cluster();
-    // umd_cluster = nullptr;
```
broskoTT, line 79: Yeah can you add another test which demonstrates creating Cluster with mock chips?

broskoTT: This is a general comment. I want to make sure we've covered all the scenarios which we want to cover. Maybe that deserves a separate api test (remember, api tests are not a place for testing code, rather for showcasing the usage):

- Creating a cluster, don' care about cluster descriptor
- Forging a custom cluster descriptor and creating a cluster using it
- Forging a custom cluster descriptor and creating a mock cluster using it
- Create a cluster and serialize cluster descriptor yaml so it can be used later, after modification or not, in one of the previous two scenarios.

@cfjchu are we missing some scenario that you folks want to have?

broskoTT (approved):

tests/api/test_cluster.cpp
cfjchu: Yes, I think these scenarios capture it. I would also like the ability to extend `Chip` or `MockChip` and provide implementations for intercepting the interactions that would've gone to an actual chip. 

For example, it would be useful  to have an ability to create a mock 64-chip cluster in an offline context where interactions to first 32-chips get transmitted as RPCs to one host galaxy and interactions with the other 32-chips get transmitted as RPCs to the other galaxy. Something similar to what is already being done with the simulator.

cfjchu: and big +1 for adding more tests that exercise this user-flow

device/tt_cluster_descriptor.cpp
pjanevskiTT: done

tests/api/test_cluster.cpp
pjanevskiTT: The idea behind adding create_mock_chips is that each constructor can be called such that you get mock chips. Mock chips should probably have all the IO functions that interact with device empty. But you can still read all the relevant info of the chip. The things is that all these constructors apart from the one I added need the device to create cluster descriptor.
The call you pasted in the comment is a way to use Cluster on a machine without silicon device

pjanevskiTT: I'll resolve the comment, feel free to write more if you have more questions

device/api/umd/device/types/cluster_descriptor_types.h
```diff
@@ -51,6 +52,33 @@ enum BoardType : uint32_t {
     UNKNOWN,
 };
 
+inline std::string board_type_to_string(const BoardType board_type) {
```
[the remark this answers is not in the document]
broskoTT: it's fine, ignore

tests/api/test_cluster.cpp
pjanevskiTT: I figured that skip_driver_allocs was actually same case as create_mock_chips (don't touch silicon). Renamed that var to create_mock_chips.

Basically your call can now be (note there is no last argument). 

```cpp
    std::unique_ptr<Cluster> cluster = std::make_unique<Cluster>(
        tt_ClusterDescriptor::create_from_yaml(cluster_path), 1, true, false, true, simulated_harvesting_masks);
```

You can take a look at the test as well

pjanevskiTT:
@broskoTT fyi, `skip_driver_allocs` had the same purpose as `create_mock_chips`. I renamed it and fixed few bugs which we had when creating mock chips

tests/api/test_cluster.cpp
broskoTT: @pjanevskiTT sounds like we might benefit from the constructor accepting vector of custom made Chips :) let's put more thought into it in the near future

PR #514, opened 2025-02-06
Minor tt_soc api update

broskoTT:
### Issue
Related to #439 

### Description
Offer custom coord system for get_*_for_channel apis. Remove unused functions

### List of the changes
- Removed is_worker_core and is_ethernet_core. These are unused in tt_metal as well
- Offer custom coord_system in get_dram_core_for_channel and get_eth_core_for_channel for more flexibility.

### Testing
No testing

### API Changes

### DIFF
--- a/device/api/umd/device/cluster.h
+++ b/device/api/umd/device/cluster.h
@@ -512,7 +512,6 @@ public:
      */
     virtual bool using_harvested_soc_descriptors() {
         throw std::runtime_error("---- tt_device:using_harvested_soc_descriptors is not implemented\n");
-        return 0;
     }
 
     /**
@@ -568,6 +567,13 @@ public:
     }
 
     /**
+     * Get set of chip ids for all chips in the cluster.
+     */
+    virtual std::set<chip_id_t> get_target_device_ids() {
+        throw std::runtime_error("---- tt_device::get_target_device_ids is not implemented\n");
+    }
+
+    /**
      * Get all logical ids for all local chips targeted by UMD.
      */
     virtual std::set<chip_id_t> get_target_mmio_device_ids() {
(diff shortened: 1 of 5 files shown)


device/api/umd/device/tt_soc_descriptor.h
```diff
@@ -85,15 +85,13 @@ class tt_SocDescriptor {
...
-
     tt_xy_pair get_core_for_dram_channel(int dram_chan, int subchannel) const;
 
     // Effectively translates between LOGICAL and VIRTUAL coordinates for DRAM and ETH cores.
-    tt::umd::CoreCoord get_dram_core_for_channel(int dram_chan, int subchannel) const;
-    tt::umd::CoreCoord get_eth_core_for_channel(int eth_chan) const;
-
-    bool is_ethernet_core(const tt_xy_pair &core) const;
+    tt::umd::CoreCoord get_dram_core_for_channel(
+        int dram_chan, int subchannel, const CoordSystem coord_system = CoordSystem::VIRTUAL) const;
```
pjanevskiTT, line 92: I think most of our default arguments for CoordinateManager take PHYSICAL as default value for coord_system. I assume you are putting virutal here because that is what we are using across the stack. I think it would be nice to make that explicit when calling this function. I am not pushing to change this default to PHYSICAL, just saying what I think would be more readable

device/api/umd/device/tt_soc_descriptor.h
```diff
@@ -85,15 +85,13 @@ class tt_SocDescriptor {
...
+    tt::umd::CoreCoord get_eth_core_for_channel(
+        int eth_chan, const CoordSystem coord_system = CoordSystem::VIRTUAL) const;
```
pjanevskiTT, line 94: same as above

pjanevskiTT (approved):

device/api/umd/device/tt_soc_descriptor.h
broskoTT: Was wondering myself as well... Changed it to PHYSICAL, and now nicelly added get_coord_system_used() to other .get_cores usages in cluster, technically all should have it...

PR #515, opened 2025-02-06
Fix ubench tests and add to CI

pjanevskiTT:
### Issue

Solving the problem that #305 was trying to solve

### Description

Fix build of ubench tests and add ubench build to CI

### List of the changes

- Fix ubench tests build
- Add ubench tests build to CI
- Remove statically programming TLB before each test. That should happen in separate tests

### Testing

Local run

### API Changes
/

### TODO

- [x] Create follow up issues for ubench tests -  https://github.com/tenstorrent/tt-umd/issues/519

### DIFF
--- a/tests/microbenchmark/device_fixture.hpp
+++ b/tests/microbenchmark/device_fixture.hpp
@@ -31,21 +31,7 @@ protected:
         };
         std::set<chip_id_t> target_devices = {0};
         uint32_t num_host_mem_ch_per_mmio_device = 1;
-        device = std::make_shared<Cluster>(
-            test_utils::GetAbsPath("tests/soc_descs/grayskull_10x12.yaml"),
-            target_devices,
-            num_host_mem_ch_per_mmio_device,
-            false,
-            true);
-
-        for (int i = 0; i < target_devices.size(); i++) {
-            // Iterate over devices and only setup static TLBs for functional worker cores
-            auto& sdesc = device->get_soc_descriptor(i);
-            for (auto& core : sdesc.workers) {
-                // Statically mapping a 1MB TLB to this core, starting from address DATA_BUFFER_SPACE_BASE.
-                device->configure_tlb(i, core, get_static_tlb_index(core), l1_mem::address_map::DATA_BUFFER_SPACE_BASE);
-            }
-        }
+        device = std::make_shared<Cluster>(target_devices);
     }
 
     void TearDown() override {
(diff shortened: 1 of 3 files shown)


tests/microbenchmark/test_rw_tensix.cpp
```diff
@@ -27,22 +23,10 @@ TEST_F(uBenchmarkFixture, WriteAllCores32Bytes) {
 
     ankerl::nanobench::Bench bench_static;
     ankerl::nanobench::Bench bench_dynamic;
-    for (auto& core : device->get_soc_descriptor(0).workers) {
+    for (auto& core_coord : device->get_soc_descriptor(0).get_cores(CoreType::TENSIX)) {
+        tt_xy_pair core = {core_coord.x, core_coord.y};
         std::stringstream wname;
         wname << "Write to device core (" << core.x << ", " << core.y << ")";
-        // Write 32 bytes through static tlbs
```
broskoTT, line 33: So these tests now write only through dynamic tlbs?

tests/microbenchmark/device_fixture.hpp
```diff
@@ -31,21 +31,7 @@ class uBenchmarkFixture : public ::testing::Test {
...
         std::set<chip_id_t> target_devices = {0};
         uint32_t num_host_mem_ch_per_mmio_device = 1;
-        device = std::make_shared<Cluster>(
-            test_utils::GetAbsPath("tests/soc_descs/grayskull_10x12.yaml"),
-            target_devices,
-            num_host_mem_ch_per_mmio_device,
-            false,
-            true);
-
-        for (int i = 0; i < target_devices.size(); i++) {
```
broskoTT, line 41: How is this not important for benchmarking?

tests/microbenchmark/test_rw_tensix.cpp
pjanevskiTT: Yup, common code was setting up the static TLBs, it should be done by the test that wants to use the static TLBs. I will write these tests later when we want the benchmark results

tests/microbenchmark/device_fixture.hpp
pjanevskiTT: Same as above. We don't want to setup static TLBs for every test, just some of the tests, so it will be done in tests which want to use static TLBs

broskoTT (approved):

PR #530, opened 2025-02-21
Add physical <-> translated mapping for all BH pcie cores

abhullar-tt:
### Issue
N/A

### Description
In Metal the BH soc descriptor lists both potential physical PCIe cores (2,0) and (11,0) but later picks which physical (noc0) PCIe core to use for the board type. 

When switching to translated coordinates Metal tests were failing to translate (11,0) because the BH coordinate manager assumes only 1 PCIe core 

### List of the changes
Update physical <-> translated mapping for BH to iterate over all PCIe cores. 

### Testing
- Tested on Metal branch `abhullar/bh-virtual`

### DIFF
--- a/device/blackhole/blackhole_coordinate_manager.cpp
+++ b/device/blackhole/blackhole_coordinate_manager.cpp
@@ -221,17 +221,20 @@ void BlackholeCoordinateManager::fill_eth_physical_translated_mapping() {
 }
 
 void BlackholeCoordinateManager::fill_pcie_physical_translated_mapping() {
-    CoreCoord logical_coord = CoreCoord(0, 0, CoreType::PCIE, CoordSystem::LOGICAL);
+    for (size_t x = 0; x < pcie_grid_size.x; x++) {
+        for (size_t y = 0; y < pcie_grid_size.y; y++) {
+            CoreCoord logical_coord = CoreCoord(x, y, CoreType::PCIE, CoordSystem::LOGICAL);
+            const tt_xy_pair physical_pair = to_physical_map[logical_coord];
 
-    const tt_xy_pair physical_pair = to_physical_map[logical_coord];
+            CoreCoord translated_coord = CoreCoord(
+                blackhole::pcie_translated_coordinate_start_x,
+                blackhole::pcie_translated_coordinate_start_y,
+                CoreType::PCIE,
+                CoordSystem::TRANSLATED);
 
-    CoreCoord translated_coord = CoreCoord(
-        blackhole::pcie_translated_coordinate_start_x,
-        blackhole::pcie_translated_coordinate_start_y,
-        CoreType::PCIE,
-        CoordSystem::TRANSLATED);
-
-    add_core_translation(translated_coord, physical_pair);
+            add_core_translation(translated_coord, physical_pair);
+        }
+    }
 }
 
 void BlackholeCoordinateManager::fill_arc_physical_translated_mapping() {


device/blackhole/blackhole_coordinate_manager.cpp
```diff
@@ -221,17 +221,20 @@ void BlackholeCoordinateManager::fill_eth_physical_translated_mapping() {
...
+        for (size_t y = 0; y < pcie_grid_size.y; y++) {
+            CoreCoord logical_coord = CoreCoord(x, y, CoreType::PCIE, CoordSystem::LOGICAL);
+            const tt_xy_pair physical_pair = to_physical_map[logical_coord];
 
-    CoreCoord translated_coord = CoreCoord(
-        blackhole::pcie_translated_coordinate_start_x,
-        blackhole::pcie_translated_coordinate_start_y,
-        CoreType::PCIE,
-        CoordSystem::TRANSLATED);
+            CoreCoord translated_coord = CoreCoord(
```
pjanevskiTT, line 229: Since we are using maps, I think having the same translated coordinate for different physical coordinates may cause problems. Is this working on all board types we have?

pjanevskiTT (commented):
I am not sure I see why this PR is needed. I left a comment which highlights some of the things that I think can be problematic in coordinate manager.
But in general, https://github.com/tenstorrent/tt-umd/pull/497 was supposed to introduce creation of Cluster for Blackhole, it should figure out for metal runtime which PCI core should be used. I don't think having the soc descriptor with pcie cores that are not available to software is the right way to go. I suppose that the problem can be the fact that Metal still can't omit the soc descriptor path out of Cluster constructor? If that is the case I think we should focus on fixing that problem rather than checking this in

device/blackhole/blackhole_coordinate_manager.cpp
pjanevskiTT: Even when order of PCI cores in soc desc yaml is switched?

pjanevskiTT:
So I thought about this and have a simple proposal on how to maybe fix this in a better way. If Blackhole Coordinate Manager had additional info on which board type it is working on (plus chip location if necessary), it could already map the proper PCI core to the translated one, and for the other PCI core do 1-1 mapping with physical. I think this could help to remove this logic out of metal runtime. Let me know what you think

pjanevskiTT:
As discussed with @abhullar-tt offline, this will be checked in so metal is not blocked, any further changes can be put through a metal blackhole ci to not break things

pjanevskiTT (approved):

abhullar-tt:
> So I thought about this and have a simple proposal on how to maybe fix this in a better way. If Blackhole Coordinate Manager had additional info on which board type it is working on (plus chip location if necessary), it could already map the proper PCI core to the translated one, and for the other PCI core do 1-1 mapping with physical. I think this could help to remove this logic out of metal runtime. Let me know what you think

I think some of your proposal might be captured by https://github.com/tenstorrent/tt-umd/pull/539. Im okay with letting that be merged and testing translation support on top of those changes + closing this PR

device/blackhole/blackhole_coordinate_manager.cpp
```diff
@@ -221,17 +221,20 @@ void BlackholeCoordinateManager::fill_eth_physical_translated_mapping() {
 }
 
 void BlackholeCoordinateManager::fill_pcie_physical_translated_mapping() {
-    CoreCoord logical_coord = CoreCoord(0, 0, CoreType::PCIE, CoordSystem::LOGICAL);
```
broskoTT, line 224: Yeah I think that now after #539 this is not needed for you for BH to work.
Sorry that my fix came after yours. Can you confirm that this now works?

That said, I think that this code does make sense though. But furthermore, that said, I don't think that everything will work well due to multiple coords being mapped to the same translated coord, as Pavle pointed out in another comment.

Bottomline, if we don't have defined translated coords for multiple PCIEs (seems like we don't), I'd vote to abandon this PR

broskoTT (changes requested):
Please confirm that this has been resolved already on UMD main, making this PR redundant.

broskoTT:
@abhullar-tt can we close this PR?

abhullar-tt:
> @abhullar-tt can we close this PR?

yes

Issue #526, opened 2025-02-18
Expose sending commands to noc1 and noc0


tt-vjovanovic:
When noc0 hangs, we would like to use noc1 to access chip.

ihamer-tt:
@tt-vjovanovic, This will also require tt-lens to support to select noc 1. Is this ready?

@broskoTT, as this feature is used for debugging a P0 (possibly HW-based) issue, and we need to look at it asap.

tt-vjovanovic:
Hacked solution with this PR: https://github.com/tenstorrent/tt-umd/pull/534

PR #534, opened 2025-02-24
Simple hack for #526 until it is properly fixed

tt-vjovanovic:
Exposing global variable for using noc1 in tlb.
Disabling coordinate translation when setting up tlb for noc1.

Note: this is bad implementation since required coordinates for communication with UMD are virtual which get converted into physical, but in tt-lens we are converting noc1 into virtual as if it is noc0, UMD reverses this so on device ends up noc1 coordinate...


broskoTT (approved):
Chatted offline, approving this to unblock using noc1. This is not a long term solution

PR #536, opened 2025-02-24
Explain harvesting mask in soc descriptor

broskoTT:
### Description
Updating a comment to make harvesting masks clearer


device/api/umd/device/tt_soc_descriptor.h
```diff
@@ -122,6 +122,15 @@ class tt_SocDescriptor {
     int eth_l1_size;
     bool noc_translation_id_enabled;
     uint64_t dram_bank_size;
+
+    // Harvesting mask is reported in logical coordinates, meaning the index of a bit that is set corresponds to the
+    // index of the row (in case of Tensix), or the index of the channel (in case of DRAM), or the index of the core
```
pjanevskiTT, line 127: nit: for ETH we also call it channel, maybe better than core. You can decide

pjanevskiTT (approved):

PR #538, opened 2025-02-24
Rename P150A to P150

pjanevskiTT:
### Issue

/

### Description

Rename P150A board type to P150. Board type naming should not be dependent on type of cooling or other stuff.

### List of the changes

- Rename P150A to P150

### Testing

CI

### API Changes

Metal uses P150A, will change it 

- [ ] tt-metal PR - ?

### DIFF
--- a/device/tt_cluster_descriptor.cpp
+++ b/device/tt_cluster_descriptor.cpp
@@ -752,7 +752,7 @@ void tt_ClusterDescriptor::load_chips_from_connectivity_descriptor(YAML::Node &y
                 board_type = BoardType::N300;
             } else if (chip_board_type.second == "p100") {
                 board_type = BoardType::P100;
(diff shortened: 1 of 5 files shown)


device/tt_cluster_descriptor.cpp
```diff
@@ -753,7 +753,7 @@ void tt_ClusterDescriptor::load_chips_from_connectivity_descriptor(YAML::Node &y
             } else if (chip_board_type.second == "p100") {
                 board_type = BoardType::P100;
             } else if (chip_board_type.second == "p150A") {
```
broskoTT, line 755: what about this string? Where is this reported from?

broskoTT: just add chip_board_type.second == "p150A" || chip_board_type.second == "p150"

broskoTT (approved):

Issue #478, opened 2025-01-23
[Topology] Wait for DRAM training to finish


pjanevskiTT:
Wait for DRAM training to finish when creating cluster descriptor/topology. This information is still not easily available on Blackhole but it should be soon. Will update the issue with more details when we have the next fw version

ejouretTT:
Please use tt-metal-dev-umd-ci with 80.18.0.0. to test this

broskoTT:
Added the machine, and the tests passed, merged the PR.

PR #542, opened 2025-02-25
Wait for Blackhole DRAM training

pjanevskiTT:
### Issue

#478 

Note: the PR is probably going to timeout on Blackhole CI until the fw from the syseng is released, so reviewers can do the review without me checking this in, to be ready when the fw is released.

### Description

Implement waiting for DRAM training on Blackhole. Format of the data provided by syseng to sw is described in https://tenstorrent.atlassian.net/browse/SYS-1241. This PR integrates the wait in our code according to the proposed ticket design.

### List of the changes

- Add function to wait for DRAM cores
- Implement the proposed design from the ticket
- Add functions to TTDevice that expose DRAM training info

### Testing
CI

### API Changes
/

### DIFF
--- a/device/api/umd/device/tt_device/blackhole_tt_device.h
+++ b/device/api/umd/device/tt_device/blackhole_tt_device.h
@@ -32,7 +32,10 @@ public:
     BoardType get_board_type() override;
 
     void dma_d2h(void *dst, uint32_t src, size_t size) override;
+
     void dma_h2d(uint32_t dst, const void *src, size_t size) override;
+
+    std::vector<DramTrainingStatus> get_dram_training_status() override;
 
 private:
     static constexpr uint64_t ATU_OFFSET_IN_BH_BAR2 = 0x1200;
--- a/device/chip/local_chip.cpp
+++ b/device/chip/local_chip.cpp
@@ -428,10 +428,6 @@ std::unique_lock<RobustMutex> LocalChip::acquire_mutex(MutexType mutex_type, int
 }
 
 void LocalChip::wait_dram_cores_training(const uint32_t timeout_ms) {
-    if (get_tt_device()->get_arch() == tt::ARCH::BLACKHOLE) {
-        return;
-    }
-
     TTDevice* tt_device = get_tt_device();
 
     auto start = std::chrono::system_clock::now();
--- a/device/tt_device/blackhole_tt_device.cpp
+++ b/device/tt_device/blackhole_tt_device.cpp
@@ -90,6 +90,7 @@ ChipInfo BlackholeTTDevice::get_chip_info() {
     chip_info.harvesting_masks.dram_harvesting_mask = telemetry->is_entry_available(blackhole::TAG_ENABLED_GDDR)
                                                           ? (~telemetry->read_entry(blackhole::TAG_ENABLED_GDDR) & 0xFF)
                                                           : 0;
+
     chip_info.harvesting_masks.eth_harvesting_mask = telemetry->is_entry_available(blackhole::TAG_ENABLED_ETH)
                                                          ? (~telemetry->read_entry(blackhole::TAG_ENABLED_ETH) & 0x3FFF)
                                                          : 0;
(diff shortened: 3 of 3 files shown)


device/chip/local_chip.cpp
```diff
@@ -94,4 +95,56 @@ void LocalChip::wait_eth_cores_training(const uint32_t timeout_ms) {
     }
 }
 
+void LocalChip::wait_dram_cores_training(const uint32_t timeout_ms) {
```
pjanevskiTT, line 98: Blackhole CI timed out as expected, don't merge the PR until the CI passes

device/api/umd/device/tt_device/blackhole_tt_device.h
```diff
@@ -23,6 +23,8 @@ class BlackholeTTDevice : public TTDevice {
 
     void wait_arc_core_start(const tt_xy_pair arc_core, const uint32_t timeout_ms = 1000) override;
 
+    std::pair<bool, uint32_t> get_dram_training_status() override;
```
pjanevskiTT, line 26: change this to return `std::optional<uint32_t>`

device/chip/local_chip.cpp
```diff
@@ -94,4 +95,56 @@ void LocalChip::wait_eth_cores_training(const uint32_t timeout_ms) {
...
+    if (get_tt_device()->get_arch() != tt::ARCH::BLACKHOLE) {
+        return;
+    }
+
+    TTDevice* tt_device = get_tt_device();
+
+    auto start = std::chrono::system_clock::now();
```
broskoTT, line 105: We now have this implemented two times. This now warrants a common implementation. Maybe a WaitOnEvent class, wrapped around a lambda which reads these values.

Not a requirement for this PR, just sharing some ideas. We don't have to have an issue now with it, it is obvious. And not critical.

No action

device/tt_device/blackhole_tt_device.cpp
```diff
@@ -141,4 +141,14 @@ void BlackholeTTDevice::wait_arc_core_start(const tt_xy_pair arc_core, const uin
     }
 }
 
+std::pair<bool, uint32_t> BlackholeTTDevice::get_dram_training_status() {
```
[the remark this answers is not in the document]
broskoTT: Yeah actually an array of those enums, since you have it per channel.

broskoTT (approved):

device/api/umd/device/tt_device/blackhole_tt_device.h
pjanevskiTT: done

device/chip/local_chip.cpp
pjanevskiTT: agree, I'll post an issue as follow up

pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/556

device/tt_device/blackhole_tt_device.cpp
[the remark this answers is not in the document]
pjanevskiTT: Generally I agree with this comment, but the problem with returning the array for all the cores is that then TTDevice knows for the number of the cores, which is at the moment always 8 for Blackhole, but not might be true for some custom architecture or something in the future. I feel like it is on Chip which knows about the cores to translate this info into something meaningful. Plus harvesting comes into play since we are going to have 00 on cores which are harvested and TTDevice shouldn't care about the info, just be able to get it. I am open for discussion here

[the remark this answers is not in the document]
broskoTT: well that telemetry for blackhole is hardcoded to 8 entries, so it is fine if we hardcode somewhere number 8 as well. That is not flexible, it is defined by firmware.
Regarding 00 for cores which are harvested, that's fine, it will return IN_PROGRESS and chip will ignore it since it knows it is harvested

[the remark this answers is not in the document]
pjanevskiTT: e7bc2ba4b078ff1c43f326f83abf6d3e808c5134

device/chip/local_chip.cpp
pjanevskiTT: Seems training and error bits were flipped from jira ticket, but that is easy change on our side. Waiting for fw to be released to CI machines

ejouretTT:
Please use tt-metal-dev-umd-ci which is on 80.18.0.0. to complete testing for this

PR #547, opened 2025-02-27
Fixing BlackholeArcTelemetryReader to use noc1 coordinate when umd_use_noc1 is specified

tt-vjovanovic:
With this change, starting `tt-lens` with noc1 should work even after noc0 is corrupted.

### DIFF
--- a/device/api/umd/device/blackhole_arc_telemetry_reader.h
+++ b/device/api/umd/device/blackhole_arc_telemetry_reader.h
@@ -50,7 +50,7 @@ private:
     bool telemetry_entry_available[NUMBER_TELEMETRY_TAGS];
     uint32_t telemetry_offset[NUMBER_TELEMETRY_TAGS];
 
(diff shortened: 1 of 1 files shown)


device/api/umd/device/blackhole_arc_telemetry_reader.h
```diff
@@ -48,7 +50,7 @@ class BlackholeArcTelemetryReader {
     bool telemetry_entry_available[NUMBER_TELEMETRY_TAGS];
     uint32_t telemetry_offset[NUMBER_TELEMETRY_TAGS];
 
-    const tt_xy_pair arc_core = tt::umd::blackhole::ARC_CORES[0];
+    const tt_xy_pair arc_core = !umd_use_noc1 ? tt::umd::blackhole::ARC_CORES[0] : tt_xy_pair(8, 11);
```
pjanevskiTT, line 53: can you just add a comment that 8,11 is ARC core in NOC1 coordinate system?

PR #562, opened 2025-03-03
Enable Blackhole AICLK test

pjanevskiTT:
### Issue

/

### Description

#548 introduced few tests to test the ARC communication, one of those being go busy/idle messages and then readouts of the Blackhole clk, but it was left disabled because 80.15 fw is not yet flashed on CI machines. This PR reenables the test and it will be merged once the new fw is released to machines.

### List of the changes

- Reenable BH AICLK test

### Testing

CI

### API Changes
/

PR #569, opened 2025-03-03
Implement TTDevice get clock

pjanevskiTT:
### Issue

Expose readouts of AICLK through TTDevice. Metal uses Cluster get_clocks which should work both for Blackhole and Wormhole.

### Description

get_clocks in Cluster should work for all architectures. Hide get_clock for each device behind TTDevice. 

### List of the changes

- Implement TTDevice get_clock
- Change Cluster get_clock to call TTDevice get_clock
- Change tests to use TTDevice get_clock

### Testing
CI

### API Changes
/

### DIFF
--- a/device/tt_device/wormhole_tt_device.cpp
+++ b/device/tt_device/wormhole_tt_device.cpp
@@ -14,4 +14,20 @@ ChipInfo WormholeTTDevice::get_chip_info() {
     throw std::runtime_error("Reading ChipInfo is not supported for Wormhole.");
 }
 
+uint32_t WormholeTTDevice::get_clock() {
+    const uint32_t timeouts_ms = 1000;
+    // There is one return value from AICLK ARC message.
+    std::vector<uint32_t> arc_msg_return_values = {0};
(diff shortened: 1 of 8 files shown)


abhullar-tt (approved):

device/tt_device/wormhole_tt_device.cpp
```diff
@@ -14,4 +14,20 @@ ChipInfo WormholeTTDevice::get_chip_info() {
     throw std::runtime_error("Reading ChipInfo is not supported for Wormhole.");
 }
 
+uint32_t WormholeTTDevice::get_clock() {
```
broskoTT, line 17: Do we have tests which use this API for both WH and BH, and do they still pass?

broskoTT (approved):

device/tt_device/wormhole_tt_device.cpp
pjanevskiTT: For WH it's on this PR, for BH it's on #562, I put a comment to not merge it before this

PR #572, opened 2025-03-04
Showcase usage of eth connectivity api's

broskoTT:
### Issue
Related to #431 

### Description
Consolidating required APIs in an api code example test

### List of the changes
- Added test
- Fill active/idle eth channels from .yaml file as well.

### Testing
CI

### API Changes

### DIFF
--- a/device/tt_cluster_descriptor.cpp
+++ b/device/tt_cluster_descriptor.cpp
@@ -8,6 +8,8 @@
 #include <memory>
 #include <sstream>
 
+#include "api/umd/device/blackhole_implementation.h"
+#include "api/umd/device/wormhole_implementation.h"
 #include "disjoint_set.hpp"
 #include "fmt/core.h"
 #include "libs/create_ethernet_map.h"
@@ -485,9 +487,8 @@ void tt_ClusterDescriptor::load_ethernet_connections_from_connectivity_descripto
 
     // Preload idle eth channels.
     for (const auto &chip : desc.all_chips) {
-        int num_channels = desc.chip_arch.at(chip) == tt::ARCH::BLACKHOLE
-                               ? 14
-                               : (desc.chip_arch.at(chip) == tt::ARCH::WORMHOLE_B0 ? 16 : 0);
+        int num_channels =
+            tt::umd::architecture_implementation::create(desc.chip_arch.at(chip))->get_num_eth_channels();
         for (int i = 0; i < num_channels; i++) {
             desc.idle_eth_channels[chip].insert(i);
         }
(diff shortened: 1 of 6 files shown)


device/tt_cluster_descriptor.cpp
```diff
@@ -482,6 +482,17 @@ std::unique_ptr<tt_ClusterDescriptor> tt_ClusterDescriptor::create_mock_cluster(
 void tt_ClusterDescriptor::load_ethernet_connections_from_connectivity_descriptor(
     YAML::Node &yaml, tt_ClusterDescriptor &desc) {
     log_assert(yaml["ethernet_connections"].IsSequence(), "Invalid YAML");
+
+    // Preload idle eth channels.
+    for (const auto &chip : desc.all_chips) {
+        int num_channels = desc.chip_arch.at(chip) == tt::ARCH::BLACKHOLE
+                               ? 14
```
pjanevskiTT, line 489: I think this needs to take into account eth harvesting mask as well, which should be encoded in cluster descriptor. We don't want to preload harvested eth channels to idle channels. I know this requires bit more work on cluster descriptor so I am ok with merging this as it is now

pjanevskiTT, line 489: we have a constant for number of ETH channels, both WH/BH, I think you should change both 14 and 16 with these constants. NUMBER_OF_ETH_CHANNELS it should be called

tests/api/test_cluster_descriptor.cpp
```diff
@@ -144,3 +144,56 @@ TEST(ApiClusterDescriptorTest, SeparateClusters) {
...
+    }
+
+    auto chips_with_mmio = cluster_desc->get_chips_with_mmio();
+    for (auto [chip, mmio_chip] : chips_with_mmio) {
+        std::cout << "Chip " << chip << " has MMIO on PCI id " << mmio_chip << std::endl;
+    }
+
+    for (auto chip : cluster_desc->get_all_chips()) {
+        // Wormhole has 16 and Blackhole has 14 ethernet channels.
+        for (int eth_chan = 0; eth_chan < 16; eth_chan++) {
```
pjanevskiTT, line 171: same comment for constant as above

tests/api/test_cluster_descriptor.cpp
```diff
@@ -144,3 +144,56 @@ TEST(ApiClusterDescriptorTest, SeparateClusters) {
         EXPECT_TRUE(chip_clusters.are_same_set(chip, closest_mmio_chip));
     }
 }
+
+TEST(ApiClusterDescriptorTest, EthernetConnectivity) {
```
pjanevskiTT, line 148: Nice :). No action

pjanevskiTT (approved):

device/tt_cluster_descriptor.cpp
pjanevskiTT: It knows from https://github.com/tenstorrent/tt-umd/pull/570. It has to know atm because our pipeline is
1. create cluster desc
2. create cluster from cluster desc
It is inverse of what it should be, but until we can't get rid of create-ethernet-map it will stay this way. I think metal folks would at least like to know which cores are harvested in serialized cluster descriptor, if not the full harvesting mask

Issue #576, opened 2025-03-04
Blackhole harvesting testing


broskoTT:
For a while, @pjanevskiTT  was using yyz-lab-125 blackhole machine, which is the only harvested bh card we currently have.
I picked it up today, and implemented some fixes.

This task tracks e2e harvesting testing.

pjanevskiTT:
To confirm that DRAM harvesting seems good on one-bank harvested machine https://github.com/tenstorrent/tt-umd/commit/fee65be3afc12e3cb0b9e0ef201dc41d2cb3bdd4 I will add a test to confirm

broskoTT:
Since we don't have machine with harvesting in CI, having tests won't help much. This task was meant to track our manual testing of the BH machine with harvesting.
@pjanevskiTT  I think we're good now after https://github.com/tenstorrent/tt-umd/pull/599, I'd close this task since it doesn't have any action item on our side

PR #577, opened 2025-03-04
Fix tests for harvested BH

broskoTT:
### Issue
Finishes #576 

### Description
Fixed a couple of items

### List of the changes
- (Didn't lead to this specific issue) Pass board type to soc descriptor
- Hardcode eth harvesting to all harvested instead of none harvested for P100
- get_num_dram_channels took size of old structure which is unharvested, switched to new structure
- (Didn't lead to this specific issue) Switch all blackhole_140_arch_no_eth to blackhole_140_arch.yaml
- Enable harvesting in CreateDestroy test. This works for WH with false, since it uses a soc descriptor which is 1x1. If we ended up on a card which has 1st row harvested, these test would probably fail for WH
- Fix hardcoded dram num channels somewhere in the tests
- Switch l1_membar to use new umd API which uses CoreCoord api

### Testing
Manually tested on yyz-lab-125. Obviously the CI tests should still pass.

### API Changes

### DIFF
--- a/device/tt_device/blackhole_tt_device.cpp
+++ b/device/tt_device/blackhole_tt_device.cpp
@@ -114,8 +114,9 @@ ChipInfo BlackholeTTDevice::get_chip_info() {
 
     chip_info.board_type = get_board_type_from_board_id(chip_info.chip_uid.board_id);
 
+    // TODO: likely not needed anymore. Firware on P100 will give 0 for TAG_ENABLED_ETH
     if (chip_info.board_type == BoardType::P100) {
-        chip_info.harvesting_masks.eth_harvesting_mask = 0;
+        chip_info.harvesting_masks.eth_harvesting_mask = 0x3FFF;
     }
 
     return chip_info;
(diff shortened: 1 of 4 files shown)


device/tt_device/blackhole_tt_device.cpp
```diff
@@ -114,8 +114,9 @@ ChipInfo BlackholeTTDevice::get_chip_info() {
 
     chip_info.board_type = get_board_type_from_board_id(chip_info.chip_uid.board_id);
 
+    // TODO: likely not needed anymore. Firware on P100 will give 0 for TAG_ENABLED_ETH
```
abhullar-tt, line 117: this FW will be part of the harvesting drop?

abhullar-tt (approved):

device/tt_device/blackhole_tt_device.cpp
broskoTT: Not sure. Currently on yyz-lab-125 it works like that, and I suspect it will work like that in the future as well. But still, marking it explicitly for now that all eth are harvested on p100.

PR #578, opened 2025-03-05
Remove cluster descriptor API from public

pjanevskiTT:
### Issue

Cluster descriptor create API always calls create-ethernet-map. This lib does not work for BH, UBB...This API should be moved from public and all calls should go through `Cluster::create_cluster_descriptor.`

### Description

Only public accessible API for creating cluster descriptor from silicon should be `Cluster::create_cluster_descriptor`. Make all tests use this API and remove create-ethernet-map call from public API.

### List of the changes

- Move `tt_ClusterDescriptor::create()` from public
- Make tests use Cluster API

### Testing
CI

### API Changes

`tt_ClusterDescriptor::create()` is not public anymore - switch all clients to `Cluster::create_cluster_descriptor()`

- [ ]  tt_metal PR - ?

### DIFF
--- a/device/api/umd/device/tt_cluster_descriptor.h
+++ b/device/api/umd/device/tt_cluster_descriptor.h
@@ -78,6 +78,8 @@ protected:
 
     void fill_chips_grouped_by_closest_mmio();
 
+    static std::unique_ptr<tt_ClusterDescriptor> create();
+
     std::map<chip_id_t, uint32_t> dram_harvesting_masks = {};
     std::map<chip_id_t, uint32_t> eth_harvesting_masks = {};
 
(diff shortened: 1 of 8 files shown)


device/api/umd/device/tt_cluster_descriptor.h
```diff
@@ -98,7 +100,6 @@ class tt_ClusterDescriptor {
     // get_cluster_descriptor_file_path will create ethernet map in the background.
     static std::string get_cluster_descriptor_file_path();
```
pjanevskiTT, line 101: Don't merge the PR until this is also moved to private. We have alternative for that in Cluster as well

pjanevskiTT:
Closing this in favour of upcoming UBB PR, this will hide evertything that is needed

Issue #544, opened 2025-02-26
Blackhole telemetry tag array size


pjanevskiTT:
https://github.com/tenstorrent/tt-umd/pull/543 fixed our unit tests on fw 80.15. We just resized the telemetry tag array. Proper solution would be to implement support for any tag or just fix the Blackhole spec that the tags don't change.

pjanevskiTT:
closed by #580

PR #580, opened 2025-03-05
Implement dynamic Blackhole telemetry

pjanevskiTT:
### Issue

#544 

### Description

Array of Blackhole tags has been statically allocated so far. Since Blackhole telemetry will change further and expand number of tags, implement telemetry to support any tag that is read from the ARC core. This should prevent bugs that we had during 80.15 fw release.

### List of the changes

- Remove limit on number of tags
- Change arrays to maps for telemetry values and offsets

### Testing

CI (every BH test goes through the telemetry)

### API Changes
/


pjanevskiTT:
Confirming that BH tests pass locally on P150 card

PR #583, opened 2025-03-06
Add support for P300 board type

pjanevskiTT:
### Issue

Add support for recognizing P300 board type.

### Description

Add recognizing P300 board type from board upi.

### List of the changes

- Add recognizing P300 board type from upi
- Add recognizing P300 from cluster descriptor yaml

### Testing

CI. This is to enable testing on P300 silicon

### API Changes
/

### DIFF
--- a/device/tt_cluster_descriptor.cpp
+++ b/device/tt_cluster_descriptor.cpp
@@ -776,7 +776,9 @@ void tt_ClusterDescriptor::load_chips_from_connectivity_descriptor(YAML::Node &y
                 chip_board_type.second == "p150" || chip_board_type.second == "p150A" ||
                 chip_board_type.second == "p150C") {
                 board_type = BoardType::P150;
-            } else if (chip_board_type.second == "p300") {
+            } else if (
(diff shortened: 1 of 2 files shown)


device/tt_cluster_descriptor.cpp
```diff
@@ -756,7 +756,9 @@ void tt_ClusterDescriptor::load_chips_from_connectivity_descriptor(YAML::Node &y
                 chip_board_type.second == "p150" || chip_board_type.second == "p150A" ||
                 chip_board_type.second == "p150C") {
                 board_type = BoardType::P150;
-            } else if (chip_board_type.second == "p300") {
+            } else if (
+                chip_board_type.second == "p300" || chip_board_type.second == "p300A" ||
```
broskoTT, line 760: We should really have "to_lower" or something like that here. We have tt::to_lower in arch.h

broskoTT (approved):

device/tt_cluster_descriptor.cpp
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/584

Issue #560, opened 2025-03-02
Wormhole galaxy topology discovery


pjanevskiTT:
Implement topology discovery on the new Wormhole galaxy. Syseng fw is changing, this should enable us to implement topology discovery from the beginning

pjanevskiTT:
Closed by #592

PR #592, opened 2025-03-07
Implement Wormhole UBB support

pjanevskiTT:
### Issue

#560 

### Description

Implement support for UBB galaxy board. ETH firmware is changing for UBB so topology discovery from scratch is needed. Also support for general recognition of UBB board type is needed.

This is a good step in implementing Wormhole topology discovery completely from scratch. Old WH boards still use legacy create-ethernet-map path.

### List of the changes

- Topology discovery for UBB
- Recognize UBB board type
- Hide cluster descriptor APIs that use create-ethernet-map by default from public API.

### Testing

CI + local testing. Hard to write UBB tests at the moment, we don't have this machines on CI

### API Changes

Cluster descriptor API

- [ ] tt_metal approved PR pointing to this branch - ?
- [ ] tt_debuda approved PR pointing to this branch - ?


device/api/umd/device/tt_cluster_descriptor.h
```diff
@@ -78,6 +78,8 @@ class tt_ClusterDescriptor {
 
     void fill_chips_grouped_by_closest_mmio();
 
+    static std::unique_ptr<tt_ClusterDescriptor> create();
```
broskoTT, line 81: I don't think this should not be public. Or did you move it due to Cluster creating it?

device/api/umd/device/tt_device/wormhole_tt_device.h
```diff
@@ -13,8 +13,12 @@ class WormholeTTDevice : public TTDevice {
 public:
     WormholeTTDevice(std::unique_ptr<PCIDevice> pci_device);
 
+    void wait_arc_core_start(const tt_xy_pair arc_core, const uint32_t timeout_ms = 1000) override;
```
broskoTT, line 16: Should this be in tt_device? Is this implemented for Blackhole?
Why pass position of the arc code?

device/cluster.cpp
```diff
@@ -3133,61 +3245,75 @@ std::unique_ptr<tt_ClusterDescriptor> Cluster::create_cluster_descriptor(
...
-                log_debug(
-                    LogSiliconDriver,
-                    "Port on eth core ({}, {}) on chip {} is in unknown state",
-                    eth_core.x,
-                    eth_core.y,
-                    chip_id);
             }
         }
+    } else {
+        ubb_eth_connections(chips, desc);
```
broskoTT, line 3316: I don't like that we do this here, but for blackhole we have whole implementation.
Can you create a separate function which encapsulates implementation for blackhole, and then Cluster::create_cluster_descriptor would remain as ~10 loc

Also this is a good signal that we should probably have a class TopologyDiscovery. Out of scope for this PR though. You can create an issue if you agree.

device/tt_device/wormhole_tt_device.cpp
```diff
@@ -11,9 +11,38 @@ WormholeTTDevice::WormholeTTDevice(std::unique_ptr<PCIDevice> pci_device) :
...
 ChipInfo WormholeTTDevice::get_chip_info() {
-    throw std::runtime_error("Reading ChipInfo is not supported for Wormhole.");
+    ChipInfo chip_info;
+
+    uint32_t niu_cfg;
+    const tt_xy_pair dram_core = {0, 0};
+    const uint64_t niu_cfg_addr = 0x1000A0000 + 0x100;
+    read_from_device(&niu_cfg, dram_core, niu_cfg_addr, sizeof(uint32_t));
+
+    bool noc_translation_enabled = (niu_cfg & (1 << 14)) != 0;
```
broskoTT, line 21: please put NOC_TRANSLATION_LOCATION from this line, and above niu_cfg_addr in some constants. wormhole::implementation probably

device/tt_device/wormhole_tt_device.cpp
```diff
@@ -11,9 +11,38 @@ WormholeTTDevice::WormholeTTDevice(std::unique_ptr<PCIDevice> pci_device) :
...
+    if (ret_code != 0) {
+        throw std::runtime_error(fmt::format("Failed to get harvesting masks with exit code {}", ret_code));
+    }
+
+    chip_info.harvesting_masks.tensix_harvesting_mask = arc_msg_return_values[0];
+
+    return chip_info;
 }
 
+void WormholeTTDevice::wait_arc_core_start(const tt_xy_pair arc_core, const uint32_t timeout_ms) {}
```
broskoTT, line 44: So you've added this without implementation? If its not used now, remove it. When we need it we will implement it.

device/api/umd/device/tt_cluster_descriptor.h
broskoTT: Actually yeah, we're going towards not calling cem anymore. So this is fine.

broskoTT (approved):

device/tt_device/wormhole_tt_device.cpp
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/595

device/tt_device/wormhole_tt_device.cpp
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/595

device/cluster.cpp
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/595

Issue #588, opened 2025-03-06
Wrong channels in ethernet_connections for harvested BH


broskoTT:
On eth harvested BH, for example eth_id can be reported as 13, and this is not channel 13 but channel 11. Add a proper map between eth_id and our logical eth_channel in the Cluster::create_cluster_descriptor.

Add a test for it.

This was surfaced by tt_metal tt_cluster get_ethernet_cores_grouped_by_connected_chips function. It called get_directly_connected_ethernet_channels_between_chips, and then interpreted returned values as logical coords (which should be true), but the (0,13) doesn't exist in logical coords.

Not sure how to test this without faking what is being read from the device... Maybe with a mock tt_device. We'd benefit from this in the future. Maybe test is out of scope for the fix.

https://docs.google.com/spreadsheets/d/1tGG4UPfABXrd97Y3VPJS7CusDFlamcml9kEw1HEwrlM/edit?gid=95904394#gid=95904394

broskoTT:
Printout from live machine
yeah we thought this was the same eth_channel 4 boot_results.local_info.eth_id 4
yeah we thought this was the same eth_channel 5 boot_results.local_info.eth_id 5
yeah we thought this was the same eth_channel 7 boot_results.local_info.eth_id 8
yeah we thought this was the same eth_channel 8 boot_results.local_info.eth_id 10
yeah we thought this was the same eth_channel 9 boot_results.local_info.eth_id 11
yeah we thought this was the same eth_channel 10 boot_results.local_info.eth_id 12
yeah we thought this was the same eth_channel 11 boot_results.local_info.eth_id 13
yeah we thought this was the same eth_channel 4 boot_results.local_info.eth_id 4
yeah we thought this was the same eth_channel 5 boot_results.local_info.eth_id 5
yeah we thought this was the same eth_channel 7 boot_results.local_info.eth_id 8
yeah we thought this was the same eth_channel 8 boot_results.local_info.eth_id 10
yeah we thought this was the same eth_channel 9 boot_results.local_info.eth_id 11
yeah we thought this was the same eth_channel 10 boot_results.local_info.eth_id 12
yeah we thought this was the same eth_channel 11 boot_results.local_info.eth_id 13

pjanevskiTT:
closed by #599

PR #599, opened 2025-03-10
Blackhole ETH id mapping fix

pjanevskiTT:
### Issue

#588 ETH id from ETH core is not considering harvesting, and we would want to have logical ETH channels which consider harvesting in cluster descriptor. We put ETH id from the core in the cluster descriptor, so now considering harvesting as well

### Description

Add mapping of ETH id from the core to the logical ETH channel to know which logical channel to put in the cluster descriptor instead od ETH id from ETH core.

### List of the changes

- Add mapping of ETH id from ETH core to logical eth channel
- Add logical eth channels to cluster descriptor
- Add verification that there is no higher ETH logical channel that number of channels

### Testing
CI + ran locally to check the mapping

### API Changes
/

### DIFF
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -40,6 +40,7 @@
 #include "api/umd/device/tt_core_coordinates.h"
 #include "logger.hpp"
 #include "umd/device/architecture_implementation.h"
+#include "umd/device/blackhole_implementation.h"
 #include "umd/device/chip/local_chip.h"
 #include "umd/device/chip/mock_chip.h"
 #include "umd/device/chip/remote_chip.h"
@@ -3238,7 +3239,6 @@ std::unique_ptr<tt_ClusterDescriptor> Cluster::create_cluster_descriptor(
         }
     }
 
-    std::map<chip_id_t, std::map<size_t, size_t>> map_eth_id_to_logical_core;
     for (auto& it : chips) {
         const chip_id_t chip_id = it.first;
         const std::unique_ptr<Chip>& chip = it.second;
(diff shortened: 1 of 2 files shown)


device/cluster.cpp
```diff
@@ -3277,8 +3290,10 @@ std::unique_ptr<tt_ClusterDescriptor> Cluster::create_cluster_descriptor(
                             remote_info.get_chip_uid().board_id);
                     } else {
                         // Adding a connection only one way, the other chip should add it another way.
-                        desc->ethernet_connections[local_chip_id][local_info.eth_id] = {
-                            remote_chip_id.value(), remote_info.eth_id};
+                        desc->ethernet_connections[local_chip_id]
+                                                  [map_eth_id_to_logical_core[local_chip_id][local_info.eth_id]] = {
```
broskoTT, line 3294: can we use this:
`soc_descriptor->translate_coord_to(tt::umd::blackhole::ETH_CORES[local_info.eth_id], CoordSystem::PHYSICAL, CoordSystem::LOGICAL)`
instead of
`map_eth_id_to_logical_core[local_chip_id][local_info.eth_id]`

Then you don't need initialization of the map above?

device/cluster.cpp
```diff
@@ -3307,6 +3322,38 @@ std::unique_ptr<tt_ClusterDescriptor> Cluster::create_cluster_descriptor(
         ubb_eth_connections(chips, desc);
     }
 
+    // This is a way to verify that cluster descriptor does not have logical eth connections
+    // that has the value bigger than number of non-harvested ETH cores for the chip.
+    for (auto& [local_chip_id, eth_connections] : desc->ethernet_connections) {
+        size_t local_chip_eth_cores_size =
```
broskoTT, line 3328: This is really out of scope for this class.
Can we instead have "ClusterDescriptor::add_eth_connection" which would do this check?
Also, if you accept the other suggestion, then you don't even have to test here since the code will automatically throw if there is no logical coordinate

device/cluster.cpp
```diff
@@ -3293,9 +3294,15 @@ std::unique_ptr<tt_ClusterDescriptor> Cluster::create_cluster_descriptor(
...
+                                                                   ->get_soc_descriptor()
+                                                                   .translate_coord_to(
+                                                                       blackhole::ETH_CORES[remote_info.eth_id],
+                                                                       CoordSystem::PHYSICAL,
+                                                                       CoordSystem::LOGICAL);
                         // Adding a connection only one way, the other chip should add it another way.
-                        desc->ethernet_connections[local_chip_id][local_info.eth_id] = {
-                            remote_chip_id.value(), remote_info.eth_id};
+                        desc->ethernet_connections[local_chip_id][eth_channel] = {
+                            remote_chip_id.value(), logical_remote_coord.y};
```
broskoTT, line 3305: it would be really nice if we could test this prior to checking in... but if not, I'm still fine with it

broskoTT (approved):

Issue #314, opened 2024-11-18
Move standard SOC descriptors to code


pjanevskiTT:
Move standard SOC descriptors to structs/class/constants instead od yaml files. This was we don't have to rely on relative paths to yaml files when we want to use it. We already have something like this in files that are arch specific, for example look at https://github.com/tenstorrent/tt-umd/blob/main/device/wormhole/wormhole_implementation.h, we just need to integrate all of this into creating soc descriptor instead of reading yaml files.

When the client wants to call the constructor of Cluster with its own descirptors, it needs to be sure that relative paths are good, meaning UMD doesn't need to figure these things out.

pjanevskiTT:
closed by #609

PR #609, opened 2025-03-13
Create standard tt_SocDescriptor without yamls

pjanevskiTT:
### Issue
#314 
Create tt_SocDescriptor without yamls. UMD should never read yamls on it's own, only if it is passed from Metal side. The yaml path had few issues

- Slower than having it in code
- Metal CI doesn't have UMD yamls checked in so it fails with some new UMD features

### Description

Remove the need to always rely on yamls for standard architectures. Get cores and all the needed info from *_implementation.h files. tt_SocDescriptor now has the option to be created just with architecture for standard TT architectures.

### List of the changes

- Add SocDescriptorInfo struct, which is used to store all the information needed to create Soc descriptor
- Make yaml path fallback on SocDescriptorInfo as well
- Add a constructor to tt_SocDescriptor to be used only with architecture
- Change tests that were using path with yaml to use path with new constructor

### Testing

CI. I think our tests are enough to confirm this works. Since BH CI is broken, confirming unit tests pass on Blackhole as well.

### API Changes
/


device/api/umd/device/umd_utils.h
```diff
@@ -0,0 +1,19 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include <vector>
+
+#include "umd/device/types/xy_pair.h"
+
+static std::vector<tt_xy_pair> flatten_vector(const std::vector<std::vector<tt_xy_pair>>& vector_of_vectors) {
```
broskoTT, line 13: if you leave this in header file, add inline

device/api/umd/device/chip/local_chip.h
```diff
@@ -16,6 +16,8 @@ class LocalChip : public Chip {
 
     LocalChip(std::string sdesc_path, std::unique_ptr<TTDevice> tt_device);
 
+    LocalChip(std::unique_ptr<TTDevice> tt_device);
```
broskoTT, line 19: Can. we remove the constructor above this one LocalChip(std::string sdesc_path, std::unique_ptr<TTDevice> tt_device);  ?

device/cluster.cpp
```diff
@@ -3157,14 +3166,15 @@ std::unique_ptr<tt_ClusterDescriptor> Cluster::create_cluster_descriptor(std::st
     if (pci_device_info.begin()->second.get_arch() == tt::ARCH::BLACKHOLE) {
         std::vector<int> pci_device_ids = PCIDevice::enumerate_devices();
 
-        if (sdesc_path.empty()) {
-            sdesc_path = tt_SocDescriptor::get_soc_descriptor_path(tt::ARCH::BLACKHOLE);
```
broskoTT, line 3161: get_soc_descriptor_path should now be called only in tests, is this true?

device/api/umd/device/chip/local_chip.h
broskoTT: Ah okay I see the if else below. Nevermind, let's leave it for now, will refactor later.

tests/wormhole/test_cluster_wh.cpp
```diff
@@ -905,7 +905,7 @@ TEST(SiliconDriverWH, RandomSysmemTestWithPcie) {
     auto target_devices = get_target_devices();
 
     Cluster cluster(
-        test_utils::GetAbsPath("tests/soc_descs/wormhole_b0_8x10.yaml"),
+        tt::ARCH::WORMHOLE_B0,
```
broskoTT, line 908: How come above contructor doesn't need arch, but this one does? Why is arch needed here?

tests/blackhole/test_cluster_bh.cpp
```diff
@@ -83,15 +83,8 @@ TEST(SiliconDriverBH, CreateDestroy) {
     std::set<chip_id_t> target_devices = get_target_devices();
```
broskoTT, line 83: Please manually run bh tests before completing PR.

broskoTT (approved):
Awesome!

tests/wormhole/test_cluster_wh.cpp
pjanevskiTT: mistake, removed it

device/cluster.cpp
pjanevskiTT: yes

device/api/umd/device/blackhole_implementation.h
```diff
@@ -12,6 +12,7 @@
 #include "umd/device/architecture_implementation.h"
```
pjanevskiTT, line 12: Don't wanna merge this until metal post commit (WH and BH) are fine with this bump

broskoTT: completed https://github.com/tenstorrent/tt-metal/actions/runs/13837501376

PR #610, opened 2025-03-14
Update create-ethernet-map to luwen 0.6.2

TTDRosen:
This picks up stability fixes from the luwen repo
- Fix a host crash when the BH card has not fully initialized
- Fix a chip detect error when using erisc fw 6.11+ on QB or TG setups


device/libs/aarch64/libcreate_ethernet_map.a
broskoTT, line 1: This is the cluster descriptor that is generated:

image

Seems like our parsing doesn't like the new "routing enabled" field.

pjanevskiTT: should be fixed with https://github.com/tenstorrent/tt-umd/pull/613

pjanevskiTT: issue to integrate this info properly https://github.com/tenstorrent/tt-umd/issues/614

broskoTT: And #611 will print the exact cluster descriptor in unit test, for future debugging efforts

TTDRosen: Thanks for taking a look! That property will let you know if a route is routable with the original routing fw. It can be helpful to debug destination unreachable errors if systems are incorrectly wired.

pjanevskiTT: if you can rebase to #613 and there are no problems with tests, I am ok with checkin this in even with that new info, to not have to make new libs

pjanevskiTT: I'll merge it to main so you can rebase directly on main and try it out

pjanevskiTT (approved):
Approving to check this in when all required CI tests pass

PR #611, opened 2025-03-14
Print cluster descriptor in contents

broskoTT:
### Issue
No issue, but inspired by #610 

### Description
Adding cluster descriptor output in our tests, just for easier debugging.

### List of the changes
- Print serialized generated cluster descriptor in case of BH and UBB
- Print cluster descriptor generated by CEM in other cases
- Changed the order so that API tests are executed first

### Testing
- CI on this PR

### API Changes

### DIFF
--- a/tests/api/test_cluster_descriptor.cpp
+++ b/tests/api/test_cluster_descriptor.cpp
@@ -4,10 +4,10 @@
 
 #include <gtest/gtest.h>
 
+#include <fstream>
 #include <string>
 #include <unordered_map>
 #include <vector>
-#include <fstream>
 
 #include "disjoint_set.hpp"
 #include "tests/test_utils/generate_cluster_desc.hpp"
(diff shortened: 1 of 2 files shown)


tests/api/test_cluster_descriptor.cpp
```diff
@@ -202,3 +203,29 @@ TEST(ApiClusterDescriptorTest, EthernetConnectivity) {
...
+    if (pci_device_ids.size() == 0) {
+        GTEST_SKIP() << "No chips present on the system. Skipping test.";
+    }
+    std::unique_ptr<TTDevice> tt_device = TTDevice::create(pci_device_ids.at(0));
+
+    // In case of u6 galaxy and blackhole, we generate the cluster descriptor.
+    // For wormhole we still use create-ethernet-map.
+    std::filesystem::path cluster_path;
+    if (tt_device->get_arch() == tt::ARCH::BLACKHOLE || tt_device->get_board_type() == BoardType::UBB) {
+        cluster_path = tt::umd::Cluster::serialize_to_file();
```
pjanevskiTT, line 218: this path should work for all architectures and board types

pjanevskiTT (approved):

tests/api/test_cluster_descriptor.cpp
pjanevskiTT: I am fine with checking this in and then we can completely remove the other function for getting cluster descriptor file path

broskoTT: But the failure in drosen's PR fails exactly during parsing of cluster descriptor. So this test would fail before printing anything in that case. That's why I added this separation. It's not ideal for sure.

PR #618, opened 2025-03-17
Coordinate manager NOC1 support

pjanevskiTT:
### Issue

part of the larger effort #526 

### Description

Add initial suppport for NOC1 coordinate translation to coordinate manager. There is still an option to create coordinate manager without NOC1 coordinates because we still want to support the path of soc descriptor which creates coordinate manager without NOC1. This will be adressed in separate PR

### List of the changes

- Add NOC1 translation support to coordinate manager
- Add guard to not run tests that require silicon on machines without silicon

### Testing

CI+ additional tests added to test bench

### API Changes
/


device/api/umd/device/blackhole_coordinate_manager.h
```diff
@@ -23,7 +23,13 @@ class BlackholeCoordinateManager : public CoordinateManager {
         const std::vector<tt_xy_pair>& arc_cores,
```
pjanevskiTT, line 23: don't merge this until the follow up issues to support NOC1 are created

device/api/umd/device/blackhole_coordinate_manager.h
```diff
@@ -23,7 +23,13 @@ class BlackholeCoordinateManager : public CoordinateManager {
...
         const tt_xy_pair& pcie_grid_size,
         const std::vector<tt_xy_pair>& pcie_cores,
-        const std::vector<tt_xy_pair>& router_cores);
+        const std::vector<tt_xy_pair>& router_cores,
+        const std::vector<tt_xy_pair>& tensix_cores_noc1 = {},
```
tt-vjovanovic, line 27: Consider transferring noc0 to noc1 mapping (x0->x1, y0->y1) instead of calculated values. It seems like less overhead to have translation on single place, instead of duplication for both nocs and choosing which one you need...

pjanevskiTT: Thanks for recommending this, changed it

device/api/umd/device/blackhole_coordinate_manager.h
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/619
#620

device/api/umd/device/tt_core_coordinates.h
```diff
@@ -41,6 +41,7 @@ enum class CoordSystem : std::uint8_t {
     PHYSICAL,
     VIRTUAL,
     TRANSLATED,
+    NOC1,
 };
```
tt-vjovanovic, line 45: Consider adding `NOC0 = PHYSICAL,` as last row of this enum... But it needs to be last row, or first row after, `PHYSICAL,` to preserve value continuation...

device/blackhole/blackhole_implementation.cpp
```diff
@@ -97,13 +97,15 @@ tt_driver_noc_params blackhole_implementation::get_noc_params() const {
 }
 
 namespace blackhole {
-std::vector<tt_xy_pair> get_pcie_cores(const BoardType board_type, const uint8_t asic_location) {
```
tt-vjovanovic, line 100: Is change to this function still needed?

tt-vjovanovic (approved):
Can you also add changes to test that verifies `NOC_NODE_ID` stores correct value when you use NOC1 coordinates? Or is it still too soon (missing more changes to send messages over noc1)?

pjanevskiTT:
I'll add a test for NOC1 noc node id separately

device/blackhole/blackhole_implementation.cpp
pjanevskiTT: no, changed it

device/api/umd/device/blackhole_coordinate_manager.h
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/621

PR #566, opened 2025-03-03
Integrate ArcMessenger into Cluster code path

pjanevskiTT:
### Issue

#549 

### Description

Integrate ARC messenger class introduced by #548 to Cluster code path

### List of the changes

- Add ArcMessenger to TTDevice class
- Call ArcMessenger from TTDevice each time PCIe arc message is needed

### Testing
CI

### API Changes
/

PR #622, opened 2025-03-19
Return mutex for arc messages

pjanevskiTT:
### Issue

While working on #616 I saw there is a mutex missing for arc messages which I removed in #566. This is not generally but could become if we have two processes sending different arc messages.

### Description

Use the arc message mutex created in cluster during the sending of arc messages

### List of the changes

- Use ARC message mutex in pcie_arc_msg

### Testing

CI

### API Changes
/


tt-vjovanovic (approved):

pjanevskiTT:
confirmed BH unit tests locally

Issue #523, opened 2025-02-13
Proper synchronization on TTDevice read/writes


pjanevskiTT:

broskoTT:
@pjanevskiTT More info here?

pjanevskiTT:
There was a bug with mutex in TTDevice, where we didn't clean it up properly so #524 was checked in to unblock metal. Basically issue here is just to add proper creation/cleaning of mutex in TTDevice. I am kinda hoping that this goes away by KMD TLB implementation, but if not we should take care of this just by mutex implementation, similar to how it's done in Cluster today.

pjanevskiTT:
This is probably going to be needed for #555 since part of the topology discovery for both WH and BH is done through TTDevice

pjanevskiTT:
Assigning @broskoTT since this may be on critical path for #555

broskoTT:
@pjanevskiTT can we close this now that locks are merged?

pjanevskiTT:
yes, closing it

PR #626, opened 2025-03-21
Add mutex to TTDevice for read and write

pjanevskiTT:
### Issue

Part of the effort towards #523 

### Description

Add mutex for read/writes for TTDevice based on the underlying PCI device

### List of the changes

- Add mutex for TTDevice read/writes io
- Add a multithreaded test

### Testing

CI + additional multithreaded test for TTDevice io

### API Changes
/

### DIFF
--- a/device/api/umd/device/tt_device/tt_device.h
+++ b/device/api/umd/device/tt_device/tt_device.h
@@ -6,6 +6,8 @@
 
 #pragma once
 
+#include <string_view>
+
 #include "umd/device/arc_messenger.h"
 #include "umd/device/architecture_implementation.h"
 #include "umd/device/pci_device.hpp"
@@ -177,9 +179,6 @@ private:
 
     std::shared_ptr<boost::interprocess::named_mutex> read_write_mutex = nullptr;
 
-    // Name of the mutex is SMALL_READ_WRITE_TLB because we need to be able to
-    // sync this TTDevice mutex with Cluster mutex for this same TLB. So final name
-    // of the mutex will be SMALL_READ_WRITE_TLB{pci_device_id} both in Cluster and TTDevice.
-    static constexpr char MUTEX_NAME[] = "SMALL_READ_WRITE_TLB";
+    static constexpr std::string_view MUTEX_NAME = "TT_SMALL_READ_WRITE_TLB";
 };
 }  // namespace tt::umd
--- a/device/api/umd/device/tt_device/tt_device.h
+++ b/device/api/umd/device/tt_device/tt_device.h
@@ -179,6 +179,6 @@ private:
 
     std::shared_ptr<boost::interprocess::named_mutex> read_write_mutex = nullptr;
 
-    static constexpr std::string_view MUTEX_NAME = "TT_SMALL_READ_WRITE_TLB";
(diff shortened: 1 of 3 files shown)


device/api/umd/device/tt_device/tt_device.h
```diff
@@ -166,5 +170,16 @@ class TTDevice {
...
+private:
+    void initialize_tt_device_mutex();
+    void clean_tt_device_mutex();
+
+    std::shared_ptr<boost::interprocess::named_mutex> read_write_mutex = nullptr;
+
+    // Name of the mutex is SMALL_READ_WRITE_TLB because we need to be able to
+    // sync this TTDevice mutex with Cluster mutex for this same TLB. So final name
+    // of the mutex will be SMALL_READ_WRITE_TLB{pci_device_id} both in Cluster and TTDevice.
+    static constexpr char MUTEX_NAME[] = "SMALL_READ_WRITE_TLB";
```
tt-vjovanovic, line 183: Use `std::string_view` instead of `char[]`.

device/tt_device/tt_device.cpp
```diff
@@ -379,4 +389,20 @@ uint32_t TTDevice::get_clock() {
...
+    permissions unrestricted_permissions;
+    unrestricted_permissions.set_unrestricted();
+    std::string mutex_name = TTDevice::MUTEX_NAME + std::to_string(get_pci_device()->get_device_num());
+    read_write_mutex = std::make_shared<named_mutex>(open_or_create, mutex_name.c_str(), unrestricted_permissions);
+}
+
+void TTDevice::clean_tt_device_mutex() {
+    std::string mutex_name = TTDevice::MUTEX_NAME + std::to_string(get_pci_device()->get_device_num());
+    read_write_mutex.reset();
+    read_write_mutex = nullptr;
```
tt-vjovanovic, line 402: No need to set it to `nullptr`, `.reset()` will do that for you.

device/api/umd/device/tt_device/tt_device.h
tt-vjovanovic: Also, consider having `TT_` in the mutex name :)

tt-vjovanovic (approved):

device/api/umd/device/tt_device/tt_device.h
pjanevskiTT: I'll revert TT naming here since mutexes in cluster are relying on TLB_NAME + something, which is the case with t device mutex as well here. TLB names don't start with TT_ and that would require much more changes that I don't think are priority right now. I will put this in an issue to address later

device/tt_device/tt_device.cpp
```diff
@@ -379,4 +389,19 @@ uint32_t TTDevice::get_clock() {
         "TTDevice is deleted.");
 }
 
+void TTDevice::initialize_tt_device_mutex() {
+    permissions unrestricted_permissions;
+    unrestricted_permissions.set_unrestricted();
```
broskoTT, line 394: I think we should have a class OurOwnMutex - with better naming of course, and just have it as member of this and other classes which need it... Not an action for this PR though, I might refactor it some day.

broskoTT (approved):

Issue #627, opened 2025-03-22
Ability to run ASan + TSan builds in CI


afuller-TT:
I'm getting TT-Metalium builds to have ASan and TSan runs in CI and as part of that we'll need TT-UMD to also be clean with both sanitizers.  It'll be much easier for you to have it in your own CI to catch issues early.

PR #628, opened 2025-03-22
Wire up ASan and TSan

afuller-TT:
### Issue
#627

### Description
Make it easy to run ASan and TSan builds.

### List of the changes
* Wire up ASan and TSan as a Build Type
* Delete the old bespoke variables that were setting those flags but were more unwieldy
* Update the CI configs so the build type can be selected either for a manual run or automatic workflows can select a build type

### Testing
Ran CI jobs

Notes:
* Non-Release build types (even Debug) are failing an assert here: https://github.com/tenstorrent/tt-umd/actions/runs/14003958381/job/39215282372#step:8:584 (release strips out asserts, so it wasn't seen in existing workflows which are all Release)
* On Multi-chip (ie: N300) there is a data race in write_to_non_mmio_device: https://github.com/tenstorrent/tt-umd/actions/runs/14003958381/job/39215282510#step:8:865

### API Changes
None


CMakeLists.txt
```diff
@@ -46,7 +46,7 @@ if(MASTER_PROJECT)
         set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Release build is the default" FORCE)
     endif()
 
-    include(sanitizer_options)
+    include(sanitizers)
```
broskoTT, line 49: So this will always be included for UMD build, which is not a build from tt_metal?
When this is included, asan/tsan will be enabled in all builds?
What does "-DCMAKE_BUILD_TYPE=" do in reference to this?

broskoTT (approved):
Thanks for improving our code infra, again! 
We'll look at the failures

broskoTT:
First failure that you mention is fixed in https://github.com/tenstorrent/tt-umd/pull/632
This data races might be more complex to fix, might need more time.

CMakeLists.txt
afuller-TT: It doesn't activate ASan or TSan.  It defines them as a choice you may select.

Just like Release, RelWithDebInfo, and Debug are all choices.  But any given build is only selecting one among them.  If you use an IDE that gives you the choice right int he GUI (eg: I think maybe CLion?  Certainly VS (non-code), KDevelop, XCode, and perhaps some others), then in your drop-down list you'll see a couple additional options: Release, RelWithDebInfo, Debug, ASan, TSan.

It's still up to the dev to choose which one that they wish to use for a particular build.

.github/workflows/build-and-run-all-tests.yml
```diff
@@ -3,6 +3,17 @@ name: Build and run all tests
...
+    inputs:
+      build-type:
+        required: false
+        default: Release
+        type: choice
+        options:
+          - Release
+          - RelWithDebInfo
+          - Debug
+          - ASan
```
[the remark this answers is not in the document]
afuller-TT: First: know thy enemey... or code... or something.

**Release** is optimized, but has all debug symbols utterly removed.
**RelWithDebInfo** is nearly as optimized, but the debug symbols aren't thrown away.  In a past life the slight additional optimization was inconsequential, and the ability to obtain a crash dump from the field was invaluable.  So in that scenario RelWithDebInfo was the hands-down winner and we pretended that Release didn't exist.
**Debug** is unoptimized and has debug symbols.  Unless you really need a pristine stack trace - better than RelWithDebInfo's which may have a bit of reordering - then this doesn't buy you much except for slower runtime.  No real harm in using it when scanning the code (eg: ClangTidy), since you aren't then actually running it.
**ASan** has the debug symbols, but also instruments for memory correctness and has a ~2x runtime penalty.  Any CI job that is "check that things are good" and don't need tight runtime tolerances can be ASan instead.
**TSan** is similar, but checks for concurrency issues.  It cannot be combined with ASan, so you'll need a parallel run for anything you wish to test with both.  The runtime penalty of TSan is ~10x (some estimates put it higher -- really dependent on the nature of the code being run).

In summary: No, there's no value in running pipelines with EVERY build type.  But correctness tests would benefit from running with ASan and again with TSan.  Performance runs should be Release or RelWithDebInfo.  That would cover the essentials, IMHO.

CMakeLists.txt
afuller-TT: The other part of your question:

> What does "-DCMAKE_BUILD_TYPE=" do in reference to this?

When a build is selected (eg: TSan) then the flags listed for TSan are the ones used.
example: when TSan is selected, and we're going to compile a .cpp file, then the value of `CMAKE_CXX_FLAGS_TSAN` is what is passed to the compiler.  When in ASan mode, then it's `..._ASAN` that is used instead.  Similarly for `..._RELWITHDEBINFO`, etc.

Issue #619, opened 2025-03-17
[NOC1] Soc descriptor NOC1 support


pjanevskiTT:
Support NOC1 coordinates in soc descriptor properly. #618 is introducing NOC1 to coordinate manager. Since coordinate manager is created through Soc descriptor, this path should be supported from tt_SocDescriptor class. There are two main things to support

- Support reading NOC0 to NOC1 mapping in soc descriptor yamls
- Support propagating the information of noc0 to noc1 mapping to coordinate manager creation

pjanevskiTT:
closed by #629

PR #629, opened 2025-03-24
Add NOC1 support to soc descriptor

pjanevskiTT:
### Issue

#619 

### Description
Impelement support for NOC1 coordinates in soc descriptor. Add default noc0 to noc1 x and y mappings, as well as mapping inside yaml files we have.

### List of the changes

- Add NOC0/1 x and y mapping field in SocDescInfo
- Add NOC0/1 x and y mapping in soc descriptor yamls we have
- Propagate into through soc descriptors to Coordinate managers
- Add tests

### Testing
CI. Additional tests for getting NOC1 coordinates from soc descriptor are added.

### API Changes
/


pjanevskiTT:
@tt-vjovanovic can you check noc0 to noc1 mappings for simulation soc descriptors, if you know how that should look like?

device/api/umd/device/tt_soc_descriptor.h
```diff
@@ -53,6 +53,8 @@ struct SocDescriptorInfo {
     uint32_t worker_l1_size;
```
broskoTT, line 53: Comment related to the previous PR.

Why do we have to have ETH_CORES_NOC1 and all other *_NOC1 constants? Shouldn't they be trivially constructed through ETH_CORES_NOC0 then NOC0_X_TO_NOC1_X ?

tests/soc_descs/blackhole_140_arch.yaml
```diff
@@ -49,6 +49,16 @@ router_only:
    8-1, 8-2, 8-3, 8-4, 8-5, 8-6, 8-7, 8-8, 8-9, 8-10, 8-11
   ]
 
+noc0_x_to_noc1_x:
```
broskoTT, line 52: Hmmm, so I'm not bought on this.

In my view, we have two sets of parameters.

First one is the set which is can't be modified once the hardware is produced. This set of parameters is generally defined in runtime by reading which generation our chip is (BH vs WH). For instance, TLB_BASE_INDEX_16M, TENSIX_SOFT_RESET_ADDR, etc.

Second ones are parameters which might be changed in a functional way on the same hardware, like tensixes used, dram used, tensix l1 size, harvesting.

So only the second one should go to yaml. We don't want to have something random in yaml which has to be fixed on a specific value, and won't be functional any other way.

I feel like these translations fall into the first category? These mappings can't be changed by FW or any other process, this is how the chip is produced?
The following conclusion from what I've said is probably that ARC and PCIE should also probably not be present in yamls at all, we should think about that as well.

I'd remove this from yaml. But if you don't remove it, bear in mind that you also might have to change it in tt_metal and simulators repos.

device/api/umd/device/tt_soc_descriptor.h
broskoTT: I see these are used in tests. Then the tests should define this constants probably

tests/api/test_soc_descriptor.cpp
```diff
@@ -449,3 +449,97 @@ TEST(SocDescriptor, BoardBasedPCIE) {
...
+                EXPECT_EQ(tensix_cores_noc1_yaml[core_coord_index].x, noc1_pair.x);
+                EXPECT_EQ(tensix_cores_noc1_yaml[core_coord_index].y, noc1_pair.y);
+                core_coord_index++;
+            }
+        }
+    }
+}
+
+TEST(SocDescriptor, BlackholeNOC1Cores) {
+    const size_t max_num_harvested_x = 14;
```
broskoTT, line 498: we have a constant for this somewhere I'm sure

tests/api/test_soc_descriptor.cpp
```diff
@@ -449,3 +449,97 @@ TEST(SocDescriptor, BoardBasedPCIE) {
...
+
+    for (size_t harvesting_mask = 0; harvesting_mask < (1 << max_num_harvested_x); harvesting_mask++) {
```
broskoTT, line 500: I feel like you're again making a mistake of reimplementing the code you're testing, in the test itself. By line 535 for example.

Can you define what is the value of testing for all possible harvesting masks? Is maybe the value for this that we know the code won't throw for all valid harvesting masks?

I'd instead hardcode a single harvesting value, and then also hardcode all the values below which you calculate to compare against.
I'll try to be clearer: When you have an EXPECT statement, you should not have calculated values on both sides. One of those values has to be a hardcoded constant, a referenced constant, or derived from constants but only with simple math operations. For example testing noc1_pair could be reduced to comparing against a hardcoded values for some values which are obvious to calculate, like the corners of our chip.

tests/api/test_soc_descriptor.cpp
pjanevskiTT: test changed, no need for this constant

broskoTT (approved):

PR #631, opened 2025-03-24
Add riscv64/libcreate_ethernet_map.a

joelsmithTT:
This was built in a Ubuntu 24.04.2 environment under riscv64.

### DIFF
--- a/device/api/umd/device/tt_device/tt_device.h
+++ b/device/api/umd/device/tt_device/tt_device.h
@@ -6,6 +6,8 @@
 
 #pragma once
 
+#include <string_view>
+
 #include "umd/device/arc_messenger.h"
 #include "umd/device/architecture_implementation.h"
 #include "umd/device/pci_device.hpp"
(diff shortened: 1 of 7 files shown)


joelsmithTT:
I know we want to get rid of this, but for now...

device/libs/riscv64/libcreate_ethernet_map.a
broskoTT, line 1: I don't see a point in adding this without adding cmake configuration for building for riscv. Are those changes complex or trivial?

pjanevskiTT: https://github.com/tenstorrent/tt-umd/blob/main/device/CMakeLists.txt#L131 should take care of it I think

pjanevskiTT (approved):

device/libs/riscv64/libcreate_ethernet_map.a
joelsmithTT: @pjanevskiTT is correct - adding the library file is sufficient to build UMD under riscv64.

Successfully _using_ UMD under riscv64 is a different story, but the considerations here are platform-specific and require KMD work too.

Issue #555, opened 2025-02-28
Support multiple UMD processes


broskoTT:
After syncing with tt-fabric team, we clarified their requirements from us.

The tools (which use Luwen) running alongside ai workload (through tt-metal and umd) is not a requirement.

The scenario is as follows: 
There will be a health check process long running in the background, which will be built somewhere in stack, but on top of UMD. (TBD: Thing whether Cluster or TTDevice is a better entry point for this).
Alongside this process, user will turn workload on and off, e.g. another UMD process will spawn up on the same set of chips.

These two should work without stepping on each other's toes. Making this functional anyway possible is a priority. Complete security is of lower priority, since clients are assumed not to be malicious. For example, if both processes start programming same TLBs, that would crash the device or workload, and that is not guaranteed with this implementation.

Further work is planned, in #462 which has this goal, so we don't need another implementation for that in the meantime.

As Aditya pointed out, currently multiple UMD processes are possible, but only if one process creates and initialises UMD driver and then forks several processes from it. This doesn't exactly match the scenario requested though.

A proposal is to make a test, or a set of tests, spawning multiple UMD processes and stress testing communication with the chip/chips and verifying nothing crashes. And then doing whatever needed to make it work.

FYI @pjanevskiTT 
Please let me know if I misunderstood your scenario or requirements, FYI @tt-asaigal @abhullar-tt @aliuTT

pjanevskiTT:
I want to add that supporting #523 is probably going to be part of this effort if we want two UMD processes doing the topology discover in parallel

pjanevskiTT:
TODO: address code organization comments https://github.com/tenstorrent/tt-umd/pull/624#discussion_r2009278971 when Cluster mutexes are consolidated to be used together with TTDevice and ArcMessenger

pjanevskiTT:
@broskoTT can we close this as feature done and focus on bugs in #654 ?

pjanevskiTT:
closed by #647

PR #647, opened 2025-03-27
Support multiple UMD processes

pjanevskiTT:
### Issue

#555 

### Description

Final changes to support UMD multiprocesses. Mutex for create ethernet map is added, for syncing multiple create eth maps running in parallel. There is still a case when create-ethernet-map and UMD can corrupt each other on TLBs and ARC messages, but that should happen very rare. Without rewriting create-ethernet-map we don't have a way to stop this. Tests in multiprocess should show the use cases that are supported for running multiple UMDs.

### List of the changes

- Add mutex for create ethernet map
- Add tests to show the usecases
- Add multiprocess tests to CI

### Testing
CI

### API Changes
/


tests/unified/multiprocess.cpp
```diff
@@ -0,0 +1,261 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent Inc.
```
pjanevskiTT, line 1: confirm tests are passing on Blackhole locally as well

broskoTT (approved):

tests/unified/multiprocess.cpp
pjanevskiTT: confirmed

Issue #418, opened 2024-12-18
LockingManager class


broskoTT:
As working towards #248 and specifically related to TLBs, the need for this class naturally arises.
This class should wrap all os calls related to locking, and provide synchronisation capabilities to other classes.

pjanevskiTT:
Useful comment https://github.com/tenstorrent/tt-umd/pull/637#discussion_r2013561731

broskoTT:
This is in a good state now

PR #658, opened 2025-03-31
LockManager

broskoTT:
### Issue
#418 

### Description
We've increased usage of locking throughout our code. This is just gathering all lock related code to a single class.

### List of the changes
- Introduced LockManager class. It offers initialization, clearing and get_mutex
- Changed all usages throughout code to go through LockManager class.

### Testing
Existing CI tests.

### API Changes


device/api/umd/device/lock_manager.h
```diff
@@ -0,0 +1,62 @@
...
+    static void initialize_default_chip_mutexes(TTDevice* tt_device, TLBManager* tlb_manager, const bool clear_mutex);
+
+private:
+    static void initialize_mutex_internal(const std::string& mutex_name, const bool clear_mutex);
+
+    static void clear_mutex_internal(const std::string& mutex_name);
+
+    static std::unique_lock<boost::interprocess::named_mutex> get_mutex_internal(const std::string& mutex_name);
+
+    static const std::unordered_map<MutexType, std::string> MutexTypeToString;
```
pjanevskiTT, line 65: nit: I think all our variables, even static, are in snake_case. I think there is no need to change it here

device/cluster.cpp
```diff
@@ -210,8 +174,12 @@ void Cluster::create_device(
                 pci_device->get_device_num(),
                 pci_device->get_device_num());
 
-            // TODO: This will be moved to a dedicated Locking class.
-            initialize_interprocess_mutexes(logical_device_id, clean_system_resources);
+            // These mutexes are intended to be based on physical devices/pci-intf not logical. Set these up ahead of
+            // time here (during device init) since its unsafe to modify shared state during multithreaded runtime.
```
pjanevskiTT, line 178: nit spelling: rename its -> it is or it's :)

device/api/umd/device/lock_manager.h
```diff
@@ -0,0 +1,72 @@
...
+    static void clear_mutex(MutexType mutex_type, TTDevice* tt_device);
+    static std::unique_lock<boost::interprocess::named_mutex> get_mutex(MutexType mutex_type, TTDevice* tt_device);
+
+    // This set of functions is used to manage mutexes which are chip specific. This variant accepts custom mutex name.
+    static void initialize_mutex(std::string mutex_prefix, TTDevice* tt_device, const bool clear_mutex);
+    static void clear_mutex(std::string mutex_prefix, TTDevice* tt_device);
+    static std::unique_lock<boost::interprocess::named_mutex> get_mutex(std::string mutex_prefix, TTDevice* tt_device);
+
+    // Commonly used set of mutexes for a chip.
+    static void initialize_default_chip_mutexes(TTDevice* tt_device, TLBManager* tlb_manager, const bool clear_mutex);
```
pjanevskiTT, line 57: I think this class should not have dependency on TTDevice and TLBManager. I think each class that wants to create mutexes should have it's own logic how to create mutexes and then call into simple API that is provided by this class. To be more precise what I am saying, instead of having initialize_default_chip_mutexes. I think better solution would be to have smh like this

```cpp

class LockManager {
     void initialize_mutex(string s, ...);
}

class TLBManager {
    void initialize_tlb_mutexes() {
          for (std::string tlb : tlbs) {
               LockManager::initialize_mutex(tlb + to_string(get_tt_device()->get_pci_device()->get_device_num()));
          } 
    }
}

```

Since in the future some other class is going to need mutex and I don't think it should have it's own function inside LockManager.

pjanevskiTT: I think other functions that accept TTDevice should accept pci_device_id instead so it could simply concatenate string of the mutex type and pci device id

pjanevskiTT (commented):
Replacement of mutex code looks good, I have some remarks about LockManager below. I think API should be changed a bit to remove dependency on other classes like TTDevice and TLBManager

device/lock_manager.cpp
```diff
@@ -0,0 +1,135 @@
...
+    {MutexType::MEM_BARRIER, "TT_MEM_BARRIER"},
+    {MutexType::CREATE_ETH_MAP, "CREATE_ETH_MAP"},
+};
+
+std::unordered_map<std::string, std::unique_ptr<boost::interprocess::named_mutex>> LockManager::mutexes;
+
+LockManager::~LockManager() {
+    // Clear out all mutexes and unlock them.
+    while (!mutexes.empty()) {
+        clear_mutex_internal(mutexes.begin()->first);
```
broskoTT, line 31: Yeah, destructor isn't ever called :)

device/api/umd/device/lock_manager.h
```diff
@@ -0,0 +1,72 @@
+/*
```
pjanevskiTT, line 1: since multiprocess tests are now disabled in CI, make sure to run it locally before merging. Make sure there are no new failures except the ones we were seeing

device/api/umd/device/lock_manager.h
broskoTT: Removed dependancies

device/api/umd/device/cluster.h
```diff
@@ -1093,19 +1089,22 @@ class Cluster : public tt_device {
         chip_id_t chip_id,
         tt_ClusterDescriptor* cluster_desc,
         tt_SocDescriptor& soc_desc,
+        const bool clean_system_resources,
```
pjanevskiTT, line 1092: we should maybe investigate whether there is a use case for this being set to both true and false. No action for this PR

pjanevskiTT (approved):
looks great now

PR #660, opened 2025-03-31
Move read/write to device to Chip

broskoTT:
### Issue
Contributing towards #248

### Description
Move local chip communication from Cluster to Chip

### List of the changes
- Moved read/write device to LocalChip
- Set some members in chip protected instead of private

### Testing
Existing CI tests

### API Changes

### DIFF
--- a/device/api/umd/device/chip/chip.h
+++ b/device/api/umd/device/chip/chip.h
@@ -41,6 +41,12 @@ public:
     virtual void write_to_sysmem(uint16_t channel, const void* src, uint64_t sysmem_dest, uint32_t size);
     virtual void read_from_sysmem(uint16_t channel, void* dest, uint64_t sysmem_src, uint32_t size);
 
+    // TODO: Currently works only for Local and not for Remote.
+    virtual void write_to_device(
+        tt_xy_pair core, const void* src, uint64_t l1_dest, uint32_t size, const std::string& fallback_tlb);
+    virtual void read_from_device(
+        tt_xy_pair core, void* dest, uint64_t l1_src, uint32_t size, const std::string& fallback_tlb);
+
     // TODO: To be removed once all usages are moved inside local chip.
     virtual std::unique_lock<boost::interprocess::named_mutex> get_mutex(std::string mutex_name, int pci_device_id);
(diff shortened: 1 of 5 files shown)


device/api/umd/device/chip/chip.h
```diff
@@ -41,6 +41,12 @@ class Chip {
     virtual void write_to_sysmem(uint16_t channel, const void* src, uint64_t sysmem_dest, uint32_t size);
     virtual void read_from_sysmem(uint16_t channel, void* dest, uint64_t sysmem_src, uint32_t size);
 
+    // TODO: Currently works only for Local and not for Remote.
```
pjanevskiTT, line 44: maybe we should have an issue for this together with a comment, seems like an important think to take care of

device/chip/local_chip.cpp
```diff
@@ -137,6 +142,105 @@ void LocalChip::read_from_sysmem(uint16_t channel, void* dest, uint64_t sysmem_s
...
+            size -= transfer_size;
+            l1_dest += transfer_size;
+            buffer_addr += transfer_size;
+        }
+        log_debug(LogSiliconDriver, "Write done Dynamic TLB with pid={}", (long)getpid());
+    }
+}
+
+void LocalChip::read_from_device(
+    tt_xy_pair core, void* dest, uint64_t l1_src, uint32_t size, const std::string& fallback_tlb) {
```
pjanevskiTT, line 190: nit: I guess all regs read/writes are going through read_mmio_device_register in cluster, so l1_src is proper naming, but just confirm that before merging this

pjanevskiTT (approved):

device/api/umd/device/chip/chip.h
broskoTT: But in the next following PRs I'll make it work for both local and remote chips, this is just a comment in the meantime. We should have Chip::write_to_device which works for all chips

pjanevskiTT: makes sense, keep it as it is

device/chip/local_chip.cpp
broskoTT: Yeah that's true. That one should also be moved to local_chip...

PR #667, opened 2025-04-01
Move remote eth core choice to local_chip

broskoTT:
### Issue
#248 

### Description
On a path of moving remote communication out of Cluster

### List of the changes
- Moved remote_transfer_ethernet_cores and active_core into Local chip. This will make it easier to move the code further.
- Removed epoch command dedicated eth cores.
- Changed non_mmio_io accordingly

### Testing
Existing CI tests

### API Changes


device/api/umd/device/cluster.h
```diff
@@ -1153,23 +1153,7 @@ class Cluster : public tt_device {
 
     std::shared_ptr<tt_ClusterDescriptor> cluster_desc;
 
-    // remote eth transfer setup
-    static constexpr std::uint32_t NUM_ETH_CORES_FOR_NON_MMIO_TRANSFERS = 6;
```
pjanevskiTT, line 1157: can you clean RemoteCommunication class as well from this legacy stuff?

device/chip/local_chip.cpp
```diff
@@ -139,6 +147,44 @@ void LocalChip::read_from_sysmem(uint16_t channel, void* dest, uint64_t sysmem_s
...
 }
 
+void LocalChip::set_remote_transfer_ethernet_cores(const std::unordered_set<CoreCoord>& active_eth_cores_per_chip) {
+    // Makes UMD aware of which ethernet cores have active links.
+    // Based on this information, UMD determines which ethernet cores can be used for host->cluster non-MMIO transfers.
+    // This overrides the default ethernet cores tagged for host to cluster routing in the constructor and must be
+    // called for all MMIO devices, if default behaviour is not desired.
+    log_assert(soc_descriptor_.arch == tt::ARCH::WORMHOLE_B0, "{} can only be called for Wormhole arch", __FUNCTION__);
+    // Cores 0, 1, 6, 7 are only available if in the active set
+    static std::unordered_set<tt_xy_pair> eth_cores_available_if_active = {
```
pjanevskiTT, line 157: this should go away with rewriting create_ethernet_map. No action for this PR

pjanevskiTT (approved):
As an isolated change, I am not sure moving all this functions to chip class makes sense, since this is something that is very coupled with LocalChip only, not sure it should be exposed through Chip interface. Having said that, I get where this is going towards having read/write in RemoteChip as well, so I guess this is going to help. I really hope all the functions that are moved in this PR to Chip go away as soon as we are able to have create-ethernet-map from source. Approved :)

device/api/umd/device/cluster.h
broskoTT: I did something, but all of this will be replaced with the original code from Cluster once I manage to extract it out.

Issue #571, opened 2025-03-04
Refactor tests


broskoTT:
Consolidate tests. Also leave a README which explains our structure of tests.

api - these should not perform any unit testing. These are example codes.
gs/wh/bh - arch specfic tests
unit - or some other name - unified, arch agnostic tests

broskoTT:
In addition:
- api/unit tests should not be arch specific, and should not have l1_address_map.h available.
- Figure out test namings, make them uniform
- Extract constants and commonly used code like writing our popular 1,2 ... 10 vector to a common test infra file.

broskoTT:
Also think about a folder with "benchmark" tests
There is something on brosko/lock_test branch

pjanevskiTT:
I think we should also have a group of tests to verify low-level device stuff. For example DRAM speed, eth speed, AICLK...all the values of the subsystems that we believe to be true. Basically the test group should say, ok the card is usable and stable. There was a requirement from syseng folks to provide groups of tests similar to this

broskoTT:
Update https://github.com/tenstorrent/tt-metal/issues/24359 if needed once this is done

PR #671, opened 2025-04-02
Add bare metal CI runs

broskoTT:
### Issue
No issue, but slightly related to #571 since I want to introduce offline unit tests
In the meantime, this issue popped up which this PR fixes: https://github.com/tenstorrent/tt-metal/issues/18087

### Description
Add baremetal machine to CI for running tests, and fix all issues with creating Cluster object on a system without chips.

### List of the changes
- add baremetal to workflow
- Fix Cluster so it doesn't throw if no chips are present
- Alter tests to skip/don't skip based on if chips are present.

### Testing
CI tests

### API Changes

### DIFF
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -111,9 +111,6 @@ void Cluster::create_device(
     // Don't buffer stdout.
     setbuf(stdout, NULL);
 
-    log_assert(
-        target_mmio_device_ids.size() > 0, "Must provide set of target_mmio_device_ids to Cluster constructor now.");
-
     for (const chip_id_t& logical_device_id : target_mmio_device_ids) {
         if (!create_mock_chips) {
             bool hugepages_initialized =
@@ -140,6 +137,9 @@ void Cluster::create_device(
 }
 
 void Cluster::construct_cluster(const uint32_t& num_host_mem_ch_per_mmio_device, const bool create_mock_chips) {
+    // TODO: work on removing this member altogether. Currently assumes all have the same arch.
+    arch_name = chips_.empty() ? tt::ARCH::Invalid : chips_.begin()->second->get_soc_descriptor().arch;
+
     if (!create_mock_chips) {
         auto available_device_ids = detect_available_device_ids();
         log_info(LogSiliconDriver, "Detected PCI devices: {}", available_device_ids);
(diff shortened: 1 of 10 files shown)


device/cluster.cpp
```diff
@@ -181,7 +181,7 @@ void Cluster::construct_cluster(const uint32_t& num_host_mem_ch_per_mmio_device,
 
     // Disable dependency to ethernet firmware for all BH devices and WH devices with all chips having MMIO (e.g. UBB
     // Galaxy), do not disable for N150, was seeing some issues in CI
-    if (remote_chip_ids_.empty() and cluster_desc->get_board_type(*local_chip_ids_.begin()) != BoardType::N150) {
```
pjanevskiTT, line 184: wasn't this added in #487 for 6u galaxy? Not sure if this fix is ok in that case

tests/api/test_cluster.cpp
```diff
@@ -346,6 +312,10 @@ TEST(TestCluster, TestClusterNocId) {
     // Some default values are set during construction of UMD, but you can override them.
     cluster->set_barrier_address_params({L1_BARRIER_BASE, ETH_BARRIER_BASE, DRAM_BARRIER_BASE});
 
+    if (cluster->get_target_device_ids().empty()) {
```
pjanevskiTT, line 315: we should probably aim to have all this TestCluster tests inside the common fixture, something similar to https://github.com/tenstorrent/tt-umd/blob/main/tests/wormhole/test_umd_remote_api_stability.cpp#L26 so we don't have to duplicate the code for skipping the test. No action for this PR, put it in a follow up issue

pjanevskiTT (commented):
Changes look good, can you just verify 6u change before approve?

device/cluster.cpp
broskoTT: Yeah but it doesn't make sense.
Anyway, this is modified in a separate PR, which I'll complete before this one. Once I figure out what's failing
https://github.com/tenstorrent/tt-umd/pull/662

tests/api/test_cluster.cpp
broskoTT: Valid idea for further refactoring

device/cluster.cpp
pjanevskiTT: ok, just wanted to make sure 6u stuff works :). Approving

pjanevskiTT (approved):

tests/api/test_cluster.cpp
broskoTT: Added to effort https://github.com/tenstorrent/tt-umd/issues/571

PR #679, opened 2025-04-03
Move read/write reg to device to Chip

broskoTT:
### Issue
#248 

### Description
Moving functions for reading and writing regs.
These will be surfaced to Cluster in the same form, when we remove fallback_tlb argument everywhere.

### List of the changes
- Moved read_mmio_device_register and write_mmio_device_register to Chip
- Renamed to read_from_device_reg and write_to_device_reg

### Testing
Existing CI tests

### API Changes


pjanevskiTT (approved):
nice :)

device/chip/local_chip.cpp
```diff
@@ -244,6 +270,41 @@ void LocalChip::read_from_device(
...
+    log_debug(LogSiliconDriver, "  dynamic tlb_index: {}", tlb_index);
+
+    auto [mapped_address, tlb_size] = tt_device_->set_dynamic_tlb(
+        tlb_index, translate_chip_coord_virtual_to_translated(core), reg_dest, tt::umd::tlb_data::Strict);
+    // Align block to 4bytes if needed.
+    auto aligned_buf = tt_4_byte_aligned_buffer(src, size);
+    if (aligned_buf.input_size != aligned_buf.block_size) {
+        // Copy value from main buffer to aligned buffer
+        std::memcpy(aligned_buf.local_storage, src, size);
+    }
```
joelsmithTT, line 286: I think there is a lot wrong with this code.

First, it doesn't deal with alignment.  It pads.
Second, if the size is e.g. 2 bytes, it writes 4 bytes to the chip.  The two extra bytes are uninitialized.

Since the case of `size % 4 != 0` appears to be completely broken, my advice is to eliminate this tt_4_byte_aligned_buffer entirely, reject sizes that are not a multiple of 4, and since this code path is ostensibly writing hardware registers, also reject addresses that are not a multiple of 4, too.

joelsmithTT: P.S. I know this already existed and is just getting moved.

device/chip/local_chip.cpp
```diff
@@ -244,6 +270,41 @@ void LocalChip::read_from_device(
...
+    tt_device_->write_regs(mapped_address, aligned_buf.block_size / sizeof(uint32_t), aligned_buf.local_storage);
+}
+
+void LocalChip::read_from_device_reg(
+    tt_xy_pair core, void* dest, uint64_t reg_src, uint32_t size, const std::string& fallback_tlb) {
+    const auto tlb_index = tlb_manager_->dynamic_tlb_config_.at(fallback_tlb);
+    auto lock = lock_manager_.get_mutex(fallback_tlb, tt_device_->get_pci_device()->get_device_num());
```
joelsmithTT, line 293: This looks incorrect.  It is not clear from reading this function whether this actually acquires and holds a lock, or if it is just getting a mutex and forgetting to lock it.  I looked at the implementation and while it appears to be technically valid, I think the name "get_mutex" is not right and should be re-named to "acquire_lock".

broskoTT: Agree. I'll fix the naming.

broskoTT: #688

device/chip/local_chip.cpp
broskoTT: Out of scope for this PR, since it already existed as you say.
Thanks for reviewing this, I created an issue to address this: #689

PR #682, opened 2025-04-03
Move read_from_non_mmio_device to RemoteChip

broskoTT:
### Issue
On a path of #248

### Description
In this PR, remote read communication is moved to RemoteChip.
There's work on:
1. Initial changes to accommodate RemoteChip to even do what's needed
2. Actual move of read and write to RemoteChip
3. Additional complexities due to a fact that remote write can be done on a LocalChip in case of ethernet broadcast.

1+2+3 seemed too much for a PR, so instead I did just 1+2 by moving only the read part, and 2+3 will be done in a next part where I'll move the write part.

### List of the changes
- RemoteCommunication now accepts LocalChip instead of TTDevice
- Copied the read code from cluster.cpp instead of existing RemoteCommunication's hacked read code
- In RemoteCommunication::write_non_mmio I did slight changes due to now having LocalChip instead of TTDevice and LockManager.
- It only made sense that get_local_chip and get_remote_chip actually return the right classes. I tried other options but I liked this one the most. In any case, those two calls should not be used a lot (in the final code where everywthing is moved to chip classes). Added get_remote_chip call.
- RemoteChip can now do a read through read_from_device, same api as is for LocalChip. This will get unified in Cluster in some of the following PRs. Also, fallback_tlbs will be removed in some of the following PRs
- Slightly changed the order of arguments in read_non_mmio so that first the target is given, then addresses, then size. Note that eth core used is not passed anymore, but it is fetched from the local_chip_. This change was actually done in https://github.com/tenstorrent/tt-umd/pull/667

### Testing
Existing CI tests.
Also tt_metal tests, since I'm uncertain on the scope of our remote tests:

### API Changes

### DIFF
--- a/device/api/umd/device/chip/chip.h
+++ b/device/api/umd/device/chip/chip.h
@@ -44,6 +44,7 @@ public:
     virtual void read_from_sysmem(uint16_t channel, void* dest, uint64_t sysmem_src, uint32_t size);
 
     // TODO: Currently works only for Local and not for Remote.
+    // Both write and read cores are defined in VIRTUAL coords.
     virtual void write_to_device(
         tt_xy_pair core, const void* src, uint64_t l1_dest, uint32_t size, const std::string& fallback_tlb);
     virtual void read_from_device(
--- a/device/api/umd/device/chip/remote_chip.h
+++ b/device/api/umd/device/chip/remote_chip.h
@@ -21,6 +21,8 @@ public:
         tt_xy_pair core, void* dest, uint64_t l1_src, uint32_t size, const std::string& fallback_tlb) override;
 
 private:
+    tt_xy_pair translate_chip_coord_virtual_to_translated(const tt_xy_pair core);
+
     eth_coord_t eth_chip_location_;
     std::unique_ptr<RemoteCommunication> remote_communication_;
 };
--- a/device/api/umd/device/chip/remote_chip.h
+++ b/device/api/umd/device/chip/remote_chip.h
@@ -7,11 +7,21 @@
 #pragma once
 
 #include "umd/device/chip/chip.h"
+#include "umd/device/remote_communication.h"
 
 namespace tt::umd {
+class LocalChip;
+
 class RemoteChip : public Chip {
 public:
-    RemoteChip(tt_SocDescriptor soc_descriptor);
+    RemoteChip(tt_SocDescriptor soc_descriptor, eth_coord_t eth_chip_location, LocalChip* local_chip);
     bool is_mmio_capable() const override;
+
+    void read_from_device(
+        tt_xy_pair core, void* dest, uint64_t l1_src, uint32_t size, const std::string& fallback_tlb) override;
+
+private:
+    eth_coord_t eth_chip_location_;
+    std::unique_ptr<RemoteCommunication> remote_communication_;
 };
 }  // namespace tt::umd
--- a/device/api/umd/device/remote_communication.h
+++ b/device/api/umd/device/remote_communication.h
@@ -16,6 +16,7 @@ public:
     RemoteCommunication(LocalChip* local_chip);
     virtual ~RemoteCommunication();
 
+    // Target core should be in translated coords.
     void read_non_mmio(
         eth_coord_t target_chip, tt_xy_pair target_core, void* dest, uint64_t core_src, uint32_t size_in_bytes);
 
--- a/device/api/umd/device/remote_communication.h
+++ b/device/api/umd/device/remote_communication.h
@@ -9,18 +9,15 @@
 
 namespace tt::umd {
 
+class LocalChip;
+
 class RemoteCommunication {
 public:
-    RemoteCommunication(TTDevice* tt_device);
+    RemoteCommunication(LocalChip* local_chip);
     virtual ~RemoteCommunication();
 
     void read_non_mmio(
-        uint8_t* mem_ptr,
-        tt_xy_pair core,
-        uint64_t address,
-        uint32_t size_in_bytes,
-        eth_coord_t target_chip,
-        const tt_xy_pair eth_core);
+        eth_coord_t target_chip, tt_xy_pair target_core, void* dest, uint64_t core_src, uint32_t size_in_bytes);
 
     void write_to_non_mmio(
         uint8_t* mem_ptr,
@@ -33,8 +30,7 @@ public:
     void wait_for_non_mmio_flush(std::vector<tt_xy_pair> remote_transfer_eth_cores);
 
 private:
-    TTDevice* tt_device;
-    LockManager lock_manager;
+    LocalChip* local_chip_;
 };
 
 }  // namespace tt::umd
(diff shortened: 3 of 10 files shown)


device/api/umd/device/chip/remote_chip.h
```diff
@@ -7,11 +7,21 @@
...
+class LocalChip;
+
 class RemoteChip : public Chip {
 public:
-    RemoteChip(tt_SocDescriptor soc_descriptor);
+    RemoteChip(tt_SocDescriptor soc_descriptor, eth_coord_t eth_chip_location, LocalChip* local_chip);
     bool is_mmio_capable() const override;
+
+    void read_from_device(
+        tt_xy_pair core, void* dest, uint64_t l1_src, uint32_t size, const std::string& fallback_tlb) override;
```
pjanevskiTT, line 21: should we add CoreCoord API as well here?

device/api/umd/device/remote_communication.h
```diff
@@ -13,18 +13,15 @@ class named_mutex;
...
     virtual ~RemoteCommunication();
 
     void read_non_mmio(
-        uint8_t* mem_ptr,
-        tt_xy_pair core,
-        uint64_t address,
-        uint32_t size_in_bytes,
-        eth_coord_t target_chip,
-        const tt_xy_pair eth_core);
+        eth_coord_t target_chip, tt_xy_pair target_core, void* dest, uint64_t core_src, uint32_t size_in_bytes);
```
pjanevskiTT, line 24: nit: rename core_src to address or l1_src, to match local chip read and writes

device/api/umd/device/remote_communication.h
```diff
@@ -13,18 +13,15 @@ class named_mutex;
...
 
+class LocalChip;
+
 class RemoteCommunication {
 public:
-    RemoteCommunication(TTDevice* tt_device);
+    RemoteCommunication(LocalChip* local_chip);
     virtual ~RemoteCommunication();
 
     void read_non_mmio(
```
pjanevskiTT, line 23: not an action for this PR, but I think we should standardize our read/write function definitions. For example I would like to see something like

```
chip->read(ptr, size, addr, core, ...<additional args>);
// same for write
```
I don't really care about the order, I would just like each read/write function in any class to respect it. If you agree you can open an issue about it

device/remote_communication.cpp
```diff
@@ -33,31 +36,27 @@ struct routing_cmd_t {
...
     using data_word_t = uint32_t;
     constexpr int DATA_WORD_SIZE = sizeof(data_word_t);
+    std::string write_tlb = "LARGE_WRITE_TLB";
+    std::string read_tlb = "LARGE_READ_TLB";
+    std::string empty_tlb = "";
 
-    tt_xy_pair translated_core = core;
-    core.x = translated_core.x;
-    core.y = translated_core.y;
+    auto translated_coord = local_chip_->get_soc_descriptor().translate_coord_to(
```
pjanevskiTT, line 51: two things here. First, I think it's a bug to use local_chip translations, we should use remote chip here?
Second, this translation based on umd_use_noc1 is problematic in itself, I added a fix in #684. So I would probably wait for that fix to land in main, we can expose that API through chip and then you use remote chip translation here

pjanevskiTT (commented):
Left few comments. I would like to see the test before approving as well. If you think more Cluster functions should be moved in order to write a test, that is probably fine as well

device/api/umd/device/remote_communication.h
broskoTT: I'm already trying to do that.
The logic is as following:
First define the destination core location, target_chip, target_core, etc
Then define src, dest in case of write, or dest,src in case of read
Then define the size of the transfer,
Then everything else

device/api/umd/device/remote_communication.h
broskoTT: But this is not just l1, this was intentional. It can be a register read/write

device/api/umd/device/chip/remote_chip.h
broskoTT: I agree.
I think we should try to use CoreCoord everywhere internally up to the point of going to the noc, or to TTDevice.
I think there are other places where this is true. I created an issue to fix that, but I'd say it's out of scope for this PR
#687

device/remote_communication.cpp
broskoTT: When I fixed this the tests started passing, so that might've been the bug :)

device/api/umd/device/chip/chip.h
```diff
@@ -44,6 +44,7 @@ class Chip {
     virtual void read_from_sysmem(uint16_t channel, void* dest, uint64_t sysmem_src, uint32_t size);
```
broskoTT, line 44: Dependant change got reverted in #713, so I can't merge this until that is resolved. It introduced hangs on TG, so make sure to check this and following changes on TG

broskoTT:
All tt_metal tests passed on https://github.com/tenstorrent/tt-umd/pull/686, 
So I'm merging the whole stack of changes:
https://github.com/tenstorrent/tt-umd/pull/714
https://github.com/tenstorrent/tt-umd/pull/682
https://github.com/tenstorrent/tt-umd/pull/685
https://github.com/tenstorrent/tt-umd/pull/686

device/cluster.cpp
```diff
@@ -175,7 +175,10 @@ std::unique_ptr<Chip> Cluster::construct_chip_from_cluster(
         return std::make_unique<LocalChip>(
             soc_desc, cluster_desc->get_chips_with_mmio().at(chip_id), num_host_mem_channels, clean_system_resources);
     } else {
-        return std::make_unique<RemoteChip>(soc_desc);
+        return std::make_unique<RemoteChip>(
```
pjanevskiTT, line 178: nit: just a suggestion to add assert on architecture to be able to make RemoteChip only for Wormhole, since eth coordinates are not available otherwise. Not sure it should be added, but it may be helpful in the future to catch potential bugs

device/cluster.cpp
```diff
@@ -404,7 +407,19 @@ Cluster::Cluster(
     std::unordered_map<chip_id_t, HarvestingMasks> simulated_harvesting_masks) {
     cluster_desc = Cluster::create_cluster_descriptor();
 
-    for (auto& chip_id : cluster_desc->get_all_chips()) {
+    std::vector<chip_id_t> all_chips_local_first;
```
pjanevskiTT, line 410: why is this needed?

device/cluster.cpp
```diff
@@ -459,7 +486,19 @@ Cluster::Cluster(
     std::unordered_map<chip_id_t, HarvestingMasks> simulated_harvesting_masks) {
     cluster_desc = Cluster::create_cluster_descriptor(sdesc_path);
 
-    for (auto& chip_id : target_devices) {
+    std::vector<chip_id_t> all_chips_local_first;
```
pjanevskiTT, line 489: having this exact code at 3 or more places (each constructor), I think this would benefit from a function

pjanevskiTT (approved):
looks good

device/cluster.cpp
broskoTT: You have to initialize the local chips first, otherwise initializing remotechips based on local ones will fail. 
Get_all_chips returns an unordered_set, so the order of traversal is not defined

PR #685, opened 2025-04-04
Move wait_for_non_mmio_flush to RemoteChip and RemoteCommunication

broskoTT:
### Issue
On a way to #248

### Description
Next one after #682 , before moving write_non_mmio

### List of the changes
- flush_non_mmio_per_chip moved from Cluster to LocalChip
- wait_for_non_mmio_flush implementation moved to RemoteCommunication. This API is also surfaced in RemoteChip
- Changed cluster.h calls to accommodate new API

### Testing
Existing CI tests

### API Changes

### DIFF
--- a/device/chip/local_chip.cpp
+++ b/device/chip/local_chip.cpp
@@ -276,6 +276,14 @@ void LocalChip::read_from_device_reg(
     tt_device_->read_regs(mapped_address, size / sizeof(uint32_t), dest);
 }
 
+void LocalChip::wait_for_non_mmio_flush() {
+    // This is a local chip, so no need to flush remote communication.
+}
+
+void LocalChip::set_flush_non_mmio(bool flush_non_mmio) { flush_non_mmio_ = flush_non_mmio; }
+
+bool LocalChip::get_flush_non_mmio() const { return flush_non_mmio_; }
+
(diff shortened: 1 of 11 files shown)


device/chip/local_chip.cpp
```diff
@@ -305,6 +305,14 @@ void LocalChip::read_from_device_reg(
     }
 }
 
+void LocalChip::wait_for_non_mmio_flush() {
+    // This is a local chip, so no need to flush remote communication.
```
pjanevskiTT, line 309: I just want to comment on this function being empty, since it looks a bit weird. I know cluster was exposing wait_for_non_mmio_flush just for remote chip ids, but this function is actually flushing the command queues on local chip, so maybe we should just think about how to expose this differently, rename the function etc... No action for this PR

pjanevskiTT (approved):

broskoTT:
All tt_metal tests passed on https://github.com/tenstorrent/tt-umd/pull/686, 
So I'm merging the whole stack of changes:
https://github.com/tenstorrent/tt-umd/pull/714
https://github.com/tenstorrent/tt-umd/pull/682
https://github.com/tenstorrent/tt-umd/pull/685
https://github.com/tenstorrent/tt-umd/pull/686

Issue #654, opened 2025-03-29
UMD multiprocess tests flaky


broskoTT:
I assume we'll have more failures, so starting an issue to gather all of the until we figure them out.
https://github.com/tenstorrent/tt-umd/actions/runs/14143817749/job/39629379291

Running IO for cluster 2
 Starting test_read_write_all_tensix_cores for cluster 140289201146784 thread_id 2
/__w/tt-umd/tt-umd/tests/unified/multiprocess.cpp:42: Failure
Expected equality of these values:
  vector_to_write
    Which is: { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 }
  readback_vec
    Which is: { 0, 1, 2, 3, 4, 5, 6, 7, 0, 1 }
Vector read back from core CoreCoord: (1, 1, TENSIX, PHYSICAL) does not match what was written

@pjanevskiTT if you see more failures, post to this issue, so we have enough relevant information when we get to it.

Issue #651, opened 2025-03-28
Hang on TTDevice mutex


joelsmithTT:
```
(gdb) bt
#0  __futex_abstimed_wait_common64 (private=<optimized out>, cancel=true, abstime=0x0, op=265, expected=0, futex_word=0x7496f14a0000)
    at ./nptl/futex-internal.c:57
#1  __futex_abstimed_wait_common (cancel=true, private=<optimized out>, abstime=0x0, clockid=0, expected=0, futex_word=0x7496f14a0000)
    at ./nptl/futex-internal.c:87
#2  __GI___futex_abstimed_wait_cancelable64 (futex_word=futex_word@entry=0x7496f14a0000, expected=expected@entry=0,
    clockid=clockid@entry=0, abstime=abstime@entry=0x0, private=<optimized out>) at ./nptl/futex-internal.c:139
#3  0x00007496f069cbdf in do_futex_wait (sem=sem@entry=0x7496f14a0000, abstime=0x0, clockid=0) at ./nptl/sem_waitcommon.c:111
#4  0x00007496f069cc78 in __new_sem_wait_slow64 (sem=0x7496f14a0000, abstime=0x0, clockid=0) at ./nptl/sem_waitcommon.c:183
#5  0x00007496f0f8713c in boost::interprocess::ipcdetail::posix_named_semaphore::wait() ()
   from /home/joel/git/tt-umd/cmake-build/device/libdevice.so
#6  0x00007496f0fec6d5 in tt::umd::TTDevice::write_to_device(void*, tt::umd::xy_pair, unsigned long, unsigned int) ()
   from /home/joel/git/tt-umd/cmake-build/device/libdevice.so
#7  0x0000635364c36708 in ApiTTDeviceTest_BasicTTDeviceIO_Test::TestBody() ()
#8  0x0000635364c6fc59 in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) ()
#9  0x0000635364c55878 in testing::Test::Run() ()
#10 0x0000635364c56870 in testing::TestInfo::Run() ()
#11 0x0000635364c57335 in testing::TestSuite::Run() ()
#12 0x0000635364c6761d in testing::internal::UnitTestImpl::RunAllTests() ()
[...]

Issue #72, opened 2024-09-23
Global mutex issues


broskoTT:
After talking with @tt-vjovanovic, he raised an issue with me that there are some issues with global mutexes used:
- Mutexes don't have "tt_" or anything like it in their name, making them hard to identify which process are they related to, on the system, from outside of our code
- Mutexes aren't cleaned up when some code crashes. This then leads to hanging mutexes, and until you manually fix it, UMD won't work properly on that system.

The task at hand is to refactor usage of these mutexes a bit to address these issues.
It is still not clear however what is the right path to achieve this:
- Refactor existing mutexes
- According to the new UMD design, it might make sense to tie them to some of the new classes. This could come naturally from the previous point and from the redesign itself.
- @joelsmithTT worked at one point on moving them to KMD, maybe that is the path to go

joelsmithTT:
The presence of these mutexes reflect a system design flaw.  Consider the case of the `ARC_MSG` mutexes: like the other mutexes in UMD, they are implemented with shared memory so that they work in a multiprocess context.  If two separate UMD-based applications attempt to message a device's ARC firmware simultaneously, the mutex will serialize the accesses.  (The same is true for a single UMD-based application running with multiple threads).

The design is flawed because UMD is not the only software that can interact with hardware.  Applications based on the Luwen library can message the ARC firmware without participating in UMD's mutex scheme.  The same is true of Python-based tooling used internally.

There are a variety of bad techniques to try to solve this:
* Don't run programs that try to message the ARC (or use any other similarly guarded chip resource) simultaneously.  This is the situation today.  This isn't realistic: users want to run e.g. `tt-smi` simultaneously with an ML workload.
* Everything that access the chip accesses it through UMD.  This is not realistic: tools built out of the Luwen library are not going away.
* Everything that accesses the chip agrees upon and respects a locking convention to guard hardware resources.  KMD has basic support for this, but there is no established convention.  Moreover, the KMD implementation does not handle lock contention properly: it forces userspace to poll.  A userspace based approach is possible: teach Luwen and the Python codes to touch the same shared memory that UMD uses.  This has other drawbacks, including the cleanup issue mentioned in this issue.

There is a good solution:
* Lower resource management to KMD.  KMD can enforce serialization.  This would require augmenting the KMD interface.

broskoTT:
Thanks for additional info.

Two approaches in solving this issue, which you identified:
[...]

PR #693, opened 2025-04-04
Robust locking mechanism

### LINKED ISSUE #691: UMD locks left uncleaned
So we've had this problem before. If UMD process crashes, the locks are left on the system. In the meantime, this was handled by clean_system_resources. In some of the recent refactors, most probably https://github.com/tenstorrent/tt-umd/pull/658, this path diverged, and now this happens again.

Somewhat related task and issue was that our multiprocess tests were unstable, due to us not properly cleaning lock resources, which was also introduced somewhat by the same referenced PR: https://github.com/tenstorrent/tt-umd/issues/654 https://github.com/tenstorrent/tt-umd/pull/673.

The true solution is to finally implement these mutexes properly, and I guess we can't avoid it anymore. Of course even better solution is not to have usage which mandates usage of locks, but I don't think we can avoid this short term.

broskoTT:
### Issue
Fixes #691 #654 #651, also completes #72 #523 , related to #418

### Description
We've had this problem for a while: Process crashing could leave locked mutexes on the system. The workaround was to do "rm /dev/shm/*".
After some refactoring and an attempt to consolidate all locking code to a single class, the non proper cleanup has hit us yet again. I've finally dedicated time to try to solve this properly.
Of course some of the usages of lock might disappear by the future redesign but we still might need some interprocess critical sections in UMD, and this class should be able to provide it robustly.
The solution is to use PTHREAD_MUTEX_ROBUST attribute, which allows for a mechanism to recover when a process which holds the lock crashes. This also meant that we had to manually write locking code, which made this effort  more involving.

### List of the changes
- Implemented RobustLock class. It is intended to be used as RAII.
- LockManager is now just a simple wrapper which concatenates strings.
- Removed all initialize and clear_mutex calls, as this is now not needed. Each "locking" will also initialize the mutex. This part might move to LockManager if we deem this is not performant enough.
- Removed initialize_default_chip_mutexes since its not needed anymore.
- Renamed get_mutex to acquire_lock to better communicate the API
- Removed includes and usages to boost::interprocess.
- Re-enabled multiprocess tests

### Testing
I've introduced crashes at many places in the code, and added some prints to all codepaths in RobustMutex. I was then manually running some test and observed what was happening. Verified everything worked as intended, verified no deadlocks occurred, verified mutex was successfully restored after crash. Also, I reenabled multiprocess tests which were previously consistently failing with the old locks, and verified they pass now, so it seems like the locks are indeed locking critical sections.

### API Changes
But removal of #include <boost/interprocess/sync/named_mutex.hpp> will likely make some people happy.


device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,149 @@
...
+    // This marks the mutex as robust. This will have the effect in the case of process crashing, another process
+    // waiting on the mutex will get the signal and will get the flag that the previous owner of mutex died, so it can
+    // recover the mutex state.
+    err = pthread_mutexattr_setrobust(&attr, PTHREAD_MUTEX_ROBUST);
+    if (err != 0) {
+        throw std::runtime_error("pthread_mutexattr_setrobust failed: " + std::to_string(err));
+    }
+    err = pthread_mutex_init(mutex_ptr, &attr);
+    if (err != 0) {
+        throw std::runtime_error("pthread_mutex_init failed: " + std::to_string(err));
```
pjanevskiTT, line 100: might be useful to store mutex name as well so we can now which one failed. Maybe that information can be useful sometimes for debugging if needed

device/api/umd/device/utils/robust_mutex.h
```diff
@@ -0,0 +1,54 @@
...
+    ~RobustMutex();
+
+    // Locks and unlocks the mutex.
+    void unlock_mutex();
+    void lock_mutex();
+
+private:
+    // Does everything related to initializing the mutex, even on first time creation.
+    void initialize_pthread_mutex(std::string mutex_name);
+    // Closes the mutex, doesn't remove the backing mutex file.
```
pjanevskiTT, line 26: nit: space between previous function and comment, for all functions

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,149 @@
...
+#include <errno.h>     // errno, ENOENT
+#include <fcntl.h>     // O_RDWR, O_CREATE
+#include <pthread.h>   // pthread_mutexattr_init, pthread_mutexattr_setpshared, pthread_mutex_t
+#include <sys/stat.h>  // for fstat
+#include <unistd.h>    // ftruncate, close
+
+#include <stdexcept>
+
+static constexpr int ALL_RW_PERMISSION = 0666;
+static constexpr std::string_view UMD_LOCK_PREFIX = "umdlock.";
```
pjanevskiTT, line 19: nit: add TT prefix. Not an official rule of course, maybe we can start adopting it

pjanevskiTT (approved):
amazing 🚀

device/utils/robust_mutex.cpp
broskoTT: Why? Do you want me to put it in tt::umd:: namespace instead?

pjanevskiTT: @tt-vjovanovic asked me to put it in mutex names some time ago, not sure if we should do it as I said. I am fine with keeping this as it is

device/api/umd/device/utils/lock_manager.h
```diff
@@ -30,37 +26,40 @@ enum class MutexType {
...
 class LockManager {
 public:
-    LockManager();
-    ~LockManager();
-
     // This set of functions is used to manage mutexes which are system wide and not chip specific.
     void initialize_mutex(MutexType mutex_type, const bool clear_mutex);
     void clear_mutex(MutexType mutex_type);
-    std::unique_lock<boost::interprocess::named_mutex> get_mutex(MutexType mutex_type);
+    std::unique_ptr<RAIIMutex> acquire_mutex(MutexType mutex_type);
```
joelsmithTT, line 39: Idiomatic C++ generally distinguishes between the mutex (the thing that gets locked/unlocked) and the lock (the wrapper that provides exception-safe lock/unlock semantics).  Please consider:
1. Making `RobustMutex` Lockable (essentially provide public lock() and unlock() methods - perhaps by renaming lock_mutex/unlock_mutex).
2. Replacing `RAIIMutex` with standard guards: `RAIIMutex` duplicates RAII pattern provided by `std::lock_guard` and `std::unique_lock`.  If `RobustMutex` were Lockable (per point 1), `RAIIMutex` becomes redundant and can be removed.
3. Instead of `acquire_mutex` returning a `std::unique_ptr<RAIIMutex>` (which involves a heap allocation), it could return a standard RAII lock guard _by value_.

device/api/umd/device/utils/lock_manager.h
```diff
@@ -30,37 +26,40 @@ enum class MutexType {
     CREATE_ETH_MAP,
 };
 
+// Note that the returned RAIIMutex should never outlive the LockManager which holds underlying RobustMutexes.
```
joelsmithTT, line 29: The lifetime documentation here is great.

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+using namespace tt::umd;
+
+RobustMutex::RobustMutex(std::string mutex_name) { initialize_pthread_mutex(mutex_name); }
```
tt-vjovanovic: There are three options to avoid copying memory:
1. Use `const std::string&` and do copy only when you need to store that value
2. Use `std::string` and then use `std::move` when you don't need it (in your example `initialize_pthread_mutex(std::move(mutex_name));` )
3. Use `std::string_view` and do copy only when you need to store that value...

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,159 @@
...
+using namespace tt::umd;
+
+RobustMutex::RobustMutex(std::string mutex_name) : mutex_name_(mutex_name) { initialize_pthread_mutex(mutex_name); }
+
+RobustMutex::~RobustMutex() { close_mutex(); }
+
+void RobustMutex::initialize_pthread_mutex(std::string mutex_name) {
+    bool created = open_shm_file(mutex_name);
+    // There is a slight chance of race condition here, where first process creates the file, and second process opens
+    // it. Then the second process "surpasses" the first process and tries using the mutex before it is initialized.
```
joelsmithTT, line 30: So:
* Process A creates the shared memory file
* Process B opens it, maps it, and tries to use the `pthread_mutex_t`
* Process A finishes calling `initialize_pthread_mutex` via `initialize_pthread_mutex_first_use`
This seems like a critical flaw.

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,159 @@
...
+bool RobustMutex::open_shm_file(std::string mutex_name) {
+    std::string shm_file_name = std::string(UMD_LOCK_PREFIX) + mutex_name;
+    bool created = false;
+    shm_fd = shm_open(shm_file_name.c_str(), O_RDWR, ALL_RW_PERMISSION);
+    // This indicates error opening the file.
+    // If the file doesn't exist, create it instead.
+    if (shm_fd == -1 && errno == ENOENT) {
+        // There is a slight chance of race condition here, since two processes can enter this same codepath, and then
+        // both can try creating the file.
+        shm_fd = shm_open(shm_file_name.c_str(), O_RDWR | O_CREAT, ALL_RW_PERMISSION);
```
joelsmithTT, line 63: `man 3 shm_open` says,

"O_EXCL If  O_CREAT was also specified, and a shared memory object with the given name already exists, return an error.  The check for the existence of the object, and its creation if it does not exist, are performed atomically."

Can this do the open atomically and handle EEXIST?

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,149 @@
...
+    open_pthread_mutex(shm_fd);
+    if (created) {
+        initialize_pthread_mutex_first_use();
+    }
+}
+
+bool RobustMutex::open_shm_file(std::string mutex_name) {
+    std::string shm_file_name = std::string(UMD_LOCK_PREFIX) + mutex_name;
+    bool created = false;
+    shm_fd = shm_open(shm_file_name.c_str(), O_RDWR, ALL_RW_PERMISSION);
```
tt-vjovanovic, line 57: You should probably switch to this order:
```cpp
shm_fd = shm_open(..., O_RDWR | O_CREAT | O_EXCL, ...);
if (shm_fd == -1 && errno == EEXIST) {
    shm_fd = shm_open(..., O_RDWR, ...);
}
```
To minimize race condition in file creation.

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,159 @@
...
+        throw std::runtime_error("shm_open failed for mutex " + mutex_name_ + " errno: " + std::to_string(errno));
+    }
+    return created;
+}
+
+void RobustMutex::open_pthread_mutex(int shm_fd) {
+    // Create a pthread_mutex based on the shared memory file descriptor
+    void *addr = mmap(NULL, sizeof(pthread_mutex_t), PROT_READ | PROT_WRITE, MAP_SHARED, shm_fd, 0);
+    if (addr == MAP_FAILED) {
+        throw std::runtime_error("mmap failed for mutex " + mutex_name_ + " errno: " + std::to_string(errno));
```
joelsmithTT, line 76: Let's assume `mmap` fails.  The exception propagates out of `open_pthread_mutex`, then out of `initialize_pthread_mutex` and finally out of `RobustMutex::RobustMutex`.  Because construction failed, the object is never fully-formed.  Therefore `RobustMutex::~RobustMutex` is never executed.  Since destructor is never executed, `close_mutex()` is never called.  So the process still holds the open fd referring to the shared memory object, but the int holding the fd is gone.

I think there are some other resource leaks on construction failure paths.

joelsmithTT (commented):
Resolving the identified race conditions during initialization is crucial for ensuring the mutex's reliability under concurrency.

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+RobustMutex::RobustMutex(std::string mutex_name) { initialize_pthread_mutex(mutex_name); }
+
+RobustMutex::~RobustMutex() { close_mutex(); }
+
+void RobustMutex::initialize_pthread_mutex(std::string mutex_name) {
+    bool created = open_shm_file(mutex_name);
+    // There is a slight chance of race condition here, where first process creates the file, and second process opens
+    // it. Then the second process "surpasses" the first process and tries using the mutex before it is initialized.
+    if (created) {
```
tt-vjovanovic, line 31: You can have two threads/processes, one will create file, one will open file. (even after comment I left in `open_shm_file`) You will have no guarantees who will get here first (one that opened file or one that creates it, it depends on CPU scheduler). So, what you need to do after opening is to lock the file, check if it is empty (that means it is just created) and do initialization.

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,149 @@
...
+        // File needs to be resized first time it is created.
+        if (ftruncate(shm_fd, sizeof(pthread_mutex_t)) != 0) {
```
tt-vjovanovic, line 33: Is `pthread_mutex_t` something that can be changed during OS runtime? Like you upgrade something with `sudo apt upgrade` and you get different size of it? Or even worse, you get structure of the same size, but different layout?
If that can happen, you should consider versioning mutex files. Like you have `umdlock.` as prefix, consider adding `pthread.h` "version" in the mutex name as well...

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,149 @@
...
+#include <stdexcept>
+
+static constexpr int ALL_RW_PERMISSION = 0666;
+static constexpr std::string_view UMD_LOCK_PREFIX = "umdlock.";
+
+using namespace tt::umd;
+
+RobustMutex::RobustMutex(std::string mutex_name) { initialize_pthread_mutex(mutex_name); }
+
+RobustMutex::~RobustMutex() { close_mutex(); }
```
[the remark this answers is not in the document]
broskoTT: You'd suggest warnings in that case?

device/utils/robust_mutex.cpp
broskoTT: Yes. But it seems unlikely enough for me to be fine with it for now. Note that also that happens only the first time on the system (after boot), not on each execution

device/utils/robust_mutex.cpp
broskoTT: this would never had occurred to me in three lifetimes

device/api/umd/device/utils/lock_manager.h
broskoTT: I ended up returning a unique_lock, but that one should also involve a heap allocation. lock_guard can't be used when you want to return it from a function.

device/utils/robust_mutex.cpp
broskoTT: The process closes file descriptor automatically when it crashes (or rather, the os closes it when the process crashes), right?

device/utils/robust_mutex.cpp
broskoTT: You know what, if I do what you suggested, I fear that it may happen I have two processes running on different pthread versions (one long running still running on old version). Instead, I'll guard this behavior by just reporting if the size is different, since I have to check the size each time when I open the file anyway. Let me know what you think of this

device/utils/robust_mutex.cpp
broskoTT: Changed to string_view

device/utils/robust_mutex.cpp
broskoTT: Yes, did exactly this and added comment to explain.

device/utils/robust_mutex.cpp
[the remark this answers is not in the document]
broskoTT: Changed to warnings and marked noexcept

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,159 @@
...
+        }
+    } else if (lock_res != 0) {
+        throw std::runtime_error(
+            "pthread_mutex_lock failed for mutex " + mutex_name_ + " errno: " + std::to_string(lock_res));
+    }
+}
+
+RAIIMutex::RAIIMutex(RobustMutex *mutex) : mutex_ptr(mutex) { mutex_ptr->lock_mutex(); }
+
+RAIIMutex::~RAIIMutex() { mutex_ptr->unlock_mutex(); }
```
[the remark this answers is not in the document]
broskoTT: Removed RAIIMutex and used unique_lock instead

device/utils/robust_mutex.cpp
broskoTT: Hopefully did this right.

broskoTT:
@joelsmithTT @tt-vjovanovic thanks for the thorough review!
I've addressed the comments, and the logic has changed now a bit, please take another look now, should be better.

Here's the comment on the change:
I've removed the race condition on file creation by inverting the order of trying to create a file and trying to open a file.
I've created a critical section for pthread initialisation, and with that removed the race condition where first process creates the file, but then another surpasses it and tries using pthread before it is initialized. The critical section was done using flock and static pthread_mutex_t
Changed throws into log_fatal
Changed throws in destructor into warnings, marked it noexcept to be explicit we don't want to throw on destruction path.
Renamed functions in RobustMutex so it implements BasicLockable requirement
Remove RAIIMutex, in favour of unique_lock which is in standard library
Map in LockManager now holds references to objects, rather than unique_ptrs. I had to implement move constructor and assignment operator in RobustMutex to allow that.
Changed arg mutex_name to string_view

device/utils/robust_mutex.cpp
joelsmithTT: That's right, the kernel will clean up the resources no matter what.  The resource leak concern is applicable if the process continues running.  For example, it is conceivable that the higher-level application could catch the exception and try again.

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,205 @@
...
+        if (ftruncate(shm_fd, sizeof(pthread_mutex_t)) != 0) {
+            log_fatal("ftruncate failed for mutex {} errno: {}", mutex_name_, std::to_string(errno));
+        }
+        // Get file size again, so that the check on the next line doesn't fail.
+        file_size = get_file_size(shm_fd);
+    }
+
+    // Verify file size.
+    if (file_size != sizeof(pthread_mutex_t)) {
+        log_fatal(
```
joelsmithTT, line 87: `log_fatal` is broken.  It is not fatal, it just throws a `std::runtime_error`.  It's conceivable that an application could catch the exception and continue running.  This is not a problem with your code here, just pointing it out for awareness.  Something that is named `log_fatal` should actually be fatal.

I think it is appropriate to treat errors in this class as fatal (i.e. the program simply terminates).

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,205 @@
...
+    // If we don't create a critical section for this, then two processes could race, one to initialize the mutex and
+    // the other one to use it before it is initialized. flock ensures only multiprocess locking, but does not guarantee
+    // multithread locking. Due to that, we need to use multithread_mutex_ which guarantees multithread locking but not
+    // multiprocess locking. Note that flock is released automatically on process crash, and static multithread_mutex_
+    // is not persistent, so we're safe even if the process crashes in the critical section. We use only a single
+    // multithread_mutex_ for all different RobustMutex instances which can affect perf of these operations, but that is
+    // fine since this is executed rarely, only on initialization and only once after booting the system. Regarding
+    // flock perf, this happens only when initializing the mutex, so it is not a big deal.
+    flock(shm_fd, LOCK_EX);
+    pthread_mutex_lock(&multithread_mutex_);
```
joelsmithTT, line 70: Should the return values of these function calls be checked?

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+    // We need a critical section here in which we test if the mutex has been initialized, and if not initialize it.
+    // If we don't create a critical section for this, then two processes could race, one to initialize the mutex and
+    // the other one to use it before it is initialized. flock ensures only multiprocess locking, but does not guarantee
+    // multithread locking. Due to that, we need to use multithread_mutex_ which guarantees multithread locking but not
+    // multiprocess locking. Note that flock is released automatically on process crash, and static multithread_mutex_
+    // is not persistent, so we're safe even if the process crashes in the critical section. We use only a single
+    // multithread_mutex_ for all different RobustMutex instances which can affect perf of these operations, but that is
+    // fine since this is executed rarely, only on initialization and only once after booting the system. Regarding
+    // flock perf, this happens only when initializing the mutex, so it is not a big deal.
```
joelsmithTT, line 68: 👍

joelsmithTT (commented):
I appreciate the attention to correctness around initialization, this looks like an effective technique.

A few housekeeping notes:

* There's a comment in `LocalChip::initialize_default_chip_mutexes` that won't be accurate after this change goes in
* The `clean_system_resources` flag (also spelled `clear_mutex`) shows up in multiple files.  With the improved robustness in mutex handling, are these flags still needed?
* Similarly, regarding the methods that delete a mutex (i.e. `LockManager::clear_mutex`, `LockManager::clear_mutex_internal`).  Is this operation still needed, after UMD does not have to worry about tidying up stale state in `/dev/shm/`?

device/utils/robust_mutex.cpp
broskoTT: right, forgot about it

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,205 @@
...
+    flock(shm_fd, LOCK_EX);
+    pthread_mutex_lock(&multithread_mutex_);
+
+    size_t file_size = get_file_size(shm_fd);
+    // This means the file was just created but nor file nor mutex were initialized.
+    bool should_initialize = (file_size == 0);
```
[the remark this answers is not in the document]
broskoTT: Sure that sounds like a breaking scenario. Any ideas how to fix it? I'd say this is unlikely to happen, so we can maybe ignore it. It is also possible that somebody manually manipulates the file in shm, which can also lead to a failure...

device/utils/robust_mutex.cpp
broskoTT: But the user of the lib might want to actually wrap everything in a try catch block, and do something else if UMD initialization fails. That should be why the exception mechanism was added, to add more choice on how to handle failures, compared to just having exit(1) or something like that.

device/utils/robust_mutex.cpp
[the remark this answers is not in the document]
joelsmithTT: How about:

* In the shared memory segment, store a `struct { pthread_mutex_t mutex; uint64_t magic; }` instead of a `pthread_mutex_t`
* Write a hard-coded constant random number to the magic field _after_ `pthread_mutex_init()` succeeds
* should_initialize logic checks for the presence of the magic number instead of the segment size

You are correct about external manipulation of the data.  That seems like a fragility inherent to using shared memory.  But that fragility is something we have today already, with the boost named_mutex scheme.  Your robust mutex represents a significant improvement.

device/utils/robust_mutex.cpp
joelsmithTT: Fatal means causing death.  `log_fatal` should be calling `std::terminate()`, but it just throws.  The program can escape death as long as there is a try block that can catch a `std::runtime_error`.  `log_fatal` is a lie.

It's fine if RobustMutex throws exceptions and leaves the choice up to the caller.  In that case:
* `log_fatal` is a confusing way to convey this behavior
* The code would need to be made exception safe -- for example, if `mmap` fails, currently the `multithread_mutex_` doesn't get unlocked and the `shm_fd` is leaked

broskoTT: Sure, log_fatal doesn't have a proper name https://github.com/tenstorrent/tt-umd/issues/696
Ok I'll try to unlock and cleanup before exception where needed...

device/utils/robust_mutex.cpp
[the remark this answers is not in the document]
broskoTT: I did exactly this. I tested with some example code failing midway in initialization, both with throwing and pure exit, everything works as expected.
User gets a warning in case of corrupted underlying file. After manually deleting it, everything starts working again (it gets reinitialized)

broskoTT:
Changes after second iteration of addressing comments
- Added check for flock and pthread_mutex_lock during initialization.
- Also wrapped this in a small RAII class, which will guarantee cleanup of these resources even if initialization code throws.
- Moved initialize() out of constructor. This allows writing proper cleanup code in the destructor, and than knowing it will be called if initialize() fails (which is not true during constructor's execution).
- Changed simple pthread_mutex to a wrapper which also holds a magic number. The shared memory file is now mapped into this struct instead of just pthread_mutex. The struct has a bool in addition, which designates whether the pthread_mutex was properly initialized, since pthread_mutex doesn't offer any other mechanism to verify itself if it is initialized.
- Add mutex_name_ to move constuructors, this was a bug, but mutex_name was used instead of mutex_name_ so it was working, now it doesn't anymore once I separated initialize and construction.

Retested again with some cases, and running multiprocess tests.

Thanks @joelsmithTT and @tt-vjovanovic for continued support on this!

broskoTT:
Time taken for 10000000 mutex initializations: 40752193209 nanoseconds, so single one is around 40752
Time taken for 10000000 lock/unlock: 654239360 nanoseconds, so single one is around 654
Time taken for 10000000 unique_lock/unlock: 680775582 nanoseconds, so single one is around 680

@tt-vjovanovic

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,263 @@
+/*
```
tt-vjovanovic, line 1: Small scenario to verify:
1. Start application that uses UMD and opens this mutex.
2. Lock this mutex.
3. Start another application that uses this mutex - this should hang since it should wait for mutex.
4. Delete file in shm directory
5. Start third application that uses this mutex - this should pass since it will create new file.

If this test behaves as explained, then this mechanism might not be the best one for our use case...
On the other hand it might be good enough solution (if we say user should not do that :D)

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,263 @@
...
+// flock ensures only multiprocess locking, but does not guarantee
+// multithread locking. Due to that, we need to use multithread_mutex_ which guarantees multithread locking but not
+// multiprocess locking. Note that flock is released automatically on process crash, and static multithread_mutex_
+// is not persistent, so we're safe even if the process crashes in the critical section.
+class CriticalSectionScopeGuard {
+public:
+    CriticalSectionScopeGuard(int fd, pthread_mutex_t* pthread_mutex, std::string_view mutex_name) :
+        fd_(fd), pthread_mutex_(pthread_mutex), mutex_name_(mutex_name) {
+        if (flock(fd_, LOCK_EX) != 0) {
+            log_fatal("flock failed for mutex {} errno: {}", mutex_name_, std::to_string(errno));
```
tt-vjovanovic, line 38: This doesn't throw any more, right?

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,263 @@
...
+        if (err != 0) {
+            // Try to unlock the flock without handling further exceptions.
+            flock(fd_, LOCK_UN);
+            log_fatal("pthread_mutex_lock failed for mutex {} errno: {}", mutex_name_, std::to_string(err));
+        }
+    }
+
+    ~CriticalSectionScopeGuard() noexcept {
+        // Use best effort to unlock the mutex and the flock and report warnings if something fails.
+        int err = pthread_mutex_unlock(pthread_mutex_);
```
tt-vjovanovic, line 50: What will happen on this line if flock fails in constructor? Can this unlock mutex for other thread?

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,263 @@
...
+        other.shm_fd_ = -1;
+        other.mutex_wrapper_ptr_ = nullptr;
+        other.mutex_name_ = "";
+    }
+    return *this;
+}
+
+void RobustMutex::initialize() {
+    int err;
+    bool created = open_shm_file(mutex_name_);
```
tt-vjovanovic, line 107: `created` variable doesn't seem to be used

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+void RobustMutex::unlock() {
+    int err = pthread_mutex_unlock(&(mutex_wrapper_ptr_->mutex));
+    if (err != 0) {
+        log_fatal("pthread_mutex_unlock failed for mutex {} errno: {}", mutex_name_, std::to_string(err));
+    }
+}
+
+void RobustMutex::lock() {
+    int lock_res = pthread_mutex_lock(&(mutex_wrapper_ptr_->mutex));
```
tt-vjovanovic, line 251: Can we verify simple scenario:
Start two applications. Open same mutex of this type. Ask for lock.
Process A will get the lock, while process B will wait for process A to unlock it.
Intentionally crash process A (with terminate or unhandled exception).
Verify that process B entered `lock_res == EOWNERDEAD` and it got the lock.

device/utils/robust_mutex.cpp
tt-vjovanovic: If `log_fatal` throws exception, then I understand how it will work :)

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,263 @@
...
+        shm_fd_ = shm_open(shm_file_name.c_str(), O_RDWR, ALL_RW_PERMISSION);
+        created = false;
+    }
+    if (shm_fd_ == -1) {
+        log_fatal("shm_open failed for mutex {} errno: {}", mutex_name_, std::to_string(errno));
+    }
+    return created;
+}
+
+void RobustMutex::open_pthread_mutex(int shm_fd_) {
```
joelsmithTT, line 176: Is the `int shm_fd_` parameter needed?

device/utils/robust_mutex.cpp
broskoTT: Since that scenario is far from somebody doing it unintentionally, I don't want to handle it. There are for sure many other ways to break our code (in whichever layer) intentionally.
The implementation relies on everyone being good, actually using the locks and not corrupt underlying files, including deleting them

device/utils/robust_mutex.cpp
broskoTT: yes, it throws :) And the answer is that the destructor won't be called

device/utils/robust_mutex.cpp
broskoTT: yeah left it from the first version, will remove

device/utils/robust_mutex.cpp
broskoTT: I did similar scenario manually, I commented on the way I did manual testing in my very long PR description. I ran one process and crashed it while holding the lock. Then ran it again and observed some added prints which showed this code path being executed.
However I didn't test if the other is waiting on a lock at the moment when the first one crashes....

broskoTT: Yeah just did this test as well

```
RobustMutex mutex("test_mutex");
mutex.initialize();
mutex.lock();
std::cout << "locked mutex" << std::endl;
while(true);
```

Also added prints to this codepath.
Ran it in two windows, then when I CTRL+C in the first, the second gets the lock.

device/api/umd/device/utils/robust_mutex.h
```diff
@@ -0,0 +1,70 @@
...
+class RobustMutex {
+public:
+    RobustMutex(std::string_view mutex_name);
+    ~RobustMutex() noexcept;
+
+    // Does everything related to initializing the mutex, even on first time creation.
+    // The initialization can fail and throw an exception. In case of failure we still want to clean up the resources.
+    // For easier handling of such case, the destructor cleans up the resources if they were taken. Having this code out
+    // of constructor will guarantee that the destructor is called in case of exception.
+    void initialize();
```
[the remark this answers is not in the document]
broskoTT: This was @tt-vjovanovic 's suggestion. Doing it separately allows me to always clean up in the destructor if the initialization fails, instead of trying to mimick some other way to implement try-catch-finally block in c++, or try to release resources on each path when possible.
You can actually see that in the constructor of CriticalSectionScopeGuard, but there is only one place where it can fail which requires cleanup, so it's not that bad. When I tried to do the same in the initialize function it got messy

device/utils/robust_mutex.cpp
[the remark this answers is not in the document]
broskoTT: But this is just the naming issue right? We can change log_fatal to log_exception or something in another PR, which would make it clearer

device/utils/robust_mutex.cpp
broskoTT: I started with another kind of implementation, so this is leftover. Will change

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,263 @@
...
+    // multithread_mutex_ for all different RobustMutex instances which can affect perf of these operations, but that is
+    // fine since this is executed rarely, only on initialization and only once after booting the system. Regarding
+    // flock perf, this happens only when initializing the mutex, so it is not a big deal.
+    // The critical_section object will get destroyed at the end of this function or when an exception is thrown, so the
+    // critical section will be released automatically.
+    CriticalSectionScopeGuard critical_section(shm_fd_, &multithread_mutex_, mutex_name_);
+
+    size_t file_size = get_file_size(shm_fd_);
+    // This means the file was just created but nor file nor mutex were initialized.
+    bool should_initialize = (file_size == 0);
```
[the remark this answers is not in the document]
joelsmithTT: The problem to solve is the case where `ftruncate()` has sized the file appropriately but the process dies before `initialize_pthread_mutex_first_use()` succeeds.  You do not need to worry about multiple simultaneous processes because the flock technique serializes access here.

As written, the following code path can get skipped even if there was never a successful call to `initialize_pthread_mutex_first_use()`:

```
    if (should_initialize) {
        // We need to initialize the mutex here, since it is the first time it is being used.
        initialize_pthread_mutex_first_use();
    }
```
because `should_initialize` is based only on the size of the segment, not on whether the mutex in the segment was actually initialized.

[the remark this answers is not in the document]
joelsmithTT: Yes, I believe automatically recovering from any partial-initialization crash scenario is preferable to requiring manual intervention and aligns better with the 'Robust' goal.

The key change is in how the decision to call `initialize_pthread_mutex_first_use()` is made inside the `CriticalSectionScopeGuard` block.  Currently, that decision hinges on whether the file size is zero.

Instead, after mapping the memory, the logic can check `if (mutex_wrapper_ptr_->initialized != INITIALIZED_FLAG)` and only call `initialize_pthread_mutex_first_use()` if that condition is true.  I think this will work even if a previous process was killed while inside of the `pthread_mutex_init()`.

Do you see any holes in this logic?

device/utils/robust_mutex.cpp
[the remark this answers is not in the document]
joelsmithTT: yes / ok

device/api/umd/device/utils/robust_mutex.h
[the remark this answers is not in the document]
joelsmithTT: I can see how it gets messy quickly.

device/utils/robust_mutex.cpp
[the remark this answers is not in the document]
broskoTT: Yes, you have to zero out the mutex memory before you try to initialize it again, but that's also doable.
Why do I fail then in case of file size being different than the expected? in that case I should also then try to resize it to proper size?

device/utils/robust_mutex.cpp
broskoTT: I'll add comment to the code

device/utils/robust_mutex.cpp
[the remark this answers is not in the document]
broskoTT: I changed the code, so it will try to initialize now in case of wrong file size, or wrong INITIALIZED_FLAG, and will report warnings if something wasn't as expected.

device/utils/robust_mutex.cpp
broskoTT: Added comment

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,295 @@
...
+#include <unistd.h>    // ftruncate, close
+
+#include <stdexcept>
+
+#include "logger.hpp"
+
+static constexpr int ALL_RW_PERMISSION = 0666;
+static constexpr std::string_view UMD_LOCK_PREFIX = "TT_UMD_LOCK.";
+// Any value which is unlikely to be found at random in the memory.
+static constexpr uint64_t INITIALIZED_FLAG = 0x5454554d444d5458;  // TTUMDMTX
```
tt-vjovanovic, line 24: Stupid question, but easy to try :)
There was character literal, might be Microsoft compiler extension, but you were able to write something like `'TEST'`, which will generate 32bit integer with character values in reverse order...

tt-vjovanovic: ChatGPT says it is Microsoft extension :)

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,295 @@
...
+        // Invalidate the other object, so the destructor doesn't try to close the same resources.
+        other.shm_fd_ = -1;
+        other.mutex_wrapper_ptr_ = nullptr;
+        other.mutex_name_ = "";
+    }
+    return *this;
+}
+
+void RobustMutex::initialize() {
+    int err;
```
tt-vjovanovic, line 111: `err` doesn't seem to be used

device/utils/robust_mutex.cpp
broskoTT: I added a comment to the class docs that this might happen, and we don't handle it well.

device/utils/robust_mutex.cpp
```diff
@@ -0,0 +1,295 @@
...
+    // flock perf, this happens only when initializing the mutex, so it is not a big deal.
+    // The critical_section object will get destroyed at the end of this function or when an exception is thrown, so the
+    // critical section will be released automatically.
+    CriticalSectionScopeGuard critical_section(shm_fd_, &multithread_mutex_, mutex_name_);
+
+    // Resize file if needed.
+    bool file_was_resized = resize_shm_file();
+
+    // We now open the mutex in the shared memory file.
+    open_pthread_mutex();
```
tt-vjovanovic, line 128: Should this be done after file initialization (call to `initialize_pthread_mutex_first_use`)?

tt-vjovanovic (approved):

joelsmithTT (approved):

.github/workflows/run-tests.yml
```diff
@@ -78,9 +78,6 @@ jobs:
         run: |
           ${{ env.TEST_OUTPUT_DIR }}/umd/misc/umd_misc_tests
 
-      # TODO: Re-enable unified tests in CI once we solve the bugs
```
broskoTT, line 81: Verify tt-metal tests pass
All runs on brosko/test_umdlocks :
- [ ] All post-commit tests : https://github.com/tenstorrent/tt-metal/actions/runs/14335808671
- [ ] Blackhole post-commit tests : https://github.com/tenstorrent/tt-metal/actions/runs/14335810810
- [ ] (Single-card) Model perf tests : https://github.com/tenstorrent/tt-metal/actions/runs/14335813192
- [ ] (Single-card) Device perf regressions : https://github.com/tenstorrent/tt-metal/actions/runs/14335816204
- [ ] (T3K) T3000 unit tests : https://github.com/tenstorrent/tt-metal/actions/runs/14335819889
- [ ] (T3K) T3000 demo tests : https://github.com/tenstorrent/tt-metal/actions/runs/14335822736
- [ ] (TG) TG unit tests : https://github.com/tenstorrent/tt-metal/actions/runs/14335825045
- [ ] (TG) TG demo tests : https://github.com/tenstorrent/tt-metal/actions/runs/14335826785

Issue #550, opened 2025-02-27
Add timeout to Blackhole ARC message queue


pjanevskiTT:
Implement timeout for sending ARC messages on Blackhole

joelsmithTT:
Added in https://github.com/tenstorrent/tt-umd/pull/698

Issue #695, opened 2025-04-04
ARC heartbeat check seems to be timing out on Metal


tt-rkim:
@asrinivasanTT noticed an error like

https://github.com/tenstorrent/tt-metal/actions/runs/14262197831/job/39976649437

where ARC doesn't seem to respond in time.

```
Timed out after waiting 1 seconds for ARC to respond
```

It's odd that ARC hangs would happen like this as AI SW like Metal shouldn't be doing anything to disturb ARC, really. We're wondering if there was a recent change to the ARC heartbeat check logic or timeout or something like that in UMD, as we didn't really see this before on the cloud VMs or in the Metal CI.

It's possible that because of the device layouts on Cloud and how large these machines are, it's possible we're just going over some threshold (1 second?) for waiting for a response. Note that these large T7000 servers which house the CI VMs have 8 cards, so who knows which card each individual VM is on.

It's non-deterministic from the current info we know.

joelsmithTT:
https://github.com/tenstorrent/tt-umd/blob/b2330028b0254e5ee2fedccc7f56efe4732b77f4/device/wormhole/wormhole_arc_messenger.cpp#L57 should say milliseconds
https://github.com/tenstorrent/tt-umd/blob/b2330028b0254e5ee2fedccc7f56efe4732b77f4/device/api/umd/device/cluster.h#L398 should say `timeout_ms = 1000`

joelsmithTT:
Fixed in https://github.com/tenstorrent/tt-umd/pull/698

pjanevskiTT:
closed by #698

PR #698, opened 2025-04-07
Fix ARC timeout message bug

joelsmithTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/695

### Description
Fixes an incongruity between Cluster and ArcMessenger with respect to timeout units.  Adds timeout functionality to BlackholeArcMessageQueue. 

### List of the changes
1. Cluster: change arc_msg timeout from seconds to ms to match lower-level implementation
2. ArcMessenger: clarify arc_messenger timeout documentation
3. Correct an error message and support indefinite wait in WormholeArcMessenger
4. Add timeout support to BlackholeArcMessageQueue

### Testing
Manual

### API Changes
This PR has API changes, but should not break anything.

### DIFF
--- a/device/blackhole/blackhole_arc_message_queue.cpp
+++ b/device/blackhole/blackhole_arc_message_queue.cpp
@@ -31,13 +31,21 @@ void BlackholeArcMessageQueue::trigger_fw_int() {
     tt_device->write_to_device((void*)&ARC_FW_INT_VAL, arc_core, ARC_FW_INT_ADDR, sizeof(uint32_t));
 }
 
-void BlackholeArcMessageQueue::push_request(std::array<uint32_t, BlackholeArcMessageQueue::entry_len>& request) {
+void BlackholeArcMessageQueue::push_request(
+    std::array<uint32_t, BlackholeArcMessageQueue::entry_len>& request, uint32_t timeout_ms) {
     uint32_t request_queue_wptr = read_word(request_wptr_offset);
 
+    auto start = std::chrono::steady_clock::now();
     while (true) {
         uint32_t request_queue_rptr = read_word(request_rptr_offset);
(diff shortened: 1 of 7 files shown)


pjanevskiTT:
this PR fixes #550 as well

device/blackhole/blackhole_arc_message_queue.cpp
```diff
@@ -31,14 +32,20 @@ void BlackholeArcMessageQueue::trigger_fw_int() {
     tt_device->write_to_device((void*)&ARC_FW_INT_VAL, arc_core, ARC_FW_INT_ADDR, sizeof(uint32_t));
 }
 
-void BlackholeArcMessageQueue::push_request(std::array<uint32_t, BlackholeArcMessageQueue::entry_len>& request) {
+void BlackholeArcMessageQueue::push_request(
+    std::array<uint32_t, BlackholeArcMessageQueue::entry_len>& request, uint32_t timeout_ms) {
     uint32_t request_queue_wptr = read_word(request_wptr_offset);
 
+    util::Timestamp timer;
```
pjanevskiTT, line 39: can this be done using std::chrono just to be in sync with other timeouts implemented

pjanevskiTT (approved):
looks good, thanks for the changes

device/blackhole/blackhole_arc_message_queue.cpp
joelsmithTT: I am making this change since you asked, although glancing over the codebase I think `util::Timestamp` is correct and lots of other places are not.  I created a separate issue to keep this PR small.

https://github.com/tenstorrent/tt-umd/issues/703

pjanevskiTT: thanks, will fix timestamping

PR #684, opened 2025-04-03
Add a test for reading NOC1 noc node id register

pjanevskiTT:
### Issue

Read NOC_NODE_ID register using NOC1 and for NOC1 

### Description

Add a test that reads noc node id register for noc1 coordinates and using noc1. 

### List of the changes

- Add a test for reading noc1 noc node id reg
- Fix translation of coordinates, there was a potential bug when using NOC1

### Testing

CI

### API Changes
/

Issue #630, opened 2025-03-24
Figure out reading NOC node id from specific cores


pjanevskiTT:
Figure out how to read NOC node id from specific set of cores. #617 is introducing reading this information for one groups of cores, it should be supported for all the core groups. Groups that are missing/have some bug in reading this information

- Blackhole ARC
- Blackhole PCIE
- Wormhole DRAM
- Wormhole PCIE
- Wormhole ARC

PR #699, opened 2025-04-07
Add NOC node id check for more cores

pjanevskiTT:
### Issue

big part of #630 

### Description

Add functions to get base address of more core types. Update the NOC0 test. NOC1 test in #684 should be updated accordingly.

### List of the changes

- Add function to get base NOC register addr for all core types
- Add function to get noc node id offset per arch
- Update the NOC0 test
- Enable readouts from more core types

### Testing

CI

### API Changes
/

### DIFF
--- a/tests/api/test_cluster.cpp
+++ b/tests/api/test_cluster.cpp
@@ -342,28 +342,11 @@ TEST(TestCluster, TestClusterLogicalETHChannelsConnectivity) {
 TEST(TestCluster, TestClusterNocId) {
     std::unique_ptr<Cluster> cluster = std::make_unique<Cluster>();
 
-    // Setup memory barrier addresses.
-    // Some default values are set during construction of UMD, but you can override them.
-    cluster->set_barrier_address_params({L1_BARRIER_BASE, ETH_BARRIER_BASE, DRAM_BARRIER_BASE});
-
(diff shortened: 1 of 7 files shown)


tests/api/test_cluster.cpp
```diff
@@ -376,10 +359,8 @@ TEST(TestCluster, TestClusterNocId) {
         const std::vector<CoreCoord>& cores = cluster->get_soc_descriptor(chip).get_cores(core_type);
         for (const CoreCoord& core : cores) {
             const auto [x, y] = read_noc_id_reg(cluster, chip, core);
-            CoreCoord translated_coord =
-                cluster->get_soc_descriptor(chip).translate_coord_to(core, CoordSystem::TRANSLATED);
-            EXPECT_EQ(translated_coord.x, x);
-            EXPECT_EQ(translated_coord.y, y);
+            EXPECT_EQ(core.x, x);
```
broskoTT, line 362: So you're now using pure physical coordinates?

broskoTT (approved):

tests/api/test_cluster.cpp
pjanevskiTT: yes we have registers for both physical and translated coordinates, I'll move the tests to physical and then write additional for translated space

Issue #697, opened 2025-04-07
Certain unit tests can hang the NOC


joelsmithTT:
ApiTLBManager.ManualTLBConfiguration is wrong because the tt_SocDescriptor is constructed without knowledge of the hardware.  Therefore the 0th Tensix core's coordinates are not guaranteed to be valid (i.e. the coordinates may correspond to a harvested core)
```
    tt_SocDescriptor soc_desc = tt_SocDescriptor(
        tt_SocDescriptor::get_soc_descriptor_path(tt_device->get_arch()), tt_device->get_arch() != tt::ARCH::GRAYSKULL);

// <snip>

    for (CoreCoord core : soc_desc.get_cores(CoreType::TENSIX)) {
        tlb_manager->configure_tlb(core, core, get_static_tlb_index(core), c_zero_address, tlb_data::Relaxed);
    }

    // So now that we have configured TLBs we can use it to interface with the TTDevice.
    auto any_worker_core = soc_desc.get_cores(CoreType::TENSIX)[0];
    tlb_configuration tlb_description = tlb_manager->get_tlb_configuration(any_worker_core);

    // TODO: Maybe accept tlb_index only?
    uint64_t address_l1_to_write = 0;
    std::vector<uint8_t> buffer_to_write = {0x01, 0x02, 0x03, 0x04};
    tt_device->write_block(
        tlb_description.tlb_offset + address_l1_to_write, buffer_to_write.size(), buffer_to_write.data());
```

ApiTTDeviceTest.BasicTTDeviceIO is wrong because for multiple PCIe devices it will use the 0th Tensix core coordinates for the 0th chip.  The coordinates of this Tensix may not be valid for a subsequent chip (i.e. that Tensix may be harvested).
```
tt_xy_pair get_any_tensix_core(tt::ARCH arch) {
    return std::make_unique<Cluster>()->get_soc_descriptor(0).get_cores(CoreType::TENSIX)[0];
}

// elsewhere...
    for (int pci_device_id : pci_device_ids) {
        std::unique_ptr<TTDevice> tt_device = TTDevice::create(pci_device_id);

        tt_xy_pair tensix_core = get_any_tensix_core(tt_device->get_arch());

        tt_device->write_to_device(data_write.data(), tensix_core, address, data_write.size() * sizeof(uint32_t));
```

pjanevskiTT:
Just take the proper non-harvested core for each chip. The fix is easy, I'm on it

[...]

PR #704, opened 2025-04-07
Fix tests to use non-harvested cores

pjanevskiTT:
### Issue

#697 

### Description

Fix TLBManager and TTDevice test to always use non-harvested tensix cores.

### List of the changes

- Fix tests to use non-harvested cores

### Testing

CI

### API Changes
/

### DIFF
--- a/tests/api/test_tlb_manager.cpp
+++ b/tests/api/test_tlb_manager.cpp
@@ -64,12 +64,15 @@ TEST(ApiTLBManager, ManualTLBConfiguration) {
     std::int32_t c_zero_address = 0;
 
     for (CoreCoord core : soc_desc.get_cores(CoreType::TENSIX)) {
-        tlb_manager->configure_tlb(core, core, get_static_tlb_index(core), c_zero_address, tlb_data::Relaxed);
+        auto virtual_core = soc_desc.translate_coord_to(core, CoordSystem::VIRTUAL);
+        auto translated_core = soc_desc.translate_coord_to(core, CoordSystem::TRANSLATED);
+        tlb_manager->configure_tlb(
+            virtual_core, translated_core, get_static_tlb_index(core), c_zero_address, tlb_data::Relaxed);
     }
 
     // So now that we have configured TLBs we can use it to interface with the TTDevice.
-    auto any_worker_core = soc_desc.get_cores(CoreType::TENSIX)[0];
-    tlb_configuration tlb_description = tlb_manager->get_tlb_configuration(any_worker_core);
+    auto any_worker_virtual_core = soc_desc.get_cores(CoreType::TENSIX, CoordSystem::VIRTUAL)[0];
+    tlb_configuration tlb_description = tlb_manager->get_tlb_configuration(any_worker_virtual_core);
 
     // TODO: Maybe accept tlb_index only?
     uint64_t address_l1_to_write = 0;
@@ -79,6 +82,6 @@ TEST(ApiTLBManager, ManualTLBConfiguration) {
 
     // Another way to write to the TLB.
     // TODO: This should be converted to AbstractIO writer.
-    tt::Writer writer = tlb_manager->get_static_tlb_writer(any_worker_core);
+    tt::Writer writer = tlb_manager->get_static_tlb_writer(any_worker_virtual_core);
     writer.write(address_l1_to_write, buffer_to_write[0]);
 }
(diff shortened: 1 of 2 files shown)


tests/api/test_tt_device.cpp
```diff
@@ -26,7 +22,12 @@ TEST(ApiTTDeviceTest, BasicTTDeviceIO) {
     for (int pci_device_id : pci_device_ids) {
         std::unique_ptr<TTDevice> tt_device = TTDevice::create(pci_device_id);
 
-        tt_xy_pair tensix_core = get_any_tensix_core(tt_device->get_arch());
+        ChipInfo chip_info = tt_device->get_chip_info();
+
+        tt_SocDescriptor soc_desc(
+            tt_device->get_arch(), chip_info.noc_translation_enabled, chip_info.harvesting_masks, chip_info.board_type);
```
tt-rkim, line 28: Interesting, if you send in `chip_info.harvesting_masks`  `soc_desc.get_cores(CoreType::TENSIX)` will return only non-harvested cores?

tt-rkim (approved):

asrinivasanTT:
Tried running this on the chips with harvest_mask 3, and tests don't seem to hang card/cause reboot.

Will test on another node too.

joelsmithTT (approved):

asrinivasanTT:
I'm noticing a reboot on n150s with harvest mask 2 (2nd row harvested). It seemed to run fine on n300s with harvest mask 3 (2nd row harvested)

tests/api/test_tt_device.cpp
broskoTT: Yes. We have separate .get_cores and .get_harvested_cores APIs, which return cores in any coord system.

broskoTT (approved):

broskoTT:
@joelsmithTT thanks again for finding these issues!
@pjanevskiTT thanks for fixing this quickly. You can't request @tt-rkim since he doesn't have those permissions on our repo, that's also why his approval is gray :)

@asrinivasanTT thanks for testing this change!
Is the reboot consistent, does it happen every time on this card? Is the last printout always the same? This info can help us pinpoint the issue easier, or we can just take the card and test ourselves

asrinivasanTT:
We have one node that has n150s marked out. Running this on both those cards is causing a reboot. The last printout is the same as what was happening previously (Same thing I got this time and  I posted above).

I will test everything again once more to be double sure.

tests/api/test_tlb_manager.cpp
```diff
@@ -29,8 +29,10 @@ TEST(ApiTLBManager, ManualTLBConfiguration) {
     }
 
     std::unique_ptr<TLBManager> tlb_manager = std::make_unique<TLBManager>(tt_device.get());
```
broskoTT, line 31: So the n300 has failed two times on this PR, right after this test. And I see it passing on other jobs, so it not a problem with card definitely. We need to debug that before this goes in

asrinivasanTT:
> I will test everything again once more to be double sure.

causing reboots only on n150s with harvest mask 2 as I reported above. Let us know if you want access to the card.

I'm not seeing any n300 failures like the CI run.

tests/api/test_tlb_manager.cpp
broskoTT: Here's a run with TensixSoftwareHarvestingAllChips
https://github.com/tenstorrent/tt-umd/actions/runs/14326907520
So the tests definitely break after ApiTLBManager.ManualTLBConfiguration

broskoTT: Yeah some other test now fails, we fixed this one

broskoTT: Here's a run with log_info in setdynamictlb
https://github.com/tenstorrent/tt-umd/actions/runs/14330344305

And another one with potentially fixed test:
https://github.com/tenstorrent/tt-umd/actions/runs/14330375797

broskoTT: Seems like TLB configuration with physical coordinates (mind that these are indeed unharvested) doesn't work for some reason,  I've switched both test to use translated coordinates and it now passes

asrinivasanTT:
I'm no longer seeing hangs on n150 either. Will add them back to the pool. Thanks for fixing this.

Issue #705, opened 2025-04-08
Add support for Security cores


pjanevskiTT:
Add support for security cores. Support should be added in

- wormhole_implementation
- blackhole_implementation
- CoordinateManager
- SocDescriptor
- soc descriptor yamls

Additionally, to confirm everything, it would be nice to add

- CoordinateManager translation test
- Soc descriptor reading of cores
- NOC node id readouts

pjanevskiTT:
closed by #710

PR #710, opened 2025-04-09
Support security cores

pjanevskiTT:
### Issue

#705 

### Description

Add support for security cores. Support for the cores is added in all places described in the issue.

### List of the changes

- Add l2cpu core type
- Add l2cpu cores to CoordinateManager
- Add l2cpu cores to soc descriptor
- Add tests for l2cpu cores

### Testing
CI

### API Changes
/


device/api/umd/device/blackhole_implementation.h
```diff
@@ -132,6 +132,9 @@ static const std::vector<tt_xy_pair> ETH_CORES_NOC0 = {
      {10, 1}}};
 
 static const std::vector<tt_xy_pair> ETH_LOCATIONS = ETH_CORES_NOC0;
+
+static const std::vector<tt_xy_pair> SECURITY_CORES_NOC0 = {{{8, 2}}};
```
broskoTT, line 136: nit, reduce to two curly braces?

device/api/umd/device/blackhole_implementation.h
```diff
@@ -227,18 +230,20 @@ static constexpr uint32_t TENSIX_L1_SIZE = 1499136;
...
-constexpr std::array<std::pair<CoreType, uint64_t>, 5> NOC0_CONTROL_REG_ADDR_BASE_MAP = {
+constexpr std::array<std::pair<CoreType, uint64_t>, 6> NOC0_CONTROL_REG_ADDR_BASE_MAP = {
     {{CoreType::TENSIX, 0xFFB20000},
      {CoreType::ETH, 0xFFB20000},
      {CoreType::DRAM, 0xFFB20000},
      {CoreType::PCIE, 0xFFFFFFFFFF000000ULL},
-     {CoreType::ARC, 0xFFFFFFFFFF000000ULL}}};
-constexpr std::array<std::pair<CoreType, uint64_t>, 5> NOC1_CONTROL_REG_ADDR_BASE_MAP = {
+     {CoreType::ARC, 0xFFFFFFFFFF000000ULL},
+     {CoreType::SECURITY, 0xFFFFFFFFFF000000ULL}}};
```
broskoTT, line 239: nit: Also why do we have 3 curly braces here?

tests/api/test_soc_descriptor.cpp
```diff
@@ -561,3 +561,37 @@ TEST(SocDescriptor, AllSocDescriptors) {
...
+    tt_SocDescriptor soc_desc_yaml(
+        test_utils::GetAbsPath("tests/soc_descs/wormhole_b0_8x10.yaml"), true, harvesting_masks);
+
+    std::vector<CoreCoord> security_cores = soc_desc_yaml.get_cores(CoreType::SECURITY);
+    for (size_t index = 0; index < security_cores.size(); index++) {
+        EXPECT_EQ(security_cores[index].x, tt::umd::blackhole::SECURITY_CORES_NOC0[index].x);
+        EXPECT_EQ(security_cores[index].y, tt::umd::blackhole::SECURITY_CORES_NOC0[index].y);
+    }
+
+    tt_SocDescriptor soc_desc_arch(tt::ARCH::WORMHOLE_B0, true, harvesting_masks);
```
broskoTT, line 590: both soc_desc_yaml and soc_desc_arch are referencing wormhole.
But besides that, I don't think this test should exist.
In the first loop you want to match what is in our standard yaml with a constant in another file. If we want to do that we should have a separate test making sure all constants match in yaml and in header file.
Second loop just repeats how this was programmed, you assign those cores using those same constants, I don't see a point in comparing them to the same constant again.

I'd just maybe test if by default blackhole reports 1 security core and thats it.

broskoTT (approved):

device/api/umd/device/blackhole_implementation.h
pjanevskiTT: constexpr array needs that kind of initialization

broskoTT:
This change broke tt_metal build, but this simple fix on tt_metal fixes that https://github.com/tenstorrent/tt-metal/pull/20590

PR #717, opened 2025-04-09
Fix: enforce alignment for register access

joelsmithTT:
Remove the complex and bugged tt_4_byte_aligned_buffer class.  This fixes a bug where uninitialized padding bytes could be written to registers if the original size was not a multiple of 4.

Add strict checks to ensure that register access is 4-byte aligned and in a multiple of 4 bytes.

### DIFF
--- a/device/chip/local_chip.cpp
+++ b/device/chip/local_chip.cpp
@@ -14,32 +14,6 @@
 
 extern bool umd_use_noc1;
 
-struct tt_4_byte_aligned_buffer {
-    // Stores a 4 byte aligned buffer
-    // If the input buffer is already 4 byte aligned, this is a nop
-    std::uint32_t* local_storage = nullptr;
-    std::uint32_t input_size = 0;
-    std::uint32_t block_size = 0;
-
-    tt_4_byte_aligned_buffer(const void* mem_ptr, uint32_t size_in_bytes) {
-        input_size = size_in_bytes;
-        local_storage = (uint32_t*)mem_ptr;
-        uint32_t alignment_mask = sizeof(uint32_t) - 1;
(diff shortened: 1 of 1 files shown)


joelsmithTT:
Some Metal CI runs

https://github.com/tenstorrent/tt-metal/actions/runs/14365369534/job/40277705311
https://github.com/tenstorrent/tt-metal/actions/runs/14365367601
https://github.com/tenstorrent/tt-metal/actions/runs/14365364868
https://github.com/tenstorrent/tt-metal/actions/runs/14365362803/job/40277709646

device/chip/local_chip.cpp
```diff
@@ -14,32 +14,6 @@
 
 extern bool umd_use_noc1;
 
-struct tt_4_byte_aligned_buffer {
```
broskoTT, line 17: ah, so if there was a register at [0-3], and I wanted to write only 1 byte to location [2], then I'd actually write zeros to 0,1,3 ? Rather than reading reg values at those addresses, and making sure that's unchanged? Yeah, this seems like it could lead to non obvious behavior...

broskoTT (approved):
Thanks for fixing and reducing code complexity!

device/chip/local_chip.cpp
joelsmithTT: Misaligned writes to a hardware register should be avoided; if the goal is to update the byte at `0x00ff0000` then a read (of the whole 32 register), modify, write (full 32 bits) is appropriate.

This code is confusing because it doesn't deal with alignment at all: it deals with padding.  So in your example, the behavior would end up being: pad the desired 1 byte with 3 bytes of uninitialized memory, and perform a misaligned write that crosses a register boundary.

Issue #522, opened 2025-02-07
get rid of / make optional: flatbuffers


alex-s168:
Flatbuffers is a heavy dependency and takes up a major amount of build time.
This is not important, but it would be more convenient if flatbuffers was optional or removed completely
It seems like it is only used in [tt_simulation_device.fbs](https://github.com/tenstorrent/tt-umd/blob/3e6e65629ad41d8c38ec20138964bd2a5b399187/device/simulation/tt_simulation_device.fbs)

pjanevskiTT:
@alex-s168 thanks for opening this issue

I also think this just makes the header file with same structs as in flatbuffer, looks like it can be moved to .h file.

Have you measured the exact time of the build time difference when we use .fbs and when this is moved to the .h file? Anyway, It would be nice to remove it no matter how big the build time difference is. 

If you want to work on it, you can assign it yourself or I can assign it to you.

alex-s168:
@pjanevskiTT 
I'll start working on this now.
Can you assign me please

broskoTT:
Fixed by #728

PR #718, opened 2025-04-09
remove flatbuffers dep

alex-s168:
### Issue
#522 

### Description
depending on flatbuffers increases compile time by a LOT

### List of the changes
custom implementation of subset of flatbuffers

### Testing
I verified that the custom serializer / deserializer produces the same bytes as the previous flatbuffers code.


afuller-TT:
TT-Metalium also uses Flatbuffers.  So removing it from UMD won't help the compile time for anyone one layer up.
That said, I'm not a fan of rolling our own when there's a good F/OSS lib out there.
Granted compile times are a major concern.

I would first want to investigate (in order of preference):
1. Can we use the system-provided flatbuffers?  We've dropped Ubuntu 20.04 now.  22.04 seems to ship `1.12.1~git20200711.33e2d80`.  Is that version good enough?  Or are we relying on some newer features?
2. If 1.12 isn't good enough, then why is our build of flatbuffers so much slower than a roll-your-own solution?  Are we building unncessary stuff?  Do they provide some flags for us to turn off unnecessary bits?

blozano-tt:
Brainstorming:

- We could bake flatbuffers into our Docker image, so its prebuilt
- We could provide a script to go download and build non system provided dependencies
- We could ingest said script in our Dockerfile

All these ideas are built upon the idea that we can bypass CPM download and build with:

https://github.com/tenstorrent/tt-metal/blob/0898308d763dad2d4b17a3da9bc6e07e89de83c0/build_metal.sh#L42

alex-s168:
Sure, but not everyone will use metalium.

I think it's okay to have a custom implementation of flatbuffers if our flatbuffer file never really changes, and we are sure that it works under all circumstances.

For metalium, we should definitely use the official flatbuffers lib

blozano-tt:
I stubmled across this today in the profiler report of tt-metalium's compilation:

```
**** Functions that took longest to compile:
  1266 ms: flatbuffers::csharp::CSharpGenerator::GenStruct(flatbuffers::StructD... (/work/.cpmcache/flatbuffers/2c4062bffa52fa4157b1b4deeae73395df475fda/src/idl_gen_csharp.cpp)
  1065 ms: flatbuffers::java::JavaGenerator::GenStruct(flatbuffers::StructDef&,... (/work/.cpmcache/flatbuffers/2c4062bffa52fa4157b1b4deeae73395df475fda/src/idl_gen_java.cpp)
   957 ms: flatbuffers::ts::TsGenerator::GenStruct(flatbuffers::Parser const&, ... (/work/.cpmcache/flatbuffers/2c4062bffa52fa4157b1b4deeae73395df475fda/src/idl_gen_ts.cpp)
   926 ms: flatbuffers::java::JavaGenerator::GenPackUnPack_ObjectAPI(flatbuffer... (/work/.cpmcache/flatbuffers/2c4062bffa52fa4157b1b4deeae73395df475fda/src/idl_gen_java.cpp)
   838 ms: flatbuffers::ts::TsGenerator::GenObjApi(flatbuffers::Parser const&, ... (/work/.cpmcache/flatbuffers/2c4062bffa52fa4157b1b4deeae73395df475fda/src/idl_gen_ts.cpp)
   806 ms: main (/work/tests/tt_metal/tt_metal/perf_microbenchmark/routing/test_vc_bi_tunnel_2ep.cpp)
   796 ms: main (/work/tests/tt_metal/tt_metal/perf_microbenchmark/routing/test_vc_bi_tunnel_4ep.cpp)
   691 ms: main (/work/tests/tt_metal/tt_metal/perf_microbenchmark/routing/test_vc_mux_demux.cpp)
   641 ms: flatbuffers::csharp::CSharpGenerator::GenPackUnPack_ObjectAPI(flatbu... 
```

blozano-tt:
Another option for UMD that we discussed internally, was to not pull in flatbuffers at all by default, because it is only needed for simulation. 

There could be an option ...

UMD_SIMULATION_MODE = OFF by default.

blozano-tt:
I propose this (Most humans using UMD will never need flatbuffers):

https://github.com/tenstorrent/tt-umd/pull/728

broskoTT:
I also don't like the idea of our own implementation for flatbuffers, seems like additional complexity in our own repo without many benefits.
Especially since https://github.com/tenstorrent/tt-umd/pull/728 is going to go in pretty soon, I don't see any other benefits to this PR change

alex-s168:
> @alex-s168 can we close this PR?

Yes

Issue #692, opened 2025-04-04
Remote communication on Wormhole not working when using NOC1


pjanevskiTT:
Remote communication is not working on wormhole when using noc1. Fix is identified and fixed on branch https://github.com/tenstorrent/tt-umd/tree/pjanevski/noc1_noc_id_fix

There should be a more general fix available (encoding noc1 to sys addr for remote transactions), gotta investigate how to do that

pjanevskiTT:
closed by #726

PR #726, opened 2025-04-10
Fix remote NOC1 communication

pjanevskiTT:
### Issue

#692 

### Description

Set flag to use NOC1 when using remote communication and umd_use_noc1 is set.

### List of the changes

- Set flag for remote NOC1 communication
- Fix the test that showed this bug
- Remove the comment in test that showed this bug

### Testing
CI. Test that showed this was reading NOC1 node id on remote chip, that works now.

### API Changes
/


device/cluster.cpp
```diff
@@ -66,6 +66,8 @@ static const uint32_t MSG_ERROR_REPLY = 0xFFFFFFFF;
 // Remove 256MB from full 1GB for channel 3 (iATU limitation)
 static constexpr uint32_t HUGEPAGE_CHANNEL_3_SIZE_LIMIT = 805306368;
 
+static constexpr uint32_t REMOTE_CMD_NOC_BIT = 9;
```
broskoTT, line 69: Is this something to be someplace else? Do we have other place with remote communication related constants? or maybe wh constants?

broskoTT (approved):

device/cluster.cpp
pjanevskiTT: we don't. I would keep making that common place out of scope of this PR. We already have double structures in remote communication and cluster. When you finish non-mmio moving to Chip, I think we can then make common place for these remote things

PR #732, opened 2025-04-11
Add tt-exalens build

broskoTT:
### Issue
No issue

### Description
exalens is now public so we can start building it to track early if we break build.

### List of the changes
- Add tt-exalens build similar to tt-metal one.
- regular gh machine should be enough, exalens builds fast

### Testing
CI Testing

### API Changes


pjanevskiTT:
the PR is fixing the issue #732

pjanevskiTT (approved):

Issue #646, opened 2025-03-27
TLBAllocator


broskoTT:
As depicted by https://docs.google.com/drawings/d/1-m1azdsBqMA0A6ATYRMfkhyeuOJuGCEI62N5a96LXj0/edit

There should be a TLBAllocator class, which has the same interface as defined by #463 
This one lives inside TTDevice.

As a helper class, there should be a StaticTLBManager, which is there not to bog down the Chip class. This one TTDevice doesn't know about.
The current TLBManager is a mix of two, and should be split.

This tasks contributes to overall completion of #248

broskoTT:
TLBManager was moved to chip.
But I won't work currently on TLBAllocator, that should be part of TTDevice and another task

pjanevskiTT:
for now PCIDevice allocates TLB handles, we are probably not going to implement an allocator, maybe rename the TLBManager at some point

Issue #463, opened 2025-01-20
TLB management in KMD


pjanevskiTT:
TLB management is going to be moved to https://github.com/tenstorrent/tt-kmd. API for TLB management is going to be exposed through few ioctl calls.

```
1. Alloc window: Size => index
2. Mmap window: index, cache(WC/UC) => ptr to the TLB window
3. Program window: TLBAttributes, index => (TLB window is programed, no return value)
4. Free window: index => (TLB window is released)
5. Write: ptr => (KMD writes 64 bytes to device)
6. Read: ptr => (KMD reads 64 bytes from device)
```

Read and write calls are going to be slow so it should be used just in case we cannot get TLB from the KMD. UMD should probably never go through this path

Issue with more details on the design -> TODO: add KMD issue

syseng design https://tenstorrent.atlassian.net/wiki/spaces/syseng/pages/537100642/PCIe+TLBs

broskoTT:
![Image](https://github.com/user-attachments/assets/12fb745e-1c2e-44a8-91f6-e0731e86b871)

pjanevskiTT:
closed by #733

PR #733, opened 2025-04-11
Structures for TLB KMD management

pjanevskiTT:
### Issue

Working towards #646 #463 

### Description

Add structures required to use KMD API for TLB management.

### List of the changes

- Add TlbHandle
- Add TlbWindow
- Add TlbAllocator
- Add tests

### Testing

CI. Added test_tlb for testing TLB KMD calls

### API Changes
/


device/pcie/pci_device.cpp
```diff
@@ -440,3 +440,34 @@ semver_t PCIDevice::read_kmd_version() {
 
     return semver_t(version_str);
 }
+
+tenstorrent_allocate_tlb_out PCIDevice::allocate_tlb(tenstorrent_allocate_tlb_in in) {
+    tenstorrent_allocate_tlb allocate_tlb{};
+    allocate_tlb.in = in;
+    if (ioctl(get_fd(), TENSTORRENT_IOCTL_ALLOCATE_TLB, &allocate_tlb) < 0) {
+        std::cout << "error" << std::endl;
+        return {};
```
joelsmithTT, line 449: This will return a zero-initialized `tenstorrent_allocate_tlb_out` which will look valid to the caller (`tenstorrent_allocate_tlb_out::id == 0` is legal).  The mmap offsets will also be zero.  Should this throw an exception?

device/tt_device/tlb_window.cpp
```diff
@@ -0,0 +1,67 @@
...
+    }
+}
+
+TlbHandle& TlbWindow::handle_ref() { return *handle; }
+
+size_t TlbWindow::get_size() const { return handle->get_size(); }
+
+// For simplicity and correctness, only allow 32-bit aligned accesses.
+// There exist platform and device specific considerations for unaligned
+// accesses which are not addressed here.
```
joelsmithTT, line 56: Wormhole has some read/modify/write logic to handle unaligned and 1/2 byte access cases.

See https://github.com/tenstorrent/tt-umd/blob/0f5841bd1eeac69f73b4c62dc4765013aa192ef3/device/api/umd/device/tt_device/tt_device.h#L172
and https://yyz-gitlab.local.tenstorrent.com/tenstorrent/syseng/-/issues/3487

device/pcie/pci_device.cpp
```diff
@@ -440,3 +440,34 @@ semver_t PCIDevice::read_kmd_version() {
...
+    }
+
+    return allocate_tlb.out;
+}
+
+tenstorrent_free_tlb_out PCIDevice::free_tlb(tenstorrent_free_tlb_in in) {
+    tenstorrent_free_tlb free_tlb{};
+    free_tlb.in = in;
+    if (ioctl(get_fd(), TENSTORRENT_IOCTL_FREE_TLB, &free_tlb) < 0) {
+        return {};
```
joelsmithTT, line 459: Consider throwing an exception here -- although if you do, you'll want to ensure that such an exception cannot escape from `~TlbHandle()`.  If you don't throw here, maybe an error should be logged.  If UMD fails to free a TLB, it is probably UMD programmer error indicative of a bug.

device/tt_device/tlb_window.cpp
```diff
@@ -0,0 +1,67 @@
...
+}
+
+uint32_t TlbWindow::read32(uint64_t offset) {
+    validate(offset, sizeof(uint32_t));
+    return *reinterpret_cast<volatile uint32_t*>(handle->get_base() + offset);
+}
+
+void TlbWindow::write_register(uint64_t offset, uint32_t value) { write32(offset, value); }
+
+uint32_t TlbWindow::read_register(uint64_t offset) { return read32(offset); }
```
joelsmithTT, line 24: Register access should use an uncached mapping for ordering and immediate visibility.  A more arcane concern is that WC mappings may be speculatively read by the CPU.  I don't know of any registers in our system that have side-effects when read.  But UC avoids this.

I don't have any immediate suggestion for how to deal with the UC/WC thing.  A few ideas:
* always map them both (pros: flexible, cons: easier to misuse; I'm not 100% convinced this is reliable)
* choose mapping type when the window is constructed
* separate abstractions (e.g. TlbWindowUC, TlbWindowWC -- I am not sure I like this, but it's definitely explicit about intent...)
* ??

tests/unified/test_tlb.cpp
```diff
@@ -0,0 +1,187 @@
...
+TEST(TestTlb, TestTlbWindowReadRegister) {
+    const uint64_t tensix_addr = 0;
+    const chip_id_t chip = 0;
+    const uint64_t two_mb_size = 1 << 21;
+
+    // Point of the test is to read NOC0 node id register.
+    // TLB needs to be aligned to 2MB so these base and offset values are
+    // how TLB should be programmed in order to get to addr 0xFFB2002C.
+    const uint64_t tlb_base = 0xFFA00000;
+    const uint64_t noc_node_id_tlb_offset = 0x12002C;
```
joelsmithTT, line 114: Longer term (not necessary for this introductory PR) I think it's worth figuring out how the TlbWindow abstraction or the logic that creates the windows and hands them out can handle this for the caller.

device/pcie/pci_device.cpp
```diff
@@ -440,3 +440,34 @@ semver_t PCIDevice::read_kmd_version() {
...
+    }
+
+    return free_tlb.out;
+}
+
+tenstorrent_configure_tlb_out PCIDevice::configure_tlb(tenstorrent_configure_tlb_in in) {
+    tenstorrent_configure_tlb configure_tlb{};
+    configure_tlb.in = in;
+    if (ioctl(get_fd(), TENSTORRENT_IOCTL_CONFIGURE_TLB, &configure_tlb) < 0) {
+        return {};
```
joelsmithTT, line 469: Similar comment to the one for `PCIDevice::allocate_tlb`

device/pcie/pci_device.cpp
```diff
@@ -440,3 +440,34 @@ semver_t PCIDevice::read_kmd_version() {
 
     return semver_t(version_str);
 }
+
+tenstorrent_allocate_tlb_out PCIDevice::allocate_tlb(tenstorrent_allocate_tlb_in in) {
+    tenstorrent_allocate_tlb allocate_tlb{};
```
joelsmithTT, line 445: Something to consider is that old tt-kmd releases don't know about the TLB API.  How should this situation be handled?  Some kind of version check seems necessary, but we don't want to obstruct anything from running against old tt-kmd releases until this integration is more mature.

joelsmithTT (approved):
Looks reasonable to me.  Since this looks like scaffolding for future integration work I am fine with it getting merged as long as it doesn't disrupt CI.  If you prefer to iterate a bit more before merging, I am happy to re-review.

device/api/umd/device/pci_device.hpp
```diff
@@ -13,6 +13,7 @@
 #include <vector>
 
 #include "fmt/format.h"
+#include "ioctl.h"
```
broskoTT, line 16: As discussed offline, try to work towards removing this include from the header file

device/api/umd/device/tt_device/tlb_handle.h
```diff
@@ -0,0 +1,35 @@
...
+ */
+#pragma once
+
+#include <sys/mman.h>
+
+#include "umd/device/pci_device.hpp"
+
+namespace tt::umd {
+
+class TlbHandle {
```
broskoTT, line 14: As discussed offline, let's try to put this class on the same level as PCIDevice. It would be a sort of a RAII handle to TLBs, where it guarantees allocate/free/configure are called in the right order.
I'm fine with leaving any amount of flexibility in the interface, such as get_base(), get_size(). 

As discussed, think about uc/wc as well, maybe we need a separate call to switch between uc and wc base? It's also fine if it is left like this for this PR, defaulting to UC, but bear in mind that we'd need to change that then before we start using it. I think some of our communication relies on WC (like to DRAM), so we'd regress perf significantly if we switched everything to UC. Maybe you could pass a flag to the constructor, and for its lifetime it would only give you UC/WC, I think that aligns well with our current usage? We never reconfigure same TLB to change is UC/WC-ness?

device/api/umd/device/pci_device.hpp
broskoTT: One potential problem I see though is with the tenstorrent_configure_tlb_in structure, since it has a lot of stuff.
You'd probably want to keep all the arguments from it when passing config, I'm fine with that. I'm also fine if you leave some of those out.

tests/unified/test_tlb.cpp
```diff
@@ -0,0 +1,187 @@
...
+    uint32_t val = 0;
+    std::vector<CoreCoord> tensix_cores =
+        cluster->get_soc_descriptor(chip).get_cores(CoreType::TENSIX, CoordSystem::TRANSLATED);
+    for (CoreCoord core : tensix_cores) {
+        cluster->write_to_device(&val, sizeof(uint32_t), chip, core, tensix_addr, "LARGE_WRITE_TLB");
+        val++;
+    }
+
+    PCIDevice* pci_device = cluster->get_tt_device(chip)->get_pci_device();
+    std::unique_ptr<TlbAllocator> tlb_allocator = std::make_unique<TlbAllocator>(pci_device);
```
broskoTT, line 30: Now after implementing other changes, I see that TLBHandle will be the element of API of PCIDevice. You'll wrap it inside TLBWindow probably, which implements read/write primitives.
But at that point I suspect that tlb_allocator would lose its point, so consider removing this class altogether.

broskoTT (commented):
Thanks for the discussion!

device/tt_device/tlb_handle.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+TlbHandle::TlbHandle(PCIDevice* pci_device, size_t size, const tenstorrent_noc_tlb_config& config) :
+    tlb_size(size), pci_device(pci_device) {
+    tenstorrent_allocate_tlb_in allocate_tlb_in;
+    allocate_tlb_in.size = size;
+    tenstorrent_allocate_tlb_out allocate_tlb_out = pci_device->allocate_tlb(allocate_tlb_in);
+    tlb_id = allocate_tlb_out.id;
+
+    try {
+        configure(config);
```
broskoTT, line 18: Consider removing the configure part from the constructor

device/tt_device/tlb_handle.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+        // mmap only UC offset for now.
+        // TODO: add choice whether to map UC or WC mapping.
+        void* uc = mmap(
+            nullptr, size, PROT_READ | PROT_WRITE, MAP_SHARED, pci_device->get_fd(), allocate_tlb_out.mmap_offset_uc);
+        if (uc == MAP_FAILED) {
+            munmap(uc, size);
+            throw std::runtime_error("Failed to map TLB UC base  using mmap call.");
```
broskoTT, line 26: Since you don't have a complex case of several places a code can throw and trying to handle all of them gracefully through a catch block, but you only have this line throwing: 
Consider eliminating try/catch, and writing here free_tlb(); throw;

tests/unified/test_tlb.cpp
```diff
@@ -0,0 +1,187 @@
...
+ */
+#include <gtest/gtest.h>
+
+#include "umd/device/cluster.h"
+#include "umd/device/tt_device/tlb_allocator.h"
+#include "umd/device/types/tlb.h"
+
+using namespace tt::umd;
+
+TEST(TestTlb, TestTlbWindowAllocateNew) {
```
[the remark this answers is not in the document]
joelsmithTT: pci_device.cpp knows how to do that: `semver_t PCIDevice::read_kmd_version()`

device/tt_device/tlb_handle.cpp
pjanevskiTT: will keep it for now...I like being able to immediately have the struct ready for read/writes

tests/unified/test_tlb.cpp
pjanevskiTT: I think it's nice to have TlbAllocator for now to have a safe point of entry. For example, TlbHandle is now directly making ioctl calls, it needs file descriptor. TlbAllocator makes you not think about passing correct fd into TlbHandle constructor

device/api/umd/device/tt_device/tlb_handle.h
pjanevskiTT: I made TlbHandle not call into PCIDevice, rather make ioctl calls directly. Other questions you asked are not part of the PR. I will make an issue to think about each one and handle it separately

device/pcie/pci_device.cpp
pjanevskiTT: guarded the tests from old kmd version. I'll implement guards for the structures during integration, not sure how that will look

tests/unified/test_tlb.cpp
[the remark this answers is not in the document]
pjanevskiTT: guarded the tests from old KMD versions

tests/unified/test_tlb.cpp
```diff
@@ -11,7 +11,18 @@
 
 using namespace tt::umd;
 
+void guard_test_kmd_version() {
+    semver_t kmd_ver = PCIDevice::read_kmd_version();
+
+    if (kmd_ver.major < 1 || (kmd_ver.major == 1 && kmd_ver.minor < 32)) {
```
joelsmithTT, line 17: I want to fix some ABI mistakes I made in the driver side: https://github.com/tenstorrent/tt-kmd/pull/55

Unfortunately this change is not backwards compatible, nor is it released yet.  Up to you how you want to handle this, either:
* hold off on merging for now; update ioctl.h when it's tt-kmd change is ready; require 1.34 as minimum version
* merge as-is (in this case I will update the ioctl.h & version check in tt-umd after tt-kmd is released with the fix)

The situation I want to avoid is a mismatch between versions of ioctl.h between tt-umd and tt-kmd.

pjanevskiTT: i am fine with holding off, not in a hurry to check this in. Thanks for pointing this out, let me know when your fix on kmd side is released

device/tt_device/tlb_window.cpp
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#include "umd/device/tt_device/tlb_window.h"
+
+#include <stdexcept>
+
+namespace tt::umd {
+
+TlbWindow::TlbWindow(std::unique_ptr<TlbHandle> handle) : handle(std::move(handle)) {}
```
broskoTT, line 12: So the tlb window becomes the owner of the tlb handle? There is no way to destroy a single tlbwindow and get back the tlbhandle to give it out to another tlbWindow?

I think TlbWindow should now own its TlbHandle. How would this work in context of dynamic tlbs that we currently have? We'd ideally get a single tlbHandle and then each time get a new TlbWindow which we can use and throw away afterwards.

I'd make this a shared_ptr, that would allow it to be used in both ways. It won't have a clear owner and it will be automatically returned to kmd when no references exist to it anymore

device/tt_device/tlb_handle.cpp
```diff
@@ -0,0 +1,76 @@
...
+TlbHandle::TlbHandle(uint32_t fd, size_t size, const tenstorrent_noc_tlb_config& config) : tlb_size(size), fd(fd) {
+    tenstorrent_allocate_tlb allocate_tlb{};
+    allocate_tlb.in.size = size;
+    if (ioctl(fd, TENSTORRENT_IOCTL_ALLOCATE_TLB, &allocate_tlb) < 0) {
+        throw std::runtime_error(fmt::format("Failed to allocate the TLB with size {}", size));
+    }
+
+    tlb_id = allocate_tlb.out.id;
+
+    configure(config);
```
broskoTT, line 27: Yeah  I still think there should be a way to get a TLBHandle without configuring it.
I'd remove the whole configure part from the constructor, and put it in a separate call

device/tt_device/tlb_handle.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+    // mmap only UC offset for now.
+    // TODO: add choice whether to map UC or WC mapping.
+    void* uc = mmap(nullptr, size, PROT_READ | PROT_WRITE, MAP_SHARED, fd, allocate_tlb.out.mmap_offset_uc);
+    if (uc == MAP_FAILED) {
+        munmap(uc, size);
+        free_tlb();
+        throw;
```
broskoTT, line 35: throw something?

device/api/umd/device/tt_device/tlb_handle.h
```diff
@@ -0,0 +1,36 @@
...
+ * SPDX-FileCopyrightText: (c) 2025 Tenstorrent Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+
+#include "ioctl.h"
```
broskoTT, line 11: I'd prefer if we don't include this in header files. This would require a separate NOC config structure. At this moment this is just a preference so I don't have a strong argument for addressing it, feel free to ignore.

tests/unified/test_tlb.cpp
```diff
@@ -0,0 +1,201 @@
...
+        config.y_start = 0;
+        config.noc = 0;
+        config.mcast = 0;
+        config.ordering = tlb_data::Strict;
+        config.linked = 0;
+        config.static_vc = 1;
+
+        std::unique_ptr<TlbWindow> tlb_window = tlb_allocator->get_tlb(two_mb_size, config);
+
+        tlb_window->handle_ref().configure(config);
```
broskoTT, line 150: you pass config in both ocnstructor and then call it here explicitly.

tests/unified/test_tlb.cpp
```diff
@@ -0,0 +1,201 @@
...
+        config.y_end = core.y;
+        config.x_start = 0;
+        config.y_start = 0;
+        config.noc = 0;
+        config.mcast = 0;
+        config.ordering = tlb_data::Relaxed;
+        config.linked = 0;
+        config.static_vc = 1;
+
+        tlb_window->handle_ref().configure(config);
```
broskoTT, line 107: Related to the other comment where I complained about the ownership... Okay, I didn't envision that you can actually reconfigure through the held ref. I had it differently in mind, and now I'm not sure if it is similar to this or if this design does have a problem. The gist of it is that I'd not allow tlb_handle to be accessed like this (i can argue that somebody can call delete on it, but that's not that probably).

Consider this solution, you should never access tlb_handle if you use it in "higher level" code like this. The code here would then look like this:
`
shared_ptr<TlbHandle> tlb_handle = TLBHandle(size);
for (...) {
  unique_ptr<TlbWindow> tlb_window = tlb_handle->configure(config);
}
`

Or if you want you can keep TlbAllocator class, then the tlb_allocator->get_tlb() would construct the tlb_handle and give you the window. In that case though, we'd have to have two different codepaths for two different scenarios:
- get a handle, keep it longer, and construct windows over it (or single window which can be reconfigured)
- get a handle and window, which should be released when window is released

But I'd remove the TlbAllocator class altogether, since it doesn't force you to explicitly think about the lifetime of tlbhandle. And it barely hides any details, it only has a couple of lines of code, to me it looks more like unnecessary complexity

broskoTT (commented):
Thanks looks much better now! But had to leave a couple of more comments

tests/unified/test_tlb.cpp
joelsmithTT: tt-kmd 1.34 is released.  You will need new ioctl.h for these TLB APIs (although you won't have to change any code).

pjanevskiTT:
After offline discussion with @broskoTT, made some changes in order to check this in as baseline and do further work from there

- Be able to create TlbHandle without configuring it
- Not putting Handle and Window in the same class because then we could create a handle which is not configured and have read/write available which could cause problems
- TlbWindow becomes the owner of TlbHandle, creating TlbWindow requires NOC config in order to configure the handle
- We need to figure out the way to keep the handle when Window is destroyed

device/tt_device/tlb_window.cpp
pjanevskiTT: gonna cover it properly with https://github.com/tenstorrent/tt-umd/issues/894

tests/unified/test_tlb.cpp
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/895

tests/unified/test_tlb.cpp
```diff
@@ -0,0 +1,206 @@
...
+        config.y_end = core.y;
+        config.x_start = 0;
+        config.y_start = 0;
+        config.noc_sel = 0;
+        config.mcast = 0;
+        config.ordering = tlb_data::Relaxed;
+        config.linked = 0;
+        config.static_vc = 1;
+
+        std::unique_ptr<TlbWindow> tlb_window =
```
broskoTT, line 55: We were talking about the case where you take one handle and own it and reconfigure it for your usage, and how that would be suppored.
Can you show that in the tests? E.g. you can but the TLBWindow out of the loop and then call ->configure inside the loop without ever releasing the tlb until you're done?

broskoTT (approved):
Thanks for the changes!

tests/unified/test_tlb.cpp
pjanevskiTT: already there, `TestTlb.TestTlbWindowReuse`

Issue #85, opened 2024-09-27
Investigate use of PCIe DMA engine in WH and beyond


joelsmithTT:
Older versions of UMD had support for programming the PCIe block's DMA controller via ARC FW.  This programming technique is unsafe if code running on the device is issuing NOC transactions to PCIe with the intent of accessing the host system bus.  Hence, the DMA controller is unused except in certain legacy code paths.

WH may offer a different technique to program the controller (registers are exposed in BAR2).
BH has different/improved addressing within the PCIe block.

Using the DMA controller for large transfers may offer a performance improvement over MMIO writes.

broskoTT:
Joel and Pavle further worked on supporting this. This is now supported in UMD: https://github.com/tenstorrent/tt-umd/pull/735

PR #735, opened 2025-04-13
Introduce a PCIe DMA API; add a basic WH implementation

joelsmithTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/85

### Description
Introduces APIs and an initial implementation for PCIe DMA transfers targeting Wormhole devices.  This aims to provide applications with a high-throughput mechanism for reading chip memory, addressing the performance limitations of MMIO reads (observed in the 3-4 MiB/s range).

The implementation uses polling via a completion flag in host memory, as driver-level interrupt support is not yet available.  It relies on an intermediate DMA buffer allocated by the driver.

### List of the changes
* Add PCIe DMA methods to `Cluster`, `Chip`, and `TTDevice` interfaces.
* Implement low-level DMA register programming and polling logic in `WormholeTTDevice`.
* Implement DMA transfer coordination (TLB management, chunking) in `LocalChip`.
* Add DMA buffer allocation and management to `PCIDevice`.
* Add simulation and mockup stubs for DMA methods.
* Add placeholder implementations for Grayskull and Blackhole `TTDevice` classes.
* Add unit tests for Wormhole DMA functionality.

### Testing
* Added unit tests covering fixed-size and randomized DMA transfers.
* Verified functionality manually on my WH dev systems.

### API Changes
Adds new public API methods for DMA transfers (`dma_write_to_device`, `dma_read_from_device` in the `Cluster` interface).  This is a non-breaking change; existing APIs are unaffected.

### DIFF
--- a/device/api/umd/device/chip/local_chip.h
+++ b/device/api/umd/device/chip/local_chip.h
@@ -46,6 +46,9 @@ public:
     void read_from_device(
         tt_xy_pair core, void* dest, uint64_t l1_src, uint32_t size, const std::string& fallback_tlb) override;
 
+    void dma_write_to_device(const void* src, size_t size, tt_xy_pair core, uint64_t addr) override;
+    void dma_read_from_device(void* dst, size_t size, tt_xy_pair core, uint64_t addr) override;
+
     void write_to_device_reg(
         tt_xy_pair core, const void* src, uint64_t reg_dest, uint32_t size, const std::string& fallback_tlb) override;
     void read_from_device_reg(
--- a/device/api/umd/device/cluster.h
+++ b/device/api/umd/device/cluster.h
@@ -331,6 +331,30 @@ public:
     }
 
     /**
+     * Use PCIe DMA to write device memory (L1 or DRAM).
+     *
+     * @param src Source data address.
+     * @param size Size in bytes.
+     * @param chip Chip to target; must be local, i.e. attached via PCIe.
+     * @param core Core to target.
+     * @param addr Address to write to.
+     */
+    virtual void dma_write_to_device(
+        const void* src, size_t size, chip_id_t chip, tt::umd::CoreCoord core, uint64_t addr) = 0;
+
+    /**
+     * Use PCIe DMA to read device memory (L1 or DRAM).
+     *
+     * @param src Source data address.
+     * @param size Size in bytes.
+     * @param chip Chip to target; must be local, i.e. attached via PCIe.
+     * @param core Core to target.
+     * @param addr Address to read from.
+     */
+    virtual void dma_read_from_device(
+        void* dst, size_t size, chip_id_t chip, tt::umd::CoreCoord core, uint64_t addr) = 0;
+
+    /**
      * Write data to specified address and channel on host (defined for Silicon).
      * This API is used to write to the host memory location that is made available to the device through
      * initialization. During the initialization the user should be able to specify how many "channels" are available to
@@ -856,6 +880,9 @@ public:
         const std::string& fallback_tlb = "LARGE_READ_TLB");
     virtual void read_from_device_reg(
         void* mem_ptr, chip_id_t chip, tt::umd::CoreCoord core, uint64_t addr, uint32_t size);
+    virtual void dma_write_to_device(
+        const void* src, size_t size, chip_id_t chip, tt::umd::CoreCoord core, uint64_t addr);
+    virtual void dma_read_from_device(void* dst, size_t size, chip_id_t chip, tt::umd::CoreCoord core, uint64_t addr);
     std::optional<std::tuple<uint32_t, uint32_t>> get_tlb_data_from_target(
         const chip_id_t chip, const tt::umd::CoreCoord core);
     tlb_configuration get_tlb_configuration(const chip_id_t chip, const tt::umd::CoreCoord core);
--- a/device/api/umd/device/pci_device.hpp
+++ b/device/api/umd/device/pci_device.hpp
@@ -38,6 +38,15 @@ struct PciDeviceInfo {
 // Do we want to put everything into this file into tt::umd namespace?
 using tt::umd::semver_t;
 
+struct DmaBuffer {
+    uint8_t *buffer{nullptr};
+    uint8_t *completion{nullptr};
+    size_t size{0};
+
+    uint32_t buffer_pa{0};
+    uint32_t completion_pa{0};
+};
+
 class PCIDevice {
     const std::string device_path;   // Path to character device: /dev/tenstorrent/N
     const int pci_device_num;        // N in /dev/tenstorrent/N
@@ -48,6 +57,7 @@ class PCIDevice {
     const tt::ARCH arch;             // e.g. Grayskull, Wormhole, Blackhole
     const semver_t kmd_version;      // KMD version
     const bool iommu_enabled;        // Whether the system is protected from this device by an IOMMU
+    DmaBuffer dma_buffer{};
 
 public:
     /**
@@ -143,6 +153,13 @@ public:
      * @return uint64_t PA (no IOMMU) or IOVA (with IOMMU) for use by the device
      */
     uint64_t map_for_dma(void *buffer, size_t size);
+
+    /**
+     * Access the device's DMA buffer.  This buffer is not guaranteed to exist.
+     * It is the caller's responsibility to check if the buffer is valid and to
+     * chunk the desired transfer size to fit within it.
+     */
+    DmaBuffer &get_dma_buffer() { return dma_buffer; }
 
 public:
     // TODO: we can and should make all of these private.
--- a/device/api/umd/device/pci_device.hpp
+++ b/device/api/umd/device/pci_device.hpp
@@ -39,12 +39,12 @@ struct PciDeviceInfo {
 using tt::umd::semver_t;
 
 struct DmaBuffer {
-    uint8_t *buffer{nullptr};
-    uint8_t *completion{nullptr};
-    size_t size{0};
+    uint8_t *buffer = nullptr;
+    uint8_t *completion = nullptr;
+    size_t size = 0;
 
-    uint32_t buffer_pa{0};
-    uint32_t completion_pa{0};
+    uint32_t buffer_pa = 0;
+    uint32_t completion_pa = 0;
 };
 
 class PCIDevice {
--- a/device/api/umd/device/tt_device/wormhole_tt_device.h
+++ b/device/api/umd/device/tt_device/wormhole_tt_device.h
@@ -6,6 +6,8 @@
 
 #pragma once
 
+#include <mutex>
+
 #include "umd/device/tt_device/tt_device.h"
 
 namespace tt::umd {
@@ -26,5 +28,13 @@ public:
     BoardType get_board_type() override;
 
     std::vector<DramTrainingStatus> get_dram_training_status() override;
+
+    void dma_d2h(void *dst, uint32_t src, size_t size) override;
+    void dma_h2d(uint32_t dst, const void *src, size_t size) override;
+
+private:
+    // Enforce single-threaded access, even though there are more serious issues
+    // surrounding resource management as it relates to DMA.
+    std::mutex dma_mutex_;
 };
 }  // namespace tt::umd
--- a/device/pcie/pci_device.cpp
+++ b/device/pcie/pci_device.cpp
@@ -298,6 +298,19 @@ PCIDevice::PCIDevice(int pci_device_number) :
 
         system_reg_start_offset = (512 - 16) * 1024 * 1024;
         system_reg_offset_adjust = (512 - 32) * 1024 * 1024;
+
+        bar2_uc_size = bar2_uc_mapping.mapping_size;
+        bar2_uc = mmap(
+            NULL,
+            bar2_uc_mapping.mapping_size,
+            PROT_READ | PROT_WRITE,
+            MAP_SHARED,
+            pci_device_file_desc,
+            bar2_uc_mapping.mapping_base);
+
+        if (bar2_uc == MAP_FAILED) {
+            throw std::runtime_error(fmt::format("BAR2 UC mapping failed for device {}.", pci_device_num));
+        }
     } else if (arch == tt::ARCH::BLACKHOLE) {
         if (bar2_uc_mapping.mapping_id != TENSTORRENT_MAPPING_RESOURCE1_UC) {
             throw std::runtime_error(fmt::format("Device {} has no BAR2 UC mapping.", pci_device_num));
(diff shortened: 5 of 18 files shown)


device/api/umd/device/tt_device/wormhole_tt_device.h
```diff
@@ -26,5 +28,13 @@ class WormholeTTDevice : public TTDevice {
...
 
     std::vector<DramTrainingStatus> get_dram_training_status() override;
+
+    void dma_d2h(void *dst, uint32_t src, size_t size) override;
+    void dma_h2d(uint32_t dst, const void *src, size_t size) override;
+
+private:
+    // Enforce single-threaded access, even though there are more serious issues
+    // surrounding resource management as it relates to DMA.
+    std::mutex dma_mutex_;
```
pjanevskiTT, line 38: can mutexes that were created in #693 be used?

device/api/umd/device/chip/local_chip.h
```diff
@@ -38,6 +38,9 @@ class LocalChip : public Chip {
     void read_from_device(
         tt_xy_pair core, void* dest, uint64_t l1_src, uint32_t size, const std::string& fallback_tlb) override;
 
+    virtual void dma_write_to_device(const void* src, size_t size, tt_xy_pair core, uint64_t addr) override;
```
pjanevskiTT, line 41: you don't need virtual when overriding the function. That is the unofficial policy we followed so far...hope to put things like this in a README soon or just point people to the official rules we are following

device/api/umd/device/cluster.h
```diff
@@ -225,7 +225,29 @@ class tt_device {
     }
 
     /**
-     * Write uint32_t data (as specified by ptr + len pair) to specified device, core and address (defined for Silicon).
+     * Read data from a specified device, core and address to host memory (defined for Silicon).
+     * This API is used for reading from both TENSIX and DRAM cores. The internal SocDescriptor can be used to determine
```
pjanevskiTT, line 229: nit: `from both TENSIX and DRAM cores.`. The API is used to read from all types of cores, not sure why mentioning just these two

device/api/umd/device/cluster.h
```diff
@@ -246,6 +268,30 @@ class tt_device {
...
+        const void* src, size_t size, chip_id_t chip, tt::umd::CoreCoord core, uint64_t addr) = 0;
+
+    /**
+     * Use PCIe DMA to read device memory (L1 or DRAM).
+     *
+     * @param src Source data address.
+     * @param size Size in bytes.
+     * @param chip Chip to target; must be local, i.e. attached via PCIe.
+     * @param core Core to target.
+     * @param addr NOC address in the target core.
```
pjanevskiTT, line 290: nit: can comment for addr param be the same as read_from_device 

>  @param addr Address to read from.

Is there some specific reason to specify this is NOC address. I think we should just keep it consistent

device/api/umd/device/chip/local_chip.h
```diff
@@ -38,6 +38,9 @@ class LocalChip : public Chip {
...
+    virtual void dma_read_from_device(void* dst, size_t size, tt_xy_pair core, uint64_t addr) override;
```
pjanevskiTT, line 42: nit: this API could also be in CoreCoord and then you can use chip's internal function for translation

tests/wormhole/test_cluster_wh.cpp
```diff
@@ -974,3 +974,163 @@ TEST(SiliconDriverWH, LargeAddressTlb) {
...
+        false,  // skip driver allocs - no (don't skip)
+        true,   // clean system resources - yes
+        true);  // perform harvesting - yes
+
+    cluster.start_device(tt_device_params{});
+
+    // This seems to return 3 cores for each controller -- not what we want.
+    /* const auto drams = cluster.get_soc_descriptor(mmio_chip_id).get_cores(CoreType::DRAM); */
+    // Instead, just hardcode the DRAM cores... they're not going to change.
+    std::vector<tt_xy_pair> drams = {{0, 6}, {0, 1}, {5, 6}, {5, 9}, {5, 4}, {5, 1}};
```
pjanevskiTT, line 1007: https://github.com/tenstorrent/tt-umd/issues/740 follow up issue for us. No action for you on this PR

tests/wormhole/test_cluster_wh.cpp
```diff
@@ -974,3 +974,163 @@ TEST(SiliconDriverWH, LargeAddressTlb) {
...
+        // Now, read back the patterns we wrote to DRAM and verify them.
+        for (const auto& op : write_ops) {
+            std::vector<uint8_t> readback(op.data.size());
+
+            // Perform the DMA read from the specific address.
+            auto now = std::chrono::steady_clock::now();
+            cluster.dma_read_from_device(readback.data(), readback.size(), chip, op.core, op.address);
+            auto end = std::chrono::steady_clock::now();
+            auto ns = std::chrono::duration_cast<std::chrono::nanoseconds>(end - now).count();
+            // print_speed("D2H", readback.size(), ns);
```
pjanevskiTT, line 1129: Since we still haven't setup the test organization properly, I am fine with seeing the print of the DMA speed for read/write. That is eventually going to end up at API section of the tests

device/pcie/pci_device.cpp
```diff
@@ -336,6 +349,67 @@ PCIDevice::PCIDevice(int pci_device_number) :
...
+    // an interrupt.
+    if (arch == tt::ARCH::WORMHOLE_B0) {
+        const uint32_t buf_size = (1 << 20);  // 1 MiB
+        tenstorrent_allocate_dma_buf dma_buf{};
+
+        dma_buf.in.requested_size = buf_size + 0x1000;
+        dma_buf.in.buf_index = 0;
+
+        if (ioctl(pci_device_file_desc, TENSTORRENT_IOCTL_ALLOCATE_DMA_BUF, &dma_buf)) {
+            // There is a chance this will fail because we're not requiring
```
pjanevskiTT, line 382: Nice comments for the explanation. Can you just open the issues to be aware of handling this use case properly?

device/tt_device/wormhole_tt_device.cpp
```diff
@@ -105,4 +106,170 @@ std::vector<DramTrainingStatus> WormholeTTDevice::get_dram_training_status() {
...
+    // Reset completion flag.
+    *completion = 0;
+
+    auto write_reg = [&](uint32_t offset, uint32_t value) {
+        *reinterpret_cast<volatile uint32_t *>(bar2 + offset) = value;
+    };
+
+    write_reg(DMA_WRITE_ENGINE_EN_OFF, 0x1);
+    write_reg(DMA_WRITE_INT_MASK_OFF, 0);
+    write_reg(DMA_CH_CONTROL1_OFF_WRCH_0, 0x04000010);
```
pjanevskiTT, line 164: what's the magic value 0x04000010?

device/tt_device/wormhole_tt_device.cpp
```diff
@@ -105,4 +106,170 @@ std::vector<DramTrainingStatus> WormholeTTDevice::get_dram_training_status() {
...
+    write_reg(DMA_WRITE_CH01_IMWR_DATA_OFF, DMA_COMPLETION_VALUE);
+    write_reg(DMA_WRITE_DONE_IMWR_HIGH_OFF, 0);
+    write_reg(DMA_WRITE_ABORT_IMWR_LOW_OFF, 0);
+    write_reg(DMA_WRITE_ABORT_IMWR_HIGH_OFF, 0);
+    write_reg(DMA_TRANSFER_SIZE_OFF_WRCH_0, size);
+    write_reg(DMA_SAR_LOW_OFF_WRCH_0, src);
+    write_reg(DMA_SAR_HIGH_OFF_WRCH_0, 0);
+    write_reg(DMA_DAR_LOW_OFF_WRCH_0, dma_buffer.buffer_pa);
+    write_reg(DMA_DAR_HIGH_OFF_WRCH_0, 0);
+    write_reg(DMA_WRITE_DOORBELL_OFF, 0);
```
pjanevskiTT, line 175: question for my understanding. Is setting the doorbell register to 0 actually triggering the DMA engine? Is it set back to 1 automatically on completion, since we don't set it as far as i can see?

pjanevskiTT (approved):
Looks good, left mostly cleanup comments and some to understand the DMA engine better

tests/wormhole/test_cluster_wh.cpp
```diff
@@ -974,3 +974,163 @@ TEST(SiliconDriverWH, LargeAddressTlb) {
...
+    auto rate = megabytes / seconds;
+    std::cout << direction << ": 0x" << std::hex << bytes << std::dec << " bytes in " << ns << " ns (" << rate
+              << " MiB/s)" << std::endl;
+};
+
+/**
+ * Test the PCIe DMA controller by using it to write random fixed-size patterns
+ * to 0x0 in several DRAM cores, then reading them back and verifying.
+ */
+TEST(SiliconDriverWH, DMA1) {
```
pjanevskiTT, line 991: I think additional test that does write using Cluster's write_to_device and reads using the DMA can be useful as an example as well

device/api/umd/device/tt_device/wormhole_tt_device.h
joelsmithTT: It could, but it's not needed.  The layer above holds a shared mutex for the TLB window that's used, so any user of UMD won't be able to break things with multiple processes.  This mutex is mainly to document that within UMD it is not possible to use multiple threads to speed things up.  (This is in contrast to MMIO access, which could have its performance improved by breaking transfers into multiple chunks with one thread per chunk, all using the same TLB window).

device/tt_device/wormhole_tt_device.cpp
joelsmithTT: Hardware sees a write to this register even if the value doesn't change.  In this case the 0 is the channel number.  I'll add some comments for this and the magic 0x04000010.

device/tt_device/wormhole_tt_device.cpp
joelsmithTT: I went trying to figure out where the 0x04000000 came from.  It shows up in a DMA programming example in the documentation for the controller.  The example is 0x04000008 (local interrupt enable=1) rather than 0x04000010 (remote interrupt enable).  I think the documentation is just wrong, but was probably used as a reference for the initial ARC FW implementation.  I might be missing something...

device/api/umd/device/cluster.h
joelsmithTT: That was already there.  Since I moved the methods I figured I should remove the stale reference to uint32_t from the documentation.  But you are right, so I wrote new documentation for both `read_from_device` and `write_to_device` which better describes their behavior.  Please take another look.

Since I hope we can abolish this fallback_tlb thing entirely, I am not going to bother reconciling the inconsistency between "tlb_to_use" and "fallback_tlb".

device/pcie/pci_device.cpp
joelsmithTT: Added https://github.com/tenstorrent/tt-umd/issues/746

tests/wormhole/test_cluster_wh.cpp
joelsmithTT: I added that in the second test.

device/api/umd/device/cluster.h
joelsmithTT: Actually, disregard that.  The API was augmented and the remark about "TENSIX and DRAM" was duplicated in more places.  I am not going to clean that up in this PR.  I backed out these documentation changes and will address this in https://github.com/tenstorrent/tt-umd/issues/753.

device/api/umd/device/pci_device.hpp
```diff
@@ -38,6 +38,15 @@ struct PciDeviceInfo {
 // Do we want to put everything into this file into tt::umd namespace?
 using tt::umd::semver_t;
 
+struct DmaBuffer {
+    uint8_t *buffer{nullptr};
```
broskoTT, line 42: nit: these {} for initialization of all members seem a bit strange, haven't seen them before for primitive types? Maybe just a regular = nullptr, also for the rest

device/api/umd/device/pci_device.hpp
```diff
@@ -48,6 +57,7 @@ class PCIDevice {
     const tt::ARCH arch;             // e.g. Grayskull, Wormhole, Blackhole
     const semver_t kmd_version;      // KMD version
     const bool iommu_enabled;        // Whether the system is protected from this device by an IOMMU
+    DmaBuffer dma_buffer{};
```
broskoTT, line 60: No need for {}

device/tt_device/wormhole_tt_device.cpp
```diff
@@ -105,4 +106,170 @@ std::vector<DramTrainingStatus> WormholeTTDevice::get_dram_training_status() {
...
+    static constexpr uint64_t DMA_WRITE_ABORT_IMWR_LOW_OFF = 0x68;
+    static constexpr uint64_t DMA_WRITE_ABORT_IMWR_HIGH_OFF = 0x6c;
+    static constexpr uint64_t DMA_TRANSFER_SIZE_OFF_WRCH_0 = 0x208;
+    static constexpr uint64_t DMA_SAR_LOW_OFF_WRCH_0 = 0x20c;
+    static constexpr uint64_t DMA_SAR_HIGH_OFF_WRCH_0 = 0x210;
+    static constexpr uint64_t DMA_DAR_LOW_OFF_WRCH_0 = 0x214;
+    static constexpr uint64_t DMA_DAR_HIGH_OFF_WRCH_0 = 0x218;
+    static constexpr uint64_t DMA_WRITE_DOORBELL_OFF = 0x10;
+
+    std::scoped_lock lock(dma_mutex_);
```
broskoTT, line 130: If this is supposed to guard between mutliple threads/processes globaly on the host for a single device, why not lock_manager?

device/tt_device/wormhole_tt_device.cpp
```diff
@@ -105,4 +106,170 @@ std::vector<DramTrainingStatus> WormholeTTDevice::get_dram_training_status() {
...
+
+        auto now = std::chrono::steady_clock::now();
+        auto elapsed_ms = std::chrono::duration_cast<std::chrono::milliseconds>(now - start).count();
+
+        if (elapsed_ms > DMA_TIMEOUT_MS) {
+            throw std::runtime_error("DMA timeout");
+        }
+    }
+
+    memcpy(dst, dma_buffer.buffer, size);
```
broskoTT, line 191: Oh so when reading, we read into a preallocated host buffer, then copy from it to our memory? It is harder to pass dst directly to the dma engine since we need a call into kmd to obtain the physical address, is that the only reason?

device/tt_device/wormhole_tt_device.cpp
```diff
@@ -105,4 +106,170 @@ std::vector<DramTrainingStatus> WormholeTTDevice::get_dram_training_status() {
...
+    static constexpr uint64_t DMA_WRITE_CH01_IMWR_DATA_OFF = 0x70;
+    static constexpr uint64_t DMA_WRITE_DONE_IMWR_HIGH_OFF = 0x64;
+    static constexpr uint64_t DMA_WRITE_ABORT_IMWR_LOW_OFF = 0x68;
+    static constexpr uint64_t DMA_WRITE_ABORT_IMWR_HIGH_OFF = 0x6c;
+    static constexpr uint64_t DMA_TRANSFER_SIZE_OFF_WRCH_0 = 0x208;
+    static constexpr uint64_t DMA_SAR_LOW_OFF_WRCH_0 = 0x20c;
+    static constexpr uint64_t DMA_SAR_HIGH_OFF_WRCH_0 = 0x210;
+    static constexpr uint64_t DMA_DAR_LOW_OFF_WRCH_0 = 0x214;
+    static constexpr uint64_t DMA_DAR_HIGH_OFF_WRCH_0 = 0x218;
+    static constexpr uint64_t DMA_WRITE_DOORBELL_OFF = 0x10;
```
broskoTT, line 128: Maybe these constants should live somewhere else, in wormhole_implementation.h for example?

tests/wormhole/test_cluster_wh.cpp
```diff
@@ -974,3 +974,211 @@ TEST(SiliconDriverWH, LargeAddressTlb) {
...
+        target_devices,
+        1,      // one "host memory channel", unused
+        false,  // skip driver allocs - no (don't skip)
+        true,   // clean system resources - yes
+        true);  // perform harvesting - yes
+
+    cluster.start_device(tt_device_params{});
+
+    // This seems to return 3 cores for each controller -- not what we want.
+    /* const auto drams = cluster.get_soc_descriptor(mmio_chip_id).get_cores(CoreType::DRAM); */
```
broskoTT, line 1005: yeah you could do a for loop up to get_num_dram_channels and then get_core_for_dram_channel(dram_chan, 0 /*subchan*/)

device/pcie/pci_device.cpp
```diff
@@ -336,6 +349,67 @@ PCIDevice::PCIDevice(int pci_device_number) :
...
+    //    - Driver initiates DMAs based on its knowledge of the buffer
+    // 5. + 0x1000 is for the completion page.  Since this entire implementation
+    // is a temporary hack until it's implemented in the driver, we'll need to
+    // poll a completion page to know when the DMA is done instead of receiving
+    // an interrupt.
+    if (arch == tt::ARCH::WORMHOLE_B0) {
+        const uint32_t buf_size = (1 << 20);  // 1 MiB
+        tenstorrent_allocate_dma_buf dma_buf{};
+
+        dma_buf.in.requested_size = buf_size + 0x1000;
```
broskoTT, line 378: Why 0x1000? And why is this not already calculated in buf size?

broskoTT (approved):

device/tt_device/wormhole_tt_device.cpp
joelsmithTT: It's not.  The purpose is to serve as documentation that interaction with the DMA registers is not thread-safe.  The multi-process protection is conferred by the TLB mutex provided by the lock manager.

The rationale was: within a single process it is legal to chunk a MMIO transfer and use multiple threads to perform the operation on non-overlapping chunks through the same TLB window.  Doing this with the DMA engine is not safe.  The mutex is an explicit mechanism to prevent anyone from attempting this optimization.

device/tt_device/wormhole_tt_device.cpp
joelsmithTT: Yes, there is a pre-allocated buffer in the way.  The memcpy imposes a performance penalty.

RE: having the DMA target dst directly -- there are additional reasons besides needing the physical address:
* dst may not be physically contiguous
* dst would need to be pinned

tt-kmd can handle this case today, but it needs IOMMU enabled to deal with the contiguity problem.

With an IOMMU-based technique we could have a zero-copy mechanism where the user buffer is accessed directly by the hardware.  However this too comes at a cost.  The driver would need to set up a mapping for the buffer prior to the DMA.  Depending on the size of the buffer, the overhead imposed by the KMD interactions:
```
uint64_t dma_addr_dst = kmd.map_for_dma(dst, size);
do_the_dma(dma_addr_dst, dma_addr_src, size);
kmd.unmap_for_dma(dma_addr_dst, size);
```
might take longer than the memcpy.  A scheme where DMA-accessible, application-allocated buffers are configured up-front (one-time mapping cost) would be appropriate here.

device/tt_device/wormhole_tt_device.cpp
joelsmithTT: That moves them further away from where they are used.  If they're used in multiple places then we are probably doing something wrong.

tests/wormhole/test_cluster_wh.cpp
joelsmithTT: Ok, fixed.

device/api/umd/device/pci_device.hpp
joelsmithTT: I'll change to use =

device/api/umd/device/pci_device.hpp
joelsmithTT: True because of the default member initializers (formerly {}, now =) in `DmaBuffer`.
But {} makes it explicit that `dma_buffer` is initialized to its default state (all nullptr/zeros), which is important because there's a code path that involves this scenario (see the remark at the `DmaBuffer &get_dma_buffer()` method documentation).

Issue #700, opened 2025-04-07
Unused PCI core on Blachkole stored in router cores


pjanevskiTT:
Unused PCI core on Blackhole is stored together with router cores on Blackhole. This is a problem since the memory map for these core types is not the same, so if someone wants to perform same operation on all "router" cores there could be problems

Example of this is reading NOC node id, problems arise because NOC register base is not same on PCI core and router cores.

Solution can be adding additional core type for unused PCI core, or storing PCI core properly and then guaranteeing that PCI core on position 0 is the one used etc...

broskoTT:
This was reported by the tt-exalens team

pjanevskiTT:
closed by #742

PR #742, opened 2025-04-14
Implement PCIE harvesting

pjanevskiTT:
### Issue

#700 

### Description

Implement PCIE harvesting in general. This is needed for Blackhole, but so far we have managed this based on the board type. This PR introduces relying on information from the device on PCIE port config. Even if this information should be 1-1 with board type, it is more reliable to trust the device configuration. 

### List of the changes

- Add pcie harvesting mask
- Add pcie harvesting in coordinate manager
- Remove fuzzy logic based on board type for blackhole
- Update tests
- Fix some tests to pass proper PCIE harvesting mask to structures

### Testing
CI

### API Changes
/

### DIFF
--- a/device/blackhole/blackhole_coordinate_manager.cpp
+++ b/device/blackhole/blackhole_coordinate_manager.cpp
@@ -199,6 +199,37 @@ void BlackholeCoordinateManager::translate_eth_coords() {
     }
 }
 
+void BlackholeCoordinateManager::translate_pcie_coords() {
+    size_t logical_x = 0;
+    for (size_t x = 0; x < pcie_grid_size.x; x++) {
+        if (harvesting_masks.pcie_harvesting_mask & (1 << x)) {
+            for (size_t y = 0; y < pcie_grid_size.y; y++) {
+                const tt_xy_pair& pcie_core = pcie_cores[x * pcie_grid_size.y + y];
+
+                CoreCoord logical_coord = CoreCoord(logical_x, y, CoreType::PCIE, CoordSystem::LOGICAL);
+                add_core_translation(logical_coord, pcie_core);
+
+                CoreCoord virtual_coord = CoreCoord(pcie_core.x, pcie_core.y, CoreType::PCIE, CoordSystem::VIRTUAL);
+                add_core_translation(virtual_coord, pcie_core);
+            }
+            logical_x++;
+        } else {
+            for (size_t y = 0; y < pcie_grid_size.y; y++) {
+                const tt_xy_pair& pcie_core = pcie_cores[x * pcie_grid_size.y + y];
+
+                CoreCoord virtual_coord = CoreCoord(pcie_core.x, pcie_core.y, CoreType::PCIE, CoordSystem::VIRTUAL);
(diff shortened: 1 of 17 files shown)


device/blackhole/blackhole_coordinate_manager.cpp
```diff
@@ -227,17 +258,38 @@ void BlackholeCoordinateManager::fill_eth_physical_translated_mapping() {
...
+    for (size_t x = 0;
+         x < pcie_grid_size.x - CoordinateManager::get_num_harvested(harvesting_masks.pcie_harvesting_mask);
+         x++) {
+        CoreCoord logical_coord = CoreCoord(x, 0, CoreType::PCIE, CoordSystem::LOGICAL);
+        const tt_xy_pair physical_pair = to_physical_map[logical_coord];
+        size_t translated_x = physical_pair.x;
+        size_t translated_y = physical_pair.y;
+
+        // Only the first PCIE core is mapped to the translated coordinates.
+        // This is in case pcie harvesting mask is set to 0.
```
broskoTT, line 270: In case harvesting mask is 0, then we'll have two physical cores mapped to the same translated core?
Can we instead remove this if (x == 0) branch, and have an assert somewhere in core so that we throw if pcie_harvesting_mask is 0, if we don't support that case?

device/blackhole/blackhole_coordinate_manager.cpp
```diff
@@ -197,6 +197,37 @@ void BlackholeCoordinateManager::translate_eth_coords() {
...
+                CoreCoord virtual_coord = CoreCoord(pcie_core.x, pcie_core.y, CoreType::PCIE, CoordSystem::VIRTUAL);
+                add_core_translation(virtual_coord, pcie_core);
+            }
+            logical_x++;
+        } else {
+            for (size_t y = 0; y < pcie_grid_size.y; y++) {
+                const tt_xy_pair& pcie_core = pcie_cores[x * pcie_grid_size.y + y];
+
+                CoreCoord virtual_coord = CoreCoord(pcie_core.x, pcie_core.y, CoreType::PCIE, CoordSystem::VIRTUAL);
+                add_core_translation(virtual_coord, pcie_core);
```
broskoTT, line 219: A surprising ammount of code is added to add these translations.. I might take a look at one point and try to make it simpler. Not action for this PR.

broskoTT (approved):
Great, thanks!

device/blackhole/blackhole_coordinate_manager.cpp
pjanevskiTT: So far all the harvesting we have is either by rows of the vector or by column. I think this can be a place to generalize this kind of stuff

device/blackhole/blackhole_coordinate_manager.cpp
pjanevskiTT: It requires changing a lot of tests, I will do this in a separate PR. We have a lot of places where we pass all zeros to harvesting masks. Should enforce the limits more strictly

PR #747, opened 2025-04-15
Implement static TLB configuration

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/44
Also on path of #248 

### Description
Mostly taken from tt_metal

### List of the changes

### Testing
TBD: testing on tt_metal tests, with removed tlb_config.cpp in tt_metal

### API Changes

### DIFF
--- a/device/api/umd/device/grayskull_implementation.h
+++ b/device/api/umd/device/grayskull_implementation.h
@@ -269,6 +269,12 @@ public:
 
     uint32_t get_dynamic_tlb_16m_cfg_addr() const override { return grayskull::DYNAMIC_TLB_16M_CFG_ADDR; }
 
+    uint32_t get_tlb_4g_base_index() const override { return 0; }
+
+    uint32_t get_dynamic_tlb_base_index() const override { return 0; }
+
+    uint32_t get_dynamic_tlb_count() const override { return 0; }
+
     uint32_t get_mem_large_read_tlb() const override { return grayskull::MEM_LARGE_READ_TLB; }
 
     uint32_t get_mem_large_write_tlb() const override { return grayskull::MEM_LARGE_WRITE_TLB; }
--- a/device/chip/local_chip.cpp
+++ b/device/chip/local_chip.cpp
@@ -74,6 +74,58 @@ void LocalChip::initialize_tlb_manager() {
     tlb_manager_->set_dynamic_tlb_config("REG_TLB", tt_device_->get_architecture_implementation()->get_reg_tlb());
     tlb_manager_->set_dynamic_tlb_config(
         "SMALL_READ_WRITE_TLB", tt_device_->get_architecture_implementation()->get_small_read_write_tlb());
+
+    auto eth_grid = soc_descriptor_.get_grid_size(CoreType::ETH);
+    auto tensix_grid = soc_descriptor_.get_grid_size(CoreType::TENSIX);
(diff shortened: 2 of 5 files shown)


device/chip/local_chip.cpp
```diff
@@ -76,6 +76,60 @@ void LocalChip::initialize_tlb_manager() {
...
+        // Revisit this when we have a more flexible UMD api
+        tlb_manager_->configure_tlb(virtual_core, translated_core, tlb_index, 0, tlb_data::Strict);
+    }
+
+    if (tt_device_->get_arch() == tt::ARCH::WORMHOLE_B0) {
+        // Setup static TLBs for MMIO mapped data space in DRAM.
+        CoreCoord virtual_core = soc_descriptor_.get_dram_core_for_channel(0, 0, CoordSystem::VIRTUAL);
+        CoreCoord translated_core = soc_descriptor_.translate_coord_to(virtual_core, CoordSystem::TRANSLATED);
+        architecture_implementation* arch_impl = tt_device_->get_architecture_implementation();
+        uint64_t peer_dram_offset = arch_impl->get_dram_channel_0_peer2peer_region_start();
```
joelsmithTT, line 115: peer2peer?!

device/chip/local_chip.cpp
```diff
@@ -76,6 +76,60 @@ void LocalChip::initialize_tlb_manager() {
...
+        for (uint32_t tlb_id = arch_impl->get_dynamic_tlb_base_index();
+             tlb_id < arch_impl->get_dynamic_tlb_base_index() + arch_impl->get_dynamic_tlb_count();
+             tlb_id++) {
+            tlb_manager_->configure_tlb(virtual_core, translated_core, tlb_id, peer_dram_offset, tlb_data::Relaxed);
+            // Align address space of 16MB TLB to 16MB boundary
+            peer_dram_offset += arch_impl->get_dynamic_tlb_16m_size();
```
joelsmithTT, line 121: The function `get_dynamic_tlb_16m_size` returns the fixed value of 16 MiB.  Using the function here adds a layer of abstraction that obscures the simple constant nature of the value, especially since its primary cross-architecture difference seems to be checking existence (via exceptions), which isn't relevant in this usage.  Suggest replacing the call with the direct constant value (16 * 1024 * 1024 or a suitable named constant like TLB_16M_SIZE) for better readability.

The variable name `peer_dram_offset` is misleading in this context, as there doesn't seem to be a "peer" component involved in this operation.

device/chip/local_chip.cpp
broskoTT: it was just copied code...

pjanevskiTT (approved):
I'm fine with checking this in if it doesn't break things. However, I would discourage efforts to make current TLB system prettier and focus on integrating TLB management from KMD started in #733

device/api/umd/device/grayskull_implementation.h
```diff
@@ -269,6 +269,12 @@ class grayskull_implementation : public architecture_implementation {
 
     uint32_t get_dynamic_tlb_16m_cfg_addr() const override { return grayskull::DYNAMIC_TLB_16M_CFG_ADDR; }
 
+    uint32_t get_tlb_4g_base_index() const override { return 0; }
```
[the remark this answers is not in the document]
broskoTT: Yeah this is dead code, so i won't put an effort to modify it. we should delete this soon

Issue #737, opened 2025-04-14
Remove fallback_tlbs from API


broskoTT:
Instead of using string to communicate which kind of read/write we want, use different API calls.

Kind of started in https://github.com/tenstorrent/tt-umd/pull/679

PR #752, opened 2025-04-15
Offer membars without fallback_tlbs in API

broskoTT:
### Issue
On a path of #737 

### Description
With these changes we can now offer whole api free of fallback_tlbs to tt_metal, and when it changes to the new API we can remove fallback_tlb on our end.

### List of the changes
- l1_membar and dram_membar versions without fallback_tlb

### Testing
No testing

### API Changes
There are API changes in this PR, to be switched to in the clients

### DIFF
--- a/device/api/umd/device/cluster.h
+++ b/device/api/umd/device/cluster.h
@@ -380,6 +380,18 @@ public:
     }
 
     /**
+     * Tensix L1 memory barrier.
+     * This should be called when the client wants to ensure that all transactions on the L1 of the specified cores have
+     * completed.
+     *
+     * @param chip Chip to target.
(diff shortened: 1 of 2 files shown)


broskoTT:
> why not just have default value for fallback TLB for the already existing functions? That is basically what this PR is doing

Because of ordering :) If I set default value, then I can't change the current tt_metal calls in such a way that as the next step I could remove fallback_tlb from our API. I could for l1_membar though probably, but not dram_membar.

We have dram_membar(arg1, fallback_tlbs, arg2)

Since metal passes arg2, there is no way for me not to pass fallback_tlbs argument

pjanevskiTT (approved):

PR #757, opened 2025-04-16
Add API to cluster desc for unique chip id for 6U

pjanevskiTT:
### Issue

Add API to cluster desc for unique chip id for 6U

### Description

Add API to cluster desc for unique chip id for 6U. Chip unique id still needs work for old Wormhole configs and Blackhole. Checking this in just for 6U for now since it's helpful for tt-metal folks working on 6U

### List of the changes

- Add struct to cluster descriptor for chip uid
- Add API to get chip uid

### Testing
CI. Manual testing on 6U

### API Changes
/

### DIFF
--- a/device/tt_cluster_descriptor.cpp
+++ b/device/tt_cluster_descriptor.cpp
@@ -868,6 +868,8 @@ const std::unordered_map<chip_id_t, eth_coord_t> &tt_ClusterDescriptor::get_chip
     return locations;
 }
 
+// Note: this API works only for Wormhole 6U galaxy at the moment.
+// TODO: implement this for Blackhole and old Wormhole configurations.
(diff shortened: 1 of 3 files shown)


device/tt_cluster_descriptor.cpp
```diff
@@ -856,6 +856,10 @@ const std::unordered_map<chip_id_t, eth_coord_t> &tt_ClusterDescriptor::get_chip
     return locations;
 }
 
+const std::unordered_map<chip_id_t, uint64_t> &tt_ClusterDescriptor::get_chip_unique_ids() const {
+    return chip_unique_ids;
```
broskoTT, line 860: Do we need to add a warning, or log_fatal or just a comment that this doesn't work for bh and regular wh?

broskoTT (approved):

Issue #750, opened 2025-04-15
read_sysfs throws for success path


broskoTT:
This is degrading debug flow, since it will stop the debugger on throwing the exception.

![Image](https://github.com/user-attachments/assets/dff65664-8446-4842-a472-129359c25ad2)

joelsmithTT:
This one was my fault.

joelsmithTT:
This should be fixed.

PR #758, opened 2025-04-16
Avoid throwing an internal exception that's caught

joelsmithTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/750

### Description
Stop throwing and catching an exception internally

### List of the changes
* Refactored the `read_sysfs` functions to return `std::optional<T>` instead of throwing exceptions when a sysfs attribute cannot be read or parsed.
* Introduced a `try_read_sysfs` helper function that encapsulates the reading and parsing logic and returns `std::nullopt` on failure.
* Updated `detect_iommu` to use `try_read_sysfs` and handle the optional return value directly, removing the need for a try-catch block.

The throwing version of `read_sysfs` is kept for callers that expect an exception on failure. The version with a default value now uses `value_or`.

This change avoids using exceptions for control flow in cases where a sysfs attribute might not exist.

### Testing
Manual

### API Changes
None

### DIFF
--- a/device/pcie/pci_device.cpp
+++ b/device/pcie/pci_device.cpp
@@ -14,6 +14,7 @@
 
 #include <cstdint>
 #include <cstring>  // for memcpy
+#include <optional>
 #include <vector>
 
 #include "assert.hpp"
@@ -36,7 +37,7 @@ using namespace tt;
 using namespace tt::umd;
 
 template <typename T>
-static T read_sysfs(const PciDeviceInfo &device_info, const std::string &attribute_name) {
+static std::optional<T> try_read_sysfs(const PciDeviceInfo &device_info, const std::string &attribute_name) {
     const auto sysfs_path = fmt::format(
         "/sys/bus/pci/devices/{:04x}:{:02x}:{:02x}.{:x}/{}",
         device_info.pci_domain,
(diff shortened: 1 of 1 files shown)


device/pcie/pci_device.cpp
```diff
@@ -62,28 +63,40 @@ static T read_sysfs(const PciDeviceInfo &device_info, const std::string &attribu
...
+    auto result = try_read_sysfs<T>(device_info, attribute_name);
+    if (!result) {
+        const auto sysfs_path = fmt::format(
+            "/sys/bus/pci/devices/{:04x}:{:02x}:{:02x}.{:x}/{}",
+            device_info.pci_domain,
+            device_info.pci_bus,
+            device_info.pci_device,
+            device_info.pci_function,
+            attribute_name);
+        TT_THROW("Failed reading or parsing sysfs attribute: {}", sysfs_path);
```
broskoTT, line 83: I don't get it, you'll throw in case of nullopt? How is this better?

joelsmithTT: The path that accepts a default parameter if the attribute does not exist does not throw/catch.
This path does.

broskoTT: So if the file doesn't exist, the try_read_sysfs will return nullopt (line 52)?
nullopt will return true for (!result) ? 
And will throw?

joelsmithTT: My understanding of the issue was that certain devices lack an `iommu_group` directory, specifically on systems where there is no IOMMU (e.g. VMs), and that the exception that gets thrown and caught in this case is troubling for developers who have configured their gdb with `catch throw` or something.

Unless I am mistaken, that should be fixed.  An exception is no longer thrown and caught in this scenario.

The exception that is thrown if the sysfs attribute _is always expected to be present_ is retained.

broskoTT: Approving to unblock, since I'm probably missing something here, otherwise the tests would fail

broskoTT (approved):

PR #759, opened 2025-04-16
Move arc_msg to Chip

broskoTT:
### Issue
Part of #248

### Description
As a next step, moving arc_msg to Chip classes.

### List of the changes
- Moved pcie_arc_msg to LocalChip::arc_msg, and from remote_arc_msg to RemoteChip::arc_msg
- Moved enable_local_ethernet_queue and enable_remote_ethernet_queue into Chip::enable_ethernet_queue, since it was the same implementation for both local and remote chips.
- Define HANG_READ_VALUE in architecture_implementation.h to be available to all

### Testing
Existing CI tests

### API Changes

### DIFF
--- a/device/chip/remote_chip.cpp
+++ b/device/chip/remote_chip.cpp
@@ -63,4 +63,78 @@ void RemoteChip::wait_for_non_mmio_flush() {
     remote_communication_->wait_for_non_mmio_flush();
 }
 
+int RemoteChip::arc_msg(
+    uint32_t msg_code,
+    bool wait_for_done,
+    uint32_t arg0,
+    uint32_t arg1,
+    uint32_t timeout_ms,
+    uint32_t* return_3,
+    uint32_t* return_4) {
+    constexpr uint64_t ARC_RESET_SCRATCH_ADDR = 0x880030060;
+    constexpr uint64_t ARC_RESET_MISC_CNTL_ADDR = 0x880030100;
+
+    auto arc_core = soc_descriptor_.get_cores(CoreType::ARC).at(0);
+
+    if ((msg_code & 0xff00) != 0xaa00) {
+        log_error("Malformed message. msg_code is 0x{:x} but should be 0xaa..", msg_code);
+    }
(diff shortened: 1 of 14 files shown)


device/wormhole/wormhole_arc_messenger.cpp
```diff
@@ -74,9 +73,9 @@ uint32_t WormholeArcMessenger::send_message(
 
             exit_code = (status & 0xffff0000) >> 16;
             break;
-        } else if (status == MSG_ERROR_REPLY) {
+        } else if (status == HANG_READ_VALUE) {
```
pjanevskiTT, line 76: I don't know if MSG_ERROR_REPLY and HANG_READ_VALUE are semantically the same constant, or they just have the same value. I think having separate variables for these two improved readability

device/chip/remote_chip.cpp
```diff
@@ -63,4 +63,78 @@ void RemoteChip::wait_for_non_mmio_flush() {
     remote_communication_->wait_for_non_mmio_flush();
 }
 
+int RemoteChip::arc_msg(
```
pjanevskiTT, line 66: I think the proper place for remote_arc_msg to live is on TTDevice layer, or whatever we use as remote TTDevice at the moment. Reason behind this is that sometimes we will need ARC msg even before we construct the chip. For example on Wormhole harvesting mask for the chip is collected using remote_arc_msg, so we would want to have an abstraction that can get us this info even before constructing RemoteChip. I needed to implemented this in #722 in TopologyDiscovery. I think ideal place for this is RemoteCommunication, this way we can reuse it both in topology discovery and here inside RemoteChip

broskoTT: It needs to be part of the chip API in anycase, but I agree that I can move the implementation to remote communication, same as tt_device encapsulates it's own arc message implementation for local chips...

broskoTT: There might be some complexities here, and I want to do it right, so I'll leave it for some further PR https://github.com/tenstorrent/tt-umd/issues/766

pjanevskiTT (approved):

PR #761, opened 2025-04-17
Move check_pcie_device to Chip

broskoTT:
### Issue
On a path of #248

### Description
Moving check pcie_device to local chip

### List of the changes
- Moved check_pcie_device_initialized and test_setup_interface to local_chip
- Started the "start_device" function in Chip, which should hold all the code from cluster start_device

### Testing
No additional testing

### API Changes

### DIFF
--- a/device/chip/local_chip.cpp
+++ b/device/chip/local_chip.cpp
@@ -11,6 +11,7 @@
 #include "umd/device/chip_helpers/tlb_manager.h"
 #include "umd/device/tt_device/tt_device.h"
 #include "umd/device/types/blackhole_eth.h"
+#include "umd/device/wormhole_implementation.h"
 
 extern bool umd_use_noc1;
 
@@ -106,6 +107,8 @@ SysmemManager* LocalChip::get_sysmem_manager() { return sysmem_manager_.get(); }
 TLBManager* LocalChip::get_tlb_manager() { return tlb_manager_.get(); }
 
 bool LocalChip::is_mmio_capable() const { return true; }
+
+void LocalChip::start_device() { check_pcie_device_initialized(); }
 
 void LocalChip::wait_eth_cores_training(const uint32_t timeout_ms) {
     if (get_tt_device()->get_arch() != tt::ARCH::BLACKHOLE) {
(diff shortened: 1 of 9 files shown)


device/chip/local_chip.cpp
```diff
@@ -497,4 +499,70 @@ int LocalChip::arc_msg(
     return exit_code;
 }
 
+void LocalChip::check_pcie_device_initialized() {
+    auto architecture_implementation = tt_device_->get_architecture_implementation();
+
+    // MT Initial BH - Add check for blackhole once access to ARC registers is setup through TLBs
```
pjanevskiTT, line 505: can we add this MT comment to the issue and delete the comment or reference the issue? We already have some kind of check whether ARC is initialized for Blackhole on TTDevice constructor, but we may need to introduce additional checks

device/cluster.cpp
```diff
@@ -532,58 +532,11 @@ void Cluster::configure_active_ethernet_cores_for_mmio_device(
...
-    }
-}
-
 void Cluster::initialize_pcie_devices() {
     log_debug(LogSiliconDriver, "Cluster::start");
 
-    for (auto chip_id : local_chip_ids_) {
-        check_pcie_device_initialized(chip_id);
+    for (auto chip_id : all_chip_ids_) {
+        get_chip(chip_id)->start_device();
```
pjanevskiTT, line 539: since we already have start_device function, I find it a bit confusing that Cluster::initialize_pcie_devices is calling Chip::start_device, why not called the function that has the same name. So in the future when everything from Cluster's start_device is moved to Chip, we can just call Chip's start device and remove Cluster::initialize_pcie_devices

broskoTT: Currently start_device in cluster calls
initialize_pcie_devices();
verify_eth_fw();
deassert_resets_and_set_power_state();

and then initialize_pcie_devices calls
check_pcie_device_initialized(chip_id);
init_pcie_iatus();
init_membars();

And I find this slightly confusing, I'd prefer to unroll initialize_pcie_devices and that cluster.start_device calls all of these. Then from that point I'd remove function calls one by one and move them to chip->start_device function.

I've just skipped the intermediate step, and immediatelly started moving stuff to chip->start_device, so that's why I moved it from initialize_pcie_devices.

pjanevskiTT (approved):

PR #762, opened 2025-04-17
Move configure_iatu_region from Cluster

broskoTT:
### Issue
On a path to #248 but also #142 

### Description
Move iatu configuration to tt_device, and also its calling to local_chip

### List of the changes
- Merged bh and wh declarations for configure_iatu_region
- Moved wh's definition for configure_iatu_region to tt_device
- Moved init_pcie_iatus  to local_chip and Calling this from start_device

### Testing
No additional testing

### API Changes

### DIFF
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -61,9 +61,6 @@ using namespace tt::umd;
 
 extern bool umd_use_noc1;
 
-// Remove 256MB from full 1GB for channel 3 (iATU limitation)
-static constexpr uint32_t HUGEPAGE_CHANNEL_3_SIZE_LIMIT = 805306368;
-
 static constexpr uint32_t REMOTE_CMD_NOC_BIT = 9;
 
 // --------------------------------------------------------------------------------------------------------------
@@ -540,8 +537,6 @@ void Cluster::initialize_pcie_devices() {
         get_chip(chip_id)->start_device();
     }
 
-    init_pcie_iatus();
-
     init_membars();
 }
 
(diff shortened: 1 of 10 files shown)


device/tt_device/wormhole_tt_device.cpp
```diff
@@ -106,6 +107,37 @@ std::vector<DramTrainingStatus> WormholeTTDevice::get_dram_training_status() {
...
+    if (region == 3) {
+        region_id_to_use = 4;  // Hack use region 4 for channel 3..this ensures that we have a smaller chan 3 address
+                               // space with the correct start offset
+    }
+
+    bar_write32(architecture_impl_->get_arc_csm_mailbox_offset() + 0 * 4, region_id_to_use);
+    bar_write32(architecture_impl_->get_arc_csm_mailbox_offset() + 1 * 4, dest_bar_lo);
+    bar_write32(architecture_impl_->get_arc_csm_mailbox_offset() + 2 * 4, dest_bar_hi);
+    bar_write32(architecture_impl_->get_arc_csm_mailbox_offset() + 3 * 4, region_size);
+    std::vector<uint32_t> arc_msg_return_values;
```
pjanevskiTT, line 125: there should be API for ARC messenger you can use if you don't want any return values

https://github.com/tenstorrent/tt-umd/blob/main/device/api/umd/device/arc_messenger.h#L56

device/cluster.cpp
```diff
@@ -539,8 +536,6 @@ void Cluster::initialize_pcie_devices() {
         get_chip(chip_id)->start_device();
     }
 
-    init_pcie_iatus();
```
pjanevskiTT, line 542: I guess we need to replace this call with calls to LocalChip objects for the same function. Maybe I am not seeing where this is called but I think you just deleted this call without replacing it

pjanevskiTT (approved):

pjanevskiTT (changes requested):
approved accidentally :). Want to see where the replacement in cluster.cpp is

device/cluster.cpp
broskoTT: It is called in LocalChip::start_device()

pjanevskiTT: I see, I probably reviewed the PRs in wrong order

pjanevskiTT (approved):

Issue #769, opened 2025-04-17
PCI harvesting not working on P100


broskoTT:
As reported by @abhullar-tt :
Im running
./build/test/tt_metal/unit_tests_device --gtest_filter=*DeviceInitializeAndTeardown/0*
 on yyz-lab-125 with branch abhullar/bh-virtual and seeing:
libc++abi: terminating due to uncaught exception of type std::runtime_error: P100 card should always have PCIE core (2, 0) harvested.

This happens on p100. There's probably a bug in logic in https://github.com/tenstorrent/tt-umd/pull/742

nxuTT:
Here is a harvesting cheatsheet, it outlines what get harvested for which product: https://tenstorrent.atlassian.net/wiki/spaces/syseng/pages/638615610/Product+Harvesting+Cheatsheet

pjanevskiTT:
closed by #772

PR #772, opened 2025-04-22
Fix PCI harvesting for P100 board

pjanevskiTT:
### Issue

#769 

### Description

We have two types of P100 cards. They use different PCI cores on Blackhole so we can't rely on the check that used PCI core is always going to be the same. This PR removes the check from soc descriptor and changes some tests. 

### List of the changes

- Remove the check for harvested PCI core on P100
- Fix reverse logic for PCI harvesting in coordinate manager
- Change the tests to test the change

### Testing
CI. Change was manually tested on P100 and P100A boards to verify that tests pass.

### API Changes
/

### DIFF
--- a/device/tt_soc_descriptor.cpp
+++ b/device/tt_soc_descriptor.cpp
@@ -83,10 +83,12 @@ void tt_SocDescriptor::create_coordinate_manager(const BoardType board_type, con
         }
     }
 
+    // TODO: P100 has two types of boards, each using different PCI core.
+    // Either have two separate enums or completely remove the check here.
     // PCIE harvesting mask 0x1 corresponds to (2, 0) and 0x2 corresponds to (11, 0).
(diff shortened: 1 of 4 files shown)


device/tt_soc_descriptor.cpp
```diff
@@ -83,10 +83,12 @@ void tt_SocDescriptor::create_coordinate_manager(const BoardType board_type, con
         }
     }
 
+    // TODO: P100 has two types of boards, each using different PCI core.
```
broskoTT, line 86: Create an issue for it

broskoTT (approved):

device/tt_soc_descriptor.cpp
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/773

PR #777, opened 2025-04-23
Add missing API file

afuller-TT:
### Issue
N/A

### Description
This file is required to build, but got missed.

### List of the changes
Add the missing file so it can be included in packages.

### Testing
Preparing tests in TT-Metalium.  This is a pre-req.

### API Changes
None

### DIFF
--- a/device/CMakeLists.txt
+++ b/device/CMakeLists.txt
@@ -115,6 +115,7 @@ if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.23)
                 api/umd/device/types/blackhole_arc.h
                 api/umd/device/types/cluster_descriptor_types.h
                 api/umd/device/types/cluster_types.h
+                api/umd/device/types/harvesting.h
                 api/umd/device/types/tlb.h
                 api/umd/device/types/xy_pair.h
                 api/umd/device/wormhole_coordinate_manager.h


device/CMakeLists.txt
```diff
@@ -115,6 +115,7 @@ if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.23)
                 api/umd/device/types/blackhole_arc.h
                 api/umd/device/types/cluster_descriptor_types.h
                 api/umd/device/types/cluster_types.h
+                api/umd/device/types/harvesting.h
```
broskoTT, line 118: Is there a way to make sure this doesn't happen in the future?

copilot mentions this, but please verify if this is the right idea:

```
# Recursively collect all source files in the "api" directory
file(GLOB_RECURSE API_HEADERS "${CMAKE_CURRENT_SOURCE_DIR}/api/*.h" "${CMAKE_CURRENT_SOURCE_DIR}/api/*.hpp")

if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.23)
    target_sources(
        device
        PUBLIC
            FILE_SET api
            TYPE HEADERS
            BASE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/api
            FILES ${API_HEADERS}
    )
endif()
```

broskoTT (approved):
Thanks for fixing this!

device/CMakeLists.txt
afuller-TT: That would technically work.  However the reason that `GLOB` is discouraged in CMake as a general rule is because they create a continous papercut for all future builds.  CMake goes out of its way to try to stay out of the way.  It wires itself up to get re-run if and only if it needs to.  So build-and-rebuild generally has no overhead.

Once there's a Glob in the mix, CMake can no longer deterministically know that "nothing has changed" unless it re-scans the disk.  Every. Single. Time.  This is a small papercut, but gets repeated on every single compile.

A preferred approach is for something to vet that list.  It'll validate that the list is correct (which a Glob won't do -- a Glob will assume that the contents of `api/` is correct, but that's not guaranteed).  That's exactly what I'm currently working on in TT-Metalium.  So we'll have the protection soon-ish once I get it in place.

An even better approach is for that same thing, but within TT-UMD.  Do you have any "examples" that show how to use TT-UMD as an external consumer?  When I'm done with TT-Metalium I could swing around and wire those up for TT-UMD and that would prove the external API is correct as far as those examples exercise it.

Issue #250, opened 2024-11-04
Nuke tt_device virtual class abstraction


broskoTT:
After #249 we can finally delete tt_device virtual class. The class that somewhat takes its place is tt::umd::Chip.
Clients now use only tt::umd::Cluster.

broskoTT:
Just chatted with @blozano-tt on this, and related to https://github.com/tenstorrent/tt-umd/pull/728
Once tt_device is removed, and we have proper Simulation and Mock chip classes, revisit the fact that metal has to build always with TT_UMD_BUILD_SIMULATION

PR #800, opened 2025-04-27
Remove tt_device

broskoTT:
### Issue
On a path of #250 

### Description
Removes tt_device abstraction

### List of the changes
- Remove tt_device
- Restructure APIs in Cluster with proper comments

### Testing
Client CIs

### API Changes

### DIFF
--- a/device/api/umd/device/cluster.h
+++ b/device/api/umd/device/cluster.h
@@ -592,21 +592,12 @@ public:
      */
     TLBManager* get_tlb_manager(chip_id_t device_id) const;
 
-private:
     /**
-     * If the tlbs are initialized, returns a tuple with the TLB base address and its size
+     * Exposes how TLBs are configured for a specific device.
      */
-    std::optional<std::tuple<uint32_t, uint32_t>> get_tlb_data_from_target(const tt_cxy_pair& target);
-
-    /**
-     * Returns a struct with the TLB configuration, or throws an exception if the target does not have a static TLB.
-     */
-    tlb_configuration get_tlb_configuration(const tt_cxy_pair& target);
-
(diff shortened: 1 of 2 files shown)


device/api/umd/device/cluster.h
```diff
@@ -51,16 +94,54 @@ class tt_device {
...
+     */
+    tt_ClusterDescriptor* get_cluster_description();
+
+    /**
+     * Get set of chip ids for all chips in the cluster.
+     */
+    std::set<chip_id_t> get_target_device_ids();
+
+    /**
+     * Get all logical ids for all local chips targeted by UMD.
```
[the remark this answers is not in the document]
broskoTT: device id = chip id = logical id
!= pci device id

We should probably consolidate all of these.

There are also variables lying around called something_chip and something_device

device/api/umd/device/cluster.h
```diff
@@ -31,14 +31,57 @@ using TLB_DATA = tt::umd::tlb_data;
...
+enum ChipType {
+    SILICON,
+    SIMULATION,
+    MOCK,
+};
+
+// All options when creating a Cluster object.
+// Each of the options provides a default value, so in general any combination of overridden options can be used when
+// constructing Cluster objects. Having this struct saves us from having a lot of different constructor overloads.
+struct ClusterOptions {
```
pjanevskiTT, line 52: seems this PR is not rebased properly at the moment, I'll take a look after the rebase

broskoTT: It is, but git diff is strange.

Probably because I tried to keep the same comments as from the tt_device.

Use the split view (under "Checks " tab there is a cogwheel, select split view, and just view the right part

PR #803, opened 2025-04-28
Introduce send_tensix_risc_reset to Chip

broskoTT:
### Issue
On a path of #250 

### Description
Introduce an API which sends risc_reset signal to all cores on a chip.
Motivation for this change was that before this change, if we tried to use Mock/Simulation chip, the code would still go through the flow of trying to setup ethernet broadcast for broadcasting tensix reset, and then collapse to write_to_device_reg in case broadcast is not enabled. This change makes this happen a bit earlier, at the beginning of the broadcast_tensix_risc_reset_to_cluster function

### List of the changes
- Introduce send_tensix_risc_reset, which sends to all cores.
- Call send_tensix_risc_reset if broadcast is not enabled in broadcast_tensix_risc_reset_to_cluster

### Testing
Client CIs

### API Changes

### DIFF
--- a/device/api/umd/device/chip/chip.h
+++ b/device/api/umd/device/chip/chip.h
@@ -64,6 +64,7 @@ public:
     virtual void dram_membar(const std::unordered_set<uint32_t>& channels = {}) = 0;
 
     virtual void send_tensix_risc_reset(tt_xy_pair core, const TensixSoftResetOptions& soft_resets);
+    virtual void send_tensix_risc_reset(const TensixSoftResetOptions& soft_resets);
 
     virtual int arc_msg(
         uint32_t msg_code,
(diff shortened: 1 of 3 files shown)


device/api/umd/device/chip/chip.h
```diff
@@ -64,6 +64,7 @@ class Chip {
     virtual void dram_membar(const std::unordered_set<uint32_t>& channels = {}) = 0;
```
pjanevskiTT, line 64: I would expect this PR to remove some code from cluster.cpp, is there some reason it's not removing any code?

pjanevskiTT (commented):
I would expect this PR to remove some code from the cluster.cpp

device/api/umd/device/chip/chip.h
broskoTT: So it reduces the code being executed on path of broadcast_tensix_reset. 
But I can't remove the code that happens after the for loop that calls newly introduced send_tensix_risc_reset, since that is a generic broadcast function which could still be called for generic writes, not just tensix reset.
Let me know if this doesn't clear up things

pjanevskiTT: all good

pjanevskiTT (approved):

PR #804, opened 2025-04-28
Simplify simulated_harvesting_mask usage

broskoTT:
### Issue
On a path of #250 

### Description
Simplify simulated_harvesting_mask usage

### List of the changes
- Change some functions from td::unordered_map<chip_id_t, HarvestingMasks>& simulated_harvesting_masks to HarvestingMasks simulated_harvesting_masks, which refer to a single chip.
- get_*_harvesting_mask functions are now too small and bring more confusion than clarity, so I removed them.
- perform_harvesting check is moved to get_harvesting_masks
- Check if harvesting_masks exist for a chip is moved to constructors

### Testing
Client CIs

### API Changes

### DIFF
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -178,7 +178,7 @@ std::unique_ptr<Chip> Cluster::construct_chip_from_cluster(
     chip_id_t chip_id,
     tt_ClusterDescriptor* cluster_desc,
     bool perform_harvesting,
-    std::unordered_map<chip_id_t, HarvestingMasks>& simulated_harvesting_masks,
+    HarvestingMasks& simulated_harvesting_masks,
     int num_host_mem_channels,
     const bool create_mock_chip) {
     HarvestingMasks harvesting_masks =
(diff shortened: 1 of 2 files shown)


device/cluster.cpp
```diff
@@ -232,94 +232,39 @@ void Cluster::add_chip(chip_id_t chip_id, std::unique_ptr<Chip> chip) {
 }
 
 uint32_t Cluster::get_tensix_harvesting_mask(
```
pjanevskiTT, line 234: I think we should also remove this function as well. Are we not removing it just because it has more complex logic than the rest of harvesting masks function?

Additionally, could you open an issue to report harvesting masks in logical layout (0 to N-1) to avoid this shuffle, once we have CEM implementation fully in source?

pjanevskiTT (approved):

device/cluster.cpp
broskoTT: https://github.com/tenstorrent/tt-umd/issues/820

Yeah, once this issue is resolved, the size of this function probably won't warrant its existence anymore . But yeah, for now I like that this shuffling code is here

PR #805, opened 2025-04-28
Implement new constructor accepting ClusterOptions

broskoTT:
### Issue
On a path of #250 

### Description
Introduce new Cluster constructor, which accepts ClusterOptions. Motivation was that our number of constructors was exploding. This one allows changing any one of the options in isolation, or in any combination.

### List of the changes
- Introduce ChipType enum for chip_type argument, so we know which kind of cluster to create.
- Introduce ClusterOptions which should provide options to the constructor
- Changed create_mock_chip usage to go through ChipType
- New construct_soc_descriptor which seems like a better way to separate responsibilities. The second two construct_chip_from_cluster are marked as deprecated (removed in following PRs)
- Implement the new constructor, which should also service Chip and Simulation chips now.
- get_chips_local_first now also verifies chips are present in the cluster.
- target_devices in ClusterOptions are now unordered_set compared to a set in other constructors
- Changed all tests to use the new ClusterOptions constructor. Note that many of the values are default ones, and this is cleaned up in one of the following PRs

### Testing
Client CIs

### API Changes

### DIFF
--- a/device/api/umd/device/cluster.h
+++ b/device/api/umd/device/cluster.h
@@ -19,6 +19,7 @@
 #include "fmt/core.h"
 #include "tt_silicon_driver_common.hpp"
 #include "umd/device/chip/chip.h"
+#include "umd/device/tt_cluster_descriptor.h"
 #include "umd/device/tt_device/tt_device.h"
 #include "umd/device/tt_io.hpp"
 #include "umd/device/types/arch.h"
--- a/device/api/umd/device/cluster.h
+++ b/device/api/umd/device/cluster.h
@@ -545,7 +545,7 @@ class RemoteChip;
 // Chip type to create under the Cluster class.
 //  - Silicon means that the chips under cluster will be connected to actual physical devices connected to the system.
 //  - Simulation is used for simulation runs.
-//  - Mock is used for testing purposes, which mocks all calls.
+//  - Mock is used for testing purposes, implementation of all functions is empty.
 enum ChipType {
     SILICON,
     SIMULATION,
--- a/device/api/umd/device/cluster.h
+++ b/device/api/umd/device/cluster.h
@@ -541,13 +541,57 @@ namespace tt::umd {
 class LocalChip;
 class RemoteChip;
 
+// Chip type to create under the Cluster class.
+//  - Silicon means that the chips under cluster will be connected to actual physical devices connected to the system.
+//  - Simulation is used for simulation runs.
+//  - Mock is used for testing purposes, which mocks all calls.
+enum ChipType {
+    SILICON,
+    SIMULATION,
+    MOCK,
+};
+
+// All options when creating a Cluster object.
+// Each of the options provides a default value, so in general any combination of overridden options can be used when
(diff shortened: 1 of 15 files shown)


device/api/umd/device/cluster.h
```diff
@@ -541,13 +542,57 @@ namespace tt::umd {
 class LocalChip;
 class RemoteChip;
 
+// Chip type to create under the Cluster class.
+//  - Silicon means that the chips under cluster will be connected to actual physical devices connected to the system.
+//  - Simulation is used for simulation runs.
+//  - Mock is used for testing purposes, which mocks all calls.
```
pjanevskiTT, line 548: nit: maybe just add info that mock calls are basically empty implementations, not sure how obvious that is from this comment

tests/api/test_cluster_descriptor.cpp
```diff
@@ -8,6 +8,7 @@
 
 #include "disjoint_set.hpp"
 #include "tests/test_utils/generate_cluster_desc.hpp"
+#include "umd/device/cluster.h"
```
pjanevskiTT, line 11: why is this needed in this file now? Seems there are no other changes

tests/api/test_cluster.cpp
```diff
@@ -59,16 +61,25 @@ TEST(ApiClusterTest, DifferentConstructors) {
...
         umd_cluster = nullptr;
     }
 
     // 4. Constructor taking cluster descriptor based on which to create cluster.
     // Create mock chips is set to true in order to create mock chips for the devices in the cluster descriptor.
     std::filesystem::path cluster_path = tt::umd::Cluster::serialize_to_file();
     std::unordered_map<chip_id_t, HarvestingMasks> simulated_harvesting_masks = {};
-    std::unique_ptr<Cluster> cluster = std::make_unique<Cluster>(
-        tt_ClusterDescriptor::create_from_yaml(cluster_path), 1, true, true, simulated_harvesting_masks);
+    std::unique_ptr<Cluster> cluster = std::make_unique<Cluster>(ClusterOptions{
```
pjanevskiTT, line 75: not needed to address it in this PR, but I would like to see the tests that hit exception paths that you added, for example passing in target devices not present in silicon cluster desc, same for simulation and mock that passes etc...You can maybe open an issue on this

pjanevskiTT (approved):
looks good, like the changes since it will lead to removing all the different constructors

tests/api/test_cluster_descriptor.cpp
broskoTT: sigh....
So ClusterOptions accept now unique_ptr, and gcc is complaining that it needs to have the definition of ClusterDescriptor. So I've added #include cluster_descriptor.h inside cluster.h. And it should be like this, Cluster knows about ClusterDescriptor, not other way around.
But now, cluster_descriptor.h shouldn't have #include cluster.h . And now that that's removed, this specific include "cluster.h" is needed at some places, including in tt_metal

tests/api/test_cluster.cpp
broskoTT: https://github.com/tenstorrent/tt-umd/issues/821

Issue #249, opened 2024-11-04
Move Simulation and Mockup into a Chip instead of Cluster abstraction.


broskoTT:
After #248 , we can finaly move tt_SimulationDevice and tt_MockupDevice into a Chip abstraction.
They would be created through cluster by passing vector of chips to cluster to work with.

broskoTT:
Think about how would this be used from tt::umd::Cluster.
We could create a constructor which accepts a custom set of chips?

PR #807, opened 2025-04-28
Derive SimulationDevice from Chip

broskoTT:
### Issue
On a path of #250 
Solves #249

### Description
Implement SimulationDevice as Chip as oppossed to tt_device.
Simulation should be created from the Cluster from now on.

### List of the changes
- Change : public tt_device to : public Chip
- Introduce simulator_directory option in ClusterOptions and propagate it.
- Change SimulationDevice header so that it matches what's in the Chip.
- Change slightly implementation in SimulationDevice so it matches the new API, the implementation is not actually changed that much.
- Change simulation tests since the API changed.

### Testing
Client CIs

### API Changes

### DIFF
--- a/device/api/umd/device/tt_simulation_device.h
+++ b/device/api/umd/device/tt_simulation_device.h
@@ -10,6 +10,7 @@
 #include <filesystem>
 #include <vector>
 
+#include "umd/device/chip/chip.h"
 #include "umd/device/cluster.h"
 #include "umd/device/tt_simulation_host.hpp"
 
@@ -28,7 +29,7 @@ private:
     tt_SocDescriptor soc_descriptor;
 };
 
-class tt_SimulationDevice : public tt_device {
+class tt_SimulationDevice : public tt::umd::Chip {
 public:
     tt_SimulationDevice(const std::filesystem::path& simulator_directory) :
         tt_SimulationDevice(tt_SimulationDeviceInit(simulator_directory)) {}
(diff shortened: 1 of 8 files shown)


device/cluster.cpp
```diff
@@ -154,13 +155,20 @@ std::unique_ptr<Chip> Cluster::construct_chip_from_cluster(
...
-        // TBD in following PRs
-        throw std::runtime_error("Simulation chip type is not supported yet.");
+#ifdef TT_UMD_BUILD_SIMULATION
+        // Note that passed soc descriptor is ignored in favor of soc descriptor from simulator_directory.
+        return std::make_unique<tt_SimulationDevice>(simulator_directory);
+#else
+        throw std::runtime_error(
+            "Simulation device is not supported in this build. Set '-DTT_UMD_BUILD_SIMULATION=ON' during cmake "
+            "configuration to enable simulation device.");
+#endif
```
broskoTT, line 171: @blozano-tt is this the right way to do it ?

blozano-tt: We can make this work, we just have to make the CMake variable TT_UMD_BUILD_SIMULATION trigger the define of C++ preprocessor define.

broskoTT: I've added that already, this works. The question was whether you approve of this method or is there another suggested way to avoid ifdefs

device/api/umd/device/tt_simulation_device.h
```diff
@@ -28,7 +29,7 @@ class tt_SimulationDeviceInit {
     tt_SocDescriptor soc_descriptor;
 };
 
-class tt_SimulationDevice : public tt_device {
+class tt_SimulationDevice : public tt::umd::Chip {
```
pjanevskiTT, line 32: rename to SimulationChip. I think it would fit our naming conventions around Chip class more than tt_SimulationDevice

pjanevskiTT (approved):
Not gonna block the PR on tt_SimulationDevice rename, would like to see it happen if possible

device/api/umd/device/tt_simulation_device.h
broskoTT: In some following PR, I didn't want to change file names https://github.com/tenstorrent/tt-umd/issues/822

device/cluster.cpp
blozano-tt: Only way to avoid would be to have a common base class, and keep simulation code path in child class? Use polymorphism?

device/simulation/tt_simulation_device.cpp
```diff
@@ -100,79 +98,67 @@ void tt_SimulationDevice::start_device(const tt_device_params& device_params) {
...
     log_info(tt::LogEmulationDriver, "Sending exit signal to remote...");
-    auto builder = create_flatbuffer(DEVICE_COMMAND_EXIT, std::vector<uint32_t>(1, 0), {0, 0, 0}, 0);
+    auto builder = create_flatbuffer(DEVICE_COMMAND_EXIT, std::vector<uint32_t>(1, 0), {0, 0}, 0);
     host.send_to_device(builder.GetBufferPointer(), builder.GetSize());
 }
 
 // Runtime Functions
-void tt_SimulationDevice::write_to_device(
-    const void* mem_ptr, uint32_t size_in_bytes, tt_cxy_pair core, uint64_t addr) {
+void tt_SimulationDevice::write_to_device(tt_xy_pair core, const void* src, uint64_t l1_dest, uint32_t size) {
```
joelsmithTT, line 140: l1_dest suggests that this never interacts with simulated registers, only memory.  Is that true?

broskoTT: Kind of. For touching registers the write_to_device_reg should be used. Although there isn't anything blocking it in this function

device/cluster.cpp
broskoTT: But then you've just moved the ifdef to the construction? Or if classes have the same name, even to CMakeLists?

I wasn't asking necessarily to avoid this, I was just asking if it was okay or clearly the wrong way to do it. I like it how it is now

PR #818, opened 2025-04-30
Ensure fmt is used as a static library

blozano-tt:
### Description
I would like our usage of fmt, to use the static library target to save on compilation time.

### List of the changes
Ensure fmt is built as a static library.
Use `fmt::fmt` target, instead of the header only target.

### Testing
CI is my test bed.

### API Changes

### DIFF
--- a/device/CMakeLists.txt
+++ b/device/CMakeLists.txt
@@ -145,7 +145,7 @@ target_link_libraries(
         hwloc
         rt
         spdlog::spdlog_header_only
-        fmt::fmt-header-only
+        fmt::fmt
         yaml-cpp::yaml-cpp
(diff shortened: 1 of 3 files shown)


broskoTT (approved):

blozano-tt:
Please don't merge this one for now. @afuller-TT and I are having a discussion. This might now be worth the risk.

device/CMakeLists.txt
```diff
@@ -145,7 +145,7 @@ target_link_libraries(
         hwloc
```
broskoTT, line 145: Adding a comment so that merge is not available. See Bryan's comment

PR #819, opened 2025-05-03
Remove spdlog logger

blozano-tt:
### Description
I am working on bringing in a new `tt-logger` based on spdlog.
The one here conflicts with mine, because it redefines some spdlog macros, and uses some hash from spdlog (it is based on a different version of spdlog, whereas I am trying to use the latest release).

This causes me a redefinition error:
https://github.com/tenstorrent/tt-umd/blob/70926004728a7f111687f32ab8bfd6b542b019eb/common/logger_.hpp#L12

### List of the changes
rm -rf the spdlog logger scheme

### Testing
CI is my test bed.

### API Changes


broskoTT (approved):
Thanks for working on this!
 I've requested review from @joelsmithTT as well since he added this originally

joelsmithTT:
Thank you Bryan for picking up the slack on this.  Will your replacement scheme be ready soon?

blozano-tt:
> Thank you Bryan for picking up the slack on this. Will your replacement scheme be ready soon?

I feel it’s ready now. I have an open PR to integrate it into tt-metal. 

I need to get approvals and buy in. (Need to sell it).

There are only two major issues:
1) I need a new repo in Tenstorrent org?
2) I need to check on any performance impact, because I didn’t want to play ifdef games with debug log functions. Ideally we should be able to control logging level at runtime. My current implementation always compiles log_trace and log_debug.

blozano-tt:
@joelsmithTT good to merge this one?

blozano-tt:
@broskoTT any clue why tt-exalens and tt-metal integration builds are failing?

joelsmithTT:
> @joelsmithTT good to merge this one?

Sure, although I use this when I hack on UMD because the existing logger infrastructure is useless to me.  Is your replacement ready?

I believe proliferating small, shared repos linked via submodules/CPM creates an overly complex dependency structure.  Has that ship sailed?

blozano-tt:
> I believe proliferating small, shared repos linked via submodules/CPM creates an overly complex dependency structure. Has that ship sailed?

IMO, that ship has sailed. UMD its self is a submodule / CPM dependency.

While I respect your opinion that submodules and CPM can increase complexity, especially compared to something that is monolithic, they do solve a problem. They allow code reuse across projects without copy paste. 

I personally dislike submodules, as the version of the dependency is implicit in the .git metadata. Also, you have to remember all the extra git submodule commands. If you forget to do `git submodule update --init --recursive` you can end up in a bad state. CPM resolves those issues for me, because everything is explicit in CMake.

One of our company initiatives is to take large problems, and break them down into smaller digestible problems. I think creating smaller component repos, and integrating with CPM is in-line with that ethos.

joelsmithTT:
Is there a technical reason not to integrate your logger into tt-umd?

blozano-tt:
> Is there a technical reason not to integrate your logger into tt-umd?

No, I was planning to start with metal, and in that context I am blocked by the mixed versions of spdlog.

However, I could start with UMD first.

Still waiting on my logger to be flipped public.

blozano-tt:
> > No, I was planning to start with metal, and in that context I am blocked by the mixed versions of spdlog.
> 
> I meant put it into tt-umd instead of a separate repo. Is there a technical reason not to do that?

I want to promote adoption of tt-logger into the 5+ repos that have copy pasted logger.hpp. 

It’s easier to say, here add this logger dependency and you get a common tool with thread safe logging. vs, add this UMD dependency, and use the logger component.

Technically it would be feasible, but in my opinion a strange coupling.

broskoTT:
> any clue why tt-exalens and tt-metal integration builds are failing?

I think you got caught up in my large set of changes. Shouldn't be failing now

Issue #788, opened 2025-04-25
Think about small set of tools for improved debugging


broskoTT:
Maybe a small tool which tries to list all archs, prints all soc descs of silicon systems and cluster descriptor?

broskoTT:
Bumping priority on this.
We're in a need of a small tool which can quickly read telemetry values

PR #828, opened 2025-05-06
UMD topology tool

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/788

### Description
Setting up infrastructure for umd tools.
Also add a smallest tool which generates cluster descriptor.

### List of the changes
- cmake changes to create umd_tools target
- cluster descriptor generator tool
- change CI to build tools

### Testing
Manually executed the tool, also builds in CI

### API Changes


tools/topology.cpp
```diff
@@ -0,0 +1,24 @@
...
+// SPDX-License-Identifier: Apache-2.0
+#include "logger.hpp"
+#include "umd/device/cluster.h"
+#include "umd/device/tt_cluster_descriptor.h"
+
+using namespace tt::umd;
+
+int main(int argc, char *argv[]) {
+    if (argc > 2) {
+        std::cerr << "Usage: topology <cluster_descriptor_path>" << std::endl;
```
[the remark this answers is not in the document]
broskoTT: @blozano-tt what do you folks think about https://github.com/docopt/docopt.cpp ?

tools/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
+add_library(tools_common INTERFACE)
```
afuller-TT, line 1: What is the purpose of breaking out the dependencies into a separate target?  Looks like these are dependencies of tool_topology.  They would go well being listed directly.

tools/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+    INTERFACE
+        umd::device
+        umd::Firmware
+        fmt::fmt-header-only
+)
+target_include_directories(
+    tools_common
+    INTERFACE
+        ${PROJECT_SOURCE_DIR}
+        ${CMAKE_CURRENT_SOURCE_DIR}
```
afuller-TT, line 13: This looks dead.  There are no headers in this dir.

tools/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+        "$<TARGET_PROPERTY:umd::device,INCLUDE_DIRECTORIES>"
```
afuller-TT, line 14: These come automatically when declaring the deps in `target_link_libraries`.  Should not be needed here.

That particular style is actually pulling in _PRIVATE_ incdirs.  That's used for whitebox tests.  It should not appear anywhere else outside the module.

tools/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+
+add_executable(tool_topology topology.cpp)
+target_link_libraries(tool_topology PRIVATE tools_common)
+set_target_properties(
+    tool_topology
+    PROPERTIES
+        RUNTIME_OUTPUT_DIRECTORY
+            ${CMAKE_BINARY_DIR}/tools/umd/
+        OUTPUT_NAME
+            topology
```
afuller-TT, line 25: Any reason for naming the target differently than what you wish the filename to be?

tools/topology.cpp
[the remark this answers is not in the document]
blozano-tt: No experience with it.

I’ve always used https://github.com/jarro2783/cxxopts

I think everyone has their personal favorite for this.

.github/workflows/build-device.yml
```diff
@@ -15,7 +15,7 @@ on:
     branches: ["main"]
 
 env:
-  BUILD_TARGET: device
+  BUILD_TARGET: device umd_tools
```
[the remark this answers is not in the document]
broskoTT: Yes, there is umd_tests... I build that as part of another workflow

tools/CMakeLists.txt
broskoTT: Because I plan making more tools which would have common dependancies, see https://github.com/tenstorrent/tt-umd/pull/832

tools/CMakeLists.txt
broskoTT: Looks like I was missing umd::Common in target link libraries

third_party/CMakeLists.txt
```diff
@@ -122,3 +122,15 @@ CPMAddPackage(
     OPTIONS
         "CMAKE_MESSAGE_LOG_LEVEL NOTICE"
 )
+
+####################################################################################################################
+# cxxopts
+####################################################################################################################
+CPMAddPackage(
+    NAME cxxopts
```
broskoTT, line 130: FYI @blozano-tt I've added this tool, seems lightweight, thanks!

third_party/CMakeLists.txt
```diff
@@ -122,3 +122,15 @@ CPMAddPackage(
     OPTIONS
         "CMAKE_MESSAGE_LOG_LEVEL NOTICE"
 )
+
+####################################################################################################################
+# cxxopts
```
pjanevskiTT, line 127: I am fine with introducing dependency, especially if it's lightweight, but would like to explore if we have any other repo using something else so we can converge on one thing

tools/topology.cpp
```diff
@@ -0,0 +1,33 @@
...
+// SPDX-License-Identifier: Apache-2.0
+#include <cxxopts.hpp>
+
+#include "logger.hpp"
+#include "umd/device/cluster.h"
+#include "umd/device/tt_cluster_descriptor.h"
+
+using namespace tt::umd;
+
+int main(int argc, char *argv[]) {
```
pjanevskiTT, line 12: comment with how to run this tool or simply explanation of all the options would go a long way I feel

.github/workflows/build-device.yml
pjanevskiTT, line 18: as a general comment, I feel tools are going to become "big" part of UMD, same as main code and tests, so I think tools deserve it's own README on what they represent, how to build/run it. It's probably going to be short README, but anyway I think that would be useful and needed

pjanevskiTT (commented):
sorry for a late review, looks good. Left some comments that I think can be addressed for tools going forward

tools/topology.cpp
broskoTT: you run topology --help and you get everything

.github/workflows/build-device.yml
broskoTT: yes readme is missing

third_party/CMakeLists.txt
broskoTT: This was suggested by Bryan, and Ivan suggested docopt, I didn't explore further.

.github/workflows/build-device.yml
broskoTT: Added it to the following PR

PR #829, opened 2025-05-06
Change L2CPU NOC ports

pjanevskiTT:
### Issue

Change L2CPU NOC ports we use in order to have proper mapping of NOC registers

### Description

Change NOC ports we use for L2CPU. We want to keep in router cores all the cores with same NOC register mapping (same address). Previously used L2CPU noc ports have the same register mapping as router cores

### List of the changes

- Change L2CPU noc ports used
- Change the address for base NOC register

### Testing
CI

### API Changes
/


tt-vjovanovic (approved):

tests/soc_descs/blackhole_140_arch_no_eth.yaml
```diff
@@ -55,7 +55,7 @@ security:
 
 l2cpu:
   [
-   8-10, 8-4, 8-8, 8-6
```
broskoTT, line 58: So the original coords were wrong?

device/api/umd/device/blackhole_arc_message_queue.h
```diff
@@ -46,7 +46,7 @@ class BlackholeArcMessageQueue {
 
 public:
     BlackholeArcMessageQueue(
```
broskoTT, line 48: You probably need to rebase first...

broskoTT (approved):

tests/soc_descs/blackhole_140_arch_no_eth.yaml
pjanevskiTT: each L2CPU core has multiple NOC endpoints, similar to DRAM. The previous one had different address for NOC registers. We are switching to these ones in order to have everything under router cores have same NOC regs address. Ideally we would want to support all NOC endpoints for L2CPU cores, but that's work for the future, no real use case for that still...

PR #831, opened 2025-05-06
Fix simulation test

broskoTT:
### Issue
Broken in #763

### Description
Seems like tensix and dram placement is not the same everywhere, so this can break simulator based on soc descriptor.

### List of the changes
- Revert testing values from CoreCoord to tt_xy_pair. The code will now find what is at that place.

### Testing
Manual testing by dzivanovicTT

### API Changes


pjanevskiTT (approved):

dzivanovicTT:
I tested this change with the `LoopbackSingleTensix` test, and the test now passes.

PR #832, opened 2025-05-06
Telemetry tool

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/788
Also, this was directly needed by ihamerTT

### Description
Added telemetry tool. You can filter the pci devices, request telemetry for any tag, and setup frequency of pooling.

### List of the changes
- Add telemetry tool
- Default behavior is that it is reading aiclk, vcore, temp
- If tag is selected it will pool on a tag
- Frequency in ms is adjustable

### Testing
Manually ran the tool to touch all codepaths on a WH card

### API Changes

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -10,6 +10,17 @@ target_link_libraries(
         cxxopts::cxxopts
 )
 
+add_executable(telemetry telemetry.cpp)
+target_link_libraries(telemetry PRIVATE tools_common)
+set_target_properties(
+    telemetry
+    PROPERTIES
+        RUNTIME_OUTPUT_DIRECTORY
+            ${CMAKE_BINARY_DIR}/tools/umd/
+        OUTPUT_NAME
+            telemetry
+)
+
 add_executable(topology topology.cpp)
 target_link_libraries(topology PRIVATE tools_common)
 set_target_properties(
@@ -21,4 +32,9 @@ set_target_properties(
             topology
 )
 
-add_custom_target(umd_tools DEPENDS topology)
+add_custom_target(
+    umd_tools
+    DEPENDS
+        telemetry
+        topology
+)
--- a/tools/README.md
+++ b/tools/README.md
@@ -1,0 +1,33 @@
+# Tools
+
+## Build flow
+
+In general, see the common build instructions in the main [README](../README.md)
+
+Short instructions for building tools:
+```
+cmake -B build -G Ninja
+cmake --build build --target umd_tools
+```
+
+## Topology tool
+
+The topology tool can be used to generate cluster descriptor which describes system topology of tenstorrent devices.
+It shows information such as pci connected chips, remote chips, ethernet connections, harvesting, etc.
+
+You can run the following for more information:
+```
+./build/tools/umd/topology --help
+```
+
+## Telemetry tool
+
+The telemetry tool can be used to read telemetry from ARC. You can provide which pci chips should be polled, the frequency of polling and which telemetry to read.
+It has a special mode where it can read some important factors for Wormhole device.
+
+If you want to save the values, you can also pass an output file to write to.
+
+You can run the following for more information:
+```
+./build/tools/umd/telemetry --help
+```
--- a/tools/telemetry.cpp
+++ b/tools/telemetry.cpp
@@ -17,9 +17,9 @@
 using namespace tt::umd;
 
 std::string run_default_telemetry(int pci_device, ArcTelemetryReader* telemetry_reader) {
-    uint32_t aiclk_fmax = telemetry_reader->read_entry(wormhole::TAG_AICLK);
-    uint32_t aiclk_current = aiclk_fmax & 0xFFFF;
-    // uint32_t aiclk_fmax = aiclk_fmax >> 16;
+    // Holds information about the max AICLK value and the current one.
+    uint32_t aiclk_info = telemetry_reader->read_entry(wormhole::TAG_AICLK);
+    uint32_t aiclk_current = aiclk_info & 0xFFFF;
     uint32_t vcore = telemetry_reader->read_entry(wormhole::TAG_VCORE);
     uint32_t tdp = telemetry_reader->read_entry(wormhole::TAG_TDP) & 0xFFFF;
     uint32_t asic_temperature = telemetry_reader->read_entry(wormhole::TAG_ASIC_TEMPERATURE);
@@ -42,7 +42,8 @@ int main(int argc, char* argv[]) {
         cxxopts::value<std::vector<std::string>>())(
         "t,tag",
         "Telemetry tag to read. If set to -1, will run default telemetry mode which works only for WH and reads aiclk, "
-        "power, temperature and vcore",
+        "power, temperature and vcore. See device/api/umd/device/types/wormhole_telemetry.h (or blackhole_telemetry.h) "
+        "for all available tags.",
         cxxopts::value<int>()->default_value("-1"))(
         "f,freq", "Frequency of polling in microseconds.", cxxopts::value<int>()->default_value("1000"))(
         "o,outfile",
--- a/tools/telemetry.cpp
+++ b/tools/telemetry.cpp
@@ -2,6 +2,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 #include <chrono>
+#include <cxxopts.hpp>
 #include <sstream>
 #include <string>
 #include <thread>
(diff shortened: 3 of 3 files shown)


tools/telemetry.cpp
```diff
@@ -0,0 +1,121 @@
...
+int main(int argc, char* argv[]) {
+    if (argc == 2 && std::string(argv[1]) == "help") {
+        std::cerr << "Usage: telemetry <telemetry_tag> <frequency_ms> <device_ids>" << std::endl;
+        std::cerr << "Example: telemetry 0,1,2,3 4 2" << std::endl;
+        std::cerr << "First argument is the list of device ids. For example 0,1,2,3. This refer to pci_ids"
+                  << std::endl;
+        std::cerr << "Second argument is the telemetry tag value. See all values in "
+                     "device/api/umd/device/types/wormhole_telemetry.h or blackhole_telemetry.h"
+                  << std::endl;
+        std::cerr << "Third argument is the frequency of pooling the telemetry in milliseconds" << std::endl;
```
ihamer-tt, line 56: pooling -> polling

tools/telemetry.cpp
```diff
@@ -0,0 +1,121 @@
...
+        return 1;
+    }
+
+    int frequency_ms = 1;
+    if (argc >= 4) {
+        frequency_ms = std::stoi(argv[3]);
```
ihamer-tt, line 62: Let's not limit this to a millisecond resolution (i.e. make it a float instead of integer)

tools/telemetry.cpp
```diff
@@ -0,0 +1,121 @@
...
+    uint32_t aiclk_current = aiclk_fmax & 0xFFFF;
+    // uint32_t aiclk_fmax = aiclk_fmax >> 16;
+    uint32_t vcore = telemetry_reader->read_entry(wormhole::TAG_VCORE);
+    uint32_t tdp = telemetry_reader->read_entry(wormhole::TAG_TDP) & 0xFFFF;
+    uint32_t asic_temperature = telemetry_reader->read_entry(wormhole::TAG_ASIC_TEMPERATURE);
+    uint32_t current_temperature = (asic_temperature & 0xFFFF) / 16.0;
+
+    log_info(
+        tt::LogSiliconDriver,
+        "Device id {} - AICLK: {} VCore: {} TDP: {} Temp: {}",
```
ihamer-tt, line 27: Rename TDP to Power

tools/CMakeLists.txt
```diff
@@ -13,6 +13,17 @@ target_include_directories(
         "$<TARGET_PROPERTY:umd::device,INCLUDE_DIRECTORIES>"
 )
 
+add_executable(tool_telemetry telemetry.cpp)
+target_link_libraries(tool_telemetry PRIVATE tools_common)
```
ihamer-tt, line 17: Since this only works for Wormhole, I would name it wh-telemetry

tools/telemetry.cpp
```diff
@@ -0,0 +1,121 @@
...
+    while (std::getline(ss, token, delim)) {
+        tokens.push_back(token);
+    }
+
+    return tokens;
+}
+
+int main(int argc, char* argv[]) {
+    if (argc == 2 && std::string(argv[1]) == "help") {
+        std::cerr << "Usage: telemetry <telemetry_tag> <frequency_ms> <device_ids>" << std::endl;
```
ihamer-tt, line 49: This should probably read:
telemetry [telemetry_tag [frequency_ms [device_ids]]]

Also, we should be explicit abou what happnes if no arguments are supplied.

tools/telemetry.cpp
```diff
@@ -0,0 +1,121 @@
...
+        std::cerr << "Example: telemetry 0,1,2,3 4 2" << std::endl;
```
ihamer-tt, line 50: Maybe add "all" to select all visible devices.

tools/telemetry.cpp
```diff
@@ -0,0 +1,121 @@
...
+    std::string token;
+
+    while (std::getline(ss, token, delim)) {
+        tokens.push_back(token);
+    }
+
+    return tokens;
+}
+
+int main(int argc, char* argv[]) {
```
ihamer-tt, line 47: Take a look at https://github.com/docopt/docopt.cpp for handling command line arguments by just writing out a help string, and getting the rest automatically.

tools/telemetry.cpp
broskoTT: I though that @ttmtrajkovic mentioned somewhere that 1ms is the fastest you can get for telemetry?

tools/CMakeLists.txt
broskoTT: Your default telemetry code does work only for wormhole, but the tool in general works for blackhole afaik.
Not sure how to best convey that

tools/telemetry.cpp
ihamer-tt: The update happens every 1ms. If you allow one to sample more often, they will be able to verify that.

tools/CMakeLists.txt
```diff
@@ -13,6 +13,17 @@ target_include_directories(
         "$<TARGET_PROPERTY:umd::device,INCLUDE_DIRECTORIES>"
```
broskoTT, line 13: Test on ubb machine

tools/telemetry.cpp
broskoTT: Arg is now in microseconds

tools/telemetry.cpp
broskoTT: By default if no option is selected, all devices will be used

tools/telemetry.cpp
broskoTT: I ended up using cxxopts

tools/telemetry.cpp
broskoTT: outdated

tools/CMakeLists.txt
broskoTT: Seems like it is working now when printed to outputfile.
I ran ./build/tools/umd/telemetry -o outfile

image

broskoTT:
Tested on ubb, seems like it works per millisecond @ttmtrajkovic 

image

tools/telemetry.cpp
```diff
@@ -0,0 +1,134 @@
...
+#include "logger.hpp"
+#include "umd/device/arc_telemetry_reader.h"
+#include "umd/device/types/wormhole_telemetry.h"
+
+using namespace tt::umd;
+
+std::string run_default_telemetry(int pci_device, ArcTelemetryReader* telemetry_reader) {
+    uint32_t aiclk_fmax = telemetry_reader->read_entry(wormhole::TAG_AICLK);
+    uint32_t aiclk_current = aiclk_fmax & 0xFFFF;
+    // uint32_t aiclk_fmax = aiclk_fmax >> 16;
```
pjanevskiTT, line 22: remove the comment

tools/telemetry.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+#include "fmt/core.h"
+#include "logger.hpp"
+#include "umd/device/arc_telemetry_reader.h"
+#include "umd/device/types/wormhole_telemetry.h"
+
+using namespace tt::umd;
+
+std::string run_default_telemetry(int pci_device, ArcTelemetryReader* telemetry_reader) {
+    uint32_t aiclk_fmax = telemetry_reader->read_entry(wormhole::TAG_AICLK);
```
pjanevskiTT, line 20: I think variable name can be better here. This has both the info on current value and max value. `aiclk_info` maybe

tools/telemetry.cpp
```diff
@@ -0,0 +1,134 @@
...
+    return fmt::format(
+        "Device id {} - AICLK: {} VCore: {} Power: {} Temp: {}",
+        pci_device,
+        aiclk_current,
+        vcore,
+        tdp,
+        current_temperature);
+}
+
+int main(int argc, char* argv[]) {
```
pjanevskiTT, line 36: similar comment as for #828 , I think having comment with example on how to run this tool would be useful

tools/telemetry.cpp
```diff
@@ -0,0 +1,134 @@
...
+        for (int i = 0; i < telemetry_readers.size(); i++) {
+            int device_id = telemetry_readers.at(i).first;
+            auto& telemetry_reader = telemetry_readers.at(i).second;
+
+            std::string telemetry_message;
+            if (telemetry_tag == -1) {
+                telemetry_message = run_default_telemetry(device_id, telemetry_reader.get());
+            } else {
+                uint32_t telemetry_value = telemetry_reader->read_entry(telemetry_tag);
+                telemetry_message = fmt::format("Device id {} - Telemetry value: 0x{:x}", device_id, telemetry_value);
```
pjanevskiTT, line 108: adding telemetry tag print to this message would be useful

tools/telemetry.cpp
```diff
@@ -0,0 +1,134 @@
...
+#include <vector>
+
+#include "fmt/core.h"
+#include "logger.hpp"
+#include "umd/device/arc_telemetry_reader.h"
+#include "umd/device/types/wormhole_telemetry.h"
+
+using namespace tt::umd;
+
+std::string run_default_telemetry(int pci_device, ArcTelemetryReader* telemetry_reader) {
```
pjanevskiTT, line 19: don't really like this. I know this is Hamer's case today, so I am fine with leaving it as is. But in general I think we should expand the argument for tags to contain the array instead of one tag and then completely remove this default telemetry thing. No action for this, just for future consideration

pjanevskiTT (commented):
few smaller comments to address

tools/telemetry.cpp
broskoTT: you can run tool with -h or --help option, which the tool generates automatically

tools/telemetry.cpp
broskoTT: Why are telemetry tag values just int constants and not an enum?

tools/telemetry.cpp
broskoTT: But that wouldn't solve this completely. There is also custom parsing of telemetry values

tools/telemetry.cpp
pjanevskiTT: not sure what was the reasoning behind it

tools/telemetry.cpp
pjanevskiTT: Ok but having README on how to even run the tool with --help option would be nice. That is probably more of a general README rather than for a specific tool

tools/telemetry.cpp
pjanevskiTT: I think we can address this. This applies only to a certain set of telemetry tags. I think for telemetry tags that this does apply to, we can simply have a function that does the custom parsing as needed. For example you can have something like 

```cpp

void print_telemetry_tag(uint32_t telem_tag, uint32_t telem_value) {
    switch(telem_tag) {
        case TELEM_TAG_CUSTOM_PARSING_0:
                // custom parsing 0...
      case TELEM_TAG_CUSTOM_PARSING_1:
              // custom parsing 1....
     default: // al the other telemetry tags
            // just print the value
    }
}

int main() {

 for (int telem_tag : telem_tags) {
     
     int telem_val = telemetry->read_value(telem_tag);

     print_telemetry_tag(telem_tag, telem_val);

 }
}

```

broskoTT: I'll have to leave this for later 
https://github.com/tenstorrent/tt-umd/issues/836

tools/telemetry.cpp
broskoTT: I'll have to leave this for later https://github.com/tenstorrent/tt-umd/issues/836

tools/README.md
```diff
@@ -0,0 +1,33 @@
+# Tools
```
pjanevskiTT, line 1: nice, thanks :)

pjanevskiTT (approved):

Issue #73, opened 2024-09-24
Revisit create-ethernet-map


broskoTT:
Investigate whether it is possible to have the code present in the repo, and build it with umd, rather than having a binary. Motivation is that this would make it easier to flag all the binaries we need to ship to the customer.

pjanevskiTT:
Improved with #245, next step would be to have source code directly inside tt-umd repo

broskoTT:
Already working on this as part of: #560 #481 #459 
However, we currently don't have a plan to move old wormhole code to CEM, but the situation today is much better.

pjanevskiTT:
Main work of having CEM from source is done by #722, follow up issues to guarantee safe introduction to main code path are open, so closing this one

PR #833, opened 2025-05-06
Run our WH topology discovery

broskoTT:
### Issue
Part of #73 , #856

### Description
We already have code which does topology discovery for wormhole systems.
We still don't have wait_arc and wait_eth implemented for WH, so our code doesn't make luwen's CEM completely obsolete.
We'll end up doing double topology discovery, but that is temporary.

### List of the changes
- Call our topology discovery
- Still call CEM

### Testing
Existing CIs
TODO: Run T3K and TG unit tests in metal

### API Changes


pjanevskiTT (approved):
🤞

Issue #751, opened 2025-04-15
Improve Device->Host and Host->Device transfer speed


pjanevskiTT:
Oritginally reported https://github.com/tenstorrent/tt-metal/issues/20518

Using this issue to dump all the numbers that we can currently measure in UMD, I'll split this into multiple issues if needed

Google doc with everything summarized for wider public to read

pjanevskiTT:
TODOs

pjanevskiTT:
Part of the issue with perf result measurements

pjanevskiTT:
Useful internal documents on any topic IO related

pjanevskiTT:
[Memory and iatu layout](https://tenstorrent.sharepoint.com/:p:/r/sites/SystemsEngineering/_layouts/15/Doc.aspx?sourcedoc=%7B6FC9EF98-C96F-486A-A84E-09EA8ADE43E0%7D&file=Memory%20&%20iATU%20Layout.pptx=&action=edit&mobileredirect=true)

pjanevskiTT:
[Presentation summarizing results and further work](https://tenstorrent-my.sharepoint.com/:p:/r/personal/pjanevski_tenstorrent_com/Documents/umd%20io%20perf/UMD%20IO%20performance.pptx?d=wceb470db34e6432b9728994d8b994243&csf=1&web=1&e=KEVaV2)

PR #837, opened 2025-05-08
Add ubench tests for TLB read/write

pjanevskiTT:
### Issue

Part of the work done for #751 

### Description

Add ubench tests for TLB that were written to measure the perf as part of the report for the issue. There are tests for read/writes using both dynamic and static TLB for tensix/DRAM. nanobench infra that was there for measuring is deleted, but there is a still dependency, we might want to use this in the future. I think there is no point in adding additional tests for TLBs, I think these showcase perf for all cases. Since perf of TLB path is not something that can be used in metal's perf path, I think this set of tests is enough for the following period.

### List of the changes

- Add test for dynamic TLB r/w tensix
- Add test for dynamic TLB r/w DRAM
- Add test for static TLB r/w tensix
- Add test for static TLB r/w DRAM

### Testing

Ran tests manually

### API Changes
/


tests/microbenchmark/test_perf_tlb.cpp
```diff
@@ -0,0 +1,227 @@
...
+     double megabytes = static_cast<double>(bytes) / (1024.0 * 1024.0);
+     auto rate = megabytes / seconds;
+     std::cout << direction << ": 0x" << std::hex << bytes << std::dec << " bytes in " << ns << " ns (" << rate
+               << " MB/s)" << std::endl;
+ }
+ 
+ TEST(TestPerf, TLBDynamicDram) {
+     const chip_id_t chip = 0;
+     const uint32_t one_mb = 1 << 20;
+     const size_t NUM_ITERATIONS = 1;
```
broskoTT, line 33: Maybe move all of the constants from all of the tests outside?
CHIP_TO_USE=0, ONE_MIB, NUM_ITERATIONS, one_mb_tlb_window_index, DRAM_TLB_INDEX

Also you could try to unify "sizes" vector

tests/microbenchmark/test_perf_tlb.cpp
```diff
@@ -0,0 +1,227 @@
...
+     const CoreCoord dram_core = CoreCoord(0, 0, CoreType::DRAM, CoordSystem::NOC0);
```
broskoTT, line 34: Can we read this from cluster instead?

tests/microbenchmark/test_perf_tlb.cpp
```diff
@@ -0,0 +1,227 @@
...
+         64 * one_mb,
+         128 * one_mb,
+         256 * one_mb,
+     };
+ 
+     std::unique_ptr<Cluster> cluster = std::make_unique<Cluster>();
+     cluster->start_device(tt_device_params{});
+ 
+     for (uint32_t buf_size : sizes) {
+         CoreCoord core(dram_core.x, dram_core.y, CoreType::DRAM, CoordSystem::NOC0);
```
broskoTT, line 51: Not needed, dram_core is already CoreCoord

tests/microbenchmark/test_perf_tlb.cpp
```diff
@@ -0,0 +1,227 @@
...
+ #include "umd/device/tt_cluster_descriptor.h"
+ #include "umd/device/tt_device/wormhole_tt_device.h"
+ #include "umd/device/wormhole_implementation.h"
+ #include "wormhole/eth_l1_address_map.h"
+ #include "wormhole/host_mem_address_map.h"
+ #include "wormhole/l1_address_map.h"
+ 
+ using namespace tt::umd;
+ 
+ static inline void print_speed(std::string direction, size_t bytes, uint64_t ns) {
```
broskoTT, line 22: Consider printing this in some other format, like some table.
Ideally something that could be directly pasted into a tests/microbenchmark/perf_results.md.
We could even automate this further with a github workflow, which would also pin machine info and commit hash...

tests/microbenchmark/test_perf_tlb.cpp
```diff
@@ -0,0 +1,227 @@
...
+         std::vector<uint8_t> readback(buf_size, 0x0);
+         auto now = std::chrono::steady_clock::now();
+         for (int i = 0; i < NUM_ITERATIONS; i++) {
+             cluster->read_from_device(readback.data(), chip, dram_core, 0x0, readback.size());
+         }
+         auto end = std::chrono::steady_clock::now();
+         auto ns = std::chrono::duration_cast<std::chrono::nanoseconds>(end - now).count();
+         print_speed("Dynamic TLB: Device DRAM -> Host", NUM_ITERATIONS * readback.size(), ns);
+ 
+         EXPECT_EQ(pattern, readback) << "Mismatch for core " << dram_core.str() << " addr=0x0"
```
broskoTT, line 75: These are perf tests for reporting performance.
Let's not conflate it with unit test which guarantee correctness.
I'd say that all of EXPECTs are not needed. They won't guard anything since our other tests should fail first.

tests/microbenchmark/test_perf_tlb.cpp
```diff
@@ -0,0 +1,227 @@
...
+                                      << " size=" << std::dec << readback.size();
+     }
+ }
+ 
+ TEST(TestPerf, TLBDynamicTensix) {
```
broskoTT, line 80: Not sure how this would work along with https://github.com/tenstorrent/tt-umd/pull/747
But that's on me to figure out when I start playing with that task

tests/microbenchmark/test_perf_tlb.cpp
```diff
@@ -0,0 +1,227 @@
...
+ 
+     cluster->configure_tlb(0, tensix_core, one_mb_tlb_window_index, 0x0, TLB_DATA::Relaxed);
+ 
+     const std::vector<uint32_t> sizes = {
+         1 * one_mb,
+     };
+ 
+     for (uint32_t buf_size : sizes) {
+         std::cout << std::endl;
+         std::cout << "Reporting results for buffer size " << (buf_size / one_mb) << " MB" << std::endl;
```
broskoTT, line 140: This should be present in all prints in all tests I think.
Consider adding buf size to "print_speed" function

tests/microbenchmark/test_perf_tlb.cpp
```diff
@@ -0,0 +1,227 @@
...
+             auto now = std::chrono::steady_clock::now();
+             for (int i = 0; i < num_io; i++) {
+                 cluster->write_to_device(pattern.data(), pattern.size(), chip, tensix_core, 0x0);
+             }
+             auto end = std::chrono::steady_clock::now();
+             auto ns = std::chrono::duration_cast<std::chrono::nanoseconds>(end - now).count();
+             print_speed("Static TLB: Host -> Device Tensix L1", num_io * pattern.size(), ns);
+         }
+ 
+         std::vector<uint8_t> readback(one_mb, 0x0);
```
broskoTT, line 157: There is really a lot of repetition in this file.
Consider making a function 'test_write' or something like that, which accepts buf_size, num_iteration, core...
Also test_read function.

If you accept my feedback that EXPECT is not needed, then you could completely decouple write/read, so they could accept "sizes" as well

broskoTT (commented):
Really happy to see this being worked on

tests/microbenchmark/test_perf_tlb.cpp
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/847

tests/microbenchmark/test_perf_tlb.cpp
pjanevskiTT: sizes depends on whether we are doing io to tensix or DRAM, I will leave it per test for now

broskoTT (approved):

PR #838, opened 2025-05-08
Move set_power_state to Chip class

broskoTT:
### Issue
Last change for #250

The previous changes in the stack of changes leading to this one:
https://github.com/tenstorrent/tt-umd/pull/725
https://github.com/tenstorrent/tt-umd/pull/801
https://github.com/tenstorrent/tt-umd/pull/802
https://github.com/tenstorrent/tt-umd/pull/803
https://github.com/tenstorrent/tt-umd/pull/804
https://github.com/tenstorrent/tt-umd/pull/805
https://github.com/tenstorrent/tt-umd/pull/806
https://github.com/tenstorrent/tt-umd/pull/807
https://github.com/tenstorrent/tt-umd/pull/808
https://github.com/tenstorrent/tt-umd/pull/809
https://github.com/tenstorrent/tt-umd/pull/810
https://github.com/tenstorrent/tt-umd/pull/811
https://github.com/tenstorrent/tt-umd/pull/800
https://github.com/tenstorrent/tt-umd/pull/835
https://github.com/tenstorrent/tt-umd/pull/839
https://github.com/tenstorrent/tt-umd/pull/840
https://github.com/tenstorrent/tt-umd/pull/838

### Description
Move set_power_state and all accompanying functions to Chip class

### List of the changes
- Move wait_for_aiclk_value and get_power_state_arc_msg to LocalChip
- Moved implementation for set_power_state to LocalChip and RemoteChip. The implementation was the same, just the chip being targeted was different
- Remove inline from a couple of functions - this wasn't needed, and the build started mysteriously failing due to it.

### Testing
Client CIs

### API Changes
This is the tip of the stack of PRs with breaking API changes.
I'll merge the whole stack once the whole stack is approved, and then merge the changes in the clients
- [x] tt_metal approved PR pointing to this branch: https://github.com/tenstorrent/tt-metal/pull/21269
- [x] tt_debuda approved PR pointing to this branch: https://github.com/tenstorrent/tt-exalens/pull/342
- [x] tt_umd_simulators CI run: https://yyz-gitlab.local.tenstorrent.com/tenstorrent/tt-umd-simulators/-/pipelines/558107

PR #839, opened 2025-05-08
Move several minor functions to Chip class

broskoTT:
### Issue
On a path of #250 

### Description
Move several minor function's implementation to Chip class.
Motivation was that code in cluster.cpp was assuming that local chips had tt_device and sysmem_manager, which is not true for Simulation chip. Moving this to Chip allows for custom implementation.

### List of the changes
- Move get_num_host_channels, get_host_channel_size, get_fast_pcie_static_tlb_write_callable, get_clock, get_numa_node to Chip
-  Removed this from Cluster
- All implementations were very simple, so not much changes, just moving code around.
- Keep LocalChip in Remote to be able to call its functions

### Testing
Client CIs

### API Changes

### DIFF
--- a/device/api/umd/device/chip/chip.h
+++ b/device/api/umd/device/chip/chip.h
@@ -44,6 +44,8 @@ public:
     virtual SysmemManager* get_sysmem_manager();
     virtual TLBManager* get_tlb_manager();
 
+    virtual int get_num_host_channels();
+    virtual int get_host_channel_size(std::uint32_t channel);
     virtual void write_to_sysmem(uint16_t channel, const void* src, uint64_t sysmem_dest, uint32_t size);
     virtual void read_from_sysmem(uint16_t channel, void* dest, uint64_t sysmem_src, uint32_t size);
 
@@ -57,6 +59,8 @@ public:
     virtual void dma_write_to_device(const void* src, size_t size, tt_xy_pair core, uint64_t addr);
     virtual void dma_read_from_device(void* dst, size_t size, tt_xy_pair core, uint64_t addr);
 
+    virtual std::function<void(uint32_t, uint32_t, const uint8_t*)> get_fast_pcie_static_tlb_write_callable();
+
     virtual void wait_for_non_mmio_flush();
 
     virtual void l1_membar(const std::unordered_set<tt::umd::CoreCoord>& cores = {}) = 0;
@@ -65,6 +69,9 @@ public:
 
     virtual void send_tensix_risc_reset(tt_xy_pair core, const TensixSoftResetOptions& soft_resets);
     virtual void send_tensix_risc_reset(const TensixSoftResetOptions& soft_resets);
+
+    virtual int get_clock() = 0;
+    virtual int get_numa_node();
 
     virtual int arc_msg(
         uint32_t msg_code,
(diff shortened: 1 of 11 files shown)


device/api/umd/device/chip/chip.h
```diff
@@ -44,6 +44,8 @@ class Chip {
     virtual SysmemManager* get_sysmem_manager();
     virtual TLBManager* get_tlb_manager();
 
+    virtual int get_num_host_channels();
```
pjanevskiTT, line 47: this is already in SysmemManager. Again this this is the topic of whether which APIs from the classes below chip we want to expose

device/api/umd/device/chip/chip.h
```diff
@@ -66,6 +70,9 @@ class Chip {
     virtual void send_tensix_risc_reset(tt_xy_pair core, const TensixSoftResetOptions& soft_resets);
     virtual void send_tensix_risc_reset(const TensixSoftResetOptions& soft_resets);
 
+    virtual int get_clock() = 0;
```
pjanevskiTT, line 73: already in TTDevice

device/chip/chip.cpp
```diff
@@ -73,6 +73,10 @@ TLBManager* Chip::get_tlb_manager() {
         "Chip::get_tlb_manager is not available for this chip, it is only available for LocalChips.");
 }
 
+int Chip::get_num_host_channels() { return 0; }
+
+int Chip::get_host_channel_size(std::uint32_t channel) { return 0; }
```
pjanevskiTT, line 78: I think this function should throw the exception. I am fine with returning 0 for num_host_channels, but for the size of the channel that doesn't exist, I would like to see the exception being thrown

device/simulation/tt_simulation_device.cpp
```diff
@@ -182,6 +184,8 @@ void tt_SimulationDevice::dram_membar(const std::unordered_set<uint32_t>& channe
 
 void tt_SimulationDevice::dram_membar(const std::unordered_set<tt::umd::CoreCoord>& cores) {}
 
+int tt_SimulationDevice::get_clock() { return 0; }
```
pjanevskiTT, line 187: Can this information be important sometimes? Is it even possible to extract from simulator? Is it fixed? Not questions for you on this PR...mostly wandering whether this value can be important sometimes for any calculations

fyi @dzivanovicTT maybe something you can explore

pjanevskiTT (approved):
Approving this since it's needed for uplifts to simulator to speed that up. In general I think what we need to address is multiple interfaces that basically call same function for underlying classes. So not blocking this PR on it, but I think this should be done sooner rather than later. I left some comments below on examples where we have duplicate APIs

device/api/umd/device/chip/chip.h
broskoTT: As discussed offline, this needs to be virtualized like this, so that it is implemented differently for Simulation Chip

device/api/umd/device/chip/chip.h
broskoTT: As discussed offline, this needs to be virtualized like this, so that it is implemented differently for Simulation Chip

device/simulation/tt_simulation_device.cpp
broskoTT: its a mocked value, i think it is fine

PR #840, opened 2025-05-08
Move deassert_risc_resets to Chip class

broskoTT:
### Issue
On a path of #250 

### Description
Move deassert_risc_resets implementation to Chip class.
Motivation was that code in cluster.cpp was assuming that local chips had tt_device and sysmem_manager, which is not true for Simulation chip. Moving this to Chip allows for custom implementation.
Minor fix for eth_fw_version

### List of the changes
- Move deassert_risc_resets to Chip.
- Implementation for both local and remote chips turned out to be the same, just the remote chip was calling it's local chip function.
- Allow eth_fw_version to be empty. We didn't have a wormhole chip without eth previously, now we have through Simulation

### Testing
Client CIs

### API Changes

### DIFF
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -1019,35 +1019,13 @@ void Cluster::deassert_resets_and_set_power_state() {
     // Assert tensix resets on all chips in cluster
     broadcast_tensix_risc_reset_to_cluster(TENSIX_ASSERT_SOFT_RESET);
 
+    for (auto& [_, chip] : chips_) {
+        chip->deassert_risc_resets();
+    }
+
     // MT Initial BH - ARC messages not supported in Blackhole
     if (arch_name != tt::ARCH::BLACKHOLE) {
-        // Send ARC Messages to deassert RISCV resets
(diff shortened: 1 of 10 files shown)


device/cluster.cpp
```diff
@@ -1064,11 +1042,18 @@ void Cluster::verify_eth_fw() {
             fw_versions.push_back(fw_version);
         }
         verify_sw_fw_versions(chip, SW_VERSION, fw_versions);
-        eth_fw_version = tt_version(fw_versions.at(0));
+        eth_fw_version = fw_versions.empty() ? tt_version() : tt_version(fw_versions.at(0));
```
pjanevskiTT, line 1045: does tt_version have some default values that basically indicate that you are reading invalid tt_version struct. Maybe we can introduce this if it doesn't exist. Not an action for this PR, probably just a follow up issue if something like that doesn't exist

pjanevskiTT (approved):

device/cluster.cpp
broskoTT: Yeah, currently it doesn't

PR #857, opened 2025-05-13
Sysmem(DMA) buffers chapter 1

pjanevskiTT:
### Issue

Part of the work for #751 

### Description

Implement system memory mapping on request. This PR introduces sysmem buffers that are allocated on request. Sysmem buffers have the information needed in order to use it directly as source/destination for DMA transfers. Sysmem buffers are managed by SysmemManager so far. Maybe SysmemManager should become responsible for IO through sysmem as well. That is partially managed by local chip right now. 

PR also introduces tests that show usages of newly introduces sysmem buffers.

This works only on systems where IOMMU is enabled. Follow up issue for adapting this work to hugepages will be opened and addressed in separate PR.

### List of the changes

- Add sysmem buffer
- Add mapping of buffers
- Add management of sysmem buffers in SysmemManager
- Add reading/writing using DMA when src/dst is already mapped
- Change Cluster, Chip and TTDevice code to support all this

### Testing

CI. There are 2 additional tests added for showing off SysmemBuffer class

### API Changes
/

### DIFF
--- a/device/api/umd/device/chip_helpers/sysmem_buffer.h
+++ b/device/api/umd/device/chip_helpers/sysmem_buffer.h
@@ -1,0 +1,33 @@
+/*
+ * SPDX-FileCopyrightText: (c) 2025 Tenstorrent Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+
+#include <stdint.h>
+
+namespace tt::umd {
+
+class SysmemBuffer {
+public:
+    SysmemBuffer(void* buffer_va, uint32_t buffer_size, uint64_t device_io_addr);
+    ~SysmemBuffer() = default;
+
+    void* get_buffer_va() const { return buffer_va; }
+
+    uint32_t get_buffer_size() const { return buffer_size; }
+
+    uint64_t get_device_io_addr() const { return device_io_addr; }
+
+private:
+    // Virtual address in process addr space.
+    void* buffer_va;
+
+    uint32_t buffer_size;
+
+    // Address that is used on the system bus to access the buffer.
+    uint64_t device_io_addr;
+};
+
+}  // namespace tt::umd
--- a/device/api/umd/device/chip_helpers/sysmem_buffer.h
+++ b/device/api/umd/device/chip_helpers/sysmem_buffer.h
@@ -5,26 +5,36 @@
  */
 #pragma once
 
+#include <stddef.h>
 #include <stdint.h>
+
+#include "umd/device/chip_helpers/tlb_manager.h"
+#include "umd/device/tt_xy_pair.h"
 
 namespace tt::umd {
 
 class SysmemBuffer {
 public:
-    SysmemBuffer(void* buffer_va, uint32_t buffer_size, uint64_t device_io_addr);
+    SysmemBuffer(TLBManager* tlb_manager, void* buffer_va, size_t buffer_size, uint64_t device_io_addr);
     ~SysmemBuffer() = default;
 
     void* get_buffer_va() const { return buffer_va; }
 
-    uint32_t get_buffer_size() const { return buffer_size; }
+    size_t get_buffer_size() const { return buffer_size; }
 
-    uint64_t get_device_io_addr() const { return device_io_addr; }
+    uint64_t get_device_io_addr(const size_t offset = 0) const { return device_io_addr + offset; }
+
+    void dma_write_to_device(size_t offset, size_t size, tt_xy_pair core, uint64_t addr);
+
+    void dma_read_from_device(size_t offset, size_t size, tt_xy_pair core, uint64_t addr);
 
 private:
+    TLBManager* tlb_manager_;
+
     // Virtual address in process addr space.
     void* buffer_va;
 
-    uint32_t buffer_size;
+    size_t buffer_size;
 
     // Address that is used on the system bus to access the buffer.
     uint64_t device_io_addr;
--- a/device/chip_helpers/sysmem_manager.cpp
+++ b/device/chip_helpers/sysmem_manager.cpp
@@ -274,11 +274,14 @@ std::shared_ptr<SysmemBuffer> SysmemManager::allocate_sysmem_buffer(uint32_t sys
 }
 
 std::shared_ptr<SysmemBuffer> SysmemManager::allocate_sysmem_buffer(void *buffer, uint32_t sysmem_buffer_size) {
-    // TODO(pjanevski): round down base buffer addr and round up sysmem_buffer_size.
+    uint32_t page_size = sysconf(_SC_PAGESIZE);
+    uint64_t buffer_addr = reinterpret_cast<uint64_t>(buffer);
+
+    if (buffer_addr % page_size != 0 || sysmem_buffer_size % page_size != 0) {
+        throw std::runtime_error("Buffer must be page-aligned with a size that is a multiple of the page size");
+    }
 
     uint64_t device_io_addr = tt_device_->get_pci_device()->map_for_dma(buffer, sysmem_buffer_size);
-
-    uint64_t buffer_addr = reinterpret_cast<uint64_t>(buffer);
 
     std::shared_ptr<SysmemBuffer> sysmem_buffer =
         std::make_shared<SysmemBuffer>(buffer, sysmem_buffer_size, device_io_addr);
--- a/device/chip_helpers/sysmem_manager.cpp
+++ b/device/chip_helpers/sysmem_manager.cpp
@@ -231,7 +231,9 @@ bool SysmemManager::init_iommu(size_t size) {
             strerror(errno));
     }
 
-    uint64_t iova = tt_device_->get_pci_device()->map_for_dma(mapping, map_size);
+    std::shared_ptr<SysmemBuffer> sysmem_buffer = allocate_sysmem_buffer(mapping, size);
+    uint64_t iova = sysmem_buffer->get_device_io_addr();
+
     log_info(LogSiliconDriver, "Mapped sysmem without hugepages to IOVA {:#x}.", iova);
 
     hugepage_mapping_per_channel.resize(num_fake_mem_channels);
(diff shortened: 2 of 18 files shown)


joelsmithTT (commented):
What UMD/Metal have historically referred to as sysmem is accessed via the NOC.  This change introduces the ability for the WH DMA engine to access this sysmem as well, which seems useful.

However, this change also allows the application to allocate and map buffers for access by the DMA engine but _not_ by the NOC, yet still refers to such buffers as "sysmem."  I think this naming convention could be confusing.

For a buffer to be NOC-accessible, WH needs the buffer to either have a 32-bit PA or have an ATU region configured for it.  UMD does the latter, both for 1G pages and for 'fake' hugepages.  Relevant to this is [PR #59](https://github.com/tenstorrent/tt-kmd/pull/59), which augments KMD's API for page pinning/IOMMU mapping with a flag to make the buffer NOC-visible, administering ATU regions as a resource.

Idon't have a concrete suggestion here, other than perhaps to document more precisely what "sysmem" means in different contexts and to clearly distinguish between:

* NOC-accessible host buffer
* DMA engine-accessible host buffer
* NOC and DMA engine-accessible host buffer (and I'm not sure if this last case has ever been tested)

A more abstract observation: this change reuses a UMD pattern that I believe UMD should move away from.  Currently, the UMD implementation tracks state and performs lookups so the application doesn't have to.  This comes at a complexity and performance cost compared to giving the application more direct control over state and resource lifecycle.  For example:

```
void *buffer = ...
size_t size = ...
auto dma_buffer_handle = cluster->get_chip(0)->map_for_dma(buffer, size);
// dma_buffer_handle knows the IOVA, VA, what chip it came from
dma_buffer_handle->dma_write_to_device(device_dst, buffer_src_offset, num_bytes);
dma_buffer_handle.reset(); // nuke the mapping: app is in control of lifecycle
```

I would be remiss not to point out that I believe a similar approach (giving the application more control over lifecycle and state) would also be beneficial for managing TLB windows.

pjanevskiTT:
@joelsmithTT thanks for the detailed review, exactly what I was looking for.

I addressed the comments, both minor and major ones regarding design. I would like you to take another look at it

Some changes I made as an onverview

- Introduced zero_copy API to TTDevice for DMA transfers
- Sysmem buffer now has read/write functions that use this zero copy API
- removed src/dst_mapped_for_dma - that API should be used as fallback path on DMA buffer that we had so far
- Remove UMD management of resources

device/chip_helpers/sysmem_manager.cpp
```diff
@@ -224,7 +228,9 @@ bool SysmemManager::init_iommu(size_t size) {
             strerror(errno));
     }
 
-    uint64_t iova = tt_device_->get_pci_device()->map_for_dma(mapping, size);
+    std::shared_ptr<SysmemBuffer> sysmem_buffer = map_sysmem_buffer(mapping, size);
```
joelsmithTT, line 231: At first glance, this looks concerning because the sysmem_buffer isn't returned or saved.  It's not wrong, but it seems like a more complicated way to do what the code was doing before.

Should instances of SysmemBuffer control the lifetime of the underlying resource?

Also, this says shared_ptr but the map_sysmem_buffer returns unique_ptr.

device/chip_helpers/sysmem_manager.cpp
```diff
@@ -257,4 +263,28 @@ void SysmemManager::print_file_contents(std::string filename, std::string hint)
...
+}
+
+std::unique_ptr<SysmemBuffer> SysmemManager::map_sysmem_buffer(void *buffer, uint32_t sysmem_buffer_size) {
+    TTDevice *tt_device_ = tlb_manager_->get_tt_device();
+    uint32_t page_size = sysconf(_SC_PAGESIZE);
+    uint64_t buffer_addr = reinterpret_cast<uint64_t>(buffer);
+
+    if (buffer_addr % page_size != 0 || sysmem_buffer_size % page_size != 0) {
+        throw std::runtime_error("Buffer must be page-aligned with a size that is a multiple of the page size");
+    }
```
joelsmithTT, line 280: Should this go into the map_for_dma() method?

device/tt_device/wormhole_tt_device.cpp
```diff
@@ -218,11 +214,9 @@ void WormholeTTDevice::dma_d2h(void *dst, uint32_t src, size_t size) {
             throw std::runtime_error("DMA timeout");
         }
     }
-
-    memcpy(dst, dma_buffer.buffer, size);
 }
 
-void WormholeTTDevice::dma_h2d(uint32_t dst, const void *src, size_t size) {
+void WormholeTTDevice::dma_h2d_transfer(uint32_t dst, const void *src, size_t size) {
```
joelsmithTT, line 219: Same comment as above, but dst/src swapped.  Can `src` be uint64_t? and so on.

device/tt_device/wormhole_tt_device.cpp
```diff
@@ -303,4 +295,27 @@ void WormholeTTDevice::dma_h2d(uint32_t dst, const void *src, size_t size) {
...
 }
 
+// TODO: This is a temporary implementation, and ought to be replaced with a
+// driver-based technique that can take advantage of multiple channels and
+// interrupts.  With a driver-based implementation we can also avoid the need to
+// memcpy into/out of a buffer, although exposing zero-copy DMA functionality to
+// the application will require IOMMU support.  One day...
+void WormholeTTDevice::dma_d2h(void *dst, uint32_t src, size_t size) {
+    DmaBuffer &dma_buffer = pci_device_->get_dma_buffer();
+    dma_d2h_transfer((void *)(uintptr_t)dma_buffer.buffer_pa, src, size);
```
joelsmithTT, line 305: I saw you are programming the upper _and_ lower registers for the host-side address now (good!) -- please consider making `buffer_pa` itself 64 bit, and also the the `completion_pa`.  If you do this, you will have to program the upper 32 bits of `completion_pa` into hardware.

Full 64 bit addressing support will be needed when KMD is fixed for WH IOMMU.

joelsmithTT (approved):
Approving in case you want to merge this and continue working in a chapter II branch.

My suggestion at this stage is to consider how the application will want to use this functionality.  If instances of `SysmemBuffer` are something the application will use, then I think it makes sense to implement an RAII-style scheme, something like the following, obviously simplified example:

```
class DmaWrapper
{
  UMD *umd;  // fake, real thing would look different
  uint64_t iova;
  void *buffer;
  size_t size;
public:
  DmaWrapper(UMD *umd, void *buffer, size_t size)
    : umd(umd), iova(umd->map_for_dma(buffer, size)), buffer(buffer), size(size) {}
  
  /* methods for doing DMA */

  ~DmaWrapper() { umd->unmap_for_dma(buffer, size);
};

```

The point is that the lifetime of the instance is tied to the lifetime of the underlying resource (DMA mapping through IOMMU, possibly even the memory if it's not a wrapper around user-supplied memory but performs the allocation itself).

An optimization to consider (which will depend on a re-worked TLB scheme in UMD) is  for such a class to acquire exclusive access to a TLB window when it is constructed.  If multi-channel DMA is implemented, this technique could allow multiple simultaneous DMAs (right now they are serialized for 1 channel and use one TLB window for RX and one for TX).

An abstraction to possibly explore is the generic IO window or IO object idea, with concrete implementations doing DMA, MMIO, or remote IO... all with the same interface.

pjanevskiTT:
Follow up issue for all the comments as I would like to work on chapter 2 to not make this PR blow up and have good baseline to work on https://github.com/tenstorrent/tt-umd/issues/884

device/api/umd/device/chip_helpers/sysmem_buffer.h
```diff
@@ -0,0 +1,43 @@
...
+
+#include <stddef.h>
+#include <stdint.h>
+
+#include "umd/device/chip_helpers/tlb_manager.h"
+#include "umd/device/tt_xy_pair.h"
+
+namespace tt::umd {
+
+class SysmemBuffer {
```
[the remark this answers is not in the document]
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/884

device/chip_helpers/sysmem_manager.cpp
```diff
@@ -257,4 +263,28 @@ void SysmemManager::print_file_contents(std::string filename, std::string hint)
...
+
+    uint64_t device_io_addr = tt_device_->get_pci_device()->map_for_dma(buffer, sysmem_buffer_size);
+
+    std::unique_ptr<SysmemBuffer> sysmem_buffer =
+        std::make_unique<SysmemBuffer>(tlb_manager_, buffer, sysmem_buffer_size, device_io_addr);
+
+    return sysmem_buffer;
+}
+
```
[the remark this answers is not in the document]
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/884

device/chip_helpers/sysmem_manager.cpp
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/884

device/tt_device/wormhole_tt_device.cpp
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/884

device/api/umd/device/chip_helpers/sysmem_buffer.h
```diff
@@ -0,0 +1,43 @@
...
+public:
+    SysmemBuffer(TLBManager* tlb_manager, void* buffer_va, size_t buffer_size, uint64_t device_io_addr);
```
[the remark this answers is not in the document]
pjanevskiTT: intended to be part of the API

device/tt_device/wormhole_tt_device.cpp
```diff
@@ -137,12 +137,7 @@ void WormholeTTDevice::configure_iatu_region(size_t region, uint64_t target, siz
         target);
 }
 
-// TODO: This is a temporary implementation, and ought to be replaced with a
-// driver-based technique that can take advantage of multiple channels and
-// interrupts.  With a driver-based implementation we can also avoid the need to
-// memcpy into/out of a buffer, although exposing zero-copy DMA functionality to
-// the application will require IOMMU support.  One day...
-void WormholeTTDevice::dma_d2h(void *dst, uint32_t src, size_t size) {
+void WormholeTTDevice::dma_d2h_transfer(void *dst, uint32_t src, size_t size) {
```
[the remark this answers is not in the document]
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/884

PR #875, opened 2025-05-14
WH: Don't map top 256MB of 4th memory channel

joelsmithTT:
### Issue
I didn't make an issue.

### Description
Don't pin memory that isn't accessible by hardware anyway.  This is in anticipation of a KMD change that introduces ATU resource management (step 1 for full WH IOMMU support).

### List of the changes
* Rehome `HUGEPAGE_CHANNEL_3_SIZE_LIMIT` constant so it can be used in sysmem_manager.cpp and elsewhere.
* Respect this limit when doing ATU configuration for WH (existing code; unchanged) _and_ pinning pages (added in this change).
* Remove comments that will become obsolete.

### Testing
Manual

### API Changes
There are no API changes in this PR.  The memory in question wasn't accessible to the device prior to this change.


device/pcie/pci_device.cpp
```diff
@@ -480,32 +480,6 @@ uint64_t PCIDevice::map_for_dma(void *buffer, size_t size) {
...
-    // contiguous (to the NOC, using iATU), but it's not clear that this is
-    // worth the effort...  the scheme this is intended to replace supports up
-    // to 4GB which is what application developers want.
-    //
-    // What can we do here?
-    // 1. Use hugepages (part of what we are trying to avoid here).
-    // 2. Use a larger value for the driver's dma_address_bits (currently 32;
-    //    has implications for non-UMD based applications -- basically that any
-    //    DMA buffer mapped beyond the 4GB boundary requires iATU configuration
-    //    for the hardware to be able to reach it via NOC).
```
joelsmithTT, line 499: FYI, this is the technique that I am implementing in KMD.

tests/wormhole/test_cluster_wh.cpp
```diff
@@ -776,10 +776,7 @@ TEST(SiliconDriverWH, RandomSysmemTestWithPcie) {
         uint64_t hi = (lo + ONE_GIG) - 1;
 
         if (channel == 3) {
-            // TODO: I thought everything past 0xffff'dddd was registers or
-            // something, but a) I don't know what's actually there, and b)
-            // the unusable range seems to be bigger than that... so
```
joelsmithTT, line 781: It's PCIE PHY registers, and the reason the unusable range seems bigger is the HUGEPAGE_CHANNEL_3_SIZE_LIMIT thing.  Relevant code doesn't show up in this diff, but exists in local_chip.cpp and wormhole_tt_deivce.cpp.

device/api/umd/device/chip_helpers/sysmem_manager.h
```diff
@@ -10,6 +10,11 @@
 
 namespace tt::umd {
 
+// Don't use the top 256MB of the 4th hugepage region on WH.  Two reasons:
+// 1. There are PCIE PHY registers at the top
+// 2. Provision for platform software to have NOC-accessible host memory
+static constexpr uint32_t HUGEPAGE_CHANNEL_3_SIZE_LIMIT = 805306368;
```
pjanevskiTT, line 16: nit: 768 * (1 << 20) would be a bit more readable imo

pjanevskiTT (approved):

device/api/umd/device/chip_helpers/sysmem_manager.h
joelsmithTT: changed

Issue #850, opened 2025-05-12
UMD tests fail on system with IOMMU enabled


pjanevskiTT:
Running UMD tests (multiple tests) fails because UMD is not unmapping IOMMU buffers that it maps so when we run multiple tests there simply is not enough IOMMU mappings to hand out by KMD so the Cluster construction fails. 

Fix for this is unmapping all the buffers that we previously mapped

joelsmithTT:
Fixed in https://github.com/tenstorrent/tt-umd/pull/880

PR #880, opened 2025-05-15
Unmap DMA buffer at PCIDevice destruction

joelsmithTT:
### Issue
N/A

### Description
Lack of munmap for the DMA buffer kept the device file descriptor alive longer than intended.  During unit tests, this had the effect of preventing resource cleanup between tests.  This manifested as unexpected test failures on WH systems with IOMMU enabled and default (limited) IOVA range.

### List of the changes
* Use named constants for DMA buffer size
* Explicitly unmap the DMA buffer if it exists
* Remove an unused method on the PCIDevice interface

### Testing
CI and Manual

### API Changes
This PR has API changes: `PCIDevice::get_fd()` is removed.  I grepped for this in tt-metal and tt-exalens projects; they do not appear to be using it.

### DIFF
--- a/device/api/umd/device/pci_device.hpp
+++ b/device/api/umd/device/pci_device.hpp
@@ -100,13 +100,6 @@ public:
     int get_numa_node() const { return numa_node; }
 
     /**
-     * @return underlying file descriptor
-     * TODO: this is an abstraction violation to be removed when this class
-     * assumes control over hugepage/DMA mapping code.
-     */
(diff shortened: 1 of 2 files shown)


pjanevskiTT:
Confirming that tests pass locally for me now on the machine with IOMMU setup. So this will close the issue #850. Thanks @joelsmithTT

pjanevskiTT (approved):

Issue #882, opened 2025-05-15
Invalid logging -> there is no device_id


blozano-tt:
https://github.com/tenstorrent/tt-umd/blob/be8b09773cbfdedb7d363fc1a28e34d96fffb840/device/chip/local_chip.cpp#L545

PR #883, opened 2025-05-15
Integrate `tt-logger`

blozano-tt:
### Issue
Closes https://github.com/tenstorrent/tt-umd/issues/882 during integration

### Description
The goal here is to integrate tt-logger a universal logger for Tenstorrent C++ projects.

### List of the changes
- Delete old `logger.hpp`
- Delete interim `logger_.[c/h]pp` that was based on spdlog and associated test
- Update CMake infrastructure to bring in `tt-logger`
- Adjust `assert.hpp` to remove dependence on old `logger.hpp`
- Convert any usage of `log_fatal` and `log_assert` to use the already existing `TT_ASSERT`.
  - These are not supported by `tt-logger`. Related logic should be handled in something like `assert.hpp`.
- Replace any `#include "logger.hpp"` with `#include <tt-logger/tt-logger>`

### Testing
Tested build locally, will use CI for runtime testing.

### API Changes


device/blackhole/blackhole_coordinate_manager.cpp
```diff
@@ -54,7 +54,7 @@ void BlackholeCoordinateManager::assert_coordinate_manager_constructor() {
     // TODO: assert that exactly 2 or all 14 (P100) ETH cores are harvested for Blackhole. This is
     // going to be true both for all Blackhole products.
     // const size_t num_harvested_eth_cores = get_num_harvested(harvesting_masks.eth_harvesting_mask);
-    // log_assert(
+    // TT_ASSERT(
```
broskoTT, line 57: Are there plans to support this usecase through tt-logger?

device/blackhole/blackhole_coordinate_manager.cpp
```diff
@@ -5,7 +5,7 @@
  */
 #include "umd/device/blackhole_coordinate_manager.h"
 
-#include "logger.hpp"
+#include <tt-logger/tt-logger.hpp>
```
broskoTT, line 8: So all log_debug and log_trace have the same definition, and will now use the one from tt-logger?
Nice

device/blackhole/blackhole_coordinate_manager.cpp
```diff
@@ -5,7 +5,7 @@
...
 
 using namespace tt::umd;
```
broskoTT, line 10: Can we add a README entry on how different logger levels are used? I'd really want to test this, I think that previously our log debug and trace might've been broken.
I know this might not be something you want to work on, I can do it instead.
Should I just look at tt-logger's readme for instructions on how to test and document this?

broskoTT (approved):
Thanks for improving umd!

pjanevskiTT (approved):

device/blackhole/blackhole_coordinate_manager.cpp
blozano-tt: Yes, they are just different logging levels.

The default logging level is info.

This reminds me, I need to at least document how to control logging level.

How was it controlled in UMD as of today? (I can look later if you don’t get to this).

Issue #870, opened 2025-05-14
Constrained toplogy


broskoTT:
Once #856 is done, we can figure out how to fit constrained topology discovery in our API well.

Does tt_ClusterTopology allow creating topologies which are constrained, and therefore we drop the assumption that this always represents the whole system?

What happens by default when .target_devices or .target_pci_devices are set?

Do we surface controlling this behavior through ClusterOptions?

pjanevskiTT:
closed by #932

PR #886, opened 2025-05-16
Constrained Cluster Topology

broskoTT:
### Issue
Working on #870 

### Description
So when we set specific target devices, we'd want the returned cluster topology to match what was open.
This makes getting info much easier for our clients.
Somebody can still open a ClusterDescriptor with a full topology, outside of Cluster.

### List of the changes
- Implement create_constrained_cluster_descriptor which copies the right stuff
- Wrote a couple of template filters to help with this.
- Wrote test for it
- Added a new cluster example configuration for easier testing.
- Change Cluster's constructor to constrain its topology if target devices are provided.
- Cluster won't take ownership of ClusterDescriptor anymore, because it only makes sense now.
- Minor change in the way opened devices are printed

### Testing
CI tests.
But will also test this as part of: https://github.com/tenstorrent/tt-metal/pull/22078
And will manually try to run something with a subset of devices

### API Changes
This PR has API changes:


device/cluster.cpp
```diff
@@ -132,12 +132,21 @@ void Cluster::construct_cluster(const uint32_t& num_host_mem_ch_per_mmio_device,
 
     if (chip_type == ChipType::SILICON) {
         std::vector<int> pci_ids;
-        for (const auto& [logical_id, pci_id] : cluster_desc->get_chips_with_mmio()) {
-            pci_ids.push_back(pci_id);
+        auto mmio_id_map = cluster_desc->get_chips_with_mmio();
+        for (chip_id_t local_chip_id : local_chip_ids_) {
```
pjanevskiTT, line 136: I am not quite sure I understand why is this loop needed. Shouldn't `local_chip_ids_` be constructed from `get_chips_with_mmio` ? What is the use case when there is some chip in `local_chip_ids_` not in mmio chips?

device/api/umd/device/cluster.h
```diff
@@ -71,7 +71,7 @@ struct ClusterOptions {
     // If not passed, topology discovery will be ran and tt_ClusterDescriptor will be constructed. If passed, and chip
     // type is SILICON, the constructor will throw if cluster_descriptor configuration shows chips which don't exist on
     // the system.
-    std::unique_ptr<tt_ClusterDescriptor> cluster_descriptor = nullptr;
+    tt_ClusterDescriptor* cluster_descriptor = nullptr;
```
pjanevskiTT, line 74: can you maybe explain why is this change needed? I can't get it from the code

device/cluster.cpp
```diff
@@ -370,16 +379,21 @@ void Cluster::ubb_eth_connections(
...
+    if (temp_full_cluster_desc == nullptr) {
+        temp_full_cluster_desc_ptr = Cluster::create_cluster_descriptor();
+        temp_full_cluster_desc = temp_full_cluster_desc_ptr.get();
     }
 
     std::unordered_set<chip_id_t> chips_to_construct = options.target_devices;
     // If no target devices are passed, obtain them from the cluster descriptor.
     if (chips_to_construct.empty()) {
-        chips_to_construct = cluster_desc->get_all_chips();
+        chips_to_construct = temp_full_cluster_desc->get_all_chips();
```
pjanevskiTT, line 392: if I'm reading this right, if `chips_to_construct` is empty, we don't need to call `create_constrained_cluster_descriptor`. I would rewrite this code as

```cpp
if (chips_to_construct.empty()) {
    chips_to_construct = cluster_desc->get_all_chips();
    chips_to_construct = temp_full_cluster_desc->get_all_chips();
    cluster_desc = temp_full_cluster_desc;
} else {
    cluster_desc = create_constrained_cluster_desc(...);
}
```

tests/api/test_cluster_descriptor.cpp
```diff
@@ -225,3 +226,91 @@ TEST(ApiClusterDescriptorTest, PrintClusterDescriptor) {
     }
     file.close();
 }
+
+TEST(ApiClusterDescriptorTest, ConstrainedTopology) {
```
pjanevskiTT, line 230: nice test. No action

device/cluster.cpp
broskoTT: Happens for mock and simulation chips

device/api/umd/device/cluster.h
broskoTT: Now that another cluster descriptor is likely to be constructed from the provided one, it doesn't make sense to pass the ownership of ClusterDescriptor when creating Cluster. 
I think it maybe didn't make much sense in the first place, since passing ownership is not something you necessarily want. The ClusterDescriptor is only a "data" structure, e.g. it doesn't do much on creation/deletion.

device/cluster.cpp
broskoTT: cluster_desc = temp_full_cluster_desc;
 
But the way code is written currently, object behind temp_full_cluster_desc might get destroyed at the end of the function, because it is maybe being held inside temp_full_cluster_desc_ptr

device/cluster.cpp
pjanevskiTT: but this code looks to be under `chip_type == ChipType::SILICON`

broskoTT: Yeah you're right on not needing the check.
I mean I still need a for loop though, to extract a vector of pci ids to be printed out, as it was done previously. 
I also want to match the order of logical_ids and pci ids, so that's why traversing over local_chip_ids.

pjanevskiTT: local_chip_ids is sorted, if that helps in printout I am fine with leaving it like that. If not, I think previous code is more readable. Your call

pjanevskiTT (approved):
nice changes :)

device/cluster.cpp
broskoTT: Yeah I'd leave it like this due to sorted printout.
I think we should change our interface to use all maps/sets maybe... should be performant for small collections

Issue #730, opened 2025-04-11
Implement TTDevice for remote device


pjanevskiTT:
There is a need for something similar to TTDevice for Remote chips. Basically lower level abstraction that knows how to get basic information about the remote chip (harvesting for example). Some of this code is in topology discovery in #722 , should be moved to separate class

pjanevskiTT:
It is not just arc_msg but also board type and everything else TTDevice has. I'll start moving the functions so maybe I just end up renaming RemoteCommunicaiton

PR #887, opened 2025-05-16
Implement remote TTDevice

pjanevskiTT:
### Issue

Initial work towards #730 

### Description

This PR introduces RemoteTTDevice class. The class is derived from base TTDevice class, in order to match the interface of the class that we use in order to get all the required information. It is not going to have PCIDevice and maybe some other things that WH and BH TTDevice has, so maybe we should think about some other hieararchy in the future. I think this should be fine for now. 

Only basic read/write from/to device functions are implemented as part of the PR, other function will come in separate PRs.

### List of the changes

- Add RemoteTTDevice class
- Override read/write functions from TTDevice 
- Add a test for read/write

### Testing

CI + additional test

### API Changes
/

### DIFF
--- a/device/tt_device/remote_tt_device.cpp
+++ b/device/tt_device/remote_tt_device.cpp
@@ -1,0 +1,70 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+#include "umd/device/tt_device/remote_tt_device.h"
+
+namespace tt::umd {
+
+RemoteTTDevice::RemoteTTDevice(LocalChip *local_chip, eth_coord_t target_chip) :
+    TTDevice(),
+    local_chip_(local_chip),
+    target_chip_(target_chip),
+    remote_communication_(std::make_unique<RemoteCommunication>(local_chip)) {
+    if (local_chip_->get_tt_device()->get_arch() != tt::ARCH::WORMHOLE_B0) {
+        throw std::runtime_error("Creating remote TTDevice is supported only for Wormhole.");
+    }
+}
+
+void RemoteTTDevice::wait_arc_core_start(const tt_xy_pair arc_core, const uint32_t timeout_ms) {}
+
+ChipInfo RemoteTTDevice::get_chip_info() {
+    throw std::runtime_error("get_chip_info() not implemented for RemoteTTDevice.");
+}
+
+uint32_t RemoteTTDevice::get_clock() { throw std::runtime_error("get_clock() not implemented for RemoteTTDevice."); }
+
+uint32_t RemoteTTDevice::get_max_clock_freq() {
+    throw std::runtime_error("get_max_clock_freq() not implemented for RemoteTTDevice.");
+}
+
+uint32_t RemoteTTDevice::get_min_clock_freq() {
+    throw std::runtime_error("get_min_clock_freq() not implemented for RemoteTTDevice.");
+}
+
+BoardType RemoteTTDevice::get_board_type() {
+    throw std::runtime_error("get_board_type() not implemented for RemoteTTDevice.");
+}
+
+std::vector<DramTrainingStatus> RemoteTTDevice::get_dram_training_status() {
+    throw std::runtime_error("get_dram_training_status() not implemented for RemoteTTDevice.");
+}
+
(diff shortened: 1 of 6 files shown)


device/tt_device/remote_tt_device.cpp
```diff
@@ -0,0 +1,70 @@
...
+BoardType RemoteTTDevice::get_board_type() {
+    throw std::runtime_error("get_board_type() not implemented for RemoteTTDevice.");
+}
+
+std::vector<DramTrainingStatus> RemoteTTDevice::get_dram_training_status() {
+    throw std::runtime_error("get_dram_training_status() not implemented for RemoteTTDevice.");
+}
+
+void RemoteTTDevice::read_from_device(void *mem_ptr, tt_xy_pair core, uint64_t addr, uint32_t size) {
+    remote_communication_->read_non_mmio(target_chip_, core, mem_ptr, addr, size);
```
broskoTT, line 43: Why did you end up creating new class, which is a passthrough, rather than just renaming RemoteCommunication to RemoteTTDevice?
(btw I know that remotecommunication doesn't hold eth_coord for the remotechip, but I'm fine if you make it hold it)

What was the choice for the set of functions which currently throw runtime_error, i.e. how do you know which functions from the TTDevice would you need as an interface?

pjanevskiTT: The idea was to derive this from TTDevice in order to have the interface matched, and implement read/write since that is already implemented just in another class. I think we should now work on implementing methods that throw runtime error now, to basically complete the class to have the same things as TTDevice for local devices. We can simply move the code from remote communication to remote tt device in the end. Method for read/writing in TTDevice doesn't have the same declaration as Remote communication, and I wanted to completely match the interface for now.

The code in the end is going to end up in RemoteTTDevice basically.

device/tt_device/tt_device.cpp
```diff
@@ -394,7 +396,11 @@ uint32_t TTDevice::get_min_clock_freq() {
         "Grayskull TTDevice is deleted.");
 }
 
-TTDevice::~TTDevice() { lock_manager.clear_mutex(MutexType::TT_DEVICE_IO, get_pci_device()->get_device_num()); }
+TTDevice::~TTDevice() {
+    if (get_pci_device() != nullptr) {
+        lock_manager.clear_mutex(MutexType::TT_DEVICE_IO, get_pci_device()->get_device_num());
```
broskoTT, line 401: I'm pretty sure clear_mutex doesn't need to be called

device/tt_device/remote_tt_device.cpp
broskoTT: Yeah okay I actually forgot that we already have TTDevice abstraction due to multiple archs :) Wondering if we're going to have to end up having WormholeRemoteTTDevice and BlackholeRemoteTTDevice.
No action

tests/api/test_tt_device.cpp
```diff
@@ -106,3 +108,47 @@ TEST(ApiTTDeviceTest, TTDeviceMultipleThreadsIO) {
...
+    for (chip_id_t remote_chip_id : cluster->get_target_remote_device_ids()) {
+        const CoreCoord tensix_core =
+            cluster->get_chip(remote_chip_id)->get_soc_descriptor().get_cores(CoreType::TENSIX)[0];
+
+        eth_coord_t remote_eth_coord = chip_locations.at(remote_chip_id);
+
+        LocalChip* closest_local_chip =
+            cluster->get_local_chip(cluster_desc->get_closest_mmio_capable_chip(remote_chip_id));
+
+        std::unique_ptr<RemoteTTDevice> remote_tt_device =
```
broskoTT, line 136: So actually, we can now have "get_tt_device" implemented for remote chip as well, right?
You don't have to add it here now, I'm just thinking aloud, no action.

broskoTT (approved):
I think this is going in the right direction if you plan to eventually remove RemoteCommunication

device/tt_device/remote_tt_device.cpp
pjanevskiTT: Probably, main things Remote tt device relies on is remote IO. Either we keep RemoteCommunication and derive that for specific ARCHs, or simply derive RemoteTTDevice further. We'll see how it goes

device/tt_device/tt_device.cpp
pjanevskiTT: Not gonna change it in this PR, this was to guard exception on destructing remote tt device

PR #888, opened 2025-05-16
Add support for logger initialization to control log level

blozano-tt:
### Description
Clients of tt-logger need to define spdlog related details such as what is the default logger? A colored multithread stdout sink, or a non-colored mutithread safe file sink?

To make this easier on clients, I've provided a customizable intializer via `tt-logger/tt-logger-initializer.hpp`.

It is integrated here, via a standalone object library.

The ramifications of this, are that UMD clients can control logging level, and log file behavior using two optional environment variables.

TT_UMD_LOGGER_FILE -> Set to a path on disk, if you wan't to log to a file.
TT_UMD_LOGGER_LEVEL -> Set to one of trace, debug, info, error, critical (If unset, defaults to info)

### List of the changes
- Add object library `logger_init` to define default logger control.

### Testing
Initializer was tested via tt-logger CI

### API Changes


device/logger_init/logger_init.cpp
blozano-tt, line 1: Don't merge this pr as of yet.

I was aiming to embed source file and line number information in log messages, in addition to enabling control of logging level and file sink. 

There is a flaw in the source file and line number collection.

Need to think through a fix.

See: https://github.com/tenstorrent/tt-logger/issues/6#issuecomment-2888098539

blozano-tt: I am decoupling the source file issue from this PR.

device/logger_init/logger_init.cpp
```diff
@@ -0,0 +1,38 @@
...
+ * @brief Implementation of UMD (User Mode Driver) logging initialization
+ *
+ * This file contains the initialization code for the UMD logging system.
+ * It creates a static instance of the TT LoggerInitializer with specific
+ * environment variable names for UMD logging configuration.
+ */
+
+#include <tt-logger/tt-logger-initializer.hpp>
+
+namespace umd_logging {
```
broskoTT, line 16: consider namespace tt::umd::logging

broskoTT (approved):

Issue #861, opened 2025-05-13
Add perf sysmem buffer tests


pjanevskiTT:
Sysmem buffers introduced by #857 

Add perf tests for sysmem buffers that show all the possible usecases with sysmem and report the perf of it

pjanevskiTT:
Closed by #889

PR #889, opened 2025-05-19
Add sysmem buffers perf tests

pjanevskiTT:
### Issue

#861 

### Description

Add tests for perf of reading/writing through sysmem buffers. Perf tests are testing zero copy transfer using DMA, mapping of the buffers, as well as reading/writing when you both have to map the buffer and read/write through it. Tests should represent the examples of how you can use the sysmem buffer.

### List of the changes

- Add test for zero copy transfer to tensix
- Add test for zero copy transfer to DRAM
- Add test for mapping the buffers
- Add test for rw with both mapping and DMA transfer

### Testing
CI

### API Changes
/

### DIFF
--- a/tests/microbenchmark/test_perf_dma.cpp
+++ b/tests/microbenchmark/test_perf_dma.cpp
@@ -3,6 +3,8 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  */
+#include <sys/mman.h>
+
 #include "gtest/gtest.h"
 #include "tests/test_utils/device_test_utils.hpp"
 #include "tests/test_utils/generate_cluster_desc.hpp"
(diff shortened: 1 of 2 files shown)


tests/microbenchmark/test_perf_dma.cpp
```diff
@@ -113,3 +153,121 @@ TEST(TestPerf, DMADram) {
...
+        .num_host_mem_ch_per_mmio_device = 0,
+    });
+
+    const chip_id_t mmio_chip = *cluster->get_target_mmio_device_ids().begin();
+
+    SysmemManager* sysmem_manager = cluster->get_chip(mmio_chip)->get_sysmem_manager();
+
+    const uint32_t one_mb = 1 << 20;
+
+    void* mapping = mmap(nullptr, one_mb, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE | MAP_POPULATE, -1, 0);
```
broskoTT, line 196: why mmap?
Could you just create a simple vector or array with fixed size? That would match our client code better, right?

broskoTT: same below

tests/microbenchmark/test_perf_dma.cpp
```diff
@@ -113,3 +153,121 @@ TEST(TestPerf, DMADram) {
...
+
+    const chip_id_t mmio_chip = *cluster->get_target_mmio_device_ids().begin();
+
+    SysmemManager* sysmem_manager = cluster->get_chip(mmio_chip)->get_sysmem_manager();
+
+    const uint32_t one_mb = 1 << 20;
+
+    const CoreCoord tensix_core = cluster->get_soc_descriptor(mmio_chip).get_cores(CoreType::TENSIX)[0];
+    {
+        void* mapping =
```
broskoTT, line 222: Why is perf_sysmem_read_write bit used in this and previous test?

tests/microbenchmark/test_perf_dma.cpp
pjanevskiTT: so far we support only page-aligned buffers, that is guaranteed by mmap, not by random allocation

tests/microbenchmark/test_perf_dma.cpp
pjanevskiTT: because this test also maps the buffer, doesn't just do read/write

Issue #884, opened 2025-05-16
Sysmem buffers chapter 2 improvements


pjanevskiTT:
Improve sysmem buffers as suggested by Joel in the PR thread https://github.com/tenstorrent/tt-umd/pull/857#pullrequestreview-2844542551

Merging chapter 1 to not blow up the PR and have baseline to work on

pjanevskiTT:
closed by #890

PR #890, opened 2025-05-19
Further improve sysmem(DMA) buffers - chapter 2

pjanevskiTT:
### Issue

#884 

### Description

Chapter 2 of improvements for Sysmem buffers. This PR addresses comments from Joel on the previous PR that should push the design in the right direction. After this, NOC address support from KMD should be integrated into sysmem buffers, so we can make this fully available for everyone to use. 

I kept the API in `SysmemManager` to get the buffers in order to be able to provide only buffer size, or both the mem pointer and buffer size, that simply calls into `SysmemBuffer` constructor. I think external user wouldn't want to have to get TLB manager in order to create `SysmemBuffer`

### List of the changes

- Make src and dst for DMA transfer uint64_t
- Address nit comments
- Make SysmemBuffer control the underlying resource
- Add IOCTL for unmapping the buffers
- Unmap the buffers through SysmemBuffer
- Change tests a bit

### Testing

CI. Manual testing for IOMMU tests

### API Changes
/

### DIFF
--- a/device/api/umd/device/chip_helpers/sysmem_buffer.h
+++ b/device/api/umd/device/chip_helpers/sysmem_buffer.h
@@ -13,10 +13,22 @@
 
 namespace tt::umd {
 
+/**
+ * SysmemBuffer class should represent the resource of the HOST memory that is visible to the device.
+ * Currently, there are two types of sysmem buffers:
+ * 1. Hugepage-based sysmem buffer, that represents old system memory scheme used, that we still want to support until
+ * transition to IOMMU is complete.
+ * 2. Sysmem buffer, that is used when the system is protected by an IOMMU. With IOMMU, the mappings can be requested at
+ * much finer granularity than hugepages.
+ *
+ * Traditionally, we have reffered to the sysmem buffer as something that is
+ * visible to device, has it's own NOC address. Without changes to KMD, this is still not fully supported for IOMMU
+ * buffers.
+ */
 class SysmemBuffer {
 public:
-    SysmemBuffer(TLBManager* tlb_manager, void* buffer_va, size_t buffer_size, uint64_t device_io_addr);
-    ~SysmemBuffer() = default;
+    SysmemBuffer(TLBManager* tlb_manager, void* buffer_va, size_t buffer_size);
+    ~SysmemBuffer();
 
     void* get_buffer_va() const { return buffer_va; }
 
--- a/device/api/umd/device/chip_helpers/sysmem_buffer.h
+++ b/device/api/umd/device/chip_helpers/sysmem_buffer.h
@@ -21,8 +21,8 @@ namespace tt::umd {
  * 2. Sysmem buffer, that is used when the system is protected by an IOMMU. With IOMMU, the mappings can be requested at
  * much finer granularity than hugepages.
  *
- * Traditionally, we have reffered to the sysmem buffer as something that is
- * visible to device, has it's own NOC address. Without changes to KMD, this is still not fully supported for IOMMU
+ * Traditionally, we have referred to the sysmem buffer as something that is
+ * visible to device, has its own NOC address. Without changes to KMD, this is still not fully supported for IOMMU
  * buffers.
  */
 class SysmemBuffer {
--- a/device/chip_helpers/sysmem_manager.cpp
+++ b/device/chip_helpers/sysmem_manager.cpp
@@ -220,6 +220,7 @@ bool SysmemManager::init_iommu(size_t size) {
     const size_t num_fake_mem_channels = size / HUGEPAGE_REGION_SIZE;
 
     TTDevice *tt_device_ = tlb_manager_->get_tt_device();
+    // Caclulate the size of the mapping in order to avoid overlap with PCIE registers.
     size_t map_size =
         (tt_device_->get_arch() == tt::ARCH::WORMHOLE_B0 && num_fake_mem_channels == 4) ? (size - carveout_size) : size;
 
@@ -237,7 +238,7 @@ bool SysmemManager::init_iommu(size_t size) {
             strerror(errno));
     }
 
-    sysmem_buffer_ = map_sysmem_buffer(mapping, size);
+    sysmem_buffer_ = map_sysmem_buffer(mapping, map_size);
     uint64_t iova = sysmem_buffer_->get_device_io_addr();
 
     log_info(LogSiliconDriver, "Mapped sysmem without hugepages to IOVA {:#x}.", iova);
--- a/device/chip_helpers/sysmem_manager.cpp
+++ b/device/chip_helpers/sysmem_manager.cpp
@@ -237,8 +237,8 @@ bool SysmemManager::init_iommu(size_t size) {
             strerror(errno));
     }
 
-    std::shared_ptr<SysmemBuffer> sysmem_buffer = map_sysmem_buffer(mapping, size);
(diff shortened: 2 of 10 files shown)


device/api/umd/device/chip_helpers/sysmem_buffer.h
```diff
@@ -13,10 +13,22 @@
...
+ * Currently, there are two types of sysmem buffers:
+ * 1. Hugepage-based sysmem buffer, that represents old system memory scheme used, that we still want to support until
+ * transition to IOMMU is complete.
+ * 2. Sysmem buffer, that is used when the system is protected by an IOMMU. With IOMMU, the mappings can be requested at
+ * much finer granularity than hugepages.
+ *
+ * Traditionally, we have reffered to the sysmem buffer as something that is
+ * visible to device, has it's own NOC address. Without changes to KMD, this is still not fully supported for IOMMU
+ * buffers.
+ */
```
joelsmithTT, line 27: This is great.  Two nitpicky things: 'we have reffered' -> 'we have referred' and 'has it's own' -> 'has its own'.

For future work - here's some context + info about KMD side of things.

Part of my mental model for 'sysmem' includes outbound ATU programming on the device.  ATU programming enables the buffer to have a NOC address.  KMD now has the ability to do ATU programming at page pinning time, returning both PA (or IOVA) _and_ NOC address to userspace.  Old behavior was just return PA/IOVA.

UMD's code paths for ATU set up are somewhat decoupled from the sysmem/hugepage/page pinning logic:
* Cluster::start_device -> LocalChip::start_device
* LocalChip::start_device -> LocalChip::init_pcie_iatus(), relies on SysmemManager::get_hugepage_mapping
* TTDevice::configure_iatu_region concrete implementations

With newer KMD and migration to full IOMMU support for WH, the init_pcie_iatus and configure_iatu_region logic will no longer be required.  However, map_for_dma will have to change slightly.  I will put together some example code for this.

device/chip_helpers/sysmem_manager.cpp
```diff
@@ -237,8 +237,8 @@ bool SysmemManager::init_iommu(size_t size) {
             strerror(errno));
     }
 
-    std::shared_ptr<SysmemBuffer> sysmem_buffer = map_sysmem_buffer(mapping, size);
-    uint64_t iova = sysmem_buffer->get_device_io_addr();
+    sysmem_buffer_ = map_sysmem_buffer(mapping, size);
```
joelsmithTT, line 240: `size` -> `map_size` - rationale is that when KMD does ATU programming in the same ioctl as page pinning, it will fail if userspace asks to map a hugepage on top of the PCIE PHY registers that live at the top of the address range we use for NOC->system bus.

A size of 4 gigabytes will conflict with this range, hence `map_size` to adjust the mapping size downwards.

Due to ATU configuration the top 256M of the 4th hugepage was never device-accessible anyway.

joelsmithTT, line 240: I think this somewhat obscures the lifetime of `mapping` - it's allocated by this method and deallocated by `~SysmemManager`.  A new engineer reading this code might be tempted to assume the SysmemBuffer instance takes ownership of the allocation.  I don't have a change to suggest here, but perhaps consider documenting lifetime assumptions on the SysmemBuffer class.

`SysmemBuffer` owns the lifetime of a buffer's _mapping for access by the device_, but not:
1. mapping of NOC access to the buffer (this is done by ATU, elsewhere in UMD)
2. lifetime of the underlying buffer itself

RE: 1 -- see my other comment about KMD doing ATU.  When we move to that, PIN_PAGES will couple DMA mapping and mapping for NOC access.

What do you think about moving ownership of the buffer into SysmemBuffer?
If SysmemBuffer doesn't own the memory, is it more of Wrapper around the buffer than a Buffer?

joelsmithTT (approved):
LGTM, the one change I request is the size/map_size thing.  Other comments are just ideas or things to consider.

Also just FYI, I noticed the other day WH DMA was timing out on me, I went back to an older UMD and it started working again.  I never debugged it.  I'll try again after this gets merged, if I run into further issues I'll open a ticket.

device/api/umd/device/chip_helpers/sysmem_buffer.h
broskoTT: Just a nit: you can install any code spell checked in vscode, it would underline these spell issues

tests/api/test_sysmem_manager.cpp
```diff
@@ -73,18 +74,16 @@ TEST(ApiSysmemManager, SysmemBuffersAllocation) {
 
         while (true) {
             try {
-                std::unique_ptr<SysmemBuffer> sysmem_buffer = sysmem_manager->allocate_sysmem_buffer(one_page);
+                sysmem_buffers.push_back(sysmem_manager->allocate_sysmem_buffer(one_mb));
```
broskoTT, line 77: This could be viewed as some kind of perf/stress testing, I don't think it should be part of our regression unit tests...
You can leave it like this now, somebody (probably me) will sweep it up and move it to another set of tests

broskoTT (approved):

device/tt_device/wormhole_tt_device.cpp
```diff
@@ -270,16 +270,16 @@ void WormholeTTDevice::dma_h2d_transfer(uint32_t dst, const void *src, size_t si
...
+    write_reg(DMA_READ_DONE_IMWR_HIGH_OFF, (uint32_t)((dma_buffer.completion_pa >> 32) & 0xFFFFFFFF));
+    write_reg(DMA_READ_CH01_IMWR_DATA_OFF, DMA_COMPLETION_VALUE);  // Read completion value
     write_reg(DMA_READ_ABORT_IMWR_LOW_OFF, 0);
     write_reg(DMA_READ_ABORT_IMWR_HIGH_OFF, 0);
     write_reg(DMA_TRANSFER_SIZE_OFF_RDCH_0, size);
-    uint64_t src_addr = reinterpret_cast<uint64_t>(src);
-    write_reg(DMA_SAR_LOW_OFF_RDCH_0, (uint32_t)(src_addr & 0xFFFFFFFF));
-    write_reg(DMA_SAR_HIGH_OFF_RDCH_0, (uint32_t)((src_addr >> 32) & 0xFFFFFFFF));
+    write_reg(DMA_SAR_LOW_OFF_RDCH_0, (uint32_t)(src & 0xFFFFFFFF));
+    write_reg(DMA_SAR_HIGH_OFF_RDCH_0, (uint32_t)((src >> 32) & 0xFFFFFFFF));
```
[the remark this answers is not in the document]
pjanevskiTT: Seems this is kernel-available macro, not in user space

device/chip_helpers/sysmem_manager.cpp
pjanevskiTT: I would like to address this in a separate PR where I would try to fully adapt SysmemBuffer scheme to already existing hugepages scheme. I agree with this not being fully clear at the moment. I was more focused on adapting SysmemBuffer scheme to IOMMU at the moment. I will probably work in chapter 3 to fully remove init_hugepage, init_iommu and other stuff and then just have the API to map the sysmem buffers. However, that is going to require metal work as well. I'll document the stuff from this thread in the code

pjanevskiTT: used map_size and documented map_size calculation

Issue #892, opened 2025-05-20
Implement new wormhole telemetry


pjanevskiTT:
Implement new wormhole telemetry. It is going to use same scheme as Blackhole telemetry implemented in #493 

Telemetry table addresses are read from NOC_NODEID_X_0 and NOC_NODEID_Y_0

pjanevskiTT:
closed by #893

PR #893, opened 2025-05-20
Implement new Wormhole telemetry

pjanevskiTT:
### Issue

https://github.com/tenstorrent/tt-umd/issues/892

### Description

Implement new Wormhole telemetry. It is going to use the same scheme as Blackhole telemetry. Reading of telemetry address is different than Blackhole. Other things stay the same.

### List of the changes

- Make both WH and BH telemetry use the same scheme
- Implement arch specific reading of telemetry address

### Testing
CI. Tests for telemetry should just pass

### API Changes
/

### DIFF
--- a/device/api/umd/device/arc_telemetry_reader.h
+++ b/device/api/umd/device/arc_telemetry_reader.h
@@ -5,7 +5,10 @@
  */
 #pragma once
 
+#include <map>
+
 #include "umd/device/tt_device/tt_device.h"
+#include "umd/device/tt_xy_pair.h"
 
 namespace tt::umd {
 
@@ -13,14 +16,41 @@ class ArcTelemetryReader {
 public:
     virtual ~ArcTelemetryReader() = default;
 
-    virtual uint32_t read_entry(const uint8_t telemetry_tag) = 0;
+    uint32_t read_entry(const uint8_t telemetry_tag);
 
-    virtual bool is_entry_available(const uint8_t telemetry_tag) = 0;
+    bool is_entry_available(const uint8_t telemetry_tag);
 
     static std::unique_ptr<ArcTelemetryReader> create_arc_telemetry_reader(TTDevice* tt_device);
 
 protected:
     ArcTelemetryReader(TTDevice* tt_device);
+
+    virtual void get_telemetry_address() = 0;
+
+    void initialize_telemetry();
+
+    // Address of the telemetry table struct on ARC core.
+    uint32_t telemetry_table_addr;
+
+    // Number of entries in the telemetry table.
+    uint32_t entry_count;
+
+    // After entry_count the telemetry contains entry_count structs of TelemetryTagEntry.
+    // Each struct contains tag and offset. Tag represents what is represented by the value.
+    // Offset is the index of the telemetry value in the telemetry_values array.
+    struct TelemetryTagEntry {
+        uint16_t tag;
+        uint16_t offset;
+    };
+
+    // Address of the telemetry data on ARC core.
+    uint32_t telemetry_values_addr;
+
+    std::map<uint32_t, uint32_t> telemetry_values;
+    std::map<uint32_t, uint32_t> telemetry_offset;
+
+    // During initialization of telemetry, if the NOC0 is hung then we need to read the telemetry values from NOC1.
+    tt_xy_pair arc_core;
 
     TTDevice* tt_device;
 };
--- a/device/arc_telemetry_reader.cpp
+++ b/device/arc_telemetry_reader.cpp
@@ -23,4 +23,46 @@ std::unique_ptr<ArcTelemetryReader> ArcTelemetryReader::create_arc_telemetry_rea
     }
 }
 
+void ArcTelemetryReader::initialize_telemetry() {
+    tt_device->read_from_device(&entry_count, arc_core, telemetry_table_addr + sizeof(uint32_t), sizeof(uint32_t));
+
+    // We offset the tag_table_address by 2 * sizeof(uint32_t) to skip the first two uint32_t values,
+    // which are version and entry count. For representaiton look at blackhole_telemetry.h
+    uint32_t tag_table_address = telemetry_table_addr + 2 * sizeof(uint32_t);
+    std::vector<TelemetryTagEntry> telemetry_tag_entries(entry_count);
+    tt_device->read_from_device(
+        telemetry_tag_entries.data(), arc_core, tag_table_address, entry_count * sizeof(TelemetryTagEntry));
+
+    std::vector<uint32_t> telemetry_data(entry_count);
+    tt_device->read_from_device(telemetry_data.data(), arc_core, telemetry_values_addr, entry_count * sizeof(uint32_t));
+
+    for (const TelemetryTagEntry& tag_entry : telemetry_tag_entries) {
+        const uint16_t tag_val = tag_entry.tag;
+        const uint16_t offset_val = tag_entry.offset;
+
+        telemetry_values.insert({tag_val, telemetry_data[offset_val]});
+        telemetry_offset.insert({tag_val, offset_val});
+    }
+}
+
+uint32_t ArcTelemetryReader::read_entry(const uint8_t telemetry_tag) {
+    if (!is_entry_available(telemetry_tag)) {
+        throw std::runtime_error(fmt::format(
+            "Telemetry entry {} not available. You can use is_entry_available() to check if the entry is available.",
+            telemetry_tag));
+    }
+
+    const uint32_t offset = telemetry_offset.at(telemetry_tag);
+    uint32_t telemetry_val;
+    tt_device->read_from_device(
+        &telemetry_val, arc_core, telemetry_values_addr + offset * sizeof(uint32_t), sizeof(uint32_t));
+
+    telemetry_values[telemetry_tag] = telemetry_val;
+    return telemetry_values[telemetry_tag];
+}
+
+bool ArcTelemetryReader::is_entry_available(const uint8_t telemetry_tag) {
+    return telemetry_values.find(telemetry_tag) != telemetry_values.end();
+}
+
 }  // namespace tt::umd
--- a/device/api/umd/device/blackhole_arc_telemetry_reader.h
+++ b/device/api/umd/device/blackhole_arc_telemetry_reader.h
@@ -6,9 +6,6 @@
 #pragma once
 
 #include "umd/device/arc_telemetry_reader.h"
-#include "umd/device/blackhole_implementation.h"
-
-extern bool umd_use_noc1;
 
 namespace tt::umd {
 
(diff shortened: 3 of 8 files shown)


device/api/umd/device/arc_telemetry_reader.h
```diff
@@ -5,23 +5,53 @@
...
 #pragma once
 
+#include <map>
+
 #include "umd/device/tt_device/tt_device.h"
+#include "umd/device/tt_xy_pair.h"
 
 namespace tt::umd {
 
 class ArcTelemetryReader {
```
pjanevskiTT, line 15: new telemetry is going to be released with new fw. Test this after new fw is released

device/api/umd/device/arc_telemetry_reader.h
```diff
@@ -5,23 +5,53 @@
  */
 #pragma once
 
+#include <map>
+
 #include "umd/device/tt_device/tt_device.h"
+#include "umd/device/tt_xy_pair.h"
 
 namespace tt::umd {
```
broskoTT, line 14: Are we going to have a case where we also need to support telemetry for old fw as well?
Is it safe just to remove the old code?

pjanevskiTT: after new telemetry is released we are not going to need the code for old telemetry anymore

device/topology_discovery.cpp
```diff
@@ -113,7 +113,7 @@ void TopologyDiscovery::get_pcie_connected_chips() {
         // figuring out ETH addresses from runtime and move it to constants.
         if (chip_id == 0) {
             eth_addresses = TopologyDiscovery::get_eth_addresses(
-                chip->get_tt_device()->get_arc_telemetry_reader()->read_entry(wormhole::TAG_ETH_FW_VERSION));
+                chip->get_tt_device()->get_arc_telemetry_reader()->read_entry(TAG_ETH_FW_VERSION));
```
pjanevskiTT, line 116: test if this still works

nanobind/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@
 # https://github.com/wjakob/nanobind/blob/master/docs/building.rst
 # Use Development.Module to support manylinux
 # Python module library should not dynamically link libpython.so or use Development.Embed
-find_package(Python REQUIRED COMPONENTS Development.Module)
+find_package(Python REQUIRED COMPONENTS Development)
```
pjanevskiTT, line 5: should we keep this change? I made it to be able to build in my custom bare metal enviorenment

broskoTT: @blozano-tt

device/arc_telemetry_reader.cpp
```diff
@@ -23,4 +23,49 @@ std::unique_ptr<ArcTelemetryReader> ArcTelemetryReader::create_arc_telemetry_rea
...
+    std::vector<TelemetryTagEntry> telemetry_tag_entries(entry_count);
+    tt_device->read_from_device(
+        telemetry_tag_entries.data(), arc_core, tag_table_address, entry_count * sizeof(TelemetryTagEntry));
+
+    std::vector<uint32_t> telemetry_data(entry_count);
+    tt_device->read_from_device(telemetry_data.data(), arc_core, telemetry_values_addr, entry_count * sizeof(uint32_t));
+
+    for (uint32_t i = 0; i < entry_count; ++i) {
+        uint32_t tag_offset;
+        tt_device->read_from_device(&tag_offset, arc_core, telemetry_table_addr + 8 + 4 * i, sizeof(uint32_t));
```
broskoTT, line 41: Comment on the 8 and 4*i values

broskoTT (approved):
Nice!

device/api/umd/device/arc/arc_telemetry_reader.h
```diff
@@ -5,24 +5,76 @@
...
+    std::map<uint32_t, uint32_t> telemetry_values;
+    std::map<uint32_t, uint32_t> telemetry_offset;
+
+    // During initialization of telemetry, if the NOC0 is hung then we need to read the telemetry values from NOC1.
+    tt_xy_pair arc_core;
+
     TTDevice* tt_device;
+
+private:
+    const std::unordered_set<uint16_t> static_entries{
```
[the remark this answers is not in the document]
broskoTT: These are cached entries, as in, they are read once?

device/api/umd/device/firmware/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+    - For each older firmware version, a class is derived from FirmwareInfoProvider.
+
+    - These derived classes override only the functions whose behavior differs from the latest firmware.
+
+    - This minimizes code duplication and isolates version-specific logic.
+
+### Version Selection
+
+The correct implementation is chosen at runtime based on the device’s firmware version.
```
broskoTT, line 25:
```suggestion
The correct implementation is chosen at runtime based on the device’s firmware version and architecture.
```

device/api/umd/device/firmware/wormhole_18_4_firmware_info_provider.h
```diff
@@ -0,0 +1,34 @@
...
+
+namespace tt::umd {
+
+/* This class captures Wormhole firmware versions 18.4.0 and later.
+ * In this firmware release there was not ASIC id information available,
+ * as well as maximum possible AICLK on the device. So these functions return
+ * placeholder values in this class.
+ * Release: https://github.com/tenstorrent/tt-firmware/releases/tag/v18.4.0
+ */
+class Wormhole_18_4_FirmwareInfoProvider : public WormholeLegacyFirmwareInfoProvider {
```
broskoTT, line 21: Hmmm... I would expect that if you want to chain derived classes, that the main base one would be FirmwareInfoProvider, then the next derived would be the next newest, then the last in the chain would we the oldest one. If I got it right

WormholeLegacyFirmwareInfoProvider = fw < 18.4
Wormhole_18_4_FirmwareInfoProvider = 18.4 <= fw < 18.7
FirmwareInfoProvider = 18.7 <= fw

Then I'd expect for WormholeLegacyFirmwareInfoProvider to derive from Wormhole_18_4_FirmwareInfoProvider, not vice versa?

This comment is given in context that you'd introduce further the changes between pre 18.7 and post 18.7

device/arc/wormhole_arc_telemetry_reader.cpp
```diff
@@ -5,58 +5,41 @@
...
 
-    uint32_t telemetry_value;
+    uint32_t telemetry_values_addr_offset;
     tt_device->read_from_device(
-        &telemetry_value, arc_core, telemetry_base_noc_addr + telemetry_tag * sizeof(uint32_t), sizeof(uint32_t));
-
-    return telemetry_value;
-}
+        &telemetry_values_addr_offset,
+        arc_core,
```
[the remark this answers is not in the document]
broskoTT: Since it is in a follow up issue: Also add that keeping the arc core is in the structures is probably not needed anymore, and then not even specialized constructors are needed

device/firmware/firmware_info_provider.cpp
```diff
@@ -0,0 +1,81 @@
...
+        case ARCH::BLACKHOLE:
+            return std::make_unique<FirmwareInfoProvider>(tt_device);
+        default:
+            throw std::runtime_error("Unsupported architecture for firmware versioner.");
+    }
+}
+
+semver_t FirmwareInfoProvider::get_firmware_version() { return firmware_version; }
+
+semver_t FirmwareInfoProvider::get_minimum_compatible_firmware_version() {
```
broskoTT, line 38: Why was this added? I don't see it used anywhere? Was it because you think it is a good idea to expose it?
In that case, we should probably return this per derived class as well? Along with the minimum compatible fw, could be changed to a pair<semver_t,semver_t>

device/tt_device/tt_device.cpp
```diff
@@ -542,4 +540,28 @@ uint64_t TTDevice::get_refclk_counter() {
...
+    std::vector<DramTrainingStatus> dram_training_status;
+    const uint32_t num_dram_channels = architecture_impl_->get_dram_banks_number();
+    // Format of the dram training status is as follows:
+    // Each channel gets two bits in the 32-bit value (16 bits used). The lower bits are for lower channels.
+    // Lower of the two bits is for training error and higher of the two bits is for training status.
+    // Example: 0b 00 00 00 00 00 00 01 10
+    // would mean that only channel 0 is trained, channel 1 has the error and other are not trained and don't have
+    // errors. If some channel is harvested the bits are always going to be zero.
+    for (uint32_t dram_channel = 0; dram_channel < num_dram_channels; dram_channel++) {
+        dram_training_status.push_back(get_firmware_info_provider()->get_dram_training_status(dram_channel));
```
broskoTT, line 559: Consider if it would be better to return a full vector of DamTrainingStatus directly from firmware info provider, rather than filling it up here

device/firmware/firmware_info_provider.cpp
pjanevskiTT: changed it to be static function based on arch

device/api/umd/device/arc/arc_telemetry_reader.h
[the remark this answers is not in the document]
pjanevskiTT: these are used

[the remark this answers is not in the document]
pjanevskiTT: yes read only once

device/tt_device/wormhole_tt_device.cpp
```diff
@@ -109,45 +109,12 @@ uint32_t WormholeTTDevice::get_clock() {
 }
 
 uint32_t WormholeTTDevice::get_max_clock_freq() {
-    uint32_t aiclk_telemetry = telemetry->read_entry(tt::umd::wormhole::TAG_AICLK);
-    return (aiclk_telemetry >> 16) & 0xFFFF;
+    // TODO: figure out if this exists in new telemetry.
```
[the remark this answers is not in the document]
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/1234

device/tt_device/tt_device.cpp
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/1234

device/CMakeLists.txt
```diff
@@ -52,13 +52,18 @@ target_sources(
         tt_device/wormhole_tt_device.cpp
```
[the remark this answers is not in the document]
pjanevskiTT: works

device/api/umd/device/firmware/wormhole_18_4_firmware_info_provider.h
broskoTT: Sure, but now you have overridden implementation which doesn't make sense, right?
I'm fine if you're touching this in the next PR

PR #897, opened 2025-05-20
Add source location logging

blozano-tt:
### Issue
NA

### Description
`tt-logger` did not have support for source location logging. Now it does.

### List of the changes
- Switch to latest version of `tt-logger`
- UMD selects its own pattern to use, which displays file:line information in the log messages.

Log messages look like this now:
```
[2025-05-20 18:15:49.113] [info] [cluster.cpp:147] [SiliconDriver] Opening local chip ids/pci ids: {0}/[0] and remote chip ids {}
```

### Testing
CI

### API Changes


pjanevskiTT (approved):
I'm generally fine with this change, doesn't bother me. I would wait for other folks to say whether they think we should have source location in the log so just we can have common agreement

blozano-tt:
> I'm generally fine with this change, doesn't bother me. I would wait for other folks to say whether they think we should have source location in the log so just we can have common agreement

I am hoping everyone is cool with it.

I want to avoid adding an ENV to control it. But if thats what people want, then its possible.

joelsmithTT:
>  I would wait for other folks to say whether they think we should have source location in the log

I want it, I find it useful during development.  "[SiliconDriver]" does not communicate any useful information to me, but "pci_device.cpp:123" does.

broskoTT (approved):
Thanks for continued improvement!

PR #906, opened 2025-05-21
Use KMD PIN_PAGES for configuring NOC access to host memory

joelsmithTT:
This is example/reference code not intended to be merged in this PR, demonstrating:
* detecting whether KMD is new enough for new APIs
* using KMD for sysmem NOC mapping

RE: KMD API detection -- the forthcoming KMD release will bump its API version number to 2.  The suggestion is for UMD to fail with an actionable error message ("Please update tt-kmd") if an older version is detected.

RE: iATU setup -- the updated PIN_PAGES API is part of the effort to enable IOMMU for WH.  More details [here](https://tenstorrent.atlassian.net/wiki/spaces/syseng/pages/903544840/32-bit+DMA+migration+to+IOMMU).  Impact to UMD:
* UMD will no longer need architecture-specific code paths for iATU setup
* KMD iATU management allows tooling/application SW coexistence based on an established but incompletely implemented convention (256 MB carveout at the top of the sysmem address space)

### DIFF
--- a/device/pcie/pci_device.cpp
+++ b/device/pcie/pci_device.cpp
@@ -183,12 +183,23 @@ PCIDevice::PCIDevice(int pci_device_number) :
         TT_THROW("Running with IOMMU support requires KMD version {} or newer", kmd_ver_for_iommu.to_string());
     }
 
+    tenstorrent_get_driver_info driver_info{};
+    driver_info.in.output_size_bytes = sizeof(driver_info.out);
+    if (ioctl(pci_device_file_desc, TENSTORRENT_IOCTL_GET_DRIVER_INFO, &driver_info) == -1) {
+        TT_THROW("TENSTORRENT_IOCTL_GET_DRIVER_INFO failed");
+    }
+
     log_info(
         LogSiliconDriver,
-        "Opened PCI device {}; KMD version: {}, IOMMU: {}",
+        "Opened PCI device {}; KMD version: {}; API: {}; IOMMU: {}",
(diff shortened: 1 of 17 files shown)


device/pcie/pci_device.cpp
```diff
@@ -470,6 +481,71 @@ uint64_t PCIDevice::map_for_hugepage(void *buffer, size_t size) {
...
+
+    if (vaddr % page_size != 0 || size % page_size != 0) {
+        TT_THROW("Buffer must be page-aligned with a size that is a multiple of the page size");
+    }
+
+    if (size > page_size && !is_iommu_enabled()) {
+        TT_THROW("Cannot map buffer of size {} to NOC with IOMMU disabled", size);
+    }
+
+    struct
```
pjanevskiTT, line 496: why not have this in ioctl.h?

pjanevskiTT (approved):
Love the changes. Obviously we are going to split this in multiple PRs down the line, leaving approval just as a sign of thinking this is good and should be implemented.
Left some comments that you don't need to address, we can address it in PRs that are going to follow up 
Specifically on sysmem buffers, I think NOC addr should become non-optional, other things are great.

pjanevskiTT:
Issues following this work #912 #913. @joelsmithTT feel free to assign yourself to these if you want to properly implement it. If not we can split it. 
I think we should split this PR into two following 2 issues above

pjanevskiTT:
closing it since work was split by #947 and #946

PR #907, opened 2025-05-22
Add set_level api for clients to control UMD log level

blozano-tt:
### Issue
Requested by @tt-vjovanovic over slack.

### Description
Clients of UMD may want to adjust log level without resorting to an ENV.
This PR adds a set_level API.

### List of the changes
- Add new public API header `<umd/logging/config.h>`
  - Currently only exposes one function and associated level enum
- Added a corresponding `logging/config.cpp` for implementation, also absorbed the init code.
- Removed the object library that was used previously for simplicity

### Testing
CI is my test bed

### API Changes
This PR has API changes:
- Since this is a newly introduced API in a new namespace, it can't possible break anyone.

### DIFF
--- a/device/api/umd/device/logging/config.h
+++ b/device/api/umd/device/logging/config.h
@@ -1,0 +1,40 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+/**
+ * @brief UMD logger configuration and utilities
+ *
+ * This namespace contains functionality for configuring the logging system
+ * used by the UMD (User Mode Driver).
+ */
+namespace tt::umd::logging {
+
+/**
+ * @brief Logging severity levels
+ *
+ * Defines the different severity levels for logging messages, from most
+ * verbose (trace) to most severe (critical), with an option to disable
+ * logging completely (off).
+ */
+enum class level {
+    trace,     ///< Most detailed logging level, for tracing program execution
+    debug,     ///< Debugging information, useful during development
+    info,      ///< General informational messages about program operation
+    warn,      ///< Warning messages for potentially harmful situations
+    error,     ///< Error messages for serious problems
+    critical,  ///< Critical errors that may lead to program termination
+    off        ///< Disables all logging
+};
+
+/**
+ * @brief Sets the global logging level
+ *
+ * @param lvl The new logging level to set. Messages with severity levels
+ *            lower than this level will not be logged.
+ */
+void set_level(level lvl);
+
+}  // namespace tt::umd::logging
--- a/device/api/umd/device/logging/config.h
+++ b/device/api/umd/device/logging/config.h
@@ -5,7 +5,7 @@
 #pragma once
 
 /**
- * @brief UMD logger configuration and utilities
+ * @brief UMD logger configuration and utilities.
  *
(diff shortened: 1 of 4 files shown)


device/logging/config.cpp
```diff
@@ -33,6 +36,27 @@ constexpr auto log_pattern = "[%Y-%m-%d %H:%M:%S.%e] [%l] [%s:%#] %v";
  */
 static tt::LoggerInitializer loggerInitializer(file_env_var, level_env_var, log_pattern);
 
-// Note - Using default logger pattern, until source_location info can be fixed
+/// Map our internal enum to spdlog's level enum
+spdlog::level::level_enum to_spdlog_level(level lvl) {
```
broskoTT, line 40: I assume you've done this to avoid surfacing spdlog publicly through umd?

device/api/umd/device/logging/config.h
```diff
@@ -0,0 +1,40 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+/**
+ * @brief UMD logger configuration and utilities
```
pjanevskiTT, line 8: nit: add . to the end of all comment in this file. 

Nice explanations in the comments :)

device/api/umd/device/logging/config.h
```diff
@@ -0,0 +1,40 @@
...
+namespace tt::umd::logging {
+
+/**
+ * @brief Logging severity levels
+ *
+ * Defines the different severity levels for logging messages, from most
+ * verbose (trace) to most severe (critical), with an option to disable
+ * logging completely (off).
+ */
+enum class level {
```
pjanevskiTT, line 22: I think renaming this to LogLevel and having it inside tt::umd namespace is a bit more understandable to me. But I'm leaving this to you to choose, I don't think it's that important to change

device/logging/config.cpp
```diff
@@ -3,14 +3,17 @@
...
+ * @file config.cpp
  * @brief Implementation of UMD (User Mode Driver) logging initialization
  *
  * This file contains the initialization code for the UMD logging system.
  * It creates a static instance of the TT LoggerInitializer with specific
  * environment variable names for UMD logging configuration.
  */
 
+#include <spdlog/spdlog.h>
+#include <umd/device/logging/config.h>
```
pjanevskiTT, line 15: We don't have strict set of rules on this one, but so far we have included our own files like

`#include "umd_file.h"`

"" instead of <>

device/logging/config.cpp
```diff
@@ -33,6 +36,27 @@ constexpr auto log_pattern = "[%Y-%m-%d %H:%M:%S.%e] [%l] [%s:%#] %v";
  */
 static tt::LoggerInitializer loggerInitializer(file_env_var, level_env_var, log_pattern);
 
-// Note - Using default logger pattern, until source_location info can be fixed
+/// Map our internal enum to spdlog's level enum
```
pjanevskiTT, line 39: nit: add . to the end of comment

pjanevskiTT (approved):
this is great!

device/logging/config.cpp
blozano-tt, line 15:
```suggestion
#include "umd/device/logging/config.h"
```

Issue #856, opened 2025-05-13
Remove CEM dependancy


broskoTT:
The point of this Milestone is to remove dependency over CEM by switching to UMD's topology discovery + UMD's wormhole wait on arc/dram/eth to be ready.

broskoTT:
This task only tracks now removing the CEM from our repo when this is ready. And switching topology discovery to our code.

pjanevskiTT:
closed by #902

PR #915, opened 2025-05-23
Fixes to topology discovery

broskoTT:
### Issue
Needed for https://github.com/tenstorrent/tt-umd/issues/856

### Description
Some fixes that I saw needed for topology discovery to work as intended

### List of the changes
- Filling up idle and active eth channels
- Filling up coords_to_chip_ids
- Filling up ethernet_connections in both ways
- Fixed an issue with the check (note the chip_0 and chip_1): 
if (desc.ethernet_connections[chip_1].find(channel_1) != desc.ethernet_connections[chip_0].end())
- Also refactored a bit not to use square brackets but explicitly .at or .insert

### Testing
Existing CI tests should be enough

### API Changes

### DIFF
--- a/device/topology_discovery.cpp
+++ b/device/topology_discovery.cpp
@@ -453,12 +453,27 @@ void TopologyDiscovery::fill_cluster_descriptor_info() {
             {chip_id, chip->get_chip_info().harvesting_masks.dram_harvesting_mask});
         cluster_desc->eth_harvesting_masks.insert(
             {chip_id, chip->get_chip_info().harvesting_masks.eth_harvesting_mask});
-        cluster_desc->chip_locations.insert({chip_id, eth_coords.at(chip_id)});
+        eth_coord_t eth_coord = eth_coords.at(chip_id);
+        cluster_desc->chip_locations.insert({chip_id, eth_coord});
+        cluster_desc->coords_to_chip_ids[eth_coord.rack][eth_coord.shelf][eth_coord.y][eth_coord.x] = chip_id;
+
+        int num_channels =
+            tt::umd::architecture_implementation::create(tt::ARCH::WORMHOLE_B0)->get_num_eth_channels() -
+            CoordinateManager::get_num_harvested(chip->get_chip_info().harvesting_masks.eth_harvesting_mask);
+        for (int i = 0; i < num_channels; i++) {
+            cluster_desc->idle_eth_channels[chip_id].insert(i);
+        }
     }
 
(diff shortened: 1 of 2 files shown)


device/topology_discovery.cpp
```diff
@@ -453,12 +453,27 @@ void TopologyDiscovery::fill_cluster_descriptor_info() {
             {chip_id, chip->get_chip_info().harvesting_masks.dram_harvesting_mask});
         cluster_desc->eth_harvesting_masks.insert(
             {chip_id, chip->get_chip_info().harvesting_masks.eth_harvesting_mask});
-        cluster_desc->chip_locations.insert({chip_id, eth_coords.at(chip_id)});
+        eth_coord_t eth_coord = eth_coords.at(chip_id);
+        cluster_desc->chip_locations.insert({chip_id, eth_coord});
+        cluster_desc->coords_to_chip_ids[eth_coord.rack][eth_coord.shelf][eth_coord.y][eth_coord.x] = chip_id;
+
+        int num_channels =
+            tt::umd::architecture_implementation::create(tt::ARCH::WORMHOLE_B0)->get_num_eth_channels() -
```
pjanevskiTT, line 461: nit: maybe use constant directly here, no need to create architecture implementation if we know which arch we are using

device/topology_discovery.cpp
```diff
@@ -453,12 +453,27 @@ void TopologyDiscovery::fill_cluster_descriptor_info() {
...
+            CoordinateManager::get_num_harvested(chip->get_chip_info().harvesting_masks.eth_harvesting_mask);
```
pjanevskiTT, line 462: again, this is always 0 for Wormhole

device/tt_cluster_descriptor.cpp
```diff
@@ -844,6 +846,7 @@ void tt_ClusterDescriptor::load_chips_from_connectivity_descriptor(YAML::Node &y
         std::string arch_str = node->second.as<std::string>();
         desc.all_chips.insert(chip_id);
         desc.chip_arch.insert({chip_id, tt::arch_from_str(arch_str)});
+        desc.ethernet_connections.insert({chip_id, {}});
```
pjanevskiTT, line 849: i don't fully understand why this is needed? What is not working without this change?

pjanevskiTT: would just like to know this before approve, other comment you can choose to address

broskoTT: This is needed since I changed the other part of the code to use .at and .insert, instead [].

The actual error doesn't have to do with this, i'd say its a personal preference to not use []

PR #902, opened 2025-05-21
Remove Luwen CEM

pjanevskiTT:
### Issue

#856. I think we have multiple duplicate issues about CEM. Will close all of it once this is merged

### Description

Remove dependency on Luwen's create ethernet map. We have implemented our own cpp implementation of create-ethernet-map in #722. Along with #901 #900, waiting on chip to safely boot should be done so we can properly integrate our own implementation to UMD main code path.
This PR removes Luwen libraries, swaps calling the lib with calling our own class `TopologyDiscovery`.

### List of the changes

- Remove .a libs for create_ethernet_map
- Swap call of the lib for the call to TopologyDiscovery
- Adapt tests

### Testing

CI. 

All runs on pjanevski/remove_cem :
- [x] All post-commit tests : https://github.com/tenstorrent/tt-metal/actions/runs/15258328541
- [x] Blackhole post-commit tests : https://github.com/tenstorrent/tt-metal/actions/runs/15258330498
- [x] (Single-card) Model perf tests : https://github.com/tenstorrent/tt-metal/actions/runs/15258332005
- [x] (Single-card) Device perf regressions : https://github.com/tenstorrent/tt-metal/actions/runs/15258333546
- [x] (T3K) T3000 unit tests : https://github.com/tenstorrent/tt-metal/actions/runs/15258335026
- [x] (T3K) T3000 demo tests : https://github.com/tenstorrent/tt-metal/actions/runs/15258339084
- [x] (TG) TG unit tests : https://github.com/tenstorrent/tt-metal/actions/runs/15258343421
- [x] (TG) TG demo tests : https://github.com/tenstorrent/tt-metal/actions/runs/15258345228

### API Changes
There are no API changes, only change needed is in usage of configure_active_ethernet_cores_for_mmio_device

PR #920, opened 2025-05-24
Fix noc1 topology discovery

broskoTT:
### Description
Taken out from https://github.com/tenstorrent/tt-umd/pull/902

### API Changes

### DIFF
--- a/device/topology_discovery.cpp
+++ b/device/topology_discovery.cpp
@@ -156,7 +156,8 @@ BoardType TopologyDiscovery::get_board_type(eth_coord_t eth_coord, Chip* mmio_ch
         nullptr,
         mmio_chip);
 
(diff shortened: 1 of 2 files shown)


device/topology_discovery.cpp
```diff
@@ -156,7 +156,8 @@ BoardType TopologyDiscovery::get_board_type(eth_coord_t eth_coord, Chip* mmio_ch
         nullptr,
```
broskoTT, line 156: @pjanevskiTT do we have a test for this?

broskoTT: Ah okay, actually NOC1 test goes through this flow once topology discovery is enabled on default path... Then separating this change doesn't make too much sense

PR #921, opened 2025-05-26
#0: Implement default telemetry for BH

ihamer-tt:
Adding support to show default telemetry readings for Blackhole devices


pjanevskiTT (approved):
changes look good. Nit comment is that there is no need to have #0 in the title of the PR, which will be a commit message as well. If there is an issue it should be linked in the description. It's ok to not have an issue as well

PR #932, opened 2025-05-29
Implement constrained topology

pjanevskiTT:
### Issue

#870 

### Description

Implement constrained topology. This PR adds the option to filter Topology discovery based on PCI device id. For example, this enabled someone who uses UMD to take 2 out of 4 cards in T3k system configuration. The idea is to not make any reads/writes to the cards that are not present in pci_target_devices. UMD is going to make reads to get board id from neighbouring chip at the moment. We should work towards reading this information from local ETH core and never have to go to the remote chip. This is going to be addressed in V2 of this feature.

### List of the changes

- Add target_pci_devices to ClusterOptions
- Implemenet constrained topology for Wormhole that is not 6U (Blackhole has all chips PCIE visible at the moment)
- Add a test

### Testing
CI

### API Changes
/

### DIFF
--- a/device/api/umd/device/cluster.h
+++ b/device/api/umd/device/cluster.h
@@ -68,6 +68,11 @@ struct ClusterOptions {
     // will be checked against the cluster descriptor. In case of MOCK and SIMULATION chip types, this check will be
     // skipped, and you can create chips regardless of the devices on the system.
     std::unordered_set<chip_id_t> target_devices = {};
+    // If set, Cluster will target only boards that have the IDs of the chips specified in this set.
+    // If not set, all discovered PCI devices will be used. This can only be used with SILICON chip type.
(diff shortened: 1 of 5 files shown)


device/api/umd/device/cluster.h
```diff
@@ -68,6 +68,11 @@ struct ClusterOptions {
     // will be checked against the cluster descriptor. In case of MOCK and SIMULATION chip types, this check will be
     // skipped, and you can create chips regardless of the devices on the system.
     std::unordered_set<chip_id_t> target_devices = {};
+    // If set, Cluster will target only boards that have the IDs of the chips specified in this set.
```
broskoTT, line 71:
```suggestion
    // If set, Cluster will target only boards that have the PCI IDs specified in this set.
```

PR #933, opened 2025-05-29
Fix active_idx set

broskoTT:
### Issue
An issue was reported by some metal folks

### Description
I assume that the code was meant to do mod, but original write knew in advance that the size wold always be a power of 2. This makes the code correct.

### List of the changes
- Fix active_eth_core_idx core assignment
- Give more descriptive message for when remote_transfer_eth_cores_.empty()

### API Changes

### DIFF
--- a/device/chip/local_chip.cpp
+++ b/device/chip/local_chip.cpp
@@ -409,13 +409,14 @@ void LocalChip::set_remote_transfer_ethernet_cores(const std::set<uint32_t>& cha
 }
 
 tt_xy_pair LocalChip::get_remote_transfer_ethernet_core() {
+    if (remote_transfer_eth_cores_.empty()) {
+        throw std::runtime_error("No remote transfer ethernet cores set.");
+    }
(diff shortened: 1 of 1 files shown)


device/chip/local_chip.cpp
```diff
@@ -409,13 +409,14 @@ void LocalChip::set_remote_transfer_ethernet_cores(const std::set<uint32_t>& cha
 }
 
 tt_xy_pair LocalChip::get_remote_transfer_ethernet_core() {
+    if (remote_transfer_eth_cores_.empty()) {
+        throw std::runtime_error("No remote transfer ethernet cores set.");
+    }
     return {remote_transfer_eth_cores_[active_eth_core_idx].x, remote_transfer_eth_cores_[active_eth_core_idx].y};
 }
 
 void LocalChip::update_active_eth_core_idx() {
```
pjanevskiTT, line 418: as a general comment, I think this function would benefit from the comment about the usacase, like when do we actually use this function. Maybe it's already documented somewhere

pjanevskiTT (approved):

PR #937, opened 2025-05-30
Implement constrained topology v2

pjanevskiTT:
### Issue

#870 

### Description

V2 of the work on constrained topology. As mentioned in the previous PR #932, UMD was doing few unnecessary reads from devices that shouldn't be included inside topology so it could discard it based on board id. Partial, but enough to distinguish boards, information about remote board id is available on local ETH core. Utilize that information to avoid reads to remote chip.

### List of the changes

- Add reading remote board id from local ETH core
- Use this information to filter topology
- Remove reads from remote chip if not necessary

### Testing
CI. Manually tested on T3K

### API Changes
/


device/topology_discovery.cpp
```diff
@@ -421,8 +423,22 @@ bool TopologyDiscovery::is_pcie_chip_id_included(int pci_id) const {
 
 // If pci_target_devices is empty, we should take all the PCI devices found in the system.
 // That is why we have the first part of the condition.
-bool TopologyDiscovery::is_board_id_included(uint64_t board_id) const {
+bool TopologyDiscovery::is_board_id_included(uint32_t board_id) const {
     return pci_target_devices.empty() || board_ids.find(board_id) != board_ids.end();
 }
 
+uint32_t TopologyDiscovery::get_remote_board_id(Chip* chip, tt_xy_pair eth_core) {
```
broskoTT, line 430: Why not a get_board_id implementation in both local and remote TTDevices?

device/topology_discovery.cpp
```diff
@@ -122,7 +116,25 @@ void TopologyDiscovery::get_pcie_connected_chips() {
         }
         std::unique_ptr<LocalChip> chip = nullptr;
         chip = std::make_unique<LocalChip>(TTDevice::create(device_id));
-        board_ids.insert(chip->get_chip_info().chip_uid.board_id);
+
+        // ETH addresses neet to be initialized after the first chip is created, so we could
+        // read the information about offsets of board IDs on ETH core.
+        // TODO: confirm that we should only suppoer one set of addresses so we can remove
```
broskoTT, line 122: nit: 
neet -> need
suppoer -> support
from in runtime -> runtime

PR #943, opened 2025-06-02
Update CPM.cmake

blozano-tt:
### Issue
NA

### Description
It was requested that we respect the environment variable, since CPM documentation advertises it.

### List of the changes
Change default initialization of CPM_SOURCE_CACHE to respect ENV.

### Testing

### API Changes


cmake/CPM.cmake
blozano-tt, line 1: We are discussing if this is worth doing with @ThisIsFineTM

broskoTT:
We'll reopen if needed.
Comment from @blozano-tt :
I really don't want environment variables impacting our build... the only problem is that there will be this weird quirk where metal's version of that file doesn't match UMDs ... but I think there are more egregious things to worry about

Issue #855, opened 2025-05-13
Detect N300 with inaccessible R chip


ihamer-tt:
When the R chip is not working, our stack opens up the card with only the L chip and continues as if working with N150. The failure then gets exposed in metal as:
```
E       RuntimeError: TT_FATAL @ tt_metal/fabric/control_plane.cpp:296: physical_chip_ids.size() == mesh_ew_size
E       info:
E       Did not find edge with expected number of East-West chips 2
```
UMD should at least warn the user about this case (and any similar case where we see a big discrepancy from what is expected).

PR #944, opened 2025-06-03
Verify cluster descriptor information

pjanevskiTT:
### Issue

Add function to verify number of chips in cluster descriptor. Motivated by #855 

### Description

Add function to verify that each chip is assigned a board and the number of chips per board is as expected according to board type.

### List of the changes

- Add function for getting number of chips based on board type
- Add function to verify number of chips and chip to board assignment

### Testing
CI

### API Changes
/

### DIFF
--- a/device/api/umd/device/types/cluster_descriptor_types.h
+++ b/device/api/umd/device/types/cluster_descriptor_types.h
@@ -129,9 +129,9 @@ inline uint32_t get_number_of_chips_from_board_type(const BoardType board_type) 
         case BoardType::P300:
             return 2;
         case BoardType::GALAXY:
-            return 32;
+            return 1;
         case BoardType::UBB:
-            return 8;
+            return 32;
         default:
             throw std::runtime_error("Unknown board type for number of chips calculation.");
     }
--- a/device/api/umd/device/types/cluster_descriptor_types.h
+++ b/device/api/umd/device/types/cluster_descriptor_types.h
@@ -116,6 +116,27 @@ inline BlackholeChipType get_blackhole_chip_type(const BoardType board_type, con
     }
 }
 
+inline uint32_t get_number_of_chips_from_board_type(const BoardType board_type) {
+    switch (board_type) {
+        case BoardType::N150:
+            return 1;
+        case BoardType::N300:
+            return 2;
+        case BoardType::P100:
+            return 1;
+        case BoardType::P150:
+            return 1;
+        case BoardType::P300:
+            return 2;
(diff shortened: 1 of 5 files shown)


device/api/umd/device/types/cluster_descriptor_types.h
```diff
@@ -115,6 +115,27 @@ inline BlackholeChipType get_blackhole_chip_type(const BoardType board_type, con
...
+            return 1;
+        case BoardType::N300:
+            return 2;
+        case BoardType::P100:
+            return 1;
+        case BoardType::P150:
+            return 1;
+        case BoardType::P300:
+            return 2;
+        case BoardType::GALAXY:
```
broskoTT, line 130: Not sure how this would work.
For 4u galaxy you'd see 4 n150 chips with different board_id, and 32 GALAXY chips with different board_id

broskoTT: So yeah, just test this change on 4U before merging

device/tt_cluster_descriptor.cpp
```diff
@@ -1235,3 +1240,25 @@ std::unordered_set<chip_id_t> tt_ClusterDescriptor::get_board_chips(const uint64
     }
     throw std::runtime_error(fmt::format("Board to chips mapping for board {:#x} not found.", board_id));
 }
+
+void tt_ClusterDescriptor::verify_cluster_descriptor_info() {
+    for (const chip_id_t chip : all_chips) {
+        if (!chip_to_board_id.empty() && chip_to_board_id.find(chip) == chip_to_board_id.end()) {
+            log_warning(LogSiliconDriver, "Chip {} does not have a board ID assigned.", chip);
```
broskoTT, line 1247: I think this will start spewing out warnings on 6U, so please test in advance.

device/api/umd/device/types/cluster_descriptor_types.h
```diff
@@ -115,6 +115,27 @@ inline BlackholeChipType get_blackhole_chip_type(const BoardType board_type, con
...
+            return 32;
+        case BoardType::UBB:
```
broskoTT, line 132: You have 8 chips per board id, but 36 chips per board type UBB. So not sure how this will resolve.

broskoTT (approved):
Please test on all configurations to verify what happens

pjanevskiTT:
fixed numbers for 4u (1 per board, it's made from modules) and 6U (32 per board, all UBBs seem to have same board id)

PR #945, opened 2025-06-03
send_tensix_risc_reset now supports core coord for Simulator

dzivanovicTT:
### Issue
NA

### Description
Update the simulation device’s send_tensix_risc_reset to support specific core coordinates.  

### List of the changes
**send_tensix_risc_reset** now accepts a core argument and passes it to the flatbuffer, enabling core-specific RISC reset commands for the simulator.

### Testing
CI

### API Changes
/


broskoTT (approved):
When I was touching this code I didn't put even minimal effort to understand that it should be done like this... thanks!

PR #948, opened 2025-06-03
Topology tool constrained topology

broskoTT:
### Description
Improve topology tool to allow constraining topology

### List of the changes
- Add --device option to filter cluster descriptor based on logical devices
- Add --pci_devices option to perform contrained topology discovery based on pci ids

### Testing
Ran manually on a 3xN300 system, confirmed cluster descriptor was as expected

### API Changes

### DIFF
--- a/tools/topology.cpp
+++ b/tools/topology.cpp
@@ -4,6 +4,7 @@
 #include <cxxopts.hpp>
 #include <tt-logger/tt-logger.hpp>
 
+#include "common.h"
 #include "umd/device/cluster.h"
 #include "umd/device/tt_cluster_descriptor.h"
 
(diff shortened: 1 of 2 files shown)


tools/topology.cpp
```diff
@@ -27,7 +33,21 @@ int main(int argc, char *argv[]) {
...
 
-    std::string output_path = tt::umd::Cluster::create_cluster_descriptor()->serialize_to_file(cluster_descriptor_path);
+    std::unordered_set<int> pci_ids = {};
+    if (result.count("pci_devices")) {
+        pci_ids = extract_int_set(result["pci_devices"]);
+    }
+
+    std::unique_ptr<tt_ClusterDescriptor> cluster_descriptor = tt::umd::Cluster::create_cluster_descriptor("", pci_ids);
+
+    if (result.count("devices")) {
```
pjanevskiTT, line 43: I think behaviour of applying both pci_device filter and logical IDs filter should be documented, at least as comment here. I don't feel it is even useful to use both at the same time but if someone does it, it should be clear what is going to happen

tools/topology.cpp
```diff
@@ -12,8 +13,13 @@ using namespace tt::umd;
 int main(int argc, char *argv[]) {
     cxxopts::Options options("topology", "Extract system topology and save it to a yaml file.");
 
-    options.add_options()("p,path", "File path to save cluster descriptor to.", cxxopts::value<std::string>())(
-        "h,help", "Print usage");
+    options.add_options()("f,path", "File path to save cluster descriptor to.", cxxopts::value<std::string>())(
+        "d,devices",
```
pjanevskiTT, line 17: consider renaming this to logical_ids or something similar. I think devices is not a proper name to use for logical IDs

pjanevskiTT: I think logical_devices is ok...to match pci_devices

Issue #895, opened 2025-05-20
Offset TlbWindow


pjanevskiTT:
#733 is going to introduce structures for managing TLBs using KMD

At the moment, TlbWindow supports only TLB size-aligned addresses. Support constructing TlbWindow with any address such that it still works seamlessly

pjanevskiTT:
closed by #951

PR #951, opened 2025-06-03
Implement TLB window with unaligned address

pjanevskiTT:
### Issue

#895 

### Description

Implement configuring TlbWindow with address that is not aligned to TLB size. Read/writes should still work. TLB size is effectively getting smaller since the user of TlbWindow should be aware of using address that is not aligned to TLB size. We would probably want to have one more layer on top of it that handles custom IO by reconfiguring TlbWindow as needed.

### List of the changes

- Implement offset inside TLB window
- Add test to confirm reading with different offsets on two windows gives same results
- Add test to confirm exception on out of bounds access

### Testing
CI + additional test

### API Changes
/

### DIFF
--- a/device/tt_device/tlb_window.cpp
+++ b/device/tt_device/tlb_window.cpp
@@ -10,17 +10,20 @@
 namespace tt::umd {
 
 TlbWindow::TlbWindow(std::unique_ptr<TlbHandle> handle, const tlb_data config) : tlb_handle(std::move(handle)) {
-    tlb_handle->configure(config);
+    tlb_data aligned_config = config;
+    aligned_config.local_offset = config.local_offset & ~(tlb_handle->get_size() - 1);
+    tlb_handle->configure(aligned_config);
+    offset_from_aligned_addr = config.local_offset - (config.local_offset & ~(tlb_handle->get_size() - 1));
 }
 
 void TlbWindow::write32(uint64_t offset, uint32_t value) {
     validate(offset, sizeof(uint32_t));
-    *reinterpret_cast<volatile uint32_t*>(tlb_handle->get_base() + offset) = value;
+    *reinterpret_cast<volatile uint32_t*>(tlb_handle->get_base() + get_total_offset(offset)) = value;
 }
 
 uint32_t TlbWindow::read32(uint64_t offset) {
     validate(offset, sizeof(uint32_t));
-    return *reinterpret_cast<volatile uint32_t*>(tlb_handle->get_base() + offset);
+    return *reinterpret_cast<volatile uint32_t*>(tlb_handle->get_base() + get_total_offset(offset));
 }
 
 void TlbWindow::write_register(uint64_t offset, uint32_t value) { write32(offset, value); }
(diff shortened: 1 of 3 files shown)


device/tt_device/tlb_window.cpp
```diff
@@ -10,17 +10,20 @@
 namespace tt::umd {
 
 TlbWindow::TlbWindow(std::unique_ptr<TlbHandle> handle, const tlb_data config) : tlb_handle(std::move(handle)) {
-    tlb_handle->configure(config);
+    tlb_data aligned_config = config;
+    aligned_config.local_offset = config.local_offset & ~(tlb_handle->get_size() - 1);
+    tlb_handle->configure(aligned_config);
+    offset_from_aligned_addr = config.local_offset - (config.local_offset & ~(tlb_handle->get_size() - 1));
```
joelsmithTT, line 16: is this needed or is a call to configure() sufficient?

joelsmithTT (approved):
I think this is a reasonable abstraction (I've had success with something similar), although I am curious to see how well it ends up working for UMD.

An edge condition to consider is the case where software wants to write across a window boundary.  At what level of abstraction should the window reconfiguration happen?  That is, should `TlbWindow::write_block` be able to write more than the window size worth of data (by reconfiguring itself)?  Or is that up to the caller?

I don't think there is a right or wrong answer here... my thought process is more along the lines of documenting the interface so that programmers know what to expect of it; make it difficult to use wrong, etc.

Another thought - something that may seem counterintuitive is the use of 0.  Let's say you want to read32 at address 0xfffb202e4ull:
```
tlb_data tlbdata{ .local_offset = 0xfffb202e4ull, /* etc */ };
TlbWindow window(std::move(handle), tlbdata);
tlb_window.read32(0); // read it
```
... it's convenient, but possibly confusing if a programmer is not familiar with the implementation.

No changes needed -- just some things to consider as this work progresses.

device/tt_device/tlb_window.cpp
pjanevskiTT: KMD ioctl was returning error for config that has unaligned address. I am ok with this behaviour, makes UMD and other clients behave properly

pjanevskiTT:
Good points.
What I am aiming to do is have TlbWindow in a state where it can be configured arbitrarily and be used more or less in a same way UMD uses TLBs today. Basically have function like setup_tlb and then have size and address to write/read. There is already code that takes care of read/writes that go across the boundary, so for now I would like to keep that code so the TlbWindow does not take care of reads/writes across boundary. So basically have TlbWindow replace 1-1 scheme that we already have today.
That work requires striping down UMD scheme of mapping the whole PCIE BAR as well. I think after those 2 are done we can start and think about abstraction (maybe TlbWindow istself) that is going to take care of reads/writes across boundary. I just feel that a lot of UMD code already relies on functions that are already implemented for this

device/tt_device/tlb_window.cpp
pjanevskiTT: https://github.com/tenstorrent/tt-kmd/blob/main/wormhole.c#L386 I would assume this was throwing...as I said, I think this is good

PR #957, opened 2025-06-04
removing mmio support for mock cluster

dzivanovicTT:
### Issue
NA

### Description
Removing mmio support for mock cluster since currently Simulator doesn't support this functionality.

### List of the changes
Remove chips with mmio from mock cluster

### Testing
CI

### API Changes
/


device/tt_cluster_descriptor.cpp
```diff
@@ -560,12 +560,10 @@ std::unique_ptr<tt_ClusterDescriptor> tt_ClusterDescriptor::create_mock_cluster(
             logical_id;
         log_debug(tt::LogSiliconDriver, "{} - adding logical: {}", __FUNCTION__, logical_id);
         desc->chip_board_type.insert({logical_id, board_type});
-        desc->chips_with_mmio.insert({logical_id, logical_id});
```
broskoTT, line 563: I'm not sure what are all the implications of this. I think our code at many places expects at least one local chip per cluster, both in umd and metal. And on some paths for local chips we expect an associated mmio with it.
Also with the removal of fill_chips_group_by_closest_mmio, not sure if tt_cluster uses that information to fill some cluster related information.

Can you consider adding a dma read/write call for Simulation, which would just call regular read/write call, if that was the only problem?

device/tt_cluster_descriptor.cpp
```diff
@@ -560,12 +560,10 @@ std::unique_ptr<tt_ClusterDescriptor> tt_ClusterDescriptor::create_mock_cluster(
             logical_id;
```
broskoTT, line 560: Please comment in the PR in the Testing section whether you manually tested this (since we don't have a CI setup) with tt-umd-simulators repo and if the tests pass with this change

device/tt_cluster_descriptor.cpp
dzivanovicTT: Ok, make sense, I will try that way.

broskoTT (approved):

Issue #820, opened 2025-05-05
Topology discovery logical harvesting mask


broskoTT:
We currently have a different format of harvesting masks being reported. HW reports it in one form, then we convert back and forth to logical layout. This will be possible to remove once we consume all CEM functionality into ourselves.

broskoTT:
Considre removing get_tensix_harvesting_mask

PR #960, opened 2025-06-06
Refactor harvesting masks usage

pjanevskiTT:
### Issue

#820 

### Description

Refactor harvesting mask usage. This PR makes all harvesting masks on all architecture be stored in a way that hides physical harvesting masks. For example, on Wormhole, tensix harvesting masks are reported in a way that bit 0 represents physical row 0 etc...This PR changes all harvesting masks to correspond to NOC0 layout (bit 0 is for NOC0 row 0, DRAM bank 0 from soc descriptor etc...). It also adds a map of harvesting masks structures rather than storing each harvesting mask (DRAM, ETH, PCIE, Tensix) separately.

### List of the changes

- Store tensix harvesting mask in NOC0 layout
- Remove separete maps for each harvesting mask type
- Add map for storing harvesting masks structures
- Change the tests to reflect the changes

### Testing
CI

### API Changes
/

### DIFF
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -271,7 +271,17 @@ HarvestingMasks Cluster::get_harvesting_masks(
         return HarvestingMasks{};
     }
 
-    return cluster_desc->get_harvesting_masks(chip_id) | simulated_harvesting_masks;
+    HarvestingMasks cluster_harvesting_masks = cluster_desc->get_harvesting_masks(chip_id);
+    log_info(
+        LogSiliconDriver,
+        "Harvesting mask for chip {} is {:#x} (NOC0: {:#x}, simulated harvesting mask: "
+        "{:#x}).",
+        chip_id,
+        cluster_harvesting_masks.tensix_harvesting_mask | simulated_harvesting_masks.tensix_harvesting_mask,
+        cluster_harvesting_masks.tensix_harvesting_mask,
+        simulated_harvesting_masks.tensix_harvesting_mask);
+
+    return cluster_harvesting_masks | simulated_harvesting_masks;
 }
 
 void Cluster::ubb_eth_connections(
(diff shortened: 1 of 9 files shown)


tests/wormhole/test_arc_messages_wh.cpp
```diff
@@ -30,7 +30,9 @@ TEST(WormholeArcMessages, WormholeArcMessagesHarvesting) {
             0,
             0);
 
-        EXPECT_EQ(arc_msg_return_values[0], harvesting_mask_cluster_desc);
+        EXPECT_EQ(
+            CoordinateManager::shuffle_tensix_harvesting_mask(tt::ARCH::WORMHOLE_B0, arc_msg_return_values[0]),
```
broskoTT, line 34: So a subtle difference now is that the cluster descriptor will now hold logical harvesting mask? And we don't surface the physical one anywhere? Do we see this as a problem maybe?

device/cluster.cpp
```diff
@@ -261,23 +261,6 @@ void Cluster::add_chip(const chip_id_t& chip_id, const ChipType& chip_type, std:
     chips_.emplace(chip_id, std::move(chip));
 }
 
-uint32_t Cluster::get_tensix_harvesting_mask(
-    chip_id_t chip_id, tt_ClusterDescriptor* cluster_desc, HarvestingMasks& simulated_harvesting_masks) {
-    uint32_t tensix_harvesting_mask_physical_layout = cluster_desc->get_harvesting_info().at(chip_id);
-    uint32_t tensix_harvesting_mask = CoordinateManager::shuffle_tensix_harvesting_mask(
-        cluster_desc->get_arch(chip_id), tensix_harvesting_mask_physical_layout);
-    log_info(
```
broskoTT, line 269: Yeah I think this log is very very useful, can we move it somewhere? Obviously you wouldn't show the physical_layout anymore, as its not available outside ttdevice

Issue #842, opened 2025-05-09
Consider adding python support


broskoTT:
Consider adding pybinds over our whole api.
See what tt-exalens already has.

PR #965, opened 2025-06-06
Add nanobind skeleton

broskoTT:
### Issue
#842 

### Description
Sets up initial nanobind structure. It only exposes Cluster constructor and one method to showcase the usage. Rest of functionality will follow up.

### List of the changes
- Nanobind cpp exposing python api
- Nanobind python example unit test
- Running this unit test in CI
- Updating docker image to support python
- Consume nanobind through CPM
- Fix write permissions for docker image generation
- gitignore __pycache__

### Testing
Added a CI test

### API Changes
Adding python API


nanobind/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+        ${CMAKE_CURRENT_SOURCE_DIR}
+        "$<TARGET_PROPERTY:umd::device,INCLUDE_DIRECTORIES>"
+)
+
+# Set properties for the Python module
+set_target_properties(
+    nanobind_libdevice
+    PROPERTIES
+        OUTPUT_NAME
+            "libdevice" # Name of the Python module
```
afuller-TT, line 28: Does `device` already have this name?  Isn't that going to conflict?

nanobind/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+        PREFIX
+            "" # Remove the default 'lib' prefix for Python modules
```
afuller-TT, line 30: .... but we just added it in on line 28;  could just say "device" up there and not need to deviate from defaults here, no?

nanobind/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+find_package(Python REQUIRED COMPONENTS Development)
+
+# Define the Python interface target
+nanobind_add_module(nanobind_libdevice MODULE cluster.cpp)
+
+# Link the Python interface target to libdevice and nanobind
+target_link_libraries(
```
afuller-TT, line 7: (Nit) on predictable structure: define from the "inside -> out".  I wrote up more details & rationale here: https://tenstorrent.atlassian.net/wiki/spaces/~71202038f3ffeaaecb4198a0bd0b8c9c0d5726/pages/1124568013/Andrew+s+Highly+Opinionated+Project+Structure#CMake-template

nanobind/CMakeLists.txt
broskoTT: I think the cmake module will be named nanobind_libdevice, but it will be exposed to Python as "libdevice", so no conflict. I mean it does work.

.gitignore
```diff
@@ -9,3 +9,4 @@ compile_commands.json
 .vscode/
 *.log
 *.csv
+__pycache__
```
pjanevskiTT, line 12: this needs empty line at the end?

pjanevskiTT (approved):
Cool, I think this is good first step towards adding python support

nanobind/CMakeLists.txt
afuller-TT: The specific (nit) recommendation here is that after we name the module we then describe itself (set_target_properties) before we go about describing its relationships to external entities (target_link_libraries).

ie: just swap those two.  Again, a nit.

PR #970, opened 2025-06-09
Delete core binary and update .gitignore

pjanevskiTT:
### Issue

Delete core binary accidentally checked in in #857 

### Description

Delete core binary. I think this is generated when running some syseng tooling from UMD folder, I was testing BW of sysmem buffers for comparison with syseng

### List of the changes

- Delete core binary

### Testing
CI

### API Changes
/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ compile_commands.json
 .vscode/
 *.log
 *.csv
+core


core
broskoTT, line 1: Please add it to .gitignore

broskoTT (approved):

Issue #971, opened 2025-06-09
Concurrent jobs on the same device.


dbabokin:
According to @ayerofieiev-tt multiple concurrent users on the same TT device are not supported by design. I would expect that two users running jobs concurrently on the same machine with single TT device, would either get an error "device is busy" or yield until device is available. None of this happens currently.

Instead, jobs run concurrently and produce race condition described in this issue: https://github.com/tenstorrent/tt-metal/issues/22754

broskoTT:
So we actually did some work on this front, you can even see some tests https://github.com/tenstorrent/tt-umd/blob/main/tests/unified/multiprocess.cpp

Our focus at that point was to enable two Clusters to be constructed in parallel over the same device, and thus allow two processes not to step on each other. Without this work, the problems you're currently seeing would've likely been less subtle.

However, we didn't do anything on the front of actually blocking two different running processes.
If I lock the start_device/stop_device functions would that be sensible and enough for this problem? From metal's perspective it should be fine, I see that both open_device then start_device is called from the same initialize_device_drivers in tt_cluster.

broskoTT:
Can you folks try with **brosko/chip_lkock** branch?
We currently have some issues in our tests, so I can't test this properly on our end.

broskoTT:
> Does both running instances need to use this branch?

Yes

dbabokin:
> > Does both running instances need to use this branch?
> 
> Yes

With both on the workspaces using your branch it works as expected.

One thing, though. It probably worth to warn the user that the job is not hang, but wait for the other job to complete. Otherwise it's not very user friendly. If you have information who owns the lock that would be even more helpful message.

broskoTT:
Constructors were fixed, now we still have multiple clusters being created in some tests

dbabokin:
Seems that all related tt-metal issues are fixed, what are the next steps for this issue?

Issue #422, opened 2024-12-18
Deassert in tests


broskoTT:
Deassert was recently removed from out unit tests: https://github.com/tenstorrent/tt-umd/pull/421
Rather, we should figure out a way to load a firmware which doesn't do anything (maybe just halts the processor?), and then reintroduce deasserts.

The way they are written now the processors can go haywire and even crash the host system though iATUs.

broskoTT:
Lazar mentioned that these scripts can be used to generate minimal C code for triscs: https://github.com/tenstorrent/tt-llk-bh/tree/ldjurovic/llk_tests/tests

This should be edited so that a fixed location in L1 is updated, so we know that triscs have started working. Loading this FW should be then done through the test infra.

broskoTT:
@tt-vjovanovic mentioned that tt-lens already has some test similar to this. Check how tt-lens does this before jumping into it.
FYI @pjanevskiTT

broskoTT:
My expectation is that multiprocess tests will fail with addition of start_device. Verify that, and then try to rebase on top of https://github.com/tenstorrent/tt-umd/pull/987, which should allow these tests to pass

PR #972, opened 2025-06-10
Chip lock in use

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/971

### Description
This PR introduces locking such that a single device can't be "started" from two processes.

### List of the changes
- Introduce CHIP_IN_USE lock. This lock is locked on start_device and released on stop_device

### Testing
Confirmed by external user that this solved the problem.
We're missing a CI test for this, should be possible after https://github.com/tenstorrent/tt-umd/issues/422

### API Changes

### DIFF
--- a/.github/docker_install_common.sh
+++ b/.github/docker_install_common.sh
@@ -13,7 +13,12 @@ apt-get update && apt-get install -y \
     libyaml-cpp-dev \
     libboost-all-dev \
     wget \
-    yamllint
+    yamllint \
+    python3-dev \
+    python3-pip
+
+# Install Python dependencies
+python3 -m pip install --no-cache-dir pytest
 
 # gcc-12 should be available only for ubuntu 22 and not 20
 if apt-cache show gcc-12 > /dev/null 2>&1; then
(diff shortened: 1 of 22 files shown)


device/utils/lock_manager.cpp
```diff
@@ -19,6 +19,7 @@ const std::unordered_map<MutexType, std::string> LockManager::MutexTypeToString
     {MutexType::NON_MMIO, "TT_NON_MMIO"},
     {MutexType::MEM_BARRIER, "TT_MEM_BARRIER"},
     {MutexType::CREATE_ETH_MAP, "CREATE_ETH_MAP"},
+    {MutexType::CHIP_IN_USE, "CHIP_IN_USE"},
```
pjanevskiTT, line 22: two nit things I see here:
- why have we diverger from adding TT_ prefix to mutex names?
- Do we still need CREATE_ETH_MAP mutex?

broskoTT: Full lock name is "TT_UMD_LOCK.TT_MEM_BARRIER_0", so we can actually remove the other TT_ prefixes.
For CREATE_ETH_MAP, probably not.
I'm waiting for unique chip id change, then we can clear up this path as well

PR #985, opened 2025-06-17
Remove close_device from tt_SimulationDevice destructor

dzivanovicTT:
### Issue
/

### Description
Remove `close_device` from `tt_SimulationDevice` destructor since it is called from cluster method close_device.
If needed, this method can still be called directly.

### List of the changes
Remove `close_device` from `tt_SimulationDevice` destructor

### Testing
CI
Manually checked test `test_simulation_device`

### API Changes
/


device/simulation/tt_simulation_device.cpp
```diff
@@ -83,8 +83,6 @@ tt_SimulationDevice::tt_SimulationDevice(const tt_SimulationDeviceInit& init) :
     uv_loop_close(loop);
 }
 
-tt_SimulationDevice::~tt_SimulationDevice() { close_device(); }
```
broskoTT, line 86: So this ends up being called two times? Is that the issue?

broskoTT (approved):

Issue #936, opened 2025-05-30
Make UMD unit tests work for simulator


pjanevskiTT:
Make as much UMD unit tests as possible work for simulator as well

Ideally make fixture for all groups of tests that decides during initialization how to make a cluster, then runs same tests both for silicon or simulator

pjanevskiTT:
this work is tracked by #2186 and #2185

PR #989, opened 2025-06-19
 Minor test code fixies

dzivanovicTT:
### Issue
#936

### Description
Minor test code fixies, applying review comments from #982

### List of the changes
- Call start_device only for SImulator until #422 is fixed
- Minor test code fixies

### Testing
CI
Manually tested on Simulator and Silicon to check if I didn't break previous functionality.

### API Changes
/

### DIFF
--- a/tests/api/test_cluster.cpp
+++ b/tests/api/test_cluster.cpp
@@ -38,13 +38,14 @@ constexpr std::uint32_t DRAM_BARRIER_BASE = 0;
 class ClusterReadWriteL1Test : public ::testing::TestWithParam<ClusterOptions> {};
 
 std::vector<ClusterOptions> get_cluster_options_for_param_test() {
+    constexpr const char* TT_UMD_SIMULATOR_ENV = "TT_UMD_SIMULATOR";
     std::vector<ClusterOptions> options;
     options.push_back(ClusterOptions{.chip_type = ChipType::SILICON});
-    if (std::getenv("TT_UMD_SIMULATOR")) {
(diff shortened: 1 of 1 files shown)


tests/api/test_cluster.cpp
```diff
@@ -417,9 +418,12 @@ TEST_P(ClusterReadWriteL1Test, ReadWriteL1) {
     if (cluster->get_target_device_ids().empty()) {
         GTEST_SKIP() << "No chips present on the system. Skipping test.";
     }
-    tt_device_params device_params;
-    device_params.init_device = true;
-    cluster->start_device(device_params);
+    if (options.chip_type == SIMULATION) {
```
pjanevskiTT, line 421: nit: would maybe like to see ChipType::SIMULATION as full naming here

tests/api/test_cluster.cpp
```diff
@@ -468,7 +472,7 @@ INSTANTIATE_TEST_SUITE_P(
             case ChipType::SILICON:
                 return "Silicon";
             case ChipType::SIMULATION:
-                return "Simulator";
+                return "Simulation";
```
pjanevskiTT, line 475: out of curiosity, what are these strings used for?

pjanevskiTT (approved):
thanks for cleaning this up

tests/api/test_cluster.cpp
dzivanovicTT: an oversight, thank you

Issue #687, opened 2025-04-04
Convert tt_xy_pair to CoreCoord where it makes sense


broskoTT:
Follow up from #682 
I think we should try to use CoreCoord everywhere internally up to the point of going to the noc, or to TTDevice.

broskoTT:
Also follow up from https://github.com/tenstorrent/tt-umd/pull/704

If it is hard, then at least rename all "core" variables to "virtual_core" , "translated_core", etc

pjanevskiTT:
@aleksamarkovicTT @broskoTT can we maybe close this now?

broskoTT:
The first iteration was done long time ago. I think only stuff in TopologyDiscovery was left with tt-xy_pair.
@aleksamarkovicTT is working anyway on setting up base API to use CoreCoord which has LITERAL, this covers implicitly this work as well

PR #997, opened 2025-06-24
Deassert in tests (only for BRISC)

nbuncicTT:
### Issue
#422 

### Description
Machine instructions for BRISC are loaded on L1 which makes the BRISC be in valid state after deasserting reset 

### List of the changes
- Added a test which sets the BRISC program, and the program sets a variable which is later compared if the test is succesul

### Testing
Manual test

### API Changes
/

### DIFF
--- a/tests/api/test_cluster.cpp
+++ b/tests/api/test_cluster.cpp
@@ -7,6 +7,7 @@
 #include <gtest/gtest.h>
 
 #include <algorithm>
+#include <cstdint>
 #include <cstdlib>  // for std::getenv
 #include <filesystem>
 #include <string>
@@ -19,6 +20,7 @@
 #include "umd/device/chip/mock_chip.h"
 #include "umd/device/cluster.h"
 #include "umd/device/tt_cluster_descriptor.h"
+#include "umd/device/tt_silicon_driver_common.hpp"
 #include "umd/device/wormhole_implementation.h"
 
 // TODO: obviously we need some other way to set this up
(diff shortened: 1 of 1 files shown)


tests/api/test_cluster.cpp
```diff
@@ -410,6 +413,57 @@ TEST(TestCluster, TestClusterAICLKControl) {
...
+    const tt_SocDescriptor& soc_desc = cluster->get_soc_descriptor(chip_id);
+    auto tensix_core = cluster->get_soc_descriptor(chip_id).get_cores(CoreType::TENSIX).at(0);
+
+    cluster->write_to_device(
+        brisc_program.data(), brisc_program.size() * sizeof(uint32_t), chip_id, tensix_core, 0x0000'0000);
+
+    cluster->wait_for_non_mmio_flush(chip_id);
+
+    // By setting these reset options, all other cores except the BRISC are/stay in reset
+    TensixSoftResetOptions brisc_deassert_reset{
```
pjanevskiTT, line 451: probably a gap in my knowledge here, but we start BRISC by writing 0 to some register? I would expect to have to set some value only for BRISC.
Whatever is true here, I think it would be helpful to expose the API which accepts the cores that you want to deassert, not do it in a way that is done here now

tests/api/test_cluster.cpp
```diff
@@ -410,6 +413,57 @@ TEST(TestCluster, TestClusterAICLKControl) {
...
+TEST(TestCluster, DeassertResetBrisc) {
+    ClusterOptions options{};
+    std::unique_ptr<Cluster> cluster = std::make_unique<Cluster>(options);
+
+    if (cluster->get_target_device_ids().empty()) {
+        GTEST_SKIP() << "No chips present on the system. Skipping test.";
+    }
+
+    // The values in brisc_program are machine instructions which represent
+    // a RISCV program. This program is downloaded at address 0x0000'0000 on L1 SRAM on a Tensix,
```
pjanevskiTT, line 425: nit: This program is "stored" at address..... You can choose whether you want to swap downloaded

tests/api/test_cluster.cpp
```diff
@@ -7,6 +7,7 @@
 #include <gtest/gtest.h>
 
 #include <algorithm>
+#include <cstdint>
```
pjanevskiTT, line 10: why we didn't need this include so far?

tests/api/test_cluster.cpp
```diff
@@ -410,6 +413,57 @@ TEST(TestCluster, TestClusterAICLKControl) {
     }
 }
 
+TEST(TestCluster, DeassertResetBrisc) {
```
pjanevskiTT, line 416: general comment, I think it would be helpful to document the process here. Small README in tests folder somewhere about how did you get instructions from source code. I think it can be helpful to someone who wants to extends these kind of tests

tests/api/test_cluster.cpp
nbuncicTT: I suppose it was used via other header files, I'll remove it.

tests/api/test_cluster.cpp
nbuncicTT: I'll do so, thanks

tests/api/test_cluster.cpp
broskoTT: Yeah I agree we can be more clear about this through the API. But in anycase, out of scope of this PR.

broskoTT: @nbuncicTT if it's a trivial change feel free to change it here. If it isn't, feel free to create a separate issue as a follow up and then don't fix it in this PR.

tests/api/test_cluster.cpp
broskoTT: +1 Yes, this is important, please

tests/api/test_cluster.cpp
```diff
@@ -410,6 +413,57 @@ TEST(TestCluster, TestClusterAICLKControl) {
...
+    cluster->wait_for_non_mmio_flush(chip_id);
+
+    // By setting these reset options, all other cores except the BRISC are/stay in reset
+    TensixSoftResetOptions brisc_deassert_reset{
+        TensixSoftResetOptions::NCRISC | TensixSoftResetOptions::TRISC0 | TensixSoftResetOptions::TRISC1 |
+        TensixSoftResetOptions::TRISC2};
+
+    auto chip = cluster->get_chip(chip_id);
+    chip->send_tensix_risc_reset(
+        cluster->get_soc_descriptor(chip_id).translate_coord_to(tensix_core, CoordSystem::VIRTUAL),
```
broskoTT, line 457: Yeah calling .translate_coord_to is bad, should be fixed as part of https://github.com/tenstorrent/tt-umd/issues/687.
No action for this PR

tests/api/assembly_programs_for_tests.hpp
```diff
@@ -0,0 +1,15 @@
+// SPDX-FileCopyrightText: © 2023 Tenstorrent Inc.
```
pjanevskiTT, line 1: maybe this file can be moved to tests/utils...

pjanevskiTT (approved):
nice work

device/chip/chip.cpp
```diff
@@ -142,7 +142,8 @@ void Chip::enable_ethernet_queue(int timeout_s) {
 void Chip::send_tensix_risc_reset(tt_xy_pair core, const TensixSoftResetOptions& soft_resets) {
     auto valid = soft_resets & ALL_TENSIX_SOFT_RESET;
     uint32_t valid_val = (std::underlying_type<TensixSoftResetOptions>::type)valid;
-    write_to_device_reg(core, &valid_val, 0xFFB121B0, sizeof(uint32_t));
+    auto architecture_implementation = tt::umd::architecture_implementation::create(get_tt_device()->get_arch());
+    write_to_device_reg(core, &valid_val, architecture_implementation->get_tensix_soft_reset_addr(), sizeof(uint32_t));
```
broskoTT, line 146: Maybe this should be moved to TTDevice class. Although I see it was already here, so we can leave it for some other PR

tests/api/test_cluster.cpp
```diff
@@ -411,6 +414,53 @@ TEST(TestCluster, TestClusterAICLKControl) {
     }
 }
 
+// This test uses the machine instructions from the header file assembly_programs_for_tests.hpp. How to generate
+// this program is explained in the GENERATE_ASSEMBLY_FOR_TESTS.md file.
+TEST(TestCluster, DeassertResetBrisc) {
```
broskoTT, line 419: Let's think about how can we introduce ->start_device() in other tests as well in future PRs

broskoTT (approved):
No action, thanks!

Issue #587, opened 2025-03-06
Read asic location from telemetry


pjanevskiTT:
Syseng folks are going to put information on asic_location of each chip inside the telemetry. Implement reading of that information in Blackhole telemetry reader class.

PR #1003, opened 2025-06-24
Read asic location from telemetry

pjanevskiTT:
### Issue

https://github.com/tenstorrent/tt-umd/issues/587

### Description

Read asic location from CMFW telemetry rather than reading it from ETH core.

### List of the changes
- Read asic location from CMFW telemetry
- Delete reading it from ETH cores

### Testing

CI

### API Changes
/

### DIFF
--- a/device/tt_device/tt_device.cpp
+++ b/device/tt_device/tt_device.cpp
@@ -571,11 +571,16 @@ std::vector<DramTrainingStatus> TTDevice::get_dram_training_status() {
 
 double TTDevice::get_asic_temperature() { return get_firmware_info_provider()->get_asic_temperature(); }
 
+uint8_t TTDevice::get_asic_location() { return get_firmware_info_provider()->get_asic_location(); }
+
 ChipInfo TTDevice::get_chip_info() {
     ChipInfo chip_info;
+
     chip_info.noc_translation_enabled = get_noc_translation_enabled();
     chip_info.chip_uid.board_id = get_board_id();
     chip_info.board_type = get_board_type();
+    chip_info.asic_location = get_asic_location();
+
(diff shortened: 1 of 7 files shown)


device/tt_device/tt_device.cpp
```diff
@@ -422,4 +422,15 @@ std::vector<DramTrainingStatus> TTDevice::get_dram_training_status() {
...
 }
 
+ChipInfo TTDevice::get_chip_info() {
+    ChipInfo chip_info;
+    chip_info.noc_translation_enabled = get_noc_translation_enabled();
+    chip_info.chip_uid.board_id = get_board_id();
+    chip_info.board_type = get_board_type();
+
+    chip_info.chip_uid.asic_location =
+        telemetry->is_entry_available(TAG_ASIC_LOCATION) ? telemetry->read_entry(TAG_ASIC_LOCATION) : 0;
```
broskoTT, line 432: Hmmm, 0 is a valid value, so having this set to 0 doesn't actually communicate that it is an invalid value, right?

device/tt_device/tt_device.cpp
```diff
@@ -422,4 +422,15 @@ std::vector<DramTrainingStatus> TTDevice::get_dram_training_status() {
...
+    return chip_info;
```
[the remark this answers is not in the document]
pjanevskiTT: it is in firmware provider now, safe for new fw

device/tt_device/tt_device.cpp
pjanevskiTT: Supporting asic location as optional is much more work, this should be always available on all multi chip cards

Issue #871, opened 2025-05-14
Figure out unique chip IDs


broskoTT:
Having unique chip ID is not clear for each configuration/chip.
Investigate how to have a unique (either global unique id, or at least unique id on the system).

Make sure this covers having a couple of  cards on system which aren't connected through ETH.

broskoTT:
@pjanevskiTT added support for unique IDs which might slightly change in the future, but everything should be there for deterministic IDs https://github.com/tenstorrent/tt-umd/pull/1024

PR #1004, opened 2025-06-24
Read unique ASIC id from telemetry

pjanevskiTT:
### Issue

#871 

### Description

Read ASIC id from telemetry. This field should represent unique chip ID. It currently works for Blackhole, Wormhole should come later. Anyway, this implementation should not break anything on Wormhole side since it basically just stores the information in cluster descriptor.

### List of the changes

- Add ASIC ID tag
- Read ASIC ID in tt device
- Store asic_id in cluster descriptor
- (De)serialize support for unique chip id

### Testing

CI

### API Changes
/

### DIFF
--- a/device/api/umd/device/tt_cluster_descriptor.h
+++ b/device/api/umd/device/tt_cluster_descriptor.h
@@ -64,6 +64,7 @@ protected:
     std::map<chip_id_t, std::set<uint32_t>> idle_eth_channels = {};
     std::map<uint64_t, std::unordered_set<chip_id_t>> board_to_chips = {};
     std::unordered_map<chip_id_t, uint64_t> chip_to_board_id = {};
+    std::unordered_map<chip_id_t, uint64_t> chip_to_unique_id = {};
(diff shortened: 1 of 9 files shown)


device/api/umd/device/tt_cluster_descriptor.h
```diff
@@ -64,6 +64,7 @@ class tt_ClusterDescriptor {
     std::map<chip_id_t, std::set<uint32_t>> idle_eth_channels = {};
     std::map<uint64_t, std::unordered_set<chip_id_t>> board_to_chips = {};
     std::unordered_map<chip_id_t, uint64_t> chip_to_board_id = {};
+    std::unordered_map<chip_id_t, uint64_t> chip_to_unique_id = {};
```
broskoTT, line 67: Seems very similar to `chip_id_to_chip_uid` , we should remove that one, right? Why is UUID used for 6u?

broskoTT (approved):

PR #1011, opened 2025-06-26
Don't implement default throw Chip implementation

### LINKED ISSUE #988: Remove all "throw" implementation from base classes
Specifically in Chip and TTDevice classes, we have a lot of implementations of functions which are throw std::runtime_error.
These should be pure virtual, and implementation should be moved to derived classes.
The problem is that we often break simulator api, which doesn't popup in compile time, so this would guard it from happening again.

broskoTT:
### Issue
Fixes #988

### Description
Change chip to be pure abstract without any throws. This will help us not miss implementing some chip functions in simulator in the future

### List of the changes
- Remove all throw std::runtime_exception from chip class.
- Implement same functions at all derived classes

### Testing
Existing CI should be enough

### API Changes

### DIFF
--- a/device/api/umd/device/topology_discovery.h
+++ b/device/api/umd/device/topology_discovery.h
@@ -103,6 +103,8 @@ private:
 
     // All board ids that should be included in the cluster descriptor.
     std::unordered_set<uint32_t> board_ids;
+
+    std::unordered_map<chip_id_t, std::set<uint32_t>> active_eth_channels_per_chip;
 };
 
 }  // namespace tt::umd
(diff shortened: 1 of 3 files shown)


device/api/umd/device/chip/chip.h
```diff
@@ -113,8 +111,6 @@ class Chip {
     ChipInfo chip_info_;
 
     tt_SocDescriptor soc_descriptor_;
-
-    std::unique_ptr<TTDevice> tt_device_ = nullptr;
```
pjanevskiTT, line 117: nit: why remove `tt_device_`

pjanevskiTT (approved):

device/api/umd/device/chip/chip.h
broskoTT: With this change we're also avoiding the code in Chip class from accessing tt_device directly, which should be the case.
Also if somebody tries using get_tt_device, they will get a more explicit error rather than nullptr

PR #1015, opened 2025-06-27
Fix active eth links for filtering by pci_target_devices

broskoTT:
### Issue
The issue is in the future which we're trying to avoid

### Description
We had a mismatch in functionality on logical id and pci id filtering, in how eth links which are outside of filtered cluster are reported.
This is important, since wrong reporting will lead to metal putting its own eth dispatch firmware on all idle cores, and then without reset you can't make the whole cluster work again.

### List of the changes
- Change the way we track active_eth_links in topology discovery

### Testing
Tested manually on an IRD T3K machine with -p filter.

### API Changes

### DIFF
--- a/device/api/umd/device/topology_discovery.h
+++ b/device/api/umd/device/topology_discovery.h
@@ -103,6 +103,8 @@ private:
 
     // All board ids that should be included in the cluster descriptor.
     std::unordered_set<uint32_t> board_ids;
+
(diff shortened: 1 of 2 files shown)


pjanevskiTT (approved):
looks good, please run this on t3k machine with pci_devices filter for some board to confirm that it works

broskoTT:
@pjanevskiTT that's exactly how i even saw this working not as expected. Here's the run:
image

Issue #1020, opened 2025-06-27
Refactor TopologyDiscovery


broskoTT:
To be unified between 6u and older WH.
Currently we have completely separate code branches.
The difference is in reading the actual eth links. Unifying that is not in scope of this PR. Once #871 lands, this code can become the same for all configs.

pjanevskiTT:
closed by #1058

PR #1023, opened 2025-06-30
Refactor topology discovery class for legacy FW Wormhole

pjanevskiTT:
### Issue

Working towards #1020 

### Description

Refactor topology discovery class. The class still works for Wormhole configurations other than 6U. This PR refactors main functions of TopologyDiscovery class to be as agnostic as possible to arch, version of FW and other things. Comments are added in .h file which functions should be overrided and which should be kept same for all configurations. It also introduces unique ID for all chips that are used to discover the topology instead of legacy ETH coordinates. ETH coordinates are used just for tunneling now, that should be deprecated with fabric. 

### List of the changes

- Refactor topology discovery class
- Add functions that should be overrided for different configurations, main functions should stay the same
- Add comments with explanations and TODOs for newly added functions
- Rely on unique ID of the chip for topology discovery - picked up from ETH core
- Trim unused fields in EthAddresses

### Testing

Testing manually on wormhole configurations

### API Changes
/

### DIFF
--- a/nanobind/CMakeLists.txt
+++ b/nanobind/CMakeLists.txt
@@ -2,7 +2,7 @@
 # https://github.com/wjakob/nanobind/blob/master/docs/building.rst
 # Use Development.Module to support manylinux
 # Python module library should not dynamically link libpython.so or use Development.Embed
-find_package(Python REQUIRED COMPONENTS Development)
+find_package(Python REQUIRED COMPONENTS Development.Module)
 
 # Define the Python interface target
 nanobind_add_module(nanobind_libdevice MODULE cluster.cpp)


device/api/umd/device/topology_discovery.h
```diff
@@ -79,15 +47,50 @@ class TopologyDiscovery {
 
     // Returns mangled remote board id from local ETH core.
     // This information can still be used to unique identify a board.
+    // TODO: override this logic for different configs. This is in group of functions
```
broskoTT, line 50: does that mean we could use different derived implementations?
So you'd create 6UTopologyDiscovery based on board id for example?

device/api/umd/device/topology_discovery.h
```diff
@@ -79,15 +47,50 @@ class TopologyDiscovery {
...
 
     // Returns mangled local board id from local ETH core.
     // This information can still be used to unique identify a board.
+    // TODO: override this logic for different configs. This is in group of functions
+    // that we should override for T3K/6U/BH...
     uint32_t get_local_board_id(Chip* chip, tt_xy_pair eth_core);
 
+    // TODO: override this logic for different configs. This is in group of functions
+    // that we should override for T3K/6U/BH...
+    uint64_t get_local_asic_id(Chip* chip, tt_xy_pair eth_core);
```
broskoTT, line 62: are xy supposed to be physical or translated? Please add a comment somewhere in the header file what is actually the case for all pairs

device/topology_discovery.cpp
```diff
@@ -130,69 +158,63 @@ void TopologyDiscovery::get_pcie_connected_chips() {
...
+
+        if (port_status == eth_unknown || port_status == eth_unconnected) {
+            channel++;
+            continue;
+        }
+
+        return get_local_asic_id(chip, eth_core);
+    }
+
+    return chip->get_tt_device()->get_board_id();
```
broskoTT, line 179: Can you just add a comment in this function on WHY is it done like this? If all eth are unconnected we'll return a board id, but if it is connected we return something else?

What happens with n150 when it's connected vs unconnected? Then it will not report the same id? Do we treat asic_id as a unique id?

device/topology_discovery.cpp
```diff
@@ -130,69 +158,63 @@ void TopologyDiscovery::get_pcie_connected_chips() {
...
 
-        eth_coords.emplace(chip_id, current_chip_unique_coord.eth_coord);
-        unique_coord_to_chip_id.emplace(current_chip_unique_coord, chip_id);
-
-        discovered_chips.insert(current_chip_unique_coord);
+        // TODO: this neeeds to be moved to specific logic for Wormhole with legacy FW.
+        eth_coords.emplace(chip_id, get_local_eth_coord(chip.get()));
     }
 
     for (const auto& [chip_id, chip] : chips) {
```
broskoTT, line 206: It doesn't have to be done in this PR, but can we please unify this loop with the next while loop, like the standard way to do BFS ?

You'd have "chips_to_discover" collection holding pointers to Chip classes (or something similar) and tread all reads/writes the same for local and remote

device/topology_discovery.cpp
```diff
@@ -206,44 +228,25 @@ void TopologyDiscovery::discover_remote_chips() {
...
-                remote_chips_to_discover.insert(unique_coord);
-                remote_unique_coord_to_mmio_chip_id.emplace(unique_coord, chip_id);
+            if (discovered_chips.find(remote_asic_id) == discovered_chips.end()) {
+                remote_chips_to_discover.insert(remote_asic_id);
+                remote_tt_devices_to_discover.emplace(remote_asic_id, std::move(remote_tt_device));
+                remote_asic_id_to_mmio_chip_id.emplace(remote_asic_id, chip_id);
             } else {
-                chip_id_t current_chip_id = unique_coord_to_chip_id.at(current_chip_unique_coord);
-                chip_id_t remote_chip_id = unique_coord_to_chip_id.at(unique_coord);
+                chip_id_t current_chip_id = asic_id_to_chip_id.at(current_chip_asic_id);
```
broskoTT, line 245: This is just chip_id?

device/topology_discovery.cpp
```diff
@@ -206,44 +228,25 @@ void TopologyDiscovery::discover_remote_chips() {
...
+                chip_id_t remote_chip_id = asic_id_to_chip_id.at(remote_asic_id);
                 Chip* remote_chip = chips.at(remote_chip_id).get();
                 CoreCoord physical_remote_eth =
-                    CoreCoord(remote_noc_x, remote_noc_y, CoreType::ETH, CoordSystem::PHYSICAL);
+                    CoreCoord(remote_eth_core.x, remote_eth_core.y, CoreType::ETH, CoordSystem::PHYSICAL);
                 CoreCoord logical_remote_eth =
                     remote_chip->get_soc_descriptor().translate_coord_to(physical_remote_eth, CoordSystem::LOGICAL);
                 ethernet_connections.push_back({{current_chip_id, channel}, {remote_chip_id, logical_remote_eth.y}});
```
broskoTT, line 252: You're only even entering this branch for connections between two local chips.
Will this add connection in both directions?

device/topology_discovery.cpp
```diff
@@ -258,61 +261,44 @@ void TopologyDiscovery::discover_remote_chips() {
...
             chip = std::make_unique<RemoteChip>(
                 tt_SocDescriptor(
-                    tt_device->get_arch(),
+                    remote_tt_device->get_arch(),
                     chip_info.noc_translation_enabled,
                     chip_info.harvesting_masks,
                     chip_info.board_type),
                 std::move(remote_tt_device));
 
+            Chip* remote_chip_ptr = chip.get();
```
broskoTT, line 288: nit: Consider not having this variable, just call get whenever you need a ptr

device/topology_discovery.cpp
```diff
@@ -324,64 +310,36 @@ void TopologyDiscovery::discover_remote_chips() {
...
-                        new_remote_chips.insert(new_unique_coord);
-                        remote_unique_coord_to_mmio_chip_id.emplace(new_unique_coord, mmio_chip_id);
+                tt_xy_pair remote_eth_core = get_remote_eth_core(chips.at(chip_id - 1).get(), eth_core);
+
+                uint64_t new_asic_id = get_remote_asic_id(chips.at(chip_id - 1).get(), {eth_core.x, eth_core.y});
+
+                if (discovered_chips.find(new_asic_id) == discovered_chips.end()) {
+                    if (remote_chips_to_discover.find(new_asic_id) == remote_chips_to_discover.end()) {
+                        std::unique_ptr<RemoteWormholeTTDevice> new_remote_tt_device =
+                            create_remote_tt_device(chips.at(chip_id - 1).get(), {eth_core.x, eth_core.y}, mmio_chip);
```
broskoTT, line 320: Please add something like 
`chip_id_t remote_chip_id = asic_id_to_chip_id.at(asic_id_to_discover);`
at the beginning of the for loop which defines asic_id_to_discover, and use that here, instead of chip_id-1

broskoTT (approved):
Glad this is going in the right direction!

device/api/umd/device/topology_discovery.h
pjanevskiTT: yes, exactly. The idea is that based on FW/config maybe we need to gather info in different ways but the functions `get_pcie_connected_chip`, `discover_remote_chips` and `fill_cluster_descriptor_info` should try and stay more or less the same across architectures

device/topology_discovery.cpp
```diff
@@ -324,64 +310,36 @@ void TopologyDiscovery::discover_remote_chips() {
...
                         remote_chip->get_soc_descriptor().translate_coord_to(physical_remote_eth, CoordSystem::LOGICAL);
                     ethernet_connections.push_back(
                         {{current_chip_id, channel}, {remote_chip_id, logical_remote_eth.y}});
                 }
-
                 channel++;
             }
         }
 
         remote_chips_to_discover = new_remote_chips;
```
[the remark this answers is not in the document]
pjanevskiTT: Won't rewrite it here, PR would blow up, there are subtle differences in the loop that I will try to adress when TopologyDiscovery matures even further

device/topology_discovery.cpp
pjanevskiTT: Won't rewrite it here, PR would blow up, there are subtle differences in the loop that I will try to adress when TopologyDiscovery matures even further

device/topology_discovery.cpp
pjanevskiTT: I already have the pointer, will use that instead of always getting from the map

device/topology_discovery.cpp
```diff
@@ -206,44 +228,25 @@ void TopologyDiscovery::discover_remote_chips() {
...
 
-            unique_coord.board_id = remote_tt_device->get_board_id();
+            tt_xy_pair remote_eth_core = get_remote_eth_core(chip.get(), eth_core);
 
-            if (discovered_chips.find(unique_coord) == discovered_chips.end()) {
-                remote_chips_to_discover.insert(unique_coord);
-                remote_unique_coord_to_mmio_chip_id.emplace(unique_coord, chip_id);
+            if (discovered_chips.find(remote_asic_id) == discovered_chips.end()) {
+                remote_chips_to_discover.insert(remote_asic_id);
+                remote_tt_devices_to_discover.emplace(remote_asic_id, std::move(remote_tt_device));
```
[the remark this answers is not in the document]
pjanevskiTT: good catch, thanks

PR #1024, opened 2025-06-30
Add unique chip IDs to cluster descriptor

pjanevskiTT:
### Issue

Add unique IDs to cluster descriptor

### Description

Add unique IDs to cluster descriptor. Unique IDs are going to be used to discover the topology starting from #1023. Store these IDs in cluster descriptor and add (de)serializing of the fields. #1004 is something similar but relies on telemetry. When that telemetry is released to all machines we can switch to using that instead of unique IDs from ETH cores.

### List of the changes

- Add field in cluster descriptor for storing unique IDs
- Store the data from topology discovery
- Add (de)serializing functions
- Add a test

### Testing

CI + tested manually

### API Changes
/


broskoTT (approved):
Maybe this cluster descriptor could be better (e.g. have ids under board section, rename "chips" to "chip_eth_coords", but let's not over engineer for now

PR #1028, opened 2025-07-01
Limit pci target devices to single host in topology discovery

pjanevskiTT:
### Description
(Add freeform description.)

### List of the changes

### Testing

### API Changes

### DIFF
--- a/device/topology_discovery.cpp
+++ b/device/topology_discovery.cpp
@@ -21,7 +21,15 @@ extern bool umd_use_noc1;
 namespace tt::umd {
 
 TopologyDiscovery::TopologyDiscovery(std::unordered_set<chip_id_t> pci_target_devices) :
-    pci_target_devices(pci_target_devices) {}
+    pci_target_devices_(pci_target_devices) {
+    // We are going to treat empty pci_target_devices as a request to enumerate all PCI devices.
+    // For now this is useful to not try and go outisde of the scope of single host.
+    // We might want to change that in the future but this is guarding us from that case at the moment.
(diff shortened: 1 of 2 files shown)


device/topology_discovery.cpp
```diff
@@ -21,7 +21,15 @@ extern bool umd_use_noc1;
 namespace tt::umd {
 
 TopologyDiscovery::TopologyDiscovery(std::unordered_set<chip_id_t> pci_target_devices) :
-    pci_target_devices(pci_target_devices) {}
+    pci_target_devices_(pci_target_devices) {
+    // We are going to treat empty pci_target_devices as a request to enumerate all PCI devices.
+    // For now this is useful to not try and go outisde of the scope of single host.
+    // We might want to change that in the future but this is guarding us from that case at the moment.
```
broskoTT, line 27: I don't agree with these changes for a couple of reasons:
- This change breaks TG and potentially some other configurations, but probably not. We need to be more careful for this
- Same filtering can be done in tt_metal or whereever, for example if this is used on 6u, just pass all 0-31 chips as pci_target_devices.
- We're limiting our flexibility. We should not have policy stuff like this, only mechanism. When somebody would want to go outside the host, they wouldn't be able to. 

Maybe we can add another option in ClusterOptions ".discover_outside_host = true" or something like that?

Also, in case of going out of the single host, we'd likely need another way to filter out which chips are discovered (so that we don't have to create a full logical topology to filter out), but this is long down the road to figure out.

pjanevskiTT:
i've addressed this in #1037, closing...

PR #1030, opened 2025-07-02
Deassert all riscs on tensix

nbuncicTT:
### Issue
#422 

### Description
Added two tests demonstrating how to assert and deassert the reset signals for BRISC, TRISC cores, and NCRISC. 

### List of the changes
- The first test shows how to make a counter program that you can check on the BRISC core
- The second test runs the same counter program on all other cores (TRISCs and NCRISC) and shows how the BRISC core sets the other cores before they execute the counter program

### Testing
Manual

### API Changes
/


tests/api/assembly_programs_for_tests.hpp
```diff
@@ -0,0 +1,94 @@
+// SPDX-FileCopyrightText: © 2023 Tenstorrent Inc.
```
pjanevskiTT: nice comments in the whole file :). No action

tests/api/test_cluster.cpp
```diff
@@ -97,7 +102,7 @@ TEST(ApiClusterTest, DifferentConstructors) {
     umd_cluster = nullptr;
 
     if (chips_available) {
-        // 2. Constructor which allows choosing a subset of Chips to open.
+        // 2. Constructor which allows choosing a unique_combination_of_risc_cores of Chips to open.
```
pjanevskiTT, line 105: not sure I understand why this change was mode. I also feel this comment doesn't make much sense now

tests/api/test_cluster.cpp
```diff
@@ -273,8 +278,8 @@ TEST(ApiClusterTest, SimpleIOSpecificSiliconChips) {
 }
 
 TEST(ClusterAPI, DynamicTLB_RW) {
-    // Don't use any static TLBs in this test. All writes go through a dynamic TLB that needs to be reconfigured for
-    // each transaction
+    // Don'configuration_of_risc_core use any static TLBs in this test. All writes go through a dynamic TLB that needs
```
pjanevskiTT, line 281: seems like you replaced t in the comment by accident here?

tests/api/test_cluster.cpp
```diff
@@ -360,8 +365,8 @@ TEST(TestCluster, PrintAllSiliconChipsAllCores) {
 }
 
 // It is expected that logical ETH channel numbers are in the range [0, num_channels) for each
-// chip. This is needed because of eth id readouts for Blackhole that don't take harvesting into
-// acount. This test verifies that both for Wormhole and Blackhole.
+// chip. This is needed because of eth id readouts for Blackhole that don'configuration_of_risc_core take harvesting
```
pjanevskiTT, line 368: same as above

tests/api/test_cluster.cpp
```diff
@@ -461,6 +467,250 @@ TEST(TestCluster, DeassertResetBrisc) {
...
+
+    if (cluster->get_target_device_ids().empty()) {
+        GTEST_SKIP() << "No chips present on the system. Skipping test.";
+    }
+
+    auto tensix_l1_size = cluster->get_soc_descriptor(0).worker_l1_size;
+    std::vector<uint32_t> zero_data(tensix_l1_size, 0x00000000);
+
+    constexpr uint64_t counter_address = 0x00010000;
+    constexpr uint64_t brisc_code_address = 0x00000000;
```
pjanevskiTT, line 481: again, might be a personal preference here, but having this many digits in literal to initialize some var to zero seems like too much. Why not simply have
`constexpr uint64_t brisc_code_address = 0;`

tests/api/test_cluster.cpp
```diff
@@ -461,6 +467,250 @@ TEST(TestCluster, DeassertResetBrisc) {
...
+
+    uint32_t first_readback_value = 0x00000000;
+    uint32_t second_readback_value = 0x00000000;
+
+    auto chip_id = *cluster->get_target_device_ids().begin();
```
pjanevskiTT, line 486: iterate over all chips instead of doing to only for one single chip_id

device/chip/chip.cpp
```diff
@@ -6,9 +6,12 @@
 
 #include "umd/device/chip/chip.h"
 
+#include <cstdint>
```
pjanevskiTT, line 9: not really sure why some includes were added without any additional code here?

tests/api/test_cluster.cpp
```diff
@@ -7,8 +7,10 @@
 #include <gtest/gtest.h>
 
 #include <algorithm>
+#include <cstdint>
 #include <cstdlib>  // for std::getenv
 #include <filesystem>
+#include <ios>
```
pjanevskiTT, line 13: why is this included?

tests/api/test_cluster.cpp
```diff
@@ -461,6 +467,250 @@ TEST(TestCluster, DeassertResetBrisc) {
...
+        auto chip = cluster->get_chip(chip_id);
+
+        cluster->write_to_device(zero_data.data(), zero_data.size() * sizeof(uint32_t), chip_id, tensix_core, 0x0);
+
+        cluster->wait_for_non_mmio_flush(chip_id);
+
+        TensixSoftResetOptions select_all_tensix_riscv_cores{TENSIX_ASSERT_SOFT_RESET};
+
+        chip->set_tensix_risc_reset(
+            cluster->get_soc_descriptor(chip_id).translate_coord_to(tensix_core, CoordSystem::VIRTUAL),
```
pjanevskiTT, line 500: I can see function call
`cluster->get_soc_descriptor(chip_id).translate_coord_to(tensix_core, CoordSystem::VIRTUAL)`
happening multiple times in a loop. Consider having only one call in the beginning and then using the variable from then on

tests/api/test_cluster.cpp
```diff
@@ -461,6 +467,250 @@ TEST(TestCluster, DeassertResetBrisc) {
...
+
+        cluster->wait_for_non_mmio_flush(chip_id);
+
+        cluster->read_from_device(
+            &first_readback_value, chip_id, tensix_core, counter_address, sizeof(first_readback_value));
+
+        cluster->read_from_device(
+            &second_readback_value, chip_id, tensix_core, counter_address, sizeof(second_readback_value));
+
+        EXPECT_NE(second_readback_value, first_readback_value);
```
pjanevskiTT, line 526: I think this check would benefit from comment, smh like
```Since we expect BRISC to work and constantly increment counter in L1, we expect values to be different on two reads from device```

File: tests/api/test_cluster.cpp [hunk 11]
```diff
@@ -461,6 +467,250 @@ TEST(TestCluster, DeassertResetBrisc) {
...
+
+        cluster->wait_for_non_mmio_flush(chip_id);
+
+        cluster->read_from_device(
+            &first_readback_value, chip_id, tensix_core, counter_address, sizeof(first_readback_value));
+
+        cluster->read_from_device(
+            &second_readback_value, chip_id, tensix_core, counter_address, sizeof(second_readback_value));
+
+        EXPECT_EQ(second_readback_value, first_readback_value);
```
pjanevskiTT, line 540: same as above

tests/api/test_cluster.cpp
```diff
@@ -461,6 +467,250 @@ TEST(TestCluster, DeassertResetBrisc) {
...
+
+    constexpr uint64_t brisc_code_address = 0x00000000;
+
+    uint32_t first_readback_value = 0x00000000;
+    uint32_t second_readback_value = 0x00000000;
+
+    auto tensix_l1_size = cluster->get_soc_descriptor(0).worker_l1_size;
+    std::vector<uint32_t> zero_data(tensix_l1_size, 0x00000000);
+
+    auto chip_id = *cluster->get_target_device_ids().begin();
```
pjanevskiTT, line 628: again, iterate over all chips

tests/api/test_cluster.cpp
```diff
@@ -461,6 +467,250 @@ TEST(TestCluster, DeassertResetBrisc) {
...
+
+    auto tensix_cores = cluster->get_soc_descriptor(chip_id).get_cores(CoreType::TENSIX);
+
+    TensixSoftResetOptions risc_cores{TensixSoftResetOptions::NONE};
+
+    for (const CoreCoord& tensix_core : tensix_cores) {
+        auto chip = cluster->get_chip(chip_id);
+
+        chip->set_tensix_risc_reset(
+            cluster->get_soc_descriptor(chip_id).translate_coord_to(tensix_core, CoordSystem::VIRTUAL),
```
pjanevskiTT, line 639: again, call this only once per tensix

tests/api/test_cluster.cpp
```diff
@@ -461,6 +467,250 @@ TEST(TestCluster, DeassertResetBrisc) {
...
+        EXPECT_EQ(second_readback_value, first_readback_value);
+    }
+}
+
+using RiscCoreProgramConfig = std::tuple<uint64_t, uint32_t, std::array<uint32_t, 6>, TensixSoftResetOptions>;
+using RiscSetUnderTest = std::vector<RiscCoreProgramConfig>;
+
+class ClusterAssertDeassertRiscsTest : public ::testing::TestWithParam<RiscSetUnderTest> {
+public:
+    static constexpr uint64_t trisc0_code_address = 0x00020000;
```
pjanevskiTT, line 549: personally, I would remove all leading 0s from literals we have in the initialization

tests/api/test_cluster.cpp
```diff
@@ -461,6 +467,250 @@ TEST(TestCluster, DeassertResetBrisc) {
...
+            &second_readback_value, chip_id, tensix_core, counter_address, sizeof(second_readback_value));
+
+        EXPECT_EQ(second_readback_value, first_readback_value);
+    }
+}
+
+using RiscCoreProgramConfig = std::tuple<uint64_t, uint32_t, std::array<uint32_t, 6>, TensixSoftResetOptions>;
+using RiscSetUnderTest = std::vector<RiscCoreProgramConfig>;
+
+class ClusterAssertDeassertRiscsTest : public ::testing::TestWithParam<RiscSetUnderTest> {
```
pjanevskiTT, line 547: Move this to some .h file, I think we should only have TEST_* code here. This is something similar to fixture implemented in #982

tests/api/test_cluster.cpp
nbuncicTT: My mistake, gonna change it

PR #1040, opened 2025-07-04
Move Blackhole code to TopologyDiscovery

pjanevskiTT:
### Issue

#1020 

### Description

Move Blackhole code to topology discovery. This PR separates logic for Wormhole and Blackhole into separate classes. Base TopologyDiscovery should have the code that is common for both arch, also not dependent on remote communication way, config etc...Derived classes for WH and BH should hide low level details of ETH connectivity and remote communication.

### List of the changes

- Make derived topology class for Wormhole
- Make derived topology class for Blackhole
- In base TopologyDiscovery keep just the common code for BH and WH

### Testing

CI + manual testing

### API Changes
/

### DIFF
--- a/device/topology/topology_discovery_wormhole.cpp
+++ b/device/topology/topology_discovery_wormhole.cpp
@@ -252,9 +252,7 @@ uint32_t TopologyDiscoveryWormhole::get_remote_eth_channel(Chip* chip, tt_xy_pai
     tt_xy_pair remote_eth_core = get_remote_eth_core(chip, local_eth_core);
 
     // TODO(pjanevski): explain in comment why we are using chip instead of remote chip.
-    return chip->get_soc_descriptor()
-        .translate_coord_to(remote_eth_core, umd_use_noc1 ? CoordSystem::NOC1 : CoordSystem::NOC0, CoordSystem::LOGICAL)
-        .y;
+    return chip->get_soc_descriptor().translate_coord_to(remote_eth_core, CoordSystem::NOC0, CoordSystem::LOGICAL).y;
 }
 
 bool TopologyDiscoveryWormhole::is_using_eth_coords() { return !is_running_on_6u; }


device/api/umd/device/topology_discovery.h
```diff
@@ -119,13 +119,15 @@ class TopologyDiscovery {
...
-    std::unordered_set<uint32_t> board_ids;
+    std::unordered_set<uint64_t> board_ids;
 
     std::unordered_map<chip_id_t, std::set<uint32_t>> active_eth_channels_per_chip;
 
     const std::string sdesc_path;
 
     bool is_running_on_6u = false;
+
+    bool is_running_blackhole = false;
```
broskoTT, line 130: So you seem to have both ways of separating specific functionality of TopologyDiscovery, one is through these flags is_running_on_6u and is_running_blackhole, and now the other is through derived classes? This can get confusing, I think you should try to choose a single design.

broskoTT: To be clear you don't have to branch out two new classes, one for non-6u and one for 6u. Rather, you can just have a TopologyDiscoveryWormhole6U : TopologyDiscoveryWormhole and just override a couple of functions where you used the is_running_on_6u flag

device/cluster.cpp
```diff
@@ -1013,24 +1015,9 @@ std::unique_ptr<tt_ClusterDescriptor> Cluster::create_cluster_descriptor(
...
-                chip = std::make_unique<LocalChip>(TTDevice::create(device_id));
-            } else {
-                chip = std::make_unique<LocalChip>(sdesc_path, TTDevice::create(device_id));
-            }
-            chips.emplace(chip_id, std::move(chip));
-            chip_id++;
-        }
-
-        return Cluster::create_cluster_descriptor(chips);
+        return TopologyDiscoveryBlackhole(pci_target_devices, sdesc_path).create_ethernet_map();
```
broskoTT, line 1018: It seems strange that the called of TopologyDiscovery has to know in advance which arch is it calling this for.
Can you wrap this up somewhere within TopologyDiscovery class? For example, you can have TopologyDiscovery::create_cluster_descriptor() which does this if else logic, and uses PCIDevice::enumerate_devices_info to jumpstart

pjanevskiTT:
Works on 6U

device/topology/topology_discovery.cpp
```diff
@@ -0,0 +1,273 @@
...
+
+extern bool umd_use_noc1;
+
+namespace tt::umd {
+
+std::unique_ptr<tt_ClusterDescriptor> TopologyDiscovery::create_cluster_descriptor(
+    std::unordered_set<chip_id_t> pci_target_devices, const std::string& sdesc_path) {
+    auto pci_devices_info = PCIDevice::enumerate_devices_info();
+    if (pci_devices_info.empty()) {
+        return std::make_unique<tt_ClusterDescriptor>();
```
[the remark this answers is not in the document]
pjanevskiTT: it will

device/topology/topology_discovery.cpp
```diff
@@ -0,0 +1,273 @@
...
+                    {{current_chip_asic_id, channel},
+                     {get_remote_asic_id(chip, eth_core), get_remote_eth_channel(chip, eth_core)}});
+
+                channel++;
+                continue;
+            }
+
+            chip->set_remote_transfer_ethernet_cores(active_eth_channels_per_chip.at(current_chip_asic_id));
+
+            uint64_t remote_asic_id = get_remote_asic_id(chip, eth_core);
```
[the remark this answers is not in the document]
pjanevskiTT: I am not gonna make this change because it extracts some info that is not always needed, I will add the information for remote chip ID that is outside of UMD cluster

PR #1064, opened 2025-07-09
Put everything in tt::umd namespace

broskoTT:
### Issue
#103 

### Description
Making our classes consistent with tt::umd namespace. Following PR will change the naming to remove tt_ prefixes.

### List of the changes
- Wrapped all in tt::umd namespace
- Removed all tt::umd:: prefixes for types
- Removed using TLB_DATA = tt::umd::tlb_data;
- Added tt::umd:: prefixes for std::hash implementations
- Removed "using namespace tt::umd;" from .cpp files, and switched to "namespace tt::umd {"
- Changed "namespace tt { namespace cpuset {" to "namespace tt::cpuset {"
- Changed "#include "umd/device/tt_cluster_descriptor.h" to "#include "umd/device/types/cluster_descriptor_types.h"" in cpuset, and added missing regex/memory includes in other files
- Added many usings at the end of the header files so that we can do two phase change in tt_metal and exalens

### Testing
Not much testing, clients still build

### API Changes
There are API changes in the naming, but I created a change so it allows for both new and old names to be used, and will create metal changes subsequently.

### DIFF
--- a/device/api/umd/device/architecture_implementation.h
+++ b/device/api/umd/device/architecture_implementation.h
@@ -17,12 +17,12 @@
 #include "umd/device/types/tlb.h"
 #include "umd/device/types/xy_pair.h"
 
+namespace tt::umd {
+
 struct tt_device_l1_address_params;
 struct tt_driver_host_address_params;
 struct tt_driver_eth_interface_params;
 struct tt_driver_noc_params;
-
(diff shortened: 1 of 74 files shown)


pjanevskiTT (approved):
fine with this change as long as it doesn't break anything

Issue #1033, opened 2025-07-02
Improve strategy for PCIe DMA buffer allocation


pjanevskiTT:
At the moment UMD always tries to allocate 1MB buffer for PCIe DMA transacations

With IOMMU, we can make this buffer significantly larger. Investigate what is the optimal size and whether bigger buffers even have the benefit

Withtout IOMMU, seems we can allocate 2MB buffers as well, bump that size

pjanevskiTT:
closed by #1068

PR #1068, opened 2025-07-09
Make PCIe DMA buffer allocation more robust

pjanevskiTT:
### Issue

#1033 

### Description

So far we have always allocated 1MB buffer for PCIe DMA transactions. This PR introduces improved strategy to allocate PCIe DMA buffer. UMD will try to allocate buffers starting from 16MB (biggest TLB size on Wormhole) up to a page size. It will probably succeed much sooner than page size, but that is being put as a limit right now. In theory larger PCIe DMA buffers should mean smaller number of DMA transactions, but in practice the perf is dependent on other stuff, such as copying different chunk of data. Even if copying larger chunks of data is slow, larger buffer could potentially make room for improvement in terms of copying more smaller chunks in parallel and then doing one bigger DMA transfer if needed. The strategy of improved e2e perf of PCIe DMA API is still to be investigated.

### List of the changes

- Improve PCIe DMA buffer allocation
- Start from 16MB in IOMMU case
- Start from 2MB in no IOMMU case

### Testing
CI  + manual testing

### API Changes
/

### DIFF
--- a/device/pcie/pci_device.cpp
+++ b/device/pcie/pci_device.cpp
@@ -762,6 +762,7 @@ bool PCIDevice::try_allocate_pcie_dma_buffer_iommu(const size_t dma_buf_size) {
 
         return true;
     } catch (...) {
+        log_debug(LogSiliconDriver, "Failed to allocate PCIe DMA buffer of size {} with IOMMU enabled.", dma_buf_size);
         munmap(dma_buf_mapping, dma_buf_alloc_size);
         return false;
     }
@@ -793,6 +794,11 @@ bool PCIDevice::try_allocate_pcie_dma_buffer_no_iommu(const size_t dma_buf_size)
             log_error(LogSiliconDriver, "Failed to map DMA buffer: {}", strerror(errno));
             return false;
         } else {
+            log_debug(
+                LogSiliconDriver,
+                "Allocated PCIe DMA buffer of size {} for PCI device {}.",
+                dma_buf_alloc_size,
+                pci_device_num);
             dma_buffer.buffer = (uint8_t *)buffer;
             dma_buffer.completion = (uint8_t *)buffer + dma_buf_size;
             dma_buffer.buffer_pa = dma_buf.out.physical_address;


device/pcie/pci_device.cpp
```diff
@@ -536,3 +475,86 @@ semver_t PCIDevice::read_kmd_version() {
...
+            } else {
+                dma_buffer.buffer = (uint8_t *)buffer;
+                dma_buffer.completion = (uint8_t *)buffer + dma_buf_size;
+                dma_buffer.buffer_pa = dma_buf.out.physical_address;
+                dma_buffer.completion_pa = dma_buf.out.physical_address + dma_buf_size;
+                dma_buffer.size = dma_buf_size;
+                break;
+            }
+        }
+        dma_buf_size >>= 1;
```
broskoTT, line 558: So each iteration of loop you're dividing the size by 2?
The intention is to try and allocate smaller and smaller buffers, where also the intention is that it will always work for a buf size == page size? Add comments explaining that, this is a bit confusing on the first look

device/pcie/pci_device.cpp
```diff
@@ -457,7 +396,7 @@ PCIDevice::~PCIDevice() {
     }
 
     if (dma_buffer.buffer != nullptr && dma_buffer.buffer != MAP_FAILED) {
-        munmap(dma_buffer.buffer, DMABUF_TOTAL_SIZE);
+        munmap(dma_buffer.buffer, dma_buffer.size);
```
joelsmithTT, line 399: + 0x1000?

device/pcie/pci_device.cpp
```diff
@@ -536,4 +475,88 @@ std::unique_ptr<TlbHandle> PCIDevice::allocate_tlb(const size_t tlb_size, const
...
+    //    - Driver initiates DMAs based on its knowledge of the buffer
+    // 5. + 0x1000 is for the completion page.  Since this entire implementation
+    // is a temporary hack until it's implemented in the driver, we'll need to
+    // poll a completion page to know when the DMA is done instead of receiving
+    // an interrupt.
+
+    // Allocation tries to allocate larger DMA buffers first. Starting size depends on whether IOMMU is enabled or not.
+    // If IOMMU is enabled, we will try to allocate 16MB buffer first.
+    // If IOMMU is not enabled, we will try to allocate 2MB buffer first.
+    // If that fails, we will try smaller sizes until we can't allocate even single page.
```
joelsmithTT, line 508: The comments about "Lack of current IOMMU support" and "we could move to an IOMMU-based scheme" are not congruent with "If IOMMU is...".

device/pcie/pci_device.cpp
```diff
@@ -806,4 +745,108 @@ uint8_t PCIDevice::read_command_byte(const int pci_device_num) {
...
+        dma_buffer.size = dma_buf_size;
+
+        return true;
+
+    } catch (...) {
+        munmap(dma_buf_mapping, dma_buf_alloc_size);
+        return false;
+    }
+
+    return true;
```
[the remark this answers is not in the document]
pjanevskiTT: addressed

PR #1074, opened 2025-07-10
Inherit stdout and stderr in simulator process

mcraigheadTT:
### Issue
none

### Description
Inherit stdout and stderr handles instead of redirecting to /dev/null, so that the simulator output is not lost

### List of the changes
Mark stdout/stderr handles UV_INHERIT_FD
Minor reordering of variable declarations for clarity

### Testing
Tested with Metal programming examples on simulator

### API Changes
/


mcraigheadTT:
I am not using run.sh (my sim binary is able to act as run.sh) but I'm doing something similar with freopen.  That said, this is suboptimal for at least a few reasons:
- failure to catch early errors as you mentioned
- fairly inconvenient vs. getting the errors in the same console, significantly slows down my edit/test cycle (made worse b/c there are some additional issues w/ telling the difference between normal and abnormal shutdown on each side of the pipe)
- sim output is not interleaved with the test output

IMO at minimum /dev/null is the wrong default for sure.  It is possible to redirect to a file or even to /dev/null from run.sh, but it is not possible to re-obtain the handles if they are not inherited to begin with.

Issue #1052, opened 2025-07-07
Add action for UMD microbenchmarks


pjanevskiTT:
Add action for UMD microbenchmarks

This does not need to be in CI necessarily, rather just as something that can be run using GH actions so someone can verify the perf expected

pjanevskiTT:
closed by #1079

PR #1079, opened 2025-07-11
Add UMD perf workflow

pjanevskiTT:
### Issue

#1052 

### Description

Add Github workflow for UMD perf tests. We are not going to run this on CI for now, because it takes lot more time than regular CI. Idea is that someone who wants to see UMD perf can run Github action and just copy-paste perf if needed from the log to the README table. For now we are running perf tests just on N150 with IOMMU enabled.

### List of the changes

- Add yaml for build and run benchmarks workflow
- Add yaml for run benchmarks workflow
- Reduce TLB tests in terms of sizes since they take too long. We should think about proper number of iterations/sizes for these tests

### Testing

Manual testing by running Github actions

### API Changes
/

### DIFF
--- a/.github/workflows/build-and-run-all-benchmarks.yml
+++ b/.github/workflows/build-and-run-all-benchmarks.yml
@@ -1,0 +1,44 @@
+# Build and then run all benchmarks, on all supported archs.
+name: Build and run all benchmarks
+
+on:
+  workflow_dispatch:
+    inputs:
+      build-type:
+        required: false
+        default: Release
+
+jobs:
+  build-tests:
+    secrets: inherit
+    strategy:
+      fail-fast: false
+      matrix:
+        ubuntu-docker-version: [
+          'ubuntu-22.04',
+          'ubuntu-24.04',
+        ]
+    uses: ./.github/workflows/build-tests.yml
+    with:
+      ubuntu-docker-version: ${{ matrix.ubuntu-docker-version}}
+      timeout: 10
+      build-type: ${{ inputs.build-type || 'Release' }}
+
+  run-benchmarks:
+    secrets: inherit
+    needs: build-tests
+    strategy:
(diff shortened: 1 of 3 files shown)


.github/workflows/build-and-run-all-benchmarks.yml
```diff
@@ -0,0 +1,44 @@
+# Build and then run all benchmarks, on all supported archs.
+name: Build and run all benchmarks
+
+on:
+  workflow_dispatch:
+    inputs:
+      build-type:
+        required: false
+        default: Release
```
broskoTT, line 9: Add valid options, it was done in some other workflow like that

.github/workflows/build-and-run-all-benchmarks.yml
```diff
@@ -0,0 +1,44 @@
...
+  build-tests:
+    secrets: inherit
+    strategy:
+      fail-fast: false
+      matrix:
+        ubuntu-docker-version: [
+          'ubuntu-22.04',
+          'ubuntu-24.04',
+        ]
+    uses: ./.github/workflows/build-tests.yml
```
broskoTT, line 21: Is ubench built by default always?

tests/microbenchmark/benchmarks/tlb/test_tlb.cpp
```diff
@@ -75,9 +75,6 @@ TEST(MicrobenchmarkTLB, TLBDynamicDram) {
         8 * one_mb,
         16 * one_mb,
         32 * one_mb,
-        64 * one_mb,
```
broskoTT, line 78: Does this take long time? Why is it removed?
In general, maybe you can just add a comment above this anyway to explain how the sizes have been chosen.
Same for the removal below.

broskoTT (approved):

Issue #847, opened 2025-05-12
Perf results print format


pjanevskiTT:
Follow up from comment https://github.com/tenstorrent/tt-umd/pull/837#discussion_r2081106311

PR #1081, opened 2025-07-14
Print UMD perf in markdown style

pjanevskiTT:
### Issue

#847 

### Description

Add UMD perf print format. Idea is that UMD perf results are printed in the format such that results can be pasted from logs directly to README files and form a table for the results. This PR also introduces common code for UMD benchmarks where some utility that is useful for all the benchmarks is going to be.

### List of the changes

- Add microbenchmark utils functions
- Print results in desired format in PCIe DMA benchmark
- Print results in desired format in TLB benchmark
- Print results in desired format in IOMMU benchmark

### Testing
manual testing

### API Changes
/

### DIFF
--- a/tests/microbenchmark/common/microbenchmark_utils.h
+++ b/tests/microbenchmark/common/microbenchmark_utils.h
@@ -15,7 +15,10 @@ namespace tt::umd::test::utils {
 
 /**
  * Prints a table in Markdown format. Headers are printed as the first row, followed by a separator row,
- * and then the data rows. Headers length must match the length of each row.
+ * and then the data rows. Headers length must match the length of each row. Example:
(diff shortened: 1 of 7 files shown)


tests/microbenchmark/common/microbenchmark_utils.h
```diff
@@ -0,0 +1,43 @@
...
+#include <sstream>
+#include <string>
+#include <vector>
+
+namespace tt::umd::test::utils {
+
+/**
+ * Prints a table in Markdown format. Headers are printed as the first row, followed by a separator row,
+ * and then the data rows. Headers length must match the length of each row.
+ *
```
broskoTT, line 19: Consider adding an example of how this looks like here

broskoTT (approved):

Issue #853, opened 2025-05-12
Support for third party UMD


broskoTT:
Do another iteration of integration guide, and over header files, to make sure our guide is up to date.
Also provude further support if needed.
Raised after a chat with Antonio

PR #1084, opened 2025-07-14
Unify discovery functions under cluster_descriptor

broskoTT:
### Issue
Related to #853 

### Description
Discovering cluster architecture and detecting mmio devices was done through PCIDevice, then everything else through higher level concepts.
This will allow tt_metal to use only ClusterDescriptor, example here: https://github.com/tenstorrent/tt-metal/blob/main/tt_metal/llrt/get_platform_architecture.hpp

### List of the changes
- Added detect_cluster_architecture and detect_mmio_devices to ClusterDescriptor. 

### Testing
No testing

### API Changes
There are API changes in this PR, which add convenience functions, to be used in metal

### DIFF
--- a/device/api/umd/device/tt_cluster_descriptor.h
+++ b/device/api/umd/device/tt_cluster_descriptor.h
@@ -40,6 +40,16 @@ public:
     /* Construction related functions. */
 
     /**
+     * Helper function to detect cluster's architecture. Assumes whole cluster has the same architecture.
+     */
+    static tt::ARCH detect_cluster_architecture();
+
+    /**
+     * Helper function to detect all PCIe chips without running full topology discovery.
+     */
+    static std::vector<chip_id_t> detect_mmio_devices();
+
+    /**
      * Serializes the cluster descriptor to a YAML string.
      */
     std::string serialize() const;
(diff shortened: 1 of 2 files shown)


device/api/umd/device/tt_cluster_descriptor.h
```diff
@@ -39,6 +39,16 @@ class tt_ClusterDescriptor {
 public:
     /* Construction related functions. */
 
+    /**
+     * Helper function to detect cluster's architecture. Assumes whole cluster has the same architecture.
+     */
+    static tt::ARCH detect_cluster_architecture();
```
pjanevskiTT, line 45: I think this function should be moved to Cluster. I would like cluster descriptor to stay just with structures that are description of the cluster, hence the name. It shouldn't have any low level logic in my opinion

device/api/umd/device/tt_cluster_descriptor.h
```diff
@@ -39,6 +39,16 @@ class tt_ClusterDescriptor {
...
+
+    /**
+     * Helper function to detect all PCIe chips without running full topology discovery.
+     */
+    static std::vector<chip_id_t> detect_mmio_devices();
```
pjanevskiTT, line 50: I don't generally see a point of this function. I think PCIDevice enumeration function is clear enough, it can be used if needed. Maybe what we can do here is return mmio device IDs directly from cluster descriptor, but certainly not call PCIDevice enumeration

Issue #851, opened 2025-05-12
Figure out proper way of TTDevice accessing ARC core


pjanevskiTT:
Follow up from https://github.com/tenstorrent/tt-umd/pull/823#discussion_r2079180876

pjanevskiTT:
closed by #1086

PR #1086, opened 2025-07-14
Arc message instead of generic write (use AXI for WH)

nbuncicTT:
### Issue

#851

### Description

Add new API in tt-device which is used for accessing ARC core

### List of the changes

- ARC core on wormhole accessed through AXI
- ARC core on blackhole accessed through NOC
- Refactored BlackholeArcMessageQueue so that is doesn't have arc_core, moved this to Blackhole tt device

### Testing

Manual

### API Changes

/

### DIFF
--- a/device/tt_device/wormhole_tt_device.cpp
+++ b/device/tt_device/wormhole_tt_device.cpp
@@ -363,7 +363,7 @@ void WormholeTTDevice::dma_d2h_zero_copy(void *dst, uint32_t src, size_t size) {
 }
 
 void WormholeTTDevice::read_from_arc(void *mem_ptr, uint64_t arc_addr_offset, size_t size) {
-    if ((arc_addr_offset < 0) || (arc_addr_offset > wormhole::ARC_XBAR_ADDRESS_END)) {
+    if (arc_addr_offset > wormhole::ARC_XBAR_ADDRESS_END) {
         throw std::runtime_error("Address is out of ARC XBAR address range");
     }
     auto result = bar_read32(wormhole::ARC_APB_BAR0_XBAR_OFFSET_START + arc_addr_offset);
@@ -371,7 +371,7 @@ void WormholeTTDevice::read_from_arc(void *mem_ptr, uint64_t arc_addr_offset, si
 }
 
 void WormholeTTDevice::write_to_arc(const void *mem_ptr, uint64_t arc_addr_offset, size_t size) {
-    if ((arc_addr_offset < 0) || (arc_addr_offset > wormhole::ARC_XBAR_ADDRESS_END)) {
+    if (arc_addr_offset > wormhole::ARC_XBAR_ADDRESS_END) {
         throw std::runtime_error("Address is out of ARC XBAR address range");
     }
     bar_write32(


device/api/umd/device/tt_device/blackhole_tt_device.h
```diff
@@ -39,6 +39,10 @@ class BlackholeTTDevice : public TTDevice {
 
     void dma_d2h_zero_copy(void *dst, uint32_t src, size_t size) override;
 
+    void read_from_arc(void *mem_ptr, uint64_t addr, size_t size) override;
```
pjanevskiTT, line 42: Please add a comment to this function explaining the API. I would rename addr to offset, since the point of the API is to read/write starting from offset 0 on the ARC core

device/api/umd/device/tt_device/tt_device.h
```diff
@@ -117,6 +117,11 @@ class TTDevice {
     virtual void read_from_device(void *mem_ptr, tt_xy_pair core, uint64_t addr, uint32_t size);
     virtual void write_to_device(const void *mem_ptr, tt_xy_pair core, uint64_t addr, uint32_t size);
 
+    // Read/write functions that will send the arc message. It is used to read/write the data needed
+    // to get the information to or form cluster the arc core.
+    virtual void read_from_arc(void *mem_ptr, uint64_t addr, [[maybe_unused]] size_t size) = 0;
```
pjanevskiTT, line 122: you can document it here, just with a bit more detail

device/api/umd/device/wormhole_implementation.h
```diff
@@ -233,6 +233,9 @@ static constexpr uint32_t GRID_SIZE_Y = 12;
 
 static constexpr uint32_t ARC_MSG_COMMON_PREFIX = 0xAA00;
 
+static constexpr uint32_t ARC_BAR0_ADDRESS_START = 0x1FE00000;
```
pjanevskiTT, line 236: I would renam this to ARC_BAR0_ADDRESS_OFFSET, since it represents offset in the BAR0, absolute address is going to be something random

device/api/umd/device/wormhole_implementation.h
```diff
@@ -233,6 +233,9 @@ static constexpr uint32_t GRID_SIZE_Y = 12;
...
+static constexpr uint32_t ARC_BAR0_ADDRESS_END = 0x1FFFFFFF;
```
pjanevskiTT, line 237: same as above

device/wormhole/wormhole_arc_messenger.cpp
```diff
@@ -50,27 +50,26 @@ uint32_t WormholeArcMessenger::send_message(
     uint32_t fw_arg = arg0 | (arg1 << 16);
     int exit_code = 0;
 
-    tt_device->write_to_device(
+    tt_device->write_to_arc(
         &fw_arg,
-        arc_core,
-        wormhole::ARC_RESET_SCRATCH_ADDR + wormhole::ARC_SCRATCH_RES0_OFFSET * sizeof(uint32_t),
+        wormhole::ARC_RESET_SCRATCH_OFFSET + wormhole::ARC_SCRATCH_RES0_OFFSET * sizeof(uint32_t),
```
pjanevskiTT, line 55: as we talked offline, this won't work for remote chip on Wormhole. It should work when API is changed to accept offset from 0 and then WormholeTTDevice and RemoteWormholeTTDevice each implement it's own logic for reading/writing ARC

device/api/umd/device/wormhole_implementation.h
nbuncicTT: Renamed to ``ARC_BAR0_ADDRESS_START``

device/api/umd/device/tt_device/tt_device.h
```diff
@@ -117,6 +117,13 @@ class TTDevice {
     virtual void read_from_device(void *mem_ptr, tt_xy_pair core, uint64_t addr, uint32_t size);
     virtual void write_to_device(const void *mem_ptr, tt_xy_pair core, uint64_t addr, uint32_t size);
 
+    // Read/write functions that will send the arc message. It is used to read/write the data needed
+    // to get the information to or form cluster the arc core. These functions on Wormhole will use the
+    // AXI interface to read/write the data if the chip is local/PCIe, while the remote chip will use the
+    // the NOC interface to read/write the data. Blackhole for now, will only use the NOC interface to read/write data.
```
pjanevskiTT, line 123: either open new issue for Blackhole access or add a comment in the original issue and don't close it

device/api/umd/device/wormhole_implementation.h
```diff
@@ -233,8 +233,8 @@ static constexpr uint32_t GRID_SIZE_Y = 12;
 
 static constexpr uint32_t ARC_MSG_COMMON_PREFIX = 0xAA00;
 
-static constexpr uint32_t ARC_RESET_SCRATCH_OFFSET = 0x1FF30060;
-static constexpr uint32_t ARC_RESET_ARC_MISC_CNTL_OFFSET = 0x1FF30100;
+static constexpr uint32_t ARC_APB_BAR0_XBAR_ADDRESS_START = 0x1FF00000;
```
pjanevskiTT, line 236: again, I think this should contain word OFFSET instead of ADDRESS because it's offset within a BAR

device/tt_device/remote_wormhole_tt_device.cpp
```diff
@@ -28,4 +30,18 @@ LocalChip *RemoteWormholeTTDevice::get_local_chip() { return local_chip_; }
 
 RemoteCommunication *RemoteWormholeTTDevice::get_remote_communication() { return remote_communication_.get(); }
 
+void RemoteWormholeTTDevice::read_from_arc(void *mem_ptr, uint64_t arc_addr_offset, size_t size) {
+    if ((arc_addr_offset < 0) || (arc_addr_offset > wormhole::ARC_XBAR_ADDRESS_END)) {
```
pjanevskiTT, line 34: `arc_addr_offset` is `uint64_t`. Can the first part of the condition ever be true?

device/tt_device/blackhole_tt_device.cpp
```diff
@@ -212,6 +214,14 @@ void BlackholeTTDevice::dma_d2h_zero_copy(void *dst, uint32_t src, size_t size)
     throw std::runtime_error("D2H DMA is not supported on Blackhole.");
 }
 
+void BlackholeTTDevice::read_from_arc(void *mem_ptr, uint64_t arc_addr_offset, size_t size) {
```
pjanevskiTT, line 217: seems BlackholeTTDevice API is not same as Wormhole in terms of accepting absolute NOC address, instead of ARC offset. Why not simply accept the offsets here as well and then add some fixed NOC offset in this function, same what is done for WormholeTTDevice. Then when we eventually add support for reading/writing stuff over BAR for Blackhole, we don't have the change the code that used BlackholeTTDevice, just this function

nbuncicTT: Sure, makes sense

device/api/umd/device/tt_device/tt_device.h
```diff
@@ -117,6 +118,34 @@ class TTDevice {
...
+    /**
+     * Read function that will send read message to the ARC core.
+     *
+     * @param mem_ptr pointer to memory which will receive the data
+     * @param arc_addr_offset address offset in ARC core
+     * @param size number of bytes
+     *
+     * NOTE: These functions on Wormhole will use the
+     * AXI interface to read/write the data if the chip is local/PCIe, while the remote chip will use the
+     * the NOC interface to read/write the data. Blackhole for now, will only use the NOC interface to read/write data.
```
broskoTT, line 130: explain why bh uses only noc

broskoTT (approved):

PR #1089, opened 2025-07-15
Add support for isolating chips on 6U

pjanevskiTT:
### Issue

Be able to isolate just specific chips on 6U galaxy

### Description

Add support for isolating specific chips on 6U galaxy. We hack the topology discovery in a way that board_id during topology discovery is actually equal to asic id so you can isolate even a single chip.

### List of the changes

- In 6U case, return remote asic ID as remote board ID
- in 6U case, return local asic ID as local board ID 

### Testing

Manual testing on 6U

### API Changes
/

### DIFF
--- a/device/api/umd/device/topology_discovery.h
+++ b/device/api/umd/device/topology_discovery.h
@@ -45,21 +45,21 @@ private:
 
     bool is_pcie_chip_id_included(int pci_id) const;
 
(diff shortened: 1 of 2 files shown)


broskoTT (approved):
I've added a comment explaining the rationale. I also checked that board id is still properly reported in cluster descriptor (haven't tested though).

Issue #507, opened 2025-02-05
Telemetry update only dynamic entries


pjanevskiTT:
Telemetry entries are mostly static, which means that we need to read it only once from the device. Have some mapping of which entries we should read more times and which are static, so we don't go to the device if it not needed

pjanevskiTT:
closed by #1095

PR #1095, opened 2025-07-16
Telemetry update only dynamic entries

nbuncicTT:
### Issue
#507 

### Description
Read static entries only once for WH and BH and store them for later re-use.

### List of the changes
- Added hashed map of static entries for WH and BH to read out static values at time of initialization

### Testing
Manual

### API Changes
/


device/api/umd/device/wormhole_arc_telemetry_reader.h
```diff
@@ -26,13 +27,35 @@ class WormholeArcTelemetryReader : public ArcTelemetryReader {
...
     uint64_t telemetry_base_noc_addr;
 
     // During initialization of telemetry, if the NOC0 is hung then we need to read the telemetry values from NOC1.
     const tt_xy_pair arc_core = !umd_use_noc1 ? wormhole::ARC_CORES_NOC0[0]
                                               : tt_xy_pair(
                                                     wormhole::NOC0_X_TO_NOC1_X[wormhole::ARC_CORES_NOC0[0].x],
                                                     wormhole::NOC0_Y_TO_NOC1_Y[wormhole::ARC_CORES_NOC0[0].y]);
+
+    bool static_telemetry_entries_initialized{false};
+    std::unordered_map<uint8_t, uint32_t> static_telemetry_entries{
```
pjanevskiTT, line 41: I would consider redesigning this in a way that this becomes (unordered) set. It can contain just the tags. Then you don't need the map to store the values. These are already in telemetry_values.

read_entry could then just check if tag is static and if yes, return value from telemetry_values. Otherwise it should go to device

pjanevskiTT, line 41: static entries look correct to me :)

device/wormhole/wormhole_arc_telemetry_reader.cpp
```diff
@@ -59,4 +65,11 @@ bool WormholeArcTelemetryReader::is_entry_available(const uint8_t telemetry_tag)
     return telemetry_tag >= 0 && telemetry_tag < wormhole::TELEMETRY_NUMBER_OF_TAGS;
 }
 
+void WormholeArcTelemetryReader::read_static_telemetry_entries() {
+    for (auto& [telemetry_tag, telemetry_value] : static_telemetry_entries) {
+        telemetry_value = read_entry(telemetry_tag);
```
pjanevskiTT, line 70: this is not a safe thing to do. Telemetry tags represent set of tags that can be available in the telemetry. However, based on FW version some of those might not be in the table where this code will throw. Safe thing to do is check with is_entry_available if entry is there. However if you redesign the map into a set for static telemetry tags this code might go away

device/api/umd/device/blackhole_arc_telemetry_reader.h
```diff
@@ -7,6 +7,7 @@
```
pjanevskiTT, line 7: as a general comment, can you maybe measure the time before/after this change for Cluster construction on some configs. Nothing official, I am just interested

device/api/umd/device/wormhole_arc_telemetry_reader.h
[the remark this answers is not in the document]
nbuncicTT: This is true on blackhole, but not on wormhole? I can maintain a set in blackhole, and if I understood, you would prefer this on wormhole also - with the telemetry_values there too

PR #1101, opened 2025-07-17
Telemetry update only dynamic entries - with new wormhole in mind

nbuncicTT:
### Issue
#507 

### Description
Read static entries only once in ARC telemetry reader and store them for later re-use.

### List of the changes
- Added hashed map of static entries in ARC telemetry reader to read out static values at time of initialization

### Testing
Manual

### API Changes
/

### DIFF
--- a/device/arc_telemetry_reader.cpp
+++ b/device/arc_telemetry_reader.cpp
@@ -45,9 +45,6 @@ void ArcTelemetryReader::initialize_telemetry() {
 
         telemetry_values.insert({tag_val, telemetry_data[offset_val]});
         telemetry_offset.insert({tag_val, offset_val});
-        if (possible_static_entries.find(tag_val) != possible_static_entries.end()) {
-            static_entries.insert(tag_val);
-        }
     }
 }
 
--- a/device/arc_telemetry_reader.cpp
+++ b/device/arc_telemetry_reader.cpp
@@ -45,6 +45,9 @@ void ArcTelemetryReader::initialize_telemetry() {
 
         telemetry_values.insert({tag_val, telemetry_data[offset_val]});
         telemetry_offset.insert({tag_val, offset_val});
+        if (possible_static_entries.find(tag_val) != possible_static_entries.end()) {
+            static_entries.insert(tag_val);
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


device/arc_telemetry_reader.cpp
```diff
@@ -55,6 +58,10 @@ uint32_t ArcTelemetryReader::read_entry(const uint8_t telemetry_tag) {
             telemetry_tag));
     }
 
+    if (static_entries.find(telemetry_tag) != static_entries.end()) {
+        return telemetry_values[telemetry_tag];
```
broskoTT, line 62: This is already done at the end of the function.
Maybe switch the logic, if not static entry update the value? In both cases return telemetry_values

device/arc_telemetry_reader.cpp
```diff
@@ -55,6 +58,10 @@ uint32_t ArcTelemetryReader::read_entry(const uint8_t telemetry_tag) {
             telemetry_tag));
     }
 
+    if (static_entries.find(telemetry_tag) != static_entries.end()) {
```
broskoTT, line 61: I don't think you need both collections.
You could just have a check here which is 
`if possible_static_entries.find(telemetry_tag) != possible_static_entries.end() && telemetry_values.find(telemetry_tag) != telemetry_values.end()`

Then you can remove the code added to `initialize_telemetry`. Then you can also remove static_entries and then maybe rename possible_static_entries to static_entries.

Also, if you accept the suggestion below, then the condition should be reversed :)

broskoTT (approved):

device/arc_telemetry_reader.cpp
nbuncicTT: I agree, thanks!

device/arc_telemetry_reader.cpp
nbuncicTT: I actually prefer premature returns, as they are more readable for me - here is a nice article on this topic: https://arne-mertz.de/2016/12/early-return/

device/arc_telemetry_reader.cpp
```diff
@@ -55,6 +55,10 @@ uint32_t ArcTelemetryReader::read_entry(const uint8_t telemetry_tag) {
...
+        return telemetry_values[telemetry_tag];
```
pjanevskiTT, line 59: nit: use .at() instead of []. [] Will create new entry even if the telemetry_tag one is not present, that's now what we want here

Issue #1054, opened 2025-07-07
Warm reset


broskoTT:
Support warm reset through UMD.
By a popular request from metal and exalens

pjanevskiTT:
pasting comments from slack

Yes I have a few usecases that all center around the same idea:
I'm a dev, working on kernels, I know I could be hacking something that might hang.  I'd much rather have the system just throw an error + the currently running kernel to yell at me, "you hung the device!!", vs waiting, seeing something's probably hung, shoot yeah it's hung, go to my other terminal window, run tt-smi, wait to make sure it retrained, make a change and repeat.
I'm a cloud user that wants to play with ttnn and I spin up a jupyter notebook and I'm developing a model.  For some reason, my instance is hanging:
Did the server crap out?
Did my wifi stop working?
Oh maybe the cloud system is timed out?
Oh the device is hung hmm
ssh to my cloud instance and run tt-smi reset
It'd be much better if I just got an exception that told me the device hung and then it auto-reset things into a working state.
I'm a developer that wants to deploy a vllm instance using TTNN.  I discover that 1 in 1000 runs hang the device, it'd be much much much better if I could just file a bug and make my script robust and handle a hang exception and retry instead of having to manage multiprocessing on the side.

pjanevskiTT:
@broskoTT @nbuncicTT anything else to be done here?

PR #1111, opened 2025-07-21
Add DRAM channel training API to TT Device

nbuncicTT:
### Issue
#1054

### Description
- Add DRAM bank training per channel API to TT Device, because it is a part of an initialization sequence which is needed after a warm reset - and may be used in these situations

### List of the changes
- Made the DRAM status API pure virtual, and provided implementations for WH and BH
- Added DRAM banks number in architecture implementations (and also for WH and BH)
- Added NOC0 DRAM coordinates in architecture implementations
- Added new API per channel in TT Device (for waiting the training)

### Testing
Manual

### API Changes
/


device/chip/chip.cpp
```diff
@@ -72,49 +72,7 @@ void Chip::wait_eth_cores_training(const uint32_t timeout_ms) {
...
-
-        auto end = std::chrono::system_clock::now();
-        auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
-        if (duration.count() > timeout_ms) {
-            throw std::runtime_error(fmt::format("DRAM training timed out after {} ms", timeout_ms));
-            break;
-        }
-    }
-}
+void Chip::wait_dram_cores_training(const uint32_t timeout_ms) { get_tt_device()->wait_dram_core_training(); }
```
broskoTT, line 75: This is now trivial ,so just move it directly to wait_chip_to_be_ready

broskoTT (approved):

pjanevskiTT (changes requested):
Nenad is currently investigating why our Cluster constructor doesn't work after warm reset sequence. I don't think this is "needed" to be done after reset sequence,reset sequence can happen even without it.  It's more of a way to ensure Chip is safe, and that's what our chip is supposed to do. I am against moving this to TTDevice

broskoTT:
Ass discussed offline, let's modify this to have an API that does wait for dram to train per core

device/api/umd/device/tt_device/tt_device.h
```diff
@@ -203,6 +203,8 @@ class TTDevice {
 
     virtual void wait_eth_core_training(const tt_xy_pair eth_core, const uint32_t timeout_ms = 60000) = 0;
 
+    void wait_dram_core_training(const tt_xy_pair eth_core, const uint32_t timeout_ms = 60000);
```
pjanevskiTT, line 206: it's not clear for me if eth_core should be in NOC0/1. I mean I can see it's NOC0 from cpp file, but I think this would benefit from a comment

pjanevskiTT, line 206: Would it make sense to maybe change this to accept DRAM channel number here? Maybe the API would be more intuitive. Then Chip class can take care of calling this appropriately while taking into consideration harvesting mask? What do you think about it?

nbuncicTT: Yeah, I agree, for me it was kind of weird to use the core coords and then to reverse engineer the DRAM channel/bank number, so I'll change it to the channel number. Thanks

Issue #600, opened 2025-03-11
Deterministic logical ID


broskoTT:
It was reported that logical IDs can change from run to run. This was specific for 6U wormhole galaxy, but it might be a deeper problem. 
Investigate code and change if needed so that our IDs always map the same to the chips on a system which doesn't change its configuration.

broskoTT:
Allan Liu
  Today at 3:29 PM
[@Bojan Roško](https://tenstorrent.slack.com/team/U06C36SEJ2V)
 the chip ids are deterministic from UMD side, it's the PCI enumeration that happens when you do a reset of the UBB that is the problem.


Bojan Roško
  Today at 3:37 PM
Aha, got it. Would you then consider an improvement if the logical chip ids are deterministic even if the pci device enumeration is random? maybe sort them by some board/asic id?


Pavle Janevski
  [1 hour ago](https://tenstorrent.slack.com/archives/C07L16Z59AP/p1741704134839929?thread_ts=1741641337.938589&cid=C07L16Z59AP)
we have something similar captured in the issue. We can open a separate one if it requires more thinking/discussion. https://github.com/tenstorrent/tt-umd/issues/137#issuecomment-2706549717


Pavle Janevski
  1 hour ago
advantage of 6u over old wh boards is that we have 64-bit unique id that we can read from the chip now
New


Allan Liu
  29 minutes ago
[@Bojan Roško](https://tenstorrent.slack.com/team/U06C36SEJ2V)
 I think what you suggest would be nice to have. UMD physical ids map to some board/asic id, regardless of the pci enumeration

broskoTT:
Can be implemented after #871

pjanevskiTT:
closed by #1112

PR #1112, opened 2025-07-21
Implement deterministic chip IDs in Topology

pjanevskiTT:
### Issue

#600 

### Description

Implement deterministic chip IDs in Topology discovery. Code has previously relied on PCI device enumeration, where now it will rely on unique chip IDs in order to assing logical IDs. These should stay consistent across reset.

### List of the changes

- Use ASIC id as key in maps in Topology Discovery
- Have separate struct for remote ETH connections in order to remap properly

### Testing

Manual testing on all configs

### API Changes
/

### DIFF
--- a/device/api/umd/device/topology_discovery.h
+++ b/device/api/umd/device/topology_discovery.h
@@ -102,20 +102,19 @@ private:
     // eth_core should be in physical (NOC0) coordinates.
     std::unique_ptr<RemoteChip> create_remote_chip(Chip* chip, tt_xy_pair eth_core, Chip* gateway_chip);
 
-    Chip* get_chip(const chip_id_t chip_id);
+    Chip* get_chip(const uint64_t asic_id);
 
-    std::map<chip_id_t, std::unique_ptr<Chip>> chips_to_discover;
-    std::map<chip_id_t, std::unique_ptr<Chip>> chips;
+    std::map<uint64_t, std::unique_ptr<Chip>> chips_to_discover;
+    std::map<uint64_t, std::unique_ptr<Chip>> chips;
 
-    std::unordered_map<uint64_t, chip_id_t> asic_id_to_chip_id;
+    std::unordered_map<uint64_t, eth_coord_t> eth_coords;
 
-    std::unordered_map<chip_id_t, eth_coord_t> eth_coords;
+    std::vector<std::pair<std::pair<uint64_t, uint32_t>, std::pair<uint64_t, uint32_t>>> ethernet_connections;
 
-    std::vector<std::pair<std::pair<chip_id_t, uint32_t>, std::pair<chip_id_t, uint32_t>>> ethernet_connections;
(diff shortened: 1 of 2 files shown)


device/topology_discovery.cpp
```diff
@@ -356,13 +352,22 @@ void TopologyDiscovery::fill_cluster_descriptor_info() {
     }
```
broskoTT, line 352: Previous loop already has both asic_id and current_chip_id, you don't need separate loop to fill up chip_unique_ids

device/topology_discovery.cpp
```diff
@@ -221,39 +221,33 @@ void TopologyDiscovery::discover_remote_chips() {
...
 
-        active_eth_channels_per_chip.emplace(current_chip_id, std::set<uint32_t>());
+        active_eth_channels_per_chip.emplace(current_chip_asic_id, std::set<uint32_t>());
         std::vector<CoreCoord> eth_cores =
             chip->get_soc_descriptor().get_cores(CoreType::ETH, umd_use_noc1 ? CoordSystem::NOC1 : CoordSystem::NOC0);
         TTDevice* tt_device = chip->get_tt_device();
 
-        uint64_t current_chip_asic_id = get_asic_id(chip);
-
         uint32_t channel = 0;
```
broskoTT, line 251: Why do we rely on eth_cores being given to us in the exact order of eth_channel?
We should be able to get the eth_channel from an eth_core, or vice versa

device/api/umd/device/topology_discovery.h
```diff
@@ -102,29 +102,28 @@ class TopologyDiscovery {
     // eth_core should be in physical (NOC0) coordinates.
```
broskoTT, line 102: Wondering why the tests on llmbox fail...

device/topology_discovery.cpp
```diff
@@ -318,16 +307,23 @@ void TopologyDiscovery::discover_remote_chips() {
...
 
 void TopologyDiscovery::fill_cluster_descriptor_info() {
-    for (const auto& [current_chip_id, chip] : chips) {
+    std::map<uint64_t, chip_id_t> asic_id_to_chip_id;
+    chip_id_t chip_id = 0;
+    for (const auto& [current_chip_asic_id, chip] : chips) {
+        asic_id_to_chip_id.emplace(current_chip_asic_id, chip_id++);
+    }
+
+    for (const auto& [current_chip_asic_id, chip] : chips) {
```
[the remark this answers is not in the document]
pjanevskiTT: done

device/topology_discovery.cpp
pjanevskiTT: https://github.com/tenstorrent/tt-umd/pull/1120 did it in separate PR

device/api/umd/device/topology_discovery.h
pjanevskiTT: passed on llmbox 🙌

PR #1123, opened 2025-07-22
Remove dead CoordinateManager code

pjanevskiTT:
### Issue

There is some dead code in CoordinateManager class.

### Description

Remove dead code from CoordinateManager class.

### List of the changes

- Remove dead code from CoordinateManager class.

### Testing
CI

### API Changes
/

### DIFF
--- a/device/api/umd/device/coordinate_manager.h
+++ b/device/api/umd/device/coordinate_manager.h
@@ -61,9 +61,6 @@ public:
     // Harvesting mask is reported by hardware in the order of physical layout. This function returns a more suitable
     // representation in logical order: Bit 0 being set means the first row in NOC0 coords is harvested.
     static uint32_t shuffle_tensix_harvesting_mask(tt::ARCH arch, uint32_t tensix_harvesting_physical_layout);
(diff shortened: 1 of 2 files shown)


device/api/umd/device/coordinate_manager.h
```diff
@@ -61,9 +61,6 @@ class CoordinateManager {
     // Harvesting mask is reported by hardware in the order of physical layout. This function returns a more suitable
     // representation in logical order: Bit 0 being set means the first row in NOC0 coords is harvested.
     static uint32_t shuffle_tensix_harvesting_mask(tt::ARCH arch, uint32_t tensix_harvesting_physical_layout);
-    // TODO: This function should be removed once the corresponding API is removed from Cluster.
-    static uint32_t shuffle_tensix_harvesting_mask_to_noc0_coords(
```
pjanevskiTT, line 65: metal build fails, investigate how they use this function, maybe they can switch to the one above or not use it at all

pjanevskiTT:
We can address this later, closing for now...

PR #1125, opened 2025-07-24
Warm reset implementation

nbuncicTT:
### Issue
#1054 

### Description

Implemented warm reset for Blackhole and Wormhole

### List of the changes

- Added Warm reset class which implements warm reset functionality through:
    - IOCTL
    - ARC messages (via AXI)
- Added PCI device static functions for reading byte command, doing reset, etc.
- Added getting DRAM banks which is architecture dependent
- Added two tests, one through cluster and second through tt device

### Testing

Manual

### API Changes

/


device/pcie/pci_device.cpp
```diff
@@ -175,6 +197,31 @@ static PciDeviceInfo read_device_info(int fd) {
         get_physical_slot_for_pcie_bdf(get_pci_bdf(info.out.pci_domain, bus, dev))};
 }
 
+static void reset_devices(uint32_t flags) {
```
broskoTT, line 200: nit: reset_device?

device/tt_device/tt_device.cpp
```diff
@@ -430,4 +430,6 @@ semver_t TTDevice::fw_version_from_telemetry(const uint32_t telemetry_data) cons
     return semver_t(major, minor, 0);
 }
 
+bool TTDevice::wait_arc_post_reset(const uint32_t timeout_ms) { return false; }
```
broskoTT, line 433: Can we make this pure virtual? Or is this a problem with other functions as well?

device/tt_device/wormhole_tt_device.cpp
```diff
@@ -487,4 +487,109 @@ WormholeTTDevice::EthAddresses WormholeTTDevice::get_eth_addresses(const uint32_
         erisc_remote_eth_id_offset};
 }
 
+bool WormholeTTDevice::wait_arc_post_reset(const uint32_t timeout_ms) {
```
broskoTT, line 490: Is all of this not valid for blackhole?

device/tt_device/wormhole_tt_device.cpp
```diff
@@ -487,4 +487,109 @@ WormholeTTDevice::EthAddresses WormholeTTDevice::get_eth_addresses(const uint32_
...
+
+    // DMA request address
+    static constexpr uint64_t ARC_CSM_ARC_PCIE_DMA_REQUEST = 0x1fef84d4;
+
+    uint32_t bar_read_arc_reset_scratch_status =
+        bar_read32(wormhole::ARC_APB_BAR0_XBAR_OFFSET_START + wormhole::ARC_RESET_SCRATCH_STATUS_OFFSET);
+
+    uint32_t bar_read_arc_post_code = bar_read32(architecture_impl_->get_arc_reset_scratch_offset());
+
+    uint32_t bar_read_arc_csm_pcie_dma_request = bar_read32(ARC_CSM_ARC_PCIE_DMA_REQUEST);
```
broskoTT, line 522: Why not put this at the beginning of while loop, and remove the "refresh" part at the end of the loop?

device/warm_reset.cpp
```diff
@@ -0,0 +1,185 @@
...
+                refclk_current[i]);
+        }
+    }
+
+    if (reset_ok) {
+        log_info(tt::LogSiliconDriver, "Reset successfully completed.");
+    }
+}
+
+uint64_t WarmReset::get_refclk_counter(TTDevice* tt_device) {
```
broskoTT, line 171: Does it make sense if this function is moved to ttdevice?

tests/api/test_cluster.cpp
```diff
@@ -504,6 +505,73 @@ TEST(TestCluster, TestClusterAICLKControl) {
...
+    hanged_tt_device->write_to_device(data.data(), {15, 15}, 0, data.size());
+
+    // TODO: Remove this check when it is figured out why there is no hang detected on Blackhole.
+    if (arch == tt::ARCH::WORMHOLE_B0) {
+        EXPECT_THROW(hanged_tt_device->detect_hang_read(), std::runtime_error);
+    }
+
+    WarmReset::warm_reset();
+
+    cluster.reset();
```
broskoTT, line 538: hmmm... This is debatable whether destructors should be call pre-reset or post-reset.
I'm fine with it if you leave it like this for now

tests/api/test_tt_device.cpp
```diff
@@ -111,6 +111,64 @@ TEST(ApiTTDeviceTest, TTDeviceMultipleThreadsIO) {
...
+
+    // TODO: Remove this check when it is figured out why there is no hang detected on Blackhole.
+    if (tt_device->get_arch() == tt::ARCH::WORMHOLE_B0) {
+        EXPECT_THROW(tt_device->detect_hang_read(), std::runtime_error);
+    }
+
+    WarmReset::warm_reset();
+
+    // Make cluster so that topology discovery does chip detection
+    auto cluster = std::make_unique<Cluster>();
```
broskoTT, line 152: If this is supposed to be purely ttdevice test, don't use Cluster

broskoTT (approved):
Thanks for pushing this, I know it wasn't easy

tests/api/test_tt_device.cpp
broskoTT: Just add a comment why this is necessary

device/warm_reset.cpp
nbuncicTT: Well, only for those who can access the chip via PCIe, not sure if that makes sense in the inheritance hierarchy that we have now..

PR #1129, opened 2025-07-25
Convert telemetry tags to enums

broskoTT:
### Issue
Helping with the cause #842 

### Description
It makes more sense these are enums. This change will also make it easier to bind these through python

### List of the changes
- Change constexpr to enum values.
- Changes to accompany it

### Testing
No significant testing

### API Changes
There are API changes in case somebody wants to use telemetry but nobody does now.


nbuncicTT (approved):
Nice work

Issue #172, opened 2024-10-16
TT_THROW macro is broken


joelsmithTT:
```
int argument = 1;
TT_THROW("Message with argument {} and more message", argument);
```
does not work as expected.   This generates the following exception info:

```
Message with argument {} and more message
1
```

Desired behavior is to format the string.

broskoTT:
At one point we should refactor how this is done in UMD, possibly related to #517

uvelimirovicTT:
Closed by #1131

PR #1131, opened 2025-07-28
TT_THROW bad formatting fix

uvelimirovicTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/172

### Description
The TT_THROW macro was broken. It couldn't format messages properly. 
Example: 

int argument = 1;
TT_THROW("Message with argument {} and more message", argument);

prints: 
Message with argument {} and more message
1

It is fixed now and works as expected. 

### Testing
Added tests in tests/misc/test_assert.cpp. 
If in the future, a more complex formatting is added, additional tests can be implemented.

### DIFF
--- a/.vscode/default.launch.json
+++ b/.vscode/default.launch.json
@@ -1,0 +1,46 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "umd_api_test",
+            "type": "cppdbg",
+            "request": "launch",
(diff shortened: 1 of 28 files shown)


common/assert.hpp
```diff
@@ -35,12 +35,47 @@ std::ostream& operator<<(std::ostream& os, tt::OStreamJoin<A, B> const& join) {
...
+template <typename T>
+std::string to_string_safe(T const& t) {
+    std::stringstream ss;
+    ss << t;
+    return ss.str();
+}
+
+// Formating a message with a {} placeholder and a vector of arguments is done
+// this way to allow a more versatile spectrum of arguments.
+// The fmt::format() function is good, but it does not handle
```
broskoTT, line 47: Do you have a specific case where fmt::format doesn't work?
We use fmt library throughout our codebase, it would be nice to be consistent

PR #1140, opened 2025-07-31
Use PID instead of timestamp to make unique pipe name

mcraigheadTT:
### Issue
When running multiple simulations in parallel, the pipe names often collide.

### Description
Don't use second-granularity system time to make pipe name, as this is almost guaranteed to collide with multiple simulations run in parallel.  Use getpid() instead.

### List of the changes
Remote strftime() code and replace with getpid().

### Testing
Ran tt-metal samples locally on simulator

### API Changes
/

### DIFF
--- a/device/simulation/tt_simulation_host.cpp
+++ b/device/simulation/tt_simulation_host.cpp
@@ -24,18 +24,12 @@ tt_SimulationHost::tt_SimulationHost() {
     host_socket = std::make_unique<nng_socket>();
     host_dialer = std::make_unique<nng_dialer>();
 
-    // Get current date time string
-    std::time_t time = std::time(nullptr);
-    std::tm local_time = *std::localtime(&time);
-    char buffer[100];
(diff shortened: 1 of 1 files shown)


mcraigheadTT:
I thought about removing the username and relying entirely on getpid().  We could even go a step further and remove the NNG_SOCKET_NAME env var, whose main purpose right now is to work around this bug.  (I'd probably be inclined to remove both in a future change and give them a name containing tt-umd.)

Yes, PID can be reused (after rather long periods of time, esp since on newer systems PIDs are no longer 16 bits), but in a well-written simulator the pipe will be unlink()'d just moments after the client connects to it and in all cases where the connection fails.  So the pipe should only be left around in /tmp in extremely rare cases such as a "kill" of the simulator process during that very brief time window.  In practice I kill simulation runs with Ctrl-\ all the time and I have yet to see this happen even once.

pjanevskiTT (approved):

PR #1141, opened 2025-07-31
Improve PCI id filtering

pjanevskiTT:
### Issue

Improve PCI id filtering in UMD

### Description

Improve PCI id filtering in UMD by doing it in PCIDevice layer. UMD will actually completely ignore pci devices not specified in pci_target_devices so logic can be simplified in TopologyDiscovery

### List of the changes

- Add pci_target_devices to enumerate functions
- Filter pci target devices in PCIDevice layer
- Remove pci id filtering logic from TopologyDiscovery

### Testing

CI + manual testing on T3K

### API Changes
/

### DIFF
--- a/device/api/umd/device/pci_device.hpp
+++ b/device/api/umd/device/pci_device.hpp
@@ -11,6 +11,7 @@
 #include <map>
 #include <memory>
 #include <unordered_map>
+#include <unordered_set>
 #include <vector>
 
 #include "fmt/format.h"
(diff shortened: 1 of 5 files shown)


pjanevskiTT:
Think about introducing this as environment variable, it could help circumvent the issues like in https://github.com/tenstorrent/tt-metal/issues/25860#issuecomment-3146805424 and similar

pjanevskiTT:
Since this is making general improvement in enabling more robust PCI id filtering, I will check it in. Final solution we should work towards is having TT_VISIBLE_DEVICES env variable used by both metal and UMD

Issue #1091, opened 2025-07-15
Cannot create Cluster with Simulator SoC


dzivanovicTT:
If I try to create Cluster with Simulator SoC, like with argument `.sdesc_path = "/localdev/dzivanovic/tt-umd-simulators/build/versim-wormhole-b0/soc_descriptor.yaml"` I get exception `No core coordinate found at location: (0, 8, ETH, LOGICAL)`

Branch for reproduction: `dzivanovic/open_cluster_versim_simulator_soc`
Board: `bgd-lab-t3002: 4x N300`

Test output:
```
[2005] dzivanovic:bgd-lab-t3002-special-dzivanovic-for-reservation-7628 /localdev/dzivanovic/tt-umd > ./build/test/umd/api/api_tests --gtest_filter=*OpenAllSiliconChips*
Running main() from /localdev/dzivanovic/tt-umd/.cpmcache/googletest/96129d89f45386492ae46d6bb8c027bc3df5f949/googletest/src/gtest_main.cc
Note: Google Test filter = *OpenAllSiliconChips*
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from ApiClusterTest
[ RUN      ] ApiClusterTest.OpenAllSiliconChips
2025-07-15 14:03:43.770 | info     |   SiliconDriver | Opened PCI device 0; KMD version: 1.27.0; API: 1; IOMMU: disabled (pci_device.cpp:197)
2025-07-15 14:03:43.772 | info     |   SiliconDriver | Opened PCI device 1; KMD version: 1.27.0; API: 1; IOMMU: disabled (pci_device.cpp:197)
2025-07-15 14:03:43.774 | info     |   SiliconDriver | Opened PCI device 2; KMD version: 1.27.0; API: 1; IOMMU: disabled (pci_device.cpp:197)
2025-07-15 14:03:43.775 | info     |   SiliconDriver | Opened PCI device 3; KMD version: 1.27.0; API: 1; IOMMU: disabled (pci_device.cpp:197)
2025-07-15 14:03:43.792 | info     |   SiliconDriver | Harvesting mask for chip 3 is 0x280 (NOC0: 0x280, simulated harvesting mask: 0x0). (cluster.cpp:307)
2025-07-15 14:03:43.793 | info     |   SiliconDriver | Opened PCI device 3; KMD version: 1.27.0; API: 1; IOMMU: disabled (pci_device.cpp:197)
unknown file: Failure
[...]

PR #1143, opened 2025-08-01
Fix Cluster creation with Simulator-like SoC descriptor

nbuncicTT:
### Issue
#1091 

### Description
Introduces changes to handle various SoC Descriptors in `TopologyDiscovery` class.

### List of the changes
- Added SoC descriptor in constructor of Topology Discovery invocation for Wormhole
- Changed method in topology discovery to handle case where no ethernet cores are present
- Added test to accept simulation-like soc descriptor 

### Testing
Manual

### API Changes
/


tests/soc_descs/one_dram_one_tensix_no_eth_descriptor.yaml
```diff
@@ -0,0 +1,63 @@
+grid:
```
pjanevskiTT, line 1: nit: I think all our soc descriptor names start with architecture, so it might be useful to add it to beginning of the name here as well

tests/api/test_soc_descriptor.cpp
```diff
@@ -12,6 +12,12 @@
 
 using namespace tt::umd;
 
+TEST(SocDescriptor, OneDRAMOneTensixDescriptor) {
```
pjanevskiTT, line 15: I would just like to move this test out of `test_soc_descriptor.cpp `file. I think all tests in this file don't require silicon to run, so we could potentially move all the test from here to some test group that don't require silicon. Since this is not explicit either
1. Add a comment that this test requires silicon
2. Move this test out of this file, to test_cluster for example
I would prefer 2

tests/soc_descs/one_dram_one_tensix_no_eth_descriptor.yaml
```diff
@@ -0,0 +1,63 @@
...
+arc:
+  [ ]
+
+pcie:
+  [ ]
+
+dram:
+  [[0-0]]
+
+dram_views:
```
pjanevskiTT, line 14: this is metal specific section of yaml, we don't need it in our soc desc yamls. I know this is something that is in Djordje's soc desc, but I think we should delete it from yamls in our repo

tests/soc_descs/one_dram_one_tensix_no_eth_descriptor.yaml
```diff
@@ -0,0 +1,63 @@
...
+  [
+    {
+      channel: 0,
+      eth_endpoint: [0,0],
+      worker_endpoint: [0,0],
+      address_offset: 0
+    }
+  ]
+
+dram_view_size:
```
pjanevskiTT, line 24: same as above

pjanevskiTT (approved):

Issue #995, opened 2025-06-22
Design: Make a stonger distinction between layers in UMD


broskoTT:
Mainly, RemoteCommunication uses Chip.
Let's figure out what is actually needed for it, and not use the higher chip abstraction

broskoTT:
This is needed to be resolved to properly implement this diagram
https://docs.google.com/drawings/d/1szsbPhZ5XjXGXQ7954mq7R1kVrrYI8ldWNzeqbsu0V0/edit

broskoTT:
Complete with removing LocalChip from RemoteTTDevice

PR #1148, opened 2025-08-04
Move remote transfer cores to RemoteCommunication

broskoTT:
### Issue
Related to https://github.com/tenstorrent/tt-umd/issues/995

### Description
Eth cores to use for remote communication are ideologically part of RemoteCommunication class, not LocalChip.
This is on a path for removing dependancy from RemoteWormholeTTDevice to LocalChip.

### List of the changes
- Move remote eth cores related functions to RemoteCommunication
- Now RemoteChip has to implement set_remote_ethernet_cores since it has to update its RemoteCommunication class object
- LocalChip also still has to do this, since LocalChip holds RemoteCommunication object for broadcast purposes.
- Callsites have to be changed, since now this has to be called for remoteChip (or remote communication) and not localchip
- RemoteWormholeTTDevice has to accept RemoteCommunication object, it can't construct it on it's own. The RemoteCommunication has to have its remote transfer eth cores already set before it is passed to the TTDevice since the constructor uses RemoteCommunication (Note: this might be changed with some of Nenad's changes in https://github.com/tenstorrent/tt-umd/issues/1162, but we can change this later again, not a problem).

### Testing
CI tests

### API Changes
This PR has API changes but they aren't directly used by tt-metal. It does affect how tt-smi will use these.

### DIFF
--- a/device/chip/local_chip.cpp
+++ b/device/chip/local_chip.cpp
@@ -374,51 +374,12 @@ void LocalChip::set_flush_non_mmio(bool flush_non_mmio) { flush_non_mmio_ = flus
 bool LocalChip::get_flush_non_mmio() const { return flush_non_mmio_; }
 
 void LocalChip::set_remote_transfer_ethernet_cores(const std::unordered_set<CoreCoord>& active_eth_cores) {
-    // Makes UMD aware of which ethernet cores have active links.
-    // Based on this information, UMD determines which ethernet cores can be used for host->cluster non-MMIO transfers.
-    // This overrides the default ethernet cores tagged for host to cluster routing in the constructor and must be
-    // called for all MMIO devices, if default behaviour is not desired.
-    TT_ASSERT(soc_descriptor_.arch == tt::ARCH::WORMHOLE_B0, "{} can only be called for Wormhole arch", __FUNCTION__);
-    remote_transfer_eth_cores_ = {};
-    for (const auto& active_eth_core : active_eth_cores) {
-        auto virtual_coord = soc_descriptor_.translate_coord_to(active_eth_core, CoordSystem::VIRTUAL);
-        remote_transfer_eth_cores_.push_back(active_eth_core);
-    }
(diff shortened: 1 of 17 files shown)


device/chip/local_chip.cpp
```diff
@@ -374,52 +374,13 @@ void LocalChip::set_flush_non_mmio(bool flush_non_mmio) { flush_non_mmio_ = flus
...
-    // Based on this information, UMD determines which ethernet cores can be used for host->cluster non-MMIO transfers.
-    // This overrides the default ethernet cores tagged for host to cluster routing in the constructor and must be
-    // called for all MMIO devices, if default behaviour is not desired.
-    TT_ASSERT(soc_descriptor_.arch == tt::ARCH::WORMHOLE_B0, "{} can only be called for Wormhole arch", __FUNCTION__);
-    remote_transfer_eth_cores_ = {};
-    for (const auto& active_eth_core : active_eth_cores) {
-        auto virtual_coord = soc_descriptor_.translate_coord_to(active_eth_core, CoordSystem::VIRTUAL);
-        remote_transfer_eth_cores_.push_back(active_eth_core);
-    }
+    // Local chips don't use remote communication, so this is a noop.
```
broskoTT, line 377: There is actually a remote_communication_ for eth broadcast, so set that one as well.

device/cluster.cpp
```diff
@@ -437,7 +437,14 @@ Cluster::Cluster(ClusterOptions options) {
 
 void Cluster::configure_active_ethernet_cores_for_mmio_device(
     chip_id_t mmio_chip, const std::unordered_set<CoreCoord>& active_eth_cores_per_chip) {
-    chips_.at(mmio_chip)->set_remote_transfer_ethernet_cores(active_eth_cores_per_chip);
+    // The ethernet cores that should be used for remote transfer are set in the RemoteCommunication structure.
```
broskoTT, line 440: There is actually a remote_communication_ for eth broadcast, so set that one as well.

PR #1155, opened 2025-08-05
SocDescriptor accepts ChipInfo

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/1163

### Description
Change soc descriptor constructors so it accepts ChipInfo. This struct, when firmware version is removed, turns out to be exactly what is needed from runtime to construct soc descriptor.

### List of the changes
- Extract get_firmware_version as a separate function
- Whenever Soc descriptor is constructer use Chip info. Should be simplified in most of the cases
- Change SocDescriptor constructors to affect ChipInfo change

### Testing
CI testing

### API Changes
There are API changes in this PR only for soc desc constructors

### DIFF
--- a/device/api/umd/device/tt_soc_descriptor.h
+++ b/device/api/umd/device/tt_soc_descriptor.h
@@ -73,7 +73,7 @@ public:
     // Default constructor. Creates uninitialized object with public access to all of its attributes.
     tt_SocDescriptor() = default;
     // Constructor used to build object from device descriptor file.
-    tt_SocDescriptor(std::string device_descriptor_path, const ChipInfo chip_info = {});
+    tt_SocDescriptor(const std::string &device_descriptor_path, const ChipInfo chip_info = {});
 
     tt_SocDescriptor(const tt::ARCH arch, const ChipInfo chip_info = {});
 
--- a/device/tt_soc_descriptor.cpp
+++ b/device/tt_soc_descriptor.cpp
@@ -422,7 +422,7 @@ void tt_SocDescriptor::load_from_yaml(YAML::Node &device_descriptor_yaml) {
     load_from_soc_desc_info(soc_desc_info);
 }
 
-tt_SocDescriptor::tt_SocDescriptor(std::string device_descriptor_path, ChipInfo chip_info) :
+tt_SocDescriptor::tt_SocDescriptor(const std::string &device_descriptor_path, ChipInfo chip_info) :
     noc_translation_enabled(chip_info.noc_translation_enabled), harvesting_masks(chip_info.harvesting_masks) {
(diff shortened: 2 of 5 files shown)


device/api/umd/device/tt_soc_descriptor.h
```diff
@@ -73,19 +73,9 @@ class tt_SocDescriptor {
...
-        const BoardType board_type = BoardType::UNKNOWN,
-        const uint8_t asic_location = 0);
-
-    tt_SocDescriptor(
-        const tt::ARCH arch,
-        const bool noc_translation_enabled,
-        const HarvestingMasks harvesting_masks = {0, 0, 0},
-        const BoardType board_type = BoardType::UNKNOWN,
-        const uint8_t asic_location = 0);
+    tt_SocDescriptor(std::string device_descriptor_path, const ChipInfo chip_info = {});
```
pjanevskiTT, line 76: can we make device_descriptor_path `const std::string& `?

device/api/umd/device/types/cluster_descriptor_types.h
```diff
@@ -194,12 +194,11 @@ struct ChipUID {
 };
 
 struct ChipInfo {
-    HarvestingMasks harvesting_masks;
-    BoardType board_type;
-    ChipUID chip_uid;
+    bool noc_translation_enabled = false;
```
pjanevskiTT, line 197: why reorder the fields?

device/cluster.cpp
```diff
@@ -276,23 +276,22 @@ tt_SocDescriptor Cluster::construct_soc_descriptor(
...
+            get_harvesting_masks(chip_id, cluster_desc, perform_harvesting, simulated_harvesting_masks);
+        chip_info.board_type = cluster_desc->get_board_type(chip_id);
+        chip_info.asic_location = cluster_desc->get_asic_location(chip_id);
+    }
 
     if (soc_desc_path.empty()) {
         tt::ARCH arch = chip_in_cluster_descriptor ? cluster_desc->get_arch(chip_id) : tt::ARCH::WORMHOLE_B0;
 
-        return tt_SocDescriptor(arch, noc_translation_table_en, harvesting_masks, chip_board_type, asic_location);
+        return tt_SocDescriptor(arch, chip_info);
```
pjanevskiTT, line 291: should we maybe also move arch to chip info? Would that be useful/problematic?

device/api/umd/device/types/cluster_descriptor_types.h
broskoTT: This was just matching our current argument order in soc descriptor constructor, so this leads to less diff in this PR...

PR #1156, opened 2025-08-05
Revise Chip constructors

broskoTT:
### Issue
#1163 

### Description
Revisited our constructors.
Previously LocalChip had to accept constructed TTDevice object, since the first thing it calls is Chip constructor, and it needs info from TTDevice for it. I've circumvented this by using static functions instead of constructors, and using a single private constructor. I've also aligned these functions a bit better.

### List of the changes
- Changed existing LocalChip constructors to more uniform "create" functions
- Make RemoteChip align with said constructors
- wait_arc_core_start remove calling it from Wormhole, and it already wasn't called from Blackhole.
- wait_arc_core_start doesn't need arc core parameter
- Constructor callsites are now simplified

### Testing
CI test should be enough

### API Changes
This PR has API changes for Chip classes, but these are not directly used by metal or exalens.

### DIFF
--- a/device/topology_discovery.cpp
+++ b/device/topology_discovery.cpp
@@ -87,7 +87,7 @@ std::unique_ptr<RemoteChip> TopologyDiscovery::create_remote_chip(Chip* chip, tt
     }
 
     auto local_chip = dynamic_cast<LocalChip*>(gateway_chip);
-    auto eth_coord = get_local_eth_coord(gateway_chip);
+    auto eth_coord = get_remote_eth_coord(chip, eth_core);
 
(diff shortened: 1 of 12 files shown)


device/topology_discovery.cpp
```diff
@@ -158,9 +149,9 @@ void TopologyDiscovery::get_pcie_connected_chips() {
         }
         std::unique_ptr<LocalChip> chip = nullptr;
         if (sdesc_path != "") {
```
nbuncicTT, line 151: Remove if-else, remote chip as well.

PR #1158, opened 2025-08-05
Remove LocalChip from RemoteTTDevice

broskoTT:
### Issue
Final change for separating out the Chip and RemoteCommunication, #995 

### Description
Use TTDevice for RemoteCommunication

### List of the changes
- RemoteCommunication accepts TTDevice instead of LocalChip. Also doesn't use CoreCoord for set_remote_eth_cores but tt_xy_pair
- RemoteWormholeTTDevice accepts remotecommunication only, it doesn't need TTDevice, it can fetch it from remotecommunication. Changed so that it doesn't hold LocalChip but TTDevice, same as RemoteCommunication.
- Adjust TopologyDiscovery remote chip creation
- Change all other constructors accordingly
- Change RemoteCommunication to use TTDevice, changed read/write functions, and removed ->get_tt_device() calls

### Testing
Existing CI tests

### API Changes
API changes for RemoteCommunication, RemoteChip and RemoteWormholeTTDevice. Nothing is directly used currently by tt-metal, so no changes needed

### DIFF
--- a/device/api/umd/device/chip/remote_chip.h
+++ b/device/api/umd/device/chip/remote_chip.h
@@ -18,12 +18,12 @@ public:
     static std::unique_ptr<RemoteChip> create(
         LocalChip* local_chip,
         eth_coord_t target_eth_coord,
-        std::unordered_set<CoreCoord>& remote_transfer_eth_cores,
+        std::unordered_set<CoreCoord> remote_transfer_eth_cores,
         std::string sdesc_path = "");
     static std::unique_ptr<RemoteChip> create(
         LocalChip* local_chip,
         eth_coord_t target_eth_coord,
-        std::unordered_set<CoreCoord>& remote_transfer_eth_cores,
+        std::unordered_set<CoreCoord> remote_transfer_eth_cores,
         tt_SocDescriptor soc_descriptor);
 
     bool is_mmio_capable() const override;
--- a/device/remote_communication.cpp
+++ b/device/remote_communication.cpp
@@ -39,7 +39,7 @@ struct routing_cmd_t {
 };
 
 RemoteCommunication::RemoteCommunication(TTDevice* local_tt_device, SysmemManager* sysmem_manager) :
-    local_chip_(local_chip), sysmem_manager_(sysmem_manager) {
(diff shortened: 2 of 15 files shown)


device/api/umd/device/chip/remote_chip.h
```diff
@@ -18,12 +18,12 @@ class RemoteChip : public Chip {
     static std::unique_ptr<RemoteChip> create(
         LocalChip* local_chip,
         eth_coord_t target_eth_coord,
-        std::unordered_set<CoreCoord>& remote_transfer_eth_cores,
+        std::unordered_set<CoreCoord> remote_transfer_eth_cores,
```
pjanevskiTT, line 21: why remove & ?

device/remote_communication.cpp
```diff
@@ -38,10 +38,9 @@ struct routing_cmd_t {
     uint32_t src_addr_tag;  // upper 32-bits of request source address.
 };
 
-RemoteCommunication::RemoteCommunication(LocalChip* local_chip, SysmemManager* sysmem_manager) :
+RemoteCommunication::RemoteCommunication(TTDevice* local_chip, SysmemManager* sysmem_manager) :
```
pjanevskiTT, line 41: rename local_chip var. To not leave this comment everywhere, please take a look and rename it where it's needed

device/remote_communication.cpp
```diff
@@ -550,59 +543,49 @@ void RemoteCommunication::write_to_non_mmio(
...
+                        erisc_txn_counters.data(), core, eth_interface_params.request_cmd_queue_base, 8);
                 } while (erisc_txn_counters[0] != erisc_txn_counters[1]);
             }
         }
         flush_non_mmio_ = false;
     }
 }
 
-void RemoteCommunication::set_remote_transfer_ethernet_cores(const std::unordered_set<CoreCoord>& active_eth_cores) {
+void RemoteCommunication::set_remote_transfer_ethernet_cores(
```
pjanevskiTT, line 577: is this function even needed now, can we just have one eth_core per RemoteCommunicaton? This also fits nicely into what we want to have for lite fabric. Is there a use case where we want multiple eth cores?

Issue #1145, opened 2025-08-04
Add physical slot info to PCIDeviceInfo


pjanevskiTT:
**Timeline**

Not tied to deadline, should work on it soon to help tt-metal team

**Is your feature request related to a problem? Please describe.**

Add physical slot info to PCIDeviceInfo

This is queryable from the system/linux, ex lspci -vvv -s 0000:4b:00 output for a card is below, where we are interested in physical slot.

```
4b:00.0 Processing accelerators: Device 1e52:401e (rev 01)
	Subsystem: Device 1e52:0014
	Physical Slot: 2
	Control: I/O- Mem+ BusMaster+ SpecCycle- MemWINV- VGASnoop- ParErr+ Stepping- SERR+ FastB2B- DisINTx+
	Status: Cap+ 66MHz- UDF- FastB2B- ParErr- DEVSEL=fast >TAbort- <TAbort- <MAbort- >SERR- <PERR- INTx-
	Latency: 0, Cache Line Size: 32 bytes
	Interrupt: pin A routed to IRQ 248
	NUMA node: 0
	Region 0: Memory at 203fc0000000 (64-bit, prefetchable) [size=512M]
	Region 2: Memory at bba00000 (32-bit, non-prefetchable) [size=1M]
	Region 4: Memory at 203fe0000000 (64-bit, prefetchable) [size=32M]
	Capabilities: <access denied>
	Kernel driver in use: tenstorrent
	Kernel modules: tenstorrent
```

**Proposed solution**

Query information about physical slot and add it to PCIDeviceInfo. Should we do this through UMD or KMD?

pjanevskiTT:
closed by #1159

PR #1159, opened 2025-08-05
Add physical slot to PCIDeviceInfo

pjanevskiTT:
### Issue

#1145 

### Description

Add optional physical slot information to PciDeviceInfo. This is something that is going to be used by metal team to build machine provisioning tool, it shouldn't be used by regular UMD flow. We should think about moving this information to some other place.

### List of the changes

- Add optional physical slot info to PciDeviceInfo
- Get physical slot info by lspci

### Testing
Manual testing on T3K

### API Changes
/

### DIFF
--- a/device/pcie/pci_device.cpp
+++ b/device/pcie/pci_device.cpp
@@ -103,6 +103,48 @@ static bool detect_iommu(const PciDeviceInfo &device_info) {
     return false;
 }
 
+static std::string get_pci_bdf(const uint16_t pci_domain, const uint16_t pci_bus, const uint16_t pci_device) {
+    return fmt::format("{:04x}:{:02x}:{:02x}", pci_domain, pci_bus, pci_device);
+}
+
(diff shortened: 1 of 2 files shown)


device/pcie/pci_device.cpp
```diff
@@ -103,6 +103,31 @@ static bool detect_iommu(const PciDeviceInfo &device_info) {
     return false;
 }
 
+static std::string get_pci_bdf(
+    const uint16_t pci_domain, const uint16_t pci_bus, const uint16_t pci_device, const uint16_t pci_function) {
+    return fmt::format("{:04x}:{:02x}:{:02x}.{:x}", pci_domain, pci_bus, pci_device, pci_function);
+}
+
+static std::optional<int> get_physical_slot(const std::string &pci_bdf) {
+    std::string cmd = "lspci -vvv -s " + pci_bdf;
```
pjanevskiTT, line 112: are there better ways to do this?

pjanevskiTT: reimplemented it to not use external cmd

tt-aho (approved):

Issue #1153, opened 2025-08-04
TT_VISIBLE_DEVICES env variable


pjanevskiTT:
**Timeline**
No specific timeline

**Is your feature request related to a problem? Please describe.**

https://github.com/tenstorrent/tt-metal/issues/25860
Could help avoid bugs like this in the future

**Proposed solution**

Implement TT_VISIBLE_DEVICES env var. This env var should serve the same purpose as TT_METAL_VISIBLE_DEVICES. It should replace metal env var and should be used both by metal/umd and other runtime components. This is better solution that existing pci_target_devices API because metal has bunch of places where they call enumeration functions from PCIDevice class and somewhere they forget to use TT_METAL_VISIBLE_DEVICES which leads to problems describes in the issue above

pjanevskiTT:
@bgoelTT @cfjchu merged the changes. @cfjchu Let me know if any help is needed when tt-metal wants to integrate this

pjanevskiTT:
Closed by #1160

bgoelTT:
Thank you @pjanevskiTT , we don't have an issue in tt-metal for integrating this feature right? If not, I will create one

cfjchu:
Yes, we need your help to integrate the changes through metal - thanks!

bgoelTT:
I created an issue here https://github.com/tenstorrent/tt-metal/issues/26658, please feel free to add any more detail you feel necessary in the comments

PR #1160, opened 2025-08-06
Implement TT_VISIBLE_DEVICES

pjanevskiTT:
### Issue

#1153 

### Description

Implement support for specifying visible devices by env var `TT_VISIBLE_DEVICES`. For now we are keeping the compatibility with specifying `pci_target_devices` as well because that's what metal relies on. In the future we should drop this API and have user only use `TT_VISIBLE_DEVICES`

### List of the changes

- Add support for `TT_VISIBLE_DEVICES` env var
- Keep compatibility with `pci_target_devices` API
- Change tests to test this as well
- Make topology tool use this path as well

### Testing

CI + manual testing on T3K

### API Changes
/

### DIFF
--- a/device/pcie/pci_device.cpp
+++ b/device/pcie/pci_device.cpp
@@ -183,6 +183,38 @@ tt::ARCH PciDeviceInfo::get_arch() const {
     return tt::ARCH::Invalid;
 }
 
+std::optional<std::unordered_set<int>> PCIDevice::get_visible_devices(
+    const std::unordered_set<int> &pci_target_devices) {
+    if (!pci_target_devices.empty()) {
+        return pci_target_devices;
+    }
+
+    const char *env_var = std::getenv(TT_VISIBLE_DEVICES_ENV);
+    if (!env_var) {
+        return std::nullopt;
+    }
+
+    std::unordered_set<int> visible_devices;
+    std::string env_str(env_var);
+    std::stringstream ss(env_str);
+    std::string token;
+
+    while (std::getline(ss, token, ',')) {
+        try {
+            visible_devices.insert(std::stoi(token));
+        } catch (const std::exception &e) {
+            throw std::runtime_error(fmt::format(
+                "Invalid device ID '{}' in {} environment variable: {}", token, TT_VISIBLE_DEVICES_ENV, e.what()));
+        }
+    }
+
+    if (visible_devices.empty()) {
+        return std::nullopt;
+    }
+
+    return visible_devices;
+}
+
 std::vector<int> PCIDevice::enumerate_devices(std::unordered_set<int> pci_target_devices) {
     std::vector<int> device_ids;
     std::string path = "/dev/tenstorrent/";
@@ -190,6 +222,9 @@ std::vector<int> PCIDevice::enumerate_devices(std::unordered_set<int> pci_target
     if (!std::filesystem::exists(path)) {
         return device_ids;
     }
+
+    std::optional<std::unordered_set<int>> visible_devices = PCIDevice::get_visible_devices(pci_target_devices);
+
     for (const auto &entry : std::filesystem::directory_iterator(path)) {
         std::string filename = entry.path().filename().string();
 
@@ -197,7 +232,8 @@ std::vector<int> PCIDevice::enumerate_devices(std::unordered_set<int> pci_target
         // is probably what we want longer-term (i.e. a UUID or something).
         if (std::all_of(filename.begin(), filename.end(), ::isdigit)) {
             int pci_device_id = std::stoi(filename);
-            if (pci_target_devices.empty() || pci_target_devices.find(pci_device_id) != pci_target_devices.end()) {
+            if (!visible_devices.has_value() ||
+                visible_devices.value().find(pci_device_id) != visible_devices.value().end()) {
                 device_ids.push_back(pci_device_id);
             }
         }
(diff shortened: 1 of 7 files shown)


tests/api/test_cluster.cpp
```diff
@@ -101,6 +102,24 @@ TEST(ApiClusterTest, OpenChipsByPciId) {
             // Always expect logical id 0 to exist, that's the way filtering by pci ids work.
             EXPECT_TRUE(actual_pci_device_ids.find(0) != actual_pci_device_ids.end());
         }
+
+        std::string value = test_utils::convert_to_comma_separated_string(target_pci_device_ids);
+
+        if (setenv(TT_VISIBLE_DEVICES_ENV, value.c_str(), 1) != 0) {
```
nbuncicTT, line 108: This is in the test body, so I don't think it makes sense to return as this is not standard for `gtest`. If you want to terminate the test, use `ASSSERT_EQ`.

tests/api/test_cluster.cpp
```diff
@@ -101,6 +102,24 @@ TEST(ApiClusterTest, OpenChipsByPciId) {
...
+
+        // Make sure that Cluster construction is without exceptions.
+        // TODO: add cluster descriptors for expected topologies, compare cluster desc against expected desc.
+        std::unique_ptr<Cluster> cluster_env_var = std::make_unique<Cluster>(ClusterOptions{
+            .pci_target_devices = {},
+        });
+
+        if (unsetenv(TT_VISIBLE_DEVICES_ENV) != 0) {
+            std::cerr << "Failed to unset environment variable." << std::endl;
+            return;
```
nbuncicTT, line 121: Same as above.

tests/test_utils/device_test_utils.hpp
```diff
@@ -44,4 +44,17 @@ inline void fill_with_random_bytes(uint8_t* data, size_t n) {
     }
 }
 
+inline std::string convert_to_comma_separated_string(const std::unordered_set<int>& devices) {
```
nbuncicTT, line 47: I think it would be cleaner to use a library for this:
```
#include <fmt/ranges.h>
 
inline std::string convert_to_comma_separated_string(const std::unordered_set<int>& devices) {
    return fmt::format("{}", fmt::join(devices, ","));
}
```

nbuncicTT (approved):
Well done

tests/test_utils/device_test_utils.hpp
pjanevskiTT: wow nice, didn't know this existed

device/pcie/pci_device.cpp
```diff
@@ -130,21 +130,60 @@ tt::ARCH PciDeviceInfo::get_arch() const {
     return tt::ARCH::Invalid;
 }
 
+std::optional<std::unordered_set<int>> PCIDevice::get_visible_devices(
+    const std::unordered_set<int> &pci_target_devices) {
+    if (!pci_target_devices.empty()) {
+        return pci_target_devices;
+    }
+
+    const char *env_var = std::getenv(TT_VISIBLE_DEVICES_ENV.data());
```
broskoTT, line 139: Reading env variable is really something which should live in common folder/lib/header.
Getting the value should not be more than one line, here or wherever

tools/topology.cpp
```diff
@@ -38,12 +38,19 @@ int main(int argc, char *argv[]) {
...
-    std::unordered_set<int> pci_ids = {};
     if (result.count("pci_devices")) {
-        pci_ids = extract_int_set(result["pci_devices"]);
+        auto vec_devices = result["pci_devices"].as<std::vector<std::string>>();
+        std::ostringstream oss;
+        oss << vec_devices[0];
+        for (const auto &device : vec_devices) {
+            oss << "," << device;
+        }
+        if (setenv(TT_VISIBLE_DEVICES_ENV.data(), oss.str().c_str(), 1) != 0) {
```
broskoTT, line 48: Yeah let's not do this in the code.... Can it happen that I have TT_VISIBLE_DEVICES set in my .bashrc since I always want to use a specific device. Then calling the topology tool completely unsets the env variable?

tests/api/test_cluster.cpp
```diff
@@ -101,6 +102,22 @@ TEST(ApiClusterTest, OpenChipsByPciId) {
...
+            ASSERT_TRUE(false) << "Failed to unset environment variable.";
+        }
+
+        // Make sure that Cluster construction is without exceptions.
+        // TODO: add cluster descriptors for expected topologies, compare cluster desc against expected desc.
+        std::unique_ptr<Cluster> cluster_env_var = std::make_unique<Cluster>(ClusterOptions{
+            .pci_target_devices = {},
+        });
+
+        if (unsetenv(TT_VISIBLE_DEVICES_ENV.data()) != 0) {
```
broskoTT, line 118: Copy the EXPECTs from above? So that we know filtering was applied?

device/pcie/pci_device.cpp
```diff
@@ -130,21 +130,60 @@ tt::ARCH PciDeviceInfo::get_arch() const {
     return tt::ARCH::Invalid;
 }
 
+std::optional<std::unordered_set<int>> PCIDevice::get_visible_devices(
```
broskoTT, line 133: You have to read this variable in TopologyDiscovery constructor right?

broskoTT (approved):
Nice.
I don't think we should remove .pci_target_devices from our ClusterOptions though. But we should remove metal code

tools/topology.cpp
pjanevskiTT: Ok makes sense

Issue #1162, opened 2025-08-06
Move init_device() out of TTDevice constructor


nbuncicTT:
**Timeline**
No specific timeline.

**Is your feature request related to a problem? Please describe.**
When `TTDevice` is being initialized, it spawns the the `ArcMessenger` and `ArcTelemetryReader` before it waits for the ARC core to be ready. This leads to potential pitfalls when the `TTDevice` is being constructed, and may invoke an exception during a warm reset and then the device isn't properly reset.

**Proposed solution**
Move the `init_tt_device()` out of the constructor and make the user of the `TTDevice` class manually call it when it's appropriate. 

This makes sense:
- It allows additional flexibility of using the `TTDevice` even if it's not in the proper state
- It doesn't retain state which is more in-line with the design principles of this class

PR #1165, opened 2025-08-07
Remove init tt device from constructor

nbuncicTT:
### Issue
#1162 

### Description
Remove `init_tt_device()` from constructor, as it essentially inits an `ArcMessenger` and `TelemetryReader` - and this can be done after `TTDevice` creation which would be the desired behavior.

### List of the changes
- Remove init_tt_device() from constructor
- Make init_tt_device non-virtual
- Add private virtual pre and post hooks in there are some necessary steps for init_tt_device for derived classes to do
- Modify tests to accept this change

### Testing
Manual

### API Changes
/


device/api/umd/device/tt_device/tt_device.h
```diff
@@ -258,15 +260,18 @@ class TTDevice {
...
     semver_t fw_version_from_telemetry(const uint32_t telemetry_data) const;
 
     TTDevice();
 
     ChipInfo chip_info;
 
     bool is_remote_tt_device = false;
+
+private:
+    virtual void pre_init_hook(){};
```
pjanevskiTT, line 272: this is nice. No action

tests/api/test_tlb_manager.cpp
```diff
@@ -19,6 +19,8 @@ TEST(ApiTLBManager, ManualTLBConfiguration) {
 
     for (int pci_device_id : pci_device_ids) {
         std::unique_ptr<TTDevice> tt_device = TTDevice::create(pci_device_id);
+        tt_device->init_tt_device();
```
pjanevskiTT, line 22: I don't like that we have to call these 2 function explicitly now....
A lot of examples below as well now require 3 lines of code to initialize TTDevice

pjanevskiTT (commented):
Not really sure what this change is supposed to introduce/solve. I feel we are just moving this out of TTDevice and calling it explicitly. I understand that now someone could use TTDevice just for read/write without having to explicitly create ARC related objects. If that is the point I feel there are better ways to solve that problem

Issue #1168, opened 2025-08-11
Consolidate device folder structure


broskoTT:
There are two ways we can structure files that are arch specific.
Option 1:
```
a.h
b.h
wormhole/wormhole_a.h
wormhole/wormhole_b.h
blackhole/blackhole_a.h
blackhole/blackhole_b.h
```

Option 2:
```
a/a.h
a/wormhole_a.h
a/blackhole_a.h
b/b.h
b/wormhole_b.h
b/blackhole_b.h
```

I personally prefer the second option, and some files in our repo are not conforming to this rule. 

This issue tracks such work

Also could be part of general guide https://github.com/tenstorrent/tt-umd/issues/356

PR #1169, opened 2025-08-11
Move arc_ files to dedicated folder

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/1168

### Description
Move arc related header and implementation files

### List of the changes
- Move arc_telemetry_reader, arc_messenger for all archs to dedicated folders
- Update file locations where needed.

### Testing
Code builds

### API Changes


device/CMakeLists.txt
```diff
@@ -56,13 +56,13 @@ target_sources(
...
         wormhole/wormhole_coordinate_manager.cpp
         blackhole/blackhole_coordinate_manager.cpp
-        arc_telemetry_reader.cpp
-        wormhole/wormhole_arc_telemetry_reader.cpp
-        blackhole/blackhole_arc_telemetry_reader.cpp
-        arc_messenger.cpp
-        wormhole/wormhole_arc_messenger.cpp
-        blackhole/blackhole_arc_messenger.cpp
-        blackhole/blackhole_arc_message_queue.cpp
+        arc/arc_telemetry_reader.cpp
```
pjanevskiTT, line 59: I don't necessarily have anything against this change, would just like to confirm that we are going to group our files based on part of the system/device, for example tensix/dram/noc etc..
Don't really have any preference, would just like to know your opinion on it.
One suggestion I would give maybe. I think there is still value to having separate directories for each chip architecture.
So I would do something like
```
arc
|---wormhole
|     |-----------wh_arc_messenger.cpp
|---blachkole
      |----------bh_arc_messenger.cpp
```

Let me know what you think

pjanevskiTT (approved):

device/CMakeLists.txt
broskoTT: But you already have it in the name, why have multiple wormhole in path? What's the benefit of arc/wh/wh_arc instead of arc/wh_arc ? To me it just seems like additional cluttering.

The idea wasn't to separate it based on part of system, but based on code design, we already have chip, chip helpers, tt_device, etc... 
Next I would add is arch_implementation, would move some to types, would maybe create cluster descriptor folder... you're already moving to topology discovery.

The name "arc" is just the best I could come up with, feel free to suggest other name

Issue #797, opened 2025-04-25
Remove virtual coordinate system


pjanevskiTT:
Remove virtual coordinate system for Core coordinates. Virtual coordinate system was originally used at API level when the API was only accepting tt_cxy_pair. Now with mature CoreCoord struct I think we can get rid of virtual coordinate system. It's just confusing and it's starting to become an overhead in maintaining code around coordinates

pjanevskiTT:
closed by #1329

PR #1173, opened 2025-08-11
Some coordinate modifications

broskoTT:
### Issue
On a path of #687 and a bit of #797 

### Description
Further modification of cleaning up coord internaly.
I've chosen for TLBManager to use tt_xy_pair instead of CoreCoord, although both options make sense to me.
Also cleaned up some other stuff along the way

### List of the changes
- TLBManager accepts translated coords always, not virtual coords for its internal map
- Use the same translation function in Chip class as other CoreCoord->tt_xy_pair transformations are using
- Some code, mainly in tests, used translation where it wasn't needed anymore, so cleaned that up a bit

### Testing
Existing CI tests should be fine.

### API Changes
There are slight TLBManager API changes in this PR, but this is not used in clients.


device/api/umd/device/chip_helpers/tlb_manager.h
```diff
@@ -19,10 +19,8 @@ class TLBManager {
 public:
     TLBManager(TTDevice* tt_device);
 
-    // TODO: Think about proper API which doesn't accept two cores.
-    // core should be in VIRTUAL coords, and translated_core should be in TRANSLATED coords.
-    void configure_tlb(
-        tt_xy_pair core, tt_xy_pair translated_core, int32_t tlb_index, uint64_t address, uint64_t ordering);
+    // All tt_xy_pairs should be in TRANSLATED coords.
```
pjanevskiTT, line 22: Since this is how we want API to be, why not have this function accept CoreCoord and then convert it to translated inside. Maybe because of introducing SocDescriptor dependency to this class? I agree that we probably wouldn't want that. Any other reasons?

tests/api/test_tlb_manager.cpp
```diff
@@ -55,16 +55,14 @@ TEST(ApiTLBManager, ManualTLBConfiguration) {
 
         std::int32_t c_zero_address = 0;
 
-        for (CoreCoord core : soc_desc.get_cores(CoreType::TENSIX)) {
-            auto virtual_core = soc_desc.translate_coord_to(core, CoordSystem::VIRTUAL);
-            auto translated_core = soc_desc.translate_coord_to(core, CoordSystem::TRANSLATED);
-            tlb_manager->configure_tlb(
-                virtual_core, translated_core, get_static_tlb_index(core), c_zero_address, tlb_data::Relaxed);
+        for (CoreCoord core : soc_desc.get_cores(CoreType::TENSIX, CoordSystem::TRANSLATED)) {
+            tt_xy_pair translated_core = core;
```
pjanevskiTT, line 59: This is not needed, right? You can just pass core var to the function below

pjanevskiTT (approved):

device/api/umd/device/chip_helpers/tlb_manager.h
broskoTT: Yeah that is exactly the reason. If I do that, then I have to pass socdescriptor.
I thought of maybe even passing raw coordinate manager pointer instead...

Issue #450, opened 2025-01-03
Quasar implementation


pjanevskiTT:
Add something similar to grayskull/wormhole/blackhole_implementation files for quasar as well. Figure out NOC endpoints and how that is going to be visible through host

pjanevskiTT:
@tt-vjovanovic don't want to assign this to you, but if you stumble upon something like this while bringing up the simulator, you can take it

PR #1174, opened 2025-08-11
Fixes needed for simulator to work

tt-vjovanovic:
There were 4 issues:
- Even though we are using simulator, we still need to provide soc_descriptor to `ClusterOptions`
- There was hardcoded version to be used to create mock cluster for simulation (wormhole_b0)
- archutecture_implementation didn't know what to do with Quasar (it uses blackhole now, but that should be fixed with #450)
- Serialization of soc descriptor created invalid yaml file if dram doesn't have 3 end points (simulators have one or two)

### DIFF
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -408,11 +408,12 @@ Cluster::Cluster(ClusterOptions options) {
             }
         }
         if (construct_mock_cluster_descriptor) {
-#ifdef TT_UMD_BUILD_SIMULATION
-            tt_SimulationDeviceInit init(options.simulator_directory);
-            auto arch = init.get_soc_descriptor().arch;
-#else
             auto arch = tt::ARCH::WORMHOLE_B0;
+#ifdef TT_UMD_BUILD_SIMULATION
+            if (options.chip_type == ChipType::SIMULATION) {
+                tt_SimulationDeviceInit init(options.simulator_directory);
+                arch = init.get_soc_descriptor().arch;
+            }
 #endif
             cluster_desc = tt_ClusterDescriptor::create_mock_cluster(chips_to_construct_vec, arch);
         }
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -409,7 +409,8 @@ Cluster::Cluster(ClusterOptions options) {
         }
         if (construct_mock_cluster_descriptor) {
             tt_SimulationDeviceInit init(options.simulator_directory);
-            cluster_desc = tt_ClusterDescriptor::create_mock_cluster(chips_to_construct_vec, init.get_soc_descriptor().arch);
+            cluster_desc =
+                tt_ClusterDescriptor::create_mock_cluster(chips_to_construct_vec, init.get_soc_descriptor().arch);
         }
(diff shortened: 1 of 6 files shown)


device/tt_cluster_descriptor.cpp
```diff
@@ -544,9 +544,10 @@ std::unique_ptr<tt_ClusterDescriptor> tt_ClusterDescriptor::create_mock_cluster(
         case tt::ARCH::WORMHOLE_B0:
             board_type = BoardType::N150;
             break;
+        case tt::ARCH::QUASAR:  // TODO (#450): Add Quasar configuration
         case tt::ARCH::BLACKHOLE:
-            board_type = BoardType::P150;
-            harvesting_masks.pcie_harvesting_mask = 0x2;
+            board_type = BoardType::UNKNOWN;
```
broskoTT, line 549: Do we need a boardtype for something even in mock cluster? Can we avoid the switch altogether and just always set it to unknown?

device/cluster.cpp
```diff
@@ -406,7 +406,12 @@ Cluster::Cluster(ClusterOptions options) {
             }
         }
         if (construct_mock_cluster_descriptor) {
-            cluster_desc = tt_ClusterDescriptor::create_mock_cluster(chips_to_construct_vec, tt::ARCH::WORMHOLE_B0);
+            tt_SimulationDeviceInit init(options.simulator_directory);
+            cluster_desc =
+                tt_ClusterDescriptor::create_mock_cluster(chips_to_construct_vec, init.get_soc_descriptor().arch);
+        }
+        if (options.sdesc_path.empty()) {
+            options.sdesc_path = options.simulator_directory / "soc_descriptor.yaml";
```
broskoTT, line 414: This code dangerously assumes that the soc_descriptor.yaml exists. If somebody wants to create a mock or simulator Chip explicitly without the yaml, that should end up working right?
Why not pass this path when calling the constructor specifically? https://github.com/tenstorrent/tt-metal/blob/main/tt_metal/llrt/tt_cluster.cpp#L350

broskoTT (approved):
Thanks for fixing

device/tt_cluster_descriptor.cpp
tt-vjovanovic: @pjanevskiTT suggested this change... Lets agree offline how to proceed...

device/cluster.cpp
```diff
@@ -408,7 +408,16 @@ Cluster::Cluster(ClusterOptions options) {
             }
         }
         if (construct_mock_cluster_descriptor) {
-            cluster_desc = tt_ClusterDescriptor::create_mock_cluster(chips_to_construct_vec, tt::ARCH::WORMHOLE_B0);
+#ifdef TT_UMD_BUILD_SIMULATION
+            tt_SimulationDeviceInit init(options.simulator_directory);
```
broskoTT, line 412: we should do this only for SIMULATION not for MOCK.
in case of MOCK if simulator_directory is empty it will still try accessing some soc_descriptor.yaml

pjanevskiTT: I think I resolved it

PR #1175, opened 2025-08-12
JTAG read/write integration.

uvelimirovicTT:
### Description
This is the first iteration of JTAG integration into tt-umd.

### List of the changes
TTDevice can now handle JTAG operations. Currently only available for Wormhole devices.


device/jtag/jtag_device.cpp
```diff
@@ -0,0 +1,351 @@
...
+    if (chip_id >= get_device_cnt()) {
+        return {};
+    }
+
+    return efuse_harvesting[chip_id];
+}
+
+bool JtagDevice::select_device(uint8_t chip_id) {
+    if (chip_id >= get_device_cnt()) {
+        return false;
```
nbuncicTT, line 80: Add log_warning here that the selected device doesn't exist or isn't available

device/jtag/jtag_device.cpp
```diff
@@ -0,0 +1,351 @@
...
+}
+
+JtagDevice::~JtagDevice() {
+    if (curr_device_idx != 0xff) {
+        jtag->close_jlink();
+        curr_device_idx = 0xff;
+    }
+}
+
+std::optional<uint32_t> JtagDevice::get_device_cnt() const { return jlink_devices.size(); }
```
nbuncicTT, line 68: Do we really need std::optional here? Let's use it where we really need it

device/jtag/jtag_device.cpp
```diff
@@ -0,0 +1,351 @@
...
+            return tt::ARCH::Invalid;
+    }
+}
+
+std::optional<int> JtagDevice::open_jlink_by_serial_wrapper(uint8_t chip_id, unsigned int serial_number) {
+    jtag->close_jlink();
+    return jtag->open_jlink_by_serial_wrapper(serial_number);
+}
+
+std::optional<int> JtagDevice::open_jlink_wrapper(uint8_t chip_id) { return jtag->open_jlink_wrapper(); }
```
nbuncicTT, line 112: Where is the invalid value here? If there is no invalid value, remove `std::optional`

device/jtag/jtag_device.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+std::optional<int> JtagDevice::write_tdr(uint8_t chip_id, const char* client, uint32_t reg_offset, uint32_t data) {
+    if (!select_device(chip_id)) {
+        return {};
+    }
+    jtag->write_tdr(client, reg_offset, data);
+    return 0;
+}
+
+std::optional<int> JtagDevice::dbus_memdump(
```
nbuncicTT, line 145: You can either declare this void, or if you want UNIX based return values (0 means successful execution, 1-255 some error code), you don't need `std::optional`

device/jtag/jtag_device.cpp
```diff
@@ -0,0 +1,351 @@
...
+    return 0;
+}
+
+std::optional<int> JtagDevice::write32_axi(uint8_t chip_id, uint32_t address, uint32_t data) {
+    if (!select_device(chip_id)) {
+        return {};
+    }
+
+    jtag->write_axi(address, data);
+    return 4;
```
nbuncicTT, line 178: We don't return the number of written bytes in our APIs

device/api/umd/device/jtag_device.h
```diff
@@ -0,0 +1,64 @@
...
+    std::optional<uint32_t> read_id_raw(uint8_t chip_id);
+    std::optional<uint32_t> read_id(uint8_t chip_id);
+    std::optional<uint8_t> get_current_device_idx() const;
+
+private:
+    bool select_device(uint8_t chip_id);
+    std::unique_ptr<Jtag> jtag;
+    std::vector<uint32_t> jlink_devices;
+    std::vector<uint32_t> efuse_harvesting;
+    uint8_t curr_device_idx = -1;
```
nbuncicTT, line 63: Why are you setting an unsigned int to a negative value?

device/jtag/jtag_device.cpp
```diff
@@ -0,0 +1,351 @@
...
+         * format */
+        efuse_harvesting.push_back(bad_row_bits);
+
+        jtag->close_jlink();
+    }
+    if (jlink_devices.empty()) {
+        throw std::runtime_error("There are no supported devices");
+    }
+
+    curr_device_idx = 0xff;
```
nbuncicTT, line 58: Why is this set from -1 to 0xff in the constructor?

device/api/umd/device/jtag.h
```diff
@@ -0,0 +1,46 @@
...
+    uint32_t read_axi(uint32_t reg_addr);
+    uint32_t read_noc_xy(uint32_t node_x_id, uint32_t node_y_id, uint64_t noc_addr);
+    std::vector<uint32_t> enumerate_jlink();
+    void close_jlink();
+    uint32_t read_id_raw();
+    uint32_t read_id();
+
+private:
+    void* handle;
+    std::unordered_map<std::string, void*> funcMap;
```
nbuncicTT, line 42: nit: let's use the same case for member variables, instead of `funcMap` use `func_map` of `function_map`

device/api/umd/device/tt_device/tt_device.h
```diff
@@ -260,13 +277,17 @@ class TTDevice {
...
+
     semver_t fw_version_from_telemetry(const uint32_t telemetry_data) const;
 
     TTDevice();
 
     ChipInfo chip_info;
 
     bool is_remote_tt_device = false;
+
+    static std::filesystem::path jtag_library_path;
```
nbuncicTT, line 290: You can make this also `private`

device/tt_device/tt_device.cpp
```diff
@@ -26,20 +34,65 @@ TTDevice::TTDevice(
...
+        if (len != -1) {
+            buffer[len] = '\0';
+            std::string path(buffer);
+            std::string::size_type pos = path.find_last_of("/");
+            binary_directory = path.substr(0, pos);
+        }
+    }
+
+    std::unique_ptr<Jtag> jtag;
+    jtag = std::make_unique<Jtag>(binary_directory.c_str());
```
nbuncicTT, line 67: Create `Jtag` in one line: `std::unique_ptr<Jtag> jtag = std::make_unique<Jtag>(binary_directory.c_str());`

device/tt_device/tt_device.cpp
```diff
@@ -26,20 +34,65 @@ TTDevice::TTDevice(
...
+    std::unique_ptr<JtagDevice> jtag_device = std::make_unique<JtagDevice>(std::move(jtag));
+
+    // Check that all chips are of the same type
+    tt::ARCH arch = *jtag_device->get_jtag_arch(0);
+    for (size_t i = 1; i < jtag_device->get_device_cnt(); i++) {
+        auto newArch = *jtag_device->get_jtag_arch(i);
```
nbuncicTT, line 73: nit: use snake case for local variables (`newArch` -> `new_arch`)

nbuncicTT: And if you use std::optional, check prior to de-referencing that it actually has a value

device/jtag/jtag_device.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+#include <cstdint>
+#include <cstring>
+#include <iostream>
+
+#include "assert.hpp"
+#include "umd/device/jtag.h"
+
+#define ROW_LEN 12
+#define WORMHOLE_ID 0x138a5
```
pjanevskiTT, line 14: I am pretty sure this is already defined somewhere, can you reuse it. If not, can you define it in a common place and then reuse it?

device/jtag/jtag_device.cpp
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    for (int jlink_id : potential_devices) {
+        uint32_t status = jtag->open_jlink_by_serial_wrapper(jlink_id);
+        if (status != 0) {
+            continue;
+        }
+        uint32_t id = jtag->read_id();
+        if (id != WORMHOLE_ID) {
+            std::cerr << "Only supporting WORMHOLE for now" << std::endl;
```
pjanevskiTT, line 33: use some log_* statement, not std::cerr

device/jtag/jtag_device.cpp
```diff
@@ -0,0 +1,351 @@
...
+        jtag->open_jlink_by_serial_wrapper(jlink_devices[curr_device_idx]);
+    }
+    return true;
+}
+
+std::optional<tt::ARCH> JtagDevice::get_jtag_arch(uint8_t chip_id) {
+    auto id_opt = read_id(chip_id);
+
+    if (!id_opt) {
+        return {};
```
pjanevskiTT, line 95: we can return invalid here, and not have to use optional

device/jtag/jtag_device.cpp
```diff
@@ -0,0 +1,351 @@
...
+    return 4;
+}
+
+std::optional<int> JtagDevice::write32(uint8_t chip_id, uint8_t noc_x, uint8_t noc_y, uint64_t address, uint32_t data) {
+    if (!select_device(chip_id)) {
+        return {};
+    }
+
+    jtag->write_noc_xy(noc_x, noc_y, address, data);
+    // sstd::this_thread::sleep_for(std::chrono::seconds(2));
```
pjanevskiTT, line 187: you probably want to remove this comment?

device/jtag/jtag_device.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+    uint8_t chip_id = get_current_device_idx().value();
+
+    const uint32_t chunk_size = sizeof(uint32_t);
+
+    while (size > 0) {
+        uint32_t transfer_size = std::min(size, chunk_size);
+
+        if (transfer_size == sizeof(uint32_t) && (addr % sizeof(uint32_t)) == 0) {
+            // Aligned 32-bit write - most efficient
```
pjanevskiTT, line 203: nit: put . at the end of the comment

device/tt_device/tt_device.cpp
```diff
@@ -16,6 +21,9 @@ bool umd_use_noc1 = false;
 
 namespace tt::umd {
 
+/* static */ std::filesystem::path TTDevice::jtag_library_path =
+    std::filesystem::path("./build/lib/libttexalens_jtag.so");
```
pjanevskiTT, line 25: we would probably want to rename this .so now? Don't have to address it in this PR

pjanevskiTT (commented):
Round 1 of comments, will continue later on

tests/api/test_jtag.cpp
```diff
@@ -0,0 +1,36 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent Inc.
```
pjanevskiTT, line 1: TODO for myself: review this file

device/tt_device/tt_device.cpp
```diff
@@ -3,11 +3,16 @@
 // SPDX-License-Identifier: Apache-2.0
```
pjanevskiTT, line 3: TODO for myself: review this file

device/api/umd/device/tt_device/tt_device.h
```diff
@@ -117,6 +127,12 @@ class TTDevice {
     virtual void read_from_device(void *mem_ptr, tt_xy_pair core, uint64_t addr, uint32_t size);
     virtual void write_to_device(const void *mem_ptr, tt_xy_pair core, uint64_t addr, uint32_t size);
 
+    /*
+     * JTAG read/write functions.
+     */
+    virtual void jtag_read_from_device(void *mem_ptr, tt_xy_pair core, uint64_t addr, uint32_t size);
```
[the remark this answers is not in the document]
nbuncicTT: I agree with this one, this if-else can be a first step, but in the long run I think we should opt out with a solution that has a base class and then we have derived classes which define and encapsulate **how** we write.

Issue #1167, opened 2025-08-11
Remove tt_ class prefix


broskoTT:
Our classes don't need to have tt_ prefix if they're in tt::umd:: namespace, tied to https://github.com/tenstorrent/tt-umd/issues/103

Somewhat related to https://github.com/tenstorrent/tt-umd/issues/324, but a separate effort

PR #1196, opened 2025-08-19
Rename simulation files and move them

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/1167

### Description
Rename Simulation related classes not to have tt_ prefix

### List of the changes
- Rename tt_SimulationDeviceInit to SimulationDeviceInit
- Rename tt_SimulationDevice to SimulationDevice
- Rename tt_SimulationHost to SimulationHost
- Renamed files accordingly
- Moved files to simulation folder

### Testing
Builds
Tested that simulation tests still pass

### API Changes
This PR has API changes, but I'll leave old header files if necessary until the clients move forward.
You will see that metal build fails on this PR, but it builds at the tip of the PR stack:
https://github.com/tenstorrent/tt-umd/pull/1218

### DIFF
--- a/device/CMakeLists.txt
+++ b/device/CMakeLists.txt
@@ -1,7 +1,7 @@
 set(POSITION_INDEPENDENT_CODE ON)
 
 if(TT_UMD_BUILD_SIMULATION)
-    set(FBS_FILE ${PROJECT_SOURCE_DIR}/device/simulation/tt_simulation_device.fbs)
+    set(FBS_FILE ${PROJECT_SOURCE_DIR}/device/simulation/simulation_device.fbs)
     get_filename_component(FBS_FILE_NAME ${FBS_FILE} NAME_WLE)
     set(FBS_GENERATED_HEADER "${CMAKE_CURRENT_BINARY_DIR}/${FBS_FILE_NAME}_generated.h")
     set(FBS_DEPENDS "")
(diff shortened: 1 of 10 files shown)


device/CMakeLists.txt
```diff
@@ -125,8 +125,8 @@ if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.23)
                 api/umd/device/tt_device/remote_blackhole_tt_device.h
                 api/umd/device/tt_io.hpp
                 api/umd/device/tt_silicon_driver_common.hpp
-                api/umd/device/tt_simulation_device.h
-                api/umd/device/tt_simulation_host.hpp
+                api/umd/device/simulation/simulation_device.h
+                api/umd/device/simulation/simulation_host.hpp
```
nbuncicTT, line 129: a question out of pure curiosity: Why do we use .hpp and .h? Is there a specific reason or it was just like that so we continued? Should we think about migrating everything to .h or .hpp? Should some directories be .hpp and some .h? What do you think?

nbuncicTT (approved):
Nice!

device/CMakeLists.txt
broskoTT: https://github.com/tenstorrent/tt-umd/pull/1219

PR #1212, opened 2025-08-22
Fix flaky warm reset test for WH

nbuncicTT:
### Issue
#1054 

### Description
The device seems to non-deterministically fail after hanging the NOC on WH, hence this PR resolves the problem by only verifying NOC hang reset against BH. For WH the test is to write into a scratch register and then do the reset to see if the register value is reset. This test is also included for BH.

### List of the changes
- Added test for WH and BH for scratch register
- Modified test so that WH is not being tested against NOC hang
- Modified arch implementation class so that it's uniform when accessing ARC registers via BAR

### Testing
Manual run on CI for all tests

### API Changes
/


broskoTT (approved):
Nice!

Issue #1136, opened 2025-07-28
Polish TTDevice constructors


broskoTT:
So we have conflicting designs here.
There is TTDevice::create which accepts pci_device num
Then there is WormholeTTDevice which has public consturctor
Then there's RemoteWormholeTTDevice which has public constructor.

Figure out the proper abstraction here.

broskoTT:
@nbuncicTT is thinking about this, he plans to come up with a design fixing this issue

PR #1220, opened 2025-08-25
Refactor device creation

nbuncicTT:
### Issue
#1136 

### Description
Make the only interface to make TTDevice's through TTDevice factory `create`.

### List of the changes
Made:
- Constructors private
  - `BlackholeTTDevice`
  - `RemoteWormholeTTDevice`
- `WormholeTTDevice` constructor protected, as `RemoteWormholeTTDevice` needs to invoke it directly (should be dealt with in a separate PR)
- Changed every instance of creation of all derived `TTDevice` instances to use `create`
- Changed `RemoteChip` to use `TTDevice` instead of `RemoteWormholeTTDevice` and used dynamic cast (should be also dealt with in a separate PR)

### Testing
Manual

### API Changes
/

### DIFF
--- a/device/api/umd/device/tt_device/tt_device.h
+++ b/device/api/umd/device/tt_device/tt_device.h
@@ -38,6 +38,7 @@ struct dynamic_tlb {
 
 class ArcMessenger;
 class ArcTelemetryReader;
+class RemoteCommunication;
 
 class TTDevice {
 public:
@@ -49,6 +50,8 @@ public:
      * Jtag support can be enabled.
      */
     static std::unique_ptr<TTDevice> create(int device_number, IODeviceType device_type = IODeviceType::PCIe);
+    static std::unique_ptr<TTDevice> create(
+        std::unique_ptr<RemoteCommunication> remote_communication, eth_coord_t target_chip);
 
     TTDevice(std::shared_ptr<PCIDevice> pci_device, std::unique_ptr<architecture_implementation> architecture_impl);
     TTDevice(
--- a/device/chip/remote_chip.cpp
+++ b/device/chip/remote_chip.cpp
@@ -10,6 +10,8 @@
 
 #include "assert.hpp"
 #include "umd/device/chip/local_chip.h"
+#include "umd/device/tt_device/remote_wormhole_tt_device.h"
+#include "umd/device/tt_device/tt_device.h"
 #include "umd/device/wormhole_implementation.h"
 
 namespace tt::umd {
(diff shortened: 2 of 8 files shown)


device/api/umd/device/tt_device/tt_device.h
```diff
@@ -49,6 +50,8 @@ class TTDevice {
      * Jtag support can be enabled.
      */
     static std::unique_ptr<TTDevice> create(int device_number, IODeviceType device_type = IODeviceType::PCIe);
+    static std::unique_ptr<TTDevice> create(
```
broskoTT, line 53: I imagine the idea is to unify these two create functions, and then accept a number of arguments or some form of options struct, which would tell it to create local/remote, pci/jtag, etc?

device/chip/remote_chip.cpp
```diff
@@ -58,17 +60,26 @@ std::unique_ptr<RemoteChip> RemoteChip::create(
...
 
 RemoteChip::RemoteChip(
-    tt_SocDescriptor soc_descriptor, LocalChip* local_chip, std::unique_ptr<RemoteWormholeTTDevice> remote_tt_device) :
+    tt_SocDescriptor soc_descriptor, LocalChip* local_chip, std::unique_ptr<TTDevice> remote_tt_device) :
     Chip(remote_tt_device->get_chip_info(), soc_descriptor), local_chip_(local_chip) {
-    remote_communication_ = remote_tt_device->get_remote_communication();
+    // Architectural design issue - this dynamic_cast reveals a leaky abstraction.
+    // The base TTDevice interface should provide access to RemoteCommunication directly,
+    // rather than requiring knowledge of the concrete RemoteWormholeTTDevice type.
+    // This violates the Liskov Substitution Principle and creates tight coupling.
```
broskoTT, line 76: Vi govorite tako narodnim jezikom, a ujedno i književnim

broskoTT (approved):
I'm surprised that you didn't have to change to ::create at more places.
Nice refactor!

device/chip/remote_chip.cpp
nbuncicTT: Yeah, maybe the comment is a bit too much 🤣  but it emphasizes how this code should evolve, thanks for the insight from your side.

device/api/umd/device/tt_device/tt_device.h
nbuncicTT: For me, having an overloaded set of functions is as elegant as having one function with a set of input parameters which define the different behavior. 

Depending on how different the implementation is for each variant of the creation, I would opt out for one of them:
- If the creation is very similar for each protocol (like JTAG and PCIe) I'd prefer to have one function
- If the creation is very different and we would have complex logic in each switch case, I'd opt out for an overloaded set of functions

For now, the complexity is fairly low and we can do both. Which do you prefer?

PR #1221, opened 2025-08-25
PCI bdf modification

broskoTT:
### Description
Slight modification to how bdf is calculated. This now matches what tt-smi expects, and still is used for pci physical slot needed for 6u.

### List of the changes
- Change bdf calculation to include .function
- Expose it in PCI Device Info
- Use it for try_read_config_byte

### Testing
Existing CI tests

### API Changes
There is an API change in that now the PCIDeviceInfo exposes pci_bdf

### DIFF
--- a/device/pcie/pci_device.cpp
+++ b/device/pcie/pci_device.cpp
@@ -108,12 +108,7 @@ static bool detect_iommu(const PciDeviceInfo &device_info) {
 }
 
 static std::optional<uint8_t> try_read_config_byte(const PciDeviceInfo &device_info, size_t offset) {
-    const auto config_path = fmt::format(
-        "/sys/bus/pci/devices/{:04x}:{:02x}:{:02x}.{:x}/config",
-        device_info.pci_domain,
-        device_info.pci_bus,
-        device_info.pci_device,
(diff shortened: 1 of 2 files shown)


device/pcie/pci_device.cpp
```diff
@@ -187,14 +183,17 @@ static PciDeviceInfo read_device_info(int fd) {
...
     return PciDeviceInfo{
         info.out.vendor_id,
         info.out.device_id,
         info.out.pci_domain,
         bus,
         dev,
         fn,
-        get_physical_slot_for_pcie_bdf(get_pci_bdf(info.out.pci_domain, bus, dev))};
+        pci_bdf,
+        get_physical_slot_for_pcie_bdf(pci_bdf)};
```
pjanevskiTT, line 196: make sure getting physical slot still works. If not, we may need to separate functions for different version of pcie bdf

pjanevskiTT (approved):

device/pcie/pci_device.cpp
broskoTT: Fixed it.
It was missing the function part

Issue #756, opened 2025-04-16
Implement remote communication for blackhole


broskoTT:
https://tenstorrent.sharepoint.com/:p:/r/sites/Jasmina/_layouts/15/doc2.aspx?sourcedoc=%7B89D48D02-563B-4DF1-B598-4FFB725FB6F3%7D&file=P300%20tunneling.pptx&action=edit&mobileredirect=true

pjanevskiTT:
closed by #1223

PR #1223, opened 2025-08-25
Add initial lite fabric support

pjanevskiTT:
### Issue

https://github.com/tenstorrent/tt-umd/issues/756

### Description

Add initial support for lite fabric. Lite fabric is the firmware that is loaded onto ETH cores connecting 2 chips which allows two blackhole chips to communicate, same as with old routing FW. Lite fabric support is still not integrated into main UMD code paths right now, the idea is just to add the classes needed and then integrate it into main Cluster path.

Two main code paths for lite fabric that are added are

- `lite_fabric_host_utils.cpp `- This is a code that loads lite fabric FW to active ETH cores and wait for it to get initialized. After `launch_lite_fabric` call, it should be ready to do IO
- ` lite_fabric.hpp` - In this file LiteFabricHostInterface is implemented. It knows the protocol on how to talk to lite fabric FW in order to issue reads/writes to remote chips

Other hpp files are with constants copied over from metal so we can create all lite fabric related structures

### List of the changes

- Add lite fabric bin and elf
- Add lite fabric loading utils 
- Add lite fabric communication interface
- Add tests 

### Testing
Manual testing on P300 with both chips visible over PCIe so we can read/write both over PCIe and lite fabric.

### API Changes
/


device/api/umd/device/lite_fabric/lite_fabric.hpp
```diff
@@ -0,0 +1,472 @@
...
+
+    // Set to 1 to enable routing.
+    volatile uint32_t routing_enabled = 1;
+} __attribute__((packed));
+
+static_assert(sizeof(LiteFabricConfig) % 16 == 0);
+static_assert(offsetof(LiteFabricConfig, primary_local_handshake) % 16 == 0);
+static_assert(offsetof(LiteFabricConfig, neighbour_handshake) % 16 == 0);
+
+class HostToLiteFabricReadEvent {
```
broskoTT, line 130: To better understand what is editable and what not, can you add an explanation to the PR description explaining which header/implementation files are effectively "third party" for us, and which classes are our own?

lite_fabric.bin
broskoTT, line 1: How does the code even work if you haven't added this to any CMakeLists?

broskoTT, line 1: Please add this file to a separate "lib/" folder, like we used to have prior to https://github.com/tenstorrent/tt-umd/pull/902/files

tests/api/test_lite_fabric.cpp
```diff
@@ -0,0 +1,153 @@
...
+    auto chip = LocalChip::create(pci_devices_ids[0]);
+
+    if (pci_devices_ids.size() < 2 || chip->get_tt_device()->get_arch() != tt::ARCH::BLACKHOLE) {
+        return true;
+    }
+
+    return false;
+}
+
+TEST_F(LiteFabricFixture, FabricReadWrite4Bytes) {
```
broskoTT, line 93: Thanks for the tests!

tests/api/test_lite_fabric.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+        if (pci_devices_ids.size() < 2) {
+            GTEST_SKIP() << "Skipping lite fabric tests. Lite fabric tests require at least two Blackhole devices to "
+                            "be connected to the host.";
+        }
+
+        fabric_chip = LocalChip::create(pci_devices_ids[0]);
+
+        if (fabric_chip->get_tt_device()->get_arch() != tt::ARCH::BLACKHOLE) {
+            GTEST_SKIP() << "Skipping lite fabric tests. Lite fabric tests require at least two Blackhole devices to "
```
broskoTT, line 40: at least two? 
Make the check and the comment match

tests/api/test_lite_fabric.cpp
```diff
@@ -0,0 +1,191 @@
...
+                            "be connected to the host.";
+        }
+
+        auto eth_cores = fabric_chip->get_soc_descriptor().get_cores(CoreType::ETH, CoordSystem::TRANSLATED);
+        eth_cores_up.clear();
+        for (auto& eth_core : eth_cores) {
+            uint32_t port_status;
+            fabric_chip->read_from_device_reg(eth_core, &port_status, 0x7CC04, sizeof(port_status));
```
broskoTT, line 48: Should this constant live somewhere else?

tests/api/test_lite_fabric.cpp
```diff
@@ -0,0 +1,191 @@
...
+        fabric_chip->set_barrier_address_params(
+            {l1_mem::address_map::L1_BARRIER_BASE, eth_l1_mem::address_map::ERISC_BARRIER_BASE, 0});
+
+        non_fabric_chip = LocalChip::create(pci_devices_ids[1]);
+        eth_core_transfer = eth_cores_up[0];
+    }
+
+    void SetUp() override {
+        if (should_skip_lite_fabric_tests()) {
+            GTEST_SKIP()
```
broskoTT, line 64: Can you put the GTEST_SKIP inside of the should_skip_lite_fabric_tests function?

tests/api/test_lite_fabric.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+std::unique_ptr<LocalChip> LiteFabricFixture::fabric_chip = nullptr;
+std::unique_ptr<LocalChip> LiteFabricFixture::non_fabric_chip = nullptr;
+std::vector<CoreCoord> LiteFabricFixture::eth_cores_up = {};
+CoreCoord LiteFabricFixture::tensix_core = CoreCoord(1, 2, CoreType::TENSIX, CoordSystem::TRANSLATED);
+// Dummy value, it will be overriden inside SetUpTestSuite.
+CoreCoord LiteFabricFixture::eth_core_transfer = CoreCoord(0, 0, CoreType::ETH, CoordSystem::TRANSLATED);
+
+TEST_F(LiteFabricFixture, FabricReadWrite4Bytes) {
+    std::cout << "i am running fabric read write 4 bytes test" << std::endl;
```
broskoTT, line 105: Remove or make more informative where you print other stuff. Also change to a log if you keep it

device/api/umd/device/lite_fabric/lite_fabric.hpp
broskoTT: So if most of these are copied header files, can we move them to some other folder? We do have some firmware constants outside device folder.
Maybe we can put all of these with the *.bin file in some lite_fabric folder? At the root of the repo?
Also, these probably don't belong in the public UMDs API

device/lite_fabric/lite_fabric_host_utils.cpp
```diff
@@ -0,0 +1,142 @@
...
+namespace {
+uint32_t get_state_address() {
+    return LITE_FABRIC_CONFIG_START + offsetof(tt::umd::lite_fabric::LiteFabricConfig, current_state);
+}
+
+}  // namespace
+
+namespace tt::umd {
+
+namespace lite_fabric {
```
broskoTT, line 23: Can you maybe put this in a LiteFabric class, with static implementation or non-static implementation, whatever makes sense?

device/lite_fabric/lite_fabric_host_utils.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    chip->l1_membar(eth_cores_set);
+
+    for (auto tunnel_1x : eth_cores) {
+        wait_for_state(chip, tunnel_1x, get_state_address(), InitState::READY);
+        log_info(LogSiliconDriver, "Lite Fabric ready on core ({}, {})", tunnel_1x.x, tunnel_1x.y);
+    }
+}
+
+void terminate_lite_fabric(Chip* chip, const std::vector<CoreCoord>& eth_cores) {
```
broskoTT, line 129: You mentioned something like using TTDevice instead of Chip for lite fabric, specifically regarding l1_membar.
Are these changes comping up or did you drop them?

device/lite_fabric/lite_fabric_host_utils.cpp
```diff
@@ -0,0 +1,142 @@
...
+    bin_file.close();
+
+    return binary_data;
+}
+
+uint32_t get_eth_channel_mask(Chip* chip, const std::vector<CoreCoord>& eth_cores) {
+    uint32_t mask = 0;
+    for (const auto& eth_core : eth_cores) {
+        CoreCoord logical_core = chip->get_soc_descriptor().translate_coord_to(eth_core, CoordSystem::LOGICAL);
+        mask |= 0x1 << logical_core.y;
```
broskoTT, line 46:
```suggestion
        mask |= 0x1 << chip->get_soc_descriptor().get_eth_channel_for_core(eth_core);
```

device/lite_fabric/lite_fabric_host_utils.cpp
```diff
@@ -0,0 +1,142 @@
...
+    } else {
+        reset_val = TENSIX_DEASSERT_SOFT_RESET &
+                    static_cast<TensixSoftResetOptions>(
+                        ~std::underlying_type<TensixSoftResetOptions>::type(TensixSoftResetOptions::TRISC0));
+
+        chip->send_tensix_risc_reset(eth_core, reset_val);
+    }
+}
+
+void set_pc(Chip* chip, CoreCoord eth_core, uint32_t pc_addr, uint32_t pc_val) {
```
broskoTT, line 68: Which PC? is this something to put in TTDevice interface?

device/lite_fabric/lite_fabric_host_utils.cpp
```diff
@@ -0,0 +1,142 @@
...
+    config.routing_enabled = true;
+
+    // Need an abstraction layer for Lite Fabric
+    auto config_addr = LITE_FABRIC_CONFIG_START;
+
+    for (const auto& tunnel_1x : eth_cores) {
+        set_reset_state(chip, tunnel_1x, true);
+        set_pc(chip, tunnel_1x, k_PcResetAddress, k_FirmwareStart);
+
+        std::vector<uint8_t> binary_data = read_binary_file("device/libs/lite_fabric.bin");
```
broskoTT, line 101: Yeah we're going to have problems with the file which we keep alongside our lib...
I'm fine if you want to merge this like this, but you'll have problem once we start using it in UMD's libdevice. You'll see failures in our CI, then on metal CI.

device/lite_fabric/lite_fabric_host_utils.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    std::unordered_set<CoreCoord> eth_cores_set(eth_cores.begin(), eth_cores.end());
+
+    chip->l1_membar(eth_cores_set);
+
+    for (auto tunnel_1x : eth_cores) {
+        set_reset_state(chip, tunnel_1x, false);
+    }
+
+    chip->l1_membar(eth_cores_set);
```
broskoTT, line 121: set_reset_state uses send_tensix_risc_reset which should not use TLBs which return before the value landed. Can you remove this l1_membar?

device/lite_fabric/lite_fabric_host_utils.cpp
```diff
@@ -0,0 +1,142 @@
...
+        std::vector<uint8_t> binary_data = read_binary_file("device/libs/lite_fabric.bin");
+        size_t bin_size = binary_data.size();
+
+        // Set up configuration
+        config.binary_addr = k_FirmwareStart;
+        config.binary_size = (bin_size + 15) & ~0xF;
+
+        chip->write_to_device(tunnel_1x, (void*)&config, config_addr, sizeof(lite_fabric::LiteFabricConfig));
+
+        chip->write_to_device(tunnel_1x, binary_data.data(), k_FirmwareStart, bin_size);
```
broskoTT, line 110: Is it worth extracting this to some other common file, outside of lite fabric, which provides functionality of putting a binary on a core?

Maybe not if you plan to work on unifying this from multiple repos...

tests/api/test_lite_fabric.cpp
```diff
@@ -0,0 +1,191 @@
...
+        non_fabric_chip->write_to_device(tensix_core, zero_data.data(), 0, zero_data.size());
+    }
+
+    void TearDown() override {
+        if (fabric_chip.get() != nullptr) {
+            lite_fabric::terminate_lite_fabric(fabric_chip.get(), eth_cores_up);
+        }
+    }
+
+    bool should_skip_lite_fabric_tests() {
```
broskoTT, line 80: Not sure if this is enough to figure out if the test is supposed to work or not? Do you need more checks, like checking if the board is p150, if there are eth connections

broskoTT (approved):
Glad to see progress on this!

device/api/umd/device/lite_fabric/lite_fabric.hpp
```diff
@@ -0,0 +1,470 @@
...
+    uint32_t eth_barrier_addr = 0;
+    uint32_t tensix_barrier_addr = 0;
+    uint32_t l1_alignment_bytes = 0;
+    // The core to process requests.
+    uint32_t mmio_device_id = 0;
+    uint32_t mmio_eth_core_x = 0;
+    uint32_t mmio_eth_core_y = 0;
+    TTDevice* tt_device = nullptr;
+
+    inline void init() volatile {
```
nhuang-tt, line 161: I'm not sure if this function is used. I made this function but I don't think I used it at all so maybe copied here?

To make it more robust, probably for the next PR, each time we use read/write on a HostInterface it might be smart to get the h2d and d2h index from the device rather than relying on syncing host/device to remove the reliance on the HostInterface instance. This will also help with having a single source of truth.

device/api/umd/device/lite_fabric/lite_fabric.hpp
```diff
@@ -0,0 +1,470 @@
...
+                    "mismatch {:#x} != {:#x}",
+                    core_type_name,
+                    virtual_core.x,
+                    virtual_core.y,
+                    read_barrier,
+                    barrier_value));
+            }
+        };
+
+        CoreCoord barrier_coord = CoreCoord(1, 2, CoreType::TENSIX, CoordSystem::TRANSLATED);
```
nhuang-tt, line 201: Is this expected to only do a barrier to the TENSIX core and not the other cores like ETH?

tests/api/test_lite_fabric.cpp
```diff
@@ -0,0 +1,191 @@
...
+    host_interface.write(write_data.data(), write_data.size(), eth_core_transfer, tensix_core, test_addr);
+
+    host_interface.barrier(eth_core_transfer);
+
+    std::vector<uint8_t> readback_data(1 << 20, 0);
+    non_fabric_chip->read_from_device(tensix_core, readback_data.data(), test_addr, readback_data.size());
+    EXPECT_EQ(write_data, readback_data);
+}
+
+TEST_F(LiteFabricFixture, FabricARC) {
```
nhuang-tt, line 176: 👍

nhuang-tt (approved):

device/api/umd/device/lite_fabric/lite_fabric_header.hpp
```diff
@@ -0,0 +1,167 @@
...
+    inline FabricLiteHeader& to_write_reg(const WriteRegCommandHeader& write_reg_command_header) {
+        this->noc_send_type = lite_fabric::NocSendType(lite_fabric::NocSendTypeEnum::WRITE_REG, 0);
+        this->command_fields.write_reg = write_reg_command_header;
+        return *static_cast<FabricLiteHeader*>(this);
+    }
+
+    // Set the number of hops along the line for this packet to the target chip
+    inline FabricLiteHeader& to_chip_unicast(uint8_t distance_in_hops) {
+        // LowLatencyPacketHeader::calculate_chip_unicast_routing_fields_value
+        uint32_t value =
```
nbuncicTT, line 148: Is there a more elegant way to generate this value?

device/api/umd/device/lite_fabric/lite_fabric_header.hpp
```diff
@@ -0,0 +1,167 @@
...
+    uint64_t noc_address;
+    uint64_t event;
+};
+
+struct WriteRegCommandHeader {
+    uint32_t reg_address;
+    uint32_t reg_value;
+};
+
+union LiteFabricCommandFields {
```
nbuncicTT, line 29: Consider using std::variant instead of a plain union, it's essentially a type-safe union

nbuncicTT (approved):
Nice job!

tests/api/test_lite_fabric.cpp
pjanevskiTT: This is the same for now for all blackhole eth constants, I will make a single PR to address all of it

device/api/umd/device/lite_fabric/lite_fabric.hpp
pjanevskiTT: it's used in UMD. Makes sense for indices, will address it in follow up PR https://github.com/tenstorrent/tt-umd/issues/1307

tests/api/test_lite_fabric.cpp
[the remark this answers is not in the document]
pjanevskiTT: calling gtest_skip in test suite for some reason continues set up for each file, left it as it is for now

device/lite_fabric/lite_fabric_host_utils.cpp
```diff
@@ -0,0 +1,142 @@
...
+    uint32_t mask = 0;
+    for (const auto& eth_core : eth_cores) {
+        CoreCoord logical_core = chip->get_soc_descriptor().translate_coord_to(eth_core, CoordSystem::LOGICAL);
+        mask |= 0x1 << logical_core.y;
+    }
+    return mask;
+}
+
+void set_reset_state(Chip* chip, CoreCoord eth_core, bool assert_reset) {
+    // Lite fabric on blackhole runs on DM1. Don't touch DM0. It is running base firmware.
```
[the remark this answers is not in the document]
pjanevskiTT: data movement

device/api/umd/device/lite_fabric/lite_fabric_header.hpp
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/1307

device/api/umd/device/lite_fabric/lite_fabric.hpp
pjanevskiTT: added barrier to ETH as well

PR #1224, opened 2025-08-25
Add handling for hidden intermesh eth links

tt-aho:
### Issue
(Link to Github issue(s))

### Description
For multihost systems where T3Ks are connected to each other, we have special SPI programming that makes the links show up as unconnected through the standard port status check, and there is an alternative path to query if these intermesh eth links are up/active. This was to originally prevent UMD and tt-smi from exploring the remote hosts, though now this is only an issue with tt-smi, not UMD.
Once tt-smi's backend is switched to UMD then we will no longer need to do this hiding of eth links, however, to have code that handles eth links in higher level stacks like tt-metal be agnostic of this, we want to push this detection code to UMD until we can get rid of the SPI programming.

### List of the changes
Add `extract_intermesh_eth_links` and `is_intermesh_eth_link_trained` apis to topology discovery (no-op on BH).
`extract_intermesh_eth_links` will return the list of links that are intermesh, and `is_intermesh_eth_link_trained` will return if the specified link is actually active/trained.

### Testing

### API Changes

### DIFF
--- a/device/api/umd/device/topology/topology_discovery.h
+++ b/device/api/umd/device/topology/topology_discovery.h
@@ -93,6 +93,15 @@ protected:
     // does not take harvesting into consideration. This function will be overridden just for Blackhole.
     virtual void patch_eth_connections();
 
+    // Intermesh links are ethernet links that are turned off during UMD's topology discovery but are
+    // otherwise physically connected. This is done since not all tools support limiting the discovery as
+    // UMD does. Once all the tools start supporting this, this feature won't be used anymore and this
+    // function will return empty set.
+    // This will extract the list of intermesh links from a config in L1.
+    virtual std::vector<uint32_t> extract_intermesh_eth_links(Chip* chip, tt_xy_pair eth_core) = 0;
+
+    virtual bool is_intermesh_eth_link_trained(Chip* chip, tt_xy_pair eth_core) = 0;
+
     std::map<uint64_t, std::unique_ptr<Chip>> chips_to_discover;
     std::map<uint64_t, std::unique_ptr<Chip>> chips;
 
--- a/device/topology/topology_discovery.cpp
+++ b/device/topology/topology_discovery.cpp
@@ -104,13 +104,25 @@ void TopologyDiscovery::discover_remote_chips() {
             chip->get_soc_descriptor().get_cores(CoreType::ETH, umd_use_noc1 ? CoordSystem::NOC1 : CoordSystem::NOC0);
         TTDevice* tt_device = chip->get_tt_device();
 
+        std::vector<uint32_t> intermesh_eth_links;
+        if (eth_cores.size() > 0) {
+            intermesh_eth_links = extract_intermesh_eth_links(chip, eth_cores.front());
+        }
+
         uint32_t channel = 0;
         for (const CoreCoord& eth_core : eth_cores) {
             uint32_t port_status = read_port_status(chip, eth_core);
 
             if (is_eth_unknown(chip, eth_core) || is_eth_unconnected(chip, eth_core)) {
-                channel++;
-                continue;
+                if (std::find(intermesh_eth_links.begin(), intermesh_eth_links.end(), channel) ==
+                    intermesh_eth_links.end()) {
+                    channel++;
+                    continue;
+                }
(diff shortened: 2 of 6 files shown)


device/api/umd/device/topology/topology_discovery.h
```diff
@@ -93,6 +93,10 @@ class TopologyDiscovery {
     // does not take harvesting into consideration. This function will be overridden just for Blackhole.
     virtual void patch_eth_connections();
 
+    virtual std::vector<uint32_t> extract_intermesh_eth_links(Chip* chip, tt_xy_pair eth_core) = 0;
```
broskoTT, line 96:
```suggestion
    // Intermesh links are ethernet links that are turned off during UMDs topology discovery but are
    // otherwise physically connected. This is done since not all tools support limiting the discovery as
    // does UMD. Once all the tools start supporting this, this feature won't be used anymore and this
    // function will return empty set.
    virtual std::vector<uint32_t> extract_intermesh_eth_links(Chip* chip, tt_xy_pair eth_core) = 0;
```

device/topology/topology_discovery.cpp
```diff
@@ -104,13 +104,26 @@ void TopologyDiscovery::discover_remote_chips() {
...
+                if (std::find(intermesh_eth_links.begin(), intermesh_eth_links.end(), channel) !=
+                    intermesh_eth_links.end()) {
+                    if (!is_intermesh_eth_link_trained(chip, eth_core)) {
+                        channel++;
+                        continue;
+                    }
+                } else {
+                    channel++;
+                    continue;
+                }
```
broskoTT, line 126: I find it more readable and easier to follow the logic if not nested, and because of continue you don't need an else branch

```suggestion
                if (std::find(intermesh_eth_links.begin(), intermesh_eth_links.end(), channel) ==
                    intermesh_eth_links.end()) {
                    channel++;
                    continue;
                }
                if (!is_intermesh_eth_link_trained(chip, eth_core)) {
                    channel++;
                    continue;
                }
```

device/topology/topology_discovery_blackhole.cpp
```diff
@@ -162,4 +162,15 @@ void TopologyDiscoveryBlackhole::patch_eth_connections() {
     }
 }
 
+// Helper to extract intermesh ports from config value
```
broskoTT, line 165: Move this comment to header file

device/topology/topology_discovery_wormhole.cpp
```diff
@@ -30,6 +31,7 @@ TopologyDiscoveryWormhole::EthAddresses TopologyDiscoveryWormhole::get_eth_addre
     uint64_t erisc_remote_eth_id_offset;
 
     if (masked_version >= 0x060000) {
+        eth_param_table = 0x1000;
```
broskoTT, line 34: @pjanevskiTT 
In the future I might insist that we move all these constants and all such reads to TTDevice.
So it would have many simple functions, "read_intermesh_link_bits", "read_port_status", "read_remote_eth_information", etc. Simply, I'd remove all address constants and almost all "read_from_device" calls from this class.

Just sharing thoughts, no action for this PR.

broskoTT (commented):
I'll leave it to Pavle to approve, thanks for the change

pjanevskiTT (approved):
Leaving approve to unblock the work, so you can merge it as soon as CI passes. Looks good anyway

device/topology/topology_discovery.cpp
```diff
@@ -104,13 +104,25 @@ void TopologyDiscovery::discover_remote_chips() {
...
+        }
+
         uint32_t channel = 0;
         for (const CoreCoord& eth_core : eth_cores) {
             uint32_t port_status = read_port_status(chip, eth_core);
 
             if (is_eth_unknown(chip, eth_core) || is_eth_unconnected(chip, eth_core)) {
-                channel++;
-                continue;
+                if (std::find(intermesh_eth_links.begin(), intermesh_eth_links.end(), channel) ==
```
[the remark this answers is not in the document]
tt-aho: If both of these both of these are false then this would be an active link, so it will be handled/incremented later like the other active links.

Issue #1226, opened 2025-08-26
Create a release process


broskoTT:
Release process should be defined and automated.
luwen's release process could be reproduced.

Issue #1178, opened 2025-08-13
UMD versioning


broskoTT:
Figure out where to place UMD version and make it used for versioning of the base library and python library

PR #1227, opened 2025-08-26
Release automation

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/1178
https://github.com/tenstorrent/tt-umd/issues/1226

### Description
Release automation workflow, which builds all artifacts, and spawns all content for the release.
The automation also sends a message to slack. And it is triggered when VERSION file is changed on main branch.

### List of the changes
- Introduced VERSION file which holds raw version
- Consume version in CMakeLists and pyproject.toml
- CHANGELOG changed to affect new releases
- Added automation workflow which does a few things
  - Builds artefacts (cpp build archives and python wheel)
  - Extracts a version, extracts changelog from the CHANGELOG file. Gets info from git commits for more detailed changelog info
  - Creates a draft release (which should be manually confirmed)
  - Sends a slack message with release info
  - The workflow will run on all pushes to main branch which change VERSION file

### Testing
Here's an example run: https://github.com/tenstorrent/tt-umd/actions/runs/17238217236/job/48908112142
Generating and example release: https://github.com/tenstorrent/tt-umd/releases/tag/untagged-69d6c36526835200eb3c

### API Changes


.github/workflows/release.yml
```diff
@@ -0,0 +1,488 @@
+# Create automated release when VERSION file is updated on main branch
+name: Automated Release
+
+on:
+  push:
+    branches:
```
vmilosevic, line 6: It is completely fine to trigger this on push to main, but perhaps is more common practise to trigger release build on tag push
```
on:
  push:
    tags:
      - 'v*.*.*'
```
Tagging with version makes it easy to link release binaries to the version of source code, while its harder to find first commit that set some value in VERSION file

CHANGELOG
```diff
@@ -1,16 +1,14 @@
 # Changelog
```
pjanevskiTT, line 1: What is the exact process of bumping the version? Can you add a readme on how is someone supposed to do it

pjanevskiTT (approved):
looks good

CHANGELOG
broskoTT: added readme

PR #1228, opened 2025-08-26
Switching how simulator and host connect to each other

tt-vjovanovic:
For aether emulation, we needed to switch to emulator connecting to UMD host (because we don't know machine name where emulation will be scheduled).
In order to have single UMD code, we switched for simulators as well.

### DIFF
--- a/device/api/umd/device/tt_simulation_host.hpp
+++ b/device/api/umd/device/tt_simulation_host.hpp
@@ -9,8 +9,6 @@
 
 #include "umd/device/tt_xy_pair.h"
 
-#define NNG_SOCKET_PREFIX "tcp://soc-zebu-01:5556"
-
 typedef struct nng_socket_s nng_socket;
 typedef struct nng_listener_s nng_listener;
 
--- a/device/api/umd/device/tt_simulation_host.hpp
+++ b/device/api/umd/device/tt_simulation_host.hpp
@@ -9,10 +9,10 @@
 
 #include "umd/device/tt_xy_pair.h"
 
-#define NNG_SOCKET_PREFIX "ipc:///tmp/"
+#define NNG_SOCKET_PREFIX "tcp://soc-zebu-01:5556"
 
 typedef struct nng_socket_s nng_socket;
-typedef struct nng_dialer_s nng_dialer;
+typedef struct nng_listener_s nng_listener;
 
 namespace tt::umd {
 
@@ -27,7 +27,7 @@ public:
 
 private:
     std::unique_ptr<nng_socket> host_socket;
-    std::unique_ptr<nng_dialer> host_dialer;
+    std::unique_ptr<nng_listener> host_listener;
 };
 
 }  // namespace tt::umd
--- a/device/simulation/tt_simulation_host.cpp
+++ b/device/simulation/tt_simulation_host.cpp
@@ -14,47 +14,81 @@
 #include <sstream>
 #include <tt-logger/tt-logger.hpp>
 #include <typeinfo>
+#include <random>
+#include <unistd.h>
 
 #include "assert.hpp"
 
 namespace tt::umd {
 
 tt_SimulationHost::tt_SimulationHost() {
-    // Initialize socket and dialer
+    // Initialize socket and listener
     host_socket = std::make_unique<nng_socket>();
-    host_dialer = std::make_unique<nng_dialer>();
+    host_listener = std::make_unique<nng_listener>();
 
-    // Get socket name, using PID to make it unique with multiple runs in parallel
-    const char *nng_socket_name = std::getenv("NNG_SOCKET_NAME") ? std::getenv("NNG_SOCKET_NAME") : "ttsim_nng_ipc";
+    // Check if NNG_SOCKET_LOCAL_PORT is set
+    const char *local_socket_port_str = std::getenv("NNG_SOCKET_LOCAL_PORT");
+    std::string nng_socket_addr_str;
+
+    // Generate socket address with hostname and random port
+    char hostname[256];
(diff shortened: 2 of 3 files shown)


device/simulation/tt_simulation_host.cpp
```diff
@@ -14,47 +14,81 @@
...
+    if (local_socket_port_str) {
+        port = atoi(local_socket_port_str);
+        log_info(tt::LogEmulationDriver, "Using specified NNG_SOCKET_LOCAL_PORT: {}", port);
+    } else {
+        // Generate random port in range 50000-59999
+        std::random_device rd;
+        std::mt19937 gen(rd());
+        std::uniform_int_distribution<> dis(50000, 59999);
+
+        port = dis(gen);
```
broskoTT, line 50: Maybe a while loop over ports until we find a free one? Meaning the first randomly generated one might be in use, right?

device/simulation/tt_simulation_host.cpp
```diff
@@ -14,47 +14,81 @@
...
-        }
-    } while (rv != 0);
+    // Start listening for connections from client
+    int rv = nng_listener_start(*host_listener, 0);
+    if (rv != 0) {
+        log_error(tt::LogEmulationDriver, "Failed to start listener: {}", nng_strerror(rv));
+        return;
+    }
+    
+    log_info(tt::LogEmulationDriver, "Server started, waiting for client to connect...");
```
broskoTT, line 90: So you start the listener, but what initiates the server (well actually, the client) on the zebu machine to try this connection?

device/simulation/tt_simulation_host.cpp
```diff
@@ -74,13 +108,13 @@ void tt_SimulationHost::send_to_device(uint8_t *buf, size_t buf_size) {
 size_t tt_SimulationHost::recv_from_device(void **data_ptr) {
     int rv;
     size_t data_size;
-    log_debug(tt::LogEmulationDriver, "Receiving messsage from remote..");
+    log_info(tt::LogEmulationDriver, "Receiving messsage from remote..");
```
broskoTT, line 111: Maybe these prints should be log_debug?

device/api/umd/device/tt_simulation_host.hpp
```diff
@@ -9,10 +9,10 @@
 
 #include "umd/device/tt_xy_pair.h"
 
-#define NNG_SOCKET_PREFIX "ipc:///tmp/"
+#define NNG_SOCKET_PREFIX "tcp://soc-zebu-01:5556"
```
broskoTT, line 12: Should we maybe make this configurable through an env variable in case it changes?

device/simulation/tt_simulation_host.cpp
```diff
@@ -14,47 +14,81 @@
...
-    // Initialize socket and dialer
+    // Initialize socket and listener
     host_socket = std::make_unique<nng_socket>();
-    host_dialer = std::make_unique<nng_dialer>();
+    host_listener = std::make_unique<nng_listener>();
 
-    // Get socket name, using PID to make it unique with multiple runs in parallel
-    const char *nng_socket_name = std::getenv("NNG_SOCKET_NAME") ? std::getenv("NNG_SOCKET_NAME") : "ttsim_nng_ipc";
+    // Check if NNG_SOCKET_LOCAL_PORT is set
+    const char *local_socket_port_str = std::getenv("NNG_SOCKET_LOCAL_PORT");
```
broskoTT, line 30: Can you add a brief comment in code in which cases is this used?
Is this for running the code on the zebu machine locally?

broskoTT (approved):
Thanks for building this!

device/api/umd/device/tt_simulation_host.hpp
```diff
@@ -9,10 +9,10 @@
 
 #include "umd/device/tt_xy_pair.h"
```
broskoTT, line 10: There is a guide here on how to run all pre-commit checks automatically, so that the job succeeds:
https://github.com/tenstorrent/tt-umd?tab=readme-ov-file#how-to-set-up-pre-commit-locally

dzivanovicTT:
Nice work!

We should test this manually first, since the changes affect both UMD and Simulator and are interdependent.
I’ll run the check on my side to confirm it doesn’t break the Versim Simulator.

device/simulation/tt_simulation_host.cpp
tt-vjovanovic: Yes, yes... This was just test printout...

device/api/umd/device/tt_simulation_host.hpp
tt-vjovanovic: This should be removed.

device/api/umd/device/tt_simulation_host.hpp
tt-vjovanovic: Thanks!

device/simulation/tt_simulation_host.cpp
broskoTT: I prob need to add a README on how to enable log_debug...

tt-vjovanovic: We were trying to do this from tt-exalens, but for some reason it didn't work as expected.
Our code for changing level is:
```
    tt::umd::logging::set_level(tt::umd::logging::level::debug);
```

dzivanovicTT:
> Nice work!
> 
> We should test this manually first, since the changes affect both UMD and Simulator and are interdependent. I’ll run the check on my side to confirm it doesn’t break the Versim Simulator.

I have tested Versim Wormhole and Blackhole Simulators with these changes and it works.

PR #1235, opened 2025-08-29
Add option to choose NOC translation enabled in simulation

pjanevskiTT:
### Issue

NOC translation enabled was fix to false for simulation and mock. Add option to choose this, both options should be valid

### Description

Add option to choose whether to turn on NOC translation for simulation and mock clusters.

### List of the changes

- Add option to choose NOC translation enabled in simulation and mock clusters

### Testing
CI

### API Changes
/


device/cluster.cpp
```diff
@@ -423,11 +423,12 @@ Cluster::Cluster(ClusterOptions options) {
...
 #ifdef TT_UMD_BUILD_SIMULATION
             if (options.chip_type == ChipType::SIMULATION) {
-                tt_SimulationDeviceInit init(options.simulator_directory);
+                tt_SimulationDeviceInit init(options.simulator_directory, options.simulation_noc_translation_enabled);
                 arch = init.get_soc_descriptor().arch;
             }
 #endif
-            cluster_desc = tt_ClusterDescriptor::create_mock_cluster(chips_to_construct_vec, arch);
+            cluster_desc = tt_ClusterDescriptor::create_mock_cluster(
+                chips_to_construct_vec, arch, options.simulation_noc_translation_enabled);
```
broskoTT, line 431: I've just confirmed, both our constructors, including the one accepting soc_descriptor yaml, will take noc_translation as passed by chip_info, which is constructed from cluster descriptor. 
So this should be fine, should be enough to pass the desired noc_translation to simulator.
We should consider in future if we want to honour what is passed through custom soc_descriptor, or if not to explicitly remove it from there to signal that.

mcraigheadTT:
Was able to pull this in and test, but ran into a regression where my tests on BH seem to have broken in the meantime and I'm not sure why

Issue #584, opened 2025-03-06
Cleanup different board types logic


pjanevskiTT:
Cleanup different board types logic when it comes to two things

1. UPI - add this to constants, not hardcoded numbers
2. Serializing from cluster yaml - we shouldn't care about upper/lower case, always put string to lowercase for example

PR #1242, opened 2025-09-01
Moved board UPI to unordered_map

aleksamarkovicTT:
### Issue
#584

### Description
Moved board UPIs to a hash table instead of using hardcoded numbers in an if-else clause.

### List of the changes
- Moved board UPIs to a hash table instead of using hardcoded numbers in an if-else clause.

### Testing
CI

### API Changes

### DIFF
--- a/device/api/umd/device/types/cluster_descriptor_types.hpp
+++ b/device/api/umd/device/types/cluster_descriptor_types.hpp
@@ -186,26 +186,34 @@ inline uint32_t get_number_of_chips_from_board_type(const BoardType board_type) 
     }
 }
 
+inline const std::unordered_map<uint64_t, BoardType> board_upi_map = {
+    // P100
+    {0x36, BoardType::P100},
+    {0x43, BoardType::P100},
+    // P150
+    {0x40, BoardType::P150},
+    {0x41, BoardType::P150},
+    {0x42, BoardType::P150},
+    // P300
+    {0x44, BoardType::P300},
+    {0x45, BoardType::P300},
+    {0x46, BoardType::P300},
+    // N150
+    {0x18, BoardType::N150},
+    // N300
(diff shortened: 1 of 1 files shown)


device/api/umd/device/types/cluster_descriptor_types.hpp
```diff
@@ -186,25 +186,34 @@ inline uint32_t get_number_of_chips_from_board_type(const BoardType board_type)
     }
 }
 
+inline const std::unordered_map<uint64_t, BoardType> board_upi_map = {
+    // P100
```
pjanevskiTT, line 190: I feel it's clear from the code for which board these constants are. I would potentially add a comment on why we have multiple constants for P100. In general we should try and focus on adding the comments on "why" we are doing something, not "what" we are doing

device/api/umd/device/types/cluster_descriptor_types.hpp
```diff
@@ -186,25 +186,34 @@ inline uint32_t get_number_of_chips_from_board_type(const BoardType board_type)
...
+    {0x43, BoardType::P100},
+    // P150
+    {0x40, BoardType::P150},
+    {0x41, BoardType::P150},
+    {0x42, BoardType::P150},
+    // P300
+    {0x44, BoardType::P300},
+    {0x45, BoardType::P300},
+    {0x46, BoardType::P300},
+    // N150
```
pjanevskiTT, line 201: no need for a comment in places where we have only a single constant for a board

pjanevskiTT (approved):
Looks good, left minor comments

device/api/umd/device/types/cluster_descriptor_types.hpp
aleksamarkovicTT: Ack

Issue #1210, opened 2025-08-22
ETH training wait for all cores


broskoTT:
I realised this functionality gap while discovering https://github.com/tenstorrent/tt-umd/issues/1208

It seems like our timeout of 60s is applied PER CORE, which a huge amount of time. Maybe we need to have a timeout for all of them? Figure out how to implement this (I know it is harder since this function is in tt_device)

PR #1243, opened 2025-09-01
Count ETH training timeout for all cores

aleksamarkovicTT:
### Issue
#1210

### Description
On chip initialization, the ETH training timeout of 60s was counted separately for every core (or every step in checking completion for every core in WH). This PR introduces a global 60s timeout for an entire chip. 
Although there is no ETH training length distribution data, having a chip-bound timeout is better than having a shorter per-core timeout assuming high variance in per-core training time lengths.

### List of the changes
- Functions waiting on ETH training completion for WH, BH return time taken to complete.
- `Chip::wait_eth_cores_training()` counts down time taken by each core and reduces timeout for the subsequent call.

### Testing
CI

### API Changes

### DIFF
--- a/device/api/umd/device/tt_device/blackhole_tt_device.hpp
+++ b/device/api/umd/device/tt_device/blackhole_tt_device.hpp
@@ -43,7 +43,7 @@ public:
 
     ChipInfo get_chip_info() override;
 
-    void wait_eth_core_training(const tt_xy_pair eth_core, const uint32_t timeout_ms = 60000) override;
+    uint32_t wait_eth_core_training(const tt_xy_pair eth_core, const uint32_t timeout_ms = 60000) override;
 
     uint64_t get_arc_noc_base_address() const override;
 
--- a/device/api/umd/device/tt_device/tt_device.hpp
+++ b/device/api/umd/device/tt_device/tt_device.hpp
@@ -230,7 +230,11 @@ public:
      */
     virtual void wait_arc_core_start(const uint32_t timeout_ms = 1000) = 0;
 
-    virtual void wait_eth_core_training(const tt_xy_pair eth_core, const uint32_t timeout_ms = 60000) = 0;
+    /**
+     * Waits for ETH core training to complete.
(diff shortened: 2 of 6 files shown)


device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -230,7 +230,11 @@ class TTDevice {
      */
     virtual void wait_arc_core_start(const uint32_t timeout_ms = 1000) = 0;
 
-    virtual void wait_eth_core_training(const tt_xy_pair eth_core, const uint32_t timeout_ms = 60000) = 0;
+    /**
```
pjanevskiTT, line 233: good for adding docs. Since you are already doing it, can you follow some general rules on how we do it (not that these are documented somewhere...) but in general use 
```
@param param_name param explanation
@returns what does function return
```

pjanevskiTT: document all params please

device/api/umd/device/tt_device/blackhole_tt_device.hpp
```diff
@@ -43,7 +43,7 @@ class BlackholeTTDevice : public TTDevice {
 
     ChipInfo get_chip_info() override;
 
-    void wait_eth_core_training(const tt_xy_pair eth_core, const uint32_t timeout_ms = 60000) override;
+    uint32_t wait_eth_core_training(const tt_xy_pair eth_core, const uint32_t timeout_ms = 60000) override;
```
pjanevskiTT, line 46: One thing I would be interested in is how much time does actual ETH training take. Is this number still too big, maybe we can reduce it by a lot such that even when training fails we don't have to wait this much

Could you maybe test that out just to get a sense if our order of magnitude is correct here? You could test that by issuing a reset and then running any test that involves waiting for ETH cores to reset

pjanevskiTT (commented):
looks good, I am just interested in timeout constant for all cores training

PR #1247, opened 2025-09-02
TTSim CI

broskoTT:
### Issue
#1277 

### Description
Establish a CI for ttsim. Build and run commands are fairly simple.
Note that the ttsim ref is hardcoded in tt_metal workflow, we extract that ref, so that new changes on ttsim don't unintentionally break the build.

### List of the changes
- Add ttsim build client workflow
- Runs only wh for now, bh is not passing.

### Testing
The added CI passes

### API Changes


.github/workflows/build-clients.yml
```diff
@@ -82,3 +82,103 @@ jobs:
...
+      - name: Prepare sim paths
+        run: |
+          echo "Preparing sim paths"
+          mkdir -p sim_wh sim_bh
+          cp ttsim/src/_out/release_wh/libttsim.so sim_wh/libttsim.so
+          cp ttsim/src/_out/release_bh/libttsim.so sim_bh/libttsim.so
+          cp $TT_METAL_HOME/tt_metal/soc_descriptors/wormhole_b0_80_arch.yaml sim_wh/soc_descriptor.yaml
+          cp $TT_METAL_HOME/tt_metal/soc_descriptors/blackhole_140_arch.yaml sim_bh/soc_descriptor.yaml
+
+      - name: Run tt-sim wormhole tests
```
pjanevskiTT, line 175: I would probably put this into a separate yaml. I would like to have build and run in different jobs in CI. Probably call it build_clients and run_clients test. Otherwise I think this yaml is good

.github/workflows/build-clients.yml
```diff
@@ -82,3 +82,103 @@ jobs:
...
+        run: |
+          export TT_METAL_SIMULATOR=$(pwd)/sim_wh/libttsim.so
+          TT_METAL_SLOW_DISPATCH_MODE=1 $TT_METAL_HOME/build/programming_examples/metal_example_add_2_integers_in_riscv
+
+# TODO: Enable this once it starts passing.
```
pjanevskiTT, line 180: Just a reminder to do this

broskoTT: BH still doesn't pass...

broskoTT: I meant I'll enable it once it starts passing from ttsim part

Issue #1237, opened 2025-08-29
Fix simulator and host connection PR


pjanevskiTT:
https://github.com/tenstorrent/tt-umd/pull/1236#pullrequestreview-3169124504 reverted a commit that changed communication between host and simulator because tt-sim was failing

Check in this commit again with proper fix for both VCS and tt-sim

broskoTT:
This change breaks our tt-umd-simulators repo.

@tt-vjovanovic can you help @mcraigheadTT implement the new protocol in tt-sim?

mcraigheadTT:
I'm confused, why doesn't the pipe work?  I would prefer to not switch to socket, as this is extra work, could cause various potential issues (e.g. may regress performance), would require significant testing, and doesn't strike me as the right long term direction.  I believe the right long term direction is to get rid of the IPC.

mcraigheadTT:
I haven't gotten around to it yet, but my plan this upcoming week was to remove the IPC altogether for ttsim and have everything run in 1 process.

mcraigheadTT:
I suppose we could still support an IPC mode in ttsim after that, though I'm not convinced that it is worth the trouble to continue supporting it.  I do know that the current IPC is very, very slow

broskoTT:
The fix is in UMD, so both codepaths work. 
Further development on this topic will be tracked by https://github.com/tenstorrent/tt-umd/issues/1254

PR #1250, opened 2025-09-02
Separate ttsim and Simulator path

broskoTT:
### Issue
Related to a couple of issues
Outdates https://github.com/tenstorrent/tt-umd/pull/1235
Related to https://github.com/tenstorrent/tt-umd/issues/1237
Also probably fixes https://github.com/tenstorrent/ttsim/issues/30 , at least to some extent
Also https://github.com/tenstorrent/ttsim/issues/15

### Description
Since we have different paths we want to take for the two simulators, separate them into two different classes. The expectation is that this might be diverging even further.

### List of the changes
- I've copied simulation_device, _host and test , and renamed with ttsim prefix
- Then reverted https://github.com/tenstorrent/tt-umd/pull/1236 , which should fix vcs simulator path
- Changed TTSim device to always use translation enabled
- Fixed an issue which always expected 0,0 in translated coords on send_tensix_reset, in both codepaths

### Testing
I've tested manually that both paths work, with tt_metal change: https://github.com/tenstorrent/tt-metal/pull/27763

### API Changes
This PR has API changes. It adds another class TTSim which should be used for the other simulator. Metal changes are accompanying this new usage

### DIFF
--- a/device/CMakeLists.txt
+++ b/device/CMakeLists.txt
@@ -86,6 +86,8 @@ if(TT_UMD_BUILD_SIMULATION)
         PRIVATE
             simulation/simulation_device.cpp
             simulation/simulation_host.cpp
+            simulation/ttsim_device.cpp
+            simulation/ttsim_host.cpp
             ${FBS_GENERATED_HEADER}
     )
     target_compile_definitions(device PRIVATE TT_UMD_BUILD_SIMULATION)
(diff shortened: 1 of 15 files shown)


tt-vjovanovic (approved):

broskoTT:
@mcraigheadTT let's continue chatting offline and on the scheduled syncs.
In any case, it is urgent and critical to re-enable the simulation path.

Issue #1251, opened 2025-09-02
Support ttsim


broskoTT:
Milestone to capture tasks for supporting ttsimulator

broskoTT:
Initial support is done, CI is up.

Issue #1253, opened 2025-09-02
Simulator lock should not be system-wide


mcraigheadTT:
The tt-umd simulator lock that was added to fix DPRINT introduced a very large performance regression running multiple simulations in parallel on the same host system.  After tracking it down, it appears that the RobustMutex used here is a shared-memory-based multi-process lock.

Such a lock should only be used when accessing system-wide resources, such as those associated with a real HW device, and should not be used to protect access to a simulator that exists only within the scope of a single application (i.e. it is not a global system-wide resource).

We probably need an assert to make sure that RobustMutex is never used in simulation, as even infrequent use of it could cause material performance leakage.  This can be especially hard to track down because you won't notice the perf regression until you run multiple simulations in parallel.

In terms of the actual fix, my recommendation would be to use a plain old pthread_mutex with a very simple RAII wrapper local to this one source file, and don't bother with using LockManager at all, as it just adds more complexity and inter-module dependencies to pull this in.

PR #1266, opened 2025-09-03
Add support for libttsim.so to fix ttsim breakage

mcraigheadTT:
### Issue
#1251 

### Description
Add logic to load libttsim.so and use entry points from it as an alternative to IPC. Fixes failures where all tests are currently failing on ttsim. Removing the IPC is also approximately a 2x speedup in some tests.

### List of the changes
- Add dlopen, dlsym, dlclose of simulator .so
- Route the various simulator messages to these entry points from .so
- Enable this code path when TT_METAL_SIMULATOR points to libttsim.so directly
- Locate soc_descriptor.yaml in the parent directory of libttsim.so

### Testing
local testing of Metal examples on ttsim

### API Changes
/

### DIFF
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -435,7 +435,11 @@ Cluster::Cluster(ClusterOptions options) {
 
     if (options.sdesc_path.empty() &&
         (options.chip_type == ChipType::SIMULATION || options.chip_type == ChipType::TTSIM)) {
-        options.sdesc_path = options.simulator_directory / "soc_descriptor.yaml";
+        if (options.simulator_directory.extension() == ".so") {
+            options.sdesc_path = options.simulator_directory.parent_path() / "soc_descriptor.yaml";
+        } else {
+            options.sdesc_path = options.simulator_directory / "soc_descriptor.yaml";
+        }
     }
 
     // Construct all the required chips from the cluster descriptor.
--- a/device/simulation/simulation_device.cpp
+++ b/device/simulation/simulation_device.cpp
@@ -6,6 +6,7 @@
 
 #include "umd/device/simulation/simulation_device.hpp"
 
+#include <dlfcn.h>
 #include <nng/nng.h>
 #include <uv.h>
 
@@ -17,6 +18,12 @@
 #include "assert.hpp"
 #include "simulation_device_generated.h"
 #include "umd/device/driver_atomics.hpp"
+
+#define DLSYM_FUNCTION(func_name)                                                    \
+    pfn_##func_name = (decltype(pfn_##func_name))dlsym(libttsim_handle, #func_name); \
+    if (!pfn_##func_name) {                                                          \
+        TT_THROW("Failed to find '%s' symbol: ", #func_name, dlerror());             \
+    }
 
 namespace tt::umd {
 
@@ -56,7 +63,10 @@ inline static void print_flatbuffer(const DeviceRequestResponse* buf) {
 }
 
 SimulationDeviceInit::SimulationDeviceInit(const std::filesystem::path& simulator_directory) :
-    simulator_directory(simulator_directory), soc_descriptor(simulator_directory / "soc_descriptor.yaml") {}
+    simulator_directory(simulator_directory),
+    soc_descriptor(
+        (simulator_directory.extension() == ".so") ? (simulator_directory.parent_path() / "soc_descriptor.yaml")
+                                                   : (simulator_directory / "soc_descriptor.yaml")) {}
 
 SimulationDevice::SimulationDevice(const SimulationDeviceInit& init) : Chip(init.get_soc_descriptor()) {
     log_info(tt::LogEmulationDriver, "Instantiating simulation device");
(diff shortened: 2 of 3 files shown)


device/simulation/simulation_device.cpp
```diff
@@ -65,75 +69,119 @@ SimulationDevice::SimulationDevice(const SimulationDeviceInit& init) : Chip(init
...
-    child_options.stdio = child_stdio;
-
-    uv_process_t child_p;
-    int rv = uv_spawn(loop, &child_p, &child_options);
-    if (rv) {
-        TT_THROW("Failed to spawn simulator process: ", uv_strerror(rv));
-    } else {
-        log_info(tt::LogEmulationDriver, "Simulator process spawned with PID: {}", child_p.pid);
+
+    if (simulator_path.extension() == ".so") {
```
broskoTT, line 77: Since we have now completely different code for the two paths, it probably makes sense to separate these into two different classes, which would derive from the same interface. The common interface could just be the Chip class. Or even, the existing TTSim class could be repurposed for this.
In any case, we can redesign this later without any functional changes, so I'm not gonna block this PR on this comment, feel free to close the comment without resolving this

device/simulation/simulation_device.cpp
```diff
@@ -163,35 +215,45 @@ void SimulationDevice::write_to_device(CoreCoord core, const void* src, uint64_t
...
+        size_t rd_rsp_sz = host.recv_from_device(&rd_resp);
 
-    auto rd_resp_buf = GetDeviceRequestResponse(rd_resp);
+        auto rd_resp_buf = GetDeviceRequestResponse(rd_resp);
 
-    // Debug level polling as Metal will constantly poll the device, spamming the logs
-    log_debug(tt::LogEmulationDriver, "Device reading vec");
-    print_flatbuffer(rd_resp_buf);
+        // Debug level polling as Metal will constantly poll the device, spamming the logs
+        log_debug(tt::LogEmulationDriver, "Device reading vec");
```
broskoTT, line 251: Do you maybe want to have log_debug in your path as well? This would allow you to turn them on on demand, although you can probably add logs on ttsim side as well.
Note that log_debugs are not even compiled by default in the Release build. We have two flags, one which enables compiling and one which controls logs at runtime.

device/simulation/simulation_device.cpp
```diff
@@ -65,75 +69,119 @@ SimulationDevice::SimulationDevice(const SimulationDeviceInit& init) : Chip(init
...
+        size_t buf_size = host.recv_from_device(&buf_ptr);
+        auto buf = GetDeviceRequestResponse(buf_ptr);
+        auto cmd = buf->command();
+        TT_ASSERT(cmd == DEVICE_COMMAND_EXIT, "Did not receive expected command from remote.");
+        nng_free(buf_ptr, buf_size);
+    }
 }
 
 void SimulationDevice::send_tensix_risc_reset(tt_xy_pair core, const TensixSoftResetOptions& soft_resets) {
     auto lock = lock_manager.acquire_mutex(MutexType::TT_SIMULATOR);
```
broskoTT, line 159: You can also consider completely skipping the locks in case of ttsim. I'll work on https://github.com/tenstorrent/tt-umd/issues/1253 anyway

device/simulation/simulation_device.cpp
broskoTT, line 77: Can you please update cluster.hpp, `struct ClusterOptions`, `simulator_directory` with a comment explaining that there are two different types of simulator which could be ran based on the contents of that option?

broskoTT (approved):
Thanks for the improvement!
I've manually tested that the versim simulation still works from tt-metal with these changes (hopefully we get a proper CI for that soon)

device/simulation/simulation_device.cpp
```diff
@@ -65,75 +69,119 @@ SimulationDevice::SimulationDevice(const SimulationDeviceInit& init) : Chip(init
     arch_name = init.get_arch_name();
     target_devices_in_cluster = {0};
 
-    // Start VCS simulator in a separate process
```
pjanevskiTT, line 68: I see that this comment is no longer true. I would however keep some comment which explains two possible scenarios here, one for running simulator with .so file and the other one that spawns separate process and does IPC

pjanevskiTT (approved):
looks good

device/cluster.cpp
```diff
@@ -448,7 +448,11 @@ Cluster::Cluster(ClusterOptions options) {
         }
         if (options.sdesc_path.empty() &&
             (options.chip_type == ChipType::SIMULATION || options.chip_type == ChipType::TTSIM)) {
-            options.sdesc_path = options.simulator_directory / "soc_descriptor.yaml";
+            if (options.simulator_directory.extension() == ".so") {
```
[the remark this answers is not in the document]
mcraigheadTT: I specifically want the env var to point to the .so though for a few reasons though.  This will work a lot better once we get rid of the need to copy the soc_descriptor.yaml and the concept of a "simulator directory" goes away altogether, you'll just point it straight at your libttsim.so in your ttsim build output.

I also don't want to require any particular filename for the .so, for example in the future I imagine there could be a librtlsim.so (or whatever, pick your favorite name), a libsocksim.so that talks to a remote simulator over a socket, etc.

mcraigheadTT:
Resolved merge conflicts, need to verify that I didn't screw anything up

device/simulation/simulation_device.cpp
```diff
@@ -65,75 +69,119 @@ SimulationDevice::SimulationDevice(const SimulationDeviceInit& init) : Chip(init
...
-        log_info(tt::LogEmulationDriver, "Simulator process spawned with PID: {}", child_p.pid);
+
+    if (simulator_path.extension() == ".so") {
+        // dlopen the simulator library
+        libttsim_handle = dlopen(simulator_path.string().c_str(), RTLD_LAZY);
+        if (!libttsim_handle) {
+            TT_THROW("Failed to dlopen simulator library: ", dlerror());
+        }
+
+        // dlsym the entry points
```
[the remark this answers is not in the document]
mcraigheadTT: I actually collapsed these two comments into 1 in my latest PR based on the way the code flowed after moving the #define

device/simulation/simulation_device.cpp
```diff
@@ -56,7 +63,10 @@ inline static void print_flatbuffer(const DeviceRequestResponse* buf) {
 }
 
 SimulationDeviceInit::SimulationDeviceInit(const std::filesystem::path& simulator_directory) :
-    simulator_directory(simulator_directory), soc_descriptor(simulator_directory / "soc_descriptor.yaml") {}
+    simulator_directory(simulator_directory),
+    soc_descriptor(
+        (simulator_directory.extension() == ".so") ? (simulator_directory.parent_path() / "soc_descriptor.yaml")
```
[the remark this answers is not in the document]
mcraigheadTT: Agree, I think what you did there may not compile (?) but I can figure it out

device/cluster.cpp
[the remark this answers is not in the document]
broskoTT: It's fine to leave it for next PR if you have urgency on this one..

PR #1267, opened 2025-09-04
Add process robust mutex

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/1253

### Description
Introduction of lock on the simulator path decreased the perf significantly.
The lock used was system wide, so it had to modify the filesystem.
I've now created an option to use more lightweight locks which are not backed by a filesystem, and are only to be used within a single process.

### List of the changes
- Move RobustMutex to RobustSystemMutex
- Created a RobustProcessMutex which is more lightweight
- Turn the old RobustMutex into an interface, and change the usage to be used as a unique_ptr rather than on the stack.

### Testing
I've manually tested that the simulator path still works, so the locks are working. I've manually removed TT_SIMULATOR from /dev/shm to test that this files weren't being created anymore

### API Changes


mcraigheadTT:
I have one other run that's tying up my machine right now, but testing this is next in line for me this morning

mcraigheadTT:
Testing now, first running my -j10 version of unit_tests_llk then will run -j10 version of ttnn suite after that.

device/api/umd/device/utils/robust_process_mutex.hpp
```diff
@@ -0,0 +1,40 @@
...
+    // Copy constructor and assignment are not allowed.
+    RobustProcessMutex(const RobustProcessMutex&) = delete;
+    RobustProcessMutex& operator=(const RobustProcessMutex&) = delete;
+
+    // Locks and unlocks the mutex.
+    void unlock() override;
+    void lock() override;
+
+private:
+    pthread_mutex_t mutex_;
```
tt-vjovanovic, line 36: Why do you need new mutex? Can't you use `std::mutex` for process level locking?

tt-vjovanovic (commented):
Even though this might involve some interesting scenarios, question is, do we actually need lock manager that is aware of process level mutex?
If I understand correctly, we decided to replace simulator mutex to be process level instead of host level mutex. If that is the case, that mutex is scoped to a single class and I don't think we need this change for that.

device/api/umd/device/utils/robust_process_mutex.hpp
mcraigheadTT: I would be fine with that or even direct usage of pthread_mutex_t directly in the source file that needs it

broskoTT: Synced offline briefly, I'll change the code to using std::mutex, no need for complications that this PR introduces

broskoTT:
Superseded by https://github.com/tenstorrent/tt-umd/pull/1279

PR #1276, opened 2025-09-05
Remove TTSimDevice

broskoTT:
### Issue
In large part a revert of https://github.com/tenstorrent/tt-umd/pull/1250

### Description
This was added to unblock both paths of simulation, but in the meantime another solution popped up which is now used by metal.
The metal change acompanying the original one never got merged, and is closed now: https://github.com/tenstorrent/tt-metal/pull/27763

This is also a part of PR https://github.com/tenstorrent/tt-umd/pull/1274 , but that one might need more time, and we all agree on these changes.

### List of the changes
- Remove all added ttsim files
- Remove ChipType::TTSIM

### Testing
Code builds

### API Changes
This PR removes added API which was a hack.


mcraigheadTT (commented):
Looks good to me

pjanevskiTT (approved):

Issue #1286, opened 2025-09-09
Move some types to tt namespace


broskoTT:
I've put everything (except tt::ARCH) to tt::umd namepsace as part of #103 
But some of it should be actually in tt namespace, as it can be used outside of interacting with the UMD.
Analyze all the types and move them

PR #1306, opened 2025-09-15
Another rename round

### LINKED ISSUE #1285: Remove _t suffix and rename to camel case
Our old types have old_naming_scheme_t, but new types are named with NewNamingScheme.
Make this consistent.

I've discussed this with Oleg on https://github.com/tenstorrent/tt-metal/pull/28063

broskoTT:
### Issue
Mostly related to comments in https://github.com/tenstorrent/tt-metal/pull/28063
Related to https://github.com/tenstorrent/tt-umd/issues/103
Related to general coding style https://github.com/tenstorrent/tt-umd/issues/356
Fixes #1285 and #1286

### Description
After offline chat with Oleg, we agreed on which structs should be in tt namespace as opposed to tt::umd namespace.
Also, this PR makes struct naming uniform as CamelCase.
There were also some forward declarations in metal which were now removed, so we could also put some of the structs in the desired namespaces.

### List of the changes
- Put HarvestingMasks in cluster descriptor types, as it was consumed by it anyway. Remove harvesting files
- Remove some forward declarations of device_ and driver_ structs
- ARCH, cluster descriptor types, core coordinates and xy_pair, are moved to tt namespace, outside of tt::umd namespace
- Everything else is supposed to have to be used with umd so it remains in umd namespace.
- BoardType and CoreType moved to namespace since the forward declaration in tt metal was removed
- Renamed chip_id_t, ethernet_channel_t, eth_coord_t to CamelCase, and put them to tt namespace
- xy_pair remains named like that since XYPair seems uglier. So we will have tt::xy_pair and tt_xy_pair
- renamed tt_device_params, barrier_address_params, device_dram_address_params, device_l1_address_params, driver_host_address_params, driver_noc_params, driver_eth_interface_params, hugepage_mapping to CamelCase
- Added using namespace tt; to many tests now that some structs live there
- Added a bunch of using declarations in header files to help with the transition of both umd and metal code.
- Actually CoreCoord won't be moved to tt namespace for now, not to clash with metal's CoreCoord

### Testing
Code builds on tt_metal and tt_exalens

### API Changes
This PR has API changes, introduces new names while keeping old ones.


nbuncicTT (approved):

device/api/umd/device/types/cluster_types.hpp
```diff
@@ -226,9 +225,13 @@ struct hugepage_mapping {
 }  // namespace tt::umd
 
 // TODO: To be removed once clients switch to namespace usage.
-namespace tt::umd {
-using device_params = ::tt_device_params;
-}
-
-using tt::umd::barrier_address_params;
 using tt::umd::tt_version;
+using barrier_address_params = tt::umd::BarrierAddressParams;
```
pjanevskiTT, line 229: I would just urge the removal of these using statements as soon as possible

device/api/umd/device/types/cluster_types.hpp
```diff
@@ -123,21 +122,21 @@ inline std::ostream& operator<<(std::ostream& os, const DevicePowerState power_s
     return os;
 }
 
-struct barrier_address_params {
+struct BarrierAddressParams {
```
pjanevskiTT, line 125: I like renaming these structs to CamelCase

pjanevskiTT (approved):

device/api/umd/device/types/cluster_types.hpp
broskoTT: I understand that you want this cleaned up as soon as possible.
However this will introduce a big change in tt_metal as well which will need time to get merged.

Another issue, is that we're changing our API significantly, and now that we have some form of versioning, we should use it to signal that.

I'll probably bump the version right after PRs which:
- Remove old header files
- Remove all of these using statements
- Remote old tensix risc reset api

All of that said, don't worry, i won't forget it

device/api/umd/device/types/cluster_types.hpp
```diff
@@ -123,21 +122,21 @@ inline std::ostream& operator<<(std::ostream& os, const DevicePowerState power_s
...
     std::uint32_t tensix_l1_barrier_base = 0;
     std::uint32_t eth_l1_barrier_base = 0;
     std::uint32_t dram_barrier_base = 0;
 };
 
-struct device_dram_address_params {
+struct DeviceDramAddressParams {
```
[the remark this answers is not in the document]
broskoTT: Yeah actually I'll leave it like this:
- I know we have PCIDevice, but I also see PciDevice
- Same should apply for NOC vs Noc, but I see we already use Noc for some stuff in lite_fabric.hpp

Maybe just PCIDevice -> PciDevice makes more sense? In any case it is probably of such low impact that we won't ever end up changing it

PR #1309, opened 2025-09-16
Avoid creating socket when using libttsim.so path

mcraigheadTT:
### Issue
https://github.com/tenstorrent/ttsim/issues/36

### Description
Socket was being created even when it was unused.  This also led to creation of a large number of unused pthreads inside the nng library and extra confusing debug output at startup time, and likely slowed down startup time somewhat as well.

### List of the changes
Split most of the SimulationHost init out of the constructor so that it can be called conditionally.

### Testing
Ran programming examples locally on simulator

### API Changes
/

### DIFF
--- a/device/api/umd/device/simulation/simulation_host.hpp
+++ b/device/api/umd/device/simulation/simulation_host.hpp
@@ -19,6 +19,7 @@ public:
     SimulationHost();
     ~SimulationHost();
 
+    void init();
     void start_host();
     void send_to_device(uint8_t *buf, size_t buf_size);
     size_t recv_from_device(void **data_ptr);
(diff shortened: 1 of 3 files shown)


device/api/umd/device/simulation/simulation_host.hpp
```diff
@@ -19,6 +19,7 @@ class SimulationHost {
     SimulationHost();
```
broskoTT, line 19: Copying my comment from: https://github.com/tenstorrent/tt-umd/pull/1274#discussion_r2339725526

Don't make further changes to the versim simulator path, since versim simulator and ttsim are two essentially different codepaths. That is true as of right now. I don't want to discuss future outside of already prepared PRs that would come after this. At the moment, it only makes sense that these two codepaths are properly separated, and this change is going against it.

Issue #1320, opened 2025-09-16
Blackhole simulation failing for full grid


broskoTT:
We should enable the eth harvesting check, and also add harvesting to mock cluster.

Unfortunately, this also means that default soc descriptor can't be instantiated without proper harvesting mask. So this requires some more changes, and also removing SimulationDeviceInit.

PR #1321, opened 2025-09-16
Fix mock eth harvesting BH

broskoTT:
### Issue
Fixes https://github.com/tenstorrent/tt-umd/issues/1320

### Description
Add default bh eth harvesting mask to mock cluster descriptor.
This is required since BH always has 2 eth cores harvested.
Following changes will further cleanup SimulationDeviceInit, I've left it out of this change so that the fix goes in faster.

### List of the changes
- Add eth harvesting to mock cluster descriptor
- Add p150 cluster descriptor yaml
- Pass soc descriptor to SimulationDevice
- I've had to create another constructor, which circumvents the SimulationDeviceInit. This is temporary.

### Testing
Tested manually that after this change, ttsim run doesn't access (33,25) coords anymore.

### API Changes

### DIFF
--- a/device/cluster_descriptor.cpp
+++ b/device/cluster_descriptor.cpp
@@ -473,7 +473,7 @@ std::unique_ptr<ClusterDescriptor> ClusterDescriptor::create_mock_cluster(
         case tt::ARCH::QUASAR:  // TODO (#450): Add Quasar configuration
         case tt::ARCH::BLACKHOLE:
             board_type = BoardType::UNKNOWN;
-            harvesting_masks.pcie_harvesting_mask = 0x0;
+            harvesting_masks.eth_harvesting_mask = 0x120;
             break;
         default:
             board_type = BoardType::UNKNOWN;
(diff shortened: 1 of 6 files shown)


device/cluster_descriptor.cpp
```diff
@@ -473,7 +473,7 @@ std::unique_ptr<ClusterDescriptor> ClusterDescriptor::create_mock_cluster(
         case tt::ARCH::QUASAR:  // TODO (#450): Add Quasar configuration
         case tt::ARCH::BLACKHOLE:
             board_type = BoardType::UNKNOWN;
-            harvesting_masks.pcie_harvesting_mask = 0x0;
+            harvesting_masks.eth_harvesting_mask = 0x120;
```
pjanevskiTT, line 476: why 0x120? If it's random, that might not be the realistic number we are seeing on silicon. If it's not random, add a comment on why we are choosing this number

pjanevskiTT: oh I guess it's 288 from cluster descriptor yaml, add a comment please

PR #1322, opened 2025-09-16
Offer alternative to SimulationDeviceInit

broskoTT:
### Issue
Related to https://github.com/tenstorrent/tt-umd/issues/1320

### Description
Offer an alternative way to get arch and grid size from soc descriptor. Also as a convenience, a method which returns exact soc_desc yaml.
It is important to add this alternative, since default soc descriptor, and this simulationDeviceInit won't be able to be created without specifying eth harvesting mask. This is added in the following PR

### List of the changes
- Add convenience functions to get arch and grid_size from soc_descriptor
- Use them in cluster.cpp

### Testing
Corresponding metal change: https://github.com/tenstorrent/tt-metal/pull/28663
Tested with manual run of ttsim

### API Changes
This PR has API changes, but they aren't breaking, as both new and old APIs will live.

### DIFF
--- a/device/api/umd/device/soc_descriptor.hpp
+++ b/device/api/umd/device/soc_descriptor.hpp
@@ -77,6 +77,11 @@ public:
 
     SocDescriptor(const tt::ARCH arch, const ChipInfo chip_info = {});
 
+    // Helpers for extracting info from soc descriptor file.
+    static std::string get_soc_descriptor_path_from_simulator_path(const std::filesystem::path &simulator_path);
+    static tt::ARCH get_arch_from_soc_descriptor_path(const std::string &soc_descriptor_path);
+    static tt_xy_pair get_grid_size_from_soc_descriptor_path(const std::string &soc_descriptor_path);
+
     // CoreCoord conversions.
     CoreCoord translate_coord_to(const CoreCoord core_coord, const CoordSystem coord_system) const;
     std::unordered_set<CoreCoord> translate_coords_to(
--- a/device/soc_descriptor.cpp
+++ b/device/soc_descriptor.cpp
@@ -353,6 +353,22 @@ std::vector<tt_xy_pair> SocDescriptor::convert_to_tt_xy_pair(const std::vector<s
     return core_pairs;
 }
 
+std::string SocDescriptor::get_soc_descriptor_path_from_simulator_path(const std::filesystem::path &simulator_path) {
+    return (simulator_path.extension() == ".so") ? (simulator_path.parent_path() / "soc_descriptor.yaml")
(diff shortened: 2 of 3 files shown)


device/api/umd/device/soc_descriptor.hpp
```diff
@@ -77,6 +77,11 @@ class SocDescriptor {
 
     SocDescriptor(const tt::ARCH arch, const ChipInfo chip_info = {});
 
+    // Helpers for extracting info from soc descriptor file.
+    static std::string get_soc_descriptor_path_from_simulator_path(const std::filesystem::path &simulator_path);
```
pjanevskiTT, line 81: I don't think this is something that should live in soc descriptor class. I think this should either be in some SimulationDevice class or anything similar. Or some simulation_utils file

device/soc_descriptor.cpp
```diff
@@ -353,6 +353,22 @@ std::vector<tt_xy_pair> SocDescriptor::convert_to_tt_xy_pair(const std::vector<s
...
 }
 
+std::string SocDescriptor::get_soc_descriptor_path_from_simulator_path(const std::filesystem::path &simulator_path) {
+    return (simulator_path.extension() == ".so") ? (simulator_path.parent_path() / "soc_descriptor.yaml")
+                                                 : (simulator_path / "soc_descriptor.yaml");
+}
+
+tt::ARCH SocDescriptor::get_arch_from_soc_descriptor_path(const std::string &soc_descriptor_path) {
+    YAML::Node device_descriptor_yaml = YAML::LoadFile(soc_descriptor_path);
+    return tt::arch_from_str(device_descriptor_yaml["arch_name"].as<std::string>());
```
pjanevskiTT, line 363: a small thing to think about here is that if we ever change yamls to use anything else that is different from "arch_name" string, there would be a need to change that in multiple places. I would solve this by either
- putting these strings into constants
- create SocDescriptor object from yaml and read arch field. Then we don't replicate arch_name string here

device/soc_descriptor.cpp
```diff
@@ -353,6 +353,22 @@ std::vector<tt_xy_pair> SocDescriptor::convert_to_tt_xy_pair(const std::vector<s
...
+}
+
+tt_xy_pair SocDescriptor::get_grid_size_from_soc_descriptor_path(const std::string &soc_descriptor_path) {
+    YAML::Node device_descriptor_yaml = YAML::LoadFile(soc_descriptor_path);
+    return tt_xy_pair(
+        device_descriptor_yaml["grid"]["x_size"].as<int>(), device_descriptor_yaml["grid"]["y_size"].as<int>());
```
pjanevskiTT, line 369: same as above

PR #1328, opened 2025-09-17
Make write method of Writer const

pjanevskiTT:
### Issue

Make method of Writer class const. This is needed for first step in refactoring metal code to new TLB scheme. The first step is to remove fast_pcie_write_callable. Related changes https://github.com/tenstorrent/tt-metal/compare/main...pjanevski/bar0_2

### Description

Make write method of Writer const

### List of the changes

- Make write method of Writer const

### Testing
CI

### API Changes
/

### DIFF
--- a/device/api/umd/device/tt_io.hpp
+++ b/device/api/umd/device/tt_io.hpp
@@ -31,7 +31,7 @@ public:
      * @param value
      */
     template <class T>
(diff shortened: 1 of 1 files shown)


nbuncicTT (approved):
Nice job

PR #1335, opened 2025-09-18
Add syntax error in cluster.cpp

dzivanovicTT:
### Issue
/

### Description
Create failing PR for GitLab CI

### List of the changes
Add syntax error in cluster.cpp

### Testing
CI

### API Changes
/

### DIFF
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -76,7 +76,6 @@ static constexpr uint32_t REMOTE_CMD_NOC_BIT = 9;
 #include "umd/device/tt_xy_pair.h"
 #include "umd/device/types/tensix_soft_reset_options.hpp"
 
-namespace tt::umd {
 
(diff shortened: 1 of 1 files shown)


dzivanovicTT:
> @dzivanovicTT can you just open PRs like this in draft, because then code owners don't get notified and you can still run CI on these?

Thanks, @pjanevskiTT  — @broskoTT  had a similar comment. Maybe I should have announced this setup earlier. The reason I opened two PRs was to have a test environment for trying out changes to the GitLab CI setup before rolling them out to production on the UMD and Metal repos.
As you may have noticed, these two PRs are on separate branches rather than on main. That way I can also test merged PRs outside of the main branch.
The PRs are opened or merged, not drafts, because GitLab CI only tests such PRs. I didn’t want my test environment to diverge too much from a real scenario.

PR #1347, opened 2025-09-19
On sim, iterate over cores and reset each one individually

mcraigheadTT:
### Issue
https://github.com/tenstorrent/ttsim/issues/49

### Description
Replace implementation of SimulationDevice::send_tensix_risc_reset with no X/Y specified with loop over cores, same as is done on silicon.

### List of the changes
Loop over Tensix cores and reset each one.  Code was taken from base class and is identical to base class implementation, so this method override could potentially be removed in a subsequent change.

### Testing
Ran programming examples locally on simulator

### API Changes
/


mcraigheadTT:
FYI, I'm open to restricting the change to `if (libttsim_handle)` if there are concerns about this on RTL, though this is extra code + would inhibit the possible downstream cleanup of removing the sim-specific override of this virtual method.

Either way, on ttsim, this change is required for subsequent ttsim reset changes that are coming down the pipe.

PR #1350, opened 2025-09-19
Uvelimirovic/add jtag translated coords test

uvelimirovicTT:
### Description
It was speculated that JTAG works with translated coordinates since they are not an integrated part of JTAG protocol but separated. This PR is a proof of that.

### List of the changes
added a test in test_jtag.cpp

### Testing
JtagTranslatedCoordsTest


pjanevskiTT (approved):
nice work

Issue #1096, opened 2025-07-16
Complete TLB perf tests


pjanevskiTT:
Complete perf tests for TLBs

- Sweep sizes
- TLB reads/writes to multiple core types

pjanevskiTT:
closed by #1370

PR #1370, opened 2025-09-24
Fully cleanup microbenchmark TLB tests

pjanevskiTT:
### Issue

#1096 

### Description

Complete TLB perf tests. 

### List of the changes

- Set NUM_ITERATIONS to 10 in order to make speed of tests reasonable
- Sweep sizes for all tests
- Report results in bytes instead of MB

### Testing

Local run

### API Changes
/


pjanevskiTT:
Merging it even if tt-sim CI step fails, since it seems as git related issue...the file changed is not included in metal build anyway

PR #1395, opened 2025-09-26
Package up the missing API headers

afuller-TT:
### Issue
N/A

### Description
Files in api/ are missing from the package.

### List of the changes
Package up the missing API headers.

### Testing
Consumed the TT-Metalium package.

### API Changes
None


broskoTT:
I actually wanted to solve this through https://github.com/tenstorrent/tt-umd/pull/1362. I'll revisit what you added, I don't think all of those headers are required. This list needs a cleanup anyhow.
Thanks for fixing

PR #1404, opened 2025-10-01
Add hardware hang check for JTAG.

uvelimirovicTT:
### Description
Jtag Device has internal check for hardware hang inside of its library for each write and read operation when accessing axi bus (which happens every time with current implementation).

This PR enables user to check hardware hang explicitly if in the future JTAG I/O implementation changes or for whatever other reason.

### List of the changes
JtagDevice

### DIFF
--- a/device/jtag/jtag_device.cpp
+++ b/device/jtag/jtag_device.cpp
@@ -319,8 +319,9 @@ bool JtagDevice::is_hardware_hung(uint8_t chip_id) {
     uint32_t timeout = 10;
     std::optional<uint32_t> status;
     do {
+        // Details about status format can be found inside JTAG library.
         status = read_tdr(chip_id, "arc", 0x4);
-        status = ((status.value() >> 16) & 0xf);
+        status = status.value() >> 16;
         timeout--;
     } while (((status.value() & 0x1) == 0) && (timeout > 0));
     return timeout == 0;
--- a/device/jtag/jtag_device.cpp
+++ b/device/jtag/jtag_device.cpp
@@ -155,7 +155,12 @@ int JtagDevice::open_jlink_wrapper(uint8_t chip_id) { return jtag->open_jlink_wr
 
 std::optional<uint32_t> JtagDevice::read_tdr(uint8_t chip_id, const char* client, uint32_t reg_offset) {
     select_device(chip_id);
-    return jtag->read_tdr(client, reg_offset);
+    try {
+        uint32_t result = jtag->read_tdr(client, reg_offset);
+        return result;
+    } catch (...) {
+        return std::nullopt;
+    }
 }
 
 std::optional<uint32_t> JtagDevice::readmon_tdr(uint8_t chip_id, const char* client, uint32_t id, uint32_t reg_offset) {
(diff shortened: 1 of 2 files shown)


device/jtag/jtag_device.cpp
```diff
@@ -155,7 +155,12 @@ int JtagDevice::open_jlink_wrapper(uint8_t chip_id) { return jtag->open_jlink_wr
 
 std::optional<uint32_t> JtagDevice::read_tdr(uint8_t chip_id, const char* client, uint32_t reg_offset) {
     select_device(chip_id);
-    return jtag->read_tdr(client, reg_offset);
+    try {
+        uint32_t result = jtag->read_tdr(client, reg_offset);
+        return result;
+    } catch (...) {
```
aleksamarkovicTT, line 161: TODO, no action needed:
Make catch narrower.

device/jtag/jtag_device.cpp
```diff
@@ -310,3 +312,16 @@ int JtagDevice::get_device_id(uint8_t chip_id) const {
     }
     return jlink_devices[chip_id];
 }
+
+bool JtagDevice::is_hardware_hung(uint8_t chip_id) {
```
nbuncicTT, line 316: Just trying to understand: As you're TODO comment states, now if there is a failure from `read_tdr`, it doesn't return `std::nullopt`, but terminates the program?

And this hang detection is only timeout based for now?

nbuncicTT (approved):
Nice job, thanks!

device/jtag/jtag_device.cpp
```diff
@@ -310,3 +312,16 @@ int JtagDevice::get_device_id(uint8_t chip_id) const {
...
+    // Timeout value 10 is chosen from JTAG library.
+    // Checkout the library implementation for more details.
+    uint32_t timeout = 10;
+    std::optional<uint32_t> status;
+    do {
+        status = read_tdr(chip_id, "arc", 0x4);
+        status = ((status.value() >> 16) & 0xf);
+        timeout--;
+    } while (((status.value() & 0x1) == 0) && (timeout > 0));
```
pjanevskiTT, line 325: if we only care about one bit in the status variable, why are we doing `& 0xf` two lines above? Why not do `& 0x1` when reading status?

device/jtag/jtag_device.cpp
```diff
@@ -310,3 +312,16 @@ int JtagDevice::get_device_id(uint8_t chip_id) const {
...
+
+bool JtagDevice::is_hardware_hung(uint8_t chip_id) {
+    // Timeout value 10 is chosen from JTAG library.
+    // Checkout the library implementation for more details.
+    uint32_t timeout = 10;
+    std::optional<uint32_t> status;
+    do {
+        status = read_tdr(chip_id, "arc", 0x4);
+        status = ((status.value() >> 16) & 0xf);
+        timeout--;
```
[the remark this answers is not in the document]
pjanevskiTT: you already have a comment that this code is taken from the library so keeping number 10 there is fine for me

device/jtag/jtag_device.cpp
uvelimirovicTT: Yes, exactly.

PR #1408, opened 2025-10-02
Update sysmem test case

joelsmithTT:
### Issue
Inspired by https://github.com/tenstorrent/tt-metal/issues/29455

### Description
Misconfiguring a VM with vIOMMU can result in failure that is not easily detectable by KMD or UMD. A recommendation is to use UMD unit tests to verify the functionality of a newly-provisioned VM. Although the existing sysmem tests catch the issue, the existing sysmem tests:
* are architecture-specific
* provide no actionable error message for the failure mode in question

### List of the changes
* Remove BH and WH-specific sysmem test cases
* Introduce a new sysmem test case that covers the old cases, but in an architecture-agnostic manner
* Fix up an unrelated code path that caused confusing error messages during testing (added as a separate commit)

### Testing
* Manual: WH, BM
* Manual: BH, BM
* Manual, BH, VM, vIOMMU, misconfigured host
* Manual, WH, VM, vIOMMU, misconfigured host
* Manual, BH, VM, vIOMMU, correctly configured host
* Manual, WH, VM, vIOMMU, correctly configured host

### API Changes


device/tt_device/blackhole_tt_device.cpp
```diff
@@ -231,7 +231,16 @@ uint64_t BlackholeTTDevice::get_arc_noc_base_address() const { return blackhole:
 bool BlackholeTTDevice::wait_arc_post_reset(const uint32_t timeout_ms) { return true; }
 
 bool BlackholeTTDevice::is_hardware_hung() {
-    throw std::runtime_error("Hardware hang detection is not supported on Blackhole.");
+    // throw std::runtime_error("Hardware hang detection is not supported on Blackhole.");
```
nbuncicTT, line 234: In the throw it's written `Hardware hang detection is not supported on Blackhole.`, so the invocation of this function will result in a throw, regardless of the reading of 0xffffffff.

So the comment below I guess kind of implicates that this is used wrongfully which I think is not right. On the other side, if you use this function in an architecturally agnostic manner (check for either WH or BH), then it may behave in the way the comment was describing it, but with a precise answer why.

My question is, is it possible to detect BH hangs? If yes, can we do it (in a separate issue/PR). If not, why not?

device/tt_device/tt_device.cpp
```diff
@@ -293,6 +293,7 @@ void TTDevice::read_block(uint64_t byte_addr, uint64_t num_bytes, uint8_t *buffe
     }
 
     if (num_bytes >= sizeof(std::uint32_t)) {
+        // TODO: looks like there is potential for undefined behavior here.
```
nbuncicTT, line 296: Can you give more context why is this the case?

tests/api/test_cluster.cpp
```diff
@@ -941,3 +942,126 @@ INSTANTIATE_TEST_SUITE_P(
...
+ * 5. Verifies that the offsets have been zeroed from host's perspective.
+ */
+TEST(TestCluster, SysmemReadWrite) {
+    {
+        Cluster cluster;
+        if (cluster.get_target_device_ids().empty()) {
+            GTEST_SKIP() << "No chips present on the system. Skipping test.";
+        }
+    }
+    const size_t ONE_GIG = 1ULL << 30;
```
nbuncicTT, line 966: nit: all literal constants can be/should be made `constexpr`

tests/api/test_cluster.cpp
```diff
@@ -941,3 +942,126 @@ INSTANTIATE_TEST_SUITE_P(
...
+    const uint64_t ALIGNMENT = sizeof(uint32_t);
+    const bool is_vm = test_utils::is_virtual_machine();
+    const bool has_iommu = test_utils::is_iommu_available();
+    const uint32_t channels = is_vm ? 1 : has_iommu ? 3 : 2;
```
nbuncicTT, line 970: Can we have a comment why is this value for channels deduced in this way? 

Why does is it correlate to the `host_dma_address` as the channel parameter?

Why does it correlate to `device_offset` as a multiplicand of `ONE_GIG`?

I'm asking as I'd like to understand how do these mappings correlate with the VM and IOMMU being enabled/disabled.

tests/api/test_cluster.cpp
```diff
@@ -941,3 +942,126 @@ INSTANTIATE_TEST_SUITE_P(
     }
 
 );
+
```
nbuncicTT, line 945: Could you provide some logs of the test cases for the misconfigured host? As we don't mock these values and would only see them if we purposefully misconfigure the host in a test case, which I think we won't do in the CI intentionally.

nbuncicTT (approved):
Nice job of simplifying the tests and making them more clear.

tests/api/test_cluster.cpp
```diff
@@ -941,3 +942,126 @@ INSTANTIATE_TEST_SUITE_P(
...
+    const auto base_address = cluster.get_pcie_base_addr_from_device(mmio_chip_id);
+
+    auto random_address_between = [&](uint64_t lo, uint64_t hi) -> uint64_t {
+        static std::random_device rd;
+        static std::mt19937_64 gen(rd());
+        std::uniform_int_distribution<uint64_t> dis(lo, hi);
+        return dis(gen);
+    };
+
+    cluster.start_device(device_params{});
```
pjanevskiTT, line 986: nit: we don't usually call start_device in tests. Random binaries can be loaded to tensix cores so there might be some unexpected behavior if we don't know what is loaded

tests/api/test_cluster.cpp
```diff
@@ -941,3 +942,126 @@ INSTANTIATE_TEST_SUITE_P(
...
+        }
+
+        // Write test - zero out the sysmem at the various offsets.
+        for (uint64_t test_offset : test_offsets) {
+            uint64_t aligned_offset = (test_offset / ALIGNMENT) * ALIGNMENT;
+            uint64_t device_offset = aligned_offset + channel * ONE_GIG;
+            uint64_t noc_addr = base_address + device_offset;
+            uint32_t value = 0;
+            cluster.write_to_device(&value, sizeof(uint32_t), mmio_chip_id, pcie_core, noc_addr);
+            cluster.read_from_device(&value, mmio_chip_id, pcie_core, noc_addr, sizeof(uint32_t));
```
pjanevskiTT, line 1054: I don't see what is the point of doing this read?

pjanevskiTT (approved):

Issue #1413, opened 2025-10-06
Move verifying ETH FW version to TopologyDiscovery


aleksamarkovicTT:
Right now, this check happens late, in construct_cluster() which happens after a ClusterDescriptor was created by TopologyDiscovery.

To better accommodate Exalens's ask for making create_cluster_descriptor() resilient, we need to make this ETH FW check happen earlier in the code, before discovery of remote chips.
When there is a problem with ETH firmware, we should be able to allow TopologyDiscovery to create a cluster only with local chips.

PR #1421, opened 2025-10-07
Move ETH FW version check to TopologyDiscovery

aleksamarkovicTT:
### Issue
#1413

### Description
Right now, this check happens late, in `construct_cluster()` which happens after a `ClusterDescriptor` was created by TopologyDiscovery.

To better accommodate Exalens's ask for making `create_cluster_descriptor()` resilient, we need to make this ETH FW check happen earlier in the code, before discovery of remote chips.
When there is a problem with ETH firmware, we should be able to allow `TopologyDiscovery` to create a cluster only with local chips.

### List of the changes
- Made `tt_version` constructors constexpr
- Added cluster `eth_fw_version` to ClusterDescriptor
- Changed poorly named `SW_VERSION` constant to `ERISC_FW_SUPPORTED_VERSION_MIN` in `cluster_types.hpp` along with other relevant ETH FW versions
- Removed `verify_eth_fw()` and `verify_sw_fw_versions()` from `Cluster` and moved them to relevant locations in TopologyDiscovery

### Testing
CI

### API Changes

### DIFF
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -56,6 +56,7 @@
 #include "umd/device/topology/topology_utils.hpp"
 #include "umd/device/types/arch.hpp"
 #include "umd/device/types/blackhole_eth.hpp"
+#include "umd/device/types/cluster_types.hpp"
 #include "umd/device/types/core_coordinates.hpp"
 #include "umd/device/types/tlb.hpp"
 #include "umd/device/utils/common.hpp"
@@ -236,6 +237,8 @@ void Cluster::construct_cluster(const uint32_t& num_host_mem_ch_per_mmio_device,
     // TODO: work on removing this member altogether. Currently assumes all have the same arch.
     arch_name = chips_.empty() ? tt::ARCH::Invalid : chips_.begin()->second->get_soc_descriptor().arch;
 
+    tt_version fw_first_eth_core = cluster_desc->eth_fw_version;
+
     if (chip_type == ChipType::SILICON) {
         std::vector<int> pci_ids;
         auto mmio_id_map = cluster_desc->get_chips_with_mmio();
@@ -251,9 +254,21 @@ void Cluster::construct_cluster(const uint32_t& num_host_mem_ch_per_mmio_device,
             remote_chip_ids_);
         verify_fw_bundle_version();
         log_device_summary();
+
         if (arch_name == tt::ARCH::WORMHOLE_B0) {
-            verify_eth_fw();
+            // Min ERISC FW version required to support ethernet broadcast is 6.5.0.
+            use_ethernet_broadcast &= fw_first_eth_core >= ERISC_FW_ETH_BROADCAST_SUPPORTED_MIN;
+            // Virtual coordinates can be used for broadcast headers if ERISC FW >= 6.8.0 and NOC translation is enabled
+            // Temporarily enable this feature for 6.7.241 as well for testing.
+            use_translated_coords_for_eth_broadcast = true;
+            for (const auto& chip : all_chip_ids_) {
+                use_translated_coords_for_eth_broadcast &=
+                    (fw_first_eth_core >= ERISC_FW_ETH_BROADCAST_VIRTUAL_COORDS_MIN ||
+                     fw_first_eth_core == tt_version(6, 7, 241)) &&
+                    get_soc_descriptor(chip).noc_translation_enabled;
+            }
         }
+
         if (cluster_desc->get_io_device_type() == IODeviceType::PCIe) {
             verify_sysmem_initialized();
         }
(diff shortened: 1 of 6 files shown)


device/cluster.cpp
```diff
@@ -251,9 +254,21 @@ void Cluster::construct_cluster(const uint32_t& num_host_mem_ch_per_mmio_device,
             remote_chip_ids_);
         verify_fw_bundle_version();
         log_device_summary();
+
         if (arch_name == tt::ARCH::WORMHOLE_B0) {
-            verify_eth_fw();
+            // Min ERISC FW version required to support ethernet broadcast is 6.5.0.
```
pjanevskiTT, line 259: I would probably move this code to separate function. Maybe name it something like `check_eth_fw_constraints` and there you can set all these variables needed for further code flow

device/topology/topology_discovery.cpp
```diff
@@ -337,4 +343,68 @@ void TopologyDiscovery::patch_eth_connections() {}
 
 void TopologyDiscovery::initialize_remote_communication(Chip* chip) {}
 
+void TopologyDiscovery::verify_eth_version_local(int device_id, Chip* chip) {
+    // ETH FW version check incomplete for BH.
+    if (chip->get_soc_descriptor().arch != tt::ARCH::WORMHOLE_B0) {
```
pjanevskiTT, line 348: we have TopologyDiscovery wormhole and blackhole specific classes. If we want different behavior of some functions these should be overrided in those classes, don't use if based on arch

device/topology/topology_discovery.cpp
```diff
@@ -337,4 +343,68 @@ void TopologyDiscovery::patch_eth_connections() {}
...
+        // Check ETH FW version
+        uint32_t eth_fw_version_read;
+        chip->read_from_device(
+            eth_core, &eth_fw_version_read, chip->l1_address_params.fw_version_addr, sizeof(uint32_t));
+
+        tt_version eth_fw_version(eth_fw_version_read);
+
+        if (!first_eth_fw_version.has_value()) {
+            log_info(LogUMD, "Established cluster ETH FW version: {}.", eth_fw_version.str());
+            log_info(LogUMD, "UMD supported minimum ETH FW version: {}", ERISC_FW_SUPPORTED_VERSION_MIN.str());
```
pjanevskiTT, line 364: I would maybe demote this log to log_debug. If ETH fw version is something we support we can just print it. Otherwise if it's something that's deprecated we can add a log saying what is the minimum version. In normal code flow I would probably demote this from log_info

device/topology/topology_discovery.cpp
```diff
@@ -337,4 +343,68 @@ void TopologyDiscovery::patch_eth_connections() {}
...
+                "ETH FW version mismatch for LocalChip {} ETH core {}, found: {}.",
+                device_id,
+                eth_core.str(),
+                eth_fw_version.str());
+        }
+    }
+}
+
+void TopologyDiscovery::verify_eth_version_remote(int asic_id, Chip* chip) {
+    // ETH FW version check incomplete for BH.
```
pjanevskiTT, line 386: same comment as above for moving to BH and WH specific classes

device/topology/topology_discovery.cpp
```diff
@@ -337,4 +343,68 @@ void TopologyDiscovery::patch_eth_connections() {}
...
+    if (chip->get_soc_descriptor().arch != tt::ARCH::WORMHOLE_B0) {
+        return;
+    }
+
+    std::vector<CoreCoord> eth_cores =
+        chip->get_soc_descriptor().get_cores(CoreType::ETH, umd_use_noc1 ? CoordSystem::NOC1 : CoordSystem::NOC0);
+    for (const CoreCoord& eth_core : eth_cores) {
+        // Check ETH FW version
```
pjanevskiTT, line 394: nit: I feel it's clear from the code what is happening, comment can probably be removed

windwardly:
@aleksamarkovicTT, thanks for you comment on my #1333 PR.
This is a little bigger PR to read through, so will get to it later today.

What would help is something in a .md file which talks about version, as when I read the code, I don't know what is right. All I know is tt-umd and luwen disagree and tt-umd had inconsistencies. I could be a better user/reviewer if I had a sense of how the hardware worked. For example, patch level is stored as 8 bits or 64 bits and acted on as though 12 bits is okay. And your diffs here reads an 8 bit patch level from BH.

How does your patch and the work in #1484 merge? Nice if they could come together with a corresponding entry in a version.md doc file in this repo and consolidate the semver_t usage as it flows back and forth from HW to FW to SW. And then reference the document in luwen (or vice-versa, but have only 1 definitive home) so that everything lines up.

Example: in this file:

device/api/umd/device/types/cluster_types.hpp

patch is 8 bits and 12 bits stored in it.

line 191 ```std::uint8_t patch = 0xff;```
line 204 ```patch = version & 0xfff;```

Thank you for working to clean this up. I hit this because trying to run tt-inference-server failed because of version test. I dug into the code and could see problems but didn't know what the right answer was. Looking forward to having this and [luwen](https://github.com/tenstorrent/luwen/issues/113) ironed out.

device/api/umd/device/utils/semver.hpp
```diff
@@ -21,16 +21,26 @@ namespace tt::umd {
...
+
+    constexpr semver_t(uint64_t major, uint64_t minor, uint64_t patch) {
         this->major = major;
         this->minor = minor;
         this->patch = patch;
     }
 
+    /*
+     * Create a semver_t from a 32-bit integer by unpacking the following bits:
+     * 0x00AABCC where A is major, B is minor and C is patch.
```
windwardly: Is the pattern 0x00AABCCC ?

device/topology/topology_discovery_blackhole.cpp
```diff
@@ -274,11 +275,52 @@ void TopologyDiscoveryBlackhole::init_topology_discovery() {
...
+    if (!first_eth_fw_version.has_value()) {
+        log_info(LogUMD, "Established cluster ETH FW version: {}", eth_fw_version.to_string());
+        log_debug(LogUMD, "UMD supported minimum BH ETH FW version: {}", BH_ERISC_FW_SUPPORTED_VERSION_MIN.to_string());
+        first_eth_fw_version = eth_fw_version;
+        if (BH_ERISC_FW_SUPPORTED_VERSION_MIN.major > eth_fw_version.major) {
+            log_warning(LogUMD, "ETH FW major version is older than UMD supported version");
+            eth_fw_problem = true;
+        }
+
+        if (BH_ERISC_FW_SUPPORTED_VERSION_MIN.minor > eth_fw_version.minor) {
```
windwardly: If `BH_ERISC_FW_SUPPORTED_VERSION_MIN.major` is 5 and `eth_fw_version.major` is 6, and `BH_ERISC_FW_SUPPORTED_VERSION_MIN.minor` is 4 and `eth_fw_version.minor` is 2, then the minor warning message is printed even though 5.6 < 6.2 . Use the semver_t `>` test to get the chained comparison.

aleksamarkovicTT: Thanks for noticing, this PR mainly moves code, didn't look into the actual logic until now.

PR #1427, opened 2025-10-08
Add caching of TLB to TTDevice class

pjanevskiTT:
### Issue

Splitting #959  

### Description

Add caching of TLBs inside TTDevice class. This is needed for UMD to not always map/unmap TLBs that it gets from KMD, since this is quite slow. There is only one UC TLB cached for all accesses.

### List of the changes

- Add field for cached TLB inside TTDevice
- Add method for getting cached TLB when needed

### Testing
Tested as part of #959 

### API Changes
/

### DIFF
--- a/device/api/umd/device/tt_device/tt_device.hpp
+++ b/device/api/umd/device/tt_device/tt_device.hpp
@@ -17,6 +17,7 @@
 #include "umd/device/firmware/firmware_info_provider.hpp"
 #include "umd/device/jtag/jtag_device.hpp"
 #include "umd/device/pcie/pci_device.hpp"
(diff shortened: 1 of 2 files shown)


device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -341,6 +342,12 @@ class TTDevice {
     virtual void pre_init_hook(){};
 
     virtual void post_init_hook(){};
+
+    std::unique_ptr<TlbWindow> cached_tlb_window = nullptr;
+
+    TlbWindow *get_cached_tlb_window(tlb_data config);
```
broskoTT, line 348: Is this intended to be used by read/write api in TTDevice?

Issue #1099, opened 2025-07-16
Send ARC messages over AXI for Blackhole where it's possible


nbuncicTT:
Follow up from https://github.com/tenstorrent/tt-umd/pull/1086#discussion_r2209616793

PR #1438, opened 2025-10-09
AXI write for BH where it's possible

nbuncicTT:
### Issue
#1099 

### Description
Send AXI messages where it's possible for blackhole

### List of the changes
- Write over AXI with JTAG and PCIe for local TTDevice if possible, if not write over NOC (board type dependent)
- Add APB address end constant

### Testing
Manual

### API Changes
/


device/tt_device/blackhole_tt_device.cpp
```diff
@@ -22,6 +23,9 @@ namespace tt::umd {
 BlackholeTTDevice::BlackholeTTDevice(std::shared_ptr<PCIDevice> pci_device) :
     TTDevice(pci_device, std::make_unique<blackhole_implementation>()) {
     arc_core = tt::umd::blackhole::get_arc_core(get_noc_translation_enabled(), umd_use_noc1);
+    // ARC tile accessibility over AXI via PCIe depends on the PCIe tile's x-coordinate:
+    // x = 2: ARC not accessible, x = 11: ARC accessible
+    arc_available_over_axi_ = (get_pcie_x_coordinate() == 11) ? true : false;
```
pjanevskiTT: change to `arc_available_over_axi_ = (get_pcie_x_coordinate() == 11)`

pjanevskiTT: I would maybe change this to function 

`is_arc_available_over_axi`

since with this design we are going to call bar_read32 even for remote device which is not that well defined. Everything is going to work with this as well since write/read for ARC core is overrided, but let's not do bar_read for remote device

Maybe we can add that function to all TTDevice classes and then it will just be true for wormhole

Issue #1289, opened 2025-09-10
Avoid creating cluster descriptor in Cluster constructor


pjanevskiTT:
At the moment Cluster constructor works by steps

- Call topology discovery to get cluster descriptor
- Topology discovery creates Chip objects based on the devices connected to the UMD host
- Create Chip objects inside Cluster from cluster descriptor information, same that were created in topology discovery

This works in this way since previously had a `create_ethernet_map` binary which created cluster descriptor and then UMD created Chip objects out of it. We kept it this way to avoid huge code change

Obvious bad side of this is creating Chip objects multiple times. This can be avoided by just not creating Cluster descriptor, rather just picking up chips from topology discovery instead of cluster descriptor and

broskoTT:
> Obvious bad side of this is creating Chip objects multiple times

This is not on a perf path, so I'd say that this shouldn't be a problem. However, merging these two will break our API which is fairly cleanly separated at the moment.

aleksamarkovicTT:
I think this can be closed.

PR #1472, opened 2025-10-15
Get chips from TopologyDiscovery

aleksamarkovicTT:
### Issue
#1289

### Description
This PR extends the TopologyDiscovery interface to return Chips along with the ClusterDescriptor.

### List of the changes
- Made TopologyDiscovery constructor and `create_ethernet_map()` private.
- Renamed `create_cluster_descriptor()` to `discover()`, now returns pair of ClusterDescriptor and map ASIC ID -> Chip.
- Moved TopologyDiscovery creation to separate private static builder method `create_topology_discovery()`.

### Testing
CI

### API Changes
There are API changes in this PR.

### DIFF
--- a/device/topology/topology_discovery.cpp
+++ b/device/topology/topology_discovery.cpp
@@ -144,7 +144,6 @@ void TopologyDiscovery::discover_remote_chips() {
             }
         }
     }
-
     while (!chips_to_discover.empty()) {
         auto it = chips_to_discover.begin();
         uint64_t current_chip_asic_id = it->first;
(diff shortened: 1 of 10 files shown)


device/topology/topology_discovery.cpp
```diff
@@ -5,41 +5,43 @@
...
-#include "umd/device/tt_device/remote_communication.hpp"
-#include "umd/device/tt_device/remote_wormhole_tt_device.hpp"
-#include "umd/device/types/cluster_types.hpp"
 
 extern bool umd_use_noc1;
 
 namespace tt::umd {
 
-std::unique_ptr<ClusterDescriptor> TopologyDiscovery::create_cluster_descriptor(
+std::unique_ptr<TopologyDiscovery> TopologyDiscovery::create_topology_discovery(
```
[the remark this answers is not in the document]
aleksamarkovicTT: The reason why this protected static method is separate is purely because of separation of concerns. The logic on how to create and configure a TopologyDiscovery object is unrelated to the process of discovery, so it's in a different method mainly because of clarity.

.github/workflows/run-tests.yml
```diff
@@ -136,7 +136,7 @@ jobs:
           source .venv/bin/activate
           python -m pip install --upgrade pip wheel pytest
           echo "Installing the wheel..."
-          pip install tt_umd-*.whl
+          pip install tt_umd-*.whl --force-reinstall
```
pjanevskiTT: I would say this change is because of bad rebase, I know you have it in separate PR

aleksamarkovicTT: True

Issue #1471, opened 2025-10-15
Shorten ReadWriteL1/Silicon on Galaxy


aleksamarkovicTT:
This test takes over 20 minutes to complete. Find a way to shorten its duration, so it can be used in metal CI.

pjanevskiTT:
@nbuncicTT also hit this. My calculations say it should take ~32 minutes for 32 chips (1 minute per chip). I am probably ok with doing read/write to single Tensix core L1 for each chip. That should make this probably work in ~20 seconds.

PR #1476, opened 2025-10-16
Change L1 read/write unit test to make it shorter

pjanevskiTT:
### Issue

#1471 

### Description

Make L1 read/write test last shorter amount of time. The problem was with galaxy devices where the test lasted 37 minutes. The test is changed to do IO with only one tensix core per chip, rather than all 80. So this should make test about 80 times faster and execute ~30seconds on galaxy, which is acceptable.

### List of the changes

- Change test to do IO to one tensix core instead of all 80

### Testing
CI

### API Changes
/

### DIFF
--- a/tests/api/test_cluster.cpp
+++ b/tests/api/test_cluster.cpp
@@ -958,26 +958,24 @@ TEST_P(ClusterReadWriteL1Test, ReadWriteL1) {
     for (auto chip_id : cluster->get_target_device_ids()) {
         const SocDescriptor& soc_desc = cluster->get_soc_descriptor(chip_id);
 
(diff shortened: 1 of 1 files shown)


aleksamarkovicTT (approved):
lgtm,
@broskoTT and I thought of testing one Tensix core only for Galaxy, while others have all Tensix cores tested. Or select a random pool of Tensixes for Galaxy instead of all of them. Thoughts?

pjanevskiTT:
I think the main point of this test was to test L1 size since there has been some misunderstanding on size of it on Wormhole/Blackhole so the purpose is to test at least one tensix read/write that is the actual size of L1. Will keep it like this

Issue #1464, opened 2025-10-14
BH galaxy discovered only a single chip


broskoTT:
More info in https://tenstorrent.atlassian.net/browse/SYS-2269

broskoTT:
Fixed, known issue related to syseng eth fw. Thanks Pavle

PR #1478, opened 2025-10-16
Improve TopologyDiscovery with getting ID for chips

pjanevskiTT:
### Issue

#1464 

### Description

Even if I feel that this issue is not a UMD problem, rather galaxy not being flashed with proper FW, there is room for improvement for getting ID of the chips that are not connected to any other chips. So far we took board ID and that can be wrong when all chips are disconnected. 

### List of the changes

- Add function for getting ID of the chip that is not connected to other chips
- Take asic ID for blackhole
- Still take board ID for wormhole

### Testing
CI

### API Changes
/

### DIFF
--- a/device/topology/topology_discovery.cpp
+++ b/device/topology/topology_discovery.cpp
@@ -330,7 +330,7 @@ uint64_t TopologyDiscovery::get_asic_id(Chip* chip) {
         return get_local_asic_id(chip, eth_core);
     }
 
-    return chip->get_tt_device()->get_board_id();
+    return get_unconnected_chip_id(chip);
 }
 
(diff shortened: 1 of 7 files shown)


device/topology/topology_discovery.cpp
```diff
@@ -330,7 +330,7 @@ uint64_t TopologyDiscovery::get_asic_id(Chip* chip) {
         return get_local_asic_id(chip, eth_core);
     }
 
-    return chip->get_tt_device()->get_board_id();
+    return get_unconnected_chip_id(chip);
```
broskoTT: I don't understand why this wasn't like this already, let's chat offline though and feel free to merge this.

Issue #354, opened 2024-11-29
Merge tt_version and semver_t


broskoTT:
Both of these classes represent a version, which can be compared, printed.

I know that they're used for totally different libraries, with different version numbers. But let's consider maybe offering two ways to parse in into a struct, and everything else stays the same? I think there might be some opportunity to merge these and reuse the code.

fyi @joelsmithTT

rekudyu:
Can I get assigned to it ?

rekudyu:
Can I get assigned to it ?

Issue #1424, opened 2025-10-08
Move verifying cluster FW bundle version to TopologyDiscovery


aleksamarkovicTT:
Followup from #1413 
For the purposes of debug mode and catching issues early, this check which can throw exceptions should be moved to an earlier part of creating a Cluster.

PR #1484, opened 2025-10-20
Merged tt_version with semver_t

rekudyu:
### Issue
https://github.com/tenstorrent/tt-umd/issues/354

### Description
This PR removes tt_version as requested 

### List of the changes
Changes to semver_t to make it compatible with tt_version

### Testing
Manually built on metal

### API Changes
None in this PR but following are the related ones.
https://github.com/tenstorrent/tt-metal/pull/31307


device/api/umd/device/utils/semver.hpp
```diff
@@ -25,14 +25,24 @@ class semver_t {
     uint64_t minor;
     uint64_t patch;
 
+    semver_t() {} // To initialize empty constructor.
```
pjanevskiTT: I feel it's clear from the code why this is added, no need for comment in my opinion.

one nit thing: maybe it would be nice to assign some default values to major, minor and patch just to know what are those values when we use empty constructor. 0 maybe? It can be done either in constructor or assign the fields in the class so the empty constructor will keep the fields

device/api/umd/device/utils/semver.hpp
```diff
@@ -25,14 +25,24 @@ class semver_t {
...
+
     semver_t(uint64_t major, uint64_t minor, uint64_t patch) {
         this->major = major;
         this->minor = minor;
         this->patch = patch;
     }
 
+    semver_t(std::uint32_t version){
```
pjanevskiTT: I might prefer removing this constructor and extracting major minor and patch outside of it, but that may be out of scope of this PR, since this was already present in tt_version. I am probably ok with keeping this

pjanevskiTT:
Approved workflows to run, we'll see how CI goes...

device/api/umd/device/cluster.hpp
```diff
@@ -18,6 +18,7 @@
 #include <unordered_set>
```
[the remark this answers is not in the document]
rekudyu: Yes, I was thinking weather everything will be okay or not. I am going to delete tt_version with one last commit.

device/api/umd/device/cluster.hpp
```diff
@@ -613,7 +614,7 @@ class Cluster {
      * Get the ethernet firmware version used by the physical cluster (only implemented for Silicon Backend).
      * Will return a bogus version if no remote chips are supported for the device.
      */
-    tt_version get_ethernet_fw_version() const;
+    tt::umd::semver_t get_ethernet_fw_version() const;
```
aleksamarkovicTT: No need to use a fully qualified type name, just `semver_t`
Also on all other occurences.

device/api/umd/device/cluster.hpp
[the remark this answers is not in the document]
pjanevskiTT: yea if all usages are removed it should be safe to delete it

aleksamarkovicTT:
Just as a tip, you can easily set up a [local precommit check](https://github.com/tenstorrent/tt-umd?tab=readme-ov-file#pre-commit-hook-integration-for-formatting-and-linting) that automatically formats your files.

device/api/umd/device/utils/semver.hpp
```diff
@@ -24,15 +24,38 @@ class semver_t {
     uint64_t major;
     uint64_t minor;
     uint64_t patch;
+    uint64_t pre_release;
```
aleksamarkovicTT: This should be a `std::string` to better conform Semantic Versioning 2.0.0

device/api/umd/device/utils/semver.hpp
```diff
@@ -24,15 +24,38 @@ class semver_t {
...
+    semver_t(uint64_t major, uint64_t minor, uint64_t patch, uint64_t pre_release = 00) {
         this->major = major;
         this->minor = minor;
         this->patch = patch;
+        if (this->pre_release == 01) {
+            std::string pre_release = "rc.1";
+        }
+    }
+
+    static semver_t fb(std::uint32_t version){
```
aleksamarkovicTT: The name of the method is too terse. Change it to `from_firmware_bundle_tag`.

windwardly:
This looks like it address some but not all of what I talked about in [tt_version bug fixes and questions](https://github.com/tenstorrent/tt-umd/pull/1333).

I'm still walking through the details of these diffs and will comment more specifically, but in the mean time wanted to your thoughts on #1333 . Thanks.

windwardly:
How does this PR and #1421 fit together?

rekudyu:
Please trigger ci workflow.

device/cluster.cpp
```diff
@@ -1076,26 +1076,26 @@ void Cluster::verify_sw_fw_versions(int device_id, std::uint32_t sw_version, std
...
         "Software version {}, Ethernet FW version {} (Device {})",
         sw.str(),
         fw_first_eth_core.str(),
         device_id);
     for (std::uint32_t& fw_version : fw_versions) {
-        tt_version fw(fw_version);
+        semver_t fw(fw_version);
         TT_ASSERT(fw == fw_first_eth_core, "FW versions are not the same across different ethernet cores");
         TT_ASSERT(sw.major <= fw.major, "SW/FW major version number out of sync");
         TT_ASSERT(sw.minor <= fw.minor, "SW version is newer than FW version");
```
windwardly: minor should only be compared if major is equal. Consider pulling my repo mentioned in the conversations. Notice how the next diff uses >= on the whole struct.

windwardly:
@rekudyu , please see my earlier comments and just-made review to fix version compare. How are you planning on meshing with PR #1421? Is there something unique here that is not there? Could they be merged?

Nice work. I'm glad for the cleanups!

windwardly:
@rekudyu , SW_VERSION is also wrong. If this PR is merged with #1424 then it doesn't matter because SW_VERSION is removed there. I'll go add a review for that.

rekudyu:
> Also in my repo I mentioned which my work is merged with yours: SW_VERSION as 0x06060000 seems wrong. The top 8 bits will be ignored, which gives 6.0 .
> If shifted right by 8 bits, then the upper 6 would be the version , but the lower 6 winds up as patch level 0x600 or 1536. I'm guessing it was intended to be 6.6.

You are wrong 

(version >> 24) & 0xFF	0x06	 Major
(version >> 16) & 0xFF	0x06	 Minor
(version >> 8)  & 0xFF	0x00 Patch

This is how version is handled in from_firmware_bundle_tag(version).

>
> If this PR is merged with #1424 then SW_VERSION goes away. If this PR is stand alone, then SW_VERSION should be fixed. What would help?

rekudyu:
> My understanding has been upgraded:
> 
> @aleksamarkovicTT said `The version comparison bug is outside of the bounty issue specification.`
> 
> Okay, never mind my comments and limit your work to the feedback of the sponsors.

Nevermind the sponsors, your PR address the seperate and crucial issue. It is better to contribute to open-source projects and not limit your scope to single problem.

aleksamarkovicTT:
> @aleksamarkovicTT ,
> 
> I read through the whole patch again. Do you see this part of my (edited) comments from above as being relevant to the limited scope of this PR?
> 
> > Which brings up: in `from_firmware_bundle_tag()`, do you want the `pre_release` string to always be a number 0-255 or better to map 0 to ""?
> 
> > My guess is all methods and constructors should have the a way to set pre_release to "" like it does in the explicit constructor `36:      semver_t(uint64_t major, uint64_t minor, uint64_t patch, std::string pre_release = "") {`.
> 
> Thanks for your guidance.

@windwardly @rekudyu 
0 should be mapped to an empty string, meaning there is no pre-release element in the version. 
It's important we follow [the semver spec.](https://semver.org/#spec-item-11): "a pre-release version has lower precedence than a normal version"

1-255 should be mapped to -rc1 to -rc255. 

And yes, they should have a default value of "" for pre-release.

rekudyu:
> @rekudyu Sorry for the inconvenience and the incomplete spec - can you implement as shared by the engineers? 

Sure.

windwardly:
@rekudyu, you wrote
> Okay, I changed the precedence and mapped 0-255 to -rc.0-255 now just testing.

I believe 0 should map to "".
Also you write a dot separating rc and the number. I don't believe there is a dot.

@aleksamarkovicTT wrote in respond to my questions:
> 0 should be mapped to an empty string, meaning there is no pre-release element in the version.
> It's important we follow [the semver spec.](https://semver.org/#spec-item-11): "a pre-release version has lower precedence than a normal version"
> 1-255 should be mapped to -rc1 to -rc255.
> And yes, they should have a default value of "" for pre-release.

Also, related to version stuff, the version parse() function in device/api/umd/device/utils/semver.hpp -- how should that look? It parses a string and if the string has an extra ".1", then pre_release is "-rc.1", and for all other values, pre-release is "". So a semver_t can't round trip - be made into a string, then have that string parsed? Again, probably outside the scope and not wanting 'spec creep'. What's the best place to put this concern? New issue?

rekudyu:
resolved conflicts, not sure if I resolved them without breaking your change please review them.

device/api/umd/device/utils/semver.hpp
```diff
@@ -21,16 +21,38 @@ namespace tt::umd {
...
         this->minor = minor;
         this->patch = patch;
+        this->pre_release = pre_release;
+    }
+
+    static semver_t from_firmware_bundle_tag(std::uint32_t version){
+        uint64_t major = (version >> 24) & 0xFF;
+        uint64_t minor = (version >> 16) & 0xFF;
+        uint64_t patch = (version >> 8) & 0xFF;
+        uint64_t pre_release = version & 0xFF;
```
aleksamarkovicTT: Map the pre-release bits to a string here as we discussed.
00 means no pre-release.
01-FF mean "rc1" to "rc255"

device/cluster.cpp
```diff
@@ -1092,14 +1092,11 @@ std::uint64_t Cluster::get_pcie_base_addr_from_device(const ChipId chip_id) cons
...
 
-// TODO: Temporary hack to pass tt-metal build
-std::optional<semver_t> Cluster::get_ethernet_firmware_version() const { return eth_fw_version; }
-
-std::optional<tt_version> Cluster::get_ethernet_fw_version() const {
-    if (!eth_fw_version.has_value()) {
-        return std::nullopt;
-    }
-    return tt_version(eth_fw_version->major, eth_fw_version->minor, eth_fw_version->patch);
+semver_t Cluster::get_ethernet_fw_version() const {
```
aleksamarkovicTT: Restore both methods you have removed when resolving the conflict:
`std::optional<tt_version> get_ethernet_fw_version() const;`
and `std::optional<semver_t> get_ethernet_firmware_version() const;`

windwardly:
> resolved conflicts, not sure if I resolved them without breaking your change please review them.

Looking forward to the comments about the change, as I’m learning so much.
I’m curious about the use and removal of constexpr and optional, to get a sense of their proper application.
Also that pre_release was switched from a string to a uint64_t.
And what initial values are when they are not specified (that for me to learn more about C++).

rekudyu:
> > resolved conflicts, not sure if I resolved them without breaking your change please review them.
> 
> Looking forward to the comments about the change, as I’m learning so much.
> I’m curious about the use and removal of constexpr and optional, to get a sense of their proper application.
> Also that pre_release was switched from a string to a uint64_t.
> And what initial values are when they are not specified (that for me to learn more about C++).

do a google search for the use of constexpr and optional.

the initial value is 0 for pre_release.


Also all things are fixed, please trigger CI workflow.

rekudyu:
hey, please trigger the CI.

marty1885:
As your wish :)

windwardly:
> do a google search for the use of constexpr and optional.

Thank you. Done! Now that I understand more about constexpr, I don't understand why the weakening of the values listed under `ERISC FW version Required by UMD`, and the changes to the semver_t constructors they use . Seems like guaranteed compile time values for constants would be better to have left as it was?

I believe I also understand optional a little better now. It looks like
```
std::optional<semver_t> Cluster::get_ethernet_firmware_version() const {
    if (!eth_fw_version.has_value()) {
        return std::nullopt;
    }   
    return semver_t(eth_fw_version->major, eth_fw_version->minor, eth_fw_version->patch);
}           
```
Could that be reduced to `return eth_fw_version;` as it is unpacking and repacking eth_fw_version? I do see they are not quite identical, as the current function filters out copying `eth_fw_version->pre_release`.

> the initial value is 0 for pre_release.

Thank you. Looks like that's the case most of the time. Looks like if there's a local (on the stack) `semver_t foo(0x00060600);` then not seeing how it gets to be 0, as it's not initialized by the constructor, nor the class definition, nor being a global variable.

> Also all things are fixed, please trigger CI workflow.

Lots there indeed! Good to see how it is going.

Any comment on device/api/umd/device/utils/semver.hpp `27:   uint64_t pre_release;`

@aleksamarkovicTT commented

> This should be a std::string to better conform Semantic Versioning 2.0.0

Additionally this looks like it will print -rc.0 :
```
    std::string str() const {
        return (pre_release) ? fmt::format("{}.{}.{}-rc.{}", major, minor, patch, pre_release)
                             : fmt::format("{}.{}.{}-rc.{}", major, minor, patch, pre_release);
    }
```
Which I get is a simple mistake of a copy with missing edit, but too, points to a missing test case of pre_release == 0.

device/cluster.cpp
```diff
@@ -1092,14 +1092,11 @@ std::uint64_t Cluster::get_pcie_base_addr_from_device(const ChipId chip_id) cons
     }
 }
 
-// TODO: Temporary hack to pass tt-metal build
-std::optional<semver_t> Cluster::get_ethernet_firmware_version() const { return eth_fw_version; }
-
-std::optional<tt_version> Cluster::get_ethernet_fw_version() const {
+std::optional<semver_t> Cluster::get_ethernet_firmware_version() const {
```
aleksamarkovicTT: As @windwardly noted, this can just be `return eth_fw_version;`

aleksamarkovicTT:
> > @rekudyu We're nearing the completion of this issue.
> > Thank you @windwardly for your insightful comments.
> > The only major thing left is correctly supporting pre-release elements of our semantic versioning type.
> > I'll go through the relevant clauses of the spec and discuss them in detail, so the design decisions for `semver_t` in this PR are more clear:
> > > 9. A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version. Identifiers MUST comprise only ASCII alphanumerics and hyphens [0-9A-Za-z-]. Identifiers MUST NOT be empty.  Numeric identifiers MUST NOT include leading zeroes.
> 
> That would require version scheme to be changed, the issue was to add the support for DD(pre_release) which will be the pre-release followed by -rc.

In #354 "Definition of done", 2nd items states:
> Conformance of semver_t to [Semantic Versioning 2.0.0](https://semver.org/), unless impractical for the requirements of UMD.

along with interpreting the firmware bundle version tag.

rekudyu:
@aleksamarkovicTT What to do then ?

aleksamarkovicTT:
> @aleksamarkovicTT What to do ?

It's alright to keep it like this. Once you resolve the conflict and CI tests pass, we can merge.
Thank you for contributing :)

rekudyu:
The pre-commit is failing without any valid reason, they are literally the [same](https://github.com/tenstorrent/tt-umd/actions/runs/19354497442/job/55374447505?pr=1484).

marty1885:
@rekudyu looks like there is a trailing whitespace. You can see it if you Ctrl-A to select text.

image

rekudyu:
> You'll need to submit a patch there too. Yes. I can help coordinate the changes to be merged. But you'll have to provide patches and make them run locally.
> 
Patch ? Can't I open a PR ?

aleksamarkovicTT:
Hello @rekudyu
If tt-metal build in CI passes, and it should, there should be no need for
a PR in tt-metal.

pet, 14. nov 2025. 07:57 Reku Dyu ***@***.***> je napisao/la:

rekudyu:
> > Since tt_version is discontinued it needs to be replaced by semver_t.
> 
> Yes, now I noticed there is code still using it, looks like a copy of old UMD code. Try renaming all the `tt_version` to `semver_t` and make a PR and link it here. If there's any problems, I'll help out.

Okay I will make the PR soon.

marty1885:
@rekudyu I made an mirror on `mirror/rekudyu/main` that is in the repo. It should help you make the Metal side CI testable.

rekudyu:
@aleksamarkovicTT 
https://github.com/tenstorrent/tt-metal/pull/33444
all test passing 🎉

rekudyu:
@windwardly the tests are passing now.

windwardly:
Hi @rekudyu ,

I'm still seeing no comparison of `pre_release` in the test:

```
--- a/tests/misc/test_semver.cpp
+++ b/tests/misc/test_semver.cpp
[..]
     for (const auto &[version_str, expected] : valid_test_cases) {
         semver_t actual(version_str);
         EXPECT_EQ(actual.major, expected.major);
         EXPECT_EQ(actual.minor, expected.minor);
         EXPECT_EQ(actual.patch, expected.patch);
     }
 }
```
I do see your new edits. My understanding of `std::getline()` is the specified terminator is not part of the token returned. It gives up to the '-' (in your new code) but not the '-'. Then the next `std::getline()` starts after the '-'. So erasing 4 chars would be `rc.<digit>`. As the code was, reading up to third dot, would set up the stream to then give the character after the third dot, which for "-rc.1" would give "1". That is, for version strings which match the pattern, it would have worked before. My comment from before was that it would get the digits in token, but it wouldn't mean anything because of the `&& pre_release != 0` that is still in the `if` and it is always `false`. A good compiler would replace the `if` with just the `std::getline()` as that call has side effects, and the body of the `if` (setting `pre_release`) will never be run.

Your current code:
```
--- a/device/api/umd/device/utils/semver.hpp
+++ b/device/api/umd/device/utils/semver.hpp
[..]
 @@ -99,12 +127,17 @@ private:
             if (std::getline(iss, token, '.')) {
                 minor = std::stoull(token);
 
-                if (std::getline(iss, token, '.')) {
+                if (std::getline(iss, token, '-')) {
                     patch = std::stoull(token);
+
+                    if (std::getline(iss, token) && pre_release != 0) {
+                                               token.erase(0, 4); // Erase first four chracter i.e -rc. leaving only the version behind.
+                        pre_release = std::stoull(token);
+                    }
                 }
             }
         }
-        return semver_t(major, minor, patch);
+        return semver_t(major, minor, patch, pre_release);
     }
 };
```

What if you undid this diff, putting it back to dot, and removed the `token.erase()` call
```
-                if (std::getline(iss, token, '.')) {
+                if (std::getline(iss, token, '-')) {
```

And removed the `pre_release` test that will always be false:
```
-                    if (std::getline(iss, token) && pre_release != 0) {
+                    if (std::getline(iss, token)) {
```

And added a `EXPECT_EQ(actual.patch, expected.pre_release);` to the testing code, then run tests?

Maybe begin by adding the `EXPECT_EQ(actual.patch, expected.pre_release);` call and verifying the tests will fail, then go do the edits in `device/api/umd/device/utils/semver.hpp`?

windwardly:
Nice to have the tests working. I see you needed to fix the test data now that pre_release is being tested. Thank you.

As these are changes going into TT's code base, I'll get out of their way to let you all finish it up. Best wishes to you.

rekudyu:
> Nice to have the tests working. I see you needed to fix the test data now that pre_release is being tested. Thank you.
> 
> As these are changes going into TT's code base, I'll get out of their way to let you all finish it up. Best wishes to you.

Thanks

aleksamarkovicTT:
@rekudyu 
Would you like to give me access to your fork of tt-umd so I can push formatting fixes, or would you like me to help you set up precommit? Which platform are you on?

rekudyu:
> @rekudyu 
> Would you like to give me access to your fork of tt-umd so I can push formatting fixes, or would you like me to help you set up precommit? Which platform are you on?

I am on linux but the precommit is little messy it generated different diff locally than on github, I can give write access to you.

Issue #740, opened 2025-04-14
Expose API to get specific NOC port of core type


pjanevskiTT:
Expose API to get one NOC port per NOC endpoint for certain core type. For example DRAM 

Motivated by Joel's use case in writing the tests 

https://github.com/tenstorrent/tt-umd/pull/735/files#diff-c8517e3916c388cd66c3227266cdd65b53c675c6688c9265a0285e2aae813b0bR1007

PR #1489, opened 2025-10-20
Channel parameter for SocDescriptor::get_cores()

aleksamarkovicTT:
### Issue
#740

### Description
/

### List of the changes
- Added optional channel parameter for `SocDescriptor::get_cores()`.

### Testing
CI

### API Changes

### DIFF
--- a/tests/baremetal/test_soc_descriptor.cpp
+++ b/tests/baremetal/test_soc_descriptor.cpp
@@ -119,8 +119,11 @@ TEST(SocDescriptor, SocDescriptorDRAMChannels) {
         soc_desc.get_cores(tt::CoreType::DRAM, tt::CoordSystem::LOGICAL, num_dram_channels + 1), std::runtime_error);
 
     for (int channel = 0; channel < num_dram_channels; channel++) {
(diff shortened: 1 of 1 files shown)


device/api/umd/device/soc_descriptor.hpp
```diff
@@ -13,6 +13,7 @@
 #include <filesystem>
```
pjanevskiTT: Can you please add a test showing how we get channel 0, 1 and 2 for DRAM cores? You can add it to baremetal tests

tests/baremetal/test_soc_descriptor.cpp
```diff
@@ -104,6 +107,21 @@ TEST(SocDescriptor, SocDescriptorWormholeETHLogicalToNOC0) {
     }
 }
 
+TEST(SocDescriptor, SocDescriptorDRAMChannels) {
```
pjanevskiTT: I would maybe expand this with getting DRAM cores for 0, 1 and 2 then comparing with hard coded values from _implementation files. To make sure we are returning the cores we want

Issue #1299, opened 2025-09-12
Consolidate constant declaration in implementation files


broskoTT:
Should be inline constexpr per https://github.com/tenstorrent/tt-umd/pull/1241#discussion_r2341013199

pjanevskiTT:
closed by #1491

PR #1491, opened 2025-10-20
Consolidate constexpr in architecture header files

pjanevskiTT:
### Issue

#1299 

### Description

Consolidate constexpr usage in architecture implementation header files.

### List of the changes

- Change `static const` to `inline constexpr`
- Add static to some vectors that were already const

### Testing
CI

### API Changes
/


nbuncicTT (approved):
Well done!

PR #1493, opened 2025-10-21
Stop downloading noc_parameters.h from github

joelsmithTT:
### Issue
N/A

### Description
At configure, cmake downloads files from GitHub. It adds unnecessary friction because the URLs change, breaking builds. It turns out these files aren't even used.

### List of the changes
* Remove logic to download per-architecture noc_parameters.h from tests/CMakeLists.txt
* Remove the one place noc_parameters.h was included

### Testing
N/A

### API Changes
N/A

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -20,37 +20,6 @@ target_include_directories(
         "$<TARGET_PROPERTY:umd::device,INCLUDE_DIRECTORIES>"
 )
 
-# Download noc_parameters.h for tests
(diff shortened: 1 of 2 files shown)


broskoTT:
Oh we don't even use this file? lol
Just in case, an APC run: 
All runs on brosko/test_removenoc :
- [ ] All post-commit tests : https://github.com/tenstorrent/tt-metal/actions/runs/18679280626

broskoTT (approved):

Issue #919, opened 2025-05-23
dram membar choose NOC port to use


pjanevskiTT:
Update the API of DRAM membar to be able to choose which NOC port of DRAM bank to use when doing a membar.

Needed for help in debugging DRAM issues

pjanevskiTT:
fyi @abhullar-tt

pjanevskiTT:
This helped metal debug the issue. We should also support choosing different NOC ports for each DRAM bank. Work on the branch above can be easily extended

pjanevskiTT:
seems fixed by #2427

PR #1502, opened 2025-10-22
Customize noc dram port for dram_membar

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/919

### Description
Update our code so that it is easily customizable which subchannel is used for dram membar operations.
I think that adding this to API is an overkill, and this approach still allows easy modifications if this is needed for debugging purposes (we only had one such instance up to now).

### List of the changes
- Add a constant to local_chip to control which subchannel is used.

### Testing
No additional testing

### API Changes

### DIFF
--- a/device/chip/local_chip.cpp
+++ b/device/chip/local_chip.cpp
@@ -25,6 +25,8 @@ static_assert(!std::is_abstract<LocalChip>(), "LocalChip must be non-abstract.")
 // TLB size for DRAM on blackhole - 4GB
 const uint64_t BH_4GB_TLB_SIZE = 4ULL * 1024 * 1024 * 1024;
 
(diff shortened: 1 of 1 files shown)


device/chip/local_chip.cpp
```diff
@@ -25,6 +25,8 @@ static_assert(!std::is_abstract<LocalChip>(), "LocalChip must be non-abstract.")
 // TLB size for DRAM on blackhole - 4GB
 const uint64_t BH_4GB_TLB_SIZE = 4ULL * 1024 * 1024 * 1024;
 
+constexpr uint32_t DRAM_SUBCHANNEL_FOR_MEMBAR = 0;
```
nbuncicTT: Make it inline or static if it's currently global scope

PR #1526, opened 2025-10-27
Warm reset with new KMD 2.4.1 IOCTL's

nbuncicTT:
### Issue
 #1574 

### Description
This pull request introduces a new warm reset implementation for PCIe devices, along with supporting changes to device reset logic and timeouts. The main goal is to support a more robust reset flow for devices that use KMD 2.4.1 or newer, including improved detection of device reappearance and additional reset types.

### List of the changes
- Added a new `warm_reset_new` method to the `WarmReset` class.
- Updated the legacy `warm_reset` method to check the kernel driver version and delegate to the new warm reset logic if the version is 2.4.1 or newer.
- Added a helper function `wait_for_device_to_reappear` that uses filesystem and glob operations to detect when a device has reappeared on the PCI bus after a reset, with configurable timeout.
- Extended the `TenstorrentResetDevice` enum to include new reset types: `USER_RESET`, `ASIC_RESET`, `ASIC_DMC_RESET`, and `POST_RESET`, providing finer control over the reset process.
- Introduced a new timeout constant `WARM_RESET_DEVICES_REAPPEAR_TIMEOUT` (10 seconds) for waiting on device reappearance after a reset.
- Added new includes for `fmt/color.h` and `glob.h` to support the new logic for device detection and logging.

### Testing
Manual

### API Changes
/

### DIFF
--- a/device/warm_reset.cpp
+++ b/device/warm_reset.cpp
@@ -40,11 +40,7 @@ void WarmReset::warm_reset(std::vector<int> pci_device_ids, bool reset_m3) {
         pci_device_ids = PCIDevice::enumerate_devices();
     }
 
-    // Perfrom check of driver version.
-    semver_t KMD_VERSION_WITH_NEW_RESET{2, 4, 1};
-
-    auto kmd_version = PCIDevice::read_kmd_version();
-    if (kmd_version >= KMD_VERSION_WITH_NEW_RESET) {
+    if (PCIDevice::is_arch_agnostic_reset_supported()) {
         warm_reset_arch_agnostic(pci_device_ids, reset_m3);
         return;
     }
--- a/device/api/umd/device/pcie/pci_device.hpp
+++ b/device/api/umd/device/pcie/pci_device.hpp
@@ -278,6 +278,12 @@ public:
      */
     static tt::ARCH get_pcie_arch();
 
+    /**
+     * Checks if architecture-agnostic reset is supported by the device by checking the KMD version which enables this
+     * feature.
+     */
+    static bool is_arch_agnostic_reset_supported();
+
 public:
     // TODO: we can and should make all of these private.
     void *bar0_uc = nullptr;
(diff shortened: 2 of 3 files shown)


device/api/umd/device/warm_reset.hpp
```diff
@@ -27,6 +27,11 @@ class WarmReset {
 
     static void warm_reset_wormhole(std::vector<int> pci_device_ids, bool reset_m3);
 
+    static void warm_reset_new(
```
pjanevskiTT: nit: I would prefer that naming is warm_reset and warm_reset_old or legacy, rather than having new. No strong opinions on this. At least that is the pattern we followed so far

device/warm_reset.cpp
```diff
@@ -48,6 +62,98 @@ void WarmReset::warm_reset(std::vector<int> pci_device_ids, bool reset_m3) {
     }
 }
 
+int wait_for_device_to_reappear(
+    const std::string& bdf, const std::chrono::milliseconds timeout_ms = timeout::WARM_RESET_DEVICES_REAPPEAR_TIMEOUT) {
```
pjanevskiTT: nit: maybe rename it to pci_bdf

device/warm_reset.cpp
```diff
@@ -48,6 +62,98 @@ void WarmReset::warm_reset(std::vector<int> pci_device_ids, bool reset_m3) {
...
+                if (std::filesystem::exists(dev_path)) {
+                    device_reappeared = true;
+                }
+            }
+        }
+
+        globfree(&glob_result);
+
+        if (!device_reappeared) {
+            std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
pjanevskiTT: do we need this timeout inside timeout constants file? Is it arbitrary?

pjanevskiTT (approved):
looks good

Issue #1431, opened 2025-10-08
Implement SPI


broskoTT:
SPI flash read/write that is available through luwen is missing in UMD
For example https://github.com/tenstorrent/luwen/blob/main/crates/luwen-if/src/chip/blackhole.rs#L344

PR #1527, opened 2025-10-27
SPI implementation

broskoTT:
### Issue
#1431 

### Description
Implement spi read write, same as implemented by luwen.
Implements protocols for both wh and bh.

### List of the changes
- Copy implementation for wormhole SPI
- Copy implementation for blackhole SPI
- Wrote test in cpp
- Add a corresponding python API and wrote a test for it.

### Testing
There are written c++ and python tests (same as the one in luwen)
TBD: test in tt-flash

### API Changes
This PR has API changes, it adds spi read/write


broskoTT:
Deprecated

Issue #1234, opened 2025-08-28
Address wormhole telemetry and firmware comments


pjanevskiTT:
Address comments that are marked to be addressed as follow up from #893

pjanevskiTT:
closed by #1541

PR #1541, opened 2025-10-28
Follow up on comments for FirmwareInfoProvider

pjanevskiTT:
### Issue

#1234 

### Description

Follow up on comments about DRAM training status. Make FirmwareInfoProvider return vector of training statuses instead of one for single channel.

### List of the changes

- Remove get_dram_training_status for TTDevice
- Change get_dram_trainingd_status inside FirmwareInfoProvider to return vector instead of single status

### Testing
CI

### API Changes
/

### DIFF
--- a/device/firmware/firmware_info_provider.cpp
+++ b/device/firmware/firmware_info_provider.cpp
@@ -99,10 +99,10 @@ uint32_t FirmwareInfoProvider::get_eth_fw_version() const {
 std::vector<DramTrainingStatus> FirmwareInfoProvider::get_dram_training_status(uint32_t num_dram_channels) const {
     // Format of the dram training status is as follows:
     // Each channel gets two bits in the 32-bit value (16 bits used). The lower bits are for lower channels.
-    // Lower of the two bits is for training error and higher of the two bits is for training status.
+    // Lower of the two bits reports the training error and higher of the two bits reports the training status.
     // Example: 0b 00 00 00 00 00 00 01 10
-    // would mean that only channel 0 is trained, channel 1 has the error and other are not trained and don't have
-    // errors. If some channel is harvested the bits are always going to be zero.
+    // would mean that only channel 0 is trained, channel 1 has the error and other channels are not trained and don't
+    // have errors. If some channel is harvested the bits are always going to be zero.
     uint32_t telemetry_data = tt_device->get_arc_telemetry_reader()->read_entry(TelemetryTag::DDR_STATUS);
     std::vector<DramTrainingStatus> statuses;
     for (uint32_t channel = 0; channel < num_dram_channels; ++channel) {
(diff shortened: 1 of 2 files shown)


device/firmware/firmware_info_provider.cpp
```diff
@@ -96,17 +96,25 @@ uint32_t FirmwareInfoProvider::get_eth_fw_version() const {
     return tt_device->get_arc_telemetry_reader()->read_entry(TelemetryTag::ETH_FW_VERSION);
 }
 
-DramTrainingStatus FirmwareInfoProvider::get_dram_training_status(uint32_t dram_channel) const {
+std::vector<DramTrainingStatus> FirmwareInfoProvider::get_dram_training_status(uint32_t num_dram_channels) const {
+    // Format of the dram training status is as follows:
+    // Each channel gets two bits in the 32-bit value (16 bits used). The lower bits are for lower channels.
+    // Lower of the two bits is for training error and higher of the two bits is for training status.
```
nbuncicTT: nit: rephrase to 
`// Lower of the two bits reports the training error and higher of the two bits reports the training status.`

device/firmware/firmware_info_provider.cpp
```diff
@@ -96,17 +96,25 @@ uint32_t FirmwareInfoProvider::get_eth_fw_version() const {
...
-    if (telemetry_data & (1 << (2 * dram_channel + 1))) {
-        return DramTrainingStatus::FAIL;
+    std::vector<DramTrainingStatus> statuses;
+    for (uint32_t channel = 0; channel < num_dram_channels; ++channel) {
+        if (telemetry_data & (1 << (2 * channel))) {
+            statuses.push_back(DramTrainingStatus::SUCCESS);
+        } else if (telemetry_data & (1 << (2 * channel + 1))) {
+            statuses.push_back(DramTrainingStatus::FAIL);
+        } else {
+            statuses.push_back(DramTrainingStatus::IN_PROGRESS);
```
nbuncicTT: Does it make sense to say that the channel status is `IN_PROGRESS` if it's harvested, as this state will always stay the same?

On the other side, there are only three states? If yes, this is okay - I'm just wondering which component knows the harvesting masks, if FirmwareInfoProvider is aware, you can add another enum value ::HARVESTED, and just put that case and for the channel and continue.

device/firmware/firmware_info_provider.cpp
```diff
@@ -96,17 +96,25 @@ uint32_t FirmwareInfoProvider::get_eth_fw_version() const {
...
     uint32_t telemetry_data = tt_device->get_arc_telemetry_reader()->read_entry(TelemetryTag::DDR_STATUS);
-    if (telemetry_data & (1 << (2 * dram_channel))) {
-        return DramTrainingStatus::SUCCESS;
-    }
-
-    if (telemetry_data & (1 << (2 * dram_channel + 1))) {
-        return DramTrainingStatus::FAIL;
+    std::vector<DramTrainingStatus> statuses;
+    for (uint32_t channel = 0; channel < num_dram_channels; ++channel) {
+        if (telemetry_data & (1 << (2 * channel))) {
```
nbuncicTT: nit: I'd prefer having a switch case for enums like in the `Wormhole_18_3_FirmwareInfoProvider::get_dram_training_status` instead of if-else's

Issue #1554, opened 2025-10-29
Add multicast API to Cluster/Chip/TTDevice


pjanevskiTT:
Add API to make clients able to multicast data by using TLBs

PR #1561, opened 2025-10-31
Add support for NOC multicast write

pjanevskiTT:
### Issue

#1554 

### Description

Add support for doing multicast NOC write from host to device. The API is added on Cluster, Chip and TTDevice layer. For now all APIs just fallback into TTDevice layer, which uses UC strict TLB. This should be improved in the future by supporting both memory and reg multicasts. For now Chip is going to assert that access is enabled only to Tensix cores, since Tensix cores fit nicely into a grid and can be used for multicast. We should figure out how to support other core types and whether UMD should impose some restrictions on higher layers.

### List of the changes

- Add API to Cluster, Chip and TTDevice layer
- Implement programming TLB to do a multicast
- Add a test to multicast data to 8x8 grid of tensix cores

### Testing
CI + manual testing

### API Changes
/

### DIFF
--- a/device/tt_device/tt_device.cpp
+++ b/device/tt_device/tt_device.cpp
@@ -628,7 +628,7 @@ void TTDevice::noc_multicast_write(void *dst, size_t size, tt_xy_pair core_start
         throw std::runtime_error("noc_multicast_write is not applicable for JTAG communication type.");
     }
     auto lock = lock_manager.acquire_mutex(MutexType::TT_DEVICE_IO, get_pci_device()->get_device_num());
-    uint8_t *buffer_addr = (uint8_t *)(uintptr_t)(dst);
+    uint8_t *buffer_addr = static_cast<uint8_t *>(dst);
     const uint32_t tlb_index = get_architecture_implementation()->get_reg_tlb();
 
     while (size > 0) {


device/api/umd/device/chip/chip.hpp
```diff
@@ -58,6 +58,7 @@ class Chip {
     virtual void read_from_device_reg(CoreCoord core, void* dest, uint64_t reg_src, uint32_t size) = 0;
```
pjanevskiTT: reminder before merging: mutlicast must be non posted, if we can control that

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -136,6 +136,8 @@ class TTDevice {
     virtual void read_from_device(void *mem_ptr, tt_xy_pair core, uint64_t addr, uint32_t size);
     virtual void write_to_device(const void *mem_ptr, tt_xy_pair core, uint64_t addr, uint32_t size);
 
+    virtual void noc_multicast_write(void *dst, size_t size, tt_xy_pair core_start, tt_xy_pair core_end, uint64_t addr);
```
broskoTT: Please comment on how the core range works in terms of skipping/targeting cores between tensix cores if a large grid is selected (we briefly discussed this offline). Comment separatelly for wh and bh if it is different

tests/api/test_cluster.cpp
```diff
@@ -842,6 +842,52 @@ TEST(TestCluster, GetEthernetFirmware) {
     }
 }
 
+TEST(TestCluster, TestMulticastWrite) {
```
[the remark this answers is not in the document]
pjanevskiTT: added low level test. For pybind I can add all IO function in follow up PR

PR #1562, opened 2025-11-01
Filter out Eth Connections with a Single Inactive Link

tt-asaigal:
### Issue
No Ticket.

### Description
 - Data Structures storing local active ethernet links and ethernet connections have a minor inconsistency
 - Based on the current implementation of `TopologyDiscovery::fill_cluster_descriptor_info()` it's possible for an inactive ethernet link to be reported in a part of an active connection, if it's peer is trained
 - This leads to issues in Cluster Descriptor clients when they query a connection and expect its links to be active
 - Especially worse for BH Scale-Out systems, which have Link Retraining Issues

### List of the changes
- Modify `TopologyDiscovery::fill_cluster_descriptor_info()` to only populate a connection between two links if they're active
- If either links in the connection is inactive, the connection is skipped/not reported

### Testing

### API Changes

### DIFF
--- a/device/topology/topology_discovery.cpp
+++ b/device/topology/topology_discovery.cpp
@@ -272,12 +272,26 @@ std::unique_ptr<ClusterDescriptor> TopologyDiscovery::fill_cluster_descriptor_in
     }
 
     for (auto [ethernet_connection_logical, ethernet_connection_remote] : ethernet_connections) {
-        ChipId local_chip_id = asic_id_to_chip_id.at(ethernet_connection_logical.first);
-        ChipId remote_chip_id = asic_id_to_chip_id.at(ethernet_connection_remote.first);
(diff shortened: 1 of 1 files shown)


pjanevskiTT (commented):
Looking at this I am not sure if https://github.com/tenstorrent/tt-umd/commit/2dca93e5e266551388c0d8613496c835614a06b1 introduced a bug. active_eth_channels_per_chip is now filled for all channels that are trained, not connected to some other active ETH core. Now it seems that all trained ETH cores get into this vector, and my understanding that ETH core is trained even if it's not connected? So I think we should change https://github.com/tenstorrent/tt-umd/blob/main/device/topology/topology_discovery.cpp#L169 to check for connectivity as well

tt-asaigal:
Hey @pjanevskiTT wanted to follow up on this.

broskoTT:
outdated

Issue #1571, opened 2025-11-03
ARM64 alignment bug


joelsmithTT:
**Describe the bug**
A SIGBUS occurs on ARM platforms when writing to a TLB register. This is caused by an unaligned 64-bit memory access in tt::umd::TTDevice::write_tlb_reg in device/tt_device/tt_device.cpp. The code attempts to write an uint64_t to an address that is not 8-byte aligned, which is not permitted on ARM architectures that enforce strict memory alignment.

**Stack trace**
```
joel@ampere ~/git/tt-umd/build $ gdb test/umd/api/api_tests
GNU gdb (Ubuntu 15.0.50.20240403-0ubuntu1) 15.0.50.20240403-git
Copyright (C) 2024 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "aarch64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from test/umd/api/api_tests...
(gdb) run
Starting program: /home/joel/git/tt-umd/build/test/umd/api/api_tests
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/aarch64-linux-gnu/libthread_db.so.1".
warning: could not find '.gnu_debugaltlink' file for /lib/aarch64-linux-gnu/libcap.so.2
Running main() from /home/joel/git/tt-umd/.cpmcache/googletest/96129d89f45386492ae46d6bb8c027bc3df5f949/googletest/src/gtest_main.cc
[==========] Running 72 tests from 16 test suites.
[----------] Global test environment set-up.
[----------] 1 test from ApiChipTest
[ DISABLED ] ApiChipTest.DISABLED_ManualTLBConfiguration
[ RUN      ] ApiChipTest.SimpleAPIShowcase

Program received signal SIGBUS, Bus error.
[...]

PR #1573, opened 2025-11-03
Misc ARM fixes

joelsmithTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/1571

### Description
Fix alignment bugs, block reset on ARM64 where it is unsupported by the driver.

### List of the changes
For ARM64 host:
* Fix alignment issue writing to TLB registers
* Fix alignment issue with memcpy
* Disable reset paths

### Testing
Local

### API Changes
N/A

### DIFF
--- a/device/warm_reset.cpp
+++ b/device/warm_reset.cpp
@@ -4,7 +4,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-#include "utils.hpp"
 #include "api/umd/device/warm_reset.hpp"
 
 #include <chrono>


device/tt_device/tt_device.cpp
```diff
@@ -276,7 +276,13 @@ void TTDevice::write_block(uint64_t byte_addr, uint64_t num_bytes, const uint8_t
     }
 
     const void *src = reinterpret_cast<const void *>(buffer_addr);
-    if (arch == tt::ARCH::WORMHOLE_B0) {
+    bool use_safe_memcpy = false;
+#if defined(__ARM_ARCH) || defined(__riscv)
```
nbuncicTT: Can we make something like:
```
constexpr bool is_arm_platform() {
#if defined(__aarch64__) || defined(__arm__)
    return true;
#else
    return false;
#endif
}
```
in `utils.hpp` and use that function instead of adding preprocessor macros everywhere?

PR #1580, opened 2025-11-04
Minor cleanup/optimization to ttsim build steps

mcraigheadTT:
### Issue
none

### Description
ttsim build steps were somewhat outdated and inefficient.

### List of the changes
Use slightly simpler and faster way to build the ttsim binaries, so that old symlink can be deleted.

### Testing
Flow is tested by the PR

### API Changes
/

### DIFF
--- a/.github/workflows/build-clients.yml
+++ b/.github/workflows/build-clients.yml
@@ -147,11 +147,8 @@ jobs:
 
       - name: Build tt-sim
         run: |
-          cd ttsim/src
-          ../scripts/make.py _out/release_wh/libttsim.so
-          ../scripts/make.py _out/release_bh/libttsim.so
-          ../scripts/make.py _out/release_qsr/libttsim.so
-          cd ../../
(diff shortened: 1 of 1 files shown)


.github/workflows/build-clients.yml
```diff
@@ -147,11 +147,8 @@ jobs:
 
       - name: Build tt-sim
         run: |
-          cd ttsim/src
-          ../scripts/make.py _out/release_wh/libttsim.so
-          ../scripts/make.py _out/release_bh/libttsim.so
-          ../scripts/make.py _out/release_qsr/libttsim.so
```
pjanevskiTT: maybe I am missing something, but do we have a replacement for qsr step?

mcraigheadTT: It can be re-added at whatever point it's being used, I just took it out because right now we're not using it. I'm fine either way

pjanevskiTT (approved):

PR #1587, opened 2025-11-06
#0: Add Active Ethernet connectivity support to ttsim chip

tt-aho:
Add support for multiprocessing for performance, using one process per chip

### Issue
(Link to Github issue(s))

### Description
Can break into multiple PRs if necessary
Adds multiprocessing support for simulation, where one process is spawned per chip.
Add support for adding ethernet connectivity between chips in simulation to support transmitting eth packets.

### List of the changes
1. Add support for multiprocessing
- Factor out common code needed for single and multiprocessing into TTSimChipImpl class
- Add a MultiProcessingTTSimChip class that adds the required SimulationChip APIs and wraps the multiprocessing code
- Add ProcessManager class, which spawns the child process and facilitates parent <-> child communication
- Add a ChildProcessTTSimChip class and executable which handles receiving commands from the parent and issuing the appropriate simulation function, and sending responses back to the parent. TODO: How do we package the executable with UMD and call it? Currently it depends on the user copying the executable into the sim directory, which is not user friendly. Ideally it's packaged with UMD and has a known location.
- For the child processes, these will automatically clock in a loop unless it has to handle messages. If needed, we can add a mode to reduce the clocking to only during reads for example to reduce cpu usage.
2. Add support for ethernet connections in sim
- Sim uses file descriptors for communicating between instances. Add an EthConnection class that handles creating the file descriptors. Currently this uses a Unix socket, but could be updated to use pipes instead if needed.
- Update the TTSimChipImpl construction to create the the required EthConnections based off of the ClusterDescriptor for both local connections and remote connections to support simulating multihost. Uses abstract sockets and creates 1 socket per ethernet queue.
- Add chips_to_clock to the single process TTSimChip. This is required so that we clock all devices in the cluster since there could be inter-device dependencies.
- Add a connect_eth_queues api to the TTSimChip classes and call this after constructing all devices in the cluster. This is used to make sure all the sockets are set up both locally, as well as to remote devices.

### Testing

### API Changes

### DIFF
--- a/device/simulation/process_manager.hpp
+++ b/device/simulation/process_manager.hpp
@@ -1,0 +1,66 @@
+/*
+ * SPDX-FileCopyrightText: (c) 2025 Tenstorrent Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "message_data.hpp"
+#include "umd/device/chip/chip.hpp"
+#include "umd/device/cluster.hpp"
+
+namespace tt::umd {
+
+// ProcessManager handles communication between parent and child processes
+class ProcessManager {
+public:
+    ProcessManager(ChipId chip_id);
+    ~ProcessManager();
+
+    // Start the child process
+    void start_child_process(const std::filesystem::path& simulator_directory, ClusterDescriptor* cluster_desc);
+
+    // Stop the child process
+    void stop_child_process();
+
+    // Send a message and wait for response (blocking)
+    void send_message_with_response(
+        MessageType type, const void* data, uint32_t data_size, void* response_data, uint32_t response_size);
+
+    // Send a message with separate header and data to avoid copying (blocking)
+    void send_message_with_data_and_response(
+        MessageType type,
+        const void* header_data,
+        uint32_t header_size,
+        const void* payload_data,
+        uint32_t payload_size);
+
+    // Check if child process is running
+    bool is_child_running() const { return child_running_; }
+
+    // Get the chip ID
+    ChipId get_chip_id() const { return chip_id_; }
+
+private:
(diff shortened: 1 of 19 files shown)


device/simulation/process_manager.hpp
```diff
@@ -0,0 +1,66 @@
...
+#include <vector>
+
+#include "message_data.hpp"
+#include "umd/device/chip/chip.hpp"
+#include "umd/device/cluster.hpp"
+
+namespace tt::umd {
+
+// ProcessManager handles communication between parent and child processes
+class ProcessManager {
```
broskoTT: Can we make this ProcessManager more generic, so it accepts arbitrary binary and arguments for starting the process, also accepts an abstract Messaging interface protocol which implements some forms of read/write.
Then we can put it in utils folder. We can have that as a separate PR. And write some tests to guarantee the functionality better. This would make it much more reusable in the future.

I realise this would be a good amount of work, so don't treat this as a requirement. If you decide not to solve this, please leave an issue in tt-umd (not assigned to anyone)

device/simulation/tt_sim_chip_impl.cpp
```diff
@@ -0,0 +1,253 @@
...
+bool TTSimChipImpl::connect_eth_links() {
+    bool result = true;
+    for (auto& [channel, eth_connection] : eth_connections_) {
+        if (eth_connection.is_connected()) {
+            continue;
+        }
+        bool connected = eth_connection.connect();
+        if (connected) {
+            auto [write_fd, read_fd] = eth_connection.get_fds();
+            pfn_libttsim_configure_eth_link(channel, write_fd, read_fd);
```
broskoTT: So the ttsim already has to have some interprocess protocol which is used to talk through eth between two processes? Then it makes even more sense that the protocol between umd and ttsim should also be defined in ttsim as well -> another argument to move child_process file there

Issue #1600, opened 2025-11-12
Improve heuristic for verifying ETH FW version


aleksamarkovicTT:
Currently, we establish ETH FW version for a Cluster by checking the first ETH core (1,0 on WH). Exalens breaks this core for their tests, which causes a wrong version to be established.

We need to assume the ETH FW version from the FW bundle version, which is read from ARC and can be trusted.

adjordjevic-TT:
We are still seeing same behavior when using latest UMD. If this issue is not resolved please reopen it. If issue is resolved, we should see what is causing us to not see remote device after running tests, make issue for that and set our [issue](https://github.com/tenstorrent/tt-exalens/issues/691)  blocked on that issue because currently we are blocked on this issue which is closed.

Issue #1008, opened 2025-06-26
Corrupting L1 memory of ETH core causes UMD not to start


tt-vjovanovic:
We have seen two different issues when corrupting L1 memory:
- UMD is hanging during initialization
- Random error regarding ETH connections

Cause: we have tests that test ETH risc and we change its L1 memory to something that is invalid in UMD eyes.
Problem: tt-exalens is debugging tool and we need to be able to start always.
Future: Instead of controlled L1 memory corruption, it can come from within the device with bugs in kernels (targeting wrong noc coordinate and ending up on ETH or having a bug inside ETH kernel), so we need to have debugging tool that can tell us what is wrong

PR #1601, opened 2025-11-12
Set expected ETH FW version according to FW bundle version

aleksamarkovicTT:
### Issue
#1600 #1008

### Description
Instead of taking the ETH FW version from the first found core, use a mapping of expected ETH FW versions in FW bundles and match ETH FW version according to FW bundle version.

### List of the changes
- Introduce `get_expected_erisc_fw_version_from_fw_bundle()` that finds most recent ETH FW version for given FW bundle version
- Change `first_eth_fw_version` to `expected_eth_fw_version` in `TopologyDiscovery`
- Move ETH FW related constants to `erisc_firmware.hpp`

### Testing
CI

### API Changes


device/api/umd/device/topology/topology_discovery.hpp
```diff
@@ -154,9 +154,11 @@ class TopologyDiscovery {
 
     virtual bool verify_fw_bundle_version(Chip* chip);
 
-    // The ETH FW version found on the first discovered local chip, that needs
-    // to match with all of the other discovered ETH FW versions on all chips.
-    std::optional<semver_t> first_eth_fw_version;
+    virtual std::optional<semver_t> get_expected_erisc_fw_version_from_fw_bundle(semver_t fw_bundle_version) const = 0;
```
pjanevskiTT: I would probably move this outside of topology discovery class, maybe to some utils file

pjanevskiTT: I know it needs arch specific behavior, I would still put arch as argument to the function and then decide which map to look into

device/topology/topology_discovery.cpp
```diff
@@ -216,6 +222,10 @@ void TopologyDiscovery::discover_remote_chips() {
             channel++;
         }
     }
+    // Do not pass an ETH FW version to ClusterDescriptor if there are no ETH cores discovered.
+    if (!eth_core_exists) {
```
pjanevskiTT: is this really needed? Can expected_eth_fw_version just have nullopt as default value and then get populated if ETH cores exist

device/topology/topology_discovery.cpp
```diff
@@ -393,6 +403,17 @@ bool TopologyDiscovery::verify_fw_bundle_version(Chip* chip) {
             latest_supported_fw_bundle_version.to_string(),
             arch_to_str(tt_device->get_arch()));
     }
+
+    // Try to infer expected ERISC FW version from FW bundle version
```
pjanevskiTT: nit: add . to the end of the comment

device/topology/topology_discovery_blackhole.cpp
```diff
@@ -293,17 +294,16 @@ bool TopologyDiscoveryBlackhole::verify_eth_core_fw_version(Chip* chip, CoreCoor
     semver_t eth_fw_version = semver_t(major, minor, patch);
 
     bool eth_fw_problem = false;
-    if (!first_eth_fw_version.has_value()) {
-        log_info(LogUMD, "Established ETH FW version: {}", eth_fw_version.to_string());
-        log_debug(LogUMD, "UMD supported minimum BH ETH FW version: {}", BH_ERISC_FW_SUPPORTED_VERSION_MIN.to_string());
-        first_eth_fw_version = eth_fw_version;
+    if (!expected_eth_fw_version.has_value()) {
+        log_info(LogUMD, "Established ETH FW version from first discovered ETH core: {}", eth_fw_version.to_string());
```
pjanevskiTT: I would move this to log debug. If we want to have any logs about ETH FW versions, we can move it to cluster where we log FW bundle as well

pjanevskiTT (approved):
Like the changes, left few small comments to address

PR #1603, opened 2025-11-13
[RFC] Enter and exit power saving mode on card open / close.

yiding:
### Description
One approach to use the power setting arc commands to take the card out of and put the card into low power states upon opening and closing the device.

Mostly putting this here to aid discussions on discord. cc @joelsmithTT 

If going with https://github.com/tenstorrent/tt-kmd/pull/169 instead, then this code would set an ioctl rather than sending the command directly.  In my understanding, an application is supposed to take full control of a card, so it seems to me adding extra indirection and bookkeeping in tt-kmd seems unnecessary, but if the design is going in the directly of tt-kmd taking more responsibilities, then it would make sense.

### List of the changes
1. Tweak some function signatures so the upper bits of arc message code can be set. Currently no other arc commands use the upper bits of the code word so the code tends to strip it in various places.
2. Add functions to LocalChip to send the message to power up and down component card.

### Testing
I ran a script to open the device a few times to ensure everything works. Also ran some pytests.

```python
import ttnn
ttnn.open_mesh_device()
```

Subsequently, check input power field of telemetry: `tt-smi -s | jq '.device_info | map(.smbus_telem.INPUT_POWER)'`

Before:
```json
[
  "0x97",
  "0x92"
]
```

After:
```json
[
  "0x4c",
  "0x47"
]
```

Cards should idle around 80W instead of 140W.

### DIFF
--- a/device/chip/local_chip.cpp
+++ b/device/chip/local_chip.cpp
@@ -201,6 +201,7 @@ void LocalChip::start_device() {
         // If this is supported by the newer KMD, UMD doesn't have to program the iatu.
         init_pcie_iatus();
     }
+    exit_powersave();
     initialize_membars();
 }
 
@@ -215,6 +216,7 @@ void LocalChip::close_device() {
             sysmem_manager_->unpin_or_unmap_sysmem();
         }
     }
+    enter_powersave();
     chip_started_lock_.reset();
 };
 
@@ -737,4 +739,31 @@ TlbWindow* LocalChip::get_cached_pcie_dma_tlb_window(tlb_data config) {
     cached_pcie_dma_tlb_window->configure(config);
     return cached_pcie_dma_tlb_window.get();
 }
+
+void LocalChip::enter_powersave() {
+    switch (soc_descriptor_.arch) {
+        case tt::ARCH::BLACKHOLE: {
+            uint32_t ret = get_tt_device()->get_arc_messenger()->send_message(
+                ((uint32_t)blackhole::ArcMessageType::POWER_SETTING) | (0x3 << 8) | (0b000 << 16));
+            TT_ASSERT(ret == 0, "Failed to set power setting with exit code: {}", ret);
+          break;
+        }
+        default:
+          break;
+    }
+}
+
+void LocalChip::exit_powersave() {
+    switch (soc_descriptor_.arch) {
+        case tt::ARCH::BLACKHOLE: {
+            uint32_t ret = get_tt_device()->get_arc_messenger()->send_message(
+                ((uint32_t)blackhole::ArcMessageType::POWER_SETTING) | (0x3 << 8) | (0b111 << 16));
+            TT_ASSERT(ret == 0, "Failed to set power setting with exit code: {}", ret);
+            break;
+        }
+        default:
+            break;
+    }
+}
+
 }  // namespace tt::umd
--- a/device/api/umd/device/chip/local_chip.hpp
+++ b/device/api/umd/device/chip/local_chip.hpp
@@ -111,6 +111,9 @@ private:
     TlbWindow* get_cached_uc_tlb_window(tlb_data config);
     TlbWindow* get_cached_pcie_dma_tlb_window(tlb_data config);
 
+    void enter_powersave();
+    void exit_powersave();
+
     std::unique_ptr<TlbWindow> cached_wc_tlb_window = nullptr;
     std::unique_ptr<TlbWindow> cached_uc_tlb_window = nullptr;
     std::unique_ptr<TlbWindow> cached_pcie_dma_tlb_window = nullptr;
--- a/device/api/umd/device/types/blackhole_arc.hpp
+++ b/device/api/umd/device/types/blackhole_arc.hpp
@@ -29,6 +29,7 @@ enum class ArcMessageType : uint8_t {
     READ_VM = 0x1D,
     I2C_MESSAGE = 0x1E,
     EFUSE_BURN_BITS = 0x1F,
+    POWER_SETTING = 0x21,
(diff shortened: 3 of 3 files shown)


device/chip/local_chip.cpp
```diff
@@ -737,4 +739,31 @@ TlbWindow* LocalChip::get_cached_pcie_dma_tlb_window(tlb_data config) {
...
+void LocalChip::exit_powersave() {
+    switch (soc_descriptor_.arch) {
+        case tt::ARCH::BLACKHOLE: {
+            uint32_t ret = get_tt_device()->get_arc_messenger()->send_message(
+                ((uint32_t)blackhole::ArcMessageType::POWER_SETTING) | (0x3 << 8) | (0b111 << 16));
+            TT_ASSERT(ret == 0, "Failed to set power setting with exit code: {}", ret);
+            break;
+        }
+        default:
+            break;
```
pjanevskiTT: same as above

device/chip/local_chip.cpp
```diff
@@ -737,4 +739,31 @@ TlbWindow* LocalChip::get_cached_pcie_dma_tlb_window(tlb_data config) {
     cached_pcie_dma_tlb_window->configure(config);
     return cached_pcie_dma_tlb_window.get();
 }
+
+void LocalChip::enter_powersave() {
+    switch (soc_descriptor_.arch) {
+        case tt::ARCH::BLACKHOLE: {
+            uint32_t ret = get_tt_device()->get_arc_messenger()->send_message(
+                ((uint32_t)blackhole::ArcMessageType::POWER_SETTING) | (0x3 << 8) | (0b000 << 16));
```
pjanevskiTT: can you please add comment on why we are modifying this message code this way? Seems very random without a comment
Related to the first comment in this PR, I would probably make this separate message code, I think it would be more readable

yiding: See `power_setting_rqst` in https://github.com/tenstorrent/tt-zephyr-platforms/blob/e10a8ec6814bd9fc465d6c454546ecc20a36bd7c/include/tenstorrent/msgqueue.h#L86

Commands use 4 byte words, with the first word being the command code. Most commands use a single byte command code, pad the remaining 3 bytes, then use the subsequent words as argument. `power_setting_rqst` uses 1 byte of the first word for the command code, and the other flags in the remaining bytes without padding. @jgrowdenTT tells me that this is intentional.

```
Other commands:

    0                   1                   2                   3
    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   | Cmd Code      | Unused                                        |
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   | Argument                                                      |
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   | Argument                                                      |
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+

power_setting_rqst:

    0                   1                   2                   3
    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   | Cmd code      | Flags valid   | Settings valid|A|M|T| Future  |
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   | ... use     |1| Future use ...                                |


   1. Reserved

```

It's just incidental that the power setting message can be expressed with a small modification of the existing send_message API, since as you can see it doesn't really follow the typical boundaries at all. I think if this kind of tighter packed message will be the norm there needs to be another API to send them, rather than send_message(Code, Arg, Arg) as it is now...

device/api/umd/device/arc/blackhole_arc_message_queue.hpp
```diff
@@ -51,7 +51,7 @@ class BlackholeArcMessageQueue {
      * Send ARC message. The call of send_message is blocking, timeout is to be implemented.
      */
     uint32_t send_message(
-        const ArcMessageType message_type,
+        const uint32_t message_type,
```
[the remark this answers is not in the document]
yiding: See the [other comment](https://github.com/tenstorrent/tt-umd/pull/1603/files#r2524617237) down below about why this is necessary. 

I think you are suggesting that I just add two entries to `ArcMessageType` enum? e.g.

```
POWER_SETTING_ALL_ENABLED = 0x21 | (0x3 << 8) | (0b111 << 16);
POWER_SETTING_ALL_DISABLED = 0x21 | (0x3 << 8) | (0b000 << 16);
```

In that case I would have to widen `ArcMessageType` to 32bits, right now it's `uint8_t`.

device/chip/local_chip.cpp
broskoTT: @jgrowdenTT What's the reason for having arguments accepted as other arguments?

yiding:
With https://github.com/tenstorrent/tt-kmd/pull/169 being merged now this code should instead use the ioctl instead of directly sending the commands. So in the current state there's no reason to commit it.

PR #1613, opened 2025-11-17
Verify ERISC FW hash in TopologyDiscovery

aleksamarkovicTT:
### Issue
#1612 #1637

### Description
UMD will now calculate a SHA256 hash of the read only segment of ETH FW containing the .text section.
The range for WH is verified to be `0x2000-0x8FFF` by the WH ERISC FW linker script.
The range for BH starts at `0x70000` and the end depends on the size of the text segment of the ERISC FW.
A warning is logged if the hash from the read L1 of the core does not match the pre-calculated hash and the core will not be used for remote chip discovery. 

### List of the changes
- Introduce third party dependency picosha2
- Added hashes of most supported WH and BH ERISC FW versions
- Verify ETH FW hash for WH and BH as a part of TopologyDiscovery

### Testing
CI

### API Changes

### DIFF
--- a/device/api/umd/device/utils/semver.hpp
+++ b/device/api/umd/device/utils/semver.hpp
@@ -114,10 +114,10 @@ namespace std {
 template <>
 struct hash<tt::umd::semver_t> {
     std::size_t operator()(const tt::umd::semver_t& v) const noexcept {
-        std::size_t h1 = std::hash<uint64_t>{}(v.major);
-        std::size_t h2 = std::hash<uint64_t>{}(v.minor);
-        std::size_t h3 = std::hash<uint64_t>{}(v.patch);
-        return h1 ^ (h2 << 1) ^ (h3 << 2);
+        // Assumption: size_t is 64-bit.
+        // Layout: [ Major (16) | Minor (16) | Patch (32) ]
+        return (static_cast<size_t>(v.major) << 48) | (static_cast<size_t>(v.minor) << 32) |
+               static_cast<size_t>(v.patch);
     }
 };
 }  // namespace std


device/CMakeLists.txt
```diff
@@ -240,6 +240,7 @@ target_link_libraries(
         rt
         spdlog::spdlog_header_only
         fmt::fmt-header-only
+        picosha2::picosha2
```
pjanevskiTT: do we need to add a dependency? Can we use something from std:: to calculate the hash? 

Have you considered alternatives? Any specific reasoning behind this one in particular?
I am not criticizing this, just asking the questions since I don't really know any hashing lib that is standard

device/api/umd/device/firmware/erisc_firmware.hpp
```diff
@@ -0,0 +1,80 @@
...
+#include <optional>
+#include <unordered_map>
+#include <vector>
+
+#include "umd/device/utils/semver.hpp"
+
+namespace tt::umd::erisc_firmware {
+
+// ERISC FW versions required by UMD.
+constexpr semver_t BH_ERISC_FW_SUPPORTED_VERSION_MIN = semver_t(1, 4, 1);
```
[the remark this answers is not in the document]
pjanevskiTT: apply to others as welll

device/firmware/firmware_utils.cpp
```diff
@@ -58,6 +65,32 @@ semver_t get_eth_fw_version_from_telemetry(const uint32_t telemetry_data, tt::AR
     return semver_t((telemetry_data >> 16) & 0xFF, (telemetry_data >> 8) & 0xFF, telemetry_data & 0xFF);
 }
 
+std::optional<bool> verify_eth_fw_integrity(
+    TTDevice* tt_device, tt_xy_pair eth_core, semver_t eth_fw_version, tt::ARCH arch) {
+    const auto* eth_fw_hashes = &erisc_firmware::WH_ERISC_FW_HASHES;
```
pjanevskiTT: why assign anything to this var, can it be nullptr, and then assign based on arch?

aleksamarkovicTT: It can, the only reason I assigned it this is `auto` cannot infer the type from `nullptr`.

device/api/umd/device/firmware/erisc_firmware.hpp
```diff
@@ -0,0 +1,80 @@
...
+constexpr semver_t WH_ERISC_FW_SUPPORTED_VERSION_MIN = semver_t(6, 0, 0);
+constexpr semver_t WH_ERISC_FW_ETH_BROADCAST_SUPPORTED_MIN = semver_t(6, 5, 0);
+constexpr semver_t WH_ERISC_FW_ETH_BROADCAST_VIRTUAL_COORDS_MIN = semver_t(6, 8, 0);
+
+// Maps firmware bundle versions to their corresponding ERISC firmware versions.
+// Bundle versions between entries inherit the ERISC version from the previous entry.
+static const std::vector<std::pair<semver_t, semver_t>> WH_ERISC_FW_VERSION_MAP = {
```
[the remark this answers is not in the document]
aleksamarkovicTT: Decided against this due to incompatible pointer types for array.

Issue #1433, opened 2025-10-08
Support tt-flash


broskoTT:

broskoTT:
Rebased again, reopened a PR which is on a branch (not fork), retested on umd mainline v0.9.5. Fixed two minor issues, provided proof it works on n150 n300 p150 , and that `--use-luwen` still works.
Waiting on review

broskoTT:
Synced with Sam on reviewing this, expecting reviews in the following period once she gets her hands on syseng pool of machines

broskoTT:
Nothing to do here anymore, but longer term I don't think it is going to be UMD backed

PR #1619, opened 2025-11-18
Fix BH arc messages

broskoTT:
### Issue
While working on https://github.com/tenstorrent/tt-umd/issues/1433

### Description
I realised that arc messages for BH are not implemented correctly. There is a difference in how the arguments are passed, and additionally the interface allows more than 2 arguments.
This PR fixes the api

### List of the changes
- Change the api throughout UMD to accept a vector of arguments rather than 2 arguments.
- Introduce checks so that wh still requires 2 args.
- Add arc_msg py binding entries
- Add two more for compatibility, one with explicit arg0, arg1, and another one with "timeout" parameter in seconds.
- Add a test in python
- arc_msg in pybinding automatically prepend common wh arc code message prefix

### Testing
Tested in tools through branch brosko/more_py_api2
Added a python test

### API Changes
Changed arc_msg in cluster api, but there is no usage of it.
Added arc_msg to python api

### DIFF
--- a/device/api/umd/device/cluster.hpp
+++ b/device/api/umd/device/cluster.hpp
@@ -564,6 +564,17 @@ public:
 
     //---------- Misc system functions
 
+    // TODO: Deprecated. To be removed once clients switch to the new arc_msg API.
+    int arc_msg(
+        int logical_device_id,
+        uint32_t msg_code,
+        bool wait_for_done = true,
+        uint32_t arg0 = 0,
+        uint32_t arg1 = 0,
+        const std::chrono::milliseconds timeout_ms = timeout::ARC_MESSAGE_TIMEOUT,
+        uint32_t* return_3 = nullptr,
+        uint32_t* return_4 = nullptr);
+
     /**
      * Issue message to device, meant to be picked up by ARC firmware.
      *
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -915,6 +915,18 @@ int Cluster::arc_msg(
     int logical_device_id,
     uint32_t msg_code,
     bool wait_for_done,
+    uint32_t arg0,
+    uint32_t arg1,
+    const std::chrono::milliseconds timeout_ms,
+    uint32_t* return_3,
+    uint32_t* return_4) {
+    return get_chip(logical_device_id)->arc_msg(msg_code, wait_for_done, {arg0, arg1}, timeout_ms, return_3, return_4);
+}
+
+int Cluster::arc_msg(
+    int logical_device_id,
+    uint32_t msg_code,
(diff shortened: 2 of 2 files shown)


device/arc/blackhole_arc_message_queue.cpp
```diff
@@ -87,10 +87,19 @@ std::array<uint32_t, BlackholeArcMessageQueue::entry_len> BlackholeArcMessageQue
 }
 
 uint32_t BlackholeArcMessageQueue::send_message(
-    const ArcMessageType message_type, uint16_t arg0, uint16_t arg1, const std::chrono::milliseconds timeout_ms) {
-    uint32_t arg = arg0 | (arg1 << 16);
+    const ArcMessageType message_type, const std::vector<uint32_t>& args, const std::chrono::milliseconds timeout_ms) {
+    if (args.size() > 7) {
```
pjanevskiTT: nit: I know there is a comment on 7 being the limit on number of arguments, might not be bad to repeat it here. Someone looking just at cpp file might get confused

device/arc/blackhole_arc_message_queue.cpp
```diff
@@ -87,10 +87,19 @@ std::array<uint32_t, BlackholeArcMessageQueue::entry_len> BlackholeArcMessageQue
...
+        throw std::runtime_error(
+            fmt::format("Too many arguments {} > 7 for message type {}", args.size(), (uint32_t)message_type));
```
pjanevskiTT: nit: expand the message to say that this is Blackhole limit

device/arc/blackhole_arc_message_queue.cpp
```diff
@@ -87,10 +87,19 @@ std::array<uint32_t, BlackholeArcMessageQueue::entry_len> BlackholeArcMessageQue
...
+    }
+
+    // Initialize with zeros for unused args
```
pjanevskiTT: nit: add . to the end of the comment

device/arc/blackhole_arc_message_queue.cpp
```diff
@@ -87,10 +87,19 @@ std::array<uint32_t, BlackholeArcMessageQueue::entry_len> BlackholeArcMessageQue
...
+    std::array<uint32_t, BlackholeArcMessageQueue::entry_len> request = {(uint32_t)message_type, 0, 0, 0, 0, 0, 0, 0};
 
-    std::array<uint32_t, BlackholeArcMessageQueue::entry_len> request = {(uint32_t)message_type, arg, 0, 0, 0, 0, 0, 0};
+    // Copy provided arguments
```
pjanevskiTT: nit: same as above

PR #1621, opened 2025-11-18
Define SPI interface

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/1431

### Description
This PR is an introductory PR to SPI. It defines the interface, how it's used, how to construct TTDevice to be able to use it, and some tests. It will remain non functional after this PR, and will be implemented per arch in separate PRs.

### List of the changes
- Introduce spi.hpp
- Add allow_spi argument to TTDevice creation which will control whether SPI access was enabled
- CPP and python tests which test the spi functionality.

### Testing
Also unit tests were carefully constructed.
Also, I executed code step by step to confirm everything matches luwen's implementation.

### API Changes
Introduces spi read/write.

### DIFF
--- a/nanobind/py_api_tt_device.cpp
+++ b/nanobind/py_api_tt_device.cpp
@@ -348,9 +348,9 @@ void bind_tt_device(nb::module_ &m) {
 
     nb::class_<SPITTDevice>(m, "SPITTDevice")
         .def(
-            [](TTDevice &device) -> std::unique_ptr<SPITTDevice> { return std::make_unique<SPITTDevice>(&device); },
+            "__init__",
+            [](SPITTDevice *self, TTDevice &device) { new (self) SPITTDevice(&device); },
             nb::arg("device"),
-            nb::rv_policy::take_ownership,
             "Create an SPITTDevice for the given TTDevice")
         .def(
             "read",


device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -354,6 +378,7 @@ class TTDevice {
     LockManager lock_manager;
     std::unique_ptr<ArcTelemetryReader> telemetry = nullptr;
     std::unique_ptr<FirmwareInfoProvider> firmware_info_provider = nullptr;
+    std::unique_ptr<SPI> spi_ = nullptr;
```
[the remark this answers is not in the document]
pjanevskiTT: I don't agree with this. I think we agreed earlier to keep anything that we have as a mechanism inside TTDevice, not dependent on how easy/difficult it is to use, as soon as we don't enforce any policy on the usage of it through TTDevice

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -52,9 +53,13 @@ class TTDevice {
     /**
      * Creates a proper TTDevice object for the given device number.
      * Jtag support can be enabled.
+     * By default, reads and writes to SPI are not allowed, you need to manually override this behavior if you want to
+     * use SPI.
      */
-    static std::unique_ptr<TTDevice> create(int device_number, IODeviceType device_type = IODeviceType::PCIe);
-    static std::unique_ptr<TTDevice> create(std::unique_ptr<RemoteCommunication> remote_communication);
+    static std::unique_ptr<TTDevice> create(
+        int device_number, IODeviceType device_type = IODeviceType::PCIe, bool allow_spi = false);
```
pjanevskiTT: allow_spi flag doesn't make too much sense to me. If this is not HW limitation, not really sure what is the point of it. It's just allowing someone to do SPI reads/writes. If we already have the protocol inside UMD, it's easy for someone to use it just by switching this to true. Same goes if this is moved to separate class, if we have the protocol I don't see why creating another object instead of explicitly using TTDevice API is any more difficult.
If SPI is that destructive, we maybe don't need to add it to UMD at all

nanobind/tests/test_py_tt_device.py
```diff
@@ -105,3 +105,91 @@ def test_arc_msg(self):
...
+    @unittest.skip("Disabled by default - potentially destructive SPI test. Remove this decorator to run.")
+    def test_spi_read_write(self):
+        """Test SPI read/write operations on discovered devices."""
+        cluster_descriptor = tt_umd.TopologyDiscovery.create_cluster_descriptor()
+        umd_tt_devices = {}
+        chip_to_mmio_map = cluster_descriptor.get_chips_with_mmio()
+
+        # Create TTDevice instances for all chips (local and remote)
+        # Note that we have to enable SPI for the test to work.
+        allow_spi = True;
```
pjanevskiTT: nit: maybe remove this variable, if we want to use SPI in this test just hardcode everything to true in tests. Doesn't look like this is something configurable

tests/api/test_tt_device.cpp
```diff
@@ -310,3 +310,135 @@ TEST(ApiTTDeviceTest, MulticastIO) {
...
+        for (uint8_t byte : read_data) {
+            if (byte != 0) {
+                has_data = true;
+                break;
+            }
+        }
+        EXPECT_TRUE(has_data) << "SPI read should return non-zero board info data for device " << chip_id;
+
+        // Test read-modify-write on spare/scratch area
+        uint32_t spare_addr = 0x20134;  // Wormhole spare area, but I also tested on Blackhole.
```
pjanevskiTT: nit: if this address works for both, why have the comment for wormhole specific area

PR #1622, opened 2025-11-18
Wormhole SPI

broskoTT:
### Issue
#1431 

### Description
Implement SPI IO for Wormhole

### List of the changes
- Implement Wormhole SPI
- Minor change in wormhole constants

### Testing
Also unit tests were carefully constructed.
Also, I executed code step by step to confirm everything matches luwen's implementation.

### API Changes

### DIFF
--- a/device/arc/smbus_arc_telemetry_reader.cpp
+++ b/device/arc/smbus_arc_telemetry_reader.cpp
@@ -25,7 +25,7 @@ void SmBusArcTelemetryReader::get_telemetry_address() {
         arc_msg_return_values,
         {0, 0});
 
-    static constexpr uint64_t noc_telemetry_offset = wormhole::ARC_CSM_OFFSET_NOC;
+    static constexpr uint64_t noc_telemetry_offset = 0x810000000;
     telemetry_base_noc_addr = arc_msg_return_values[0] + noc_telemetry_offset;
 }
 
--- a/device/arc/wormhole_arc_telemetry_reader.cpp
+++ b/device/arc/wormhole_arc_telemetry_reader.cpp
@@ -20,7 +20,7 @@ WormholeArcTelemetryReader::WormholeArcTelemetryReader(TTDevice* tt_device) : Ar
 }
 
 void WormholeArcTelemetryReader::get_telemetry_address() {
-    static constexpr uint64_t noc_telemetry_offset = wormhole::ARC_CSM_OFFSET_NOC;
+    static constexpr uint64_t noc_telemetry_offset = 0x810000000;
     uint32_t telemetry_table_addr_offset;
     tt_device->read_from_device(
         &telemetry_table_addr_offset,


device/arc/wormhole_spi.cpp
```diff
@@ -0,0 +1,451 @@
...
+
+    val = spi_ser_slave_disable(0);
+    tt_device_->write_to_arc_apb(&val, SPI_SER, sizeof(val));
+
+    // Wait for lock operation to complete
+    uint8_t status_val;
+    uint8_t prev_status = 0;
+    uint32_t status_count = 0;
+    bool first_status = true;
+    while (((status_val = read_status(SPI_RD_STATUS_CMD)) & 0x1) == 0x1) {
```
nbuncicTT: I understand that we need to wait, but does it make sense to add a timeout here if it doesn't finish and to have a return value if the lock failed?

nanobind/py_api_tt_device.cpp
```diff
@@ -87,7 +91,128 @@ void bind_tt_device(nb::module_ &m) {
...
             nb::arg("core_x"),
             nb::arg("core_y"),
-            nb::arg("addr"));
+            nb::arg("addr"))
+        .def(
+            "arc_msg",
+            [](TTDevice &self,
+               uint32_t msg_code,
+               bool wait_for_done = true,
+               std::vector<uint32_t> args = {},
```
nbuncicTT: nit: If you opt out for the variadic template I mentioned above, you need a different entry-point for the python binding as it can bind to a template function.

So instead of that interface (which would be used only for C++ users) we would have a different API for python bindings:
```cpp
uint32_t send_message_dynamic(uint32_t msg_code,
                                  std::vector<uint32_t>& return_values,
                                  const std::array<uint32_t, 7>& args,
                                  size_t arg_count,
                                  std::chrono::milliseconds timeout) {
        return send_message_impl(msg_code, return_values, args, arg_count, timeout);
    }
```
Note: send_message_impl is the same one in the first comment.

tests/api/test_tt_device.cpp
```diff
@@ -310,3 +310,135 @@ TEST(ApiTTDeviceTest, MulticastIO) {
         }
     }
 }
+
+// This test can be destructive, and should not normally run.
+// Make sure to only run it on hardware which has recovery support.
+// The test is disabled by default. To enable it, run with --gtest_also_run_disabled_tests
+TEST(ApiTTDeviceTest, DISABLED_SPIReadWrite) {
```
nbuncicTT: Instead of having this one big test, try to chunk it out in 2 or 3 tests where you test the read, read-modify-write, and reading a wider area (or something similar).

Also, as we've discussed, consider moving this test out of ApiTTDeviceTest to SPIFlashTest or something similar as it shouldn't be in the TTDevice class, I'd also make a separate file for SPI tests and put them to disabled (so when somebody runs the tests for SPI, he must enable them explicitly - because of the dangers of using the SPI interface).

nbuncicTT (approved):
Well done adding this big feature to UMD.

PR #1624, opened 2025-11-18
Implement get_spi_fw_bundle_version for BH

broskoTT:
### Issue
Related to #1431  but on a path of https://github.com/tenstorrent/tt-umd/issues/1433

### Description
Blackhole provides a specific way to get the spi fw bundle version. Implement it, and surface through the api.

### List of the changes
- Implement get_spi_fw_bundle_version as it is implemented in luwen
- Surface through python API

### Testing

### API Changes
Introduces get_spi_fw_bundle_version in python api

### DIFF
--- a/device/arc/blackhole_spi_tt_device.cpp
+++ b/device/arc/blackhole_spi_tt_device.cpp
@@ -8,10 +8,9 @@
 
 #include <algorithm>
 #include <chrono>
-#include <functional>
-#include <iostream>
 #include <stdexcept>
 #include <thread>
+#include <tt-logger/tt-logger.hpp>
 #include <utility>
 #include <vector>
 
@@ -29,7 +28,7 @@ namespace {
 std::optional<uint32_t> read_varint(const uint8_t* data, size_t size, size_t& pos) {
     uint32_t value = 0;
     int shift = 0;
-    while (pos < size && shift < 32) {
+    while ((pos < size) && (shift < 32)) {
         uint8_t byte = data[pos++];
         value |= static_cast<uint32_t>(byte & 0x7F) << shift;
         if (!(byte & 0x80)) {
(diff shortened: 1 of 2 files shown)


device/arc/blackhole_spi_tt_device.cpp
```diff
@@ -45,6 +69,82 @@ static SpiBufferInfo get_spi_buffer_info(TTDevice* device) {
...
 
+// Template member function implementation.
+template <typename Reader>
+std::optional<TtBootFsFd> BlackholeSPITTDevice::find_boot_fs_tag(const Reader& reader, const std::string& tag_name) {
+    uint32_t curr_addr = 0;
+    uint32_t entry_count = 0;
+
+    while (true) {
+        if (entry_count >= BOOT_FS_MAX_ENTRIES_SCAN) {
+            std::cout << "Safety exit, tag not found" << std::endl;
```
nbuncicTT: nit: Let's not use std::cout here, we could either use log_warning or log_debug.

device/arc/blackhole_spi_tt_device.cpp
```diff
@@ -45,6 +69,82 @@ static SpiBufferInfo get_spi_buffer_info(TTDevice* device) {
     return {buffer_addr, buffer_size};
 }
 
+// Template member function implementation.
+template <typename Reader>
+std::optional<TtBootFsFd> BlackholeSPITTDevice::find_boot_fs_tag(const Reader& reader, const std::string& tag_name) {
```
nbuncicTT: I think this function is over-engineered. The function has a template for an abstract type which is being invoked (and I would be okay with this if this is generalized) - and then we pass in a lambda which invokes a method from the same class (via the this pointer).

Omit the template and callable completely and invoke the function directly instead of calling std::invoke on a template object.

i.e.:

`this->read(curr_addr, reinterpret_cast<uint8_t*>(&fd), sizeof(TtBootFsFd));`

instead of 

`std::invoke(reader, curr_addr, sizeof(TtBootFsFd), reinterpret_cast<uint8_t*>(&fd));`

nbuncicTT (approved):

Issue #1547, opened 2025-10-28
Gather machine specification info in runtime for perf benchmarks


pjanevskiTT:
Together with IO performance in perf microbenchmarks, we should also gather machine specification info and report it in tests. That includes

- CPU 
- number of PCIe lanes connected to the card
- Potentially GDDR on host machine

PR #1627, opened 2025-11-18
Host spec gathering script

nbuncicTT:
### Issue
#1547 

### Description
This pull request adds a new tool and updates the benchmark workflow to improve the collection and reporting of host machine specifications during CI runs. The main addition is a robust Python script for gathering detailed system information, with both automatic and manual fallback modes, and exporting the data in multiple formats.

### List of the changes
* Added a new script `tools/gather_host_specs.py` that collects comprehensive machine specifications, including OS, kernel, memory, CPU details, Tenstorrent driver version, and PCIe information. The script first attempts to use `tt-smi` for data collection and falls back to manual sysfs and psutil methods if unavailable. It supports output in JSON, CSV, and Markdown formats via command-line flags.
* Updated the `.github/workflows/run-benchmarks.yml` workflow to install the necessary Python dependencies (`pandas`, `psutil`, `tabulate`) and invoke the new `gather_host_specs.py` script before running TLB benchmarks, ensuring host specs are always collected on supported architectures.

### Testing
CI and Manual

### API Changes
/


tools/gather_host_specs.py
```diff
@@ -0,0 +1,415 @@
+# SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+This script gathers machine specification information.
+
+It is divided into two parts based on the availability of the `tt-smi` tool:
+1. **If `tt-smi` is not available**, information is gathered manually.
```
pjanevskiTT: nit: I would probably swap steps 1 and 2 in this comment. First step should be if tt-smi is available, second is manual option

tools/gather_host_specs.py
```diff
@@ -0,0 +1,415 @@
+# SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
```
pjanevskiTT: not gonna review this file in that much detail if you have tested it and it works

tools/gather_host_specs.py
```diff
@@ -0,0 +1,415 @@
...
+        "tt_pcie_lane_info": tt_pcie_lane_info
+    }
+    return data
+
+def gather_specs():
+    """
+    Dispatcher function. Tries tt-smi first, then falls back to manual.
+    """
+    try:
+        # Try to run tt-smi -s
```
pjanevskiTT: as a general comment, I don't really like relying on external tool outside of UMD (tt-smi in this case). I feel if we know how to gather this information we should implement just that in this script. If tt-smi has the use case for the same stuff as we do, probably a signal we should have common code with them that they can call into. So my opinion here is to remove tt-smi calls and gather specs manually

.github/workflows/run-benchmarks.yml
```diff
@@ -68,6 +68,21 @@ jobs:
         shell: bash
```
pjanevskiTT: can we export this data directly into main .md file for UMD benchmarks after running the script. I think we should at least add an example of how report looks like. For example run it on CI machine we are running benchmarks and paste it into md file, if it's not possible to do it automatically

PR #1650, opened 2025-11-24
Rename ttsim to ttsim-private

mcraigheadTT:
### Issue
none

### Description
tenstorrent/ttsim is being renamed to ttsim-private to make room for public repo

### List of the changes
Change repo path

### Testing
n/a

### API Changes
/


broskoTT:
I've pinged somebody over slack to update our token so we have access to ttsim-private

Issue #636, opened 2025-03-25
Investigate data race issues


broskoTT:
These issues are surfaced by TSAN run here: https://github.com/tenstorrent/tt-umd/actions/runs/14003958381/job/39215282510#step:8:865
Relevant PR: https://github.com/tenstorrent/tt-umd/pull/628

broskoTT:
Hoping this might fix some other issues we were seeing.

broskoTT:
This task tracks enabling Asan Tsan runs on our CI

pjanevskiTT:
+1 if it's not already there from https://tenstorrent.slack.com/archives/C07L16Z59AP/p1763528462295799

PR #1651, opened 2025-11-25
Fix TSAN data race for remote communication

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/636

### Description
TSAN found these failures, specifically the flush_non_mmio_ was not guarded by the locks.

### List of the changes
- Move locking in read/write non_mmio to the top of the function

### Testing
CI TSAN tests, with some other changes combined: https://github.com/tenstorrent/tt-umd/actions/runs/19664175096

### API Changes

### DIFF
--- a/device/tt_device/remote_communication_legacy_firmware.cpp
+++ b/device/tt_device/remote_communication_legacy_firmware.cpp
@@ -340,6 +340,7 @@ void RemoteCommunicationLegacyFirmware::write_to_non_mmio(
     std::vector<int> broadcast_header,
     const std::chrono::milliseconds timeout_ms) {
     auto lock = lock_manager_.acquire_mutex(MutexType::NON_MMIO, local_tt_device_->get_communication_device_id());
+    flush_non_mmio_ = true;
 
     using data_word_t = uint32_t;
     constexpr int DATA_WORD_SIZE = sizeof(data_word_t);


device/tt_device/remote_communication_legacy_firmware.cpp
```diff
@@ -100,6 +100,8 @@ void RemoteCommunicationLegacyFirmware::read_non_mmio(
     uint64_t core_src,
     uint32_t size_in_bytes,
     const std::chrono::milliseconds timeout_ms) {
+    auto lock = lock_manager_.acquire_mutex(MutexType::NON_MMIO, local_tt_device_->get_communication_device_id());
```
pjanevskiTT: do we need to move this lock this early in the function? Seems to me that code between this place and the one where this was before is safe to execute in multithreaded environment

device/tt_device/remote_communication_legacy_firmware.cpp
```diff
@@ -342,6 +339,7 @@ void RemoteCommunicationLegacyFirmware::write_to_non_mmio(
     bool broadcast,
     std::vector<int> broadcast_header,
     const std::chrono::milliseconds timeout_ms) {
+    auto lock = lock_manager_.acquire_mutex(MutexType::NON_MMIO, local_tt_device_->get_communication_device_id());
```
pjanevskiTT: same as above, apart from flush_non_mmio, can we maybe move other code to execute before the lock?

PR #1653, opened 2025-11-25
Version bump for changes to support TLBs from KMD

broskoTT:
Version bump after https://github.com/tenstorrent/tt-umd/commit/364e45c163247c940694a86f2c88b40e58bc4a6e


CHANGELOG
```diff
@@ -1,5 +1,12 @@
 # Changelog
```
pjanevskiTT: please change commit message from PR "Version bump" to something bit more verbose like "Version bump for changes to support TLBs from KMD"

broskoTT:
> although it might need bump from 0 to 1 for major

Still hesitant to make that jump over 1.0.0
I want for some discussions to happen and for us to have a clear version bumping guidance before that happens.
Also, our releases are still marked as "pre-release"

PR #1655, opened 2025-11-25
Set expected ETH FW version according to FW bundle version

aleksamarkovicTT:
### Issue
#1600 #1008

### Description
Instead of taking the ETH FW version from the first found core, use a mapping of expected ETH FW versions in FW bundles and match ETH FW version according to FW bundle version.

### List of the changes
- Introduce `get_expected_erisc_fw_version_from_fw_bundle()` that finds most recent ETH FW version for given FW bundle version
- Change `first_eth_fw_version` to `expected_eth_fw_version` in `TopologyDiscovery`
- Move ETH FW related constants to `erisc_firmware.hpp`

### Testing
CI

### API Changes

### DIFF
--- a/device/firmware/firmware_utils.cpp
+++ b/device/firmware/firmware_utils.cpp
@@ -6,9 +6,11 @@
 #include "umd/device/firmware/firmware_utils.hpp"
 
 #include <cstdint>
+#include <iterator>
 #include <optional>
 #include <thread>
 #include <tt-logger/tt-logger.hpp>
+#include <utility>
 
 #include "umd/device/arc/smbus_arc_telemetry_reader.hpp"
 #include "umd/device/firmware/erisc_firmware.hpp"
@@ -70,12 +72,16 @@ std::optional<semver_t> get_expected_eth_firmware_version_from_firmware_bundle(
 
     // Find the most recently updated ERISC FW version from a given firmware
     // bundle version.
-    for (const auto& [fw_bundle_it, eth_fw_version] : *version_map) {
-        if (semver_t::compare_firmware_bundle(fw_bundle_it, fw_bundle_version) >= 0) {
-            return eth_fw_version;
+    for (auto it = version_map->cbegin(); it != version_map->cend(); ++it) {
+        if (semver_t::compare_firmware_bundle(it->first, fw_bundle_version) > 0) {
+            if (it == version_map->cbegin()) {
+                return std::nullopt;
+            } else {
+                return std::prev(it)->second;
+            }
         }
     }
-    return std::nullopt;
+    return version_map->back().second;
 }
 
 semver_t get_eth_fw_version_from_telemetry(const uint32_t telemetry_data, tt::ARCH arch) {
--- a/tests/misc/test_semver.cpp
+++ b/tests/misc/test_semver.cpp
@@ -8,6 +8,9 @@
 
 #include <map>
 
+#include "umd/device/firmware/erisc_firmware.hpp"
+#include "umd/device/firmware/firmware_utils.hpp"
+#include "umd/device/types/arch.hpp"
 #include "umd/device/utils/semver.hpp"
 
 using namespace tt::umd;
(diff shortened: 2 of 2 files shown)


device/firmware/firmware_utils.cpp
```diff
@@ -50,6 +54,39 @@ semver_t get_firmware_version_util(TTDevice* tt_device) {
                : semver_t(0, 0, 0);
 }
 
+std::optional<semver_t> get_expected_eth_firmware_version_from_firmware_bundle(
+    semver_t fw_bundle_version, tt::ARCH arch) {
+    const auto* version_map = &erisc_firmware::WH_ERISC_FW_VERSION_MAP;
```
broskoTT: Can we make these constants a std::map<ARCH, whatever_is_now>, so you don't need this switch?

device/firmware/firmware_utils.cpp
```diff
@@ -50,6 +54,39 @@ semver_t get_firmware_version_util(TTDevice* tt_device) {
...
+        case ARCH::BLACKHOLE:
+            version_map = &erisc_firmware::BH_ERISC_FW_VERSION_MAP;
+            break;
+        default:
+            return std::nullopt;
+    }
+
+    // Find the most recently updated ERISC FW version from a given firmware
+    // bundle version.
+    auto it = std::upper_bound(
```
broskoTT: Can we have a set instead of vector? Specifically ordered container, that would allow us to trivially find the item if it is present or the closest one? If "closest one" is not a requirement, than unordered_set as well.

If you implement both suggestions , then get_expected_eth_firmware_version_from_firmware_bundle turns into a oneline function, so maybe it is not even needed

device/topology/topology_discovery_wormhole.cpp
```diff
@@ -320,17 +324,23 @@ bool TopologyDiscoveryWormhole::verify_eth_core_fw_version(Chip* chip, CoreCoord
     semver_t eth_fw_version = semver_t::from_eth_fw_tag(eth_fw_version_read);
 
     bool eth_fw_problem = false;
-    if (!first_eth_fw_version.has_value()) {
-        log_info(LogUMD, "Established ETH FW version: {}", eth_fw_version.to_string());
-        log_debug(LogUMD, "UMD supported minimum WH ETH FW version: {}", ERISC_FW_SUPPORTED_VERSION_MIN.to_string());
-        first_eth_fw_version = eth_fw_version;
-        if (ERISC_FW_SUPPORTED_VERSION_MIN > eth_fw_version) {
+    if (!expected_eth_fw_version.has_value()) {
```
broskoTT: The WH and BH functions look very similar, can we maybe have one in the base class?

device/topology/topology_discovery_blackhole.cpp
```diff
@@ -301,17 +304,23 @@ bool TopologyDiscoveryBlackhole::verify_eth_core_fw_version(Chip* chip, CoreCoor
...
+        expected_eth_fw_version =
+            get_expected_eth_firmware_version_from_firmware_bundle(first_fw_bundle_version.value(), ARCH::BLACKHOLE);
+        if (expected_eth_fw_version.has_value()) {
+            log_debug(LogUMD, "Expected ETH FW version: {}", expected_eth_fw_version->to_string());
+        } else {
+            expected_eth_fw_version = eth_fw_version;
+            log_debug(
+                LogUMD, "Established ETH FW version from first discovered ETH core: {}", eth_fw_version.to_string());
+        }
+        if (erisc_firmware::BH_ERISC_FW_SUPPORTED_VERSION_MIN > eth_fw_version) {
```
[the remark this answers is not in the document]
aleksamarkovicTT: Yes, it's the minimum supported version.

device/topology/topology_discovery_wormhole.cpp
aleksamarkovicTT: I will do this in a separate PR, given the trouble I had with Lite Fabric it's useful to keep them separate to revert code quickly.

device/firmware/firmware_utils.cpp
aleksamarkovicTT: I prefer the switch because it is certainly faster than using a map or unordered_map for only two entries.

device/topology/topology_discovery_wormhole.cpp
broskoTT: Sure, thanks

device/firmware/firmware_utils.cpp
broskoTT: The perf benefits are minimal, and are also not that important here on the initialization path. As mentioned in another comment, I value more code simplicity at places like this.

I'm fine with whatever you choose looks better in code

Issue #1340, opened 2025-09-18
Make static version of UMD device library


tt-vjovanovic:
**Timeline**
Metal folks are complaining that inside their environment, tt-exalens doesn't work.

**Is your feature request related to a problem? Please describe.**
Problem is that they override LD_LIBRARY_PATH variable and since both metal and tt-exalens depend on libdevice.so they try to load the same file, but they are depending on different versions.

**Proposed solution**
Have a static library version to avoid loading wrong device library.
You can take a look at https://github.com/tenstorrent/tt-umd/pull/1294 as a starting point.
It shouldn't be done like that. We should allow users to have both versions and to choose which one they would like to take. Follow some other open source library as an example.

PR #1662, opened 2025-11-26
Introduce static lib

broskoTT:
### Issue
#1340 

### Description
Create a version of static library. Also use it for nanobind builds, so that we end up with one .so

### List of the changes
- Introduce an option to build umd as static lib
- Different way that nanobind is installed based on this, whether to include libdevice.so
- Use that in pyproject.toml as default for python builds
- Failure for install command in case of static lib
- Failure to run cpack in case of static lib

### Testing
So I tested this more diligently after a discussion iteration with Nenad.
- `rm -rf build; umdc -DTT_UMD_BUILD_PIP=ON -DTT_UMD_BUILD_STATIC=OFF; umdb; cmake --install build --prefix installtest; rm -rf installtest; cd build; cpack -G DEB; cd ..`

cmake install passed
cpack installed umd-runtime and umd-dev

- `rm -rf build; umdc -DTT_UMD_BUILD_PIP=OFF -DTT_UMD_BUILD_STATIC=OFF; umdb; cmake --install build --prefix installtest; rm -rf installtest; cd build; cpack -G DEB; cd ..`

cmake install passed
cpack installed umd-runtime and umd-dev and umd-python

- `rm -rf build; umdc -DTT_UMD_BUILD_PIP=ON -DTT_UMD_BUILD_STATIC=ON; umdb; cmake --install build --prefix installtest; rm -rf installtest; cd build; cpack -G DEB; cd ..`

cmake install fails early with:
```
CMake Error at build/cmake_install.cmake:46 (message):
  Installation not supported when TT_UMD_BUILD_STATIC=ON.  Static libraries
  are meant for internal linking only, not for installation.
```

cpack fails showing the same error, and additionally:
`CPack Error: Error when generating package: umd`

- `rm -rf build; umdc -DTT_UMD_BUILD_PIP=OFF -DTT_UMD_BUILD_STATIC=ON; umdb; cmake --install build --prefix installtest; rm -rf installtest; cd build; cpack -G DEB; cd ..`

Same as previous

### API Changes

### DIFF
--- a/cmake/packaging.cmake
+++ b/cmake/packaging.cmake
@@ -91,4 +91,7 @@ cpack_add_component(
         umd-runtime # Makes the python package depend on the runtime
 )
 
-include(CPack)
+# Include CPack configuration only if not building as a static library
+if(NOT TT_UMD_BUILD_STATIC)
+    include(CPack)
+endif()


nanobind/CMakeLists.txt
```diff
@@ -107,8 +107,11 @@ if(TT_UMD_BUILD_PIP)
...
     # Only when building Python bindings to avoid conflicts with main install
-    if(TT_UMD_BUILD_PYTHON)
+    # Skip for static libraries as they are linked statically into the Python module
+    if(TT_UMD_BUILD_PYTHON AND NOT TT_UMD_BUILD_STATIC)
         install(FILES $<TARGET_FILE:umd::device> DESTINATION . COMPONENT pip_wheel)
+    elseif(TT_UMD_BUILD_PYTHON AND TT_UMD_BUILD_STATIC)
+        message(STATUS "Skipping device library install for pip wheel: device is built as static library")
     endif()
 else()
     # --- FHS/CPACK LOGIC ---
```
nbuncicTT: On line 127, there's an if/else block which defines the path for Fedora/Ubuntu. This logic is only valid for .so, hence it needs to go in some if/else branch dependent on `TT_UMD_BUILD_STATIC`. Something like this:
```bash
    if(TT_UMD_BUILD_STATIC)
        # STATIC MODE: The module is standalone (fat binary).
        # We do NOT want an RPATH because there is no external dependency.
        set_target_properties(
            nanobind_tt_umd
            PROPERTIES
                INSTALL_RPATH ""
        )
    else()
        # SHARED MODE: The module needs to find libdevice.so.
        
        # Fedora/RHEL (lib64) rpmbuild forbids relative RUNPATH like 'lib64'.
        # Keep Ubuntu's previous behavior; clear RUNPATH only on lib64 platforms.
        if(CMAKE_INSTALL_LIBDIR MATCHES "lib64")
            set_target_properties(
                nanobind_tt_umd
                PROPERTIES
                    INSTALL_RPATH ""
            )
        else()
            set_target_properties(
                nanobind_tt_umd
                PROPERTIES
                    INSTALL_RPATH "${CMAKE_INSTALL_LIBDIR}"
            )
        endif()
    endif()
```

I haven't tested either of these two things in CMake, hence it's necessary to test them by generating .deb and .rpm packages for python and running them to see if they work.

PR #1667, opened 2025-11-28
Implement SysmemManager for simulation

pjanevskiTT:
### Issue

/

### Description

Implement SysmemManager for simulation chips. The simulation doesn't have the actual system memory, we can't actually allocate it, so we need to simulate it by simply creating a vector of the size that is equal to number of host memory channels. Functions in Chip class need to overridden in order to fully support integration of SysmemManager inside SimulationChip.

### List of the changes

- Implement SimulationSysmemManager class
- Integrate it inside SimulationChip
- Override functions in SimulationChip that are needed to support system memory in simulation
- Add tests for SimulationSysmemManager

### Testing
Added new test. E2E tests to be run by Matt

### API Changes
/

### DIFF
--- a/device/api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
+++ b/device/api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
@@ -24,6 +24,9 @@ public:
         void* buffer, size_t sysmem_buffer_size, const bool map_to_noc = false) override;
 
 protected:
+    bool init_sysmem(uint32_t num_host_mem_channels) override;
+
+private:
     std::vector<uint8_t> system_memory_;
 };
 
--- a/device/chip_helpers/silicon_sysmem_manager.cpp
+++ b/device/chip_helpers/silicon_sysmem_manager.cpp
@@ -19,15 +19,7 @@
 
 namespace tt::umd {
 
-SiliconSysmemManager::SiliconSysmemManager(TLBManager *tlb_manager, uint32_t num_host_mem_channels) :
-    SysmemManager()
-// tlb_manager_(tlb_manager),
-// tt_device_(tlb_manager_->get_tt_device()),
-// pcie_base_(
-//     tlb_manager->get_tt_device()->get_arch() == tt::ARCH::WORMHOLE_B0
-//         ? 0x800000000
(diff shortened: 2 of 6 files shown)


device/api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
```diff
@@ -0,0 +1,59 @@
...
+     */
+    bool init_hugepages(uint32_t num_host_mem_channels) override;
+    /**
+     * Allocate sysmem without hugepages and map it through IOMMU.
+     * This is used when the system is protected by an IOMMU.  The mappings will
+     * still appear as hugepages to the caller.
+     * @param num_fake_mem_channels number of fake mem channels to allocate
+     * @return whether allocation/mapping succeeded.
+     */
+    bool init_iommu(uint32_t num_fake_mem_channels) override;
```
broskoTT: It seems like you don't need these functions? You only implemented init_hugepages from the protected section, others aren't called anywhere (since you implemented default constructor in SysmemManager

pjanevskiTT: right, removed all protected functions in sim sysmem manager

device/chip_helpers/simulation_sysmem_manager.cpp
```diff
@@ -0,0 +1,119 @@
...
+            num_host_mem_channels);
+    }
+
+    uint64_t total_size = num_host_mem_channels * (1ULL << 30);
+
+    if (num_host_mem_channels == 4) {
+        total_size -= 256 * (1ULL << 20);
+    }
+
+    system_memory_.resize(total_size, 0);
```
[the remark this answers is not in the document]
pjanevskiTT: I agree but there is some specific code in silicon case that is bit more difficult to merge. I would like to keep this as it is right now and change it bit later on

device/chip_helpers/silicon_sysmem_manager.cpp
```diff
@@ -0,0 +1,354 @@
+/*
```
broskoTT: I'm going to assume this is fully copied

pjanevskiTT: it is

PR #1668, opened 2025-11-28
Enable Asan and Tsan in our CI

broskoTT:
### Description
Set asan builds by default on PRs, and add both asan and tsan builds on main.

### List of the changes
- Change Release to Asan for PR runs
- Change Release to Asan, Tsan for main runs
- A minor fix after #1649 

After reviving these changes since November, I had to make a couple of more fixes. Yes, this means that all of these issues would've happened if this was turned on, so can serve as examples. Most of the issues were successfully fixed by claude, since the tsan/asan output can be a bit overwhelming:
- Testing for init result of TTDevice inside LocalChip and RemoteChip
- Remove __tsan_acquire, apparently this is not needed anymore?
- get_tsan_mutex_id const arg
- get_tsan_mutex_id lock tsan_storage_mutex
- Add io_mutex for weak_io in warm reset notifications
- Explicitly clear allocated_tlbs in tlb_virus test

### Testing
TBD for brosko, report perf: 
| type | current | asan (on PR) | tsan (on main) |
|--------|----------|-----|-----|
| build | 2m 29s | 3m 22s | 3m 6s |
| wheel build | 3m 49s | / | / |
| n150 | 4m 28s | 5m 1s | 5m 42s |
| n300 | 8m 40s | 9m 57s | 13m 49s |
| llmbox wh | 5m 22s | 7m 15s | 7m 31s |
| 6u wh | 15m 25s | 19m 10s | 21m 22s |
| p150 | 7m 34s | 8m 49s | 9m 43s |
| p300 | 8m 23s |  | 9m 46s |
| llmbox bh | 13m 2s | 14m 46s | 16m 7s |

Checks on this PR run Asan build
Here is a Tsan run on main: https://github.com/tenstorrent/tt-umd/actions/runs/19770130989
New run (March): https://github.com/tenstorrent/tt-umd/actions/runs/23608749856

### API Changes
There are no API changes in this PR.

PR #1672, opened 2025-11-28
Fix BH p300 machine runs

broskoTT:
### Issue
Blackhole dedicated machine gets broken after asan tsan runs from https://github.com/tenstorrent/tt-umd/pull/1668

### Description
The problem is that the cleanup of the folders was not done properly. Artefacts from the previous runs were left, so the python tests ended up using Asan builds.
It is actually the same problem that #1652 tried to resolve.
And I also managed to confirm that #1662 will solve it as well, but nevertheless, clean slate is better anyway.

### List of the changes
- rm -rf everything in workspace folder before running tests

### Testing
CI passes on this run, and subsequent scheduled runs on other PRs also pass after this fix.

### API Changes

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -72,6 +72,7 @@ jobs:
         # This might include: .pytest_cache, .venv, artifact.tar, build, tests, tt-umd, tt_umd-*.whl
         # Just nuke everything to be safe.
         run: |
+          echo "Cleaning up working directory from previous runs: $(pwd)"
           rm -rf *
 
       - name: Use wheel artifact


.github/workflows/build-and-run-all-tests.yml
```diff
@@ -71,20 +71,20 @@ jobs:
...
-          {arch: blackhole, card: tt-ubuntu-2204-p100a-viommu-stable},
-          {arch: blackhole, card: tt-ubuntu-2204-p150b-stable},
-          {arch: blackhole, card: tt-ubuntu-2204-p150b-viommu-stable},
+          # {arch: blackhole, card: tt-ubuntu-2204-p100a-viommu-stable},
+          # {arch: blackhole, card: tt-ubuntu-2204-p150b-stable},
+          # {arch: blackhole, card: tt-ubuntu-2204-p150b-viommu-stable},
           # bh-gh-07 is a P300 runner.
           {arch: blackhole, card: bh-gh-07},
-          {arch: baremetal, card: ubuntu-22.04},
+          # {arch: baremetal, card: ubuntu-22.04},
```
broskoTT: Will revert this

.github/workflows/run-tests.yml
```diff
@@ -67,17 +67,12 @@ jobs:
           echo build-tests - build-type: ${{ inputs.build-type }}
           echo build-tests - log-level: ${{ inputs.log-level }}
 
-      - name: Cleanup tt-umd dir, and change directory as if we were in a github.repository
+      - name: Cleanup workspace directory from previous runs.
+        # If this is not a new VM, we might have some artefacts from previous runs that need to be cleaned up.
+        # This might include: .pytest_cache, .venv, artifact.tar, build, tests, tt-umd, tt_umd-*.whl
+        # Just nuke everything to be safe.
         run: |
```
nbuncicTT: What's the directory your running in to use this command? Can you print it prior to calling this command to be sure it's ok to invoke this?

nbuncicTT (approved):

.github/workflows/run-tests.yml
broskoTT: It is workspace folder. Added pwd command

broskoTT: Confirmed from the runs it works fine

PR #1682, opened 2025-12-02
Read logical remote eth id for all BH systems

tt-asaigal:
### Issue
No Ticket.

### Description
- `get_logical_remote_eth_channel` is currently guarded for use on P150 systems only
- This is because the original implementation had a SW workaround that was applicable to P150. The SW workaround is not needed after TT-Flash 18.12.0, so this API can be used on all BH systems
- As part of Multi-Host bringup on BH Galaxy systems, we need to use this API on these systems as well.
- This PR uses this API as part of topology discovery for all multi-host BH systems

### List of the changes
- Unconditionally call `get_logical_remote_eth_channel` as part of topology discovery
- For BH this requires FW Bundle >= 18.12.0. WH behaviour is unchanged
- This does not affect any existing BH workloads, since they're all single process/host

### Testing
- Testing changes by upstreaming to metal

### API Changes

### DIFF
--- a/device/topology/topology_discovery_blackhole.cpp
+++ b/device/topology/topology_discovery_blackhole.cpp
@@ -356,7 +356,6 @@ uint64_t TopologyDiscoveryBlackhole::get_unconnected_chip_id(Chip* chip) {
     return (static_cast<uint64_t>(asic_id_hi) << 32) | asic_id_lo;
 }
 
-void TopologyDiscoveryBlackhole::validate_routing_firmware_state(
-    const std::map<uint64_t, std::unique_ptr<Chip>>& chips) {}
+bool TopologyDiscoveryBlackhole::verify_routing_firmware_state(Chip* chip, const tt_xy_pair eth_core) { return true; }
 
 }  // namespace tt::umd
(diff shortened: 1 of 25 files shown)


device/topology/topology_discovery_blackhole.cpp
```diff
@@ -163,26 +163,18 @@ uint32_t TopologyDiscoveryBlackhole::get_remote_eth_channel(Chip* chip, tt_xy_pa
 }
 
 uint32_t TopologyDiscoveryBlackhole::get_logical_remote_eth_channel(Chip* chip, tt_xy_pair local_eth_core) {
-    if (chip->get_chip_info().board_type != BoardType::P150) {
+    auto fw_bundle_version = chip->get_tt_device()->get_firmware_version();
+    if (fw_bundle_version < semver_t(18, 12, 0)) {
```
pjanevskiTT: hm I think this is going to break anything that is running FW older than 18.12. Since we are calling this function now for all systems, UMD would just throw for FW older than 18.12. We don't quite want. We are trying to keep larger set of FWs supported, so I think we need a way to handle FW older than 18.12 here as well. I would suggest something like
```
if (fw_bundle_version < semver_t(18, 12, 0)) {
    return get_remote_eth_channel(chip, local_eth_core);
}
... // same code as yours.
```
Does this make sense?

Issue #1680, opened 2025-12-01
Delete/Disable Lite Fabric


nhuang-tt:
We don't need Lite Fabric anymore. Maintaining it is more work and we already deleted it on tt-metal.

PR #1683, opened 2025-12-03
Remove all lite fabric related code

pjanevskiTT:
### Issue

#1680 

### Description

Remove all lite fabric related code. Blackhole will have all chips visible over PCIe, and since Wormhole already has a solution for remote communication we won't port it over.

### List of the changes

- Remove all lite fabric related code

### Testing
CI

### API Changes
/


device/api/umd/device/lite_fabric/fabric_edm_types.hpp
broskoTT: Are you sure about this? We wouldn't have any kind of way to make a remote communication to the other blackhole chip? I'd like to sync again with Davor and folks, so that they don't need to reimplement all the discovery protocols for remote chips if needed.

broskoTT: Also, is the base firmware that is left over enough to do link discovery between two ethernet connected chips? Aleksa mentioned something about having to communicate with ARC to get the noc translation/harvesting so we know which core we are actually connected to?

Issue #1687, opened 2025-12-04
UMD assumes local board ID from corrupt core


aleksamarkovicTT:
UMD assumes the local board ID from the first ETH core that reports it, even though this core might be corrupt. We need to verify ETH FW before obtaining local board ID.

PR #1691, opened 2025-12-05
Calculate WH-ERISC Board ID instead of reading it

aleksamarkovicTT:
### Issue
#1687 

### Description
Fix a bug where a corrupt WH-ERISC board ID causes neighboring WH chip to be reported as remote.

### List of the changes
- Calculate mangled board ID instead of reading it from potentially corrupt ETH core.

### Testing
CI

### API Changes


broskoTT (commented):
LGTM

Issue #1061, opened 2025-07-08
Show stack trace on exceptions


broskoTT:
Check the behavior for throw std::runtime_exception
Check the behavior for TT_THROW
Check the behavior for throwing in tt_metal

Somewhere in metal I've seen stack traces, let's figure out how they did it and replicate it.

aleksamarkovicTT:
This is now complete. Use `UMD_THROW` and `UMD_ASSERT` to have automatic stack traces printed.

PR #1694, opened 2025-12-08
Replace throw std::runtime_error() with TT_THROW

aleksamarkovicTT:
### Issue
#1061 

### Description
- Replaces all instances of throwing `std::runtime_error()` with `TT_THROW` which automatically prints stack trace.

### List of the changes
- Replaced all instances of `std::runtime_error()` outside of tests.
- Cleaned up some exception messages.
- Added `common` into include dirs for `device`

### Testing
CI

### API Changes

### DIFF
--- a/device/api/umd/device/arch/blackhole_implementation.hpp
+++ b/device/api/umd/device/arch/blackhole_implementation.hpp
@@ -343,7 +343,9 @@ public:
 
     uint32_t get_arc_message_test() const override { return static_cast<uint32_t>(blackhole::arc_message_type::TEST); }
 
-    uint32_t get_arc_csm_bar0_mailbox_offset() const override { UMD_THROW("Not implemented for Blackhole arch"); }
+    uint32_t get_arc_csm_bar0_mailbox_offset() const override {
+        UMD_THROW("Not implemented for Blackhole architecture.");
+    }
 
     uint32_t get_arc_axi_apb_peripheral_offset() const override { return blackhole::ARC_APB_BAR0_XBAR_OFFSET_START; }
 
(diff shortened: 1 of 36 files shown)


device/api/umd/device/arch/blackhole_implementation.hpp
```diff
@@ -376,20 +375,11 @@ class blackhole_implementation : public architecture_implementation {
 
     uint32_t get_dynamic_tlb_2m_size() const override { return blackhole::DYNAMIC_TLB_2M_SIZE; }
 
-    uint32_t get_dynamic_tlb_16m_base() const override {
-        throw std::runtime_error("No 16MB TLBs for Blackhole arch");
-        return 0;
-    }
+    uint32_t get_dynamic_tlb_16m_base() const override { TT_THROW("No 16MB TLBs for Blackhole arch"); }
```
pjanevskiTT: same as above. Pls revise all this for all other messages

device/api/umd/device/utils/assert.hpp
```diff
@@ -80,36 +80,35 @@ void tt_assert_message(std::ostream& os, T const& t, Ts const&... ts) {
...
     }
 }
 
-}  // namespace tt::assert
+}  // namespace tt::umd::assert
 
-#define TT_ASSERT(condition, ...) \
-    ::tt::assert::tt_assert(__FILE__, __LINE__, "TT_ASSERT", (condition), #condition, ##__VA_ARGS__)
-#define TT_THROW(...) ::tt::assert::tt_throw(__FILE__, __LINE__, "TT_THROW", "tt::exception", ##__VA_ARGS__)
+#define UMD_ASSERT(condition, ...) \
```
broskoTT: Why not TT_THROW though? Is it to avoid potential confusion with the one from tt_metal for example? This code is not specific to UMD, right?

PR #1696, opened 2025-12-09
Use TTDevice in TopologyDiscovery

aleksamarkovicTT:
### Issue
#1548 

### Description
This PR removes all usages of Chip classes in TopologyDiscovery, relying only on initialized TTDevices.
It is noticeable that SocDescriptor should be moved to TTDevice, because of extensive its extensive use for coordinate system operations.

### List of the changes
- Use TTDevice instead of Chip in all TopologyDiscovery code.
- Rename members referencing chips to reference devices.
- Move `Chip::translate_chip_coord_to_translated()` to SocDescriptor.
- Remove duplicate EthAddresses code from WormholeTTDevice.
- Fixes BH ERISC FW update to 1.7.0 from CMFW 18.12.1 instead of 18.12.0

### Testing
CI

### API Changes
This PR has API changes.

### DIFF
--- a/device/api/umd/device/chip_helpers/tlb_manager.hpp
+++ b/device/api/umd/device/chip_helpers/tlb_manager.hpp
@@ -23,7 +23,6 @@ public:
 
     // All tt_xy_pairs should be in TRANSLATED coords.
     void configure_tlb(tt_xy_pair core, size_t tlb_size, uint64_t address, uint64_t ordering);
-    void configure_tlb_kmd(tt_xy_pair core, size_t tlb_size, uint64_t address, uint64_t ordering);
     bool is_tlb_mapped(tt_xy_pair core);
     bool is_tlb_mapped(tt_xy_pair core, uint64_t address, uint32_t size_in_bytes);
 
--- a/device/arc/blackhole_arc_message_queue.cpp
+++ b/device/arc/blackhole_arc_message_queue.cpp
@@ -43,8 +43,7 @@ void BlackholeArcMessageQueue::push_request(
             break;
         }
 
-        auto now = std::chrono::steady_clock::now();
-        utils::check_timeout(now, timeout_ms, "Timeout waiting for ARC msg request queue.");
+        utils::check_timeout(start, timeout_ms, "Timeout waiting for ARC msg request queue.");
     }
 
     // Offset in words.
@@ -69,8 +68,7 @@ std::array<uint32_t, BlackholeArcMessageQueue::entry_len> BlackholeArcMessageQue
             break;
         }
 
-        auto now = std::chrono::steady_clock::now();
-        utils::check_timeout(now, timeout_ms, "Timeout waiting for ARC msg request queue.");
+        utils::check_timeout(start, timeout_ms, "Timeout waiting for ARC msg request queue.");
     }
 
     uint32_t response_entry_offset =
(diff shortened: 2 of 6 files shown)


device/topology/topology_discovery.cpp
```diff
@@ -69,26 +67,26 @@ TopologyDiscovery::TopologyDiscovery(const TopologyDiscoveryOptions& options) :
...
     std::unique_ptr<ClusterDescriptor> cluster_desc = td->create_ethernet_map();
-    return std::make_pair(std::move(cluster_desc), std::move(td->chips));
+    return std::make_pair(std::move(cluster_desc), std::move(td->devices));
 }
 
 void TopologyDiscovery::init_topology_discovery() {}
 
-void TopologyDiscovery::get_connected_chips() {
+void TopologyDiscovery::get_connected_devices() {
     std::vector<int> device_ids;
```
nbuncicTT: nit: Rename this `local_device_ids` or `pcie_device_ids`

device/topology/topology_discovery.cpp
```diff
@@ -130,92 +129,92 @@ void TopologyDiscovery::get_connected_chips() {
     }
 }
 
-void TopologyDiscovery::discover_remote_chips() {
+void TopologyDiscovery::discover_remote_devices() {
     std::set<uint64_t> discovered_chips = {};
```
nbuncicTT: nit: Rename this to `discovered_devices`

device/api/umd/device/topology/topology_discovery.hpp
```diff
@@ -140,26 +143,30 @@ class TopologyDiscovery {
...
 
-    // The FW bundle version found on the first discovered local chip, that needs
-    // to match with all of the other discovered FW bundle versions on all chips.
+    // The FW bundle version found on the first discovered local device, that needs
+    // to match with all of the other discovered FW bundle versions on all devices.
     std::optional<semver_t> first_fw_bundle_version;
+
+private:
+    // Hack used to cache SocDescriptors.
+    std::unordered_map<TTDevice*, SocDescriptor> soc_descriptor_cache;
```
pjanevskiTT: Nice, hope this saves us from having some kind of perf penalty in topology discovery. No action if perf of topology discovery is ok

device/chip/chip.cpp
```diff
@@ -243,15 +243,7 @@ void Chip::set_power_state(DevicePowerState state) {
 }
 
 tt_xy_pair Chip::translate_chip_coord_to_translated(const CoreCoord core) const {
-    // Since NOC1 and translated coordinate space overlaps for Tensix cores on Blackhole,
```
pjanevskiTT: why not remove this function completely now? Code in Chip class can also call into soc_descriptor method if needed, no need to have a wrapper that just calls into SocDescriptor

device/topology/topology_discovery.cpp
```diff
@@ -397,4 +397,26 @@ bool TopologyDiscovery::verify_fw_bundle_version(Chip* chip) {
     return true;
 }
 
+SocDescriptor TopologyDiscovery::get_soc_descriptor(TTDevice* tt_device) {
```
pjanevskiTT: as mentioned above as well, make sure this doesn't have any significant perf impact on TopologyDiscovery. I would assume we are fine with caching

device/api/umd/device/topology/topology_discovery.hpp
aleksamarkovicTT: I ran OpenCluster microbenchmarks on n150 and there seems to be no time difference, perf should be alright.

device/api/umd/device/soc_descriptor.hpp
```diff
@@ -90,6 +90,7 @@ class SocDescriptor {
         const tt_xy_pair core_location,
```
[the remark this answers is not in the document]
aleksamarkovicTT: I compared ClusterDescriptors from this PR and newest main and found them identical on n300, 6U and t3k

Issue #1689, opened 2025-12-05
UMD start-up fails on 6u galaxy


adjordjevic-TT:
**Describe the bug**
UMD start-up fails on 6u galaxy (`wh-glx6u-05`) with following exception `Routing Firmware should not be enabled on 6U-Galaxy Systems.`. 

**Stack trace**
[logs.txt](https://github.com/user-attachments/files/23961485/logs.txt)

**To reproduce**
Try starting `tt-exalens`.

**Expected behavior**
Start-up is successful.

PR #1697, opened 2025-12-09
Make routing firmware check not throw exception

aleksamarkovicTT:
### Issue
#1689 

### Description
TopologyDiscovery should not throw exceptions because of routing firmware state. Change to report warning and skip discovery from ETH core.

### List of the changes
- Verifying routing fw state now happens just before discovery.
- Verifying routing fw state no longer throws exception.

### Testing
CI

### API Changes

### DIFF
--- a/device/api/umd/device/firmware/erisc_firmware.hpp
+++ b/device/api/umd/device/firmware/erisc_firmware.hpp
@@ -33,7 +33,8 @@ static const std::vector<std::pair<semver_t, semver_t>> WH_ERISC_FW_VERSION_MAP 
     {{18, 4, 0}, {6, 15, 0}},
     {{18, 6, 0}, {7, 0, 0}},
     {{18, 12, 0}, {7, 1, 0}},
-    {{19, 0, 0}, {7, 2, 0}}};
+    {{19, 0, 0}, {7, 2, 0}},
+    {{19, 4, 0}, {7, 3, 0}}};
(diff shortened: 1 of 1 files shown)


broskoTT (approved):

aleksamarkovicTT:
> Hey @aleksamarkovicTT Im trying to understand the motivation behind this change. The exception was added because running SW on a 6U Galaxy with routing firmware enabled leads to undefined behaviour.
> 
> As such, it made sense for us to catch this as low in the stack as possible - hence, we decided to put this in UMD, as part of topology discovery over ethernet.
> 
> My worry is that a warning may not be enough, especially for machines running automated jobs. An exception immediately lets the user know whats wrong.

I completely agree with your remarks. The motivation is that UMD needs to support a "best-effort" discovery for tt-exalens. A coworker stumbled upon a 6U left in an invalid state (routing firmware was enabled on every 4th chip) after some firmware test suite was run on it - UMD would not start due to this check, and I was asked to make this change.

I will modify this PR to retain the exception, however make it a warning if there is a special TopologyDiscovery option set (TopologyDiscoveryOptions::no_eth_firmware_strictness). The option is off by default and only meant for the purposes of debugging/triage. I hope this clears up the reasoning behind the change.

tt-asaigal:
An option here to unblock other users makes sense. Thanks!

PR #1700, opened 2025-12-11
Driver version mismatch removal

broskoTT:
### Issue
Aftermath of https://github.com/tenstorrent/tt-umd/pull/1556

### Description
Seems like a check was leftover that effectivelly ties us to a single kmd version. And looks like all our CI machines are using the same KMD version.

### List of the changes
- Remove the checks for driver version

### Testing
No testing

### API Changes

### DIFF
--- a/device/tt_kmd_lib/tt_kmd_lib.c
+++ b/device/tt_kmd_lib/tt_kmd_lib.c
@@ -81,33 +81,6 @@ int tt_device_open(const char* chardev_path, tt_device_t** out_dev) {
         return -e;
     }
 
-    uint64_t major = 0;
-    uint64_t minor = 0;
-    uint64_t patch = 0;
-
-    int ret;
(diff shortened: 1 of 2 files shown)


device/tt_kmd_lib/tt_kmd_lib.c
```diff
@@ -81,33 +81,6 @@ int tt_device_open(const char* chardev_path, tt_device_t** out_dev) {
         return -e;
     }
 
-    uint64_t major = 0;
-    uint64_t minor = 0;
-    uint64_t patch = 0;
-
-    int ret;
-    if ((ret = tt_driver_get_attr(dev, TT_DRIVER_SEMVER_MAJOR, &major)) != 0 ||
```
aleksamarkovicTT: Here, the check gets removed entirely because it is too narrow.
The check should exist however and we should determine the range of versions kmd_lib can work with. CC: @joelsmithTT

PR #1715, opened 2025-12-23
Setup default static TLBs in UMD

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/44

### Description
Copy code for setting up static TLBs to UMD from metal.
This code is designed to work in the meantime as well, if both the client and UMD setup these, it should still work.
The next step would be to remove this code from tt-metal's codebase.

Mind that as part of this change, we have another subtle functional change: On close device, all mappings will be cleared. This should help significantly with the multiprocess issues with TLBs we've been seeing in the fleet in the last couple of weeks. Example: https://github.com/tenstorrent/tt-metal/issues/34034

### List of the changes
- Add function to cleanup a specific TLB, which is now called each time configure is called, to verify we don't have two TLBs setup over the same core.
- Add functions for setting up default static TLBs for all tensix, eth, and dram cores.
- Cleanup all mappings when close_device is called
- Added relevant log_debugs and log_info
- Removed unused get_static_tlb_size, and changed the get_cached_tbl_size to use constants which I also fixed

### Testing
Manually ran a metal tests which showed expected log outputs.
tt_metal runs: 
APC: https://github.com/tenstorrent/tt-metal/actions/runs/20462642669
BH: https://github.com/tenstorrent/tt-metal/actions/runs/20462654157
GX: https://github.com/tenstorrent/tt-metal/actions/runs/20462660845
T3K: https://github.com/tenstorrent/tt-metal/actions/runs/20462657817
T3K nightly: https://github.com/tenstorrent/tt-metal/actions/runs/20463137941

### API Changes

### DIFF
--- a/device/chip_helpers/tlb_manager.cpp
+++ b/device/chip_helpers/tlb_manager.cpp
@@ -137,6 +137,11 @@ std::unique_ptr<TlbWindow> TLBManager::allocate_tlb_window(
 }
 
 void TLBManager::map_default_static_tlbs(SocDescriptor& soc_descriptor) {
+    if (tlb_windows_.size() > 0) {
+        log_info(LogUMD, "Static TLBs already mapped. Skipping mapping default static TLBs.");
+        return;
+    }
+
     log_info(LogUMD, "Mapping default static TLBs.");
 
     uint32_t static_tlb_size = tt_device_->get_architecture_implementation()->get_cached_tlb_size();


device/chip/local_chip.cpp
```diff
@@ -189,6 +189,9 @@ void LocalChip::start_device() {
         init_pcie_iatus();
     }
     initialize_membars();
+    if (tlb_manager_) {
+        tlb_manager_->map_default_static_tlbs(soc_descriptor_);
```
pjanevskiTT: i am not entirely sure if this is the right thing to do. Few things I can think of from top of my head

- Metal used to setup TLBs this way since they had some perf requirements. What happens now when they want to change this scheme? Do we need to take care of it?
- I think we should provide the mechanism on how to allocate/map TLBs and then metal and other clients should setup how they like it. If they have some bugs in MetalContext stuff we shouldn't fix it by just putting it into UMD. 
- For example now with multicast support, they probably don't need all TLBs allocated for each Tensix. How do we communicate these changes and know when to change this map_default_static_tlbs function strategy on how we do it?

pjanevskiTT (commented):
I think code changes are good if we even want to do this. I am not convinced we should so left a comment to discuss

Issue #1719, opened 2025-12-24
'Firmware bundle version 18.5.0 on the system is older than the minimum compatible version 18.5.0 for blackhole architecture'


changh95:
**Describe the bug**
Thanks to @swote-git and @gnh1201 for finding the bug!

When trying to run any TT programs with tt-firmware 18.5.0 on BH devices, `Firmware bundle version 18.5.0 on the system is older than the minimum compatible version 18.5.0 for blackhole architecture` assert appears and stops the program.

**Stack trace**

<img width="1714" height="1303" alt="Image" src="https://github.com/user-attachments/assets/b467e6a7-89bd-48e1-bf2f-f295e48ec5ff" />

**To reproduce**
1. Set tt-firmware 18.5.0 on any BH devices
2. Run any TT programs
3. See bug behavior

**Expected behavior**
It should pass, since version 18.5.0 should actually work on BH devices.

PR #1720, opened 2025-12-24
fix/Asserting on correct Blackhole firmware version check (18.5.0)

changh95:
### Issue
https://github.com/tenstorrent/tt-umd/issues/1719

### Description
The minimum version for Blackhole devices is 18.5.0. 
When tt-firmware is set to 18.5.0, UMD will prevent running tt programs with assert with the following error:

'Firmware bundle version 18.5.0 on the system is older than the minimum compatible version 18.5.0 for blackhole architecture.'

When looking at the code that prints this assert message (`device/topology/topology_discovery.cpp`), we see that the assert condition prevents any FW that is not newer than 18.5.0, but the message implies that the the asset condition prevents any FW that is older than 18.5.0. 

This excludes the case of FW being 18.5.0, which is the SAME as the condition.

So we need to consider the case for 18.5.0 as well.

### List of the changes

- Add equality for tt-firmware 18.5.0.

### Testing

### API Changes

### DIFF
--- a/device/topology/topology_discovery.cpp
+++ b/device/topology/topology_discovery.cpp
@@ -378,7 +378,7 @@ bool TopologyDiscovery::verify_fw_bundle_version(Chip* chip) {
         latest_supported_fw_bundle_version.to_string());
 
     TT_ASSERT(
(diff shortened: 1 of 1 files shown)


aleksamarkovicTT (approved):
Thanks for noticing this bug.

PR #1722, opened 2025-12-24
Warm reset tool

broskoTT:
### Issue
No issue

### Description
Small tool for running warm reset. 
Can be convenient to quickly reset a device when tt-smi is not readily available.

### List of the changes
- Implemented warm_reset tool

### Testing
Ran it manually on a machine

### API Changes


tools/warm_reset.cpp
```diff
@@ -0,0 +1,44 @@
...
+    auto result = options.parse(argc, argv);
+
+    if (result.count("help")) {
+        std::cout << options.help() << std::endl;
+        return 0;
+    }
+
+    try {
+        log_info(tt::LogUMD, "Performing warm reset on all available devices...");
+        WarmReset::warm_reset();
```
aleksamarkovicTT: Maybe do a check for 6U and perform UBB Warm reset in that case?

PR #1726, opened 2025-12-25
Add benchmark to compare unicast with multicast

pjanevskiTT:
### Issue

/

### Description

This PR adds simple benchmark which compares time needed to multicast data to all tensix cores and time to unicast data to every single Tensix core. SInce multicast BW is the same as unicast, it just saves time by having multiple endpoints as the destination. Because of that, the test is added, disabled by default, so someone can run it if needed for info, but it's not part of the official benchmarking. 

### List of the changes

- Add test to compare unicast with multicast

### Testing
Manual testing

### API Changes
/

### DIFF
--- a/tests/microbenchmark/benchmarks/tlb/test_tlb.cpp
+++ b/tests/microbenchmark/benchmarks/tlb/test_tlb.cpp
@@ -304,7 +304,7 @@ TEST(MicrobenchmarkTLB, CompareMulticastandUnicast) {
 
         {
             double total_ns = 0;
-            for (auto tensix_core : tensix_cores) {
+            for (auto &tensix_core : tensix_cores) {
                 auto start = std::chrono::steady_clock::now();
                 cluster->write_to_device(buffer.data(), buf_size, chip, tensix_core, 0);
                 auto end = std::chrono::steady_clock::now();


tests/microbenchmark/benchmarks/tlb/test_tlb.cpp
```diff
@@ -266,3 +266,72 @@ TEST(MicrobenchmarkTLB, TLBStaticEth) {
     }
     test::utils::print_markdown_table_format(headers, rows);
 }
+
+// Since multicast has multiple endpoints as targets, it's not compeletely fair to compare
+// the bandwidth, which is still tied to TLB bandwidth. BW of multicast writes will be the same in terms
+// of BW as unicast writes. The benefit of multicast is in saving time by writing to multiple endpoints in one go.
+// However, it is interesting to see the time taken for unicast vs multicast writes to multiple endpoints.
+// That is why this test is disabled by default. It's meant for someone to run it manually if needed.
+TEST(MicrobenchmarkTLB, DISABLED_CompareMulticastandUnicast) {
```
broskoTT: Why Disabled?

nbuncicTT (approved):

aleksamarkovicTT:
I don't think these benchmarks should "compare speedup" they should just measure multicast performance.
Unicast performance is measured in existing benchmarks, so this is performing existing measurements and the data given here can be calculated at a later stage, and I am also not sure what information "speedup" brings. It will not show any change when there is change in both unicast and multicast performance, and regressions in unicast will show up as better perf for multicast.

PR #1727, opened 2025-12-25
Skip sysmem setup in tests for local only configurations

broskoTT:
### Description
While working on some 6U tests, I noticed they take really long as each test is setting up 32GB iommu buffers. Note this doesn't take that long in hugepage case though.
I've changed test to only setup system memory in tests that actually use it, and on configurations which have remote chips.
Note that this is potentially even a design issue for our tests. We don't have to just Cluster() in each test, we should separate across components.

### List of the changes
- Changed default sysmem channels to 0
- Alter tests to use sysmem 1 if needed
- Alter tests to use sysmem 1 on a system with remote chips if it would help perf
- Remove verification of at least one sysmem for remote chips
- Change the condition a bit in remote communication, allow large transfers even without sysmem
- Print warning in case of large transfers

### Testing
CI runs on this PR + manual run on 6u
n150 viommu - lowered approx from 2m45s to 2m15s
n300 viommu - lowered approx from 8m to 7m
n300 llmbox viommu - lowered approx from 9m to 7m
p150 viommu - from 4m to 2m30s
6u previously timed out, now runs in 10m

### API Changes

### DIFF
--- a/device/tt_device/remote_communication_legacy_firmware.cpp
+++ b/device/tt_device/remote_communication_legacy_firmware.cpp
@@ -145,8 +145,13 @@ void RemoteCommunicationLegacyFirmware::read_non_mmio(
     erisc_q_rptr.resize(1);
     erisc_q_rptr[0] = erisc_q_ptrs[4];
 
-    bool use_host_dram = size_in_bytes > 256 * DATA_WORD_SIZE && sysmem_manager_ != nullptr &&
-                         sysmem_manager_->get_num_host_mem_channels() > 0;
+    bool system_mem_available = sysmem_manager_ != nullptr && sysmem_manager_->get_num_host_mem_channels() > 0;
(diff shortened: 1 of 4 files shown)


tests/api/test_chip.cpp
```diff
@@ -24,7 +24,7 @@ using namespace tt::umd;
 
 // TODO: Once default auto TLB setup is in, check it is setup properly.
 TEST(ApiChipTest, DISABLED_ManualTLBConfiguration) {
-    std::unique_ptr<Cluster> umd_cluster = std::make_unique<Cluster>();
+    std::unique_ptr<Cluster> umd_cluster = get_default_cluster();
```
pjanevskiTT: should we maybe handle this by setting param for num_host_mem_channels to 0 for cluster and then someone constructing it should explicitly request sysmem? Maybe that change can go additionally with this one since sometimes we need sysmem in tests for N300

PR #1734, opened 2025-12-26
Consolidate start_device in tests

broskoTT:
### Issue
Should finally close https://github.com/tenstorrent/tt-umd/issues/422

### Description
We had issues with start_device historically, but in the meantime we got the infrastructure to safely start the cluster in tests by setting up no-op kernels.
I now went again through our tests, and removed start_device where not needed, and used the new infra where it wasn't.
Also finally implemented a multiprocess tests which starts/stops the cluster, which is now guarded by chip_lock.

### List of the changes
- Removed start_device where unnecessary.
- Where needed changed to safe_test_cluster_start
- Change the helper function to also start the cluster
- Implemented a multiprocess tests which starts/stops the cluster, which is now guarded by chip_lock

### Testing
CI should be enough

### API Changes

### DIFF
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -982,7 +982,7 @@ void Cluster::deassert_resets_and_set_power_state() {
 }
 
 void Cluster::start_device(const DeviceParams& device_params) {
-    log_info(LogUMD, "Starting cluster");
+    log_info(LogUMD, "Starting devices in cluster");
     if (device_params.init_device) {
         for (auto chip_id : all_chip_ids_) {
             get_chip(chip_id)->start_device();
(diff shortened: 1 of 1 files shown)


tests/unified/multiprocess.cpp
```diff
@@ -139,21 +139,16 @@ TEST(Multiprocess, MultipleThreadsMultipleClustersRunning) {
...
-        clusters.emplace_back(std::make_unique<Cluster>());
-        std::cout << "Setup risc cores for cluster " << i << std::endl;
-        test_utils::setup_risc_cores_on_cluster(clusters[i].get());
-    }
     for (int i = 0; i < NUM_PARALLEL; i++) {
         threads.push_back(std::thread([&, i] {
-            std::cout << "Starting cluster " << i << std::endl;
+            std::unique_ptr<Cluster> cluster = std::make_unique<Cluster>();
+            std::cout << "Setting up risc cores and starting cluster " << i << std::endl;
+            test_utils::safe_test_cluster_start(cluster.get());
```
nbuncicTT: I see that the start is now used with all these API's, but shouldn't we have a `stop_device()`/`close_device()` at the end to test that API? Or you think it should be a part of another task/PR?

nbuncicTT: So after a small analysis here are the two tests that use `test_utils::safe_test_cluster_start`, but don't call `close_device()` at the end:
- SysmemReadWrite
- MultipleThreadsMultipleClustersOpenClose

PR #1736, opened 2025-12-29
Add stub generation and installation for type checking in nanobind module

tt-vjovanovic:
tt-exalens needs stub file for mypy to work. This will also enable other static analyzers to work as well.
In order for static analyzers to pick up stub file, it needs to be part of the package. So, tt-umd is not just module, but now needs to be package. Hence directory structure looks like this:
```
tt_umd/
    __init__.py        # Auto-generated file. Imports everything from nanobind module
    __init__.pyi       # Stub file from top-level module
    logging.py         # Auto-generated file based on existence of tt_umd.logging submodule
    logging.pyi        # Stub file for tt_umd.logging
    wormhole.py        # Auto-generated file based on existence of tt_umd.wormhole submodule
    wormhole.pyi       # Stub file for tt_umd.wormhole
    tt_umd....so       # Nanobind module
    py.typed           # Says that package has stub file
```
nanobind stub generator generates stubs for top-level module and all submodules. `generate_init.py` script during compile time generates `__init__.py` and all `[submodule].py` files that do imports so that `tt_umd` package works as expected.
This allows bindings project to be named differently if needed (currently `__init__.py` imports everything from `.tt_umd`, but can be anything else)

Needed to update nanobind to latest version (to get stub generation for submodules), which requires python 3.9. This means that tt-umd doesn't support Python 3.8.

### DIFF
--- a/device/api/umd/device/tt_device/remote_wormhole_tt_device.hpp
+++ b/device/api/umd/device/tt_device/remote_wormhole_tt_device.hpp
@@ -28,7 +28,7 @@ public:
 
     void wait_for_non_mmio_flush() override;
 
-    RemoteCommunication* get_remote_communication();
+    RemoteCommunication* get_remote_communication() const override;
 
     /*
      * RemoteWormholeTTDevice uses RemoteCommunication and doesn't have an underlying I/O device,
--- a/device/api/umd/device/tt_device/tt_device.hpp
+++ b/device/api/umd/device/tt_device/tt_device.hpp
@@ -263,6 +263,8 @@ public:
 
     FirmwareInfoProvider *get_firmware_info_provider() const;
 
+    virtual RemoteCommunication *get_remote_communication() const { return nullptr; }
+
     virtual uint32_t get_clock() = 0;
 
     uint32_t get_max_clock_freq();
--- a/device/tt_device/remote_wormhole_tt_device.cpp
+++ b/device/tt_device/remote_wormhole_tt_device.cpp
@@ -37,7 +37,7 @@ void RemoteWormholeTTDevice::write_to_device(const void *mem_ptr, tt_xy_pair cor
 
 void RemoteWormholeTTDevice::wait_for_non_mmio_flush() { remote_communication_->wait_for_non_mmio_flush(); }
 
-RemoteCommunication *RemoteWormholeTTDevice::get_remote_communication() { return remote_communication_.get(); }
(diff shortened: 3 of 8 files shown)


pyproject.toml
```diff
@@ -37,11 +37,12 @@ classifiers = [
 cmake.build-type = "Release"
 build-dir = "build/"
 # Only build the Python-related targets (new key in scikit-build-core >= 0.10)
-build.targets = ["nanobind_tt_umd", "device"]
+build.targets = ["nanobind_tt_umd", "nanobind_tt_umd_stub", "device"]
 # Exclude development files from the wheel
 wheel.exclude = ["CMakeCache.txt", "*.cmake", "CMakeFiles/", "Makefile"]
 # Install only the 'pip_wheel' component defined in CMake (new key path)
 install.components = ["pip_wheel"]
+wheel.py-api = "py3"
```
nbuncicTT: Why do we need this line? 

From what I've looked around it says that this does the following:
> By setting `wheel.py-api = "py3"` it tells pip your C++ binary works on every version of Python 3, which will cause a crash if, for example, a user on Python 3.12 installs a wheel that you compiled specifically for Python 3.10.

We have support from python 3.8 to 3.13
image

And by adding this change, it would collapse all of these artifacts into one, like this:
`tt_umd-0.6.0-py3-none-manylinux_2_34_x86_64.whl`

which wouldn't work, so I'd omit this line from the PR.

I haven't tested any of this, I've researched a bit with LLM and with docs, so maybe I'm wrong, but that's my assumption based on the investigation I've done.

nanobind/CMakeLists.txt
```diff
@@ -103,13 +178,9 @@ if(TT_UMD_BUILD_PIP)
...
 else()
     # --- FHS/CPACK LOGIC ---
-    # Configures the build for a .deb/.rpm package.
-    # RPATH is set to the system lib dir to find the 'umd-runtime' package.
-
     message(STATUS "Configuring Python bindings for FHS package")
 
-    # Fedora/RHEL (lib64) rpmbuild forbids relative RUNPATH like 'lib64'.
-    # Keep Ubuntu's previous behavior; clear RUNPATH only on lib64 platforms.
+    # Set RPATH for system lib directory (clear for lib64 per rpmbuild requirements)
```
nbuncicTT: nit: Can you leave the previous comment? I think it nicely states that Fedora/RHEL explicitly need a blank `INSTALL_RPATH`, but Ubuntu and other Debian based distros need `CMAKE_INSTALL_LIBDIR`

PR #1742, opened 2025-12-31
Exposing RTL simulation to TTDevice API

tt-vjovanovic:
Adding missing pure virtual functions so that `RtlSimulationTTDevice` class can compile.
Exposing its functions to Python and adding simulation to python build.
Adding missing functionality that `tt-exalens` needs.
Bumping version to `0.7.2` so that `tt-exalens` can consume simulation as well.

### DIFF
--- a/device/tt_device/rtl_simulation_tt_device.cpp
+++ b/device/tt_device/rtl_simulation_tt_device.cpp
@@ -68,7 +68,7 @@ RtlSimulationTTDevice::RtlSimulationTTDevice(
     start_host_communication();
 }
 
-static inline flatbuffers::FlatBufferBuilder create_flatbuffer(
+static inline flatbuffers::FlatBufferBuilder _create_flatbuffer(
     DEVICE_COMMAND rw, std::vector<uint32_t> vec, tt_xy_pair core_, uint64_t addr, uint64_t size_ = 0) {
     flatbuffers::FlatBufferBuilder builder;
     auto data = builder.CreateVector(vec);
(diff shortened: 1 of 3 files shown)


device/api/umd/device/tt_device/rtl_simulation_tt_device.hpp
```diff
@@ -8,11 +8,46 @@
...
 
+#include "umd/device/simulation/simulation_host.hpp"
+#include "umd/device/soc_descriptor.hpp"
 #include "umd/device/tt_device/tt_device.hpp"
 
 namespace tt::umd {
 class RtlSimulationTTDevice : public TTDevice {
 public:
-    RtlSimulationTTDevice();
+    RtlSimulationTTDevice(const std::filesystem::path &simulator_directory, SocDescriptor soc_descriptor);
```
nbuncicTT: Does it make sense to make this constructor private as you'll only make objects via the create function?

device/tt_device/rtl_simulation_tt_device.cpp
```diff
@@ -4,11 +4,222 @@
...
+    child_stdio[2].flags = UV_INHERIT_FD;
+    child_stdio[2].data.fd = 2;
+
+    uv_process_options_t child_options = {0};
+    child_options.file = simulator_path_string.c_str();
+    child_options.flags = UV_PROCESS_DETACHED;
+    child_options.stdio_count = 3;
+    child_options.stdio = child_stdio;
+
+    uv_process_t child_p;
```
nbuncicTT: Does it make sense for this to be a local variable with automatic lifetime duration?

Should it's ownership be governed by the class itself (make it a member of the class), or allocate on the heap and give it to some other owner?

PR #1746, opened 2026-01-03
Moving C++ from tt-exalens

tt-vjovanovic:
Moving `c++.natvis` file from tt-exalens as we don't have C++ code any more, while here it might be useful (even for Metal folks as well).
Exposing compile commands by default, so you get code references when you compile code by installing clangd.


compile_commands.json
```diff
@@ -0,0 +1 @@
+build/compile_commands.json
```
nbuncicTT: Nice, now this can be seen by clangd without having to modify the `settings.json` file. For which tools did you add this? Asking out of curiosity, no action needed.

nbuncicTT (approved):

PR #1747, opened 2026-01-03
Exposing few more methods for tt-exalens

tt-vjovanovic:
Exposed methods:
- `ClusterDescriptor.create_from_yaml_content`
- `ClusterDescriptor.get_ethernet_connections`
- `ClusterDescriptor.get_io_device_type`
- `TTDevice.get_communication_device_type`

Closing simulator on `~RtlSimulationTTDevice()`.

Bumping VERSION to `0.7.3`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,7 @@ set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED True)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON) # This also impacts dependencies brought in through CPM
 set(CMAKE_POLICY_VERSION_MINIMUM 3.5)
-set(CMAKE_EXPORT_COMPILE_COMMANDS 1)
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON) # Generate compile_commands.json for IDEs and tools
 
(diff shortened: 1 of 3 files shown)


device/cluster_descriptor.cpp
```diff
@@ -322,16 +322,22 @@ ChipId ClusterDescriptor::get_closest_mmio_capable_chip(const ChipId chip) {
...
     }
+    std::stringstream buffer;
+    buffer << fdesc.rdbuf();
     fdesc.close();
+    return create_from_yaml_content(buffer.str());
+}
+
+std::unique_ptr<ClusterDescriptor> ClusterDescriptor::create_from_yaml_content(
+    const std::string &cluster_descriptor_file_content) {
+    std::unique_ptr<ClusterDescriptor> desc = std::unique_ptr<ClusterDescriptor>(new ClusterDescriptor());
```
nbuncicTT: I'm aware that in our `ClusterDescriptor` class we make the new ClusterDescriptor this way, but I'd prefer to make it via `make_unique` instead of this way as the constructor for this class is public.

Once we move (or if we ever move) the constructor to private we will revert to this way of making the class (for now I'll modify the ClusterDescriptor class to use make_unique in some other PR).

If you encounter a problem with `make_unique`, leave it this way.

PR #1748, opened 2026-01-04
Initial support for noc1

### LINKED ISSUE #526: Expose sending commands to noc1 and noc0

tt-vjovanovic:
`use_noc1` transferred through whole code base.
It defaults to `false` on public interfaces.

Closes #526.

### DIFF
--- a/nanobind/py_api_telemetry.cpp
+++ b/nanobind/py_api_telemetry.cpp
@@ -140,25 +140,25 @@ void bind_telemetry(nb::module_ &m) {
 
     nb::class_<FirmwareInfoProvider>(m, "FirmwareInfoProvider")
         .def("get_firmware_version", &FirmwareInfoProvider::get_firmware_version)
-        .def("get_board_id", &FirmwareInfoProvider::get_board_id)
-        .def("get_eth_fw_version", &FirmwareInfoProvider::get_eth_fw_version)
-        .def("get_asic_location", &FirmwareInfoProvider::get_asic_location)
-        .def("get_aiclk", &FirmwareInfoProvider::get_aiclk)
-        .def("get_axiclk", &FirmwareInfoProvider::get_axiclk)
-        .def("get_arcclk", &FirmwareInfoProvider::get_arcclk)
-        .def("get_fan_speed", &FirmwareInfoProvider::get_fan_speed)
-        .def("get_tdp", &FirmwareInfoProvider::get_tdp)
-        .def("get_tdc", &FirmwareInfoProvider::get_tdc)
-        .def("get_vcore", &FirmwareInfoProvider::get_vcore)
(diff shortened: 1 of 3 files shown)


device/api/umd/device/arc/arc_messenger.hpp
```diff
@@ -22,9 +22,10 @@ class ArcMessenger {
      * Create an ArcMessenger object.
      *
      * @param tt_device TTDevice object used to communicate with the ARC of the device.
+     * @param use_noc1 Whether to use NOC1 for communication during construction.
```
nbuncicTT: I am very hesitant to introduce this flag to every API that we have for a couple of reasons:
- Using `bool use_noc1` through 85 files and all public APIs creates significant technical debt and changes the API signatures everywhere
- It couples us tightly to the current hardware (where 0/1 is the only choice) and will break when we introduce Grendel (which has System NOCs etc.) so it will be hard to migrate everything
- Significant interface pollution where components have to "think" about the NOC even though that API is just for reading some parameter (A method like get_asic_temperature() asks "What is the temperature?". It should not ask "Which NOC should I use to find the temperature?" - I'm aware that it's defaulted to false on some places, but still it incurs this behavior in many cases)

I propose to use the something like the proxy pattern:
```cpp
// 1. Define semantic modes
enum class NocMode { NOC0, NOC1, SystemNOC };

class TTDevice {
public:
    // Existing clean API (unchanged, defaults to NocMode::Data)
    void write(uint64_t addr, const void* data, size_t size);

    // New: Returns the lightweight proxy
    TTDeviceNocView as_noc(NocMode mode);

protected:
    // The actual logic (protected, accessible only by the view)
    friend struct TTDeviceNocView;
    virtual void write_internal(uint64_t addr, const void* data, size_t size, NocMode mode);
};

// 2. The Helper Struct (Stack-allocated View)
struct TTDeviceNocView {
    TTDevice* device;
    NocMode mode;
    
    // Mirrors the main API but passes the mode to internal logic
    void write(uint64_t addr, const void* data, size_t size) {
        device->write_internal(addr, data, size, mode);
    }
};
```
and the usage becomes trivial to implement for other stuff in the same manner, so we just add this where we need it and most of the API's stay the same.

If you want, we can also discuss this offline, maybe I'm overlooking a constraint that makes the boolean flag necessary.

pjanevskiTT: Thinking same as Nenad here, adding this to all functions seems problematic. I think his proposal makes sense, but maybe we can come up with something better. Not quite sure what is wrong with current scheme that apart from umd_use_noc1 being global variable. TTDevice could expose function that just switches NOC it operates on and from there on client knows NOC1 (or any other NOC in the future) will be used

PR #1749, opened 2026-01-04
TLS implementation of Noc selector for device communication

### LINKED ISSUE #526: Expose sending commands to noc1 and noc0

tt-vjovanovic:
As seen in previous [PR](https://github.com/tenstorrent/tt-umd/pull/1748), exposing the NOC selection throughout the entire API creates a poor experience for both library users and maintainers.

This change replaces that approach with a simple thread-local global variable that is used for all device communication. Conceptually, this is the same as the previous single global variable, except that users can now select the desired `NocId` on a per-thread basis via `NocIdSwitcher`.

For Python users, a `set_thread_noc_id()` API is provided to select the `NocId` prior to invoking any device communication.

Closes #526


pjanevskiTT:
Follow up https://github.com/tenstorrent/tt-umd/issues/1750

tt-vjovanovic:
> lgtm, I think this is much better than #1748. Next step would probably be for this to be wired up for each TTDevice separately, but we can work on that

When you say wire up for each `TTDevice`, you mean allowing users to easily configure which NocId to be used for each `TTDevice` instead of always setting it up before invoking API calls? That makes sense, we want to have that in tt-exalens. On the other hand, you need to know that there are calls that can be done only using selected NocId (for example reading NIU registers on certain noc locations), so you will to make sure that what is selected in `TTDevice` can be overridden, etc...

nbuncicTT (approved):

PR #1751, opened 2026-01-05
Expose TTSim simulation to TTDevice API

pjanevskiTT:
### Issue

#1467 

### Description

Expose TTSim simulation to TTDevice layer. We are moving all simulators to TTDevice layer and this PR moves existing code for TTSim from Chip to TTDevice layer. 

### List of the changes

- Move code from Chip layer for TTSim to TTDevice layer
- Make Chip use TTDevice code

### Testing
CI + local ttsim testing

### API Changes
/

### DIFF
--- a/device/tt_device/tt_sim_tt_device.cpp
+++ b/device/tt_device/tt_sim_tt_device.cpp
@@ -47,7 +47,7 @@ TTSimTTDevice::TTSimTTDevice(
     ChipId chip_id,
     bool copy_sim_binary) :
     simulator_directory_(simulator_directory),
-    soc_descriptor_(soc_descriptor),
+    soc_descriptor_(std::move(soc_descriptor)),
     chip_id_(chip_id),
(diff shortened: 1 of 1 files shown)


device/api/umd/device/simulation/tt_sim_chip.hpp
```diff
@@ -10,6 +10,7 @@
 #include <filesystem>
```
broskoTT: Do we need a different tt_sim chip now? What would be the gaps to use TTSimTTDevice in LocalChip?

broskoTT (approved):
Great job!

device/api/umd/device/simulation/tt_sim_chip.hpp
pjanevskiTT: we can use the same TTSimChip, but we are working towards moving out all simulator logic out of Chip. We can't use TTSimDevice in LocalChip still since it doesn't have everything supported in the same way as silicon TTDevice

PR #1753, opened 2026-01-06
Remove unused variable 'start' from methods

blozano-tt:
Dead code?

This is getting flagged by Clang Static Analyzer.


broskoTT (approved):
I'm fine with the fix, but would like us to think about briefly if this could've been caught with a test? If it is exceptionally hard to test, then this is fine, if it is fairly easy let's have that test.

blozano-tt:
I am scanning UMD as part of metal if you are interested in peeking at the results from time to time:

https://blozano-tt.github.io/clangsa-results/#sort=file-path&asc=1

blozano-tt:
I'll let you folks take ownership of this PR. Please merge it when you are ready.

Issue #1731, opened 2025-12-25
Investigate failing EriscFirmwareHashCheck


broskoTT:
This test is failing on tt-metal's 6u galaxy. I wasn't able to reproduce manually on IRD galaxy
The test itself is not failing, but when this test is run, the following tests start failing:

```
[  FAILED  ] 16 tests, listed below:
[  FAILED  ] ApiTTDeviceTest.TTDeviceRegIO
[  FAILED  ] AllTriscNcriscCoreCombinations/ClusterAssertDeassertRiscsTest.TriscNcriscAssertDeassertTest/0, where GetParam() = { (131072, 8192, { 10039, 466979, 468867, 1542035, 16195619, 4284477551 }, TRISC0) }
```

aleksamarkovicTT:
Closing because feature is removed.

PR #1766, opened 2026-01-12
Reenable TestCluster.EriscFirmwareHashCheck

aleksamarkovicTT:
### Issue
#1731 

### Description
Reenable EriscFirmwareHashCheck, but skip for Galaxy 6U, to keep verifying that the ETH FW hash check functions correctly.

### List of the changes
- Reenabled TestCluster.EriscFirmwareHashCheck
- Skip TestCluster.EriscFirmwareHashCheck for Galaxy 6U.

### Testing
CI

### API Changes


tests/api/test_cluster.cpp
```diff
@@ -1351,10 +1354,6 @@ TEST(TestCluster, DISABLED_EriscFirmwareHashCheck) {
     std::cout << "Passed hash check." << std::endl;
 
     // Revert ERISC FW state with warm reset.
-    if (is_galaxy_configuration(cluster.get())) {
```
pjanevskiTT: I would maybe keep this if, so when we enable this test for Galaxy we don't this to change this again. If this is always going to be non-galaxy test, then it's fine

pjanevskiTT (approved):

broskoTT (approved):
Did you end up figuring out why it doesn't work on 6u?

aleksamarkovicTT:
> Did you end up figuring out why it doesn't work on 6u?

The test never failed on 6U. It was some other test you saw failing, no repro, no stack trace.
I couldn't repro this test failing locally on FW 19.5 wh-glx6u-03, so I'll count it as a one off failure.
I will actually reenable it for 6U, and see what happens.

PR #1780, opened 2026-01-12
Tools-UMD Migration

kaplanz:
This PR will track and ultimately be used to merge tools' migration to UMD. This will be implemented by introducing a low-level API to UMD as a stable common interface to export low-level device operation primitives.


pjanevskiTT:
looks like stale work, will close

PR #1793, opened 2026-01-14
Fix n150 eth coord

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/1820

### Description
Observed while testing tt-smi with umd.
Seems like we regressed at some point, and stop filling up eth_coord in case of a single N150

### List of the changes
- Bubble up the eth_coord assignment a bit, so it is populated even if the ethernet core is not connected to anything.

### Testing
I observed the new test failing without this fix and then it passes after this fix

### API Changes


broskoTT:
Due to the nature of changes, and the combination of tests that already passed on previous commit and the latest one, I'm bypassing rules and merging without waiting on all tests to complete, as I'm fairly confident that all the important ones passed

PR #1798, opened 2026-01-14
warm_reset: Add secondary bus reset flag to make it optional

sbansalTT:
### Description
- On the glx machines we need to skip the secondary bus reset by disabling the RESET_PCIE_LINK flag to kmd
- Added the flag to warm reset function and the python binding
- Added subsystem id and subsystem vendor id into pci device and python bindings
- Tested against WH galaxy with the warm_reset pytest  that was added

### List of the changes
- added secondary bus reset flag to cpp and pybindings
- added subsystem id and subsystem vendor id to cpp and pybindings

### Testing
-Tested against WH galaxy with the warm_reset pytest

### API Changes
- No API changes against metal should have happened

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -6,7 +6,8 @@ Checks: >
     bugprone-use-after-move,
     bugprone-redundant-branch-condition,
     cppcoreguidelines-use-enum-class,
-    cppcoreguidelines-pro-type-cstyle-cast
+    cppcoreguidelines-pro-type-cstyle-cast,
+    readability-container-size-empty
 
 HeaderFilterRegex: '^(?!.*\.cpmcache).*$'
 WarningsAsErrors: '*'
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -138,9 +138,11 @@ add_compile_options(
     -Werror
     -Wno-unused-variable
     -Wno-unused-function
-    -Wno-unused-but-set-variable
     -Wno-sign-compare
 )
+
+# Fixed this warning on clang, but not yet gcc
+add_compile_options("$<$<NOT:$<CXX_COMPILER_ID:Clang>>:-Wno-unused-but-set-variable>")
 
 # Conditionally add -Wno-maybe-uninitialized for GCC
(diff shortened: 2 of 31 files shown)


nanobind/tests/test_py_warm_reset.py
```diff
@@ -5,7 +5,7 @@
 import tt_umd
 
 class TestWarmReset(unittest.TestCase):
-    @unittest.skip("Skipping warm reset test to avoid resetting cards during unit tests")
+    # @unittest.skip("Skipping warm reset test to avoid resetting cards during unit tests")
```
nbuncicTT: @broskoTT you skipped this test for some reason - is the reason of this test being skipped now solved?

I assume that you didn't call topology discovery after reset here so that's why you skipped it, to not mess up other tests.

nanobind/tests/test_py_warm_reset_wh_galaxy.py
```diff
@@ -0,0 +1,46 @@
+# SPDX-FileCopyrightText: © 2025 Tenstorrent Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+import unittest
+import tt_umd
+
+class TestWarmReset(unittest.TestCase):
+    # @unittest.skip("Skipping warm reset test to avoid resetting cards during unit tests")
```
nbuncicTT: Similar comment like the one from above, should we do topology discovery so that other tests can run without crashing afterwards (or am I missing something)?

device/warm_reset.cpp
```diff
@@ -123,7 +126,9 @@ void WarmReset::warm_reset_arch_agnostic(
     }
 
     log_info(tt::LogUMD, "Starting reset on devices at PCI indices: {}", fmt::join(pci_device_id_set, ", "));
-    PCIDevice::reset_device_ioctl(pci_device_id_set, TenstorrentResetDevice::RESET_PCIE_LINK);
+    if (secondary_bus_reset) {
```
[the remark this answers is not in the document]
broskoTT: I mentioned this offline to Sam, but I'm also fine if we leave it in the interface for potential future flexibility, so we can leave it as is

PR #1808, opened 2026-01-15
Add step to run tt-exalens tests in CI workflow

tt-vjovanovic:
As agreed, we should run tt-exalens tests on top of UMD wheel.

### DIFF
--- a/.github/workflows/build-device.yml
+++ b/.github/workflows/build-device.yml
@@ -269,8 +269,8 @@ jobs:
           python -m cibuildwheel --output-dir ${{ env.WHEEL_OUTPUT_DIR }}
           echo "Build complete."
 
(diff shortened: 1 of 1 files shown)


.github/workflows/run-tests.yml
```diff
@@ -99,6 +99,30 @@ jobs:
           echo "Running tests..."
           .venv/bin/python -m pytest -s tests
 
+      - name: Run tt-exalens tests
```
pjanevskiTT: how stable are these tests? Do we want to make this required in order to check in UMD PRs? @broskoTT 

@tt-vjovanovic how long do these tests take?

pjanevskiTT (approved):

PR #1811, opened 2026-01-15
Trying to speed up build by caching previous build

tt-vjovanovic:
When new PR is sent, it should reuse existing `main` branch builds. When new iteration is sent, builds from that branch should be reused.

Missing: Docker tag should be correctly set for workflow instead of using latest.

### DIFF
--- a/.github/actions/build_with_cache/action.yml
+++ b/.github/actions/build_with_cache/action.yml
@@ -348,7 +348,7 @@ runs:
       working-directory: ${{ inputs.repo-path }}
       run: |
         # Script is always at the same relative location from within repo-path
-        python3 .github/actions/save_build_cache/adjust_artifact_timestamps.py ${{ inputs.build-output-dir }}
+        python3 $GITHUB_ACTION_PATH/adjust_artifact_timestamps.py ${{ inputs.build-output-dir }}
 
     - name: Save build cache
       uses: actions/cache/save@v4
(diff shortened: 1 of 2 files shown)


blozano-tt (changes requested):
I would suggest using ccache. The infrastructure is already in place for tt-metal. 

I think it’s time for UMD build flow to align more closely with tt-metal. This would give access to faster more powerful build machines, redis ccache, and allow for more thorough clang-tidy scans.

cc: @afuller-TT

tt-vjovanovic:
> I would suggest using ccache. The infrastructure is already in place for tt-metal.
> 
> I think it’s time for UMD build flow to align more closely with tt-metal. This would give access to faster more powerful build machines, redis ccache, and allow for more thorough clang-tidy scans.
> 
> cc: @afuller-TT

ccache solves some problems, but also introduces overhead.
Something I was interested in was this issue: https://github.com/ninja-build/ninja/issues/1459
Anyways, this is experiment. If it succeeds, I'll compare build time from this experiment and with ccache, so I can give you numbers. What I hate with tt-metal builds (where ccache is setup) as well is that when I change something in tt-triage, I need to wait ~10-15 minutes to get build ready before I can go to step of running tt-triage tests. This experiment would return result immediately as build is already present in cache.

afuller-TT:
Always good to explore options to see what our choices are.  Thanks @tt-vjovanovic !

Would you like a hand setting up CCache for UMD?

Issue #1820, opened 2026-01-16
Publishing tt-smi with tt-umd efforts


broskoTT:
There are some efforts in fixing all the last minute issues, provisioning efforts, etc.
This issue tracks those efforts.

broskoTT:
This was done a while back, tt-smi with umd is being used internally

PR #1821, opened 2026-01-16
Manylinux 2.34 -> 2.28

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/1820

### Description
There is a problem with deploying umd wheels on CIv1 fleet.
Specifically, this will make umd wheels supported on ubuntu 20.04

### List of the changes
- Lower manylinux image to 2.28
- Minor fix in robust mutex to make umd build on older linux

### Testing
I manually tested, but the CI should show if this build is passing.
If it is passing, I have high confidence the deployment will actually work on ubuntu 20

### API Changes


device/utils/robust_mutex.cpp
```diff
@@ -414,7 +415,7 @@ void RobustMutex::lock() {
     }
 
     // lock_res is 0, so this is a success case.
-    mutex_wrapper_ptr_->owner_tid = gettid();
+    mutex_wrapper_ptr_->owner_tid = static_cast<pid_t>(::syscall(SYS_gettid));
```
nbuncicTT: Small explanation why this cast is needed?

broskoTT: Not sure for the cast actually, this is what ai produced

Issue #1836, opened 2026-01-19
Delete all code related to 4U galaxy


pjanevskiTT:
Delete all code from UMD codebase related to 4U galaxy systems since these are now deprecated

PR #1837, opened 2026-01-19
Remove 4U galaxy related code from codebase

pjanevskiTT:
### Issue

#1836 

### Description

Remove all code related to 4U galaxy. These systems are now deprecated and we don't need to support it anymore. A lot of code used for 4U galaxy performed computations for other systems that were not needed in those cases. It also simplifies logic in ClusterDescriptor quite a bit.

### List of the changes

- Remove calculations based on ETH coordinates inside ClusterDescriptor
- Remove functions that were implemented specifically for 4U
- Remove GALAXY enum and it's usages
- Remove cluster descriptor for galaxy
- Remove all code specifying 4U in test and other utils

### Testing
CI

### API Changes
/

### DIFF
--- a/tests/test_utils/test_api_common.hpp
+++ b/tests/test_utils/test_api_common.hpp
@@ -86,9 +86,8 @@ private:
 
 // Helper function to detect if the cluster is a Galaxy configuration, including 4U and 6U configurations.
 inline bool is_galaxy_configuration(Cluster* cluster) {
-    bool is_6u_galaxy_configuration = !cluster->get_target_device_ids().empty() &&
-                                      cluster->get_cluster_description()->get_board_type(0) == tt::BoardType::UBB;
-    return is_6u_galaxy_configuration;
+    return !cluster->get_target_device_ids().empty() &&
+           cluster->get_cluster_description()->get_board_type(0) == tt::BoardType::UBB;
 }
 
 inline bool has_remote_chips() {


tests/test_utils/test_api_common.hpp
```diff
@@ -84,18 +84,11 @@ class ClusterAssertDeassertRiscsTest : public ::testing::TestWithParam<RiscSetUn
...
-// Helper function to detect if the cluster is a 4U Galaxy configuration.
-inline bool is_4u_galaxy_configuration(Cluster* cluster) {
-    return cluster != nullptr && !cluster->get_target_remote_device_ids().empty() &&
-           cluster->get_cluster_description()->get_board_type(*cluster->get_target_remote_device_ids().begin()) ==
-               tt::BoardType::GALAXY;
-}
-
 // Helper function to detect if the cluster is a Galaxy configuration, including 4U and 6U configurations.
 inline bool is_galaxy_configuration(Cluster* cluster) {
     bool is_6u_galaxy_configuration = !cluster->get_target_device_ids().empty() &&
```
broskoTT: consider reducing to single return statement.
Also consider changing the naming to be more specific, maybe just a simple is_6u()?

broskoTT (approved):
Thanks for doing this!

tests/cluster_descriptor_examples/wormhole_4xN300_mesh.yaml
```diff
@@ -1,75 +0,0 @@
-arch: {
```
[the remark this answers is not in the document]
pjanevskiTT: Duplicates

tests/test_utils/test_api_common.hpp
pjanevskiTT: will reduce, I think naming is descriptive. We should look at ubb and 6u naming, but won't address this here

PR #1839, opened 2026-01-19
Make predicting ETH FW version optional

aleksamarkovicTT:
### Issue
#1861

### Description
Make predicting ETH FW optional

### List of the changes
- Make predicting ETH FW optional, set to false by default in TopologyDiscoveryOptions.

### Testing
CI

### API Changes


device/topology/topology_discovery_blackhole.cpp
broskoTT: I thought this is something where you'd have to apply changes? In this case to return no eth_fw_problem, so that discovery can continue?

broskoTT: Discussed offline. Understood the changes

PR #1857, opened 2026-01-19
Update build-clients

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/1858

### Description
As a follow up of https://github.com/tenstorrent/tt-umd/pull/1833 I looked at our workflows.
Also after recent changes in how ttsim tests are run in https://github.com/tenstorrent/tt-umd/issues/1788 we now don't even need another job which just build tt-metal, since ttsim one already does that

### List of the changes
- Remove build tt-metal job, as it is included in ttsim job
- Change docker download to use harbor.ci

### API Changes

### DIFF
--- a/.github/workflows/build-clients.yml
+++ b/.github/workflows/build-clients.yml
@@ -13,6 +13,8 @@ on:
     branches: ["main"]
 
 jobs:
+  # Note: this job also serves as the tt-metal build verification job —
+  # building tt-metal is a required step before running ttsim tests.
   build-tt-sim:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-clients.yml
```diff
@@ -13,46 +13,6 @@ on:
     branches: ["main"]
 
 jobs:
-  build-tt-metal:
```
pjanevskiTT: can we just add a comment somewhere that ttsim job is basically used as a job for testint tt-metal build?

PR #1866, opened 2026-01-20
Enable performance inefficient vector operation

blozano-tt:
### Issue
NA

### Description
Enable the currently disabled performance inefficient vector operation clang tidy check.
https://clang.llvm.org/extra/clang-tidy/checks/performance/inefficient-vector-operation.html

### List of the changes
- Enable check, so new code cannot be introduced that violates this rule.
- Apply automatic FIXITs

### Testing
CI

### API Changes


pjanevskiTT (approved):
lgtm

PR #1868, opened 2026-01-20
Use RobustMutex for multiprocess PCIe DMA usage

pjanevskiTT:
### Issue
/

### Description

Fix multiprocess usage of PCIe DMA. Instead of using std::mutex lock use RobustMutex. This fixes the problem because except TLB, which can be isolated by local std mutex, there is also DMA resource which is distributed through BAR so needs proper multiprocess isolation, instead of per TTDevice object. This PR also adds a test which failed before fixes. 

### List of the changes

- Add a test which demonstrated bad isolation between processes
- Replace std mutex with RobustMutex for PCIe DMA usage

### Testing
CI + local testing

### API Changes
/

### DIFF
--- a/tests/unified/multiprocess.cpp
+++ b/tests/unified/multiprocess.cpp
@@ -318,6 +318,7 @@ TEST(Multiprocess, DMAWriteReadRaceCondition) {
     std::cout << "Testing DMA race condition on PCI device " << test_device_id << std::endl;
 
     std::vector<std::thread> process_threads;
+    process_threads.reserve(num_processes);
 
     for (int process_id = 0; process_id < num_processes; process_id++) {
         process_threads.push_back(std::thread([=]() {


tests/unified/multiprocess.cpp
```diff
@@ -294,3 +294,81 @@ TEST(Multiprocess, ClusterAndTTDeviceTest) {
...
+// use TTDevice objects with the same underlying PCIDevice
+TEST(Multiprocess, DMAWriteReadRaceCondition) {
+    std::vector<int> pci_device_ids = PCIDevice::enumerate_devices();
+
+    if (pci_device_ids.empty()) {
+        GTEST_SKIP() << "No chips present on the system. Skipping test.";
+    }
+
+    // Use the first available PCI device for this test.
+    const int test_device_id = pci_device_ids.at(0);
```
nbuncicTT: nit: change const things to constexpr

Issue #1846, opened 2026-01-19
Investigate Timeout waiting for Ethernet core service remote IO request.


broskoTT:
This started showing up in our CI.
It was made significantly worse by this PR, which was then reverted but we still see this issue (stating this as it might help investigation) https://github.com/tenstorrent/tt-umd/pull/1696

Potentially related: https://github.com/tenstorrent/tt-umd/issues/1535

pjanevskiTT:
TODOs for @aleksamarkovicTT 

- try to disable ethernet_broadcast_write, use unicasts instead
- try to repro everything locally

aleksamarkovicTT:
PR to disable ethernet_broadcast_write(): https://github.com/tenstorrent/tt-umd/pull/1862

pjanevskiTT:
PR to use WC in topology discovery as well #1869

broskoTT:
There's a lot of details on: https://tenstorrent.slack.com/archives/C07L16Z59AP/p1770394958164549

As last discussed offline @aleksamarkovicTT:
- Create an issue on this board https://tenstorrent.atlassian.net/jira/software/c/projects/SWE/boards/1351, and ping the thread with the info
- Provide instructions on how to create a eth FW which triggers this issue more easily.
- Provide the script from our side, and instructions on how to provision tt-umd to be able to run the script.

broskoTT:
Complete by introducing heartbeat checks and posted the investigation results here: https://tenstorrent.atlassian.net/wiki/spaces/LLK/pages/2142339983/WH-ERISC+post-reset+hang+investigation

PR #1924, opened 2026-01-27
Implement the changes to mitigate ETH FW routing bugs

pjanevskiTT:
### Issue

#1846 

### Description

Implement the changes needed to mitigate the bug in ETH firmware. After multiple changes we have seen bug with ETH firmware timing out when we want to issue requests. This in particular happens after performing warm reset. This PR applies multiple changes that could help mitigate this behavior. 

### List of the changes

- Add WC API to TTDevice for read and write
- Use that API in remote communication and topology discovery
- Wait for ETH training in more that one way, heartbeat and one more field for training
- 

### Testing
Was tested by running warm reset test on N300 and T3K machines in over 400 jobs multiple times. Around 1000 jobs passed in total

### API Changes
/


aleksamarkovicTT (approved):

broskoTT (changes requested):
Given that these changes don't seem to solve the issue reported, we should either drop or revisit if this changes make sense

pjanevskiTT:
we can close it...there are useful changes that we should integrate into separate PRs

PR #1928, opened 2026-01-27
Specify min compiler versions gcc 11 and clang 13

broskoTT:
### Issue
A follow up to https://github.com/tenstorrent/tt-umd/pull/1926

### Description
Be specific in readme and CI what are the oldest compilers we support

### List of the changes
- Lower gcc 12 to gcc 11
- Add clang 13 installation to docker
- Add jobs which test clang 13 build

### Testing
CI should be enough

### API Changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Required Ubuntu dependencies:
 sudo apt install -y libhwloc-dev cmake ninja-build
 ```
 
-UMD currently supports gcc-11 and newer compilers, and clang-13 and newer compilers.
+UMD currently supports gcc-11 and newer gcc versions, and clang-13 and newer clang versions.
 
 ## IOMMU and Hugepage requirements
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -18,12 +18,7 @@ Required Ubuntu dependencies:
 sudo apt install -y libhwloc-dev cmake ninja-build
 ```
 
-Suggested third-party dependency is Clang 17:
-```
-wget https://apt.llvm.org/llvm.sh
-chmod u+x llvm.sh
-sudo ./llvm.sh 17
-```
+UMD currently supports gcc-11 and newer compilers, and clang-13 and newer compilers.
```
nbuncicTT: nit: `UMD currently supports gcc-11 and newer gcc versions, and clang-13 and newer clang versions.`

PR #2028, opened 2026-02-02
Capture init. state in TTDevice

aleksamarkovicTT:
### Issue
#2063

### Description
Capture init. state in TT Device.

### List of the changes
- Define `TTDeviceInitResult` enum.
- Return `TTDeviceInitResult` in `init_tt_device()`.
- Keep default behavior of throwing exception on ARC startup failure. Make it optional.

### Testing
CI

### API Changes
This PR has API changes.

### DIFF
--- a/device/api/umd/device/tt_device/tt_device.hpp
+++ b/device/api/umd/device/tt_device/tt_device.hpp
@@ -28,7 +28,7 @@ class ArcMessenger;
 class ArcTelemetryReader;
 class RemoteCommunication;
 
-enum class TTDeviceInitResult : uint8_t {
+enum class TTDeviceInitResult {
     UNKNOWN = 0,
     UNINITIALIZED,
     ARC_STARTUP_FAILED,
@@ -319,8 +319,6 @@ public:
      */
     virtual void dma_multicast_write(void *src, size_t size, tt_xy_pair core_start, tt_xy_pair core_end, uint64_t addr);
 
-    TTDeviceInitResult get_last_init_result() const;
-
 protected:
     std::shared_ptr<PCIDevice> pci_device_;
     std::shared_ptr<JtagDevice> jtag_device_;
@@ -341,8 +339,6 @@ protected:
     tt_xy_pair arc_core;
 
 private:
-    TTDeviceInitResult last_init_result_ = TTDeviceInitResult::UNINITIALIZED;
-
     void probe_arc();
 
     TlbWindow *get_cached_tlb_window();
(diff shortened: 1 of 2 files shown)


device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -268,7 +269,8 @@ class TTDevice {
 
     bool is_remote();
 
-    void init_tt_device(const std::chrono::milliseconds timeout_ms = timeout::ARC_STARTUP_TIMEOUT);
+    TTDeviceInitResult init_tt_device(
+        std::chrono::milliseconds timeout_ms = timeout::ARC_STARTUP_TIMEOUT, bool throw_on_arc_failure = true);
```
pjanevskiTT: can we capture all these options in something like TTDeviceInitOptions? I would assume we will have more variables like throw_on_arc_failure?

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -328,6 +341,8 @@ class TTDevice {
     tt_xy_pair arc_core;
 
 private:
+    TTDeviceInitResult last_init_result_ = TTDeviceInitResult::UNINITIALIZED;
```
nbuncicTT: I don't recall why are we adding state to this? Wasn't the offline discussion conclusion that the user should evaluate the result and we should not cache this info?

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -28,6 +28,16 @@ class ArcMessenger;
 class ArcTelemetryReader;
 class RemoteCommunication;
 
+enum class TTDeviceInitResult : uint8_t {
```
nbuncicTT: Why are you using uint8_t here? If you're not casting the value from/to somewhere where uint8_t is needed - I'd remove it.

nbuncicTT (commented):
Overall good job, I'd just omit the state and make the user wrap the call with a try catch block or evaluate the return value. 

Not sure if init_tt_device should ever throw though - I'd like it to have the [[nodiscard]] attribute and the top level component that uses it should determine what to do (but I'm okay with the bool flag here, just seems ambiguous to me).

device/api/umd/device/types/tt_device_health.hpp
```diff
@@ -0,0 +1,22 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+#include <cstdint>
+
+namespace tt::umd {
+
+enum class TTDeviceInitResult {
```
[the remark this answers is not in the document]
nbuncicTT: Recommendation (seen this quite a few times)

Unless you are constrained by hardware register layouts or extremely tight memory packing requirements, avoid uint8_t for enums.

PR #2065, opened 2026-02-04
Implement TTSimCommunicator class

pjanevskiTT:
### Issue

General improvement to TTSim flow.

### Description

This PR introduces the class TTSimCommunicator which should encapsulate communication with TTSim .so file. While moving TTSim code to TTDevice layer, a lot of boilerplate code was required to move. Now we could just move TTSimCommunicator class. It represent an API towards TTSim .so file, basically we wrap @mcraigheadTT's API into c++ API that we could use. This will also help with moving more code towards PCIDevice layer for simulation and unifying the code with silicon.

### List of the changes

- Create TTSimCommunicator class
- Wrap .so file API into class API
- use TTSimCommunicator in TTDevice

### Testing

CI

### API Changes
/

### DIFF
--- a/device/api/umd/device/simulation/tt_sim_communicator.hpp
+++ b/device/api/umd/device/simulation/tt_sim_communicator.hpp
@@ -17,7 +17,7 @@ namespace tt::umd {
  *
  * This class can be used independently of TTSimTTDevice for direct simulator communication.
  */
-class TTSimCommunicator {
+class TTSimCommunicator final {
 public:
     /**
      * Constructor for TTSimCommunicator.


device/api/umd/device/tt_device/tt_sim_tt_device.hpp
```diff
@@ -59,31 +63,16 @@ class TTSimTTDevice : public TTDevice {
...
-    void (*pfn_libttsim_pci_mem_rd_bytes)(uint64_t paddr, void *p, uint32_t size) = nullptr;
-    void (*pfn_libttsim_pci_mem_wr_bytes)(uint64_t paddr, const void *p, uint32_t size) = nullptr;
-    void (*pfn_libttsim_tile_rd_bytes)(uint32_t x, uint32_t y, uint64_t addr, void *p, uint32_t size) = nullptr;
-    void (*pfn_libttsim_tile_wr_bytes)(uint32_t x, uint32_t y, uint64_t addr, const void *p, uint32_t size) = nullptr;
-    void (*pfn_libttsim_clock)(uint32_t n_clocks) = nullptr;
     uint64_t bar0_base = 0;
-    uint32_t tlb_region_size = 0;
-
+    uint32_t tlb_region_size_ = 0;
+    std::unique_ptr<TTSimCommunicator> communicator_;
```
nbuncicTT: As discussed offline, `TTSimTTDevice` doesn't work without the new helper class `TTSimCommunicator` - hence consider using a reference instead of a pointer.

device/api/umd/device/simulation/tt_sim_communicator.hpp
```diff
@@ -0,0 +1,138 @@
...
+     *
+     * @param simulator_directory Path to the simulator binary/directory
+     * @param copy_sim_binary If true, copy the simulator binary to memory for security
+     */
+    TTSimCommunicator(const std::filesystem::path &simulator_directory, bool copy_sim_binary = false);
+
+    /**
+     * Destructor that properly cleans up library handles and file descriptors.
+     */
+    ~TTSimCommunicator();
```
nbuncicTT: If you ever intend to use this class as a base class, make the destructor public. If not, make the class final:
```cpp
class TTSimCommunicator final {
```

device/api/umd/device/tt_device/tt_sim_tt_device.hpp
nbuncicTT: Sorry, my bad, I misread the code, I thought that you were passing a long-lived object into TTSim via a pointer so that makes sense to pass as a reference, but TTSim is the manager of the lifetime of the object and this isn't feasible.

Why don't you use TTSimCommunicator as just a plain object instead of a pointer?

pjanevskiTT: Because all our helper classes are pointers, we can change that for all classes if needed, not gonna change that just for this class

PR #2066, opened 2026-02-04
Adding TT_VISIBLE_DEVICES visibility to mock cluster descriptors

Riddy21:
### Issue
Enabled TT_VISIBLE_DEVICES device filtering to work for Mock Clusters using cluster descriptors


pjanevskiTT (commented):
Yes this won't work. I would close this...I will open an issue put up a PR addressing the use case we discussed offline

PR #2093, opened 2026-02-07
Remove -Wno-unused-function (again)

aleksamarkovicTT:
### Issue
/

### Description
Remove -Wno-unused-function
Accidentally reintroduced in #1883.

### List of the changes
- Remove -Wno-unused-function
- Add inline to some static header defined functions.

### Testing
Build passing

### API Changes

### DIFF
--- a/device/common/utils.hpp
+++ b/device/common/utils.hpp
@@ -17,7 +17,7 @@
 
 namespace tt::umd::utils {
 
-static inline std::optional<std::string> get_env_var_value(const char* env_var_name) {
+inline std::optional<std::string> get_env_var_value(const char* env_var_name) {
     const char* env_var = std::getenv(env_var_name);
     if (!env_var) {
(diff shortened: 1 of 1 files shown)


device/common/utils.hpp
```diff
@@ -17,15 +17,15 @@
 
 namespace tt::umd::utils {
 
-static std::optional<std::string> get_env_var_value(const char* env_var_name) {
+static inline std::optional<std::string> get_env_var_value(const char* env_var_name) {
```
blozano-tt:
```suggestion
inline std::optional<std::string> get_env_var_value(const char* env_var_name) {
```

device/common/utils.hpp
blozano-tt: I think static is redundant here.

pjanevskiTT (approved):

PR #2109, opened 2026-02-13
Introduce ERISC FW heartbeat check in TopologyDiscovery

aleksamarkovicTT:
### Issue
#1846 #2162

### Description
Introduce ERISC FW heartbeat check in TopologyDiscovery. This check is done on every ETH core of every discovered device.
The check reads from a fixed value in ETH L1 for a uin32_t in the format of `ABCDxxxx`. If the ETH FW is running, this value should be updated between two consecutive reads.

### List of the changes
- Introduce helper methods to read heartbeat and postcode for ERISC FW.
- Introduce method to check if the heartbeat is functioning.
- Added new topology discovery option `eth_fw_heartbeat_failure`.
- Remove heartbeat check in WormholeTTDevice::wait_eth_core_training().
 
### Testing
CI
Manually tested on n300 with hung local and remote ERISC after WarmReset.

### API Changes
There are behavior changes in this PR:
- tt-exalens: https://github.com/tenstorrent/tt-exalens/pull/931

### DIFF
--- a/device/topology/topology_discovery_wormhole.cpp
+++ b/device/topology/topology_discovery_wormhole.cpp
@@ -19,6 +19,7 @@
 #include "umd/device/firmware/firmware_utils.hpp"
 #include "umd/device/tt_device/remote_communication.hpp"
 #include "umd/device/tt_device/tt_device.hpp"
+#include "umd/device/types/wormhole_eth.hpp"
 #include "umd/device/types/xy_pair.hpp"
 #include "umd/device/utils/semver.hpp"
 


device/topology/topology_discovery.cpp
```diff
@@ -469,4 +488,46 @@ SocDescriptor TopologyDiscovery::get_soc_descriptor(TTDevice* tt_device) {
...
+            break;
+        }
+        if (utils::check_timeout(
+                start,
+                timeout::ETH_STARTUP_TIMEOUT,
+                "Timed out waiting for ETH hearbeat.",
+                utils::TimeoutAction::Return)) {
+            return false;
+        }
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
pjanevskiTT: nit: could we implement more precise timeout somehow? Or if 1 ms is already good enough, add a comment explaining why

device/topology/topology_discovery_blackhole.cpp
```diff
@@ -32,15 +32,8 @@ TopologyDiscoveryBlackhole::TopologyDiscoveryBlackhole(const TopologyDiscoveryOp
...
-    // ETH coord is not used for Blackhole, as Blackhole does not have a concept of ETH coordinates.
-    std::unique_ptr<RemoteCommunication> remote_communication =
-        RemoteCommunication::create_remote_communication(gateway_device, {0, 0, 0, 0});
-    remote_communication->set_remote_transfer_ethernet_cores(
-        get_soc_descriptor(gateway_device)
-            .get_eth_xy_pairs_for_channels(gateway_eth_channels, CoordSystem::TRANSLATED));
-    std::unique_ptr<TTDevice> remote_tt_device = TTDevice::create(std::move(remote_communication));
-    remote_tt_device->init_tt_device();
-    return remote_tt_device;
+    TT_THROW("Creating remote devices in unsupported on Blackhole.");
```
pjanevskiTT: again, can we maybe move this to separate PR?

device/topology/topology_discovery.cpp
```diff
@@ -469,4 +503,38 @@ SocDescriptor TopologyDiscovery::get_soc_descriptor(TTDevice* tt_device) {
     return soc_descriptor;
 }
 
+bool TopologyDiscovery::eth_heartbeat_running(TTDevice* tt_device, tt_xy_pair eth_core) {
```
pjanevskiTT: https://github.com/tenstorrent/tt-umd/blob/main/device/tt_device/wormhole_tt_device.cpp#L411 can you look at this code and maybe remove it if it does the similar thing to your code?

pjanevskiTT (commented):
i would prefer splitting some of the changes, otherwise I am fine with the changes. Will approve it once it's split

device/topology/topology_discovery.cpp
```diff
@@ -191,6 +196,20 @@ void TopologyDiscovery::discover_remote_devices() {
         for (const CoreCoord& eth_core : eth_cores) {
             const uint32_t channel = get_soc_descriptor(tt_device).get_eth_channel_for_core(eth_core);
 
+            if (!eth_heartbeat_running(tt_device, eth_core)) {
```
broskoTT: So seems like we're checking this for all eth cores.
Maybe comment somewhere if that's the case.
Specifically: The code might throw if some unconnected eth core fails, which we don't care that much about, on n300 for example. 

I do think it is still good to throw in that case, so we're aware of the problem

device/api/umd/device/topology/topology_discovery.hpp
```diff
@@ -81,6 +81,12 @@ class TopologyDiscovery {
     // eth_core should be in NoC 0 coordinates.
     virtual uint64_t get_remote_asic_id(TTDevice* tt_device, tt_xy_pair eth_core) = 0;
```
broskoTT: Just thinking, in general we've defined:
- DeviceProtocol - providing abstraction of IO to our devices
- TTDevice - manages lifetime of our devices
- TopologyDiscovery - performs device discovery procedure

And where do we with health checks?
Feels a bit more natural for them to be part of TTDevice.
But also, we could just change the definition for TopologyDiscovery - performs device discovery and evaluates healthness of the cluster of devices.

Just thinking out loud, no action

Issue #1222, opened 2025-08-25
Use proper resource management for PCIe Device and JTAG Device


nbuncicTT:
## Summary
The codebase currently uses `std::shared_ptr` in several places where `std::unique_ptr` would be more appropriate and provide clearer ownership semantics. This change would improve code clarity, performance, and maintainability.

## Background
Smart pointers should reflect the actual ownership model of the objects they manage. In most cases throughout this codebase, objects have clear single ownership with well-defined transfer points, making `unique_ptr` the more appropriate choice.

### 1. Unclear Ownership Semantics
- `shared_ptr` implies shared ownership, but in most cases we have single ownership
- Makes it harder to reason about object lifetimes and destruction order
- Can mask design issues where ownership responsibilities are unclear

### 2. Performance Overhead (not critical, but still important)
- `shared_ptr` has reference counting overhead (atomic operations)
- Larger memory footprint (control block allocation)
- Unnecessary when single ownership is sufficient

### 3. Examples in Codebase
```cpp
// Current usage that could be improved:
std::shared_ptr<PCIDevice> pci_device_;

// In TTDevice constructor:
TTDevice(std::shared_ptr<PCIDevice> pci_device, ...)

// Factory method:
auto pci_device = std::make_shared<PCIDevice>(device_number);

Issue #925, opened 2025-05-28
Remote TTDevice design v2


pjanevskiTT:
#917 is going to introduce remote TTDevice that is going to have the same API as TTDevice base class and can be used in other code paths to get all the needed info for remote devices

However, there are few things that can be made a bit better, which should be a follow up work on the PR above

Sketch of the design is written in https://github.com/tenstorrent/tt-umd/pull/917#discussion_r2111243147

broskoTT:
To make this task better scoped:
Whomever takes this task they should figure out a proper polymorphism (or other mechanisms) design within TTDevice across ARCH dimension and across Local/ETH dimension

PR #2112, opened 2026-02-14
Device Protocol intro

nbuncicTT:
### Issue
#925
#1222

### Description
This pull request introduces a major refactor to the device communication layer by abstracting communication protocols into dedicated protocol classes and interfaces. It removes the legacy `remote_wormhole_tt_device` and replaces direct device handling with protocol-based interfaces, improving modularity and extensibility.

### List of the changes

**1. Introduction of Protocol Abstractions**
- Added new protocol interface and implementation classes: `DeviceProtocol`, `PcieProtocol`, `JtagProtocol`, and `RemoteProtocol`, along with their respective interface headers (`pcie_interface.hpp`, `jtag_interface.hpp`, `remote_interface.hpp`). These define clear APIs for device communication over PCIe, JTAG, and remote connections.

**2. Refactoring Remote Communication**
- Updated `RemoteCommunication` and `RemoteCommunicationLegacyFirmware` to depend on the new `DeviceProtocol` abstraction instead of directly on `TTDevice`, and updated all relevant constructors and methods accordingly.

**3. Removal of Legacy Device Implementation**
- Removed all references to `remote_wormhole_tt_device` in both source and header lists, as well as in includes, reflecting the protocol-based refactor.

**4. Cleanup of BlackholeTTDevice**
- Removed DMA and multicast-related methods from `BlackholeTTDevice`, as these are now handled by protocol interfaces.

**5. Build System Updates**
- Updated `CMakeLists.txt` to add new protocol source files and remove the obsolete device implementation. 

These changes lay the groundwork for a more modular and maintainable device communication codebase, enabling easier extension and protocol-specific optimizations in the future.

### Testing
Manual and CI

### API Changes
/


device/api/umd/device/chip/remote_chip.hpp
```diff
@@ -6,7 +6,6 @@
```
broskoTT: Please create a confluence page with the design explanation (or if there is already one, ignore this step). Then attach the last photo we brainstormed around on the whiteboard. Then attach the link to the design in this PR's description

device/api/umd/device/tt_device/wormhole_tt_device.hpp
```diff
@@ -55,6 +47,8 @@ class WormholeTTDevice : public TTDevice {
 protected:
     WormholeTTDevice(std::shared_ptr<PCIDevice> pci_device, bool use_safe_api);
     WormholeTTDevice(std::shared_ptr<JtagDevice> jtag_device, uint8_t jlink_id);
+    WormholeTTDevice(std::unique_ptr<RemoteCommunication> remote_communication);
```
broskoTT: Why protected? Isn't WormholeTTDevice final?

broskoTT (commented):
Thanks, in general this looks like a nice improvement of our design.

I think it would be beneficial for the review process if you'd split this PR into several PRs which would be both easier to review and have better scoped per commit functional changes. It's specifically hard/impossible to see the diff that happened with the stuff you moved around, for instance I'm not sure for dma code did you copy paste it 100% or also changed something minor.

Here's a suggestion how you could separate your PR into several ones:
- Introduce DeviceProtocol class, with commented out full interface (empty class), you can tag along the specific protocol classes as well, without any protocol specific functions.
- A series of PRs where you transfer one group of functions in each of those PRs from having implementation in tt_device / remote_tt_device/ wormhole_device to protocol implementation. Probably around 5 PRs with this types of changes. As you're doing it, the remote and non-remote class end up being more and more similar. If they both just have to do device_protocol->read, then you don't need an override in the remote class for example.
- One or two PRs with some additional functional changes which I'm missing at the moment if they're not covered with previous PRs. Example, maybe during previous PRs   you changed tt_device to tt_device->get_device_protocol() in RemoteCommunicationLegacyFirmware, but now you just use device_protocol in it, so a valid PR is to change the interface to start using device_protocol
- PRs with leftover refactoring, renames etc. 
- Removing tt_device specific implementations if they just pass through to underlying protocol, if we chose to do that.

I really think this would greatly help, but won't block on it.

device/api/umd/device/tt_device/protocol/pcie_protocol.hpp
```diff
@@ -0,0 +1,108 @@
...
+    PcieProtocol() = delete;
+
+    /* DeviceProtocol */
+    virtual ~PcieProtocol() = default;
+
+    void write_to_device(const void *mem_ptr, tt_xy_pair core, uint64_t addr, uint32_t size) override;
+
+    void read_from_device(void *mem_ptr, tt_xy_pair core, uint64_t addr, uint32_t size) override;
+
+    /* MmioProtocol */
```
pjanevskiTT: I guess we should remove this comment? We have just DeviceProtocol and PcieInterface, not sure what MmioProtocol represents here

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -335,6 +339,13 @@ class TTDevice {
      */
     virtual void dma_multicast_write(void *src, size_t size, tt_xy_pair core_start, tt_xy_pair core_end, uint64_t addr);
 
+    PcieInterface *get_pcie_interface();
```
pjanevskiTT: not sure how much I like this. That is why maybe I was more in favor of having all functions in DeviceProtocol and then being able to query DeviceProtocol for ability to do certain operations. My motivation was a use case of adding another protocol, then we would need to expand this and user code as well, where with having all functions in DeviceProtocol, you could simply implement new class and don't change TTDevice and/or user code

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -86,7 +92,7 @@ class TTDevice {
      * @param size number of bytes
      * @throws std::runtime_error if the DMA transfer fails
      */
-    virtual void dma_d2h(void *dst, uint32_t src, size_t size) = 0;
+    void dma_d2h(void *dst, uint32_t src, size_t size);
```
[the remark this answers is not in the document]
pjanevskiTT: +1 on this comment, I understood we are not going to keep this API around TTDevice layer anymore

pjanevskiTT (changes requested):

device/api/umd/device/tt_device/protocol/device_protocol.hpp
```diff
@@ -0,0 +1,34 @@
...
+    virtual void read_from_device(void* mem_ptr, tt_xy_pair core, uint64_t addr, uint32_t size) = 0;
+
+    virtual tt::ARCH get_arch() = 0;
+    virtual architecture_implementation* get_architecture_implementation() = 0;
+
+    virtual int get_communication_device_id() const = 0;
+    virtual IODeviceType get_communication_device_type() = 0;
+
+    virtual void detect_hang_read(uint32_t data_read = HANG_READ_VALUE) = 0;
+    virtual bool is_hardware_hung() = 0;
```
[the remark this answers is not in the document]
nbuncicTT: reply same as above

device/tt_device/tt_device.cpp
```diff
@@ -130,109 +150,65 @@ TTDeviceInitResult TTDevice::init_tt_device(const std::chrono::milliseconds time
...
-    if (communication_device_type_ == IODeviceType::JTAG) {
-        // Jtag protocol uses different communication paths from pci therefore
-        // there's no need to check hang which is in this case pci-specific.
-        return;
-    }
-    if (data_read == HANG_READ_VALUE && is_hardware_hung()) {
-        throw std::runtime_error("Read 0xffffffff from PCIE: you should reset the board.");
+PcieInterface *TTDevice::get_pcie_interface() {
+    if (pcie_capabilities_ == nullptr) {
+        throw std::runtime_error("TTDevice was built with a non-PCIe protocol.");
```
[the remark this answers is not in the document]
nbuncicTT: So this decision is a product of discussion with @aleksamarkovicTT. He is greatly opposed to getting a segmentation fault if some protocol isn't used - and I don't have a strong opinion here, and for me it's not that important if it's an exception or a nullptr, although, I think you can overcome the obstacle with a try catch block here instead of an if else that you have previewed.

nbuncicTT:
Closed because `DeviceProtocol` was merged via other smaller PRs.

PR #2124, opened 2026-02-18
Move warm reset tests to separate file

pjanevskiTT:
### Issue

Consolidate tests such that test_cluster is not piled with all tests

### Description

Move all warm reset tests to separate file.

### List of the changes

- Move warm reset tests to separate file

### Testing
CI

### API Changes
/

### DIFF
--- a/tests/api/test_warm_reset.cpp
+++ b/tests/api/test_warm_reset.cpp
@@ -57,10 +57,6 @@ bool is_ipmitool_ready() {
     return true;
 }
 
-// ============================================================================
-// TTDevice Warm Reset Tests (moved from test_tt_device.cpp)
-// ============================================================================
-
 TEST(WarmResetTest, DISABLED_TTDeviceWarmResetAfterNocHang) {
     std::vector<int> pci_device_ids = PCIDevice::enumerate_devices();
     if (pci_device_ids.empty()) {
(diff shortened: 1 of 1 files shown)


tests/api/test_warm_reset.cpp
```diff
@@ -0,0 +1,777 @@
...
+
+    if (system("timeout 2 ipmitool power status > /dev/null 2>&1") != 0) {
+        std::cout << "ipmitool power status command failed." << std::endl;
+        return false;
+    }
+
+    return true;
+}
+
+// ============================================================================
```
nbuncicTT: nit: Remove the comment, I think it's irrelevant

tests/api/test_warm_reset.cpp
```diff
@@ -0,0 +1,777 @@
...
+    WarmReset::warm_reset();
+
+    for (pid_t p : pids) {
+        int status;
+        waitpid(p, &status, 0);
+        EXPECT_EQ(WEXITSTATUS(status), 0) << "Child process " << p << " failed.";
+    }
+}
+
+// ============================================================================
```
nbuncicTT: nit: Same as above, I think the moved tests are a part of warm reset tests and don't need to be distinguished by the file they came from.

PR #2128, opened 2026-02-19
Remove duplicate methods in semver_t

aleksamarkovicTT:
### Issue
/

### Description
Remove duplicate methods in semver_t

### List of the changes
- Removed `semver_t::from_eth_fw_tag()`, identical methods exists under different name.
- Removed unused fw bundle tag constructor, better to use explicit static method.

### Testing
CI

### API Changes
There are API changes in this PR.

### DIFF
--- a/device/api/umd/device/utils/semver.hpp
+++ b/device/api/umd/device/utils/semver.hpp
@@ -26,9 +26,6 @@ public:
 
     constexpr semver_t() : major(0), minor(0), patch(0), pre_release(0) {}
 
-    constexpr semver_t(std::uint32_t version) :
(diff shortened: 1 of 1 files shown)


device/api/umd/device/utils/semver.hpp
```diff
@@ -26,9 +26,6 @@ class semver_t {
 
     constexpr semver_t() : major(0), minor(0), patch(0), pre_release(0) {}
 
-    constexpr semver_t(std::uint32_t version) :
```
nbuncicTT: Why was this added in the first place? Nice cleanup.

PR #2131, opened 2026-02-19
Add MultiArchTopologyDiscovery for heterogeneous device systems

aperezvicente-TT:
Introduces a new topology discovery class that supports systems with multiple Tenstorrent architectures (e.g., Wormhole + Blackhole) present simultaneously on the PCI bus.

The existing TopologyDiscovery::discover() validates all devices under a single architecture assumption. On mixed-arch nodes this causes failures because validation logic conflicts across architectures. This change works around that by running discovery in per-architecture isolated contexts using TT_VISIBLE_DEVICES filtering.

Changes:
- device/topology/multi_arch_topology_discovery.cpp: implementation
- device/api/umd/device/topology/multi_arch_topology_discovery.hpp: public API
- device/CMakeLists.txt: register new source and header in build

API surface:
  MultiArchTopologyDiscovery::discover_by_architecture()  - discover all archs MultiArchTopologyDiscovery::discover_single_architecture() - target one arch MultiArchTopologyDiscovery::get_available_architectures() - PCI enumeration

### Issue
(Link to Github issue(s))

### Description
(Add freeform description.)

### List of the changes

### Testing

### API Changes

### DIFF
--- a/device/CMakeLists.txt
+++ b/device/CMakeLists.txt
@@ -108,6 +108,7 @@ target_sources(
         topology/topology_discovery.cpp
         topology/topology_discovery_blackhole.cpp
         topology/topology_discovery_wormhole.cpp
+        topology/multi_arch_topology_discovery.cpp
         tt_device/remote_communication.cpp
         tt_device/remote_communication_legacy_firmware.cpp
         jtag/jtag_device.cpp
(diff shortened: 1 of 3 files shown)


aleksamarkovicTT (changes requested):
Thanks for your contribution.
We've been planning on introducing this feature in a different way - this seems like a lot more work than what we had in mind.
My idea was to introduce a preferred architecture in TopologyDiscovery options. The user could select a specific architecture and get a ClusterDescriptor for it. The user could thus create two independent clusters and manage them accordingly.

One of the reasons why this was not implemented yet is that we do not have heterogeneous machines in our CI pools or lab machines. Until we can actually test this functionality, I'd prefer leaving it out. Reach out to #tt-umd if you have such a machine.

Support for actual heterogeneous clusters is not something we've been planning to do, but I can imagine it being implemented with some caveats.

PR #2133, opened 2026-02-20
Added Fan RPM Support to FirmwareInfoProvider 

btrzynadlowski-tt:
### Issue                                                                                                                                                                                           
N/A           
                                                                                                                                                                                                      
  ### Description                                                                                                                                                                                     
  This PR adds support for reading actual fan RPM values from firmware telemetry, separate from the existing fan speed percentage metric. Previously, `get_fan_speed()` was incorrectly documented as
  returning RPM when it actually returns a percentage (0-100). This has been clarified in the documentation, and a new `get_fan_rpm()` method has been added that reads from the previously unused
  `TelemetryTag::FAN_RPM` (tag 41).

  The implementation follows the same pattern as other ARC telemetry metrics and includes backward compatibility for older firmware versions where FAN_RPM telemetry is not available.

  ### List of the changes
  - Added `get_fan_rpm()` method to `FirmwareInfoProvider` base class that reads from `TelemetryTag::FAN_RPM`
  - Added `fan_rpm_available` member variable to track availability of FAN_RPM telemetry
  - Implemented `get_fan_rpm()` override in `Wormhole_18_3_FirmwareInfoProvider` to return `std::nullopt` (FAN_RPM not available in legacy firmware ≤18.3)
  - Updated `get_fan_speed()` documentation comment to clarify it returns percentage (0-100), not RPM
  - Added proper handling of 0xFFFFFFFF sentinel value in `get_fan_rpm()` to indicate fans not present/controllable

  ### Testing
  - Existing CI tests should pass (no breaking changes to existing API)
  - Manual testing recommended on devices with firmware ≥18.4 to verify FAN_RPM telemetry is read correctly
  - Manual testing on Wormhole devices with firmware ≤18.3 to verify `get_fan_rpm()` returns `std::nullopt` gracefully

  ### API Changes
  This PR has API changes:
  - Added new `get_fan_rpm()` virtual method to `FirmwareInfoProvider` (non-breaking addition)
  - Updated documentation comment for `get_fan_speed()` to clarify return value semantics (non-breaking)


device/api/umd/device/firmware/firmware_info_provider.hpp
```diff
@@ -86,11 +86,17 @@ class FirmwareInfoProvider {
...
+     * Get fan speed as a percentage (0-100), if fans are present and controllable by firmware.
+     * @returns Fan speed [percent]
      */
     virtual std::optional<uint32_t> get_fan_speed() const;
 
+    /*
+     * Get fan speed in RPM, if fans are present and controllable by firmware.
+     * @returns Fan speed [RPM]
+     */
+    virtual std::optional<uint32_t> get_fan_rpm() const;
```
nbuncicTT: This change seems fine, but I'm just wondering, should we add a test to it? @pjanevskiTT do we have and FirmwareInfoProvider tests? If not, should we add them?

btrzynadlowski-tt: Do we know which firmware versions have support for the fan? My understanding is that many of our systems don't have fans, so I'm not sure if there is a clear way to determine this ahead of time (would be nice to have is_supported() type functions so I don't even need to create this metric when it's not there but not strictly necessary). 

Not sure how to resolve this.

aleksamarkovicTT:
We'll wait for #1807 to be merged before we merge this one.

PR #2137, opened 2026-02-21
Select architecture in TopologyDiscovery for multi-arch hosts

aleksamarkovicTT:
### Issue
#2205

### Description
This PR resolves an issue for hosts with multiple cards of differing architectures by converging to the first observed architecture. Also support for selecting a preferred architecture for hosts with devices of different architectures.

To clear any confusion, this PR **does not** introduce support for any kind of heterogeneous cluster. It simply avoids crashes caused by UMD trying to create a heterogeneous cluster.

### List of the changes
- Added options.architecture.
- Introduce `determine_architectures()` to scan host for architectures.
- In case of no arch. specified, `determine_architectures()` is called.
- Filtering devices per selected architecture.
- Specifying architecture using `TT_ARCH` env. var.

### Testing
Manually verified on sjc-lab-t7003 (KMD dev machine).

### API Changes
There are no API changes in this PR


aperezvicente-TT:
Do we have a new PR for this?

PR #2138, opened 2026-02-23
Implement system memory access from device for TTSim

pjanevskiTT:
### Issue

/

### Description

Add support for accessing host system memory from device for TTSim. This mechanism is used by FastDispatch and this PR is work towards supporting FD in TTSim.

### List of the changes

- Add support for setting callbacks for read/write to system memory from device
- Implement those callbacks
- Add test

### Testing
CI + local testing

### API Changes
/


pjanevskiTT:
https://github.com/tenstorrent/ttsim-private/pull/282 matching TTSim PR

Issue #1466, opened 2025-10-15
Expose RTLSim in TTDevice layer


broskoTT:
This is a requirement for exalens to switch to this layer

PR #2171, opened 2026-02-24
Implement RtlSimCommunicator class

pjanevskiTT:
### Issue

Code for communicating with remote machine for RTL sim was all over the code, basically working towards #1466 

### Description

Implement RtlSimCommunicator class. It should encapsulate the communication with remote machine where RTL sim is actually running. The point of this class is to not copy all of the code for communication with simulator when we want to move the logic to TTDevice layer. 

### List of the changes

- Implement RtlSimCommunicator
- Move all the code related to communication to this class
- Make RtlSimulationChip use RtlSimCommunicator

### Testing
CI

### API Changes
/


aleksamarkovicTT (approved):
I don't know this part of UMD in detail but design-wise the refactor looks good.

PR #2187, opened 2026-02-27
Fix CMake targets

RossComputerGuy:
### Issue
(Link to Github issue(s))

### Description

Fixes various CMake issues which prevent tt-umd from configuring and building correctly for Nix.

### List of the changes

- Fixes `hwloc.h` not found
- Fixes `tt-logger::tt-logger` target not found
- Fixes `spdlog::spdlog_header_only` target not found
- Fixes `fmt::fmt-header-only` target not found
- Fixes CMake trying to do `-ltt-logger`
- Fixes CMake trying to do `-tcxxopts`

### Testing

Try building `tt-umd` from https://github.com/NixOS/nixpkgs/pull/494239 via `nix-build -A tt-umd` after cloning the PR.

### API Changes

### DIFF
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -43,6 +43,7 @@ CPMAddPackage(
         "YAML_CPP_BUILD_TESTS OFF"
         "YAML_CPP_BUILD_TOOLS OFF"
         "YAML_BUILD_SHARED_LIBS OFF"
+    FIND_PACKAGE_ARGUMENTS GLOBAL
 )
 
 if(yaml-cpp_ADDED)
@@ -146,7 +147,13 @@ CPMAddPackage(
 ####################################################################################################################
 # tt-logger
 ####################################################################################################################
-CPMAddPackage(NAME tt-logger GITHUB_REPOSITORY tenstorrent/tt-logger VERSION 1.1.8 SYSTEM YES)
+CPMAddPackage(
+    NAME tt-logger
+    GITHUB_REPOSITORY tenstorrent/tt-logger
+    VERSION 1.1.8
+    SYSTEM YES
+    FIND_PACKAGE_ARGUMENTS GLOBAL
+)
 
 ####################################################################################################################
 # cxxopts
@@ -159,6 +166,7 @@ CPMAddPackage(
     SYSTEM YES
     OPTIONS
         "CMAKE_MESSAGE_LOG_LEVEL NOTICE"
+    FIND_PACKAGE_ARGUMENTS GLOBAL
 )
 
 ####################################################################################################################


third_party/CMakeLists.txt
```diff
@@ -145,20 +145,34 @@ CPMAddPackage(
 ####################################################################################################################
 # tt-logger
 ####################################################################################################################
-CPMAddPackage(NAME tt-logger GITHUB_REPOSITORY tenstorrent/tt-logger VERSION 1.1.8 SYSTEM YES)
+# CPMAddPackage doesn't support "GLOBAL"
+find_package(tt-logger GLOBAL)
+
+if(NOT tt-logger_FOUND)
```
nbuncicTT: So I have started the CI to see how your changes affect the repo and it breaks our builds (which are not sandboxed). 

I'll propose a solution here, but maybe there's a better one.

So, revert the changes that delete `tt-logger::tt-logger` from `target_link_libraries` and instead add `target_include_directories` (this also goes for cxxopts) - and let's make the GLOBAL part manually but to not break the non-sandbox builds in the following manner:

After the line:
`CPMAddPackage(NAME tt-logger GITHUB_REPOSITORY tenstorrent/tt-logger VERSION 1.1.8 SYSTEM YES)`

Add this:
```
# Check if the target exists
if(TARGET tt-logger::tt-logger)
    # Check if CPM fetched the package from github or locally
    get_target_property(is_tt_logger_imported tt-logger::tt-logger IMPORTED)
    if(is_tt_logger_imported)
        # For local fetching set the property to "GLOBAL"
        set_property(TARGET tt-logger::tt-logger PROPERTY IMPORTED_GLOBAL TRUE)
    endif()
endif()
```

Do a similar thing for cxxopts.

I'd test this thing myself, but I don't have the NixOS setup to test this (or any other sandbox build approach for that matter).

blozano-tt (changes requested):
@RossComputerGuy thank you for trying to improve our CMake infra with Nix support.

At the moment I am taking issue with the direction of these changes.

First, I question the assertion that CPMAddPackage doesn't support GLOBAL for find_package. It didn't make sense to me, so I asked my AI friend @Claude to create a test repo to demonstrate that it works. Please take a look: https://github.com/blozano-tt/test-cpm-find-package-global - its possible its incorrect, but I don't see why it can't just forward arguments to find_package.

If the AI finding is correct, then the premise of this PR is invalid.

Likely some of your findings are correct.

I don't like moving us away from using CMake targets for linking. I'd prefer that we continue using modern idiomatic CMake as much as possible.

I would suggest rechecking your original assertion about CPM, and try fixing these issues one at a time.

Again, thanks for your time on this.

blozano-tt:
@RossComputerGuy 

Any interest in testing this Claude Opus generated nix PR?

https://github.com/tenstorrent/tt-umd/pull/2429/files

broskoTT:
@RossComputerGuy is this PR outdated, or needs rework, or waiting for testing of Bryan's solution? It is not clear to me from the last few messages what is the current state of it

RossComputerGuy:
Reworked after git conflicts, still necessary

device/CMakeLists.txt
```diff
@@ -298,8 +298,8 @@ target_link_libraries(
         umd::Firmware
         hwloc
         rt
-        spdlog::spdlog_header_only
-        fmt::fmt-header-only
+        spdlog
+        fmt
```
blozano-tt: I mentioned in some other PR, that we would like to stay header only on fmt. 

Is there no nix package for fmt::fmt-header-only ? 

Is it possible for you to supply a patch to the nix recipe? I think you are doing that now.

RossComputerGuy: There's no `fmt::fmt-header-only` with Nix. But I'm going to redo this PR after #2188 merges.

third_party/CMakeLists.txt
```diff
@@ -33,17 +33,32 @@ endif()
...
-    GIT_TAG
-        2f86d13775d119edbb69af52e5f566fd65c6953b # 0.8.0 + patches
-    SYSTEM YES
-    OPTIONS
-        "YAML_CPP_BUILD_TESTS OFF"
-        "YAML_CPP_BUILD_TOOLS OFF"
-        "YAML_BUILD_SHARED_LIBS OFF"
-)
+
+# CPMAddPackage doesn't support "GLOBAL"
```
blozano-tt: I don't follow ... 

CPMAddPackage adds things globally by default

device/CMakeLists.txt
blozano-tt: **1. spdlog / fmt target names are probably wrong** (device/CMakeLists.txt)cmake
- spdlog::spdlog_header_only
- fmt::fmt-header-only
+ spdlog
+ fmt
Plain (non-namespaced) targets fall back to -lspdlog / -lfmt linker flags when no such CMake target exists — which is one of the things this PR claims to fix. The correct targets from find_package(spdlog) / find_package(fmt) are spdlog::spdlog and fmt::fmt. Unless there's a specific reason Nix's packages expose bare spdlog / fmt targets, these should be namespaced.

third_party/CMakeLists.txt
```diff
@@ -145,20 +160,50 @@ CPMAddPackage(
...
-CPMAddPackage(
-    NAME cxxopts
-    GITHUB_REPOSITORY jarro2783/cxxopts
-    GIT_TAG
-        dbf4c6a66816f6c3872b46cc6af119ad227e04e1 #version 3.2.1 + patches
-    SYSTEM YES
-    OPTIONS
-        "CMAKE_MESSAGE_LOG_LEVEL NOTICE"
-)
+
```
blozano-tt: Should be able to use FIND_PACKAGE_ARGUMENTS GLOBAL

RossComputerGuy:
Finally got the time to redo this PR, it works. Noticed that nanobind isn't a problem for us in nixpkgs anymore so we'll have tt-umd packaged soon I'd say.

broskoTT (approved):

blozano-tt (approved):

PR #2202, opened 2026-03-01
Add uv support

lu-zero:


CMakeLists.txt
```diff
@@ -122,11 +122,10 @@ if(TT_UMD_ENABLE_CLANG_TIDY)
     endif()
 endif()
 
-# Enable all warnings and treat them as errors
+# Enable all warnings but be more permissive
 add_compile_options(
     "$<$<STREQUAL:${CMAKE_SYSTEM_PROCESSOR},x86_64>:-march=x86-64-v3>"
     -Wall
-    -Werror
```
aleksamarkovicTT: Thank you for the contribution.

I agree with all the changes except this one (and the one in .clang-tidy). Which warnings were you facing when building UMD with these newer compilers?

Issue #2199, opened 2026-02-27
[Cluster Stability] SW Support for DRAM Retraining


tt-asaigal:
## Motivation

A significant source of instability for BH Galaxy clusters is DRAM training failing after `tt-smi` resets are issued.

Today this shows up as the following error message `DRAM training failed`. The goal is to handle this more gracefully by trying to recover the DRAM controllers, instead of instantly killing the workload.

## Integrating CMFW DRAM Retraining Functionality from Syseng into UMD

Please see the following tickets for some background on the work Syseng has done:
- https://github.com/tenstorrent/tt-metal/issues/37941
- https://tenstorrent.atlassian.net/browse/SYS-3226

The idea is to use the ARC Mailbox to issue the reset request (please see Tyler's comments regarding the arguments that need to be passed in and how we can check status).

I think `TTDevice` can have a dedicated `retrain_dram_controller` API, which takes in an index argument. The logic inside `TTDevice::wait_dram_channel_training` can then handle directed retrains for this condition:

```
if (dram_training_status.at(dram_channel) == DramTrainingStatus::FAIL) {
         throw std::runtime_error("DRAM training failed");
}
```

We can have a scheme where we try to recover the controller up to _N_ times when a failure is detected (_N_ can be provided by Syseng), instead of failing immediately.

Assigning to @broskoTT @pjanevskiTT - feel free to delegate as needed.

fyi @cfjchu @fahmedchanTT

tt-asaigal:
I think it would be good for the DRAM retraining logic to be entirely contained inside UMD. The reason for this is that recovering DRAM does not require any multi-host communication or coordination.

I see this as something thats very similar to de-asserting core level resets. UMD on each host does this independently and user SW just gets functional cores.

We had more complexity with ethernet, since we have multi-host and topology aware SW outside of UMD that does cross host link re-training, etc. This will not be the case with DRAM retraining.

tldr: I think its best to start with UMD handling this automatically under the hood during discovery. Its okay to expose an API, though I don't see a use-case for it at the moment.

broskoTT:
Pavle to look again and to add missing Bist failure support

broskoTT:
Pending review

pjanevskiTT:
closed by #2768

PR #2220, opened 2026-03-03
Support retrain of DRAM cores

pjanevskiTT:
### Issue

#2199 

### Description

Add support for retraining DRAM cores on Blackhole. When DRAM training fails, run retrain mechanism which sends ARC message to retrain specific DRAM core. This is to avoid simply throwing an exception when DRAM training fails for the first time.

### List of the changes

- Add retrain_dram_core function
- Call it N number of times if DRAM training fails 

### Testing

Not really tested. I ran on p150 machine to confirm we can send a message, not really sure how to get a card into a state where DRAM training fails

### API Changes
/

### DIFF
--- a/device/tt_device/wormhole_tt_device.cpp
+++ b/device/tt_device/wormhole_tt_device.cpp
@@ -609,6 +609,8 @@ bool WormholeTTDevice::is_hardware_hung() {
     return (scratch_data == HANG_READ_VALUE);
 }
 
-void WormholeTTDevice::retrain_dram_core(const uint32_t dram_channel) {}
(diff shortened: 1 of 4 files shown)


device/tt_device/wormhole_tt_device.cpp
```diff
@@ -624,4 +624,6 @@ bool WormholeTTDevice::is_hardware_hung() {
     return (scratch_data == HANG_READ_VALUE);
 }
 
+void WormholeTTDevice::retrain_dram_core(const uint32_t dram_channel) {}
```
aleksamarkovicTT: nit: Throw if not supported?

PR #2221, opened 2026-03-03
Clang tidy warnings suppression

lu-zero:
Here the warnings that trigger over the codebase. the patch body has the list with the occurrences.

https://gist.github.com/lu-zero/fa9cab21a385384358d55ccae94662fe has the cuter list, not sure if I'm doing something wrong, but that's what I'm seeing with clang-20 as well as with clang-21 and clang-22


.clang-tidy
```diff
@@ -16,7 +16,21 @@ Checks: >
     -bugprone-suspicious-stringview-data-usage,
     -bugprone-switch-missing-default-case,
     -bugprone-unchecked-optional-access,
+    -bugprone-shadow,
```
nbuncicTT: The bugprone checks are very useful for catching bugs and we see value in them, so I'd rather not omit them.

nbuncicTT: As most of these checks are are introduced in clang 21 and 22, I'm okay omitting them for now, and we'll introduce fixes for them in the future.

Thanks for the help.

To be sure of this change, @blozano-tt can you take a look at these if they are ok to omit for now? Our driver doesn't follow these rules (and I think neither metal does).

blozano-tt: Nice, I did not know there was a clang-tidy check for this. I knew there was a compiler warning.

PR #2236, opened 2026-03-06
Report telemetry from all devices in telemetry tool

pjanevskiTT:
## Description
Update the telemetry tool to discover and report telemetry from all devices in the system, including remote (ETH-connected) chips, not just PCIe-connected ones.

Users who need to restrict which devices are visible can use the `TT_VISIBLE_DEVICES` environment variable, which is already respected by the PCIe enumeration layer inside topology discovery.

## List of the changes

**Use `TopologyDiscovery::discover()` instead of `PCIDevice::enumerate_devices()`:**
- Replaces direct PCIe device enumeration with topology discovery, which also finds ETH-connected remote chips.
- Devices are now identified by logical `ChipId` instead of PCIe device ID.

**Remove `--devices` CLI option:**
- The option for filtering by PCIe device IDs is removed. Use `TT_VISIBLE_DEVICES` to control which devices are visible.

## Testing
Manual

## API Changes
/

### DIFF
--- a/tools/telemetry.cpp
+++ b/tools/telemetry.cpp
@@ -33,7 +33,7 @@
 
 using namespace tt::umd;
 
-std::string run_default_telemetry(tt::ChipId chip_id, FirmwareInfoProvider* firmware_info_provider, tt::ARCH arch) {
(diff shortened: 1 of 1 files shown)


aleksamarkovicTT (approved):
looks good, just apply copilot suggestions

Issue #1688, opened 2025-12-04
Support KMD power API


joelsmithTT:
With the upcoming release of KMD 2.6.0, the driver introduces a new power management policy that automatically lowers the device's idle power. To ensure backward compatibility, the driver treats any application that opens the device without the `O_APPEND` flag as a "legacy" client, defaulting it to a high-power state (enabling Tensix, L2CPU, etc.).

Currently, the UMD opens the device in this legacy mode. This ensures safety but prevents the device from entering a low-power idle state as long as the application is connected, even if it is not actively computing.

## Proposed Solution
Update the UMD to explicitly manage device power states.

1.  **Opt-out of Legacy Mode:** Pass `O_APPEND` to `open()` to signal the KMD that the UMD is power-aware.
2.  **Explicit Power Requests:** The UMD must use `TENSTORRENT_IOCTL_SET_POWER_STATE` to explicitly request the power domains required for the current workload.
    *   **Idle:** When no workloads are running, release power flags to allow the device to idle.
    *   **Active:** Before dispatching work to Tensix, L2CPU, or accessing memory, explicitly enable the necessary domains.
    *   **Safety:** Failure to enable the correct domains before access may result in hardware hangs or undefined behavior.

## Benefits
- **Idle Efficiency:** Long-running applications holding a device handle but not performing compute (e.g., telemetry/health daemon) will allow the device to power down when inactive.
- **Granular Control:** Allows for more fine-grained power management strategies in the future.

## Strategic Context
As tooling migrates from Luwen to UMD, the UMD must support diverse power profiles:
- **Monitoring Tools (e.g., `tt-smi`):** Require device access without forcing high-power states to ensure accurate readings and efficiency.
- **Diagnostics (e.g., `tt-burnin`) & ML Workloads:** Require explicit high-power states to ensure performance and stability.

## Environment
- **Driver:** KMD 2.6.0+
- **Device:** Blackhole

PR #2252, opened 2026-03-10
Support KMD power API

pjanevskiTT:
### Issue
#1688

### Description
KMD 2.6.0 introduced automatic idle power reduction but treats applications that open devices without `O_APPEND` as "legacy" clients, forcing them into high-power states.

This PR opts UMD into the new KMD power management for Blackhole devices, adding explicit power domain control via the new `TENSTORRENT_IOCTL_SET_POWER_STATE` ioctl. By default, devices open in high-power (legacy) mode to preserve backward compatibility. A new `low_power` option in `TopologyDiscoveryOptions` allows callers to opt into low-power mode for discovery workloads that don't need full power.

### List of the changes
- Add `TENSTORRENT_IOCTL_SET_POWER_STATE` and `tenstorrent_power_state` struct with `TT_POWER_FLAG_*` / `TT_POWER_VALIDITY_*` defines (synced from KMD 2.6.0).
- Add `KMD_POWER_STATE = SemVer(2, 6, 0)` sentinel to `kmd_versions.hpp`.
- Add `PCIDevice::set_power_state(bool busy)` to issue the new ioctl; restricted to Blackhole, no-ops on other architectures and KMD < 2.6.0.
- Add `TTDevice::set_power_state(bool busy)` as a public virtual method (overridden by `PCIDevice`).
- Add `low_power` option to `TopologyDiscoveryOptions` (default `false`). When `true`, devices are opened with `O_APPEND` to opt out of KMD legacy mode on Blackhole with KMD >= 2.6.0.
- Thread `low_power` flag through `TTDevice::create` → `PCIDevice` constructor → `open_pci_device` and `tt_device_open`.
- Add `extra_flags` parameter to `tt_device_open()` in tt_kmd_lib.
- Use `open_pci_device()` for `read_command_byte()` instead of raw `open()`.
- Add `log_warning` when `set_power_state` is called on unsupported KMD version.
- Remove `set_power_state(true)` call from `LocalChip::start_device()` — not needed since devices open in high-power (legacy) mode by default.
- Add test verifying TopologyDiscovery produces equivalent results in both power modes.

### Testing
CI + new `TopologyDiscoveryPowerModes` test

### API Changes
- `PCIDevice::set_power_state(bool busy)` added as a public method.
- `TTDevice::set_power_state(bool busy)` added as a public virtual method.
- `PCIDevice` constructor now accepts optional `bool low_power = false`.
- `TTDevice::create()` now accepts optional `bool low_power = false`.
- `tt_device_open()` now takes an `int extra_flags` parameter.
- `TopologyDiscoveryOptions::low_power` added (default `false`).

### DIFF
--- a/device/pcie/pci_device.cpp
+++ b/device/pcie/pci_device.cpp
@@ -1032,8 +1032,8 @@ void PCIDevice::set_power_state(bool busy) {
     power_state.validity = TT_POWER_VALIDITY(4, 0);
 
     if (busy) {
-        power_state.power_flags = TT_POWER_FLAG_MAX_AI_CLK | TT_POWER_FLAG_MRISC_PHY_WAKEUP |
-                                  TT_POWER_FLAG_TENSIX_ENABLE | TT_POWER_FLAG_L2CPU_ENABLE;
+        power_state.power_flags =
+            TT_POWER_FLAG_MRISC_PHY_WAKEUP | TT_POWER_FLAG_TENSIX_ENABLE | TT_POWER_FLAG_L2CPU_ENABLE;
     } else {
         power_state.power_flags = 0;
     }


tests/api/test_cluster_descriptor.cpp
```diff
@@ -323,3 +324,32 @@ TEST(TestClusterDescriptor, TestClusterLogicalETHChannelsConnectivity) {
...
+    default_options.discover_remote_devices = false;
+    auto [desc_default, devices_default] = TopologyDiscovery::discover(default_options);
+
+    // Low-power mode.
+    TopologyDiscoveryOptions low_power_options;
+    low_power_options.discover_remote_devices = false;
+    low_power_options.low_power = true;
+    auto [desc_low_power, devices_low_power] = TopologyDiscovery::discover(low_power_options);
+
+    // Both modes should discover the same number of chips.
```
broskoTT: Not sure if this guards anything...
Anyway, please consider adding a telemetry readout and verify that the power reported is smaller?

device/pcie/pci_device.cpp
```diff
@@ -386,10 +386,19 @@ std::optional<int> PCIDevice::get_pci_device_id(int umd_logical_id) {
     return enumerated_ids[umd_logical_id];
 }
 
+static int open_pci_device(const std::string &device_path) {
+    // O_APPEND opts out of legacy mode in KMD >= 2.6.0, allowing the device to enter low-power idle states.
+    int flags = O_RDWR | O_CLOEXEC;
+    if (PCIDevice::read_kmd_version() >= KMD_POWER_STATE) {
```
broskoTT: consider adding a log debug, which prints out "Opening PCI device with flags {} in {low/high} power mode"

PR #2256, opened 2026-03-10
Refactor is_board_id_included() in TopologyDiscovery

aleksamarkovicTT:
### Issue
/

### Description
Removed unused `board_type` argument in `is_board_id_included()`. This argument was used to support 4U/TG. Since we deprecated support for it, it has been unused.

### List of the changes
- Removed `board_type` argument from `is_board_id_included()`.
- Removed now unused `get_remote_board_type()`.

### Testing
CI

### API Changes

### DIFF
--- a/device/api/umd/device/topology/topology_discovery.hpp
+++ b/device/api/umd/device/topology/topology_discovery.hpp
@@ -56,18 +56,12 @@ protected:
     virtual void wait_eth_cores_training(
         TTDevice* tt_device, std::chrono::milliseconds timeout_ms = timeout::ETH_TRAINING_TIMEOUT);
 
-    // board_type is not used for all configs.
-    virtual bool does_board_belong_to_host(uint64_t board_id) const;
+    virtual bool is_board_id_included(uint64_t board_id) const;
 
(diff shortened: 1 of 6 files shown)


device/api/umd/device/topology/topology_discovery.hpp
```diff
@@ -57,8 +57,7 @@ class TopologyDiscovery {
         TTDevice* tt_device, std::chrono::milliseconds timeout_ms = timeout::ETH_TRAINING_TIMEOUT);
 
     // board_type is not used for all configs.
-    // We need to know that we are seeing TG board and that we should include it in the topology.
-    virtual bool is_board_id_included(uint64_t board_id, uint64_t board_type) const = 0;
+    virtual bool does_board_belong_to_host(uint64_t board_id) const;
```
pjanevskiTT: we should rename this function. It's not just used for all boards on the host. It is also used when we want to apply TT_VISIBLE_DEVICES and run topology discovery. Then we also don't want to take some boards, even if they are physically connected to the host. So you should revert this rename

Issue #2146, opened 2026-02-24
Fast Dispatch ttsim


broskoTT:
Support FD in ttsim simulation.
Some work already done in
https://github.com/tenstorrent/tt-umd/pull/2138

pjanevskiTT:
completed with all the changes merged in UMD and metal, final commit was https://github.com/tenstorrent/tt-metal/commit/999f99375f7800d0ce22a97d89397bcec82bbff5

PR #2278, opened 2026-03-12
Add changes to support running FastDispatch on TTSim

pjanevskiTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/2146

### Description
Adds initial changes to support running FastDispatch on the TTSim simulation device.

### List of the changes
- Expose `TLBManager` from `TTSimTTDevice` and override `get_tlb_manager()` in `TTSimChip`.
- Set `pcie_base_` in `SimulationSysmemManager` constructor using an explicit arch switch (matching `SiliconSysmemManager` pattern).
- Wire internal PCIe DMA read/write callbacks to `SimulationSysmemManager` for simulator-side sysmem access. The public `dma_d2h`/`dma_h2d` overrides still throw, as host-initiated DMA is not yet supported.
- Short-circuit `get_closest_mmio_capable_chip` to return 0 for simulation.

### Testing
CI

### API Changes
- Added public `TTSimTTDevice::get_tlb_manager()` method (used by `TTSimChip::get_tlb_manager()` override).

### DIFF
--- a/device/tt_device/rtl_simulation_tt_device.cpp
+++ b/device/tt_device/rtl_simulation_tt_device.cpp
@@ -45,7 +45,7 @@ RtlSimulationTTDevice::RtlSimulationTTDevice(
     simulator_directory_(simulator_directory),
     soc_descriptor_(std::move(soc_descriptor)),
     architecture_impl_(architecture_implementation::create(soc_descriptor_.arch)),
-    sysmem_manager_(std::make_unique<SimulationSysmemManager>(num_host_mem_channels)) {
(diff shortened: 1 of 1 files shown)


device/chip_helpers/simulation_sysmem_manager.cpp
```diff
@@ -21,7 +21,18 @@
 
 namespace tt::umd {
 
-SimulationSysmemManager::SimulationSysmemManager(uint32_t num_host_mem_channels) {
+SimulationSysmemManager::SimulationSysmemManager(uint32_t num_host_mem_channels, tt::ARCH arch) {
+    switch (arch) {
```
broskoTT: Should we put this in base class?

Issue #1441, opened 2025-10-09
Fast Dispatch RTL simulation


broskoTT:
Current simulator implementation doesn't support host memory.
There might be other blocking issues, to be investigated.
Simulator folks have provided us with a prototype of this.

pjanevskiTT:
we implemented the mechanism in #2311, will reopen if more thing pop up

PR #2311, opened 2026-03-17
Handle RTL sim sysmem callbacks

pjanevskiTT:
### Issue

Work towards #1441 

#### The notification thread

`RtlSimCommunicator` runs a **background notification thread** that continuously receives messages from the simulator. It demultiplexes them:

- **Regular command responses** (e.g. responses to read/write requests UMD initiated) → queued for the calling thread
- **AXI RAM notifications** (device-initiated memory access) → handled immediately in the notification thread via callbacks

#### How AXI RAM access works

When the simulated device executes a memory transaction targeting host memory (e.g. a DMA write to sysmem), the simulator can't perform it directly. Instead:

**Device writes to host memory (`AXI_RAM_WRITE_NOTIFICATION`):**
1. Simulator detects an AXI transaction targeting the host memory address range
2. Simulator sends a `AXI_RAM_WRITE_NOTIFICATION` message containing the target address, size, and data payload
3. UMD's notification thread receives it and calls the **write callback**
4. The callback translates the physical address to the correct channel + offset, then `memcpy`s the data into the mmap'd sysmem buffer

**Device reads from host memory (`AXI_RAM_READ_NOTIFICATION`):**
1. Simulator detects an AXI read targeting host memory and needs the data to proceed
2. Simulator sends a `AXI_RAM_READ_NOTIFICATION` with the address and size
3. UMD's notification thread receives it and calls the **read callback**
4. The callback reads the data from the mmap'd sysmem buffer
5. UMD sends a response message back with the data
6. Simulator receives the data and completes the AXI transaction

#### Address translation

The address in the notification is a NOC/physical address as the device sees it. Translation to host VA:

```
offset_from_base  = address - pcie_base
channel           = offset_from_base / 1GB
offset_in_channel = offset_from_base % 1GB
host_va           = hugepage_mapping[channel].mapping + offset_in_channel
```

#### Callback registration

The callbacks are `std::function` objects set on `RtlSimCommunicator` via `set_ram_callbacks()`. `RtlSimulationTTDevice` registers them in its constructor, capturing a pointer to its `SimulationSysmemManager`. This keeps the communicator layer decoupled from the memory management layer — the communicator doesn't know about sysmem, channels, or address translation; it just invokes the callbacks.

#### What we expect from the remote (simulator) side

1. The simulator must send `AXI_RAM_WRITE_NOTIFICATION` whenever the simulated device writes to an address in the host memory range (`pcie_base` to `pcie_base + total_sysmem_size`). The message must include the full data payload.
2. The simulator must send `AXI_RAM_READ_NOTIFICATION` whenever the simulated device reads from host memory, and **block the simulated AXI transaction** until it receives the response from UMD containing the actual data.
3. Both notification types use the FlatBuffers `DeviceRequestResponse` schema with the `address`, `size`, and `data` fields. The `core.x` field carries a `ram_idx` identifier.

### List of the changes
- Add `AXI_RAM_WRITE_NOTIFICATION` and `AXI_RAM_READ_NOTIFICATION` commands to the FlatBuffers schema.
- Add `recv_from_device(void**, int timeout_ms)` overload to `SimulationHost` for non-blocking receives with configurable timeout.
- Add a notification handler thread in `RtlSimCommunicator` that demuxes incoming messages into AXI RAM notifications vs regular command responses.
- Add `RamWriteCallback` and `RamReadCallback` types and `set_ram_callbacks()` on `RtlSimCommunicator`.
- `RtlSimulationTTDevice` registers callbacks that translate NOC addresses to sysmem channels and perform actual `memcpy` via `SimulationSysmemManager`.
- Add `get_pcie_base()` getter to `SysmemManager`.
- Refactor `tile_read_bytes` to use the command response queue instead of blocking directly on the socket.

### Testing
CI

### API Changes
- Add `get_pcie_base()` getter to `SysmemManager` (public API).
- Add `recv_from_device(void**, int timeout_ms)` overload to `SimulationHost`.

### DIFF
--- a/device/simulation/simulation_device.fbs
+++ b/device/simulation/simulation_device.fbs
@@ -12,6 +12,9 @@ enum DEVICE_COMMAND : byte {
     NEO_DM_RESET_ASSERT = 8,
     NEO_DM_RESET_DEASSERT = 9,
     // AXI RAM notifications (remote → host).
+    // TODO: The simulator release currently sends these as tags 8 and 9, which conflict with
+    // NEO_DM_RESET_ASSERT and NEO_DM_RESET_DEASSERT. The simulator must be updated to use 10 and 11
+    // to match these values before this can work end-to-end.
     AXI_RAM_WRITE_NOTIFICATION = 10,
     AXI_RAM_READ_NOTIFICATION = 11,
 }


device/simulation/rtl_sim_communicator.cpp
```diff
@@ -54,7 +54,24 @@ RtlSimCommunicator::RtlSimCommunicator(const std::filesystem::path &simulator_di
     }
 }
 
-RtlSimCommunicator::~RtlSimCommunicator() = default;
+RtlSimCommunicator::~RtlSimCommunicator() {
+    if (notification_thread_running_.load()) {
+        notification_thread_running_.store(false);
+        if (notification_thread_.joinable()) {
+            notification_thread_.join();
```
broskoTT: So we're trying here to gracefully signal the thread it should close. But the user passed us their custom callbacks. If something hangs there, this will also hang.
Instead, please consider other options to instead forcefully close the notification_thread

device/tt_device/rtl_simulation_tt_device.cpp
```diff
@@ -48,6 +48,29 @@ RtlSimulationTTDevice::RtlSimulationTTDevice(
     sysmem_manager_(std::make_unique<SimulationSysmemManager>(num_host_mem_channels, soc_descriptor_.arch)) {
     log_info(tt::LogEmulationDriver, "Instantiating RTL simulation TTDevice");
     arch = soc_descriptor_.arch;
+
+    // Register sysmem callbacks so the simulator can read/write host memory.
+    if (num_host_mem_channels > 0) {
```
broskoTT: Can you also add a test, or make an adjustment to an existing test which would show this behavior is working?
If there was already a test, then please add to PR description how you've observed a difference in test behavior through our CI before and after the change

device/api/umd/device/simulation/simulation_host.hpp
```diff
@@ -24,6 +24,7 @@ class SimulationHost {
     void start_host();
     void send_to_device(uint8_t *buf, size_t buf_size);
     size_t recv_from_device(void **data_ptr);
```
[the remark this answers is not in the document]
pjanevskiTT: changed to one function

device/tt_device/rtl_simulation_tt_device.cpp
pjanevskiTT: tested it locally

pjanevskiTT:
OK this change has been tested with simulator release that has sysmem support, but tags for the commands for AXI RAM read/write have values 8 and 9 (which has been taken by NEO DM functions). Ideally we should just be able to move it to some other values (like 10 and 11 in this PR). fyi @mlausinTT

Issue #2336, opened 2026-03-19
Ctrl+C doesn't work when waiting for device lock


OctopusET:
**Describe the bug**
When a process waits for `CHIP_IN_USE` lock held by another process, Ctrl+C is ignored. The only way to stop it is `kill -9` from another terminal.

`lock()` calls `pthread_mutex_lock()` after the first 1-second timeout, which blocks forever with no way to interrupt.

**Stack trace**
``` 
Waiting for lock 'CHIP_IN_USE_0_PCIe' which is currently held by thread TID: 2243607, PID: 2243607 (robust_mutex.cpp:407)
```

**To reproduce**
1. Terminal 1: start any process that opens a TT device
2. Terminal 2: start another process that opens the same device
3. Press Ctrl+C on terminal 2 — nothing happens

**Expected behavior**
The second process should exit when Ctrl+C is pressed.

Related: #973, #2287

PR #2337, opened 2026-03-19
Make RobustMutex::lock() interruptible by SIGINT

OctopusET:
Replace pthread_mutex_lock() (blocks forever) with a pthread_mutex_timedlock() loop. Install a temporary SIGINT handler during lock contention so Ctrl+C can abort the wait.

### Issue
#2336 

### Description
When a process waits for a device lock (e.g. `CHIP_IN_USE`) held by another process, Ctrl+C is ignored. The only way to stop it is `kill -9` from another terminal.

This happens because `lock()` calls `pthread_mutex_lock()` after the first 1-second timeout, which blocks forever with no way to interrupt.

### List of the changes
  - Replace blocking `pthread_mutex_lock()` with a `pthread_mutex_timedlock()` loop (1-second retries) via the existing `probe_lock()` helper
  - Add `SigintHandlerScope` (RAII): refcounted, process-wide install/restore of a SIGINT handler so concurrent `RobustMutex::lock()` callers don't race on `sigaction()` state
  - Use a monotonic `sigint_seq` counter (incremented by the handler, never reset) so concurrent waiters can't clear each other's interrupt notification
  - On interrupt: restore the previous handler, `raise(SIGINT)` so the original (e.g. Python) handler observes the signal, then throw `std::runtime_error("Interrupted")`
  - Add 3 unit tests (normal lock, same-process thread contention, cross-process via fork)

### Testing
- Unit tests pass (no hardware needed)
- Manually tested on Blackhole:
> Terminal 1: 
```python
python -c "import ttnn; device = ttnn.open_device(device_id=0); input('Press Enter to close'); ttnn.close_device(device)"
```
> Terminal 2:
```python
python -c "import ttnn; device = ttnn.open_device(device_id=0)"
```
> Ctrl+C on terminal 2:
```sh
2026-03-20 02:26:44.191 | info     |             UMD | Starting devices in cluster (cluster.cpp:946)                                                                                                                                                                                                                                                                                            
2026-03-20 02:26:45.191 | warning  |             UMD | Waiting for lock 'CHIP_IN_USE_0_PCIe' which is currently held by thread TID: 2550360, PID: 2550360 (robust_mutex.cpp:429)                                                                                                                                                                                                                
^C2026-03-20 02:26:46.191 | warning  |             UMD | Lock acquisition for 'CHIP_IN_USE_0_PCIe' interrupted (robust_mutex.cpp:419)                                                                                                                                                                                                                                                           
object address  : 0x7f38aacba2c0                                                                                                                                                                                                                                                                                                                                                                
object refcount : 3                                                                                                                                                                                                                                                                                                                                                                             
object type     : 0x13fae58                                                                                                                                                                                                                                                                                                                                                                     
object type name: RuntimeError                                                                                                                                                                                                                                                                                                                                                                  
object repr     : RuntimeError('Interrupted')                                                                                                                                                                                                                                                                                                                                                   
lost sys.stderr                              

```
### API Changes
No API changes here

### Note
- #2287 also changes `lock()` -- happy to rebase.
- The `object address` / `lost sys.stderr` output on exit is a tt-metal issue (exception not caught during interpreter shutdown), not a tt-umd issue.
```
object address  : 0x7f38aacba2c0                                                                                                                                                                                                                                                                                                                                                                
object refcount : 3                                                                                                                                                                                                                                                                                                                                                                             
object type     : 0x13fae58                                                                                                                                                                                                                                                                                                                                                                     
object type name: RuntimeError                                                                                                                                                                                                                                                                                                                                                                  
object repr     : RuntimeError('Interrupted')                                                                                                                                                                                                                                                                                                                                                   
lost sys.stderr 
```

### DIFF
--- a/device/utils/robust_mutex.cpp
+++ b/device/utils/robust_mutex.cpp
@@ -1,4 +1,5 @@
 // SPDX-FileCopyrightText: © 2025 Tenstorrent Inc.
+// SPDX-FileCopyrightText: © 2026 Sungjoon Moon
 //
 // SPDX-License-Identifier: Apache-2.0
 
@@ -15,9 +16,12 @@
 
 #include <cerrno>  // errno, ENOENT
 #include <chrono>
+#include <csignal>  // sigaction, SIGINT
 #include <cstdint>
 #include <ctime>  // clock_gettime, timespec
+#include <mutex>
 #include <optional>
+#include <stdexcept>
 #include <string>
 #include <tt-logger/tt-logger.hpp>
 #include <utility>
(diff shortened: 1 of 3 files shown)


OctopusET:
let me rebase it

device/utils/robust_mutex.cpp
```diff
@@ -95,6 +99,53 @@ class CriticalSectionScopeGuard {
...
 
+namespace {
+
+volatile sig_atomic_t sigint_seq = 0;
+
+void sigint_handler(int /*sig*/) { ++sigint_seq; }
+
+class SigintHandlerScope {
+public:
+    SigintHandlerScope() {
```
broskoTT, line 110: Not sure I like adding custom sigint handler to the core library...

Maybe a plausible solution would be for lock to accept a "should_lock" lambda, which would in case of nanobind be set to call "PyErr_CheckSignals" which would interrupt the locking mechanism?

The locking mechanism has to be rewritten anyway to loop and probe lock with some interval...

Issue #1606, opened 2025-11-13
Add soname/versioning to .so library


nbuncicTT:
What to do:
Add SONAME versioning to `libtt-umd.so` (usermode driver dynamic library).

Example: 
 - Real installed file is `libtt-umd.so.0.1.0`
 - SONAME is `libtt-umd.so.1`
 - devel symlink is `libtt-umd.so`

Hint - how to do:
- CMake: set OUTPUT_NAME tt-umd, SOVERSION 0, VERSION 0.1.0 on the shared library target; install and run ldconfig.
- GCC/LD: build libtt-umd.so.0.1.0 with -Wl,-soname,libtt-umd.so.0, then create symlinks libtt-umd.so.0 and libtt-umd.so.

Why it's necessary: Communicates ABI level, allows safe upgrades, avoids runtime breakage, and complies with distro/ldconfig expectations.

PR #2347, opened 2026-03-23
Set SOVERSION for tt-umd target

aleksamarkovicTT:
### Issue
#1606 

### Description
Set SOVERSION for tt-umd target.

### List of the changes
- Added SOVERSION to tt-umd CMake target.
- Idiomatic way to place output libraries to `build/lib`.
- Fixed DEB packaging configuration to work for soname'd so.

### Testing
CI, manual.

### API Changes

### DIFF
--- a/device/CMakeLists.txt
+++ b/device/CMakeLists.txt
@@ -359,27 +359,18 @@ else()
     message(STATUS "Install functionality disabled: CMake ${CMAKE_VERSION} < 3.23 required for FILE_SET support")
 endif()
 
-# Add a custom command to copy the library to build/lib
-add_custom_command(
-    TARGET tt-umd
-    POST_BUILD
-    COMMAND
-        ${CMAKE_COMMAND} -E make_directory ${CMAKE_BINARY_DIR}/lib
-    COMMAND
-        ${CMAKE_COMMAND} -E copy $<TARGET_FILE:tt-umd> ${CMAKE_BINARY_DIR}/lib/
-)
-if(NOT TT_UMD_BUILD_STATIC)
-    add_custom_command(
-        TARGET tt-umd
-        POST_BUILD
-        COMMAND
-            ${CMAKE_COMMAND} -E make_directory ${CMAKE_BINARY_DIR}/lib
-        COMMAND
-            ${CMAKE_COMMAND} -E create_symlink $<TARGET_FILE_NAME:tt-umd>
-            ${CMAKE_BINARY_DIR}/lib/$<TARGET_SONAME_FILE_NAME:tt-umd>
-        COMMAND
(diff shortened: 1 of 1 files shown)


device/CMakeLists.txt
```diff
@@ -37,6 +37,22 @@ else()
...
 
+# Set version
+set_target_properties(
+    tt-umd
+    PROPERTIES
+        VERSION
+            ${TT_UMD_VERSION_CMAKE}
+)
+# SOVERSION should contain only the major component (e.g. 0.9.4 -> 0)
+string(REGEX REPLACE "^([0-9]+)\..*" "\\1" TT_UMD_SOVERSION ${TT_UMD_VERSION_CMAKE})
```
nbuncicTT: If possible, replace this line with:
`SOVERSION ${PROJECT_VERSION_MAJOR}`

device/CMakeLists.txt
```diff
@@ -357,5 +373,19 @@ add_custom_command(
         ${CMAKE_COMMAND} -E make_directory ${CMAKE_BINARY_DIR}/lib
     COMMAND
         ${CMAKE_COMMAND} -E copy $<TARGET_FILE:tt-umd> ${CMAKE_BINARY_DIR}/lib/
-    COMMENT "Copying tt-umd library to build/lib"
 )
+if(NOT TT_UMD_BUILD_STATIC)
```
nbuncicTT: I don't like these `add_custom_command`s that we are having in this file. So I'd propose to fix those like this:
```cpp
if(NOT TT_UMD_BUILD_STATIC)
    set_target_properties(tt-umd PROPERTIES
        LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib
    )
else()
    set_target_properties(tt-umd PROPERTIES
        ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib
    )
endif()
```
and  for the symlinks - CMake creates them automatically. When you set `VERSION` + `SOVERSION` on a SHARED library, CMake's build system generators (both Makefile and Ninja) natively produce all three files in the `LIBRARY_OUTPUT_DIRECTORY`:


build/lib/libtt-umd.so.0.9.5   ← real file
build/lib/libtt-umd.so.0       ← soname symlink → libtt-umd.so.0.9.5
build/lib/libtt-umd.so         ← linker symlink → libtt-umd.so.0

This is built into the generator itself — it's not a post-build step, it's part of the library link rule. The same mechanism that creates these symlinks in the default build directory will create them in `LIBRARY_OUTPUT_DIRECTORY` if set.

If you feel this is an overkill for this PR, you can do it in another.

Issue #813, opened 2025-04-29
Simulated harvesting doesn't work for TRANSLATED


broskoTT:
TRANSLATED relies on same translation being implemented on the device, which won't be true in case of simulated harvesting in UMD,
Think about the design around this

broskoTT:
Simulated harvesting is broking in essence.
Think about whether we even need this

PR #2348, opened 2026-03-23
Remove simulated harvesting

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/813

### Description
Simulated harvesting never worked properly. The reason is that we would build TRANSLATED coords as if this was true harvesting on the device, but that would lead to TRANSLATED coords not matching translation on the actual device, so you'd end up hitting wrong cores.

Given that we haven't used this once, removing it.
There is a syseng tool to change the harvesting mask on the device itself, that can be used to simulate all different harvesting masks.

### List of the changes
- Removed all simulation_harvesting related code
- Removed related tests

### Testing
Removal of features, no testing

### API Changes

### DIFF
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -268,6 +268,16 @@ SocDescriptor Cluster::construct_soc_descriptor(
         chip_info.asic_location = cluster_desc->get_asic_location(chip_id);
     }
 
+    log_info(
+        LogUMD,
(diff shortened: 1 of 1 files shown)


device/cluster.cpp
```diff
@@ -304,10 +302,7 @@ void Cluster::add_chip(const ChipId& chip_id, const ChipType& chip_type, std::un
 }
 
 HarvestingMasks Cluster::get_harvesting_masks(
```
pjanevskiTT: can we maybe remove this function now as well since it just combines regular harvesting masks with simulated ones?

pjanevskiTT (approved):

Issue #1265, opened 2025-09-03
Follow up from bh104, update memcpy


broskoTT:
As a follow up from the issue https://tenstorrent.atlassian.net/browse/BH-104
1. add proper documentation to our cluster.hpp header documenting what API is used for what:
Some example:

> My current understanding is that we can declare the following for the APIs,
> write_to_device_reg:
> - synchronous (when the function returns, the value is in its destination)
> - if used from single thread, synchronous necessarily means ordered
> - deterministic number of writes
> 
> write_to_device:
> - Supposed to be more performant
> - Async
> - Don’t expect ordering guarantees
> - Don’t expect no double writes (what this issue uncovered, and we didn’t expect previously)
> - Calling l1_membar guarantees all previous writes completed.

2. avoid memcpy if we can on our codepaths

> Paul Keller
>   [9:58 PM](https://tenstorrent.slack.com/archives/C09DBQ5FNNL/p1756497537118549)
> IMHO, the fix should be to not use memcpy, no documentation necessary
> 
> Bojan Roško
>   [9:59 PM](https://tenstorrent.slack.com/archives/C09DBQ5FNNL/p1756497553217559)
> would that hurt the perf?
> 
> Paul Keller
>   [10:00 PM](https://tenstorrent.slack.com/archives/C09DBQ5FNNL/p1756497605159649)
> Generally role your own is faster than memcpy in my experience, especially if you validate a cpu specific copy routine (using eg streaming writes) and you can guarantee alignment on one or both sides
> [10:00](https://tenstorrent.slack.com/archives/C09DBQ5FNNL/p1756497626520429)
> We have a copy routine in runtime using streaming writes fwiw
> [10:00](https://tenstorrent.slack.com/archives/C09DBQ5FNNL/p1756497642247979)
> To write to “hugepage” in our case
> [10:01](https://tenstorrent.slack.com/archives/C09DBQ5FNNL/p1756497674675179)
> It doesn’t seem that our use of UMD today needs peak perf, though that may change in the future
> 
> 
> Bojan Roško
>   [10:07 PM](https://tenstorrent.slack.com/archives/C09DBQ5FNNL/p1756498042274559)
> so the action item is then to use something other than memcpy in umd
> :git+1:
> 1
> 
> Joel Smith
>   [10:08 PM](https://tenstorrent.slack.com/archives/C09DBQ5FNNL/p1756498136104159)
> If I remember right, memcpy is already avoided in a Wormhole-specific code path.
[...]

PR #2357, opened 2026-03-24
Replace raw memcpy with streaming stores for TLB window writes/reads

pjanevskiTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/1265

#### Problem

glibc's `memcpy` implementation can emit **overlapping stores** to the same address. For example, when copying between 4 and 7 bytes, glibc's `between_4_7` routine reads two overlapping 4-byte chunks and writes them both — meaning the overlapping bytes get written twice. On normal (cacheable) memory this is invisible, but when the destination is **device memory mapped through a TLB window**, each store is a separate PCIe write transaction. A double write corrupts data on the device side because the device sees two distinct writes to the same address with potentially different values (since the second write may carry stale data from the overlapping read).

This has been independently hit by multiple teams (BH-104, LLK test infra) and was expensive to debug.

#### Solution

Replace all raw `memcpy` calls that target device-mapped memory with custom copy routines (`streaming_memcpy_to`, `streaming_memcpy_from`) that guarantee **each address is written/read exactly once**.

On x86_64, these use **non-temporal (streaming) SIMD instructions**:
- **Writes**: AVX2 `_mm256_stream_si256` for 32-byte stores, SSE `_mm_stream_si128` for 16-byte, `_mm_stream_si32` for 4-byte, with an `_mm_sfence()` at the end.
- **Reads**: SSE4.1 `_mm_stream_load_si128` (MOVNTDQA) for 16-byte loads, with regular `_mm_storeu_si128` stores to host memory.

On other architectures (aarch64), a portable fallback uses explicit 4-byte volatile loads/stores.

#### Why streaming (non-temporal) stores/loads?

1. **No double writes**: Each SIMD instruction writes a fixed, non-overlapping region — the hardware guarantees a single store per address.
2. **Cache bypass**: Non-temporal stores bypass the CPU cache hierarchy and write directly to the write-combining buffer. This is faster for device memory (which is typically mapped as write-combining/uncacheable) because the data won't be read back from cache anyway — caching it just wastes cache lines and evicts useful data.
3. **Write combining**: The CPU's write-combining buffer can merge multiple streaming stores into fewer, larger PCIe transactions, improving bus utilization.
4. **~1.4-1.8x throughput improvement** over `memcpy` for device memory writes (based on tt-metal's experience with a similar approach).

#### Alignment handling

TLB window offsets can be arbitrary (not necessarily 4-byte aligned), and transfer sizes can be any number of bytes. The functions handle this with a phased approach:
- **Phase 0**: Byte-wide volatile stores/loads to align the device pointer to 4 bytes
- **Phase 1-5**: Progressively larger SIMD operations (4B → 32B → 256B bulk → tail)
- **Final phase**: Byte-wide volatile stores/loads for trailing bytes (0-3)

The existing Wormhole-specific `memcpy_to_device` / `memcpy_from_device` (which handle WH-specific device bugs like GDDR5 1-byte write adjacency) are also updated to use the streaming functions for their aligned middle section, while keeping their own RMW handling for leading/trailing partial words.

#### Trade-offs

- **x86_64 only for SIMD path**: The AVX2/SSE path is `#ifdef`'d to x86_64. Other architectures get a correct but non-optimized fallback using explicit 4-byte volatile stores.
- **`_mm_sfence()` cost**: Each call to `streaming_memcpy_to` ends with a store fence. For many small writes this adds overhead, but for the typical use case (large block transfers) it's negligible.
- **Write ordering**: Non-temporal stores are weakly ordered — they may reach the device in any order. The `_mm_sfence()` at the end ensures all stores from this call are visible before the function returns, but there is no ordering guarantee between individual stores within the call. This matches the existing `write_to_device` semantics (async, no ordering, use `l1_membar` for ordering).

### List of the changes
- Add `device_memcpy.hpp` with `streaming_memcpy_to` (AVX2 non-temporal stores) and `streaming_memcpy_from` (SSE4.1 non-temporal loads) — handles arbitrary alignment and size
- Replace raw `memcpy` in `write_block`/`read_block` for non-Wormhole architectures (Blackhole) with streaming functions
- Replace word-by-word copy loop in WH-specific `memcpy_to_device`/`memcpy_from_device` aligned middle with streaming functions
- Portable fallback (aarch64) uses explicit 4-byte volatile stores/loads

### Testing
CI

### API Changes

## Performance Results (CI comparison vs main)

| | WH N300 | WH N150 | WH N300-viommu | WH N150-viommu | WH N300-llmbox | BH P150b | BH P150b-viommu | BH P100a-viommu |
|---|---|---|---|---|---|---|---|---|
| **API Tests** | 1499→1379 ms (-8%) | 147→118 ms (-20%) | 6024→5779 ms (-4%) | 4499→4093 ms (-9%) | 6717→6391 ms (-5%) | 190→211 ms (+11%) | 2667→2588 ms (-3%) | 3764→3618 ms (-4%) |
| **Arch Tests** | 20178→6753 ms (**-67%**) | 19174→2447 ms (**-87%**) | 292236→123352 ms (**-58%**) | 29430→12837 ms (**-56%**) | 48434→28145 ms (**-42%**) | 15159→15433 ms (+2%) | 21834→20915 ms (-4%) | 20703→15593 ms (**-25%**) |
| **ReadWriteL1** | 2465→451 ms (**-82%**) | 1074→38 ms (**-96%**) | 2341→381 ms (**-84%**) | 1101→88 ms (**-92%**) | 8836→1956 ms (**-78%**) | 192→170 ms (-11%) | 236→185 ms (-22%) | 294→74 ms (**-75%**) |

### DIFF
--- a/device/api/umd/device/pcie/device_memcpy.hpp
+++ b/device/api/umd/device/pcie/device_memcpy.hpp
@@ -5,12 +5,6 @@
 #pragma once
 
 #include <cstddef>
-#include <cstdint>
-#include <cstring>
-
-#if defined(__x86_64__) || defined(_M_X64)
-#include <immintrin.h>
-#endif
 
 namespace tt::umd {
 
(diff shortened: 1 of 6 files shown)


device/api/umd/device/pcie/device_memcpy.hpp
```diff
@@ -0,0 +1,242 @@
...
+        _mm256_stream_si256(reinterpret_cast<__m256i*>(d_aligned + 64), v2);
+        _mm256_stream_si256(reinterpret_cast<__m256i*>(d_aligned + 96), v3);
+        _mm256_stream_si256(reinterpret_cast<__m256i*>(d_aligned + 128), v4);
+        _mm256_stream_si256(reinterpret_cast<__m256i*>(d_aligned + 160), v5);
+        _mm256_stream_si256(reinterpret_cast<__m256i*>(d_aligned + 192), v6);
+        _mm256_stream_si256(reinterpret_cast<__m256i*>(d_aligned + 224), v7);
+
+        d_aligned += 256;
+        s += 256;
+        size -= 256;
```
nbuncicTT: I can image that after this line in the while loop we could have the hang detection here for the PCIe controller.

Something like:
```cpp
        _mm_sfence();
        // PCIe health check: read a known register to detect a hung controller.
        // 0xFFFFFFFF on read = PCIe link down / completion timeout.
        // TODO: replace with a suitable register address for the target device.
        constexpr uintptr_t PCIE_HEALTH_CHECK_ADDR = 0x...;
        uint32_t readback = *reinterpret_cast<volatile uint32_t*>(PCIE_HEALTH_CHECK_ADDR);
        if (readback == 0xFFFFFFFF) {
            throw std::runtime_error("PCIe controller hung: health check readback returned 0xFFFFFFFF");
        }
```

Maybe a comment here? If you think it's redundant we can omit it for now.

PR #2379, opened 2026-03-27
Add DRAM_WORKER CoreType

jbaumanTT:
## Description
We're adding support to metal for running kernels on DRAM cores on blackhole. Each DRAM bank has 3 endpoints (noc addresses), and each endpoint has its own RISC core.

CoreType::DRAM is used inside metal to signify the DRAM bank itself and not a specific endpoint, so there is a non-trivial mapping between the two (e.g. the same logical coordinate may specify a different endpoint, depending on which NOC is used). To allow us to uniquely identify the endpoint (and RISC itself), add a new DRAM_WORKER type that metal won't adjust that way.

### List of the changes
Add CoreType::DRAM_WORKER and add support to a couple of places that need it.

### Testing
Builds.

### API Changes
This PR has API changes: these changes aren't breaking.

### DIFF
--- a/device/api/umd/device/types/core_coordinates.hpp
+++ b/device/api/umd/device/types/core_coordinates.hpp
@@ -34,6 +34,9 @@ enum class CoreType {
     HARVESTED,
     ETH,
     WORKER,
+    // Represents programmable DRISCs on hardware DRAM cores, distinct from CoreType::DRAM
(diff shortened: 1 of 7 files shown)


broskoTT:
Not needed at the moment, might reopen if something changes

Issue #2152, opened 2026-02-24
CoordIO - automatic coordinate translation in base API


broskoTT:
It doesn't matter how its called, the point is to have optional object which has soc descriptor, and has read/write for CoreCoords

PR #2397, opened 2026-04-01
LITERAL coord. system in CoreCoord

aleksamarkovicTT:
### Issue
#2152 

### Description
This PR introduces a LITERAL coordinate system for CoreCoord, signifying no translation or processing should be done to the x and y components of the coordinate, effectively treating the coordinate as if it was xy_pair.

### List of the changes
- Introduce `LITERAL` coord. system and `UNSPECIFIED` core type.
- Update constructors.
- Enhance CoreCoord string serialization.

### Testing
CI

### API Changes
This PR has API changes.


device/api/umd/device/types/core_coordinates.hpp
```diff
@@ -7,6 +7,7 @@
 #include <fmt/format.h>
 
 #include <cstdint>
+#include <sstream>
```
pjanevskiTT, line 10: `<sstream>` is a heavy include to pull into a widely-included header. `<fmt/format.h>` is already included on line 7 — `CoreCoord::str()` could use `fmt::format` instead of `std::stringstream`, and then this include can be dropped. Something like:
```cpp
std::string str() const {
    if (coord_system == CoordSystem::LITERAL) {
        return tt_xy_pair::str();
    }
    char shorthand = type_shorthand(core_type);
    if (shorthand != '\0') {
        return fmt::format("{}{}-{} ({})", shorthand, x, y, to_str(coord_system));
    }
    return fmt::format("{}-{} ({})", x, y, to_str(coord_system));
}
```

device/api/umd/device/types/core_coordinates.hpp
```diff
@@ -97,14 +134,21 @@ static inline std::string to_str(const CoordSystem coord_system) {
...
 
-    CoreCoord(const tt_xy_pair core, const CoreType type, const CoordSystem coord_system) :
-        tt_xy_pair(core), core_type(type), coord_system(coord_system) {}
+    constexpr CoreCoord(
+        const xy_pair core,
+        const CoreType type = CoreType::UNSPECIFIED,
+        const CoordSystem coord_system = CoordSystem::LITERAL) :
+        xy_pair(core), core_type(type), coord_system(coord_system) {}
 
     CoreType core_type;
```
pjanevskiTT: `core_type` and `coord_system` have no default member initializers, so the defaulted default constructor (`CoreCoord() = default;` on line 138) leaves them uninitialized. The other constructors now default to `UNSPECIFIED`/`LITERAL` — adding matching member initializers would make the default constructor consistent and avoid UB if someone default-constructs a `CoreCoord` and reads these fields:
```cpp
CoreType core_type = CoreType::UNSPECIFIED;
CoordSystem coord_system = CoordSystem::LITERAL;
```

device/api/umd/device/types/core_coordinates.hpp
aleksamarkovicTT: From experience I felt fmt was slower. Wrote the following benchmark to measure:
implementations:
```
    std::string str() const {
        if (coord_system == CoordSystem::LITERAL) {
            return tt_xy_pair::str();
        }
        std::stringstream ss;
        char shorthand = type_shorthand(core_type);
        if (shorthand != '\0') {
            ss << shorthand;
        }
        ss << x << '-' << y;
        ss << ' ' << '(' << to_str(coord_system) << ')';
        return ss.str();
    }

    std::string str2() const {
        if (coord_system == CoordSystem::LITERAL) {
            return tt_xy_pair::str();
        }
        char shorthand = type_shorthand(core_type);
        if (shorthand != '\0') {
            return fmt::format("{}{}-{} ({})", shorthand, x, y, to_str(coord_system));
        }
        return fmt::format("{}-{} ({})", x, y, to_str(coord_system));
    }
```

```
TEST(CooreCoord, Print) {
    CoreCoord x(1, 2, tt::CoreType::ARC, tt::CoordSystem::NOC1);
    auto bench =
        ankerl::nanobench::Bench().maxEpochTime(std::chrono::seconds(30)).title("print corecoord").unit("strings");
    bench.name("default").run([&] {
        auto out = x.str();
        ankerl::nanobench::doNotOptimizeAway(out);
    });
    bench.name("fmt").run([&] {
        auto out = x.str2();
        ankerl::nanobench::doNotOptimizeAway(out);
    });
}
```

Results on epyc machine:
```

|          ns/strings |           strings/s |    err% |     total | print corecoord
|--------------------:|--------------------:|--------:|----------:|:----------------
|              250.73 |        3,988,373.80 |    0.4% |      0.01 | `default`
|              585.45 |        1,708,096.45 |    2.1% |      0.01 | `fmt`
```

device/api/umd/device/types/core_coordinates.hpp
```diff
@@ -35,6 +36,7 @@ enum class CoreType {
     ETH,
     WORKER,
     COUNT,
+    UNSPECIFIED,
```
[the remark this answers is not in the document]
aleksamarkovicTT: This is a hallucination, COUNT is used for grid size, and is not being used in UMD currently.

device/soc_descriptor.cpp
```diff
@@ -224,6 +223,13 @@ CoreCoord SocDescriptor::get_coord_at(const tt_xy_pair core, const CoordSystem c
 
 CoreCoord SocDescriptor::translate_coord_to(
     const tt_xy_pair core_location, const CoordSystem input_coord_system, const CoordSystem target_coord_system) const {
+    if (input_coord_system == CoordSystem::LITERAL) {
```
nbuncicTT: Add this check to `translate_chip_coord_to_translated_coord` also, and evaluate if there are parts in the codebase where we also depend on the old `CoreCoord` information which `LITERAL` doesn't contain and add these checks if needed.

nbuncicTT (approved):
As said, after the discussion, just try in subsequent PRs to tackle the edge cases where an API would accept only CoreCoord which wouldn't throw because there would be no translations or similar, and in these APIs restrict the usage to not include literal type if there is some dependency on core type or coord system.

Otherwise, well done.

PR #2406, opened 2026-04-01
Add SWEmuleChip: software emulation chip for UMD

xanderchin:
### Description

  Introduce SWEmuleChip, a new Chip subclass for host-side software emulation via the https://github.com/tenstorrent/tt-emule project. SWEmuleChip provides real memory-backed read/write I/O (backed by
  tt_emule::Core mmap'd regions) while stubbing out all hardware operations (barriers, resets, power management, host memory channels) as no-ops. This allows tt-metal to run kernel programs on an emulated device
  without requiring physical Tenstorrent hardware.

  Worker cores are allocated from a contiguous L1Pool (MAP_32BIT mmap with 2 MB aligned slots for bitmask offset extraction). DRAM cores use individual mmaps sized to the full DRAM bank size.

 ### List of the changes

  - device/api/umd/device/cluster.hpp: Add ChipType::SWEMULE enum value
  - device/api/umd/device/chip/sw_emule_chip.hpp (new): SWEmuleChip class declaration extending Chip — memory-backed I/O via tt_emule::Core, lazy core creation with WORKER/DRAM role dispatch, L1Pool for worker L1
  allocation
  - device/chip/sw_emule_chip.cpp (new): Full implementation — constructor reads SOC descriptor for L1/DRAM sizes and DRAM core layout, get_core() lazy-creates cores from pool or individual mmap,
  write_to_device/read_from_device delegate to memcpy on core storage, all non-memory operations are no-ops
  - device/cluster.cpp: Wire SWEMULE into construct_chip_from_cluster() factory, handle SWEMULE alongside MOCK/SIMULATION in Cluster constructor (mock cluster descriptor, NOC translation enabled)
  - device/CMakeLists.txt: Conditionally compile sw_emule_chip.cpp when TT_UMD_BUILD_EMULE=ON, add TT_UMD_BUILD_EMULE compile definition and optional TT_EMULE_PATH include directory

### Testing

  Build-tested with TT_UMD_BUILD_EMULE=ON as part of the full tt-metal emulated-mode integration. End-to-end emulated tests (L1 buffer read/write, eltwise add, matmul, ttnn ops) pass through this code path via
  ChipType::SWEMULE.

 ### API Changes

  This PR has API changes:
  - Adds ChipType::SWEMULE enum value to the public ChipType enum
  - Adds new public header umd/device/chip/sw_emule_chip.hpp


pjanevskiTT (commented):
Changes look good in general, I would like to see a CI step to make sure we are not breaking your usage and maybe few nit details addressed

tests/baremetal/CMakeLists.txt
```diff
@@ -10,6 +10,9 @@ set(BAREMETAL_TESTS_SRCS
 
 add_executable(baremetal_tests ${BAREMETAL_TESTS_SRCS})
 target_link_libraries(baremetal_tests PRIVATE test_common)
+if(TT_UMD_BUILD_EMULE)
+    target_compile_definitions(baremetal_tests PRIVATE TT_UMD_BUILD_EMULE)
```
broskoTT: Not sure what this does?
If TT_UMD_BUILD_EMULE is set to on, then we'd have this line?
target_compile_definitions(baremetal_tests PRIVATE ON)

What is the effect if this is turned on and off? Also leave a comment, since others might be confused as well by this if I was now...

third_party/CMakeLists.txt
```diff
@@ -251,3 +251,16 @@ if(TT_UMD_ENABLE_TRACY)
             "CMAKE_MESSAGE_LOG_LEVEL NOTICE"
     )
 endif()
+
+############################################################################################################################
+# tt-emule (software emulation support)
+############################################################################################################################
+if(TT_UMD_BUILD_EMULE)
+    CPMAddPackage(
```
broskoTT, line 232: This command will fail if current git user doesn't have access to this repo? Maybe add that as a comment here.

tests/baremetal/test_cluster_descriptor_offline.cpp
```diff
@@ -415,3 +415,70 @@ TEST(ApiMockClusterTest, CreateMockClustersFromAllDescriptors) {
         }
     }
 }
+
+#ifdef TT_UMD_BUILD_EMULE
```
broskoTT: Can we move these tests to a dedicated file instead?

tests/baremetal/CMakeLists.txt
broskoTT: Ah you define the compiler constant to be used in this file...
Can we instead (as noted in other comment) have a dedicated file for emule tests, then compile the test file into baremetal_tests target only if TT_UMD_BUILD_EMULE?

tests/baremetal/test_cluster_descriptor_offline.cpp
```diff
@@ -415,3 +415,70 @@ TEST(ApiMockClusterTest, CreateMockClustersFromAllDescriptors) {
...
+        std::vector<uint8_t> data(1024, 0);
+        for (auto chip_id : emule_cluster->get_target_device_ids()) {
+            CoreCoord any_tensix_core = emule_cluster->get_soc_descriptor(chip_id).get_cores(CoreType::TENSIX)[0];
+            emule_cluster->write_to_device(data.data(), data.size(), chip_id, any_tensix_core, 0);
+            emule_cluster->read_from_device(data.data(), chip_id, any_tensix_core, 0, data.size());
+        }
+    }
+}
+
+TEST(ApiEmuleClusterTest, EmuleRoundtripIO) {
```
broskoTT: This is fine, but consider parametrizing some of the other tests if you want to test those as well. Some of our silicon tests can be parametrized to accept either SILICON or SWEMULE chip type and behave the same either way.
No action REQUIRED for this comment, feel free to ignore.

CMakeLists.txt
```diff
@@ -60,6 +60,13 @@ option(TT_UMD_BUILD_EXAMPLES "Enables build of tt_umd examples" OFF)
 option(TT_UMD_BUILD_ALL "Enables build of all tt_umd components (tests, simulation, python, tools, examples)" OFF)
 option(TT_UMD_BUILD_STATIC "Enables build of tt_umd as a static library" OFF)
 option(TT_UMD_ENABLE_TRACY "Enable Tracy profiling support" OFF)
+option(TT_UMD_BUILD_EMULE "Enable software emulation chip (SWEmuleChip)" OFF)
+set(TT_EMULE_PATH "" CACHE PATH
+    "Local path to tt-emule source (overrides CPM fetch from GitHub)")
+
+if(TT_UMD_BUILD_EMULE AND TT_EMULE_PATH)
+    set(CPM_tt_emule_SOURCE "${TT_EMULE_PATH}")
```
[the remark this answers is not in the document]
broskoTT: +1, please don't just silently ignore if PATH is not set

third_party/CMakeLists.txt
broskoTT: Also maybe we need a specific if/else to handle such case? In case somebody doesn't have permissions maybe we can print out that as an error message?

device/chip/sw_emule_chip.cpp
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    // Allocate L1Pool for worker cores.
+    // Use a generous count covering Tensix + Ethernet + Router + other non-DRAM cores,
+    // since all non-DRAM cores go through the pool for consistent bitmask offset extraction.
+    // Add extra headroom for cores created via translated coords that differ from physical coords.
+    size_t num_tensix = soc.get_cores(tt::CoreType::TENSIX).size();
+    // 128 is a safe upper bound on Tensix cores across known architectures (Wormhole=72,
+    // Blackhole~120). Used as fallback if SOC descriptor reports zero.
+    size_t pool_size = (num_tensix > 0 ? num_tensix : 128) * 2;  // 2× headroom
```
broskoTT, line 49: If user explicitly passed a soc descriptor with 0 tensix, then what is the reason for that? I feel it might be better to just throw or report as warning, rather than providing fallback logic. Won't block on this though, you're the user of it :)

device/cluster.cpp
```diff
@@ -53,6 +53,9 @@
 #include "umd/device/arch/wormhole_implementation.hpp"
 #include "umd/device/chip/local_chip.hpp"
 #include "umd/device/chip/mock_chip.hpp"
+#ifdef TT_UMD_BUILD_EMULE
```
broskoTT: this ifdef is not needed, as you don't guard the inclusion of the header file in cmakelists with this definition.

I'm saying you can always include the header. And bellow you still need the ifdef for calling the destructor obviously.

If this is not intentional, then change the inclusion in CMakeLists to include this header only if TT_UMD_BUILD_EMULE is enabled

device/api/umd/device/chip/sw_emule_chip.hpp
```diff
@@ -0,0 +1,114 @@
...
+#include <cstdint>
+#include <memory>
+#include <mutex>
+#include <set>
+#include <unordered_map>
+#include <unordered_set>
+
+#include "umd/device/chip/chip.hpp"
+
+namespace tt_emule {
```
broskoTT, line 16: Just something to consider in general for emule:
the essence of namespace name tt_emule is that this is something called emule but tied to something called tt.
Consider being more explicit by having namespace tt::emule instead :)

broskoTT (approved):
Sorry for waiting on the review, and good luck on your endeavours :)

pjanevskiTT (approved):

tests/baremetal/CMakeLists.txt
xanderchin: fixed up in latest commit

tests/baremetal/test_cluster_descriptor_offline.cpp
xanderchin: yes. updated.

device/chip/sw_emule_chip.cpp
xanderchin: Yep. I think the logic here can definitely be improved in the future. Wasn't a point of focus for this initial bringup.

device/cluster.cpp
xanderchin: updated.

device/api/umd/device/chip/sw_emule_chip.hpp
xanderchin: will address this in future updates.

PR #2415, opened 2026-04-03
Support returning unhealthy devices in TopologyDiscovery

aleksamarkovicTT:
### Issue
#2388

### Description
Support returning unhealthy devices through TopologyDiscovery. Enable by setting `device_init_failure_action` to `IGNORE`.
For local unhealthy devices:
- TTDevices returned from `discover()`, check in `ClusterDescriptor::get_unhealthy_devices()` for unhealthy ChipIDs.
- No information about unhealthy devices in YAML.

For remote unhealthy devices:
- TTDevices returned from `discover()`, check in `ClusterDescriptor::get_unhealthy_devices()` for unhealthy ChipIDs.
- Information about links to unhealthy devices is still present in ClusterDescriptor.
- Discovery will **not** be retried for over another core for a device which has been deemed unhealthy through the first ETH core it appeared in.

This option is forced to `THROW` for Cluster, because the impact has not been researched completely.

### List of the changes
- Introduce topology discovery option `device_init_failure_action`.
- Introduce cluster descriptor field `unhealthy_devices` containing ChipIDs of unhealthy devices.
- Topology discovery now can return devices which are considered "unhealthy" (ARC startup failure).

### Testing
CI, manually tested on n300.
Waiting on tt-exalens to confirm PR functionality.

### API Changes
This PR has API changes.


device/cluster.cpp
```diff
@@ -1009,7 +1010,12 @@ std::unique_ptr<ClusterDescriptor> Cluster::create_cluster_descriptor(
     IODeviceType device_type,
     const TopologyDiscoveryOptions& topology_discovery_options) {
     ZoneScopedC(tracy::Color::DarkGreen);
-    return TopologyDiscovery::discover(topology_discovery_options, device_type, sdesc_path).first;
+    auto adjusted_topology_options = topology_discovery_options;
+    if (adjusted_topology_options.device_init_failure_action != TopologyDiscoveryOptions::Action::THROW) {
```
broskoTT: Ah I see what you mean by "hack".. Although it already set its default to THROW, so even without this code, I don't expect anyone in metal to hit this issue, nobody will touch this option through Cluster.
But I'm okay with this hack.
But why the IF, maybe just always set it? :) Whatever you choose its fine.

broskoTT (approved):

device/topology/topology_discovery.cpp
```diff
@@ -146,7 +147,22 @@ void TopologyDiscovery::get_connected_devices() {
...
 
         // When coming out of reset, devices can take on the order of minutes to become ready.
-        tt_device->init_tt_device(timeout::ARC_LONG_POST_RESET_TIMEOUT);
+        auto init_result = tt_device->init_tt_device(
+            timeout::ARC_LONG_POST_RESET_TIMEOUT,
+            /*throw_on_arc_failure=*/options.device_init_failure_action == TopologyDiscoveryOptions::Action::THROW);
+        if (init_result != TTDeviceInitResult::SUCCESSFUL) {
+            uint64_t asic_id = get_unhealthy_asic_id(device_id);
+            devices_to_discover.emplace(asic_id, std::move(tt_device));
+            asic_id_to_chip_id.emplace(asic_id, chip_id);
```
tt-vjovanovic: Should you add device to `unhealthy_devices`?

device/topology/topology_discovery.cpp
```diff
@@ -146,7 +147,22 @@ void TopologyDiscovery::get_connected_devices() {
         ChipId chip_id = get_next_chip_id();
 
         // When coming out of reset, devices can take on the order of minutes to become ready.
-        tt_device->init_tt_device(timeout::ARC_LONG_POST_RESET_TIMEOUT);
+        auto init_result = tt_device->init_tt_device(
+            timeout::ARC_LONG_POST_RESET_TIMEOUT,
+            /*throw_on_arc_failure=*/options.device_init_failure_action == TopologyDiscoveryOptions::Action::THROW);
+        if (init_result != TTDeviceInitResult::SUCCESSFUL) {
+            uint64_t asic_id = get_unhealthy_asic_id(device_id);
```
pjanevskiTT: `get_unhealthy_asic_id(device_id)` uses the PCI device index (`int`) rather than the `ChipId` that was just allocated on line 148. This diverges from the remote path (line 342) which correctly uses `chip_id`. If a local device is skipped due to architecture mismatch (line 144), `device_id` and `chip_id` will drift apart, producing an inconsistent mock ASIC ID. Should this be `get_unhealthy_asic_id(chip_id)`?

---
🤖 Generated by Claude (claude-opus-4-6).

device/topology/topology_discovery.cpp
```diff
@@ -305,11 +327,34 @@ void TopologyDiscovery::discover_remote_devices() {
...
 
-                devices_to_discover.emplace(remote_asic_id, std::move(remote_device));
+                try {
+                    remote_device->init_tt_device();
+                    if (options.wait_on_ethernet_link_training) {
+                        wait_eth_cores_training(remote_device.get());
+                    }
+                    // Put device in discovery queue only if it is healthy.
+                    devices_to_discover.emplace(remote_asic_id, std::move(remote_device));
+                } catch (std::runtime_error& re) {  // Catches both remote comm. failure and ARC init failure.
```
pjanevskiTT: The try-catch wraps both `init_tt_device()` and `wait_eth_cores_training()`. If init succeeds but ETH training times out, the device gets marked unhealthy even though ARC startup was fine. Is this intentional? The comment says "Catches both remote comm. failure and ARC init failure" but doesn't mention training failure. If training failure should be treated differently, consider narrowing the try scope or checking which call actually threw.

Also nit: `re` is unused — consider `catch (const std::runtime_error&)`.

---
🤖 Generated by Claude (claude-opus-4-6).

device/cluster.cpp
```diff
@@ -1009,7 +1010,12 @@ std::unique_ptr<ClusterDescriptor> Cluster::create_cluster_descriptor(
...
+        log_warning(LogUMD, "Ignoring device init. failures is not supported in Cluster. Overrriding to THROW.");
```
pjanevskiTT: Typo: "Overrriding" (triple 'r') — should be "Overriding".

---
🤖 Generated by Claude (claude-opus-4-6).

device/topology/topology_discovery.cpp
```diff
@@ -305,11 +327,34 @@ void TopologyDiscovery::discover_remote_devices() {
...
+
+                    log_debug(
+                        LogUMD,
+                        "Discovered remote device ASIC ID: {} is unhealthy. Assigned mock ASIC ID: {}",
+                        remote_asic_id,
+                        mock_asic_id);
+                }
+                // These actions are safe for both healthy and unhealthy devices.
                 asic_id_to_chip_id.emplace(remote_asic_id, chip_id);
                 active_eth_channels_per_device.emplace(remote_asic_id, std::set<uint32_t>());
```
pjanevskiTT: For unhealthy remote devices, the device is stored under `mock_asic_id` in `devices_to_discover` (line 343), but `active_eth_channels_per_device` and `remote_asic_id_to_mmio_device_id` are keyed on `remote_asic_id` here. In `fill_cluster_descriptor_info`, the channel loop iterates `active_eth_channels_per_device` and will populate `idle_eth_channels`/`active_eth_channels` for this chip_id, even though the chip was excluded from `all_chips`. This inconsistency is probably harmless but could confuse consumers that iterate channel maps.

---
🤖 Generated by Claude (claude-opus-4-6).

device/topology/topology_discovery.cpp
aleksamarkovicTT: fixed.

pjanevskiTT (approved):

PR #2442, opened 2026-04-08
[Bug fix] Fix integer overflow in TLB window read/write reconfigure

sagarwalTT:
Fixes an integer overflow issue in TLB window configuration by correcting type mismatches in `read_block_reconfigure()` and `write_block_reconfigure()`.

The functions were storing the result of `get_size()` (which returns `size_t`) into a `uint32_t` variable, causing overflow when sizes exceeded 32-bit range. The fix uses `size_t` for `tlb_size` to preserve the full value returned by `get_size()`, and casts the `size` parameter to `size_t` when comparing with `tlb_size` in `std::min()`.

### DIFF
--- a/device/pcie/tlb_window.cpp
+++ b/device/pcie/tlb_window.cpp
@@ -36,8 +36,8 @@ void TlbWindow::read_block_reconfigure(
 
     while (size > 0) {
         configure(config);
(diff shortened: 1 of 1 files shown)


sagarwalTT:
Closing as this will be fixed in #2444

Issue #791, opened 2025-04-25
Memcpy performance


pjanevskiTT:
Part of the larger effort #751 

As seen by doing initial analysis in #751, memcpy bandwidth seems very inconsistent across different chunk sizes (#751 for details). We should investigate why this happens

- Involve CPU guys to help

Some of the potential solutions to improve the situation

- Implement SIMD memcpy for x86

pjanevskiTT:
we are going to close this after #2455

pjanevskiTT:
closed by #2455, we have our own implementation of copy right now

PR #2455, opened 2026-04-12
Re-land #2357: avoid overlapping memcpy stores on TLB-mapped device memory

pjanevskiTT:
### Issue
closing https://github.com/tenstorrent/tt-umd/issues/1265
closing #68
closing #791

### Description
Re-lands #2357 which was reverted in #2430.

See the original PR for the full description: https://github.com/tenstorrent/tt-umd/pull/2357

#### Summary

Replace all raw `memcpy` calls that target device-mapped memory with custom copy routines (`umd_memcpy_to_device`, `umd_memcpy_from_device`) that guarantee **each address is written/read exactly once**.

glibc's `memcpy` can emit overlapping stores, which corrupts data on device memory mapped through TLB windows because each store is a separate PCIe write transaction. The replacement routines step through the buffer in non-overlapping chunks so no destination address is ever written twice. On x86_64 the bulk uses regular AVX2 unaligned 256-bit loads/stores (`VMOVDQU`).

### List of the changes
- Add internal header `device/pcie/device_memcpy.hpp` with `umd_memcpy_to_device` / `umd_memcpy_from_device`
- Replace raw `memcpy` in `write_block`/`read_block` for non-Wormhole architectures with the new functions
- Replace word-by-word copy loop in WH-specific `SiliconTlbWindow::memcpy_to_device`/`memcpy_from_device` aligned middle with the new functions
- Portable fallback (non-x86_64) uses explicit 4-byte volatile stores/loads
- Add `WriteCountMatchesPostedWrites` test in `test_cluster_bh.cpp`

### Testing
CI

### Performance

TLB-IO + ReadWriteL1 test time (PR vs main `40a65dc2`, DMA tests excluded as they use random per-iteration sizes):

| Machine | PR (s) | main (s) | Δ% |
|---|---:|---:|---:|
| **Wormhole** | | | |
| n150-stable | 5.55 | 18.26 | **−69.6%** |
| n150-viommu-stable | 8.98 | 21.74 | **−58.7%** |
| n300-stable | 5.92 | 18.66 | **−68.3%** |
| n300-viommu-stable | 10.04 | 23.36 | **−57.0%** |
| n300-llmbox-stable | 11.93 | 28.93 | **−58.8%** |
| n300-llmbox-viommu-stable | 27.38 | 50.13 | **−45.4%** |
| **Blackhole** | | | |
| p100a-viommu-stable | 10.13 | 9.58 | +5.7% |
| p150b-stable | 4.21 | 4.32 | −2.5% |
| p150b-viommu-stable | 6.88 | 6.95 | −0.9% |
| **TOTAL** | **91.01** | **181.93** | **−50.0%** |

Test set: `TestDeviceIOFixture.*`, `CoreTypes/TestDeviceIOFixture.*` (`DynamicTLB_RW`, `SimpleIO*`, `RemoteFlush`), `SiliconDriver{WH,BH}.{Unaligned,Static,Dynamic}TLB_RW`, `SiliconDriver{WH,BH}.MultiThreadedMemBar`, `ClusterReadWriteL1Test.ReadWriteL1/Silicon`.

- **Wormhole:** 45–70% faster across all six machines (old 4-byte word loop → AVX2 256-bit `VMOVDQU`).
- **Blackhole:** essentially flat (−2.5% to +5.7%); the old path was already AVX-vectorized glibc memcpy.

### API Changes

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,8 +1,0 @@
-{
-    "permissions": {
-        "allow": [
-            "Read(//home/pjanevski/.config/gh/**)",
-            "Bash(python3 -c ' *)"
-        ]
-    }
-}
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,0 +1,8 @@
+{
+  "permissions": {
+    "allow": [
+      "Read(//home/pjanevski/.config/gh/**)",
+      "Bash(python3 -c ' *)"
+    ]
+  }
+}
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,8 +1,8 @@
 {
-  "permissions": {
-    "allow": [
-      "Read(//home/pjanevski/.config/gh/**)",
-      "Bash(python3 -c ' *)"
-    ]
-  }
+    "permissions": {
+        "allow": [
+            "Read(//home/pjanevski/.config/gh/**)",
+            "Bash(python3 -c ' *)"
+        ]
+    }
 }
--- a/device/pcie/device_memcpy.cpp
+++ b/device/pcie/device_memcpy.cpp
@@ -17,31 +17,19 @@ void streaming_memcpy_to_device(volatile void* dest, const void* src, std::size_
     auto* d = static_cast<volatile std::uint8_t*>(dest);
     auto* s = static_cast<const std::uint8_t*>(src);
 
-    // Phase 0: Align device destination address to 4 bytes using byte-wide volatile stores.
+    // Phase 0: Align device destination to 4 bytes using byte-wide volatile stores.
     while (size > 0 && (reinterpret_cast<std::uintptr_t>(d) % 4) != 0) {
         *d++ = *s++;
         size--;
     }
 
 #if defined(__x86_64__) || defined(_M_X64)
-    // d is now 4-byte aligned. Cast to non-volatile for SIMD intrinsics — the intrinsics
-    // are opaque to the compiler and won't be reordered or eliminated.
+    // Cast to non-volatile for SIMD intrinsics — the intrinsics are opaque to the
+    // compiler and won't be reordered or eliminated.
     // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast).
-    auto* d_aligned = const_cast<std::uint8_t*>(static_cast<const volatile std::uint8_t*>(d));
+    auto* d_simd = const_cast<std::uint8_t*>(static_cast<const volatile std::uint8_t*>(d));
 
-    // Phase 1: Align destination to 32 bytes using 4-byte streaming stores.
-    // This prepares for Phase 2 which uses _mm256_stream_si256 (VMOVNTDQ) to write
-    // 256 bits (32 bytes) at a time — that instruction requires 32-byte alignment.
-    while (size >= 4 && (reinterpret_cast<std::uintptr_t>(d_aligned) % 32) != 0) {
-        std::uint32_t tmp;
-        std::memcpy(&tmp, s, sizeof(tmp));
-        _mm_stream_si32(reinterpret_cast<int*>(d_aligned), static_cast<int>(tmp));
-        d_aligned += 4;
-        s += 4;
-        size -= 4;
-    }
-
-    // Phase 2: Bulk 256-byte blocks (8 x 32-byte AVX2 streaming stores).
+    // Phase 1: Bulk 256-byte blocks (8 x 32-byte AVX2 unaligned stores).
     while (size >= 256) {
         __m256i v0 = _mm256_loadu_si256(reinterpret_cast<const __m256i*>(s));
         __m256i v1 = _mm256_loadu_si256(reinterpret_cast<const __m256i*>(s + 32));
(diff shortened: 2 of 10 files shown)


.claude/settings.local.json
```diff
@@ -0,0 +1,8 @@
+{
```
pjanevskiTT: This file should not be committed: per the workspace convention, Claude-tooling config lives under `~/.claude/projects/<sanitized-dir>/`, not in the repo. Delete the file (and add `.claude/` to `.gitignore` if it isn't already) before merging.

device/pcie/device_memcpy.cpp
```diff
@@ -0,0 +1,173 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "umd/device/pcie/device_memcpy.hpp"
+
+#include <cstdint>
+#include <cstring>
+
+#if defined(__x86_64__) || defined(_M_X64)
```
pjanevskiTT, line 10: `__x86_64__` does **not** imply AVX2. AVX2 is a CPU extension introduced with Haswell; older x86-64 CPUs (Nehalem, Sandy Bridge, AMD pre-Excavator, etc.) and many cloud SKUs without AVX2 will SIGILL on `_mm256_loadu_si256`. Options: (a) compile with `-mavx2` and document the runtime requirement; (b) gate on `__AVX2__` instead of `__x86_64__`; (c) runtime `__builtin_cpu_supports("avx2")` dispatch. If the target hardware for UMD is guaranteed to be AVX2-capable, at least add a static_assert or compile-time `__AVX2__` guard so the dependency is explicit.

device/pcie/device_memcpy.cpp
```diff
@@ -0,0 +1,173 @@
...
+    // Phase 0: Align device destination to 4 bytes using byte-wide volatile stores.
+    while (size > 0 && (reinterpret_cast<std::uintptr_t>(d) % 4) != 0) {
+        *d++ = *s++;
+        size--;
+    }
+
+#if defined(__x86_64__) || defined(_M_X64)
+    // Cast to non-volatile for SIMD intrinsics — the intrinsics are opaque to the
+    // compiler and won't be reordered or eliminated.
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast).
```
pjanevskiTT, line 30: The comment "intrinsics are opaque to the compiler and won't be reordered or eliminated" overstates the guarantee: `_mm256_loadu_si256` / `_mm256_storeu_si256` are *not* volatile and the compiler is free to reorder them relative to other non-volatile accesses (it can't reorder against the volatile byte loops you have at the boundaries, which is what saves you here). Either tighten the comment to "the surrounding `volatile` byte loops bound the reordering window" or use `_mm256_stream_si256` + `_mm_sfence()` if you actually need ordering. Same comment applies on line 108 for the load side.

tests/blackhole/test_cluster_bh.cpp
```diff
@@ -510,6 +510,46 @@ TEST(SiliconDriverBH, DISABLED_VirtualCoordinateBroadcast) {  // same problem as
     cluster.close_device();
 }
 
+TEST(SiliconDriverBH, DebugDoubleWrite) {
```
pjanevskiTT: `TEST(SiliconDriverBH, DebugDoubleWrite)` looks like a verification harness used to chase the original double-write bug rather than a regression test someone should run in CI: the 100k iteration loop is expensive, the name starts with "Debug", and it asserts a NIU counter equals an in-process counter (which only works in single-writer single-process configurations). If it's intended to stay, rename to something like `WriteCountMatchesPostedWrites`, drop the iteration count, and document the assumption. If it's leftover debugging, remove it.

tests/api/test_device_io.cpp
```diff
@@ -497,6 +497,13 @@ TEST_P(ClusterReadWriteL1Test, ReadWriteL1) {
 
         cluster->read_from_device(readback_data.data(), chip_id, tensix_core, SAFE_IO_L1_ADDRESS, test_size);
 
+        for (int i = 0; i < zero_data.size(); i++) {
+            if (zero_data[i] != readback_data[i]) {
+                std::cout << "diff at index " << i << " write " << (uint32_t)zero_data[i] << " "
```
pjanevskiTT: Debug `std::cout` dump on mismatch that bypasses gtest's reporting (also at lines 517, 825, 900, 954). The `EXPECT_EQ` / `ASSERT_EQ` immediately after each block already produces a structured failure message, and a per-byte dump to stdout will drown CI logs on the first regression. Please remove these blocks before merging.

tests/unified/multiprocess.cpp
```diff
@@ -77,9 +98,12 @@ static void test_read_write_all_tensix_cores_impl(
         for (const CoreCoord& core : cluster->get_soc_descriptor(0).get_cores(CoreType::TENSIX)) {
             cluster->write_to_device(vector_to_write.data(), data_size, 0, core, address);
             cluster->l1_membar(0, {core});
-            test_utils::read_data_from_device(*cluster, readback_vec, 0, core, address, data_size);
+            const int attempts = read_until_match(vector_to_write, readback_vec, [&]() {
```
pjanevskiTT: `read_until_match` retries the readback up to 100 times. For the DMA tests below this is plausible (DMA completion is asynchronous w.r.t. the host's read path), but `test_read_write_all_tensix_cores_impl` already issues `cluster->l1_membar(0, {core})` between the write and the read — if a readback there ever needs *any* retry, the membar isn't doing what it's supposed to and the streaming-stores patch is exposing missing ordering. Same concern for line 326 / 345 (no DMA involved). Suggest: gate `read_until_match` to the DMA-only tests, and let the L1-membar tests fail loudly if a retry is needed — otherwise this re-land will mask the very ordering bug that prompted the original revert.

pjanevskiTT: this was reverted from multirprocess.cpp tests

tests/api/test_device_io.cpp
pjanevskiTT: removed

device/pcie/device_memcpy.cpp
pjanevskiTT: working on newer CPUs, should work, maybe we can define more strict rules about this

device/pcie/device_memcpy.cpp
```diff
@@ -0,0 +1,175 @@
...
+#include <cstdint>
+#include <cstring>
+
+#if defined(__x86_64__) || defined(_M_X64)
+#include <immintrin.h>
+#endif
+
+namespace tt::umd {
+
+void umd_memcpy_to_device(volatile void* dest, const void* src, std::size_t size) {
```
broskoTT: nit: no need for umd_, it is already in tt::umd namespace

device/pcie/device_memcpy.cpp
```diff
@@ -0,0 +1,175 @@
...
+    auto* d = static_cast<volatile std::uint8_t*>(dest);
+    auto* s = static_cast<const std::uint8_t*>(src);
+
+    // Phase 0: Align device destination to 4 bytes using byte-wide volatile stores.
+    while (size > 0 && (reinterpret_cast<std::uintptr_t>(d) % 4) != 0) {
```
broskoTT, line 21: does this do a reinterpret cast + mod operation on each byte?
reinterpret cast is probably something trivial but mod operation can be expensive?

device/pcie/device_memcpy.cpp
```diff
@@ -0,0 +1,175 @@
...
+        *d++ = *s++;
+        size--;
+    }
+
+#if defined(__x86_64__) || defined(_M_X64)
```
broskoTT, line 26: You can leave this for another PR:
But we can probably move this to the CMakeLists rather than having it in cpp file. By having a specific function with two implementations based on which cpp is included by cmake

device/pcie/device_memcpy.cpp
```diff
@@ -0,0 +1,175 @@
...
+        _mm256_storeu_si256(reinterpret_cast<__m256i*>(d_simd + 64), v2);
+        _mm256_storeu_si256(reinterpret_cast<__m256i*>(d_simd + 96), v3);
+        _mm256_storeu_si256(reinterpret_cast<__m256i*>(d_simd + 128), v4);
+        _mm256_storeu_si256(reinterpret_cast<__m256i*>(d_simd + 160), v5);
+        _mm256_storeu_si256(reinterpret_cast<__m256i*>(d_simd + 192), v6);
+        _mm256_storeu_si256(reinterpret_cast<__m256i*>(d_simd + 224), v7);
+
+        d_simd += 256;
+        s += 256;
+        size -= 256;
```
broskoTT, line 56: size is of unsigned type? This underflows.
Please make sure to fix

broskoTT: Ah I now see the check in the while loop... this is not expected to drop to zero or below, right.
Maybe nit: add a comment just above this line that the loop breaks once size ends up being 0 <= size < 256, just for minor clarity here

device/pcie/device_memcpy.cpp
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    // Phase 5: Trailing bytes (0-3) using byte-wide volatile stores.
+    while (size > 0) {
+        *d++ = *s++;
+        size--;
+    }
+}
+
+void umd_memcpy_from_device(void* dest, const volatile void* src, std::size_t size) {
```
broskoTT: remove umd_

broskoTT: Then use umd::memcpy_from_device at the callsites.
(it is bad that the original memcpy_from_device is in the global namespace)

broskoTT (approved):

device/pcie/device_memcpy.cpp
pjanevskiTT: but we do it at most 3 times until we align it, shouldn't be too expensive

pjanevskiTT:
passed almost all in merge queue, it was an infra problem for p300, so force merging it to not queue it again

PR #2457, opened 2026-04-14
Add NOC hang check to init_tt_device

aleksamarkovicTT:
### Issue
#2388 

### Description
Adds a NOC hang check and PCIe hang check in `init_tt_device()`. The function will throw if the currently selected NOC is hung or the PCIe interface to the device is hung. A `NocHangError` can be caught to signify the user should use a different NOC.

### List of the changes
- Added checking for NOC hangs in `init_tt_device()`.
- Removed `TTDeviceInitResult`. `init_tt_device()` returns void or throws errors.
- Introduced new error types `NocHangError`, `PCIeHangError` and `ArcStartupError`.
- Utility function `noc_to_str()`.

### Testing
CI, manual testing.

### API Changes
This PR has API changes.

### DIFF
--- a/device/api/umd/device/types/noc_id.hpp
+++ b/device/api/umd/device/types/noc_id.hpp
@@ -5,11 +5,23 @@
 #pragma once
 
 #include <cstdint>
+#include <string>
 
 namespace tt::umd {
 
 // NOC Identifiers that can be selected when communicating with the device.
 enum class NocId : uint8_t { DEFAULT_NOC = 0, NOC0 = 0, NOC1 = 1, SYSTEM_NOC = 2 };
+
+std::string noc_to_str(NocId noc_id) noexcept {
+    switch (noc_id) {
+        case NocId::NOC0:
+            return "NOC0";
+        case NocId::NOC1:
+            return "NOC1";
+        case NocId::SYSTEM_NOC:
+            return "System NOC";
+    }
+}
 
 // Set the NocId for the current thread.
 // All subsequent device communications from this thread will use the selected NocId.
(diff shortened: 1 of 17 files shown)


nbuncicTT (approved):

device/api/umd/device/types/noc_id.hpp
```diff
@@ -5,12 +5,25 @@
...
 
 #include <cstdint>
+#include <string>
 
 namespace tt::umd {
 
 // NOC Identifiers that can be selected when communicating with the device.
 enum class NocId : uint8_t { DEFAULT_NOC = 0, NOC0 = 0, NOC1 = 1, SYSTEM_NOC = 2 };
 
+inline std::string noc_to_str(NocId noc_id) noexcept {
```
nbuncicTT, line 15: Add a comment here that this is temporary because soon we will have "stateless NOC API" which won't have state and that's why it's okay to omit the `DEFAULT_NOC` from the print.

device/api/umd/device/utils/error.hpp
```diff
@@ -4,11 +4,24 @@
...
+#include "umd/device/tt_device/tt_device.hpp"
+#include "umd/device/types/arch.hpp"
+#include "umd/device/types/cluster_descriptor_types.hpp"
+#include "umd/device/types/communication_protocol.hpp"
+#include "umd/device/types/noc_id.hpp"
+#include "umd/device/types/xy_pair.hpp"
 #include "umd/device/utils/error_detail.hpp"
 
+namespace tt::umd {
+class TTDevice;
```
nbuncicTT, line 21: Do you need both the header file and the forward declaration?

aleksamarkovicTT: Keeping the forward decl.

PR #2471, opened 2026-04-20
Add CI job to smoke-test Tracy-enabled builds

### LINKED ISSUE #2305: CI pipeline integration for Tracy builds
### Objective
Ensure Tracy-enabled builds are tested in CI to prevent regressions.

### Tasks
- Add a CI job that builds UMD with `TT_UMD_ENABLE_TRACY=ON` and runs the test suite
- Ensure Tracy-enabled builds don't break any existing tests
- Optionally add a CI step that captures a basic trace and generates a FlameGraph as a build artifact

ajovanovicTT:
### Issue
#2305

### Description
Add CI job to smoke-test Tracy-enabled build

Builds UMD with TT_UMD_ENABLE_TRACY=ON and runs the OpenAllSiliconChips
test on an n150 runner, wired into the PR/push pipeline so Tracy
regressions are caught before merge.

### List of the changes
 - Added build-tracy matrix axis to build-tests.yml
 - Added run-tracy-test.yml (test-only, no build job)
 - Wired build-tracy: true and run-tracy job into build-and-run-all-tests.yml

### Testing
 - CI

### API Changes


.github/workflows/build-and-run-tracy-test.yml
```diff
@@ -0,0 +1,129 @@
...
+      # CI runners may lack invariant TSC support (required by Tracy for high-res timestamps).
+      # This skips Tracy's runtime TSC check to prevent initialization failures.
+      TRACY_NO_INVARIANT_CHECK: 1
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Build umd_tests with Tracy enabled
```
broskoTT: Let's not have build jobs all around...
I think you can integrated this better in build-and-run-all-tests (not build-and-run-all-tests-multiple-builds).
Look at how some build parameters are parametrized via `use_release_flags: true` in build-device.yml.
Then create a same kind of matrix in build-tests.yml to `build` job.
You will have something like `-DTT_UMD_ENABLE_TRACY=${{ inputs.matrix.use_tracy }}`
Also make sure it uploads an artifact which has -tracy suffix.

Then add a run-tracy job in build-and-run-all-tests.
That run-tracy should call your current build-and-run-tracy-test.yml, to be renamed to run-tracy-test.yaml, and you can have just the second test job in it

broskoTT (commented):
I'll take another look, ping me after you resolve the comment

.github/workflows/build-tests.yml
```diff
@@ -22,6 +22,13 @@ on:
         required: false
         type: boolean
         default: true
+      build-tracy:
```
broskoTT, line 25: Add to workflow_dispatch section as well (this is what is shown up if you manually want to run the job)

broskoTT (approved):
Thanks for adjusting!
As discussed offline, review if it makes sense to just enable it on all builds and with all tests, which would simplify the workflow files.
In that case we could just add TracyRelease and TracyDebug build types to our cmake config

PR #2473, opened 2026-04-20
Simulation: add advance_device_execution API to enable fast-dispatch on tt-sim

pjanevskiTT:
### Issue
/

### Description
Enables fast-dispatch execution on the TT simulator (`tt-sim`). Adds a `Cluster::advance_device_execution` API that lets clients advance the simulation clock from the calling thread, plus a couple of simulation-only fixes encountered while bringing up fast-dispatch.

**Why no clocking thread?** Simulator clocking must remain deterministic so runs are reproducible. A background thread driving `advance_clock` would race with reads/writes and produce non-reproducible behavior, so the clock is advanced synchronously from the calling thread via the explicit `advance_device_execution` API. The same rationale is captured in code on `TTDevice::advance_device_execution` and `TTSimTTDevice::advance_device_execution`.

### List of the changes
- Added `virtual void advance_device_execution()` on `TTDevice` (no-op default), overridden by `TTSimTTDevice` to call `communicator_->advance_clock(1)` — the API lives at the device layer so chip/cluster code is a thin pass-through.
- Added `void Chip::advance_device_execution()` that delegates to its underlying `TTDevice`.
- Added `Cluster::advance_device_execution(ChipId)` public API that forwards to the chip.
- Made `TLBManager::configure_tlb` debug logging safe when `get_pci_device()` returns nullptr (simulation case) and added an inline comment explaining the chip-id 0 fallback.
- Fixed `TTSimTlbHandle`'s destructor to avoid teardown issues (defaulted, no `free_tlb()` call — the manager/communicator may already be torn down by the time the handle dies during simulation chip shutdown).
- Minor pointer-style cleanup in `SimulationSysmemManager` (`void *` → `void*`, picked up by clang-format).

### Testing
CI

### API Changes
This PR adds a new function to the public API: `Cluster::advance_device_execution(ChipId device_id)`. This is a non-breaking additive change.

### DIFF
--- a/device/api/umd/device/chip/chip.hpp
+++ b/device/api/umd/device/chip/chip.hpp
@@ -115,6 +115,11 @@ public:
     virtual int get_clock() = 0;
     virtual int get_numa_node() = 0;
 
+    // Advance the chip by one clock cycle. Delegates to the underlying TTDevice, which
+    // is a no-op for chips without a controllable clock and drives the simulator clock
+    // synchronously (no background thread) for deterministic simulation.
+    void advance_device_execution();
+
     virtual int arc_msg(
         uint32_t msg_code,
         bool wait_for_done = true,
--- a/device/api/umd/device/tt_device/tt_device.hpp
+++ b/device/api/umd/device/tt_device/tt_device.hpp
@@ -350,6 +350,12 @@ public:
 
     virtual uint32_t get_min_clock_freq() = 0;
 
+    // Advance the device by one clock cycle. No-op by default; overridden by devices with a
+    // controllable clock (e.g. simulation). Simulator clocking must be deterministic, so the
+    // clock is advanced synchronously from the calling thread rather than driven by a
+    // background thread.
+    virtual void advance_device_execution();
+
     uint64_t get_board_id();
 
     uint8_t get_asic_location();
--- a/device/chip_helpers/tlb_manager.cpp
+++ b/device/chip_helpers/tlb_manager.cpp
@@ -45,10 +45,11 @@ void TLBManager::configure_tlb(tt_xy_pair core, size_t tlb_size, uint64_t addres
     config.static_vc = get_tt_device()->get_architecture_implementation()->get_static_vc();
     std::unique_ptr<TlbWindow> tlb_window = allocate_tlb_window(config, TlbMapping::WC, tlb_size);
 
+    // Simulation TTDevices have no PCI device; report chip 0 in the log line in that case.
     log_debug(
         LogUMD,
         "Configured TLB window for chip: {} core: {} size: {} address: {} ordering: {} tlb_id: {}",
-        tt_device_->get_pci_device()->get_device_num(),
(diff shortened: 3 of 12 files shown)


device/chip_helpers/tlb_manager.cpp
```diff
@@ -41,7 +41,7 @@ void TLBManager::configure_tlb(tt_xy_pair core, size_t tlb_size, uint64_t addres
     log_debug(
         LogUMD,
         "Configured TLB window for chip: {} core: {} size: {} address: {} ordering: {} tlb_id: {}",
-        tt_device_->get_pci_device()->get_device_num(),
+        tt_device_->get_pci_device() != nullptr ? tt_device_->get_pci_device()->get_device_num() : 0,
```
aleksamarkovicTT, line 52: Add a comment explaining why 0

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -328,6 +328,12 @@ class TTDevice {
 
     virtual uint32_t get_min_clock_freq() = 0;
 
+    // Advance the device by one clock cycle. No-op by default; overridden by devices with a
+    // controllable clock (e.g. simulation). Simulator clocking must be deterministic, so the
+    // clock is advanced synchronously from the calling thread rather than driven by a
+    // background thread.
+    virtual void yield() {}
```
aleksamarkovicTT: I don't like that this is in TTDevice, but I'm not sure where else it could be placed, given that there's no generic SimulationTTDevice for RTL, ttsim and swemule.

broskoTT: If we thing about our device, it behaves like a background thread, from high level perspective... And for ttsim we explicitly don't want that behavior. So this is an essential need for simulation which is not needed for real device.
For that reason, I'm fine with having something like this in the API. 

RTL is already driven in the background though, and SWemule is fast enough to always just (in some way) drive it on any call. So not sure if it will be used for anything else other than ttsim.

I'm fine with having this in the api, but "yield" is very very ambiguous.. Maybe something specific, like from your description which is much clearer. "advance_device_execution" or "advance_simulator" or "advance_simulator_execution"

broskoTT: Same for function in chip

broskoTT (approved):

device/api/umd/device/chip/chip.hpp
```diff
@@ -104,6 +104,15 @@ class Chip {
     virtual int get_clock() = 0;
     virtual int get_numa_node() = 0;
 
+    // Advance the chip by one clock cycle. Delegates to the underlying TTDevice, which
+    // is a no-op for chips without a controllable clock and drives the simulator clock
+    // synchronously (no background thread) for deterministic simulation.
+    void yield() {
+        if (auto* td = get_tt_device()) {
+            td->yield();
```
broskoTT: Let's not have implementation in hpp

device/api/umd/device/tt_device/tt_device.hpp
broskoTT: Let's not have implementation in hpp

PR #2474, opened 2026-04-20
Wait on telemetry initialization

broskoTT:
### Issue
https://tenstorrent.atlassian.net/browse/SYS-3873

### Description
Stumbled upon this while looking at https://tenstorrent.atlassian.net/browse/SYS-3873.
There is a possible race condition where we waited on ARC to report completion on booting, but the telemetry is not ready. This PR adds a path which also waits for the last telemetry entry to be filled up. This was already present on one path, so unified the approaches (it is slightly different for smbus telemetry and new telemetry).

Regarding the referenced issue, the issue grows deeper and telemetry won't ever become available.

### List of the changes
- add wait_for_telemetry_initialized to arc telemetry header
- Implement it for both variants
- Change get_firmware_version_util to use this new function
- Wait on this when initializing tt_device

### Testing
the path should be covered by ci tests. I haven't tested the case from the ticket, as I don't have access to the machine

### API Changes


device/api/umd/device/arc/arc_telemetry_reader.hpp
```diff
@@ -23,6 +24,12 @@ class ArcTelemetryReader {
 
     static std::unique_ptr<ArcTelemetryReader> create_arc_telemetry_reader(TTDevice* tt_device);
 
+    // Wait until the telemetry table is fully populated by ARC firmware.
+    // FLASH_BUNDLE_VERSION is the last entry written by ARC, so a non-zero value
+    // guarantees all other entries are also present.
+    // Retries reinitializing the telemetry table until the entry appears or timeout expires.
+    virtual void wait_for_telemetry_initialized(std::chrono::milliseconds timeout_ms = std::chrono::milliseconds(1000));
```
pjanevskiTT: nit: put 1000 in timeouts.hpp (or smh similar, not sure what is called)

device/api/umd/device/arc/arc_telemetry_reader.hpp
```diff
@@ -23,6 +24,12 @@ class ArcTelemetryReader {
 
     static std::unique_ptr<ArcTelemetryReader> create_arc_telemetry_reader(TTDevice* tt_device);
 
+    // Wait until the telemetry table is fully populated by ARC firmware.
+    // FLASH_BUNDLE_VERSION is the last entry written by ARC, so a non-zero value
+    // guarantees all other entries are also present.
+    // Retries reinitializing the telemetry table until the entry appears or timeout expires.
```
pjanevskiTT: I think for new style telemetry, register that contains telemetry address will be set to 0 until it is populated. Did you see that this is not the case? (That address is non-zero but data inside telemetry is garbage)

Can you try and confirm with someone from syseng that this is the case? I can try and find the link for that

device/arc/arc_telemetry_reader.cpp
```diff
@@ -98,4 +100,24 @@ bool ArcTelemetryReader::is_entry_available(const uint8_t telemetry_tag) {
     return telemetry_values.find(telemetry_tag) != telemetry_values.end();
 }
 
+void ArcTelemetryReader::wait_for_telemetry_initialized(std::chrono::milliseconds timeout_ms) {
+    auto start = std::chrono::steady_clock::now();
+    constexpr auto poll_interval = std::chrono::milliseconds(10);
+
+    while (!is_entry_available(TelemetryTag::FLASH_BUNDLE_VERSION) ||
```
pjanevskiTT: so do we need to change this logic every time we add another entry inside new telemetry?

device/firmware/firmware_utils.cpp
```diff
@@ -33,27 +33,13 @@ namespace tt::umd {
...
-        log_warning(
-            tt::LogUMD, "Timeout reading firmware bundle version (250ms), returning potentially invalid version");
+        SmBusArcTelemetryReader smbus_reader(tt_device);
+        smbus_reader.wait_for_telemetry_initialized();
         return FirmwareBundleVersion::from_firmware_bundle_tag(
-            smbus_telemetry_reader->read_entry(wormhole::LegacyTelemetryTag::FW_BUNDLE_VERSION));
+            smbus_reader.read_entry(wormhole::LegacyTelemetryTag::FW_BUNDLE_VERSION));
     }
     ArcTelemetryReader* telemetry = tt_device->get_arc_telemetry_reader();
+    telemetry->wait_for_telemetry_initialized();
```
pjanevskiTT: I see this needs to be called after constructor. Do we want to to put the wait inside the constructor, is there any use case where we want telemetry object without calling wait method?

device/api/umd/device/arc/arc_telemetry_reader.hpp
broskoTT: This was the related issue: https://tenstorrent.atlassian.net/browse/SYS-3873

device/arc/arc_telemetry_reader.cpp
broskoTT: I don't think so... I let claude analyze the codebase and this one is always populated last

device/firmware/firmware_utils.cpp
broskoTT: added to constructor

device/firmware/firmware_utils.cpp
```diff
@@ -33,27 +33,13 @@ namespace tt::umd {
...
-                smbus_telemetry_reader->read_entry(wormhole::LegacyTelemetryTag::FW_BUNDLE_VERSION);
-            if (fw_bundle_version != 0) {
-                return FirmwareBundleVersion::from_firmware_bundle_tag(fw_bundle_version);
-            }
-            std::this_thread::sleep_for(std::chrono::milliseconds(10));
-        }
-        log_warning(
-            tt::LogUMD, "Timeout reading firmware bundle version (250ms), returning potentially invalid version");
+        SmBusArcTelemetryReader smbus_reader(tt_device);
+        smbus_reader.wait_for_telemetry_initialized();
```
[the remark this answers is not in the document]
broskoTT: the exact timeout doesn't matter

device/arc/arc_telemetry_reader.cpp
broskoTT: Changed the logic after analysis, now it should be more general

device/tt_device/tt_device.cpp
```diff
@@ -114,6 +114,9 @@ void TTDevice::init_tt_device(const std::chrono::milliseconds timeout_ms) {
     wait_arc_core_start(timeout_ms);
     arc_messenger_ = ArcMessenger::create_arc_messenger(this);
     telemetry = ArcTelemetryReader::create_arc_telemetry_reader(this);
+    // Note: The timeout passed is not accounted for when waiting for telemetry initialization, but this is expected
+    // to be very quick.
+    telemetry->wait_for_telemetry_initialized();
```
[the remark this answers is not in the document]
broskoTT: outdated

device/arc/arc_telemetry_reader.cpp
```diff
@@ -26,25 +28,31 @@ static constexpr FirmwareBundleVersion FW_NEW_TELEMETRY = FirmwareBundleVersion(
...
+    TTDevice* tt_device, std::chrono::milliseconds timeout_ms) {
+    std::unique_ptr<ArcTelemetryReader> reader;
     switch (tt_device->get_arch()) {
         case tt::ARCH::WORMHOLE_B0: {
             FirmwareBundleVersion fw_bundle_version = get_firmware_version_util(tt_device);
 
             if (fw_bundle_version >= FW_NEW_TELEMETRY) {
                 log_debug(tt::LogUMD, "Creating new-style telemetry reader.");
-                return std::make_unique<WormholeArcTelemetryReader>(tt_device);
+                reader = std::make_unique<WormholeArcTelemetryReader>(tt_device);
```
nbuncicTT, line 41: Why would we silently fallback to old style telemetry? What's the use case here? If it's not the expected behavior, I'd omit the fallback and fail (i.e. throw).

device/arc/arc_telemetry_reader.cpp
```diff
@@ -102,4 +110,21 @@ bool ArcTelemetryReader::is_entry_available(const uint8_t telemetry_tag) {
     return telemetry_values.find(telemetry_tag) != telemetry_values.end();
 }
 
+void ArcTelemetryReader::wait_for_telemetry_initialized(std::chrono::milliseconds timeout_ms) {
+    auto start = std::chrono::steady_clock::now();
+    constexpr auto poll_interval = std::chrono::milliseconds(10);
+
+    get_telemetry_address();
+    while (telemetry_table_addr_reg == 0) {
```
nbuncicTT: We have the utilities `check_timeout`/`poll_until`, can you use one of them instead of doing this manually?

nbuncicTT (approved):
Nice catch

device/arc/arc_telemetry_reader.cpp
broskoTT: well it's separating fw version which implemented new telemetry on wormhole, 18.4.0. It's not a fallback in a sense that it will create old telemetry if something fails for new one.
And this PR doesn't change that behavior...

PR #2478, opened 2026-04-21
Add NEO DM uncore reset commands for RTL simulation

pjanevskiTT:
### Issue
/

### Description
Add support for NEO DM uncore reset assert/deassert commands in the RTL simulation
communicator, with both per-core and all-cores variants.

### List of the changes
- Add 4 new DEVICE_COMMAND enum values to the FlatBuffers schema (14-17)
- Add 4 new methods to RtlSimCommunicator for uncore reset assert/deassert
- Add RiscType::ALL_NEO_DMS_UNCORE and RiscType::NEO_DM_UNCORE flags
- Wire up uncore reset in RtlSimulationTTDevice::assert_risc_reset/deassert_risc_reset

### Testing
CI

### API Changes
This PR has API changes:
- New RiscType flags: ALL_NEO_DMS_UNCORE, NEO_DM_UNCORE
- New RtlSimCommunicator methods for uncore reset

### DIFF
--- a/device/api/umd/device/types/risc_type.hpp
+++ b/device/api/umd/device/types/risc_type.hpp
@@ -83,6 +83,10 @@ enum class RiscType : std::uint64_t {
     ALL_NEO_TRISCS = ALL_NEO0_TRISCS | ALL_NEO1_TRISCS | ALL_NEO2_TRISCS | ALL_NEO3_TRISCS,
     ALL_NEO_DMS = DM0 | DM1 | DM2 | DM3 | DM4 | DM5 | DM6 | DM7,
     ALL_NEO = ALL_NEO_TRISCS | ALL_NEO_DMS,
+
+    // Uncore reset flags for NEO DM cores.
+    ALL_NEO_DMS_UNCORE = 1ULL << 32,
+    NEO_DM_UNCORE = 1ULL << 33,
 };
 
 std::string RiscTypeToString(RiscType value);
--- a/device/simulation/simulation_device.fbs
+++ b/device/simulation/simulation_device.fbs
@@ -17,6 +17,10 @@ enum DEVICE_COMMAND : byte {
     // to match these values before this can work end-to-end.
     AXI_RAM_WRITE_NOTIFICATION = 10,
     AXI_RAM_READ_NOTIFICATION = 11,
+    ALL_NEO_DMS_UNCORE_RESET_DEASSERT = 14,
+    ALL_NEO_DMS_UNCORE_RESET_ASSERT = 15,
+    NEO_DM_UNCORE_RESET_ASSERT = 16,
+    NEO_DM_UNCORE_RESET_DEASSERT = 17,
 }
 
 struct tt_vcs_core {
--- a/device/tt_device/rtl_simulation_tt_device.cpp
+++ b/device/tt_device/rtl_simulation_tt_device.cpp
@@ -145,8 +145,15 @@ void RtlSimulationTTDevice::assert_risc_reset(tt_xy_pair core, const RiscType se
     // If the architecture is Quasar, a special case is needed to control the NEO Data Movement cores.
     if (get_soc_descriptor().arch == tt::ARCH::QUASAR) {
         if (selected_riscs == RiscType::ALL_NEO_DMS) {
-            // Reset all DM cores.
             communicator_->all_neo_dms_reset_assert(core.x, core.y);
+            return;
+        }
+        if (selected_riscs == RiscType::ALL_NEO_DMS_UNCORE) {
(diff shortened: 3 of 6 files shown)


device/api/umd/device/types/risc_type.hpp
```diff
@@ -83,6 +83,10 @@ enum class RiscType : std::uint64_t {
     ALL_NEO_TRISCS = ALL_NEO0_TRISCS | ALL_NEO1_TRISCS | ALL_NEO2_TRISCS | ALL_NEO3_TRISCS,
     ALL_NEO_DMS = DM0 | DM1 | DM2 | DM3 | DM4 | DM5 | DM6 | DM7,
     ALL_NEO = ALL_NEO_TRISCS | ALL_NEO_DMS,
+
+    // Uncore reset flags for NEO DM cores.
+    ALL_NEO_DMS_UNCORE = 1ULL << 32,
+    NEO_DM_UNCORE = 1ULL << 33,
```
broskoTT, line 89: ALL_NEO_DMS_UNCORE = ALL_NEO_DMS | NEO_DM_UNCORE ?
In that case shift NEO_DM_UNCORE = 1ULL << 32

device/simulation/simulation_device.fbs
```diff
@@ -17,6 +17,10 @@ enum DEVICE_COMMAND : byte {
     // to match these values before this can work end-to-end.
     AXI_RAM_WRITE_NOTIFICATION = 10,
     AXI_RAM_READ_NOTIFICATION = 11,
+    ALL_NEO_DMS_UNCORE_RESET_DEASSERT = 14,
```
broskoTT, line 20: Is this command NEO_DM_UNCORE_RESET + reset on all DMs?
Is it strictly different than calling them separately?

I see we already have a similar case for ALL_NEO_DMS_RESET_DEASSERT and NEO_DM_RESET_DEASSERT...

In any case, even though there is a special case for how simulator is implemented, I think we don't need to separate this in our interface, as suggested in the other comment...

device/tt_device/rtl_simulation_tt_device.cpp
```diff
@@ -141,11 +141,24 @@ void RtlSimulationTTDevice::assert_risc_reset(tt_xy_pair core, const RiscType se
...
+            communicator_->all_neo_dms_uncore_reset_assert();
+            return;
+        }
         if (selected_riscs == RiscType::ALL_NEO_DMS) {
-            // Reset all DM cores.
             communicator_->all_neo_dms_reset_assert(core.x, core.y);
             return;
         }
+        if (selected_riscs == RiscType::ALL_NEO_DMS_UNCORE) {
+            communicator_->all_neo_dms_uncore_reset_assert();
```
broskoTT, line 158: This is not per core?

device/tt_device/rtl_simulation_tt_device.cpp
```diff
@@ -141,11 +141,24 @@ void RtlSimulationTTDevice::assert_risc_reset(tt_xy_pair core, const RiscType se
     log_debug(tt::LogEmulationDriver, "Sending 'assert_risc_reset' signal for risc_type {}", selected_riscs);
     // If the architecture is Quasar, a special case is needed to control the NEO Data Movement cores.
     if (soc_descriptor_.arch == tt::ARCH::QUASAR) {
+        if (selected_riscs == RiscType::ALL) {
```
broskoTT, line 147: I think the flow would look better with slightly different logic here, although won't block on it.

If ALL and QUASAR, selected_riscs = RISCTYPE::ALL_NEO

Then you don't need this if here, rather passing through bellow ifs should handle all the resets nicely regarding ALL_NEO_DMS and ALL_NEO_TRISCS, you'd also have to remove the return statements

broskoTT (approved):

device/tt_device/rtl_simulation_tt_device.cpp
pjanevskiTT: it's on all cores, not per core

device/api/umd/device/types/risc_type.hpp
pjanevskiTT: UNCORE is just for UNCORE part, that is why we have ALL_RISCS as well

device/simulation/simulation_device.fbs
pjanevskiTT: it's just a different enum, the client would use the same API just with different parameters

device/tt_device/rtl_simulation_tt_device.cpp
pjanevskiTT: NEO and uncore need to be resetted by separate API, so will keep the flow as it is

PR #2483, opened 2026-04-22
Remove SocDescriptor hack in TopologyDiscovery

aleksamarkovicTT:
### Issue
#1741 

### Description
Remove SocDescriptor hack in TopologyDiscovery. Since TTDevice now owns a SocDescriptor, TopologyDiscovery no longer needs to keep track of its own copies.

### Testing
CI

### API Changes

### DIFF
--- a/device/topology/topology_discovery_blackhole.cpp
+++ b/device/topology/topology_discovery_blackhole.cpp
@@ -33,15 +33,7 @@ TopologyDiscoveryBlackhole::TopologyDiscoveryBlackhole(
 
 std::unique_ptr<TTDevice> TopologyDiscoveryBlackhole::create_remote_device(
     std::optional<EthCoord> eth_coord, TTDevice* gateway_device, std::set<uint32_t> gateway_eth_channels) {
-    // ETH coord is not used for Blackhole, as Blackhole does not have a concept of ETH coordinates.
-    std::unique_ptr<RemoteCommunication> remote_communication =
-        RemoteCommunication::create_remote_communication(gateway_device, {0, 0, 0, 0});
-    remote_communication->set_remote_transfer_ethernet_cores(
-        get_soc_descriptor(gateway_device)
-            .get_eth_xy_pairs_for_channels(gateway_eth_channels, CoordSystem::TRANSLATED));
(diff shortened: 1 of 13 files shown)


device/topology/topology_discovery_blackhole.cpp
```diff
@@ -33,15 +33,7 @@ TopologyDiscoveryBlackhole::TopologyDiscoveryBlackhole(
...
-    // ETH coord is not used for Blackhole, as Blackhole does not have a concept of ETH coordinates.
-    std::unique_ptr<RemoteCommunication> remote_communication =
-        RemoteCommunication::create_remote_communication(gateway_device, {0, 0, 0, 0});
-    remote_communication->set_remote_transfer_ethernet_cores(
-        get_soc_descriptor(gateway_device)
-            .get_eth_xy_pairs_for_channels(gateway_eth_channels, CoordSystem::TRANSLATED));
-    std::unique_ptr<TTDevice> remote_tt_device = TTDevice::create(std::move(remote_communication));
-    remote_tt_device->init_tt_device();
-    return remote_tt_device;
+    return nullptr;
```
nbuncicTT, line 36: Just a question, why was this, on the right-hand side ever here? Why wasn't it returning `nullptr` previously?

nbuncicTT (approved):
Well done.

device/topology/topology_discovery_blackhole.cpp
```diff
@@ -33,15 +33,7 @@ TopologyDiscoveryBlackhole::TopologyDiscoveryBlackhole(
 
 std::unique_ptr<TTDevice> TopologyDiscoveryBlackhole::create_remote_device(
     std::optional<EthCoord> eth_coord, TTDevice* gateway_device, std::set<uint32_t> gateway_eth_channels) {
-    // ETH coord is not used for Blackhole, as Blackhole does not have a concept of ETH coordinates.
```
pjanevskiTT, line 36: can we make this change a separate PR?

device/topology/topology_discovery_blackhole.cpp
aleksamarkovicTT: We had lite fabric support but it was removed in a way so this function never got called. TTDevice::create would return a nullptr anyway.

pjanevskiTT (approved):

Issue #2449, opened 2026-04-09
UMD bump breaking tt-exalens CI


adjordjevic-TT:
Recent bump of UMD version (`0.9.4.260326` -> `0.9.5.dev260402`) with this [PR](https://github.com/tenstorrent/tt-exalens/commit/b5dbb3e5fa243916ba6c4c6bc983df1897fd8698) caused tt-exalens CI to fail.

Reading address `0xffb121b0`through NOC1 on blackhole p150b triggers our heuristic for bad device state (read takes unusually long time (more than 2ms) and returns `0xFFFFFFFF`). CI was run multiple times and failed every time. Unsure how this even got merged since this is required workflow.
Failed run: https://github.com/tenstorrent/tt-exalens/actions/runs/24194229234/job/70636524185
After reverting UMD bump run passed: https://github.com/tenstorrent/tt-exalens/actions/runs/24195806157/job/70625667836

Could not reproduce this locally.

PR #2512, opened 2026-04-24
Run tt-exalens tests with NOC1 on both tests

nbuncicTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/2449

### Description
Delete hack from PR https://github.com/tenstorrent/tt-umd/pull/2467 and don't use O_APPEND until it's investigated why it's making a problem.

### List of the changes
- Removed O_APPEND usage
- Removed hack for workflow

### Testing
CI testing with 50 consecutive passes on the tests that were failing: https://github.com/tenstorrent/tt-umd/actions/runs/24941251006

### API Changes
/

### DIFF
--- a/.github/workflows/run-exalens-tests.yml
+++ b/.github/workflows/run-exalens-tests.yml
@@ -96,12 +96,6 @@ jobs:
           TTEXALENS_TESTS_USE_NOC1: 1
         shell: bash
         run: |
-          # HACK: By running tt-exalens application in the background,
-          # we prevent KMD from setting the devices to low power mode.
-          # Exalens tests recreate context very often, which causes KMD to
-          # oscillate between high power and low power mode, causing some
-          # transient hangs on NOC1.
-          python3 tt-exalens.py > /dev/null 2>&1 &
           python3 -m unittest discover -v -t . -s test/ttexalens -p *test*.py
 
       - name: Run Python tests for tt-exalens app (NOC1)
--- a/device/pcie/pci_device.cpp
+++ b/device/pcie/pci_device.cpp
@@ -403,14 +403,9 @@ std::optional<int> PCIDevice::get_pci_device_id(int umd_logical_id) {
 }
 
 static int open_pci_device(const std::string &device_path) {
-    // O_APPEND opts out of legacy mode in KMD >= 2.6.0, allowing the device to enter low-power idle states.
+    // Force legacy mode (no O_APPEND) to test if O_APPEND causes NOC1 issues.
     int flags = O_RDWR | O_CLOEXEC;
-    if (PCIDevice::read_kmd_version() >= KMD_POWER_STATE && PCIDevice::get_pcie_arch() == tt::ARCH::BLACKHOLE) {
-        log_debug(LogUMD, fmt::format("Opening device {} in power aware mode.", device_path));
-        flags |= O_APPEND;
-    } else {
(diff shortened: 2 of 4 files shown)


broskoTT (approved):

.github/workflows/run-exalens-tests.yml
```diff
@@ -91,21 +91,23 @@ jobs:
...
-          # oscillate between high power and low power mode, causing some
-          # transient hangs on NOC1.
           python3 tt-exalens.py > /dev/null 2>&1 &
           python3 -m unittest discover -v -t . -s test/ttexalens -p *test*.py
 
       - name: Run Python tests for tt-exalens app (NOC1)
         env:
           TTEXALENS_TESTS_USE_NOC1: 1
         run: |
+          python3 tt-exalens.py > /dev/null 2>&1 &
```
broskoTT: Consider just adding this to all tests

nbuncicTT: done.

.github/workflows/run-exalens-tests.yml
```diff
@@ -83,13 +83,22 @@ jobs:
...
 
+      # MITIGATION: Look as the HACK explanation in the lower part of the file.
+      # The behavior was never observed on NOC0, but if it's a race condition,
+      # it may just be very hard to hit.
       - name: Run Python tests for tt-exalens library
         run: |
+          python3 tt-exalens.py > /dev/null 2>&1 &
+          EXALENS_PID=$!
           python3 -m unittest discover -v -t . -s test/ttexalens -p *test*.py
+          kill $EXALENS_PID 2>/dev/null || true
```
broskoTT: I think we don't need to run this in each step (I know I might've said otherwise earlier).

Seems like you can "leak" the process from the previous step? As it, all steps run in the same environment.

Here's what gemini suggests:

Option B: The "GitHub Way" (Cleanest)
Split the startup and the cleanup into different steps.

YAML
      - name: Start Background Library
        run: python3 tt-exalens.py > /dev/null 2>&1 & echo $! > .exalens_pid

      - name: Run Python tests
        run: python3 -m unittest discover -v -t . -s test/ttexalens -p *test*.py

      - name: Cleanup
        if: always()
        run: |
          PID=$(cat .exalens_pid)
          kill $PID 2>/dev/null || true

aleksamarkovicTT: There's no need to copy&paste this mitigation to every test, it doesn't help to prevent the problem as we saw in the previous run. I'll try adding sleep to wait out the exalens app startup but I doubt it will do anything.

device/pcie/pci_device.cpp
```diff
@@ -403,14 +403,9 @@ std::optional<int> PCIDevice::get_pci_device_id(int umd_logical_id) {
 }
 
 static int open_pci_device(const std::string &device_path) {
-    // O_APPEND opts out of legacy mode in KMD >= 2.6.0, allowing the device to enter low-power idle states.
+    // Force legacy mode (no O_APPEND) to test if O_APPEND causes NOC1 issues.
```
nbuncicTT: Note: Going to add a comment here on why we don't use O_APPEND and link it to an issue.

.github/workflows/run-exalens-tests.yml
nbuncicTT: Managed the mitigation another way, so I'll resolve this conversation as it's now a different change.

aleksamarkovicTT (approved):

PR #2515, opened 2026-04-24
TTSimTTDevice: make it a complete TTDevice for LLK test use

pavlejosipovic:
## Summary

Four fixes that together let tt-exalens (and by extension the LLK test harness) drive a TTSim simulation end-to-end. Validated on vanilla ttsim v1.5.4 (\`libttsim_{bh,wh}.so\` from [github.com/tenstorrent/ttsim/releases](https://github.com/tenstorrent/ttsim/releases)) — BH and WH LLK smoke tests now pass with \`BootMode.BRISC\` default, no arch-specific workarounds required.

- **Populate the base-class \`arch\` field** from the soc descriptor in the constructor. TTSim does not go through \`init_tt_device()\` (no PCI probe), so without this \`arch\` stays \`tt::ARCH::Invalid\` and downstream consumers (e.g. tt-exalens constructing a \`SocDescriptor\` from the device) see the wrong arch.
- **Return \`false\` from \`get_noc_translation_enabled()\`** instead of throwing. TTSim addresses cores with logical/virtual coordinates end-to-end; NOC translation is never applied.
- **Override \`get_chip_info()\`** to return defaults consistent with \`TTSimTTDevice::create()\`. On Blackhole the \`SocDescriptor\` constructor rejects an empty \`eth_harvesting_mask\` (\"Exactly 2 or 14 ETH cores should be harvested on full Blackhole\"), so set the \`0x120\` default here too.
- **Advance the simulated clock in \`write_to_device()\`.** Without this, host sequences like \"load BRISC ELF, deassert BRISC, write first command mailbox\" all land at the same simulated timestamp — the newly deasserted core doesn't execute any instructions between deassert and the subsequent command-mailbox write, so its own CRT init (which clears the mailbox at startup) clobbers host-written commands and the handshake hangs. Mirror the read path's auto-clock, with a floor of 1000 cycles so a 4-byte deassert write still gives the just-started core enough time to get past CRT into \`main()\` before the next host operation.

Companion PRs: tenstorrent/tt-exalens \`pjosipovic/ttsim-llk-support\`, tenstorrent/tt-metal \`pjosipovic/llk-tests-ttsim\`.

## Test plan

- [x] \`test_eltwise_unary_datacopy\` (3x Float16_b variants) passes on BH + WH against vanilla ttsim v1.5.4
- [ ] Existing tt-umd CI (silicon + RTL sim paths unchanged)


device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -99,6 +103,13 @@ void TTSimTTDevice::write_to_device(const void* mem_ptr, tt_xy_pair core, uint64
     } else {
         communicator_->tile_write_bytes(core.x, core.y, addr, mem_ptr, size);
     }
+    // Advance the sim on host writes. Without this, sequences like "load BRISC ELF, deassert
```
pjanevskiTT, line 106: > so BRISC's clear of
>     // brisc_command_buffer at startup clobbers the host's command and the handshake hangs.

if this is specific for LLK test infra, I would maybe change the behavior such that code in test infra can check when BRISC has cleated the buffer and is ready to receive commands (which you would do by doing read_from_device, hence clocking the simulator). I am not against clocking the sim on write as well, I think that is fundamentally ok
I know @mcraigheadTT will have his worries for sim performance in this case, but maybe we need more general way of clocking the sim. 

Leaving a comment for you to consider if you can change LLK test infra code to avoid adding clocking here, if not feel free to check it in and we can work on more general sim clocking in the future

pjanevskiTT (approved):
Left one comment to consider, otherwise changes look good, leaving approve to unblock

device/tt_device/tt_sim_tt_device.cpp
pavlejosipovic: you are right but I'm gonna go with this for now, as it doesn't require ttsim changes

PR #2517, opened 2026-04-24
Move silicon SysmemManager ownership into TTDevice (unified)

pjanevskiTT:
### Issue
/

### Description

Stacked on #2516 (TLBManager unification). Completes the pattern for `SysmemManager`: every `TTDevice` flavor — silicon PCIe, `TTSim`, RTL sim — stores its `SysmemManager` in one base-class member, and the `Chip` layer forwards instead of owning.

#### Background

Before this PR the ownership split was:
- **Silicon:** `LocalChip::sysmem_manager_` (a `std::unique_ptr<SysmemManager>` holding a `SiliconSysmemManager`), constructed in `LocalChip::create` when the `TTDevice` is PCIe. `SiliconSysmemManager` needs a `TLBManager*` plus `num_host_mem_channels`.
- **Simulation:** `TTSimTTDevice::sysmem_manager_` / `RtlSimulationTTDevice::sysmem_manager_` (each a `std::unique_ptr<SimulationSysmemManager>`), constructed in the sim subclass ctor from `num_host_mem_channels` + `soc_descriptor.arch`.

#2476 added `virtual SimulationSysmemManager* get_sysmem_manager() { return nullptr; }` to `TTDevice` so simulation could be reached polymorphically. That return type made silicon polymorphism impossible (`SiliconSysmemManager` is a sibling of `SimulationSysmemManager`, not derived).

#### What changes here

- `TTDevice` base gains `std::unique_ptr<SysmemManager> sysmem_manager_` as the single home. Silicon stores a `SiliconSysmemManager` there, simulation stores a `SimulationSysmemManager` there (implicit upcast, public inheritance).
- `TTDevice::get_sysmem_manager()` — added by #2476 as a `virtual` returning `SimulationSysmemManager*` / `nullptr` — becomes a non-virtual accessor returning `SysmemManager*` (base type, so both flavors fit). No subclass overrides it anymore.
- `num_host_mem_channels` is now threaded through `TTDevice::create(int device_number, IODeviceType, bool use_safe_api, int num_host_mem_channels = 0)` and the PCIe `TTDevice` / `Wormhole` / `Blackhole` ctors. `LocalChip::create` — which already has the value — passes it through. Default `0` keeps all other callers (`warm_reset`, `topology_discovery`, test helpers) working unchanged.
- The PCIe `TTDevice` ctor now constructs `SiliconSysmemManager` immediately after `TLBManager` (which it depends on).
- Simulation subclass ctors move their `SimulationSysmemManager` construction from the initializer list into the ctor body — can't initialize an inherited base member from a derived initializer list. The construction is placed before any code that reads `sysmem_manager_` (for `TTSimTTDevice`: before `initialize_sysmem_functions`; for `RtlSimulationTTDevice`: a local `SimulationSysmemManager*` handle keeps the ram-callbacks block unchanged, then moves into the base slot).
- `LocalChip` loses the field + ctor param + dtor reset; `LocalChip::get_sysmem_manager()` forwards to `tt_device_`; six internal direct-access call sites (`start_device`, `close_device`, `get_num_host_channels`, `get_host_channel_size`, `write_to_sysmem`, `read_from_sysmem`, `init_pcie_iatus`) are rewired through the forwarder with a cached local to avoid repeated getter calls.

#### Include-cycle fix

Adding `#include "umd/device/chip_helpers/sysmem_manager.hpp"` to `tt_device.hpp` surfaced an existing circular include: `sysmem_manager.hpp` was pulling `tt_device.hpp` to use `TTDevice*`. `TTDevice*` in that header only needs a forward decl, so `sysmem_manager.hpp` now does `class TTDevice;` and the `.cpp` files that actually dereference `tt_device_` (`sysmem_manager.cpp`, `silicon_sysmem_manager.cpp`) include `tt_device.hpp` directly.

#### Commits

1. **`Move silicon SysmemManager ownership from LocalChip to TTDevice`** — introduces the base member, populates it from the PCIe ctor, threads `num_host_mem_channels`, fixes the circular include, drops `LocalChip`'s field + call sites. Simulation subclasses untouched.
2. **`Unify SysmemManager ownership on TTDevice, drop virtual accessor`** — migrates both simulation subclasses to the base member, removes their field and override, and de-virtualizes `TTDevice::get_sysmem_manager()`.

### List of the changes

- `device/api/umd/device/tt_device/tt_device.hpp`: add `#include "umd/device/chip_helpers/sysmem_manager.hpp"`; remove `class SimulationSysmemManager;` forward decl; add protected `std::unique_ptr<SysmemManager> sysmem_manager_`; change `get_sysmem_manager()` return type to `SysmemManager*`, body to return the member, drop `virtual`; add `int num_host_mem_channels = 0` to the PCIe ctor signature and to the `TTDevice::create(...)` factory signature.
- `device/tt_device/tt_device.cpp`: populate `sysmem_manager_` in the PCIe ctor after `tlb_manager_`; update `TTDevice::create` to pass `num_host_mem_channels` through to `Wormhole/BlackholeTTDevice`.
- `device/api/umd/device/tt_device/wormhole_tt_device.hpp` + `device/tt_device/wormhole_tt_device.cpp`: add `int num_host_mem_channels = 0` to PCIe ctor and forward to base; update friend decl of `TTDevice::create`.
- `device/api/umd/device/tt_device/blackhole_tt_device.hpp` + `device/tt_device/blackhole_tt_device.cpp`: same.
- `device/api/umd/device/chip_helpers/sysmem_manager.hpp`: replace `#include "umd/device/tt_device/tt_device.hpp"` with a forward decl `class TTDevice;` to break the new include cycle.
- `device/chip_helpers/sysmem_manager.cpp` + `device/chip_helpers/silicon_sysmem_manager.cpp`: add `#include "umd/device/tt_device/tt_device.hpp"` (previously transitive).
- `device/api/umd/device/chip/local_chip.hpp`: remove `sysmem_manager_` field and ctor param.
- `device/chip/local_chip.cpp`: drop `SiliconSysmemManager` construction in `LocalChip::create`; source `sysmem_ptr` from `tt_device->get_sysmem_manager()`; drop ctor param/init and dtor reset; make `LocalChip::get_sysmem_manager()` forward; rewire the six internal call sites with cached locals.
- `device/api/umd/device/tt_device/tt_sim_tt_device.hpp` + `.cpp`: remove `sysmem_manager_` field and `get_sysmem_manager()` override; move construction into ctor body before `initialize_sysmem_functions()`.
- `device/api/umd/device/tt_device/rtl_simulation_tt_device.hpp` + `.cpp`: same, using a local typed handle to preserve the ram-callbacks lambda captures.

### Testing

CI. Locally: `cmake --build build` + `pre-commit run --all-files` pass after each commit.

No new tests — pure ownership refactor, behavior unchanged. Existing silicon coverage (`Cluster::get_sysmem_manager()`, `LocalChip` sysmem APIs) and simulation coverage (`TTSimChip` / `RtlSimulationChip` forwarders) exercise both paths through the same non-virtual accessor.

### API Changes

Non-breaking internal changes:
- `TTDevice::get_sysmem_manager()` return type changes from `SimulationSysmemManager*` (as #2476 introduced) to `SysmemManager*`. Callers that were already dealing in `SysmemManager*` (via the sim overrides' covariant return) still compile.
- `TTDevice::get_sysmem_manager()` loses `virtual`. No remaining overrides.
- `TTDevice::create(...)` and `TTDevice(PCIDevice, ...)` ctor gain a trailing `int num_host_mem_channels = 0` param. Default keeps existing callers source-compatible.
- `WormholeTTDevice` / `BlackholeTTDevice` PCIe ctors gain the same trailing param. Not part of the public API (protected, constructed via `TTDevice::create` friend).
- `LocalChip`'s private ctor signature loses the `std::unique_ptr<SysmemManager>` param. Not public.

No header exports or public-facing types change shape.

### DIFF
--- a/device/api/umd/device/chip/local_chip.hpp
+++ b/device/api/umd/device/chip/local_chip.hpp
@@ -80,11 +80,9 @@ private:
     LocalChip(
         SocDescriptor soc_descriptor,
         std::unique_ptr<TTDevice> tt_device,
-        std::unique_ptr<SysmemManager> sysmem_manager,
         std::unique_ptr<RemoteCommunication> remote_communication,
         int num_host_mem_channels);
 
-    std::unique_ptr<SysmemManager> sysmem_manager_;
     LockManager lock_manager_;
     // Used only for ethernet broadcast to all remote chips.
     std::unique_ptr<RemoteCommunication> remote_communication_;
--- a/device/api/umd/device/chip_helpers/sysmem_manager.hpp
+++ b/device/api/umd/device/chip_helpers/sysmem_manager.hpp
@@ -6,11 +6,12 @@
 
 #include "sysmem_buffer.hpp"
 #include "umd/device/chip_helpers/sysmem_buffer.hpp"
-#include "umd/device/tt_device/tt_device.hpp"
 #include "umd/device/types/arch.hpp"
 #include "umd/device/types/cluster_types.hpp"
 
 namespace tt::umd {
+
+class TTDevice;
 
 class SysmemManager {
 public:
--- a/device/api/umd/device/tt_device/blackhole_tt_device.hpp
+++ b/device/api/umd/device/tt_device/blackhole_tt_device.hpp
@@ -45,7 +45,7 @@ public:
     EthTrainingStatus read_eth_core_training_status(tt_xy_pair eth_core) override;
 
 protected:
-    BlackholeTTDevice(std::unique_ptr<PCIDevice> pci_device, bool use_safe_api);
+    BlackholeTTDevice(std::unique_ptr<PCIDevice> pci_device, bool use_safe_api, int num_host_mem_channels = 0);
     BlackholeTTDevice(std::unique_ptr<JtagDevice> jtag_device, uint8_t jlink_id);
 
     virtual bool is_arc_available_over_axi();
@@ -58,7 +58,8 @@ protected:
 private:
     int get_pcie_x_coordinate();
 
-    friend std::unique_ptr<TTDevice> TTDevice::create(int device_number, IODeviceType device_type, bool use_safe_api);
+    friend std::unique_ptr<TTDevice> TTDevice::create(
+        int device_number, IODeviceType device_type, bool use_safe_api, int num_host_mem_channels);
 
     static constexpr uint64_t ATU_OFFSET_IN_BH_BAR2 = 0x1000;
     std::set<size_t> iatu_regions_;
--- a/device/api/umd/device/tt_device/rtl_simulation_tt_device.hpp
+++ b/device/api/umd/device/tt_device/rtl_simulation_tt_device.hpp
@@ -59,8 +59,6 @@ public:
 
     RtlSimCommunicator* get_communicator() { return communicator_.get(); }
 
-    SimulationSysmemManager* get_sysmem_manager() override { return sysmem_manager_.get(); }
-
 protected:
     void retrain_dram_core(const uint32_t dram_channel) override;
 
@@ -69,7 +67,6 @@ private:
     std::recursive_mutex device_lock;
 
     std::filesystem::path simulator_directory_;
-    std::unique_ptr<SimulationSysmemManager> sysmem_manager_;
     std::unique_ptr<TlbWindow> cached_tlb_window_;
 };
 }  // namespace tt::umd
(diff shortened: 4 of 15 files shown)


pjanevskiTT:
deprecated by multiple other PRs supporting this refactor

PR #2518, opened 2026-04-24
Automatically set number of host mem channels in Cluster constructor

aleksamarkovicTT:
### Issue
/

### Description
Automatically set number of host mem channels in Cluster constructor. Initial work for removing passing already created ClusterDescriptors to Cluster for SILICON chip type.

### List of the changes
- Renamed `g_MAX_HOST_MEM_CHANNELS` to `MAX_HOST_MEM_CHANNELS`.
- Made `ClusterOptions::num_host_mem_ch_per_mmio_device` an optional.
- Automatically determine num_host_mem_ch_per_mmio_device by max amount of chips connected through one PCIe interface in the ClusterDescriptor.

### Testing
CI

### API Changes
This PR has API changes.

### DIFF
--- a/device/api/umd/device/cluster.hpp
+++ b/device/api/umd/device/cluster.hpp
@@ -62,8 +62,10 @@ struct ClusterOptions {
 
     /**
      * Number of host memory channels (hugepages) per MMIO device.
+     * If not provided, the value is determined automatically by determining the max
+     * amount of chips connected through one PCIe interface in the ClusterDescriptor.
      */
-    uint32_t num_host_mem_ch_per_mmio_device = 0;
+    std::optional<uint32_t> num_host_mem_ch_per_mmio_device = 0;
 
     /**
      * If set, this soc descriptor will be used to construct devices on this cluster. If not set, the default soc
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -395,6 +395,16 @@ Cluster::Cluster(ClusterOptions options) {  // NOLINT(performance-unnecessary-va
             UMD_THROW(error::RuntimeError, "Unsupported chip type.");
     }
 
+    if (!options.num_host_mem_ch_per_mmio_device.has_value()) {
+        auto grouped_chips = cluster_desc->get_chips_grouped_by_closest_mmio();
+        uint32_t max_chips_per_mmio = 0;
+        for (const auto& [mmio_device_id, chips] : grouped_chips) {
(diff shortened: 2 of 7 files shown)


device/api/umd/device/cluster.hpp
```diff
@@ -62,8 +62,10 @@ struct ClusterOptions {
 
     /**
      * Number of host memory channels (hugepages) per MMIO device.
+     * If not provided, the value is determined automatically by determining the max
+     * amount of chips connected through one PCIe interface in the ClusterDescriptor.
      */
-    uint32_t num_host_mem_ch_per_mmio_device = 0;
+    std::optional<uint32_t> num_host_mem_ch_per_mmio_device = 0;
```
broskoTT, line 68: Well actually this won't work :) You probably wanted it set to nullopt by default

device/cluster.cpp
```diff
@@ -396,6 +395,16 @@ Cluster::Cluster(ClusterOptions options) {  // NOLINT(performance-unnecessary-va
             UMD_THROW(error::RuntimeError, "Unsupported chip type.");
     }
 
+    if (!options.num_host_mem_ch_per_mmio_device.has_value()) {
+        auto grouped_chips = cluster_desc->get_chips_grouped_by_closest_mmio();
+        uint32_t max_chips_per_mmio = 0;
+        for (const auto& [mmio_device_id, chips] : grouped_chips) {
+            max_chips_per_mmio = std::max(max_chips_per_mmio, static_cast<uint32_t>(chips.size()));
+        }
+        options.num_host_mem_ch_per_mmio_device = std::min(MAX_HOST_MEM_CHANNELS, max_chips_per_mmio);
```
broskoTT, line 404: Just to confirm, on today's system this would result to 1GB for n150 and 2GB for N300?
What's the motivation for adding this logic? Is this making the client's code simpler?

device/api/umd/device/cluster.hpp
aleksamarkovicTT: No, it's intentional to retain existing behavior.

device/cluster.cpp
aleksamarkovicTT: Yes and yes. Note that if there is 1xn300, every device will get 2GB. This is existing behavior in metal.

pjanevskiTT (approved):
Looks good, I think this will simplify tt-metal hose code, maybe linking a PR would make it easier to understand how

PR #2522, opened 2026-04-27
Adding SMN support

adjordjevic-TT:
### Issue
/

### Description
Added support for SMN reads and writes.
**Disclaimer**: Simulator with SMN is still not official version, added comment explaining that, will remove once this gets into official tag.

### List of the changes
- Extended `DEVICE_COMMAND` enum with `SMN_WRITE=12` and `SMN_READ=13` matching simulator with SMN support
- Added logic to `tile_read_bytes` and `tile_write_bytes` in `RtlSimCommunicator` to know which `DEVICE_COMMAND` to sent based on selected `NocId`.

### Testing
/

### API Changes

### DIFF
--- a/device/api/umd/device/simulation/rtl_sim_communicator.hpp
+++ b/device/api/umd/device/simulation/rtl_sim_communicator.hpp
@@ -15,7 +15,6 @@
 #include <thread>
 
 #include "umd/device/simulation/simulation_host.hpp"
-#include "umd/device/types/arch.hpp"
 
 namespace tt::umd {
 
@@ -33,7 +32,7 @@ public:
      *
      * @param simulator_directory Path to the simulator binary/directory
      */
-    RtlSimCommunicator(const std::filesystem::path &simulator_directory, tt::ARCH arch);
+    explicit RtlSimCommunicator(const std::filesystem::path &simulator_directory);
 
     /**
      * Destructor that properly cleans up simulation host.
@@ -73,6 +72,28 @@ public:
      * @param size Number of bytes to write
      */
     void tile_write_bytes(uint32_t x, uint32_t y, uint64_t addr, const void *data, uint32_t size);
+
+    /**
+     * Read data from a tile core via SMN.
+     *
+     * @param x Core X coordinate
+     * @param y Core Y coordinate
+     * @param addr Address to read from
+     * @param data Buffer to store read data
+     * @param size Number of bytes to read
+     */
+    void smn_tile_read_bytes(uint32_t x, uint32_t y, uint64_t addr, void *data, uint32_t size);
+
+    /**
+     * Write data to a tile core via SMN.
+     *
+     * @param x Core X coordinate
+     * @param y Core Y coordinate
+     * @param addr Address to write to
+     * @param data Data to write
+     * @param size Number of bytes to write
+     */
+    void smn_tile_write_bytes(uint32_t x, uint32_t y, uint64_t addr, const void *data, uint32_t size);
 
     /**
      * Assert reset for all Tensix cores.
@@ -202,9 +223,6 @@ private:
 
     // Simulator directory path.
     std::filesystem::path simulator_directory_;
-
-    // Architecture of the device.
-    tt::ARCH arch_;
 
     // Simulation host for communication.
     SimulationHost host_;
--- a/device/api/umd/device/simulation/rtl_sim_communicator.hpp
+++ b/device/api/umd/device/simulation/rtl_sim_communicator.hpp
@@ -15,6 +15,7 @@
 #include <thread>
 
 #include "umd/device/simulation/simulation_host.hpp"
+#include "umd/device/types/arch.hpp"
 
 namespace tt::umd {
 
@@ -32,7 +33,7 @@ public:
      *
      * @param simulator_directory Path to the simulator binary/directory
      */
-    explicit RtlSimCommunicator(const std::filesystem::path &simulator_directory);
+    RtlSimCommunicator(const std::filesystem::path &simulator_directory, tt::ARCH arch);
 
     /**
      * Destructor that properly cleans up simulation host.
@@ -72,28 +73,6 @@ public:
      * @param size Number of bytes to write
      */
     void tile_write_bytes(uint32_t x, uint32_t y, uint64_t addr, const void *data, uint32_t size);
-
-    /**
-     * Read data from a tile core via SMN.
-     *
-     * @param x Core X coordinate
-     * @param y Core Y coordinate
-     * @param addr Address to read from
-     * @param data Buffer to store read data
-     * @param size Number of bytes to read
-     */
-    void smn_tile_read_bytes(uint32_t x, uint32_t y, uint64_t addr, void *data, uint32_t size);
-
-    /**
-     * Write data to a tile core via SMN.
-     *
-     * @param x Core X coordinate
-     * @param y Core Y coordinate
-     * @param addr Address to write to
-     * @param data Data to write
-     * @param size Number of bytes to write
-     */
-    void smn_tile_write_bytes(uint32_t x, uint32_t y, uint64_t addr, const void *data, uint32_t size);
 
     /**
      * Assert reset for all Tensix cores.
@@ -223,6 +202,9 @@ private:
 
     // Simulator directory path.
     std::filesystem::path simulator_directory_;
+
+    // Architecture of the device.
+    tt::ARCH arch_;
 
     // Simulation host for communication.
     SimulationHost host_;
--- a/device/api/umd/device/simulation/rtl_sim_communicator.hpp
+++ b/device/api/umd/device/simulation/rtl_sim_communicator.hpp
@@ -74,6 +74,28 @@ public:
     void tile_write_bytes(uint32_t x, uint32_t y, uint64_t addr, const void *data, uint32_t size);
 
     /**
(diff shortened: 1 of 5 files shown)


aleksamarkovicTT (approved):

device/api/umd/device/simulation/rtl_sim_communicator.hpp
```diff
@@ -31,7 +32,7 @@ class RtlSimCommunicator final {
      *
      * @param simulator_directory Path to the simulator binary/directory
      */
-    explicit RtlSimCommunicator(const std::filesystem::path &simulator_directory);
+    RtlSimCommunicator(const std::filesystem::path &simulator_directory, tt::ARCH arch);
```
nbuncicTT: nit: Can you leave the constructor `explicit`?

device/simulation/rtl_sim_communicator.cpp
```diff
@@ -148,12 +149,23 @@ void RtlSimCommunicator::shutdown() {
 }
 
 void RtlSimCommunicator::tile_read_bytes(uint32_t x, uint32_t y, uint64_t addr, void *data, uint32_t size) {
+    NocId noc_id = get_selected_noc_id();
+
+    if (noc_id == NocId::SYSTEM_NOC && arch_ != tt::ARCH::QUASAR) {
```
nbuncicTT: Can you make two small tests for these cases?

i.e. you create a test where we use grendel arch, and then set the NocId to be NOC1 and do a read? Or even easier, do it with bh arch and set the NocId to be SYSTEM_NOC and do a read? 

This will test this the behavior when using unexpected NOCs.

nbuncicTT (approved):

device/simulation/rtl_sim_communicator.cpp
```diff
@@ -178,6 +190,15 @@ void RtlSimCommunicator::tile_read_bytes(uint32_t x, uint32_t y, uint64_t addr,
 }
 
 void RtlSimCommunicator::tile_write_bytes(uint32_t x, uint32_t y, uint64_t addr, const void *data, uint32_t size) {
+    NocId noc_id = get_selected_noc_id();
+
+    if (noc_id == NocId::SYSTEM_NOC && arch_ != tt::ARCH::QUASAR) {
```
pjanevskiTT: nit: this is code is duplicated in read, maybe put it inside some small function to reuse it for these constraints

device/simulation/rtl_sim_communicator.cpp
```diff
@@ -186,7 +207,10 @@ void RtlSimCommunicator::tile_write_bytes(uint32_t x, uint32_t y, uint64_t addr,
     const auto *data_ptr = static_cast<const uint32_t *>(data);
     std::vector<uint32_t> data_vec(data_ptr, data_ptr + num_elements);
 
-    send_command_to_simulation_host(host_, create_flatbuffer(DEVICE_COMMAND_WRITE, data_vec, core, addr));
+    // Send write request.
+    // System NOC only available with aether-main-v2026.W10.1_smn_support tag and 2x3_SMU config.
+    DEVICE_COMMAND command = (noc_id == NocId::SYSTEM_NOC) ? DEVICE_COMMAND_SMN_WRITE : DEVICE_COMMAND_WRITE;
```
pjanevskiTT: can we maybe move this logic to TTDevice layer and add explicit SMN API to communicator. Communicator should have all the possible commands that you can send to remote part of simulator as part of the API. I think it shouldn't care about NOC which user selected, that should be logic inside TTDevice

So my proposal is

- Add explicit SMN API to communicator
- Handle deciding which API to call based on NOC selected on TTDevice layer in read/write functions
- This is going to remove arch from the constructor of communicator class

pjanevskiTT:
Taking over to merge this PR

device/simulation/rtl_sim_communicator.cpp
pjanevskiTT: will create a follow up issue, RTL sim tests are a bigger topic in UMD repo

device/api/umd/device/simulation/rtl_sim_communicator.hpp
pjanevskiTT: will do

device/simulation/rtl_sim_communicator.cpp
```diff
@@ -189,6 +189,36 @@ void RtlSimCommunicator::tile_write_bytes(uint32_t x, uint32_t y, uint64_t addr,
...
+    // Send SMN read request.
+    send_command_to_simulation_host(host_, create_flatbuffer(DEVICE_COMMAND_SMN_READ, {0}, core, addr, size));
+
+    // Get read response.
+    size_t rd_rsp_sz = host_.recv_from_device(&rd_resp);
+    auto rd_resp_buf = GetDeviceRequestResponse(rd_resp);
+
+    log_debug(tt::LogEmulationDriver, "Device SMN reading {} bytes from address {} in core ({}, {})", size, addr, x, y);
+
+    std::memcpy(data, rd_resp_buf->data()->data(), rd_resp_buf->data()->size() * sizeof(uint32_t));
```
pjanevskiTT: Two correctness bugs here vs. the existing `tile_read_bytes` pattern just above: (1) the `memcpy` length is `rd_resp_buf->data()->size() * sizeof(uint32_t)` (the response size) instead of the caller-requested `size`, so a response longer than requested will overflow the caller's `data` buffer, and a shorter response will leave the tail uninitialized with no error. (2) There is no assertion/error if the response is smaller than requested (`tile_read_bytes` has a `TT_ASSERT(response_bytes >= size, ...)`). Please mirror the `tile_read_bytes` logic: assert `response_bytes >= size`, then `std::memcpy(data, rd_resp_buf->data()->data(), size);`.

device/simulation/rtl_sim_communicator.cpp
```diff
@@ -189,6 +189,36 @@ void RtlSimCommunicator::tile_write_bytes(uint32_t x, uint32_t y, uint64_t addr,
...
+void RtlSimCommunicator::smn_tile_read_bytes(uint32_t x, uint32_t y, uint64_t addr, void *data, uint32_t size) {
+    std::lock_guard<std::mutex> lock(device_lock_);
+    tt_xy_pair core = {x, y};
+    void *rd_resp;
+
+    // Send SMN read request.
+    send_command_to_simulation_host(host_, create_flatbuffer(DEVICE_COMMAND_SMN_READ, {0}, core, addr, size));
+
+    // Get read response.
+    size_t rd_rsp_sz = host_.recv_from_device(&rd_resp);
```
pjanevskiTT: `smn_tile_read_bytes` uses `host_.recv_from_device(&rd_resp)` directly while the sibling `tile_read_bytes` uses `wait_for_command_response()` (the notification-thread/command-queue path). If the notification thread is consuming inbound messages, this direct recv will race with it and either deadlock or pull a frame out from under the notification thread. Please use the same `wait_for_command_response()` path unless there's a reason SMN must bypass it.

device/simulation/rtl_sim_communicator.cpp
```diff
@@ -189,6 +189,36 @@ void RtlSimCommunicator::tile_write_bytes(uint32_t x, uint32_t y, uint64_t addr,
...
+    std::memcpy(data, rd_resp_buf->data()->data(), rd_resp_buf->data()->size() * sizeof(uint32_t));
+    nng_free(rd_resp, rd_rsp_sz);
+}
+
+void RtlSimCommunicator::smn_tile_write_bytes(uint32_t x, uint32_t y, uint64_t addr, const void *data, uint32_t size) {
+    std::lock_guard<std::mutex> lock(device_lock_);
+    log_debug(tt::LogEmulationDriver, "Device SMN writing {} bytes to address {} in core ({}, {})", size, addr, x, y);
+
+    tt_xy_pair core = {x, y};
+    const uint32_t num_elements = size / sizeof(uint32_t);
```
pjanevskiTT, line 237: `size / sizeof(uint32_t)` silently drops the last 1-3 bytes when `size` is not a multiple of 4. `tile_write_bytes` has the same issue, but since this is a new code path consider either rounding up / asserting alignment or documenting the precondition in the header.

device/tt_device/rtl_simulation_tt_device.cpp
```diff
@@ -100,6 +113,15 @@ RtlSimulationTTDevice::~RtlSimulationTTDevice() { communicator_->shutdown(); }
 void RtlSimulationTTDevice::write_to_device(const void* mem_ptr, tt_xy_pair core, uint64_t addr, size_t size) {
     std::lock_guard<std::recursive_mutex> lock(device_lock);
     log_debug(tt::LogEmulationDriver, "Device writing {} bytes to l1_dest {} in core {}", size, addr, core.str());
+
+    NocId noc_id = get_selected_noc_id();
+    validate_noc_for_arch(noc_id, get_soc_descriptor().arch);
```
pjanevskiTT, line 127: `validate_noc_for_arch` is now called on every `read_from_device` / `write_to_device`. If `get_selected_noc_id()` ever defaults to NOC1 on a Quasar device (or to SYSTEM_NOC on a non-Quasar device that previously worked) every read/write will throw. Worth verifying the default selected NOC for each arch covered by `RtlSimulationTTDevice`, and adding a unit test that exercises both the SMN and non-SMN paths so the routing doesn't silently regress.

PR #2524, opened 2026-04-27
Merge queue jobs in workflows

broskoTT:
### Description
Adding jobs for merge queue.
We will run all important variants prior to merging to main (Release, Asan, Tsan).
Prior to merge queue, and on the main branch itself, Release run is enough.

### List of the changes
- Introduce merge queue jobs which runs Release Asan Tsan

### Testing
Well, no real way to test this other then trying it out live in production

### API Changes


nbuncicTT (approved):

.github/workflows/build-and-run-all-tests-multiple-builds.yml
```diff
@@ -40,26 +41,97 @@ jobs:
     with:
```
pjanevskiTT, line 41: Behavior change worth confirming: previously PRs ran ASan (the rationale in the deleted comment was "ASan adds a marginal perf hit but catches memory access bugs and runs all asserts that are optimized out in Release"); after this change PRs run Release only (ASan/TSan are present in the matrix but skipped by the `if:` guard above) and `merge_group` runs Release+ASan+TSan. That intentionally trades earlier ASan signal on each PR for ASan/TSan signal only at merge time. Also, push-to-main went from Release+ASan+TSan to Release only, so post-merge-on-main no longer catches the TSan-only race regressions the previous comment specifically called out. If that's intentional please reflect it in an inline comment; if not, push should keep ASan/TSan.

broskoTT: Refined the comment

Issue #2250, opened 2026-03-10
Separate base and high layer into different compilation units


broskoTT:
In order to guarantee separation.
Probably also different folders

broskoTT:
I'll track this in a milestone #2678

PR #2536, opened 2026-04-28
Cleaning up includes with IWYU

aleksamarkovicTT:
### Issue
Prerequisite for #2250.

### Description
Cleaning up includes with IWYU.
An automated IWYU workflow is probably possible, but there were many problems with our 3rd party dependencies, which could not be fixed with config or a mapping file, so I had to manually fix them.
Clang-tidy's `misc-include-cleaner` also struggles with 3rd party dependencies and some transitive includes, but is even more complicated to configure.

### List of the changes
- Ran IWYU over /device and /tests
- FIxed build errors after IWYU.
- Added `misc-header-include-cycle` to clang-tidy.

### Testing
CI

### API Changes

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -60,7 +60,6 @@ Checks: >
     -llvmlibc-*,
     -misc-const-correctness,
     -misc-definitions-in-headers,
-    -misc-header-include-cycle,
     -misc-include-cleaner,
     -misc-non-private-member-variables-in-classes,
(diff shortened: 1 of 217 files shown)


broskoTT (approved):
Glad to see this.
Had to manually do some of these changes when I tested out separating our packages in multiple modules.

If this is an automated tool, consider adding it to our pre-commit

Issue #2521, opened 2026-04-26
Fix Remote Multicast


broskoTT:
The multicast on remote chips is half-broken features. It doesn't work the same as for the local chips, because it doesn't skip all the cores required.
While at it, analyze ethernetbroadcast feature and integrate it properly. And refactor it out of the Cluster

broskoTT:
Merging PRs is going a bit slower, progressing

broskoTT:
Finally done

PR #2537, opened 2026-04-29
Multicast test and fix for remote

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/2521

### Description
There was a mismatch between local chip's multicast and remote chip's, in that it wasn't limited to TENSIX cores.

### List of the changes
- Enhance multicast tests greatly
- Fix the multicast/unicast fallback path in case of remote

### Testing
The test on this PR

### API Changes


device/tt_device/tt_device.cpp
```diff
@@ -450,11 +450,50 @@ tt_xy_pair TTDevice::get_arc_core() const { return arc_core; }
 
 void TTDevice::noc_multicast_write(void *src, size_t size, tt_xy_pair core_start, tt_xy_pair core_end, uint64_t addr) {
     if (is_remote_tt_device) {
-        // Remote devices don't have direct NOC multicast support.
-        // Fallback to unicast for all cores in the range.
-        for (std::size_t x = core_start.x; x <= core_end.x; ++x) {
-            for (std::size_t y = core_start.y; y <= core_end.y; ++y) {
-                write_to_device(src, xy_pair(x, y), addr, size);
+        // TODO: implement multicast over remote communication.
```
broskoTT: This should probably move now under WormholeTTDevice... Since the code is wormhole specific

device/tt_device/tt_device.cpp
```diff
@@ -458,11 +458,50 @@ tt_xy_pair TTDevice::get_arc_core() const { return arc_core; }
...
+        // For now, fall back to unicast, targeting only non-harvested TENSIX cores to match hardware
+        // multicast behavior.
+        //
+        // Coordinates may be in translated or NOC0 space; we check both since their ranges don't overlap.
+        // In translated space, non-harvested TENSIX rows occupy a prefix of the TENSIX translated rect.
+        // In NOC0 space, we look up the core in TENSIX_CORES_NOC0 and check its row against the mask.
+        const uint32_t tensix_harvesting_mask = get_chip_info().harvesting_masks.tensix_harvesting_mask;
+
+        uint32_t num_harvested = 0;
+        for (uint32_t mask = tensix_harvesting_mask; mask; mask >>= 1) {
```
pjanevskiTT: `is_active_tensix` hardcodes `wormhole::TENSIX_GRID_SIZE`, `wormhole::tensix_translated_coordinate_start_x/y`, and `wormhole::TENSIX_CORES_NOC0`. Only WH supports remote chips today, but this code lives in the generic `TTDevice` base class with no arch guard. If/when BH (or any other arch) gains remote support, this fallback will silently filter out all cores (or worse, mis-classify them) and the multicast will become a no-op without any failure. Either assert `get_arch() == tt::ARCH::WORMHOLE_B0` here, or push this fallback down to `WormholeTTDevice`.

device/tt_device/tt_device.cpp
```diff
@@ -458,11 +458,50 @@ tt_xy_pair TTDevice::get_arc_core() const { return arc_core; }
...
+            num_harvested += mask & 1;
+        }
+        const uint32_t num_active_rows = wormhole::TENSIX_GRID_SIZE.y - num_harvested;
+
+        auto is_active_tensix = [&](uint32_t x, uint32_t y) -> bool {
+            // Translated space: non-harvested TENSIX rows are the first num_active_rows rows of the rect.
+            if (x >= wormhole::tensix_translated_coordinate_start_x &&
+                x < wormhole::tensix_translated_coordinate_start_x + wormhole::TENSIX_GRID_SIZE.x &&
+                y >= wormhole::tensix_translated_coordinate_start_y &&
```
pjanevskiTT: The translated-space test `y < tensix_translated_coordinate_start_y + num_active_rows` relies on the fact that non-harvested rows are laid out as a contiguous prefix in translated y. That's true today for WH because `wormhole_coordinate_manager` assigns translated y starting at the prefix and skips harvested rows, but it's a non-obvious invariant. A core in the harvested-row "translated y" position would then be classified as non-active even though no such row exists in translated space. Worth a comment pointing at the invariant in `WormholeCoordinateManager::fill_tensix_noc0_translated_mapping` so future readers don't break it.

tests/api/test_device_io.cpp
```diff
@@ -349,6 +306,154 @@ TEST_F(TestDeviceIOFixture, TestDmaMulticastWrite) {
...
+                representative_cores.push_back(cores[0]);
+            }
+        }
+
+        // Keep one bystander TENSIX core on the same chip. This is used to verify when sending multicast to one TENSIX
+        // it doesn't spill to another.
+        CoreCoord bystander_tensix_core;
+        std::vector<uint32_t> bystander_original(num_words, 0);
+        // if (!full_grid) {
+        for (const auto& c : soc_desc.get_cores(CoreType::TENSIX, CoordSystem::TRANSLATED)) {
```
pjanevskiTT, line 368: `bystander_tensix_core` is default-initialized and only assigned inside the for-loop searching for a core not in `representative_cores`. If every TENSIX core happens to be in `representative_cores` (small grid / heavy harvesting), the subsequent `read_from_device(bystander_original.data(), bystander_tensix_core, ...)` reads from a default `CoreCoord{0,0,...}`, which may resolve to a non-existent or non-Tensix core and corrupt the test. Skip the bystander assertion when no bystander is found (or assert at least one is found before reading).

tests/test_utils/test_api_common.hpp
```diff
@@ -105,4 +105,22 @@ inline bool has_remote_chips() {
...
 
+// Returns the top-left (lowest x, lowest y) and bottom-right (highest x, highest y) TENSIX cores
+// in translated coordinates for the given SoC descriptor.
+inline std::vector<CoreCoord> get_tensix_corners(const SocDescriptor& soc_desc) {
+    const auto cores = soc_desc.get_cores(tt::CoreType::TENSIX, tt::CoordSystem::TRANSLATED);
+    EXPECT_FALSE(cores.empty()) << "No TENSIX cores found in SoC descriptor";
+    CoreCoord top_left = cores[0];
+    CoreCoord bottom_right = cores[0];
+    for (const auto& core : cores) {
+        if (core.x < top_left.x || core.y < top_left.y) {
```
pjanevskiTT, line 119: `if (core.x < top_left.x || core.y < top_left.y) top_left = core;` does not compute element-wise min: it picks any core with EITHER coordinate smaller, so `top_left` can end up with a smaller x but a larger y than the true top-left (and analogously for `bottom_right` on the next branch). This corrupts the "tensix corners" returned to the multicast tests, which are then used as the multicast rectangle. Compute min/max for x and y independently and combine them, or use `std::min_element` twice.

device/tt_device/tt_device.cpp
```diff
@@ -458,11 +458,50 @@ tt_xy_pair TTDevice::get_arc_core() const { return arc_core; }
 void TTDevice::noc_multicast_write(void *src, size_t size, tt_xy_pair core_start, tt_xy_pair core_end, uint64_t addr) {
     ZoneScopedC(tracy::Color::Orange);
     if (is_remote_tt_device) {
-        // Remote devices don't have direct NOC multicast support.
-        // Fallback to unicast for all cores in the range.
-        for (std::size_t x = core_start.x; x <= core_end.x; ++x) {
-            for (std::size_t y = core_start.y; y <= core_end.y; ++y) {
-                write_to_device(src, xy_pair(x, y), addr, size);
+        // TODO: implement multicast over remote communication.
```
pjanevskiTT: code looks a bit fuzzy imo. Is this basically to decide whether we are using translated or NOC0 and calculate those coordinates? If so, will moving SocDescriptor to TTDevice make this code less fuzzy? 
Anyway, I would maybe move this code to separate function and think about whether it can be more simple

pjanevskiTT (approved):

device/tt_device/tt_device.cpp
broskoTT: No, it's actually trying to figure out which tensix cores to write to.
I'll try to rewrite it using soc_descriptor, I could loop over all tensix from soc_descriptor and filter out the ones included in the range, instead of the other way around.

tests/api/test_device_io.cpp
```diff
@@ -349,6 +306,154 @@ TEST_F(TestDeviceIOFixture, TestDmaMulticastWrite) {
...
+
+// Parametrized over (use_noc0, full_grid, sysmem_enabled):
+//   use_noc0       - true: NOC0 coordinates; false: translated coordinates
+//   full_grid      - true: multicast spans the entire chip grid; false: isolated single-core multicast
+//   sysmem_enabled - true: 1 host mem channel (needed for remote chips); false: 0 (fallback path)
+// For full_grid+NOC0 the range is {0,0}–{grid_size-1}; for full_grid+translated the tensix corners are used.
+// Bystander verification is only performed for isolated-core multicasts.
+TEST_P(TestMulticastWriteFixture, TestMulticastWrite) {
+    // TODO: full_grid parameter to be added in following PR.
+    // TODO: sysmem_enabled parameter to be added in the following PR.
```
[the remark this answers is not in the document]
broskoTT: This will be uncommented in immediate next PRs

tests/test_utils/test_api_common.hpp
broskoTT: this can't ever happen for a rectangle

tests/test_utils/test_api_common.hpp
```diff
@@ -105,4 +105,22 @@ inline bool has_remote_chips() {
...
+            top_left = core;
+        }
+        if (core.x > bottom_right.x || core.y > bottom_right.y) {
+            bottom_right = core;
+        }
```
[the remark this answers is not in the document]
broskoTT: no

tests/api/test_device_io.cpp
```diff
@@ -349,6 +306,154 @@ TEST_F(TestDeviceIOFixture, TestDmaMulticastWrite) {
...
+    auto [use_noc0] = GetParam();
+
+    std::unique_ptr<Cluster> cluster = std::make_unique<Cluster>(ClusterOptions{.num_host_mem_ch_per_mmio_device = 0});
```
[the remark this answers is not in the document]
broskoTT: I actually want  to control this explicitly in the following commits

device/tt_device/tt_device.cpp
broskoTT: Rewrote, should be much much clearer now

device/tt_device/tt_device.cpp
broskoTT: not applicable anymore

device/tt_device/tt_device.cpp
broskoTT: not applicable anymore

tests/api/test_device_io.cpp
broskoTT: Added an assert

tests/test_utils/test_api_common.hpp
```diff
@@ -105,4 +105,22 @@ inline bool has_remote_chips() {
 
 inline uint32_t get_num_host_ch_for_test() { return has_remote_chips() ? 1UL : 0UL; }
 
+// Returns the top-left (lowest x, lowest y) and bottom-right (highest x, highest y) TENSIX cores
+// in translated coordinates for the given SoC descriptor.
+inline std::vector<CoreCoord> get_tensix_corners(const SocDescriptor& soc_desc) {
+    const auto cores = soc_desc.get_cores(tt::CoreType::TENSIX, tt::CoordSystem::TRANSLATED);
+    EXPECT_FALSE(cores.empty()) << "No TENSIX cores found in SoC descriptor";
```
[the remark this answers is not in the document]
broskoTT: changed to assert

Issue #2317, opened 2026-03-18
Refactor TLBManager: separate TLB allocation/bookkeeping from Handle+Window factory logic


pjanevskiTT:
### Context

Follow-up from PR #2306 ([comment](https://github.com/tenstorrent/tt-umd/pull/2306#discussion_r2949929925)) by @nbuncicTT.

### Problem

`TLBManager` currently mixes two independent responsibilities:

1. **TLB allocation/bookkeeping** — which indices are free, what sizes exist, what addresses they map to. This is backend-agnostic and only needs architecture config (TLB counts, sizes, base addresses).
2. **TlbWindow factory** — creating `TlbHandle` + `TlbWindow` objects. This requires knowledge of communicators, handle types, window types, and concrete backend details.

Combining both in one class forces backend-specific construction logic into a class that should be pure bookkeeping. This leads to `std::function` factories, lambda captures, and other glue to inject backend knowledge where it doesn't belong.

### Proposed Solution

Split into two concerns:

- **TLB allocator**: knows architecture, manages indices, computes addresses. No knowledge of handles or windows. Pure bookkeeping.
- **Each device**: uses the allocator to get an index + size, then constructs its own `Handle` + `Window` with the types it already knows about.

### References

- `device/api/umd/device/chip_helpers/simulation_tlb_manager.hpp`
- `device/chip_helpers/tlb_manager.cpp`

pjanevskiTT:
closed by #2544

PR #2538, opened 2026-04-29
Extract SimulationTlbAllocator out of SimulationTlbManager

pjanevskiTT:
### Issue

Refs #2317. First PR in a stack of five for #2317. Followed by #2540.

### Description

Pure refactor — no behavior change. First step toward the `TLBManager` allocation/factory split tracked in #2317.

Today `SimulationTlbManager` mixes two unrelated responsibilities: TLB index bookkeeping (size-class bitmaps, free/used tracking, BAR0-relative address math) and `TlbHandle`/`TlbWindow` construction (the `TlbWindowFactory` callable used by simulation backends).

This PR carves the bookkeeping half out into a new `SimulationTlbAllocator` class. `SimulationTlbManager` holds one as a member and forwards the index/address methods to it. The factory-half stays put for now; it moves out in later PRs in the stack.

### List of the changes

- New class `SimulationTlbAllocator` (`device/api/umd/device/chip_helpers/simulation_tlb_allocator.hpp`, `device/chip_helpers/simulation_tlb_allocator.cpp`) owning: per-size-class allocation bitmaps, allocation mutex, per-arch sizes/counts/start indices, `bar0_base_`, `arch_impl_`, and the methods `allocate_tlb_index`, `deallocate_tlb_index`, `get_tlb_size_from_index`, `get_tlb_address_from_index`, `get_tlb_reg_address_from_index`, `get_architecture_impl`.
- `SimulationTlbManager` now holds a `SimulationTlbAllocator allocator_` member. The same index/address methods stay on `SimulationTlbManager` as thin forwarders so simulation `TlbHandle` subclasses (`TTSimTlbHandle`, `RtlSimTlbHandle`) that hold a `SimulationTlbManager*` back-pointer keep working unchanged. Handle back-pointer migration is the next PR in the stack (#2540).
- `allocate_tlb_window` and `allocate_default_tlb_window` keep their semantics: they consult the allocator for index/size and pass `this` to the existing `TlbWindowFactory`.
- `device/CMakeLists.txt`: registered the new header in the public API file set and the new cpp in the simulation sources block.

### Testing

CI.

### API Changes

### DIFF
--- a/device/api/umd/device/chip_helpers/simulation_tlb_allocator.hpp
+++ b/device/api/umd/device/chip_helpers/simulation_tlb_allocator.hpp
@@ -1,0 +1,100 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+#include <mutex>
+#include <vector>
+
+#include "umd/device/types/arch.hpp"
+
+namespace tt::umd {
+
+class architecture_implementation;
+
+/**
+ * In-process allocator for simulation TLB indices.
+ *
+ * Tracks which TLB indices are allocated per size class, and computes BAR0-relative
+ * addresses for a given index. Counterpart to KMD-managed allocation on silicon —
+ * no knowledge of TlbHandle / TlbWindow types.
+ */
+class SimulationTlbAllocator {
+public:
+    SimulationTlbAllocator(uint64_t bar0_base, const architecture_implementation* arch_impl);
+
+    /**
+     * Allocate a TLB index that fits the requested size. If size is 0, allocate
+     * any available TLB, preferring smaller sizes first.
+     *
+     * @param size Requested TLB size in bytes (0 means any available).
+     * @return TLB index if successful, -1 if no TLB available.
+     */
+    int allocate_tlb_index(size_t size);
+
+    /**
+     * Mark a TLB index as free.
+     */
+    void deallocate_tlb_index(int tlb_index);
+
+    /**
+     * Size of the TLB at the given index, in bytes.
+     */
+    size_t get_tlb_size_from_index(int tlb_index);
+
+    /**
+     * BAR0-relative address mapped by the TLB at the given index.
+     */
+    uint64_t get_tlb_address_from_index(int tlb_index);
+
+    /**
+     * Address of the TLB configuration register for the given index.
+     */
+    uint64_t get_tlb_reg_address_from_index(int tlb_index);
+
+    const architecture_implementation* get_architecture_impl() const;
+
+private:
+    void initialize_architecture_config();
+
+    // Allocate a TLB index assuming allocation_mutex_ is already held by the caller.
+    // Used for the size==0 path which recurses internally.
+    int allocate_tlb_index_locked(size_t size);
+
+    uint64_t bar0_base_ = 0;
+    const architecture_implementation* arch_impl_ = nullptr;
+
+    // Architecture-specific TLB configuration.
+    tt::ARCH architecture_;
+    size_t tlb_reg_size_bytes_ = 8;  // Default to Wormhole size.
+
+    // TLB size constants (set based on architecture).
+    size_t tlb_1mb_size_ = 0;
+    size_t tlb_2mb_size_ = 0;
+    size_t tlb_16mb_size_ = 0;
+    size_t tlb_4gb_size_ = 0;
+
+    // TLB count constants (set based on architecture).
+    size_t tlb_1mb_count_ = 0;
+    size_t tlb_2mb_count_ = 0;
+    size_t tlb_16mb_count_ = 0;
+    size_t tlb_4gb_count_ = 0;
+
+    // TLB index ranges (set based on architecture).
+    size_t tlb_1mb_start_index_ = 0;
+    size_t tlb_2mb_start_index_ = 0;
+    size_t tlb_16mb_start_index_ = 0;
+    size_t tlb_4gb_start_index_ = 0;
+
+    // TLB allocation tracking (dynamically sized based on architecture).
+    std::mutex allocation_mutex_;
+    std::vector<bool> tlb_1mb_allocated_;   // Wormhole: 156 TLBs, Blackhole: 0.
+    std::vector<bool> tlb_2mb_allocated_;   // Wormhole: 10 TLBs, Blackhole: 202.
+    std::vector<bool> tlb_16mb_allocated_;  // Wormhole: 20 TLBs, Blackhole: 0.
+    std::vector<bool> tlb_4gb_allocated_;   // Wormhole: 0 TLBs, Blackhole: 8.
+};
+
+}  // namespace tt::umd
--- a/device/api/umd/device/chip_helpers/simulation_tlb_allocator.hpp
+++ b/device/api/umd/device/chip_helpers/simulation_tlb_allocator.hpp
@@ -76,10 +76,6 @@ private:
 
     void initialize_architecture_config();
 
-    // Allocate a TLB index assuming allocation_mutex_ is already held by the caller.
-    // Used for the size==0 path which recurses internally.
-    int allocate_tlb_index_internal(size_t size);
-
     // Returns the pool that owns `tlb_index`, or nullptr if no pool covers it
     // (including for negative indices).
     TlbSizeClass* find_size_class_for_index(int tlb_index);
--- a/device/api/umd/device/chip_helpers/simulation_tlb_allocator.hpp
+++ b/device/api/umd/device/chip_helpers/simulation_tlb_allocator.hpp
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <array>
 #include <cstddef>
 #include <cstdint>
 #include <mutex>
@@ -61,43 +62,35 @@ public:
     const architecture_implementation* get_architecture_impl() const;
 
 private:
+    // A pool of TLB indices that all share a single size.
+    struct TlbSizeClass {
+        size_t size = 0;         // Bytes per TLB; 0 means unused on the current arch.
+        size_t count = 0;        // Number of TLBs in the pool.
+        size_t start_index = 0;  // First global TLB index belonging to this pool.
+        std::vector<bool> allocated;
+    };
+
+    // Indices into size_classes_; ordered smallest -> largest so allocate-with-escalation
+    // and address-layout iteration both Just Work.
+    enum SizeClass : size_t { ONE_MB = 0, TWO_MB = 1, SIXTEEN_MB = 2, FOUR_GB = 3, NUM_SIZE_CLASSES = 4 };
+
     void initialize_architecture_config();
 
     // Allocate a TLB index assuming allocation_mutex_ is already held by the caller.
     // Used for the size==0 path which recurses internally.
     int allocate_tlb_index_internal(size_t size);
 
+    // Returns the pool that owns `tlb_index`, or nullptr if no pool covers it
+    // (including for negative indices).
+    TlbSizeClass* find_size_class_for_index(int tlb_index);
+
     uint64_t bar0_base_ = 0;
     const architecture_implementation* arch_impl_ = nullptr;
-
-    // Architecture-specific TLB configuration.
     tt::ARCH architecture_;
     size_t tlb_reg_size_bytes_ = 8;  // Default to Wormhole size.
 
-    // TLB size constants (set based on architecture).
-    size_t tlb_1mb_size_ = 0;
-    size_t tlb_2mb_size_ = 0;
-    size_t tlb_16mb_size_ = 0;
-    size_t tlb_4gb_size_ = 0;
-
-    // TLB count constants (set based on architecture).
-    size_t tlb_1mb_count_ = 0;
-    size_t tlb_2mb_count_ = 0;
-    size_t tlb_16mb_count_ = 0;
-    size_t tlb_4gb_count_ = 0;
-
-    // TLB index ranges (set based on architecture).
-    size_t tlb_1mb_start_index_ = 0;
-    size_t tlb_2mb_start_index_ = 0;
-    size_t tlb_16mb_start_index_ = 0;
-    size_t tlb_4gb_start_index_ = 0;
-
-    // TLB allocation tracking (dynamically sized based on architecture).
     std::mutex allocation_mutex_;
-    std::vector<bool> tlb_1mb_allocated_;   // Wormhole: 156 TLBs, Blackhole: 0.
-    std::vector<bool> tlb_2mb_allocated_;   // Wormhole: 10 TLBs, Blackhole: 202.
-    std::vector<bool> tlb_16mb_allocated_;  // Wormhole: 20 TLBs, Blackhole: 0.
-    std::vector<bool> tlb_4gb_allocated_;   // Wormhole: 0 TLBs, Blackhole: 8.
+    std::array<TlbSizeClass, NUM_SIZE_CLASSES> size_classes_;
 };
 
 }  // namespace tt::umd
--- a/device/api/umd/device/chip_helpers/simulation_tlb_allocator.hpp
+++ b/device/api/umd/device/chip_helpers/simulation_tlb_allocator.hpp
@@ -27,8 +27,11 @@ public:
     SimulationTlbAllocator(uint64_t bar0_base, const architecture_implementation* arch_impl);
 
     /**
-     * Allocate a TLB index that fits the requested size. If size is 0, allocate
-     * any available TLB, preferring smaller sizes first.
+     * Allocate the smallest TLB whose size class is >= the requested size. If no
+     * TLB is free in that size class, escalate to the next larger size class and
+     * retry, continuing until a TLB is allocated or all size classes are exhausted.
+     *
+     * If size is 0, allocate any available TLB, preferring smaller size classes first.
      *
      * @param size Requested TLB size in bytes (0 means any available).
      * @return TLB index if successful, -1 if no TLB available.
@@ -62,7 +65,7 @@ private:
 
     // Allocate a TLB index assuming allocation_mutex_ is already held by the caller.
     // Used for the size==0 path which recurses internally.
-    int allocate_tlb_index_locked(size_t size);
+    int allocate_tlb_index_internal(size_t size);
 
     uint64_t bar0_base_ = 0;
     const architecture_implementation* arch_impl_ = nullptr;
--- a/device/api/umd/device/chip_helpers/simulation_tlb_manager.hpp
+++ b/device/api/umd/device/chip_helpers/simulation_tlb_manager.hpp
@@ -8,10 +8,8 @@
 #include <cstdint>
 #include <functional>
 #include <memory>
-#include <mutex>
-#include <vector>
 
-#include "umd/device/arch/architecture_implementation.hpp"
+#include "umd/device/chip_helpers/simulation_tlb_allocator.hpp"
 #include "umd/device/chip_helpers/tlb_manager.hpp"
 #include "umd/device/pcie/tlb_window.hpp"
 #include "umd/device/types/tlb.hpp"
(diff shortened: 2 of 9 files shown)


device/chip_helpers/simulation_tlb_allocator.cpp
```diff
@@ -0,0 +1,275 @@
...
+            return bar0_base_ + (static_cast<uint64_t>(tlb_index) * tlb_2mb_size_);
+        } else if (tlb_4gb_count_ > 0 && tlb_index >= tlb_4gb_start_index_) {
+            // 4GB TLBs are in BAR4, not BAR0 for Blackhole.
+            UMD_THROW(
+                error::RuntimeError, "4GB TLBs in Blackhole are not currently supported in simulation TLB allocation.");
+        }
+    }
+
+    // Invalid TLB index.
+    return 0;
```
pjanevskiTT: Returning `0` for an invalid index is silent and ambiguous — a caller can't distinguish "this is the BAR0-relative offset 0" from "you passed a bogus index". Pre-existing behavior carried over from `SimulationTlbManager`, but worth tightening (e.g. `UMD_THROW(error::RuntimeError, ...)`, or returning `std::optional<uint64_t>`) since this is the right moment to fix it while moving the code.

device/chip_helpers/simulation_tlb_allocator.cpp
```diff
@@ -0,0 +1,275 @@
...
+        tlb_16mb_start_index_ = 0;  // Not applicable.
+        tlb_4gb_start_index_ = tlb_2mb_count_;
+
+        // Initialize allocation tracking.
+        // tlb_1mb_allocated_ and tlb_16mb_allocated_ remain empty.
+        tlb_2mb_allocated_.resize(tlb_2mb_count_, false);
+        tlb_4gb_allocated_.resize(tlb_4gb_count_, false);
+
+    } else {
+        log_debug(
```
pjanevskiTT, line 148: Unsupported architectures silently produce a useless allocator: counts/sizes stay zero, every `allocate_tlb_index` returns `-1`, and the failure surfaces only at the first allocation attempt as a generic "No available TLB". Carried over from the old code, but consider either throwing in the constructor for arches without a config, or at least leaving an explicit `architecture_ = ...` and a comment so the empty-pool state is intentional rather than accidental.

device/chip_helpers/simulation_tlb_allocator.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+    return -1;  // No available TLB.
+}
+
+void SimulationTlbAllocator::deallocate_tlb_index(int tlb_index) {
+    ZoneScopedC(tracy::Color::Cyan);
+    std::lock_guard<std::mutex> lock(allocation_mutex_);
+
+    // Check 1MB TLBs (Wormhole only).
+    if (tlb_1mb_count_ > 0 && tlb_index >= tlb_1mb_start_index_ && tlb_index < tlb_2mb_start_index_) {
```
pjanevskiTT: Mixing signed `int tlb_index` with `size_t` `tlb_*_start_index_` is a footgun — a negative `tlb_index` (e.g. `-1` from a failed allocate) silently casts to `SIZE_MAX` and ends up in `get_tlb_address_from_index`'s "Invalid TLB index" path returning `0` rather than a clear failure. Either change `tlb_index` to `size_t` everywhere with an early `if (tlb_index < 0) return;` guard at each entry point, or add the guard inline. Same pattern in `get_tlb_size_from_index` and `deallocate_tlb_index`.

device/chip_helpers/simulation_tlb_allocator.cpp
```diff
@@ -0,0 +1,275 @@
...
+        // Initialize allocation tracking.
+        // tlb_1mb_allocated_ and tlb_16mb_allocated_ remain empty.
+        tlb_2mb_allocated_.resize(tlb_2mb_count_, false);
+        tlb_4gb_allocated_.resize(tlb_4gb_count_, false);
+
+    } else {
+        log_debug(
+            LogUMD,
+            fmt::format(
+                "Architecture {} does not yet have support for TLB management in simulation. UMD will use legacy "
```
pjanevskiTT, line 151: The `size==0` path now correctly recurses into `_locked` to avoid re-entering `allocation_mutex_`, which is a real fix vs. the old `SimulationTlbManager::allocate_tlb_index` (it took the non-recursive `allocation_mutex_` and then recursively called the also-locking public method, which would self-deadlock if anything ever exercised the size==0 fall-through). Worth a regression test that actually exercises the cross-class fall-through under a held lock — `WormholeAllocateZeroPicksSmallest` only hits the first branch and so wouldn't have caught the original deadlock either. e.g., drain the 1MB pool, then call `allocate_tlb_index(0)` and expect a 2MB index back.

device/api/umd/device/chip_helpers/simulation_tlb_manager.hpp
```diff
@@ -51,81 +49,22 @@ class SimulationTlbManager : public TLBManager {
...
 
 private:
-    /**
-     * Initialize architecture-specific TLB configuration based on the device architecture.
-     */
-    void initialize_architecture_config();
-
-    uint64_t bar0_base_ = 0;
-    const architecture_implementation* arch_impl_ = nullptr;
+    SimulationTlbAllocator allocator_;
```
broskoTT, line 66: I still think these two classes should not be entangled.
The comment wasn't that there's to many stuff in one class, rather that TLBManager should not hold primitives that SimulationTlbAllocator holds

broskoTT: Now that I properly read the description, seems this is removed in further PRs. Ignore this comment

device/api/umd/device/chip_helpers/simulation_tlb_allocator.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * Mark a TLB index as free.
+     */
+    void deallocate_tlb_index(int tlb_index);
+
+    /**
+     * Size of the TLB at the given index, in bytes.
+     */
+    size_t get_tlb_size_from_index(int tlb_index);
```
broskoTT, line 50: What is the point of this API? If the point is to get what is the size allocated in case you called allocate_tlb_index with a size==0, maybe that function should return both index and size instead?

broskoTT: Actually looking at the next two functions as well:
Isn't it cleaner if SimulationTlbAllocator just returns TTSimTLBHandle?

device/chip_helpers/simulation_tlb_allocator.cpp
```diff
@@ -0,0 +1,275 @@
...
+const architecture_implementation* SimulationTlbAllocator::get_architecture_impl() const { return arch_impl_; }
+
+void SimulationTlbAllocator::initialize_architecture_config() {
+    architecture_ = arch_impl_->get_architecture();
+
+    if (architecture_ == tt::ARCH::WORMHOLE_B0) {
+        // Wormhole B0 configuration.
+        tlb_reg_size_bytes_ = 8;  // wormhole::TLB_CFG_REG_SIZE_BYTES.
+
+        const auto& tlb_sizes = arch_impl_->get_tlb_sizes();
```
broskoTT, line 122: This seems like a mix of two approaches:
- Check if the architecture is WH or BH and set hardcoded constants
- Call generic functions on architecture_implementation class

Can we avoid two branches for WH and BH?

device/api/umd/device/chip_helpers/simulation_tlb_allocator.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+    // Architecture-specific TLB configuration.
+    tt::ARCH architecture_;
+    size_t tlb_reg_size_bytes_ = 8;  // Default to Wormhole size.
+
+    // TLB size constants (set based on architecture).
+    size_t tlb_1mb_size_ = 0;
+    size_t tlb_2mb_size_ = 0;
+    size_t tlb_16mb_size_ = 0;
+    size_t tlb_4gb_size_ = 0;
```
broskoTT: how come these are configurable? This have const values

device/api/umd/device/chip_helpers/simulation_tlb_allocator.hpp
```diff
@@ -0,0 +1,100 @@
...
+    uint64_t get_tlb_reg_address_from_index(int tlb_index);
+
+    const architecture_implementation* get_architecture_impl() const;
+
+private:
+    void initialize_architecture_config();
+
+    // Allocate a TLB index assuming allocation_mutex_ is already held by the caller.
+    // Used for the size==0 path which recurses internally.
+    int allocate_tlb_index_locked(size_t size);
```
broskoTT: consider renaming to "_internal" instead of "_locked"

device/api/umd/device/chip_helpers/simulation_tlb_allocator.hpp
```diff
@@ -0,0 +1,100 @@
...
+    SimulationTlbAllocator(uint64_t bar0_base, const architecture_implementation* arch_impl);
+
+    /**
+     * Allocate a TLB index that fits the requested size. If size is 0, allocate
+     * any available TLB, preferring smaller sizes first.
+     *
+     * @param size Requested TLB size in bytes (0 means any available).
+     * @return TLB index if successful, -1 if no TLB available.
+     */
+    int allocate_tlb_index(size_t size);
```
broskoTT, line 40: Can you be clearer in the comment that "Allocate a TLB index that fits the requested size" means it will try to allocate the smallest tlb which is not smaller than the size requested, and increase the size further until it allocates something or all tlbs are taken? It wasn't clear on the first read that it will try allocating multiple sizes

device/chip_helpers/simulation_tlb_allocator.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+int SimulationTlbAllocator::allocate_tlb_index(size_t size) {
+    ZoneScopedC(tracy::Color::Cyan);
+    std::lock_guard<std::mutex> lock(allocation_mutex_);
+    return allocate_tlb_index_locked(size);
+}
+
+int SimulationTlbAllocator::allocate_tlb_index_locked(size_t size) {
+    if (size == 0) {
+        // Allocate any available TLB, prefer smaller sizes first.
```
broskoTT: What's the point of this branch? Why doesn't the code below work if size is 0?

device/chip_helpers/simulation_tlb_allocator.cpp
```diff
@@ -0,0 +1,275 @@
...
+        for (size_t i = 0; i < tlb_1mb_count_; ++i) {
+            if (!tlb_1mb_allocated_[i]) {
+                tlb_1mb_allocated_[i] = true;
+                return static_cast<int>(tlb_1mb_start_index_ + i);
+            }
+        }
+    }
+
+    // Try 2MB TLBs (both architectures).
+    if (tlb_2mb_size_ > 0 && size <= tlb_2mb_size_) {
```
broskoTT: Could this can be rewritten to be more readable with a for loop? You have the same logic, just different containers

device/chip_helpers/simulation_tlb_allocator.cpp
```diff
@@ -0,0 +1,275 @@
...
+        size_t local_index = tlb_index - tlb_1mb_start_index_;
+        if (local_index < tlb_1mb_count_) {
+            tlb_1mb_allocated_[local_index] = false;
+        }
+    }
+    // Check 2MB TLBs (both architectures).
+    else if (
+        tlb_2mb_count_ > 0 && tlb_index >= tlb_2mb_start_index_ &&
+        (tlb_16mb_count_ == 0 || tlb_index < tlb_16mb_start_index_) &&
+        (tlb_4gb_count_ == 0 || tlb_index < tlb_4gb_start_index_)) {
```
broskoTT: this is super unreadable.
Can you have a small dedicated internal function which just gives you which tlb type is in question based on index?

device/chip_helpers/simulation_tlb_allocator.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+uint64_t SimulationTlbAllocator::get_tlb_address_from_index(int tlb_index) {
+    if (architecture_ == tt::ARCH::WORMHOLE_B0) {
+        // Wormhole B0 address calculation.
+        if (tlb_1mb_count_ > 0 && tlb_index >= tlb_1mb_start_index_ && tlb_index < tlb_2mb_start_index_) {
+            // 1MB TLBs: indices 0-155.
+            return bar0_base_ + (static_cast<uint64_t>(tlb_index) * tlb_1mb_size_);
+        } else if (tlb_2mb_count_ > 0 && tlb_index >= tlb_2mb_start_index_ && tlb_index < tlb_16mb_start_index_) {
+            // 2MB TLBs: indices 156-165.
+            uint64_t offset_1mb_region = tlb_1mb_count_ * tlb_1mb_size_;
```
broskoTT: aren't these already defined somewhere in wormhole_implementation.hpp?

tests/baremetal/test_simulation_tlb_allocator.cpp
```diff
@@ -0,0 +1,196 @@
...
+        allocator.get_tlb_reg_address_from_index(5),
+        TEST_BAR0_BASE + TLB_CONFIG_REG_BASE_OFFSET + 5 * BH_TLB_REG_STRIDE);
+}
+
+TEST(SimulationTlbAllocator, GetArchitectureImpl) {
+    auto arch_impl = architecture_implementation::create(tt::ARCH::WORMHOLE_B0);
+    const architecture_implementation* raw = arch_impl.get();
+    SimulationTlbAllocator allocator(TEST_BAR0_BASE, raw);
+
+    EXPECT_EQ(allocator.get_architecture_impl(), raw);
```
broskoTT: Is this api needed?

tests/baremetal/test_simulation_tlb_allocator.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+TEST(SimulationTlbAllocator, BlackholeRegAddressFromIndex) {
+    auto arch_impl = architecture_implementation::create(tt::ARCH::BLACKHOLE);
+    SimulationTlbAllocator allocator(TEST_BAR0_BASE, arch_impl.get());
+
+    // Blackhole uses a 12-byte register stride.
+    EXPECT_EQ(allocator.get_tlb_reg_address_from_index(0), TEST_BAR0_BASE + TLB_CONFIG_REG_BASE_OFFSET);
+    EXPECT_EQ(
+        allocator.get_tlb_reg_address_from_index(5),
+        TEST_BAR0_BASE + TLB_CONFIG_REG_BASE_OFFSET + 5 * BH_TLB_REG_STRIDE);
```
broskoTT: You're programming tests.
You're not testing the API, you're locking in the implementation.
Write a raw number here

tests/baremetal/test_simulation_tlb_allocator.cpp
```diff
@@ -0,0 +1,196 @@
...
+}
+
+TEST(SimulationTlbAllocator, BlackholeAddressFromIndex) {
+    auto arch_impl = architecture_implementation::create(tt::ARCH::BLACKHOLE);
+    SimulationTlbAllocator allocator(TEST_BAR0_BASE, arch_impl.get());
+
+    EXPECT_EQ(allocator.get_tlb_address_from_index(0), TEST_BAR0_BASE);
+    EXPECT_EQ(
+        allocator.get_tlb_address_from_index(BH_NUM_2MB_TLBS - 1),
+        TEST_BAR0_BASE + (BH_NUM_2MB_TLBS - 1) * BH_TLB_2MB_SIZE);
```
broskoTT: Same as below, write a raw number

tests/baremetal/test_simulation_tlb_allocator.cpp
```diff
@@ -0,0 +1,196 @@
...
+    EXPECT_EQ(allocator.get_tlb_address_from_index(0), TEST_BAR0_BASE);
+    // Last 1MB TLB.
+    EXPECT_EQ(
+        allocator.get_tlb_address_from_index(WH_NUM_1MB_TLBS - 1),
+        TEST_BAR0_BASE + (WH_NUM_1MB_TLBS - 1) * WH_TLB_1MB_SIZE);
+    // First 2MB TLB sits right after the 1MB region.
+    const uint64_t end_of_1mb_region = TEST_BAR0_BASE + WH_NUM_1MB_TLBS * WH_TLB_1MB_SIZE;
+    EXPECT_EQ(allocator.get_tlb_address_from_index(WH_NUM_1MB_TLBS), end_of_1mb_region);
+    // First 16MB TLB sits right after the 2MB region.
+    const uint64_t end_of_2mb_region = end_of_1mb_region + WH_NUM_2MB_TLBS * WH_TLB_2MB_SIZE;
```
broskoTT: Same as below, use raw numbers in tests

broskoTT (approved):

device/api/umd/device/chip_helpers/simulation_tlb_allocator.hpp
pjanevskiTT: it should also be used for RTL. That's why we have the factory. Allocator is designed to be similar to KMD approach which just returns index of available TLB so you can do whatever you want with it. So I would like to keep this API of getting the index and then we do whatever we want with it. There will be get_io_window API, and then you don't even care about TLB handles.
I can think about merging size and offset API

tests/baremetal/test_simulation_tlb_allocator.cpp
pjanevskiTT: fixed all constants in newly added tests

tests/baremetal/test_simulation_tlb_allocator.cpp
pjanevskiTT: used by tlb manager for now, will remove it at one point

device/chip_helpers/simulation_tlb_allocator.cpp
pjanevskiTT: I think we can address this as follow up. Basically allocator takes architecture, which defines some TLB layout and from there on it works as generic. We could derive multiple classes that just have separate init based on architecture, but the point of this for now is to hide tlb allocation and decouple it, so I would address it as follow up

device/api/umd/device/chip_helpers/simulation_tlb_allocator.hpp
pjanevskiTT: look at comment above

device/chip_helpers/simulation_tlb_allocator.cpp
pjanevskiTT: yes but they should not be hardcoded here, they are arch dependent

device/chip_helpers/simulation_tlb_allocator.cpp
pjanevskiTT: rewritten

device/chip_helpers/simulation_tlb_allocator.cpp
pjanevskiTT: rewritten, should be more clear

device/api/umd/device/chip_helpers/simulation_tlb_allocator.hpp
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/2586

device/chip_helpers/simulation_tlb_allocator.cpp
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/2586

device/chip_helpers/simulation_tlb_allocator.cpp
pjanevskiTT: it just loops over different sizes, code below is looking for exact size

device/chip_helpers/simulation_tlb_allocator.cpp
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/2586

device/chip_helpers/simulation_tlb_allocator.cpp
pjanevskiTT: changes that logic, comments doesn't make much sense now

device/chip_helpers/simulation_tlb_allocator.cpp
pjanevskiTT: rewritten it to be in the same branch

device/chip_helpers/simulation_tlb_allocator.cpp
pjanevskiTT: addressed

PR #2539, opened 2026-04-29
Add TlbAllocator interface; introduce KmdTlbAllocator; SimulationTlbAllocator inherits

pjanevskiTT:
### Issue

Refs #2317. Stacked on #2538.

### Description

Introduces a backend-agnostic `TlbAllocator` abstract base with two concrete implementations:

- `KmdTlbAllocator` (silicon) — delegates to `PCIDevice::allocate_tlb`. Owns the size-fallback loop when the caller passes `size=0`.
- `SimulationTlbAllocator` (simulation) — already-existing class, now inherits from `TlbAllocator` and implements `allocate()` by combining its existing index allocation with a backend-supplied `TlbHandleFactory`.

The interface is handle-first: `allocate(size, mapping) -> unique_ptr<TlbHandle>`. This matches silicon's natural API shape (KMD returns an opaque handle directly) and avoids forcing a synthetic "index" concept onto the silicon path. The simulation backend pays a small cost — a `TlbHandleFactory` callable injected at construction — so that `allocate()` can return a fully-built handle.

To make this clean, simulation `TlbHandle` subclasses now hold a `SimulationTlbAllocator*` back-pointer (instead of a `SimulationTlbManager*` back-pointer). The methods they called on the manager (`deallocate_tlb_index`, `get_tlb_address_from_index`, `get_tlb_reg_address_from_index`, `get_architecture_impl`) all live on the allocator now, so this is a pointer-type rename rather than new logic. The forwarding methods previously kept on `SimulationTlbManager` for back-compat are removed.

`SimulationTlbManager` itself no longer owns a `TlbWindowFactory` that knows about cores, ids, and configs all at once. Instead it gets two simpler callables at construction: a `TlbHandleFactory` (passed straight to its `SimulationTlbAllocator`) and a `TlbWindowBuilder` (`(handle, config) -> window`) that wraps the allocator-produced handle into the right backend-specific `TlbWindow`. The simulation `TTDevice` subclasses provide both lambdas.

This is the wiring step for #2317's allocator/factory split. The next task (add `TTDevice::get_io_window`) will route allocation through this interface.

### List of the changes

- New abstract interface `TlbAllocator` (`device/api/umd/device/chip_helpers/tlb_allocator.hpp`) with single virtual method `allocate(size, mapping) -> unique_ptr<TlbHandle>`.
- New `KmdTlbAllocator` (`device/api/umd/device/chip_helpers/kmd_tlb_allocator.hpp`, `device/chip_helpers/kmd_tlb_allocator.cpp`): silicon implementation, delegates to `PCIDevice::allocate_tlb`. Owns the size-fallback loop previously in `TLBManager::allocate_tlb_window`.
- `SimulationTlbAllocator` inherits from `TlbAllocator`. Constructor now takes a `TlbHandleFactory`. `allocate()` allocates an index and calls the factory. Adds `build_handle_for_index()` for backends like Quasar that need a handle without going through the bitmap.
- `SimulationTlbManager`:
  - Constructor takes a `TlbHandleFactory` (passed to allocator) plus a `TlbWindowBuilder` (`(handle, cfg) -> window`).
  - `allocate_tlb_window` now: `allocator_.allocate(...)` + `window_builder_(...)`.
  - The `allocate_tlb_index` / `deallocate_tlb_index` / `get_tlb_address_from_index` / `get_tlb_reg_address_from_index` / `get_architecture_impl` forwarders are removed (no longer needed).
- Simulation `TlbHandle` subclasses (`TTSimTlbHandle`, `RtlSimTlbHandle`):
  - Back-pointer changes from `SimulationTlbManager*` to `SimulationTlbAllocator*`.
  - Constructor and `create()` signatures updated.
  - `free_tlb()` calls `allocator_->deallocate_tlb_index()` directly.
  - `get_tlb_manager()` accessor renamed to `get_tlb_allocator()`. Single use site (`get_arch()` in the matching `TlbWindow` subclass) updated to chain through `allocator_->get_architecture_impl()->get_architecture()`.
- `tt_sim_tt_device.cpp` and `rtl_simulation_tt_device.cpp`: factory lambdas split into a handle-factory and a window-builder.
- `device/CMakeLists.txt`: registered `kmd_tlb_allocator.{cpp,hpp}` and `tlb_allocator.hpp`.

### Testing

CI. Locally: `cmake --build build` clean; `pre-commit run --all-files` clean; `baremetal_tests` (121 tests) and `test_utils_tests` (3 tests) pass.

### API Changes

### DIFF
--- a/device/CMakeLists.txt
+++ b/device/CMakeLists.txt
@@ -74,6 +74,7 @@ target_sources(
         chip_helpers/simulation_sysmem_manager.cpp
         chip_helpers/silicon_sysmem_manager.cpp
         chip_helpers/sysmem_buffer.cpp
+        chip_helpers/kmd_tlb_allocator.cpp
         chip_helpers/tlb_manager.cpp
         pcie/tlb_window.cpp
         pcie/silicon_tlb_window.cpp
@@ -190,7 +191,9 @@ if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.23)
                 api/umd/device/chip_helpers/sysmem_manager.hpp
                 api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
                 api/umd/device/chip_helpers/silicon_sysmem_manager.hpp
+                api/umd/device/chip_helpers/kmd_tlb_allocator.hpp
                 api/umd/device/chip_helpers/sysmem_buffer.hpp
+                api/umd/device/chip_helpers/tlb_allocator.hpp
                 api/umd/device/chip_helpers/tlb_manager.hpp
                 api/umd/device/chip_helpers/simulation_tlb_allocator.hpp
                 api/umd/device/chip_helpers/simulation_tlb_manager.hpp
--- a/device/api/umd/device/chip_helpers/kmd_tlb_allocator.hpp
+++ b/device/api/umd/device/chip_helpers/kmd_tlb_allocator.hpp
@@ -1,0 +1,31 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include "umd/device/chip_helpers/tlb_allocator.hpp"
+
+namespace tt::umd {
+
+class PCIDevice;
+class architecture_implementation;
+
+/**
+ * Silicon TLB allocator. Delegates to KMD via PCIDevice::allocate_tlb.
+ *
+ * KMD owns the actual bookkeeping. This class only handles the size-fallback
+ * loop when the caller passes size=0.
+ */
+class KmdTlbAllocator : public TlbAllocator {
+public:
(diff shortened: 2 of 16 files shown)


pjanevskiTT:
Closing in favor of a simpler design — see follow-up PR.

The unified TlbAllocator abstract base added here was superficially symmetric across silicon and simulation, but introduced unnecessary plumbing:

- KmdTlbAllocator was a one-line passthrough to PCIDevice::allocate_tlb plus a size-fallback loop — not a real abstraction over silicon
- SimulationTlbAllocator had to take a TlbHandleFactory at construction so allocate() could return a fully-built handle, mixing bookkeeping with handle construction

Cleaner alternative: drop the abstraction entirely. Each TTDevice subclass'''s eventual get_io_window() will do end-to-end construction with whatever primitive its backend exposes. SimulationTlbAllocator stays a pure bookkeeping class. The simulation TlbHandle back-pointer migration (the actual cleanup here) will be carried over into the replacement PR.

PR #2540, opened 2026-04-29
Migrate simulation TlbHandle back-pointer to SimulationTlbAllocator

pjanevskiTT:
### Issue

Refs #2317. Stacked on #2538. Followed by #2541.

### Description

The simulation `TlbHandle` subclasses (`TTSimTlbHandle`, `RtlSimTlbHandle`) hold a back-pointer to `SimulationTlbManager` purely for `deallocate_tlb_index` on destruction and a few address/arch lookups. Since #2538 moved that bookkeeping into `SimulationTlbAllocator`, the manager is now a passthrough for these calls — the natural owner is the allocator itself.

This PR migrates the back-pointer from `SimulationTlbManager*` to `SimulationTlbAllocator*` end-to-end:

- `TTSimTlbHandle` and `RtlSimTlbHandle` hold a `SimulationTlbAllocator*` and call `allocator_->deallocate_tlb_index()` / `get_tlb_address_from_index()` / `get_tlb_reg_address_from_index()` / `get_architecture_impl()` directly.
- `TlbWindowFactory` typedef on `SimulationTlbManager` takes `SimulationTlbAllocator*` as its first argument instead of `SimulationTlbManager*`. The simulation `TTDevice` factory lambdas pass through unchanged in shape.
- `SimulationTlbManager`'s passthrough forwarders introduced in #2538 (`allocate_tlb_index`, `deallocate_tlb_index`, `get_tlb_size_from_index`, `get_tlb_address_from_index`, `get_tlb_reg_address_from_index`, `get_architecture_impl`) are removed — nothing reaches into the manager for these anymore. A `get_allocator()` accessor is added so callers that need the allocator can still get to it.
- The single use site of the old `get_tlb_manager()` accessor on each handle (`get_arch()` in the matching `TlbWindow` subclass) is updated to chain through `allocator_->get_architecture_impl()->get_architecture()` instead of going `manager_ → tt_device_ → arch`.

This keeps `SimulationTlbAllocator` purely about bookkeeping — no `TlbHandleFactory` injection, no `allocate()` method, no abstract `TlbAllocator` base. Subsequent PRs in the stack add `TTDevice::get_io_window(...)` (#2541) and then move end-to-end construction into the simulation `TTDevice` subclasses (#2544) — silicon calls `pci_device_->allocate_tlb()` directly, simulation calls `allocator_->allocate_tlb_index()` and builds the handle/window inline. No abstract `TlbAllocator` interface is needed because each backend speaks its own primitive at that layer.

[...]

PR #2541, opened 2026-04-29
Add TTDevice::get_io_window

pjanevskiTT:
### Issue

Refs #2317. Stacked on #2540. Followed by #2544.

### Description

Adds a virtual `TTDevice::get_io_window(config, mapping, size)` entry point for TLB-window allocation, with a default silicon implementation that uses `PCIDevice::allocate_tlb` plus the size-fallback loop and wraps the resulting handle in a `SiliconTlbWindow`. `WormholeTTDevice` and `BlackholeTTDevice` use the default unchanged.

Simulation `TTDevice`s (`TTSimTTDevice`, `RtlSimulationTTDevice`) override `get_io_window` to forward to their existing `tlb_manager_->allocate_tlb_window` — `SimulationTlbManager` and the bitmap allocator + factory it owns continue to do the actual sim allocation. Removing `SimulationTlbManager` is the next PR in the stack (#2544).

`TLBManager::allocate_tlb_window` collapses to a one-liner that delegates to `tt_device_->get_io_window`. Size-fallback logic moves into the silicon default impl. `TLBManager::configure_tlb` logging switches from `pci_device->get_device_num` (silicon-only, would crash on sim) to `tt_device->get_communication_device_id` which is universal.

### List of the changes

- **`TTDevice` base** (`tt_device.hpp` / `tt_device.cpp`): add `virtual std::unique_ptr<TlbWindow> get_io_window(tlb_data config, TlbMapping mapping = WC, size_t size = 0)` with default impl that uses `PCIDevice::allocate_tlb` + size fallback + `SiliconTlbWindow` wrapping.
- **`TTSimTTDevice` / `RtlSimulationTTDevice`**: override `get_io_window` with a one-line forward to `tlb_manager_->allocate_tlb_window(...)`. `SimulationTlbManager` keeps doing the bitmap allocation + factory work for now.
- **`TLBManager::allocate_tlb_window`** (`tlb_manager.cpp`): one-liner — `return tt_device_->get_io_window(config, mapping, tlb_size);`.
- **`TLBManager::configure_tlb`** logging: `pci_device->get_device_num()` → `tt_device->get_communication_device_id()`.

### Testing

CI.

### API Changes

### DIFF
--- a/device/api/umd/device/tt_device/tt_device.hpp
+++ b/device/api/umd/device/tt_device/tt_device.hpp
@@ -445,6 +445,20 @@ public:
 
     virtual TLBManager *get_tlb_manager() { return nullptr; }
 
+    /**
+     * Allocate a TlbWindow for use by callers (typically TLBManager).
+     *
+     * Default implementation uses PCIDevice::allocate_tlb (silicon path) and
+     * wraps the resulting handle in a SiliconTlbWindow. Simulation TTDevice
+     * subclasses override this to allocate from their backend-specific path.
+     *
+     * @param config tlb_data configuration applied to the new window.
+     * @param mapping UC or WC.
+     * @param size Requested TLB size in bytes (0 means try arch-supported sizes in order).
+     */
+    virtual std::unique_ptr<TlbWindow> get_io_window(
+        tlb_data config, TlbMapping mapping = TlbMapping::WC, size_t size = 0);
+
     virtual void dma_write_to_device(const void *src, size_t size, tt_xy_pair core, uint64_t addr);
 
     virtual void dma_read_from_device(void *dst, size_t size, tt_xy_pair core, uint64_t addr);
--- a/device/api/umd/device/tt_device/tt_sim_tt_device.hpp
+++ b/device/api/umd/device/tt_device/tt_sim_tt_device.hpp
@@ -84,6 +84,9 @@ public:
 
     TLBManager *get_tlb_manager() override;
 
+    std::unique_ptr<TlbWindow> get_io_window(
+        tlb_data config, TlbMapping mapping = TlbMapping::WC, size_t size = 0) override;
+
     uint64_t bar0_base = 0;
     uint64_t bar4_base = 0;
 
--- a/device/api/umd/device/tt_device/tt_sim_tt_device.hpp
+++ b/device/api/umd/device/tt_device/tt_sim_tt_device.hpp
@@ -84,8 +84,7 @@ public:
 
     TLBManager *get_tlb_manager() override;
 
-    std::unique_ptr<TlbWindow> get_io_window(
-        tlb_data config, TlbMapping mapping = TlbMapping::WC, size_t size = 0) override;
+    std::unique_ptr<TlbWindow> get_io_window(tlb_data config, TlbMapping mapping, size_t size) override;
 
     uint64_t bar0_base = 0;
     uint64_t bar4_base = 0;
--- a/device/tt_device/tt_device.cpp
+++ b/device/tt_device/tt_device.cpp
@@ -266,6 +266,29 @@ bool TTDevice::is_noc_hung(NocId noc, TTDevice::HangAction action) {
     return false;
 }
 
+std::unique_ptr<TlbWindow> TTDevice::get_io_window(tlb_data config, TlbMapping mapping, size_t size) {
+    PCIDevice *pci = get_pci_device();
+    if (pci == nullptr) {
+        UMD_THROW(error::RuntimeError, "TTDevice::get_io_window default implementation requires a PCIDevice.");
+    }
+
+    if (size != 0) {
+        return std::make_unique<SiliconTlbWindow>(pci->allocate_tlb(size, mapping), config);
+    }
+
+    // Caller didn't specify a size — try arch-supported sizes in preference order.
+    const std::vector<size_t> &possible_sizes = get_architecture_implementation()->get_tlb_sizes();
+    for (const auto &s : possible_sizes) {
+        try {
+            return std::make_unique<SiliconTlbWindow>(pci->allocate_tlb(s, mapping), config);
+        } catch (const std::exception &e) {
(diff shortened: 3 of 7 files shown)


device/tt_device/tt_device.cpp
```diff
@@ -265,6 +265,29 @@ bool TTDevice::is_noc_hung(NocId noc, TTDevice::HangAction action) {
     return false;
 }
 
+std::unique_ptr<TlbWindow> TTDevice::get_io_window(tlb_data config, TlbMapping mapping, size_t size) {
```
pjanevskiTT, line 269: Putting the silicon implementation in the base `TTDevice` is a small layering inversion — the base now depends on `SiliconTlbWindow` and silently assumes `get_pci_device()` is non-null (the runtime check at line 271 is the only safety net). The cleaner shape would be to make `get_io_window` pure-virtual on `TTDevice` and put this body on a silicon-specific intermediate (or directly on `WormholeTTDevice` / `BlackholeTTDevice`). Not necessarily a blocker for this PR — but #2544 is going to push more of this allocation work into the subclasses anyway, so worth weighing whether to fix the layering now while it's still a small change rather than after #2544 lands.

device/api/umd/device/tt_device/tt_sim_tt_device.hpp
```diff
@@ -80,6 +80,9 @@ class TTSimTTDevice : public TTDevice {
 
     TLBManager *get_tlb_manager() override;
 
+    std::unique_ptr<TlbWindow> get_io_window(
+        tlb_data config, TlbMapping mapping = TlbMapping::WC, size_t size = 0) override;
```
pjanevskiTT: Default arguments are re-declared on the override (`mapping = TlbMapping::WC, size = 0`). C++ resolves default arguments based on the static type of the callee, so a base-pointer call uses base defaults and a derived-pointer call uses the derived defaults — if these ever diverge, you get silently inconsistent behaviour at different call sites. Convention is to put the defaults only on the base declaration: `std::unique_ptr<TlbWindow> get_io_window(tlb_data config, TlbMapping mapping, size_t size) override;`. (Same applies in `rtl_simulation_tt_device.hpp:77`.)

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -430,6 +430,20 @@ class TTDevice {
 
     virtual TLBManager *get_tlb_manager() { return nullptr; }
 
+    /**
+     * Allocate a TlbWindow for use by callers (typically TLBManager).
```
broskoTT, line 449: Well... not TLB if this header is supposed to be generic

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -430,6 +430,20 @@ class TTDevice {
...
+     * Default implementation uses PCIDevice::allocate_tlb (silicon path) and
+     * wraps the resulting handle in a SiliconTlbWindow. Simulation TTDevice
+     * subclasses override this to allocate from their backend-specific path.
+     *
+     * @param config tlb_data configuration applied to the new window.
+     * @param mapping UC or WC.
+     * @param size Requested TLB size in bytes (0 means try arch-supported sizes in order).
+     */
+    virtual std::unique_ptr<TlbWindow> get_io_window(
+        tlb_data config, TlbMapping mapping = TlbMapping::WC, size_t size = 0);
```
broskoTT, line 460: tlb_data and TlbMapping should be generic

device/api/umd/device/tt_device/tt_device.hpp
broskoTT: Ok so this is supposed to be addressed in https://github.com/tenstorrent/tt-umd/pull/2543 I guess

device/api/umd/device/tt_device/tt_device.hpp
broskoTT: I understand it will be addressed in following PRs

broskoTT: This one seems to be going in the right direction anyway

device/chip_helpers/simulation_tlb_manager.cpp
broskoTT, line 54: consider just replacing here the getter since you use architecture at a single place

device/tt_device/tt_device.cpp
```diff
@@ -266,6 +266,29 @@ bool TTDevice::is_noc_hung(NocId noc, TTDevice::HangAction action) {
...
+    PCIDevice *pci = get_pci_device();
+    if (pci == nullptr) {
+        UMD_THROW(error::RuntimeError, "TTDevice::get_io_window default implementation requires a PCIDevice.");
```
broskoTT: Consider using UMD_ASSERT

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -318,4 +309,8 @@ void TTSimTTDevice::noc_multicast_write(void* src, size_t size, uint64_t addr) {
 
 TLBManager* TTSimTTDevice::get_tlb_manager() { return static_cast<TLBManager*>(tlb_manager_.get()); }
 
+std::unique_ptr<TlbWindow> TTSimTTDevice::get_io_window(tlb_data config, TlbMapping mapping, size_t size) {
+    return tlb_manager_->allocate_tlb_window(config, mapping, size);
```
broskoTT, line 322: I know I saw it somewhere in stack, but I assume this will end up being call to SimulationTlbAllocator. Just double checking

broskoTT (approved):

device/tt_device/tt_sim_tt_device.cpp
pjanevskiTT: it will

device/tt_device/tt_device.cpp
pjanevskiTT: has a lot of code duplication, won't change it for now

PR #2542, opened 2026-04-29
Use `UMD_ASSERT` in codebase

aleksamarkovicTT:
### Issue
/

### Description
Change all `TT_ASSERT` into `UMD_ASSERT(..., RuntimeError, ...)`.
Removed `TT_ASSERT` and related files as they are now completely unused.

### List of the changes
- Changed all `TT_ASSERT` to `UMD_ASSERT`.
- Removed `assert.hpp` and `test_assert.hpp`
- Made `error_detail.hpp` private, users should include `error.hpp`.

### Testing
CI

### API Changes


device/utils/robust_mutex.cpp
```diff
@@ -61,12 +59,18 @@ class CriticalSectionScopeGuard {
...
+        UMD_ASSERT(
+            flock(fd_, LOCK_EX) == 0,
+            error::RuntimeError,
+            fmt::format("flock failed for mutex {} errno: {}", mutex_name_, std::to_string(errno)));
         int err = pthread_mutex_lock(pthread_mutex_);
         if (err != 0) {
             // Try to unlock the flock without handling further exceptions.
             flock(fd_, LOCK_UN);
-            TT_ASSERT(false, "pthread_mutex_lock() failed for mutex {} errno: {}", mutex_name_, std::to_string(err));
+            UMD_ASSERT(
```
pjanevskiTT: The condition here is a literal `false` — this branch is reached only after `pthread_mutex_lock` already failed and we want to throw unconditionally. `UMD_ASSERT(false, ...)` is awkward (and is the reason the `// NOLINT(readability-simplify-boolean-expr)` had to be added in `error_detail.hpp`). `UMD_THROW(error::RuntimeError, fmt::format(...))` expresses the intent more clearly here.

device/api/umd/device/utils/error_detail.hpp
```diff
@@ -301,7 +306,7 @@ class UmdException : public UmdBaseException {
  */
 #define UMD_ASSERT(condition, error_type, ...)                                                                       \
     do {                                                                                                             \
-        if (!(condition)) {                                                                                          \
+        if (!(condition)) { /* NOLINT(readability-simplify-boolean-expr) */                                          \
```
pjanevskiTT, line 309: The mass conversion produced lots of `UMD_ASSERT(cond, error::RuntimeError, fmt::format(fmt_str, args...))` callsites — the `fmt::format` wrapper is mechanical noise that the old `TT_ASSERT` macro handled automatically. Worth considering an additional macro (e.g. `UMD_ASSERT_FMT(cond, error_type, fmt_str, args...)`) that internally calls `fmt::format`, or a varadic `RuntimeError(fmt_str, args...)` constructor. Not blocking, but it would make future call sites and any further migrations cleaner.

aleksamarkovicTT: > the fmt::format wrapper is mechanical noise that the old TT_ASSERT macro handled automatically

This is not possible because it would leak fmt as a public dependency, which we want to avoid. TT_ASSERT could not be a public UMD API because it would require spdlog and fmt for every API client.

device/utils/robust_mutex.cpp
aleksamarkovicTT: Will change, but this is not the reason for the nolint.

pjanevskiTT (approved):

device/api/umd/device/utils/error_detail.hpp
```diff
@@ -4,6 +4,11 @@
 
 #pragma once
 
+// IWYU pragma: private, include "umd/device/utils/error.hpp".
+
+#ifndef UMD_ERROR_HPP_INTERNAL_INCLUDE
+#error "error_detail.hpp is a private header. Include umd/device/utils/error.hpp instead."
```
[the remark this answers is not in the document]
aleksamarkovicTT: Will remove includes. #pragma once is the reason it doesn't error out.

PR #2543, opened 2026-04-29
Rename TlbWindow -> IOWindow, TlbHandle -> IOHandle

pjanevskiTT:
### Issue

Refs #2317. Stacked on #2544. Last PR in the #2317 stack.

### Description

Mechanical rename — the base `TlbWindow` and `TlbHandle` classes generalize to `IOWindow` and `IOHandle` so the same abstraction can later represent windows backed by something other than a TLB (e.g., a slice of BAR space). All current methods (including `*_reconfigure` and `configure(tlb_data)`) are preserved as-is — splitting the access-only base from the TLB-specific reconfigure surface is a follow-up.

Concrete subclasses keep their `*Tlb*` names — `SiliconTlbWindow`, `SiliconTlbHandle`, `TTSimTlbWindow`, `TTSimTlbHandle`, `RtlSimTlbWindow`, `RtlSimTlbHandle` are still TLB-specific implementations of the now-generalized `IOWindow` / `IOHandle` interface. Their inheritance updates from `: public TlbWindow/TlbHandle` to `: public IOWindow/IOHandle`.

### List of the changes

- File renames:
  - `device/api/umd/device/pcie/tlb_window.hpp` → `io_window.hpp`
  - `device/api/umd/device/pcie/tlb_handle.hpp` → `io_handle.hpp`
  - `device/pcie/tlb_window.cpp` → `io_window.cpp`
- Class renames in those files: `class TlbWindow` → `class IOWindow`, `class TlbHandle` → `class IOHandle`. Methods, members, and internal logic unchanged.
- Word-boundary rename `\bTlbWindow\b` → `IOWindow` and `\bTlbHandle\b` → `IOHandle` across the repo. Subclass names (`SiliconTlbWindow`, `TTSimTlbHandle`, etc.) are unaffected because the boundary regex doesn't match across an internal word.
- Include path swaps `umd/device/pcie/tlb_window.hpp` → `umd/device/pcie/io_window.hpp` and `umd/device/pcie/tlb_handle.hpp` → `umd/device/pcie/io_handle.hpp`.
- `device/CMakeLists.txt`: source filename and `FILE_SET` header paths updated to the new names.

### Testing

CI.

### API Changes

[...]

PR #2544, opened 2026-04-29
Delete SimulationTlbManager; sim chips own their own TLBManager

pjanevskiTT:
### Issue

Refs #2317. Stacked on #2541. Followed by #2543.

### Description

With `TTDevice::get_io_window` in place (#2541), `SimulationTlbManager` no longer contributes anything beyond a thin override that uses its own `SimulationTlbAllocator` + `TlbWindowFactory`. This PR moves the allocator and the handle/window construction into the simulation `TTDevice` subclasses, then deletes `SimulationTlbManager`.

After this PR:
- `SimulationTlbAllocator` is owned directly by the simulation `TTDevice` subclass that uses it.
- `TLBManager` lives entirely on the `Chip` layer. `TTDevice::get_tlb_manager` virtual is removed; `TTSimChip` and `RtlSimulationChip` each own their own `TLBManager` (parallel to how `LocalChip` owns one for silicon).
- The simulation `TTDevice`'s `get_io_window` override does end-to-end construction: allocate index → build sim handle → wrap in sim window.

### List of the changes

- **`TTSimTTDevice` / `RtlSimulationTTDevice`**:
  - Hold a `std::unique_ptr<SimulationTlbAllocator> tlb_allocator_` directly (was inside `SimulationTlbManager`).
  - `get_io_window` override now does the full allocate → handle → window pipeline inline (was forwarding to `tlb_manager_->allocate_tlb_window`).
  - Drop `tlb_manager_` field and `get_tlb_manager()` override.
  - Constructor populates `cached_tlb_window_` directly: `get_io_window({}, WC, default_size)` for Wormhole/Blackhole, special bypass with index 0 for Quasar (no real TLBs).
- **`TTDevice` base** (`tt_device.hpp`): remove `virtual TLBManager* get_tlb_manager() { return nullptr; }`.
- **`TTSimChip` / `RtlSimulationChip`**: own a `std::unique_ptr<TLBManager> tlb_manager_` field, constructed from their `tt_device_`. `get_tlb_manager()` returns `tlb_manager_.get()` (was forwarding to `tt_device_->get_tlb_manager()`).
- **Delete** `device/api/umd/device/chip_helpers/simulation_tlb_manager.hpp` and `device/chip_helpers/simulation_tlb_manager.cpp`. Remove from `device/CMakeLists.txt`.
- **Includes** in `tt_sim_tlb_window.cpp` / `rtl_sim_tlb_window.cpp` swap the deleted `simulation_tlb_manager.hpp` for `simulation_tlb_allocator.hpp`.

### Testing

CI.

### API Changes


device/tt_device/rtl_simulation_tt_device.cpp
```diff
@@ -95,16 +95,45 @@ RtlSimulationTTDevice::RtlSimulationTTDevice(
...
-            return std::make_unique<RtlSimTlbWindow>(std::move(handle), comm, cfg);
-        });
-    cached_tlb_window_ = tlb_manager_->allocate_default_tlb_window();
+    tlb_allocator_ = std::make_unique<SimulationTlbAllocator>(/*bar0_base=*/0, architecture_impl_.get());
+
+    // Allocate the cached default TLB window. Quasar has no real TLBs; the communicator handles
+    // all I/O underneath. The 4GB size for Quasar is a dummy value — it just needs to be large
+    // enough so that TlbWindow::validate doesn't reject any valid access (size 0 would cause
+    // division by zero in RtlSimTlbHandle::configure).
+    static constexpr size_t SIZE_2MB = 2 * 1024 * 1024;
```
broskoTT, line 114: We should really have this constants somewhere in some utils user around the whole codebase... You can ignore for this PR

broskoTT (approved):
Thanks for cleaning this up

device/tt_device/rtl_simulation_tt_device.cpp
```diff
@@ -95,16 +95,45 @@ RtlSimulationTTDevice::RtlSimulationTTDevice(
...
+            log_debug(
+                LogUMD,
+                "Architecture {} does not support TLB allocation, leaving cached_tlb_window_ null.",
+                tt::arch_to_str(arch));
+            break;
+    }
+}
+
+std::unique_ptr<TlbWindow> RtlSimulationTTDevice::get_io_window(tlb_data config, TlbMapping mapping, size_t size) {
+    int tlb_index = tlb_allocator_->allocate_tlb_index(size);
```
pjanevskiTT, line 137: On Quasar, `tlb_allocator_->allocate_tlb_index(size)` will always return `-1` because `initialize_architecture_config` only populates pools for `WORMHOLE_B0` / `BLACKHOLE` — Quasar's `tlb_*_count_` all stay 0. Constructor-time Quasar handling sidesteps the allocator via a direct `RtlSimTlbHandle::create(...)`, but any later `TLBManager::allocate_tlb_window` call routed through this path will throw. If this PR lands after #2588 (Quasar bypass), the dynamic-allocation behavior #2588 was specifically adding for Quasar is silently removed. Either (a) replicate the Quasar bypass here, or (b) confirm Quasar truly never needs more than the cached window and document the invariant.

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -95,16 +95,46 @@ TTSimTTDevice::TTSimTTDevice(
...
+            log_debug(
+                LogUMD,
+                "Architecture {} does not support TLB allocation, leaving cached_tlb_window_ null.",
+                tt::arch_to_str(arch));
+            break;
+    }
+}
+
+std::unique_ptr<TlbWindow> TTSimTTDevice::get_io_window(tlb_data config, TlbMapping mapping, size_t size) {
+    int tlb_index = tlb_allocator_->allocate_tlb_index(size);
```
pjanevskiTT, line 134: Same Quasar regression as `rtl_simulation_tt_device.cpp:130`. If #2588 is meant to stay, fold the bypass into `get_io_window` (or `TLBManager::allocate_tlb_window`) here too.

device/tt_device/tt_device.cpp
```diff
@@ -266,6 +266,29 @@ bool TTDevice::is_noc_hung(NocId noc, TTDevice::HangAction action) {
...
+        return std::make_unique<SiliconTlbWindow>(pci->allocate_tlb(size, mapping), config);
+    }
+
+    // Caller didn't specify a size — try arch-supported sizes in preference order.
+    const std::vector<size_t> &possible_sizes = get_architecture_implementation()->get_tlb_sizes();
+    for (const auto &s : possible_sizes) {
+        try {
+            return std::make_unique<SiliconTlbWindow>(pci->allocate_tlb(s, mapping), config);
+        } catch (const std::exception &e) {
+            log_error(LogUMD, "Failed to allocate TLB window of size {}: {}", s, e.what());
```
pjanevskiTT: `log_error("Failed to allocate TLB window of size {}: {}", s, e.what())` inside the size-fallback loop fires for every size that fails before eventually succeeding — which is the expected control flow when `size==0`, not an error. This is moved from the old `TLBManager::allocate_tlb_window` (same issue pre-existing), but worth downgrading to `log_debug` while you're touching this code.

Issue #2150, opened 2026-02-24
Stateless NOC


broskoTT:
https://tenstorrent.atlassian.net/wiki/spaces/LLK/pages/2016215192/Design+agreements+Feb+2026

Per agreement, remove statefulness for NOC.
This means adding "noc" argument on read/write api, and probably on other functions such as  TTDev::init() and many constructors such as  Chip, TopoDiscovery

broskoTT:
In progress, 2 PRs already merged

broskoTT:
Progressing

nbuncicTT:
2 PRs published, 1 PR in draft, and upcoming 3 PRs.

nbuncicTT:
Added 1 PR and two are pending verification.

nbuncicTT:
Update: 2 PRs more merged, more upcoming

broskoTT:
Progressing slowly, but PRs are flowing.
There will be some two phase changes with exalens and metal and telemetry.

PR #2546, opened 2026-04-29
Make TlbWindow accept NOC parameter

nbuncicTT:
### Issue
#2150 

### Description
This pull request updates the TLB window APIs and their usage throughout the device codebase to explicitly require a `NocId` parameter for all reconfigurable block read/write and multicast operations. This change makes the NoC (Network-on-Chip) selection explicit.

### List of the changes
- Modified APIs of TlbWindow to accept NOC (hence make it "stateless" in terms of the APIs on this layer).
- Changed usage where it was needed for the new API to be used.

### Testing
CI

### API Changes
- Modified APIs of TlbWindow to accept NOC (hence make it "stateless" in terms of the APIs on this layer).


pjanevskiTT (approved):

aleksamarkovicTT (approved):
I agree with the choice of not making noc_id have a default arg. value.

device/pcie/tlb_window.cpp
```diff
@@ -24,13 +23,14 @@ TlbWindow::TlbWindow(std::unique_ptr<TlbHandle> handle, const tlb_data config) :
...
-void TlbWindow::read_block_reconfigure(void* mem_ptr, tt_xy_pair core, uint64_t addr, size_t size, uint64_t ordering) {
+void TlbWindow::read_block_reconfigure(
+    void* mem_ptr, tt_xy_pair core, uint64_t addr, size_t size, NocId noc_id, uint64_t ordering) {
     uint8_t* buffer_addr = static_cast<uint8_t*>(mem_ptr);
     tlb_data config{};
     config.local_offset = addr;
     config.x_end = core.x;
     config.y_end = core.y;
-    config.noc_sel = is_selected_noc1() ? 1 : 0;
+    config.noc_sel = static_cast<uint64_t>(noc_id);
```
pjanevskiTT, line 33: `static_cast<uint64_t>(noc_id)` writes the raw enum value into `noc_sel`, so `NocId::SYSTEM_NOC` (2) ends up in a field that previously was clamped to 0/1 by `is_selected_noc1() ? 1 : 0`. `noc_sel` is a single bit in hardware; passing 2 would silently truncate/corrupt the TLB config. Either restrict to NOC0/NOC1 here (assert/mask) or match the old semantics with `noc_id == NocId::NOC1 ? 1 : 0`. Same applies to the analogous assignments further down in `write_block_reconfigure` and `noc_multicast_write_reconfigure`.

nbuncicTT: I don't feel like this make sense, because the low level component will get this enum which will be cast to the correct value. If we pass SMN (which is 2) in the API for WH/BH, this should be caught earlier.

Issue #1314, opened 2025-09-16
Flaky VirtualCoordinateBroadcast test


pjanevskiTT:
Flaky VirtualCoordinateBroadcast test was disabled since it was blocking merge queue

Enable them again and investigate

Issue #1312, opened 2025-09-16
Investigate n300 Broadcast failures


broskoTT:
One such failure here: https://github.com/tenstorrent/tt-umd/actions/runs/17762368311/job/50478334029?pr=1241

This seems to have started happening on main.

As far as I see it happens only on n300 machines, I've seen it on both viommu and nonviommu.

Try to repro locally.
Try to bisect.

Then analyse further.

broskoTT:
This was likely happening due to deassert/assert not functioning properly, which now does. The tests aren't flaky anymore.

PR #2547, opened 2026-04-29
Reenable Ethernet Broadcast tests

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/2521
https://github.com/tenstorrent/tt-umd/issues/1314
https://github.com/tenstorrent/tt-umd/issues/1312
https://github.com/tenstorrent/tt-umd/issues/797

### Description
These tests were disabled in https://github.com/tenstorrent/tt-umd/commit/207cb3c1ab8aba2aef9e08066d5c0bbbb0d4e953
They were likely flaky due to start_device not being safe for tests at the moment.
And there was a removal of VIRTUAL coords effort, which mistakenly changed some of the variables from "virtual" to "translated". The broadcast operates on VIRTUAL coords still.

### List of the changes
- Added TRANSLATED to VIRTUAL vector in wormole_implementation
- Reenable Broadcast tests.
- Add another which performs broadcast per chip
- Remove all Broadcast Blackhole specific code, since that is dead code.

### Testing
This PR includes tests

### API Changes


device/api/umd/device/cluster.hpp
```diff
@@ -758,8 +758,8 @@ class Cluster {
     ClusterOptions options_;
 
     std::map<std::set<ChipId>, std::unordered_map<ChipId, std::vector<std::vector<int>>>> bcast_header_cache;
-    bool use_ethernet_broadcast = true;
-    bool use_translated_coords_for_eth_broadcast = true;
+    bool use_ethernet_broadcast;
```
pjanevskiTT: Removing the in-class default initializer for both `use_ethernet_broadcast` and `use_translated_coords_for_eth_broadcast` makes them default-uninitialized. `use_translated_coords_for_eth_broadcast` is only assigned inside the `arch_name == tt::ARCH::WORMHOLE_B0` branch in `construct_cluster`, so for non-WH architectures (Blackhole, Grendel, mock/sim) the field is left uninitialized. While the new `broadcast_write_to_cluster` throws for non-WH today, any future code path or constructor failure that reads the flag (e.g. logging, assertions during teardown) will read indeterminate memory. Either restore the `= true` default initializers or assign in the constructor unconditionally.

device/cluster.cpp
```diff
@@ -170,9 +170,9 @@ void Cluster::construct_cluster(const uint32_t& num_host_mem_ch_per_mmio_device,
 
     // Disable dependency to ethernet firmware for all BH devices and WH devices with all chips having MMIO (e.g. UBB
     // Galaxy, or P300).
-    if (remote_chip_ids_.empty() || chip_type != ChipType::SILICON || arch_name == tt::ARCH::BLACKHOLE) {
-        use_ethernet_broadcast = false;
-    }
+    // The ethernet firmware also requires host memory for broadcasting.
+    use_ethernet_broadcast = chip_type == ChipType::SILICON && arch_name == tt::ARCH::WORMHOLE_B0 &&
+                             !remote_chip_ids_.empty() && num_host_mem_ch_per_mmio_device > 0;
```
pjanevskiTT, line 174: The new condition `use_ethernet_broadcast = chip_type == ChipType::SILICON && arch_name == tt::ARCH::WORMHOLE_B0 && !remote_chip_ids_.empty() && num_host_mem_ch_per_mmio_device > 0` silently disables ethernet broadcast (and falls back to per-chip loops) whenever a user constructs a multi-chip WH cluster with `num_host_mem_ch_per_mmio_device == 0`. The previous behavior used eth broadcast in that case. This behavior change isn't documented in the new comment ("ethernet firmware also requires host memory") — please confirm host memory is actually required by the eth FW broadcast (vs. only by the new tests), and consider asserting/logging when a user explicitly requests broadcast but host mem is absent rather than silently degrading.

device/cluster.cpp
```diff
@@ -984,7 +935,7 @@ void Cluster::broadcast_tensix_risc_reset_to_cluster(const TensixSoftResetOption
         return;
     }
     // If ethernet broadcast is not supported, do it one by one.
-    if (!use_ethernet_broadcast) {
+    if (!use_ethernet_broadcast || arch_name == tt::ARCH::BLACKHOLE) {
```
pjanevskiTT: arch_name == Blackhole should maybe be arch_name != WORMHOLE? What about grendel?

device/cluster.cpp
broskoTT: I've added a comment already. Yes, this is intentional

device/api/umd/device/cluster.hpp
broskoTT: I'll do both

tests/wormhole/test_cluster_wh.cpp
```diff
@@ -561,13 +568,129 @@ TEST(SiliconDriverWH, DISABLED_VirtualCoordinateBroadcast) {
...
+
+TEST(SiliconDriverWH, VirtualCoordinateBroadcastPerChip) {
+    // Broadcast multiple vectors to tensix and dram grid. Verify broadcasted data is read back correctly.
+    Cluster cluster(ClusterOptions{.num_host_mem_ch_per_mmio_device = 1});
+    set_barrier_params(cluster);
+    auto mmio_devices = cluster.get_target_mmio_device_ids();
+
+    test_utils::safe_test_cluster_start(&cluster);
+    auto eth_version = cluster.get_ethernet_firmware_version();
+    bool virtual_bcast_supported = (eth_version >= SemVer(6, 8, 0) || eth_version == SemVer(6, 7, 241)) &&
```
aleksamarkovicTT: I removed these checks because they are below the min. supported ERISC FW (6.14.+) version, that part of the condition will never be be false, so you can remove them.
It's just magic numbers by this point, and they don't amount to anything.

aleksamarkovicTT: It would throw on cluster creation if the version is below 6.14.

device/cluster.cpp
```diff
@@ -799,109 +799,60 @@ void Cluster::broadcast_write_to_cluster(
...
-                use_translated_coords_for_eth_broadcast);
+                cols_to_exclude_for_col_5_bcast,
+                false);
         }
+    } else {
+        UMD_ASSERT(
+            use_translated_coords_for_eth_broadcast or
+                valid_tensix_broadcast_grid(rows_to_exclude, columns_to_exclude, architecture_implementation.get()),
+            error::RuntimeError,
+            "Must broadcast to all tensix rows when ERISC FW is < 6.8.0.");
```
aleksamarkovicTT, line 869: How is the message related to the assert condition?

device/cluster.cpp
```diff
@@ -799,109 +799,60 @@ void Cluster::broadcast_write_to_cluster(
...
-                    false);
-            }
-        } else {
-            UMD_ASSERT(
-                use_translated_coords_for_eth_broadcast or
-                    valid_tensix_broadcast_grid(rows_to_exclude, columns_to_exclude, architecture_implementation.get()),
-                error::RuntimeError,
-                "Must broadcast to all tensix rows when ERISC FW is < 6.8.0.");
+    if (arch_name != tt::ARCH::WORMHOLE_B0) {
+        UMD_THROW(error::RuntimeError, "Broadcast write is only supported on Wormhole architecture.");
```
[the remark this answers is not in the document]
broskoTT: This will be cleared up in the following PR which will implement a fallback mechanism for non wormhole implementations

device/cluster.cpp
broskoTT: This is updated in the following PR.
The code is unchanged here, git shows it as changed due to whitespace change only

tests/wormhole/test_cluster_wh.cpp
```diff
@@ -436,17 +438,19 @@ TEST(SiliconDriverWH, MultiThreadedMemBar) {
     cluster.close_device();
 }
 
-TEST(SiliconDriverWH, DISABLED_BroadcastWrite) {
+TEST(SiliconDriverWH, BroadcastWrite) {
     // Broadcast multiple vectors to tensix and dram grid. Verify broadcasted data is read back correctly.
-    Cluster cluster;
+    Cluster cluster(ClusterOptions{.num_host_mem_ch_per_mmio_device = 1});
     set_barrier_params(cluster);
     auto mmio_devices = cluster.get_target_mmio_device_ids();
```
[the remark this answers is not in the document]
broskoTT: Nope, I want to run it for local only as well. I do want to have fallback mechanism in the loop

tests/wormhole/test_cluster_wh.cpp
broskoTT: Right, will update the test to skip this check, it is just a leftover

device/cluster.cpp
```diff
@@ -158,21 +158,19 @@ void Cluster::construct_cluster(const ChipType& chip_type) {
...
+    }
 
-        if (arch_name == tt::ARCH::WORMHOLE_B0) {
-            // Virtual coordinates can be used for broadcast headers if NOC translation is enabled.
-            use_translated_coords_for_eth_broadcast = true;
-            for (const auto& chip : all_chip_ids_) {
-                use_translated_coords_for_eth_broadcast &= get_soc_descriptor(chip).noc_translation_enabled;
-            }
-        }
+    // Virtual coordinates can be used for broadcast headers if NOC translation is enabled.
```
pjanevskiTT: I think comment and the code are out of sync, one says translated, the other one says virtual

pjanevskiTT (approved):

device/cluster.cpp
broskoTT: Yes, and it is already included in use_ethernet_broadcast so not needed here. Removed it

PR #2548, opened 2026-04-29
Make Device Protocol NOC stateless

nbuncicTT:
### Issue
#2150 

### Description
This pull request refactors device communication APIs and implementations to explicitly require a `NocId` parameter that specifies the target NoC (Network-on-Chip) for memory and DMA operations. The change is applied consistently across protocol interfaces, PCIe/JTAG/Remote protocol implementations.

### List of the changes
- DeviceProtocol in it's API uses the NOC ID and the users of these APIs are modified to accept this new API.

### Testing
CI

### API Changes
Mentioned in list of changes.


aleksamarkovicTT (approved):

device/tt_device/protocol/jtag_protocol.cpp
```diff
@@ -18,15 +17,15 @@ JtagProtocol::JtagProtocol(std::unique_ptr<JtagDevice> jtag_device, uint8_t jlin
 
 JtagProtocol::~JtagProtocol() = default;
 
-void JtagProtocol::write_to_device(const void* mem_ptr, tt_xy_pair core, uint64_t addr, size_t size) {
-    jtag_device_->write(mmio_id_, mem_ptr, core.x, core.y, addr, size, is_selected_noc1() ? 1 : 0);
+void JtagProtocol::write_to_device(const void* mem_ptr, tt_xy_pair core, uint64_t addr, size_t size, NocId noc_id) {
+    jtag_device_->write(mmio_id_, mem_ptr, core.x, core.y, addr, size, static_cast<uint64_t>(noc_id));
```
pjanevskiTT: `JtagDevice::write` / `read` take `uint8_t noc_id` (see `device/api/umd/device/jtag/jtag_device.hpp:81,86`). Casting `NocId` (an enum class with `uint8_t` underlying type) to `uint64_t` here will be implicitly narrowed at the call. Should be `static_cast<uint8_t>(noc_id)`. Same applies on line 25.

device/tt_device/protocol/remote_protocol.cpp
```diff
@@ -16,15 +16,17 @@ namespace tt::umd {
 RemoteProtocol::RemoteProtocol(std::unique_ptr<RemoteCommunication> remote_communication) :
     remote_communication_(std::move(remote_communication)) {}
 
-void RemoteProtocol::write_to_device(const void* mem_ptr, tt_xy_pair core, uint64_t addr, size_t size) {
+void RemoteProtocol::write_to_device(const void* mem_ptr, tt_xy_pair core, uint64_t addr, size_t size, NocId noc_id) {
```
pjanevskiTT, line 19: The new `noc_id` parameter is silently dropped — `RemoteCommunication::write_to_non_mmio` / `read_non_mmio` don't accept a `NocId`, so a caller selecting NOC1 still gets default-NOC behaviour on the remote path with no warning. Either plumb `noc_id` through `RemoteCommunication`, or at minimum mark the parameter `[[maybe_unused]]` and add a comment explaining the limitation so this isn't surprising. (Same applies to `read_from_device` on line 23.)

device/tt_device/protocol/remote_protocol.cpp
```diff
@@ -16,15 +16,17 @@ namespace tt::umd {
...
     remote_communication_->write_to_non_mmio(core, mem_ptr, addr, size);
```
pjanevskiTT, line 20: `RemoteProtocol::write_to_device`, `read_from_device`, and `write_to_core_range` accept `NocId noc_id` but silently drop it — `RemoteCommunication::write_to_non_mmio` / `read_non_mmio` are not given the value. With the rest of the stack now plumbing the NOC id through, callers passing `NocId::NOC1` to a remote chip will still execute on whatever NOC the underlying `RemoteCommunication` picks (effectively NOC0). Please either (a) propagate the `noc_id` into `RemoteCommunication`, or (b) `UMD_THROW` (or at minimum log) when `noc_id != NOC0` on the remote path so users aren't silently downgraded.

device/tt_device/hang_detection/blackhole_hang_detector.cpp
```diff
@@ -30,7 +30,7 @@ uint32_t BlackholeHangDetector::read_hang_check_reg_via_noc(NocId noc) {
     uint64_t addr = get_arch_impl()->get_noc_reg_base(CoreType::PCIE, static_cast<uint32_t>(noc)) +
                     get_arch_impl()->get_noc_node_id_offset();
     uint32_t value = 0;
-    get_protocol()->read_from_device(&value, core, addr, sizeof(value));
+    get_protocol()->read_from_device(&value, core, addr, sizeof(value), noc);
```
pjanevskiTT, line 33: Subtle behavior change: previously the address was computed for the queried `noc` parameter but the read transport used the thread-local selected NOC (often NOC0); now both address and transport use `noc`. If the hang check is invoked while NOC `noc` is itself hung, the read can now hang where before it would have succeeded over the healthy NOC. Worth confirming this doesn't regress hang-detector reliability when the queried NOC is the broken one.

device/tt_device/protocol/remote_protocol.cpp
nbuncicTT: Correct, but, this is a bigger problem:
- `RemoteCommunication` depends on `TTDevice` which is currently stateful - this means that the parameter is ignored, but in all of the codebase, DeviceProtocol and it's derived classes are only used in TTDevice and they get the same NOC.

So for this PR - it's ok because TTDevice is stateful for now.

Action items:
- Remove dependency on `TTDevice`, and plumb in `RemoteCommunication` API the parameter `NocId`.

device/tt_device/protocol/remote_protocol.cpp
nbuncicTT: Answered in another comment.

device/tt_device/hang_detection/blackhole_hang_detector.cpp
nbuncicTT: Essentially it's the same behavior. In is_noc_hung we thread local storage is changed to the queried NOC and DeviceProtocol obeyed the thread local storage value. Now this value is passed into the API and it doesn't rely on state, but rather on the passed parameter.

PR #2549, opened 2026-04-29
Use CoreCoord in TopologyDiscovery

aleksamarkovicTT:
### Issue
#2173 

### Description
Port TopologyDiscovery to use CoreCoord.

### List of the changes
- Changed all usages of xy_pair to CoreCoord.
- Removed manually translating coordinates.
- Removed `retrain_eth_core()` from `WormholeTTDevice`.

### Testing
CI

### API Changes


device/topology/topology_discovery_wormhole.cpp
```diff
@@ -263,11 +258,10 @@ void TopologyDiscoveryWormhole::retrain_eth_cores() {
         bool all_eth_cores_trained = true;
 
         for (const auto& [asic_id, tt_device] : devices_to_discover) {
-            auto* wormhole_tt_device = dynamic_cast<WormholeTTDevice*>(tt_device.get());
-
-            for (const CoreCoord& eth_core : tt_device->get_soc_descriptor().get_cores(
-                     CoreType::ETH, is_selected_noc1() ? CoordSystem::NOC1 : CoordSystem::NOC0)) {
-                EthTrainingStatus status = tt_device->read_eth_core_training_status(eth_core);
+            for (const CoreCoord& eth_core : tt_device->get_soc_descriptor().get_cores(CoreType::ETH)) {
```
pjanevskiTT, line 260: Pre-PR, this loop iterated over `get_cores(CoreType::ETH, is_selected_noc1() ? CoordSystem::NOC1 : CoordSystem::NOC0)`, branching on the selected NOC. Now it implicitly uses the `NOC0` default. Was the NOC1 branch intentionally dropped here? If users running with NOC1 selected ever hit `retrain_eth_cores()`, this looks like a behaviour change worth confirming or calling out in the PR description.

aleksamarkovicTT: > Now it implicitly uses the NOC0 default.


The NOC switch was moved to `translate_chip_coord_to_translated()` in the line below, similarly to what was done in #2331

device/tt_device/wormhole_tt_device.cpp
```diff
@@ -296,11 +296,6 @@ EthTrainingStatus WormholeTTDevice::read_eth_core_training_status(tt_xy_pair eth
     return static_cast<EthTrainingStatus>(training_status);
 }
 
-void WormholeTTDevice::retrain_eth_core(tt_xy_pair eth_core) {
```
pjanevskiTT, line 298: why move this out of TTDevice? TTDevice should provide a mechanism on how to do it, and then other clients in UMD can use it

aleksamarkovicTT: I don't think this should be in TTDevice, especially since there's no guards for calling it, meanwhile it's only for eth cores, only on 6U, only for some ETH FW versions, it's a private method by all counts. That's why I removed it and will probably make it public in EthFirmwareProvider.

device/topology/topology_discovery_wormhole.cpp
```diff
@@ -286,7 +280,8 @@ void TopologyDiscoveryWormhole::retrain_eth_cores() {
 
                 log_debug(
                     LogUMD, "Retraining ETH core {} on device {}, attempt {}.", eth_core.str(), asic_id, attempt + 1);
-                wormhole_tt_device->retrain_eth_core(eth_core);
+                uint32_t trigger_val = wormhole::ETH_TRIGGER_RETRAIN_VAL;
```
pjanevskiTT, line 282: Behavior change worth flagging: previously `retrain_eth_cores` iterated ETH cores in NOC0 or NOC1 (depending on `is_selected_noc1()`) and called `wormhole_tt_device->retrain_eth_core(eth_core)`, which forwarded the coord straight to the `tt_xy_pair` `write_to_device` overload (no translation). After this change, the loop iterates default-coordinate `CoreCoord` and the `CoreCoord` `write_to_device` overload translates via `translate_chip_coord_to_translated`. For unharvested/older paths NOC0 and TRANSLATED can differ on Wormhole (NOC translation table), so the retrain trigger is now written to a different physical NOC target than before. If that's intended (the new path is more correct), please confirm in the PR description; otherwise this could regress retrain on configurations where translated and NOC0 differ.

device/tt_device/wormhole_tt_device.cpp
aleksamarkovicTT: Also, a dynamic_cast is required to call it in the only place it is called, another sign it is out of place.

pjanevskiTT: can we just create an issue to track this work? Or if we already have one issue can you add this item to it?

pjanevskiTT (approved):

device/tt_device/wormhole_tt_device.cpp
broskoTT: Lets not go back and forth, I added it to TTDevice and now it's being removed. 
Can we have an issue to track this, and have a proposal there on where should this kind of function live longer term?

device/topology/topology_discovery_wormhole.cpp
```diff
@@ -263,11 +258,10 @@ void TopologyDiscoveryWormhole::retrain_eth_cores() {
...
         for (const auto& [asic_id, tt_device] : devices_to_discover) {
-            auto* wormhole_tt_device = dynamic_cast<WormholeTTDevice*>(tt_device.get());
-
-            for (const CoreCoord& eth_core : tt_device->get_soc_descriptor().get_cores(
-                     CoreType::ETH, is_selected_noc1() ? CoordSystem::NOC1 : CoordSystem::NOC0)) {
-                EthTrainingStatus status = tt_device->read_eth_core_training_status(eth_core);
+            for (const CoreCoord& eth_core : tt_device->get_soc_descriptor().get_cores(CoreType::ETH)) {
+                xy_pair translated_eth_core =
+                    tt_device->get_soc_descriptor().translate_chip_coord_to_translated(eth_core);
+                EthTrainingStatus status = tt_device->read_eth_core_training_status(translated_eth_core);
```
broskoTT, line 263: So you still use the version in ttdevice which accepts xy_pair?
Hmm I see you've added CoreCoord version for some of the tt_device api, but not all?

broskoTT (approved):

device/topology/topology_discovery_wormhole.cpp
aleksamarkovicTT: I didn't yet, because I don't think this method belongs in TTDevice in its current form, so I left it out in the PR where I added the CoreCoords, but I will clear this up in another PR.

device/tt_device/wormhole_tt_device.cpp
aleksamarkovicTT: https://github.com/tenstorrent/tt-umd/issues/2575

device/topology/topology_discovery_wormhole.cpp
aleksamarkovicTT: Yes, this should be translated_eth_core.

aleksamarkovicTT: Actually, it goes through the same function so the effect is the same, it's just not clear whether it's xy_pair or CoreCoord.

Issue #1303, opened 2025-09-13
Remove old tensix reset API


broskoTT:
Once the new one is completely in.

broskoTT:
Finally closed by https://github.com/tenstorrent/tt-umd/pull/2553

PR #2552, opened 2026-04-29
Remove usage of TensixSoftReset

broskoTT:
### Issue
#1303 
Part of effort #2521 

### Description
This enum was being purged, and this was the only place left. Now we have much better test coverage than when this effort was started.

### List of the changes
- Remove TensixSoftResetOptions from broadcast_tensix_risc_reset_to_cluster
- Expose noc_multicast_write in Chip
- deassert_resets_and_set_power_state calls assert_rist_reset()

### Testing
Existing CI should cover this flow

### API Changes
There are no API changes in this PR, changes are internal

### DIFF
--- a/device/chip/chip.cpp
+++ b/device/chip/chip.cpp
@@ -299,4 +299,8 @@ void Chip::noc_multicast_write(void* dst, size_t size, CoreCoord core_start, Cor
         addr);
 }
 
+void Chip::noc_multicast_write(void* dst, size_t size, uint64_t addr) {
+    get_tt_device()->noc_multicast_write(dst, size, addr);
+}
+
 }  // namespace tt::umd
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -1037,13 +1037,6 @@ void Cluster::broadcast_tensix_risc_reset_to_cluster(uint32_t reg_value) {
         // Nowhere to broadcast to.
         return;
     }
-    // If ethernet broadcast is not supported, do it one by one.
-    if (!use_ethernet_broadcast) {
-        for (auto& chip_id : all_chip_ids_) {
-            get_chip(chip_id)->noc_multicast_write(&reg_value, sizeof(uint32_t), 0xFFB121B0);
-        }
-        return;
-    }
 
     std::set<ChipId> chips_to_exclude = {};
     std::set<uint32_t> rows_to_exclude;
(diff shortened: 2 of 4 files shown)


device/chip/chip.cpp
```diff
@@ -272,4 +272,8 @@ void Chip::noc_multicast_write(void* dst, size_t size, CoreCoord core_start, Cor
         addr);
 }
 
+void Chip::noc_multicast_write(void* dst, size_t size, uint64_t addr) {
```
pjanevskiTT, line 302: should this be a part of the separate PR?

pjanevskiTT (approved):
I would just move exposing noc_multicast_write to separate PR, otherwise seems fine

device/cluster.cpp
```diff
@@ -428,9 +428,19 @@ std::set<ChipId> Cluster::get_target_mmio_device_ids() { return local_chip_ids_;
...
+    // Raise reset signal for all cores.
+    uint32_t reset_reg_value =
+        architecture_implementation::create(arch_name)->get_soft_reset_reg_value(RiscType::ALL_TENSIX);
+    broadcast_tensix_risc_reset_to_cluster(reset_reg_value);
+}
 
-void Cluster::deassert_risc_reset() { broadcast_tensix_risc_reset_to_cluster(TENSIX_DEASSERT_SOFT_RESET); }
+void Cluster::deassert_risc_reset() {
+    // Lower the reset signal for BRISC only.
+    uint32_t reset_reg_value = architecture_implementation::create(arch_name)->get_soft_reset_reg_value(
```
pjanevskiTT: The previous `TENSIX_DEASSERT_SOFT_RESET` mask included `STAGGERED_START` (bit 31), but `get_soft_reset_reg_value(RiscType::ALL_TENSIX & ~RiscType::BRISC)` only sets BRISC/TRISC/NCRISC bits — `STAGGERED_START` is added separately via `get_soft_reset_staggered_start()`. This silently drops the staggered-start bit from the broadcast deassert path, which is a behavior change vs. the prior `TENSIX_DEASSERT_SOFT_RESET` value.

pjanevskiTT: do we even need staggered start?

device/chip/chip.cpp
broskoTT: to minor to move to a separate PR

device/cluster.cpp
broskoTT: added staggered start by default. I think it was important at some point, would rather leave it be

Issue #814, opened 2025-04-29
Verify our TRACE logging capabilities


broskoTT:
We should have a way to log everything that happens through UMD. 
Especially useful for simulators.
We should be covered here mostly, but another pass is needed to verify this.

PR #2556, opened 2026-04-30
Add log_info on main metal used init functions

broskoTT:
### Issue
No specific issue.
Vaguely related to https://github.com/tenstorrent/tt-umd/issues/814

### Description
We got an inflow of bug requests where people think UMD's init is stuck, so this should clear up the confusion in cases when  that is not the case.

### List of the changes
Added logs at the beginning and end of main functions called from metal:
- Cluster::Cluster
- Cluster::~Cluster
- Cluster::start_device
- Cluster::close_device
- TopologyDiscovery::discover

### Testing
Checked in the logs of our tests that these logs are visible and not spamming to much

### API Changes


device/topology/topology_discovery.cpp
```diff
@@ -99,6 +99,7 @@ std::unique_ptr<ClusterDescriptor> TopologyDiscovery::create_ethernet_map() {
 std::pair<std::unique_ptr<ClusterDescriptor>, std::map<ChipId, std::unique_ptr<TTDevice>>> TopologyDiscovery::discover(
     const TopologyDiscoveryOptions& options, IODeviceType io_device_type, const std::string& soc_descriptor_path) {
     ZoneScopedC(tracy::Color::DarkGreen);
+    log_info(LogUMD, "Topology discovery started.");
```
aleksamarkovicTT: This already exists, you should remove it here and change the existing ones from debug to info.

device/cluster.cpp
```diff
@@ -1037,6 +1040,7 @@ void Cluster::start_device(const DeviceParams& device_params) {
 
         deassert_resets_and_set_power_state();
     }
+    log_info(LogUMD, "Done starting devices in cluster.");
```
aleksamarkovicTT: Use "completed" instead of colloquial "done" in all of the logs.

pjanevskiTT (approved):

aleksamarkovicTT (approved):

PR #2557, opened 2026-04-30
TTSimTTDevice: localize the brisc-CRT clock pre-pay to the deassert paths

pavlejosipovic:
## Summary

PR #2515 added `advance_clock(max(1000, size))` to `TTSimTTDevice::write_to_device` so that a freshly-deasserted RISC has time to clear CRT init before the next host op lands. That fixed the LLK ttsim hang, but the floor fires on **every** host write — including register pokes that have nothing to do with brisc — and amplifies them ~228× in simulated cycles.

On the galaxy `unit_tests_api` runs (`*StreamRegWrite*` / `*InlineWrite*`, 32 chips × many small writes) this turned a ~16 min CI pipeline into ~56 min ([example run](https://github.com/tenstorrent/tt-umd/actions/runs/24886375506/job/72867340367)).

### Where the cost actually was

Local repro (1000 4-byte writes through `write_to_device`, libttsim_bh.so v1.5.4):

| Floor | Per write (wall) | Total |
|---|---|---|
| `max(1000, size)` (current) | ~430 µs | 0.43 s |
| `max(1, size)` (this PR) | ~1.9 µs | 0.0019 s |

Galaxy unit tests do thousands of small register writes per chip × 32 chips × multiple test cases — that's where the regression compounds.

### Fix

Move the 1000-cycle pre-pay to the only writes that actually need it:

- **`write_to_device`**: `advance(max(1, size))` — same shape as `read_from_device`'s `advance(1)`. Enough to keep back-to-back host ops from piling up at the same simulated timestamp without paying 1000 cycles per register poke.
- **`send_tensix_risc_reset`**: `advance_clock(1000)` after the reset write. This is the path tt-metal slow dispatch goes through (`Cluster::deassert_risc_reset_at_core` → `driver_->deassert_risc_reset`), so the brisc-CRT pre-pay is preserved end-to-end.
- **`deassert_risc_reset`**: same, for callers that use the granular per-RISC API.

Also adds a `TTDevice::advance_simulated_clock(cycles)` Python binding for callers that bypass umd's reset API entirely (e.g. the LLK test harness writes `RISCV_DEBUG_REG_SOFT_RESET_0` directly via register_store). No-op on silicon.

### Companion PRs

Both required; this PR alone doesn't make the LLK harness happy because that path bypasses umd's reset API.

- tt-exalens PR https://github.com/tenstorrent/tt-exalens/pull/1002 — thin `advance_simulated_clock` wrapper.
- tt-metal LLK harness PR https://github.com/tenstorrent/tt-metal/pull/43412 — calls `advance_simulated_clock(1000)` in `commit_tensix_soft_reset`'s deassert path.

## Test plan

- [x] Standalone repro: 1000 × 4-byte `write_to_device` on libttsim_bh.so v1.5.4 → 228× faster (0.43 s → 0.0019 s)
- [x] LLK Float16_b 6-test smoke (`test_eltwise_unary_datacopy` on BH ttsim) with companion tt-exalens + tt-metal changes: 6/6 pass
- [ ] CI: galaxy `unit_tests_api` `*StreamRegWrite*` / `*InlineWrite*` step recovers from ~30 min back toward ~1-2 min on BH (and similar on WH)
- [ ] tt-metal slow dispatch programming examples (single-chip WH/BH `metal_example_add_2_integers_in_riscv`) still pass — expected since they go through `deassert_risc_reset_at_core` which retains the pre-pay


pjanevskiTT:
closing in favor of #2562

PR #2562, opened 2026-04-30
TTSim: drop sim-clock advance from TTSimTTDevice::write_to_device

### LINKED ISSUE #2561: TTSim: don't advance the sim clock from TTSimTTDevice::write_to_device — it's slow
`tt-metal/tt_metal/tt-llk/tests/TTSIM.md` references a workaround in (some
revision of) `TTSimTTDevice::write_to_device` that advances the sim clock on
every host write, in order to mask a startup race in the LLK test infra.

We've fixed that race at the protocol layer in tt-metal — boot-ready sentinel
handshake on the BRISC counter mailbox, see
https://github.com/tenstorrent/tt-metal/tree/pjanevski/llk_infra_fix
(commit https://github.com/tenstorrent/tt-metal/commit/9d9a976d9ee2f819407dff221883f5e08f124d73).
With that in place, the umd-side workaround is no longer needed.

Please revert it (or don't land it, if it's still in flight): test runs do many
more writes than reads (ELF loads, stimuli, runtime args), so ticking the sim
on every write is a meaningful slowdown for no benefit now that the underlying
race is gone. Reads can keep ticking the sim for spin-loop purposes.

Companion tt-metal issue: tenstorrent/tt-metal#43438

pjanevskiTT:
### Issue
Closes #2561

### Description
The clock-advance in `TTSimTTDevice::write_to_device` (`advance_clock(max(1000, size))`) was a workaround for an LLK-test infra startup race where BRISC's CRT init clobbered the host's first command mailbox write. That race has been fixed at the protocol layer in tt-metal via a boot-ready sentinel handshake on the BRISC counter mailbox (tenstorrent/tt-metal#43438, branch `pjanevski/llk_infra_fix`, commit tenstorrent/tt-metal@9d9a976), so the umd-side workaround is no longer needed.

Test runs do many more writes than reads (ELF loads, stimuli, runtime args), so ticking the sim on every write was a meaningful slowdown. `read_from_device` keeps its `advance_clock(1)` for spin-loop purposes.

### List of the changes
- Remove `communicator_->advance_clock(std::max<uint32_t>(1000, size))` from `TTSimTTDevice::write_to_device` (and the explanatory comment that went with it).

### Testing
- Functional validation runs as part of the tt-metal companion change (tenstorrent/tt-metal#43438).

### API Changes

### DIFF
--- a/device/tt_device/tt_sim_tt_device.cpp
+++ b/device/tt_device/tt_sim_tt_device.cpp
@@ -114,13 +114,6 @@ void TTSimTTDevice::write_to_device(const void* mem_ptr, tt_xy_pair core, uint64
     } else {
         communicator_->tile_write_bytes(core.x, core.y, addr, mem_ptr, size);
     }
-    // Advance the sim on host writes. Without this, sequences like "load BRISC ELF, deassert
-    // BRISC, write first command mailbox" all sit at the same simulated timestamp — BRISC never
-    // gets cycles to finish its CRT init before the command mailbox write, so BRISC's clear of
-    // brisc_command_buffer at startup clobbers the host's command and the handshake hangs. The
-    // floor guarantees that a tiny 4-byte deassert write still gives the newly-started core
-    // enough cycles to make meaningful progress before the next host operation lands.
-    communicator_->advance_clock(std::max<uint32_t>(1000, size));
 }
 
 void TTSimTTDevice::read_from_device(void* mem_ptr, tt_xy_pair core, uint64_t addr, size_t size) {


aleksamarkovicTT (approved):

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -114,13 +114,6 @@ void TTSimTTDevice::write_to_device(const void* mem_ptr, tt_xy_pair core, uint64
     } else {
         communicator_->tile_write_bytes(core.x, core.y, addr, mem_ptr, size);
     }
```
pjanevskiTT, line 116: (Note: ideally on the deleted `advance_clock` block at old lines 117-122 — placed on the closing `}` of the if/else immediately above the deletion, which is the last context line in the diff.) The removed block had a load-bearing comment about the BRISC ELF-load → deassert → mailbox-write sequence hanging because all three writes shared a simulated timestamp. Dropping the floor without an alternate source of clock progression risks reintroducing that hang. The PR description doesn't say what now guarantees forward progress between back-to-back host writes (the read path still has its own `advance_clock(1)`, but there's no equivalent here). Either document in the commit body what makes this safe (sim-side change, a previously-hanging test that now passes, etc.) or keep a smaller floor.

pjanevskiTT: it's fine, we shouldn't have comments about specific use cases in the code, clients should know how to handle this

PR #2563, opened 2026-04-30
Fix offset because of doppler for safe start of Baby RISCs

nbuncicTT:
### Issue
Tied to #2402 

### Description
This pull request updates the BRISC program loading and startup sequence in the test suite to support a new program layout in L1 memory. The main change is introducing a trampoline jump instruction at address 0x0 to redirect execution to the actual BRISC code at address 0x20, ensuring firmware data at 0x10 is skipped. This required updating test code, helper utilities, and the default BRISC configuration program.

### List of the changes
**Test harness and program loading updates:**

* All test cases and helper functions now write a trampoline jump instruction (`BRISC_TRAMPOLINE_JMP`) to address 0x0 before loading the BRISC program at address 0x20, ensuring correct program start and skipping firmware data at 0x10.
* The BRISC program load address in all relevant tests and helpers is updated from `0x0` to `0x20` to match the new layout.

**Assembly program and trampoline definition:**

* Added a new constant, `BRISC_TRAMPOLINE_JMP`, representing the trampoline jump instruction (`jal zero, 0x20`), to be written at address `0x0`.
* Updated the default BRISC configuration program to load the correct value for the new program layout (changed immediate value from `52` to `0x4C`).

### Testing
CI

### API Changes
/

### DIFF
--- a/tests/test_utils/assembly_programs_for_tests.hpp
+++ b/tests/test_utils/assembly_programs_for_tests.hpp
@@ -147,7 +147,11 @@ inline constexpr std::array<uint32_t, 14> bh_brisc_configuration_program{
 
 /*
 This program is architecture-agnostic and configures all RISC cores to execute an infinite loop
-at the same address (0x34 = 52 bytes from start of L1).
+at the same address (0x4C = 76 bytes from start of L1).
+
+The 0x4C value is intentional: with the 12-word program loaded at 0x20 (runtime-inserted
+architecture-specific `lui` + 11 instructions below), 0x4C is the address of the trailing
+`jal zero, 0` instruction, so all override PCs start in the self-loop.
 
 The first instruction (lui a5, <base>) sets the architecture-specific base address:
   - Wormhole (WH):  a5 = 0xFFEF'0000  (instruction: 0xffef07b7)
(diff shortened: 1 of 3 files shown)


aleksamarkovicTT (approved):

broskoTT (approved):

tests/test_utils/assembly_programs_for_tests.hpp
```diff
@@ -189,7 +189,7 @@ inline constexpr std::array<uint32_t, 11> brisc_configuration_program_default{
     0x00100713,  // li a4, 1
     0x28e7a623,  // sw a4, 652(a5)
     0x00078713,  // mv a4, a5
-    0x02c00793,  // li a5, 52
+    0x04c00793,  // li a5, 0x4C
```
pjanevskiTT, line 196: (Note: ideally on lines 176-179 of the pseudo-source/comment block above this array — that block is not in the diff window, so placing this comment on the changed `li a5, 0x4C` line.) The pseudo-source/comment block above `brisc_configuration_program_default` still documents `*trisc0_code_start_reg_addr = 0x34;` for all four PCs but the encoded `0x04c00793 // li a5, 0x4C` here now writes 0x4C. Update the C-style source comments to 0x4C so the comment matches what the override PC will be set to (currently the comment lies). The new PC value 0x4C must equal the offset of the trailing `jal zero, 0` instruction within the 12-word program loaded at 0x20, which is a non-obvious invariant worth documenting.

PR #2564, opened 2026-04-30
Fix GddrTelemetry test so it only focuses on API

nbuncicTT:
### Issue
/

### Description
This pull request makes several cleanups and simplifications to the `GddrTelemetry` test in `test_firmware_info_provider.cpp`. The main changes involve removing unused includes and redundant test assertions, and simplifying the validation of maximum GDDR temperature logic.

### List of the changes
* Removed unused headers `algorithm` and `unordered_map` from `test_firmware_info_provider.cpp` to reduce unnecessary dependencies.
* Simplified the GDDR temperature test by removing the manual maximum temperature calculation and instead only checking that `max_temp` is available. This reduces redundant logic and makes the test more robust to implementation changes.
* Removed redundant assertions that checked individual module temperature values against the aggregated data, further streamlining the test and focusing it on essential behavior.

### Testing
CI

### API Changes
/


aleksamarkovicTT (approved):

tests/api/test_firmware_info_provider.cpp
```diff
@@ -466,14 +464,7 @@ TEST_F(TestFirmwareInfoProvider, GddrTelemetry) {
...
 
-        double max_temp_from_modules = 0.0;
-        for (const auto& [gddr_index, module_telemetry] : gddr_telemetry->modules) {
-            max_temp_from_modules = std::max(max_temp_from_modules, module_telemetry.dram_temperature_top);
-            max_temp_from_modules = std::max(max_temp_from_modules, module_telemetry.dram_temperature_bottom);
-        }
-
-        EXPECT_DOUBLE_EQ(max_temp.value(), max_temp_from_modules)
-            << "Max temperature should match the maximum from all module temperatures.";
+        EXPECT_TRUE(max_temp.has_value()) << "Max GDDR temperature should be available on Blackhole.";
```
pjanevskiTT, line 469: Replacing the cross-check with `EXPECT_TRUE(max_temp.has_value())` weakens the test: nothing now verifies that the aggregator's value is consistent with the per-module values it's supposedly aggregating. If this was removed because aggregator and per-module reads can be sampled at different points in time and drift (which is the reasonable explanation), please leave a one-line comment in the test explaining that, otherwise the next reader will be tempted to put it back. If aggregation is actually expected to match exactly, the original assertion was the right one and the fix is in the implementation, not the test.

nbuncicTT: That was actually the idea - not to semantically test the APIs, but rather to test if the query is valid (not the returned value).

I'll add a comment.

Issue #1716, opened 2025-12-23
Move `SocDescriptor` to `TTDevice`


aleksamarkovicTT:
To enable using coord. manager from TTDevice and some other use cases.

pjanevskiTT:
is this fully fixed now? @aleksamarkovicTT

Issue #1163, opened 2025-08-07
Refactor constructors


broskoTT:
Our LocalChip constructors are a mess.
RemoteChip is also unmatched with it.

tt_SocDescriptor is also not great.
The issue captures refactoring of said constructors.
This was mostly captured while doing other changes on a path of #1118

pjanevskiTT:
are we more happy with the state of this today? @broskoTT @aleksamarkovicTT

broskoTT:
Aleksa continued cleaning up constructors, see linked PRs

broskoTT:
Refactor chip constructor also includes introducing a new exception UninitializedDevice

PR #2565, opened 2026-05-01
Clean up SocDescriptor constructors

aleksamarkovicTT:
### Issue
#1716 #1163

### Description
Clean up SocDescriptor constructors. SocDescriptor should only be constructable with a SocArchDescriptor and runtime info, the reason being to promote sharing SocArchDescriptors between individual SocDescriptors, something that will be implemented in a follow-up PR.

The PR also goes through each individual SocDescriptor construction and removes ones deemed unnecessary since the introduction of SocDescriptors in TTDevice, saving up memory.

### List of the changes
- Remove default. arch and YAML constructors of SocDescriptor. The only one remaining requires an existing SocArchDescriptor.
- Move private `format_node()` helper from SocDescriptor to SocArchDescriptor where it is used, remove include.
- Change SocArchDescriptor builder method `create` to be a constructor. This class does not need a builder method as it will not be inherited from.
- Remove unnecessary creation of duplicate SocDescriptors by using the one owned by TTDevice.

### Testing
CI

### API Changes
There are API changes in this PR:
- tt-metal: https://github.com/tenstorrent/tt-metal/pull/43472/changes (already merged)

### DIFF
--- a/device/api/umd/device/soc_descriptor.hpp
+++ b/device/api/umd/device/soc_descriptor.hpp
@@ -30,8 +30,6 @@ namespace tt::umd {
 class CoordinateManager;
 class SocArchDescriptor;
 
-tt_xy_pair format_node(const std::string& str);
-
 //! SocDescriptor contains information regarding the SOC configuration targetted.
 /*!
     Should only contain relevant configuration for SOC.
--- a/device/api/umd/device/soc_descriptor.hpp
+++ b/device/api/umd/device/soc_descriptor.hpp
@@ -41,8 +41,6 @@ public:
     // Default constructor. Creates uninitialized object with public access to all of its attributes.
     SocDescriptor() = default;
 
-    SocDescriptor(const tt::ARCH arch, const ChipInfo chip_info = {});
-
     // Constructor with explicit arch descriptor (enables sharing).
     SocDescriptor(std::shared_ptr<const SocArchDescriptor> arch_desc, const ChipInfo chip_info = {});
 
--- a/device/api/umd/device/soc_descriptor.hpp
+++ b/device/api/umd/device/soc_descriptor.hpp
@@ -38,10 +38,6 @@ tt_xy_pair format_node(const std::string& str);
 */
 class SocDescriptor {
 public:
-    // Default constructor. Creates uninitialized object with public access to all of its attributes.
-    SocDescriptor() = default;
-
-    // Constructor with explicit arch descriptor (enables sharing).
     SocDescriptor(std::shared_ptr<const SocArchDescriptor> arch_desc, const ChipInfo chip_info = {});
 
     // Helpers for extracting info from soc descriptor file.
--- a/device/api/umd/device/soc_descriptor.hpp
+++ b/device/api/umd/device/soc_descriptor.hpp
@@ -40,8 +40,6 @@ class SocDescriptor {
 public:
     // Default constructor. Creates uninitialized object with public access to all of its attributes.
     SocDescriptor() = default;
-    // Constructor used to build object from device descriptor file.
-    SocDescriptor(const std::string& device_descriptor_path, const ChipInfo chip_info = {});
 
     SocDescriptor(const tt::ARCH arch, const ChipInfo chip_info = {});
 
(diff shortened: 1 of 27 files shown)


tests/api/test_spi_tt_device.cpp
```diff
@@ -54,35 +54,14 @@ struct SpiTestDevices {
...
 SpiTestDevices setup_spi_test_devices() {
-    auto [cluster_desc, _] = TopologyDiscovery::discover({});
+    auto [cluster_desc, tt_devices] = TopologyDiscovery::discover({});
     SpiTestDevices result;
 
     for (ChipId chip_id : cluster_desc->get_chips_local_first(cluster_desc->get_all_chips())) {
         std::cout << "Setting up device " << chip_id << " local: " << cluster_desc->is_chip_mmio_capable(chip_id)
                   << std::endl;
 
-        if (cluster_desc->is_chip_mmio_capable(chip_id)) {
```
aleksamarkovicTT, line 64: Note to reviewers: Removed because there is no need to reconstruct devices, discover() returns them initialized.

tests/wormhole/test_remote_communication_wh.cpp
```diff
@@ -121,23 +121,10 @@ TEST(RemoteCommunicationWormhole, LargeTransferNoSysmem) {
         GTEST_SKIP() << "No remote chips found. Test requires at least one remote chip. Skipping test.";
     }
```
aleksamarkovicTT, line 123: Note to reviewers: like in SPI tests, no need to reconstruct devices.

device/api/umd/device/soc_descriptor.hpp
```diff
@@ -36,6 +36,7 @@ class SocArchDescriptor;
 */
 class SocDescriptor {
 public:
+    SocDescriptor() = default;  // TODO(#331): Delete when metal_SocDescriptor is deleted.
```
aleksamarkovicTT: Waiting on https://github.com/tenstorrent/tt-metal/pull/43472

device/chip/local_chip.cpp
```diff
@@ -48,15 +48,14 @@ std::unique_ptr<LocalChip> LocalChip::create(
     auto tt_device = TTDevice::create(physical_device_id, device_type);
     tt_device->init_tt_device();
 
-    SocDescriptor soc_descriptor;
+    std::shared_ptr<SocArchDescriptor> sad = nullptr;
```
pjanevskiTT: nit: rename the variable to something more descriptive

device/tt_device/rtl_simulation_tt_device.cpp
```diff
@@ -46,7 +46,8 @@ static constexpr ChipId DEFAULT_CHIP_ID = 0;
 std::unique_ptr<RtlSimulationTTDevice> RtlSimulationTTDevice::create(
     const std::filesystem::path& simulator_directory, int num_host_mem_channels) {
     auto soc_desc_path = SimulationChip::get_soc_descriptor_path_from_simulator_path(simulator_directory);
-    SocDescriptor soc_descriptor = SocDescriptor(soc_desc_path);
+    std::shared_ptr<SocArchDescriptor> sad = std::make_shared<SocArchDescriptor>(soc_desc_path);
```
pjanevskiTT: nit: same as above, apply it to all other places where you create SocArchDescriptor variable

device/chip/local_chip.cpp
pjanevskiTT: we probably don't even need the variable here, we can maybe create it directly in ctor call, remove the need to store it

tests/wormhole/test_remote_communication_wh.cpp
```diff
@@ -175,6 +162,4 @@ TEST(RemoteCommunicationWormhole, LargeTransferNoSysmem) {
         ASSERT_EQ(data_to_write[i], data_read[i]) << "Data mismatch at index " << i << ": expected 0x" << std::hex
                                                   << data_to_write[i] << " but got 0x" << data_read[i];
     }
-
-    local_tt_device->set_power_state(false);
```
pjanevskiTT, line 179: why are we removing set_power_state?

aleksamarkovicTT: Will revert.

aleksamarkovicTT: Actually, this was removed because there is no local_tt_device anymore, and I thought the call had no effect as the TTDevices are destructed right after. I don't think there is a need to return it to low power.

pjanevskiTT (approved):

PR #2566, opened 2026-05-01
Refactor Chip and TTDevice constructors

aleksamarkovicTT:
### Issue
#1163

### Description
Refactoring Chip and TTDevice constructors to make the code easier to understand.

### List of the changes
- Make TTDevice ctors protected, only create method is public.
- Remove unused TTDevice ctor taking architecture_implementation.
- Removed {Local,Remote}Chip create methods not taking an existing TTDevice.
- Enforced safe I/O api option for PCIe with exception.

### Testing
CI

### API Changes
This PR has API changes.


device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -469,8 +457,18 @@ class TTDevice {
     std::unique_ptr<ArcTelemetryReader> telemetry = nullptr;
     std::unique_ptr<FirmwareInfoProvider> firmware_info_provider = nullptr;
 
-    TTDevice();
-    TTDevice(std::unique_ptr<architecture_implementation> architecture_impl);
+    TTDevice() = default;
```
aleksamarkovicTT, line 470: Note: should delete.

aleksamarkovicTT: Can't be deleted as it's used by Simulator TTDevices. Will resolve in later PR.

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -477,8 +465,18 @@ class TTDevice {
...
+    TTDevice(
```
pjanevskiTT, line 471: this seems that we just moved ctors in the file? Is that because of some implicit ordering that we want to follow?

pjanevskiTT (approved):

device/api/umd/device/tt_device/tt_device.hpp
aleksamarkovicTT: They became protected instead of public.

PR #2568, opened 2026-05-03
Remove ethernet broadcast's dependancy on Cluster and Chip

broskoTT:
### Issue
(Link to Github issue(s))

### Description
In further preparation to extract EthernetBroadcast out of Cluster, remove some of the dependencies of broadcast_write_to_cluster and ethernet_broadcast_write functions.

### List of the changes
- Remove get_shelf_local_physical_chip_coords's only usage, and it's definition in cluster_descriptor
- Make ethernet broadcast use RemoteCommunication directly.
- That also means that LocalChip doesn't need RemoteCommunication anymore, so remove it from there, and keep it in Cluster

### Testing
Existing CI tests should cover this codepath

### API Changes
get_shelf_local_physical_chip_coords removed from ClusterDescriptor and ethernet_broadcast removed from LocalChip

### DIFF
--- a/device/chip/local_chip.cpp
+++ b/device/chip/local_chip.cpp
@@ -28,7 +28,6 @@
 #include "umd/device/pcie/pci_device.hpp"
 #include "umd/device/pcie/silicon_tlb_window.hpp"
 #include "umd/device/soc_descriptor.hpp"
-#include "umd/device/tt_device/remote_communication.hpp"
 #include "umd/device/tt_device/tt_device.hpp"
 #include "umd/device/types/arch.hpp"
 #include "umd/device/types/cluster_types.hpp"
@@ -75,7 +74,6 @@ std::unique_ptr<LocalChip> LocalChip::create(
     std::unique_ptr<TTDevice> tt_device, const SocDescriptor& soc_descriptor, int num_host_mem_channels) {
     std::unique_ptr<TLBManager> tlb_manager = nullptr;
     std::unique_ptr<SysmemManager> sysmem_manager = nullptr;
-    std::unique_ptr<RemoteCommunication> remote_communication = nullptr;
 
     // The variables below are only needed when using PCIe.
     // JTAG(currently the only communication protocol other than PCIe) has no use of them.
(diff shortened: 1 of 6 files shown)


device/cluster.cpp
```diff
@@ -621,6 +634,9 @@ void Cluster::wait_for_non_mmio_flush() {
     for (auto& [chip_id, chip] : chips_) {
         chip->wait_for_non_mmio_flush();
     }
+    for (auto& [chip_id, remote_comm] : remote_communications_) {
+        remote_comm->wait_for_non_mmio_flush();
```
pjanevskiTT, line 648: Null deref on Blackhole. `RemoteCommunication::create_remote_communication` returns `nullptr` for BH (`remote_communication.cpp:34`), and `remote_communications_[bh_chip_id]` is set to that `nullptr`. The previous `LocalChip::wait_for_non_mmio_flush` guarded with `if (remote_communication_)`; that guard is gone here, so this loop will SIGSEGV on any BH cluster that calls `Cluster::wait_for_non_mmio_flush()`.

device/chip/local_chip.cpp
```diff
@@ -402,39 +389,11 @@ void LocalChip::read_from_device_reg(CoreCoord core, void* dest, uint64_t reg_sr
...
-    // Depending on the device type, the implementation may vary.
-    // Currently JTAG doesn't support remote communication.
-    if (!remote_communication_) {
-        UMD_THROW(
-            error::RuntimeError,
-            fmt::format(
-                "Ethernet remote transfer is currently not supported for {} devices.",
-                DeviceTypeToString.at(tt_device_->get_communication_device_type())));
-    }
+void LocalChip::wait_for_non_mmio_flush() {}
```
pjanevskiTT, line 395: `LocalChip::wait_for_non_mmio_flush` is now a no-op, but `Cluster::wait_for_non_mmio_flush(chip_id)` (`cluster.cpp:615`) still routes through `get_chip(chip_id)->wait_for_non_mmio_flush()`. The single-chip variant therefore silently no longer flushes ethernet broadcast traffic. Either route the per-chip wait through `remote_communications_` too, or document this regression.

device/cluster.cpp
```diff
@@ -418,7 +427,9 @@ void Cluster::configure_active_ethernet_cores_for_mmio_device(
         }
     }
     // Local chips hold communication primitives for broadcasting, so we have to set this up for them as well.
-    get_local_chip(mmio_chip)->set_remote_transfer_ethernet_cores(active_eth_cores_per_chip);
+    remote_communications_.at(mmio_chip)->set_remote_transfer_ethernet_cores(
```
pjanevskiTT, line 439: Same null-deref risk: `remote_communications_.at(mmio_chip)->set_remote_transfer_ethernet_cores(...)` dereferences a `nullptr` when `mmio_chip` is BH.

device/chip/local_chip.cpp
broskoTT: Yeah this is very unimportant and not how it was used... if broadcast was used on cluster, the other variant for wait_non_mmio was called

device/cluster.cpp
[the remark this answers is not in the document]
broskoTT: fixed

device/cluster.cpp
```diff
@@ -418,7 +427,9 @@ void Cluster::configure_active_ethernet_cores_for_mmio_device(
...
+        get_local_chip(mmio_chip)->get_soc_descriptor().translate_coords_to_xy_pair(
+            active_eth_cores_per_chip, CoordSystem::TRANSLATED));
```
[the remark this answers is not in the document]
broskoTT: fixed

device/cluster.cpp
broskoTT: fixed

device/cluster.cpp
broskoTT: fixed

device/cluster.cpp
```diff
@@ -641,8 +661,8 @@ std::unordered_map<ChipId, std::vector<std::vector<int>>>& Cluster::get_ethernet
         for (const auto& chip : all_chip_ids_) {
             if (chips_to_exclude.find(chip) == chips_to_exclude.end()) {
                 // Get shelf local physical chip id included in broadcast.
-                ChipId physical_chip_id = cluster_desc->get_shelf_local_physical_chip_coords(chip);
                 EthCoord eth_coords = cluster_desc->get_chip_locations().at(chip);
+                ChipId physical_chip_id = 8 * eth_coords.x + eth_coords.y;
```
pjanevskiTT, line 667: why this logic? can you add a comment

pjanevskiTT (approved):

PR #2569, opened 2026-05-03
Create dedicated EthernetBroadcast

broskoTT:
### Issue
#2521 

### Description
Scope out EthernetBroadcast class out of Cluster class with clear interface

### List of the changes
- Introduce EthernetBroadcast
- Used in Cluster for broadcast_write_to_cluster, but the fallback is still there.

### Testing
Existing Broadcast tests are covering this codepath

### API Changes
This PR has API changes, it introduces a dedicated ethernet_broadcast class


device/cluster.cpp
```diff
@@ -527,7 +542,17 @@ void Cluster::refresh_cluster_description() {
...
+    if (ethernet_broadcast_) {
+        ChipId first_mmio_chip = *local_chip_ids_.begin();
+        std::unordered_map<ChipId, ChipId> chip_to_mmio_chip;
+        for (const ChipId chip : all_chip_ids_) {
+            const EthCoord& eth_coords = cluster_desc->get_chip_locations().at(chip);
+            chip_to_mmio_chip[chip] = (eth_coords.rack == 0 && eth_coords.shelf == 0)
+                                          ? cluster_desc->get_closest_mmio_capable_chip(chip)
+                                          : first_mmio_chip;
+        }
+        ethernet_broadcast_->clear_header_cache(cluster_desc->get_chip_locations(), chip_to_mmio_chip);
```
pjanevskiTT: `clear_header_cache` refreshes `chip_locations_` and `chip_to_mmio_chip_`, but `mmio_remote_comms_` is never updated on refresh. If `refresh_cluster_description` ever rebuilds `remote_communications_`, `EthernetBroadcast` keeps dangling `RemoteCommunication*` pointers and the next broadcast can use-after-free or route to the wrong chip. Either rebuild `mmio_remote_comms_` inside `clear_header_cache` too, or document/enforce that `remote_communications_` outlive `EthernetBroadcast` and don't change identity.

device/api/umd/device/tt_device/ethernet_broadcast.hpp
```diff
@@ -0,0 +1,59 @@
...
+    void broadcast_write_to_cluster(
+        const void* mem_ptr,
+        uint32_t size_in_bytes,
+        uint64_t address,
+        const std::set<ChipId>& chips_to_exclude,
+        std::set<uint32_t>& rows_to_exclude,
+        std::set<uint32_t>& columns_to_exclude,
+        bool use_translated_coords);
+
+    void clear_header_cache(
```
pjanevskiTT: `clear_header_cache` is misleading: it also overwrites `chip_locations_` and `chip_to_mmio_chip_`. Consider renaming (e.g. `reset_topology` / `refresh`) and have it accept all three pieces of state that the constructor takes so they can't drift out of sync.

device/tt_device/ethernet_broadcast.cpp
```diff
@@ -0,0 +1,281 @@
...
+            use_translated_coords ? wormhole::TRANSLATED_TO_VIRTUAL_Y.at(row - wormhole::translated_coordinate_start_y)
+                                  : row);
+    }
+    for (const auto& col : columns_to_exclude) {
+        cols_to_exclude_virtual.insert(
+            use_translated_coords ? wormhole::TRANSLATED_TO_VIRTUAL_X.at(col - wormhole::translated_coordinate_start_x)
+                                  : col);
+    }
+
+    auto arch_impl = architecture_implementation::create(tt::ARCH::WORMHOLE_B0);
```
pjanevskiTT, line 275: `architecture_implementation::create(tt::ARCH::WORMHOLE_B0)` is hardcoded inside `broadcast_write_to_cluster`, but nothing prevents a caller (PR #2571 exercises this) from constructing `EthernetBroadcast` for a non-WH chip. The WH-specific column/row constants and `TRANSLATED_TO_VIRTUAL_X/Y` tables used downstream would silently misroute writes on another arch. Take/check the arch at construction.

pjanevskiTT (approved):

PR #2571, opened 2026-05-03
Implement RemoteProtocol's write to core range

broskoTT:
### Issue
#2521 

### Description
Further prepare usage of EthernetBroadcast by adding it to remote protocol

### List of the changes
- RemoteProtocol now holds EthernetBroadcast instance
- Will return false if it is not available for broadcast
- implement a test for this path

### Testing
Wrote a dedicated test to test this functionality

### API Changes
This PR has API changes RemoteProtocol now implements write_to_core_range

### DIFF
--- a/device/tt_device/protocol/remote_protocol.cpp
+++ b/device/tt_device/protocol/remote_protocol.cpp
@@ -6,15 +6,22 @@
 
 #include "umd/device/tt_device/protocol/remote_protocol.hpp"
 
+#include <tt-logger/tt-logger.hpp>
 #include <utility>
 
+#include "umd/device/arch/wormhole_implementation.hpp"
+#include "umd/device/tt_device/ethernet_broadcast.hpp"
 #include "umd/device/tt_device/remote_communication.hpp"
 #include "umd/device/tt_device/tt_device.hpp"
 
 namespace tt::umd {
 
 RemoteProtocol::RemoteProtocol(std::unique_ptr<RemoteCommunication> remote_communication) :
-    remote_communication_(std::move(remote_communication)) {}
+    remote_communication_(std::move(remote_communication)) {
+    if (remote_communication_->has_sysmem_manager()) {
+        ethernet_broadcast_ = std::make_unique<EthernetBroadcast>(remote_communication_.get());
+    }
+}
 
 void RemoteProtocol::write_to_device(const void* mem_ptr, tt_xy_pair core, uint64_t addr, size_t size, NocId noc_id) {
     remote_communication_->write_to_non_mmio(core, mem_ptr, addr, size);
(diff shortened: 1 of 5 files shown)


device/tt_device/protocol/remote_protocol.cpp
```diff
@@ -6,15 +6,22 @@
...
+#include "umd/device/tt_device/ethernet_broadcast.hpp"
 #include "umd/device/tt_device/remote_communication.hpp"
 #include "umd/device/tt_device/tt_device.hpp"
 
 namespace tt::umd {
 
 RemoteProtocol::RemoteProtocol(std::unique_ptr<RemoteCommunication> remote_communication) :
-    remote_communication_(std::move(remote_communication)) {}
+    remote_communication_(std::move(remote_communication)) {
+    if (remote_communication_->has_sysmem_manager()) {
```
pjanevskiTT: The only gate before constructing `EthernetBroadcast` is `has_sysmem_manager()`; there is no architecture check. `EthernetBroadcast` hardcodes `WORMHOLE_B0` internally, so if `RemoteProtocol` is ever used for a non-WH arch you'll silently apply WH masks against the wrong NOC layout. Add an arch check, or pass arch into `EthernetBroadcast` and assert there.

device/tt_device/protocol/remote_protocol.cpp
```diff
@@ -24,7 +31,55 @@ void RemoteProtocol::read_from_device(void* mem_ptr, tt_xy_pair core, uint64_t a
...
 }
 
-bool RemoteProtocol::write_to_core_range(const void*, tt_xy_pair, tt_xy_pair, uint64_t, uint32_t) { return false; }
+bool RemoteProtocol::write_to_core_range(
+    const void* mem_ptr, tt_xy_pair core_start, tt_xy_pair core_end, uint64_t address, uint32_t size_in_bytes) {
+    if (ethernet_broadcast_ == nullptr) {
+        return false;
+    }
+
+    const bool use_translated_coords = core_start.x >= wormhole::tensix_translated_coordinate_start_x;
```
pjanevskiTT, line 49: `use_translated_coords = core_start.x >= wormhole::tensix_translated_coordinate_start_x` misclassifies ETH translated coordinates (`eth_translated_coordinate_start_x` is also 18 on WH) as tensix-translated. The translated branch then forces `rows_to_exclude={16,17}`, which are exactly the ETH rows, so an ETH multicast looks like it succeeds but writes nothing. Either reject non-tensix translated ranges explicitly or document the tensix-only contract on `noc_multicast_write`.

device/tt_device/protocol/remote_protocol.cpp
```diff
@@ -24,7 +31,55 @@ void RemoteProtocol::read_from_device(void* mem_ptr, tt_xy_pair core, uint64_t a
...
+
+    std::set<uint32_t> rows_to_exclude;
+    std::set<uint32_t> columns_to_exclude;
+
+    if (!use_translated_coords) {
+        // NOC0 coordinates: only the full tensix grid is supported.
+        if (core_start.x > 1 || core_start.y > 1 || core_end.x < 9 || core_end.y < 11) {
```
pjanevskiTT, line 57: Partial NOC0 ranges silently fall back to unicast (return false), while partial translated ranges are supported via row/col exclusion. The asymmetry isn't obvious to callers; either symmetrize NOC0 (using the same exclusion approach in `TENSIX_CORES_NOC0` space) or at least move the magic numbers (1, 9, 11) into named constants — the current values also bake in that NOC0 col=5 is the tensix gap, which isn't mentioned.

pjanevskiTT (approved):

device/tt_device/protocol/remote_protocol.cpp
broskoTT: noc_multicast_write is tensix-only already

device/tt_device/protocol/remote_protocol.cpp
```diff
@@ -24,7 +31,55 @@ void RemoteProtocol::read_from_device(void* mem_ptr, tt_xy_pair core, uint64_t a
...
+            log_debug(
+                LogUMD,
+                "broadcast_write_to_tensix: partial NOC0 ranges not supported for ethernet broadcast, falling back to "
+                "unicast");
```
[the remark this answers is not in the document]
broskoTT: it is fine to silently fail outer ring

tests/wormhole/test_cluster_wh.cpp
```diff
@@ -799,6 +799,96 @@ TEST(SiliconDriverWH, EthernetBroadcastSingleRemotePerChip) {
...
+
+                if (in_range) {
+                    ASSERT_EQ(vector_to_write, readback_vec)
+                        << "Core " << core.str() << " (translated " << translated.str()
+                        << ") is inside range and should have written data for size " << size;
+                } else {
+                    ASSERT_EQ(zeros, readback_vec) << "Core " << core.str() << " (translated " << translated.str()
+                                                   << ") is outside range and should still be zero for size " << size;
+                }
+            }
```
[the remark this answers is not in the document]
broskoTT: yeah let's not overengineer.
also some of those paths are covered by other broadcast tests

Issue #2520, opened 2026-04-26
close_device on Cluster destructor


broskoTT:
Make sure close_device is called on Cluster desctruction.

Also try to write a test which would fail. Maybe open a Cluster, call start_device, destroy it, then open TTDevice and check if tensixes are still running.

PR #2572, opened 2026-05-03
noc_multicast over EthernetBroadcast for Remote

broskoTT:
### Issue
Finalizing #2521 

### Description
As a crown jewel of the whole stack of changes, enable ethernet broadcast when calling multicast for remote chip.

### List of the changes
- integrate RemoteProtocol::write_to_core_ranges into TTDevice's noc_multicast_write
- Adjust the existing Multicast test to end up using EthernetBroadcast, by adding sysmem.

### Testing
The test adjusted in this PR should test the new codepath

### API Changes
There are no API changes in this PR


device/tt_device/tt_device.cpp
```diff
@@ -457,56 +457,71 @@ tt_xy_pair TTDevice::get_arc_core() const { return arc_core; }
...
+    bool broadcast_success = device_protocol_->write_to_core_range(src, core_start, core_end, addr, size);
+
+    log_info(
+        LogUMD,
+        "Remote multicast write to cores ({}, {}) - ({}, {}) {}",
+        core_start.x,
+        core_start.y,
+        core_end.x,
+        core_end.y,
+        broadcast_success ? "succeeded" : "fallbacked to unicast");
```
pjanevskiTT: Typo: "fallbacked to unicast" → "fell back to unicast" (or "falling back").

device/tt_device/tt_device.cpp
```diff
@@ -457,56 +457,71 @@ tt_xy_pair TTDevice::get_arc_core() const { return arc_core; }
...
-                    "noc_multicast_write fallback unicast to core at ({}, {}) is_tensix {}",
-                    x,
-                    y,
-                    is_active_tensix(x, y));
-                if (is_active_tensix(x, y)) {
-                    write_to_device(src, xy_pair(x, y), addr, size);
-                }
+    bool broadcast_success = device_protocol_->write_to_core_range(src, core_start, core_end, addr, size);
+
+    log_info(
```
pjanevskiTT: `log_info` per remote multicast will be very chatty in workloads that do many remote writes (every NOC multicast logs once). Consider `log_debug` so the line only fires when UMD logging is verbose.

device/tt_device/tt_device.cpp
```diff
@@ -457,56 +457,71 @@ tt_xy_pair TTDevice::get_arc_core() const { return arc_core; }
...
+
+    for (uint32_t x = core_start.x; x <= core_end.x; ++x) {
+        for (uint32_t y = core_start.y; y <= core_end.y; ++y) {
+            log_trace(
+                LogUMD,
+                "noc_multicast_write fallback unicast to core at ({}, {}) is_tensix {}",
+                x,
+                y,
+                is_active_tensix(x, y));
+            if (is_active_tensix(x, y)) {
```
pjanevskiTT, line 557: The fallback only writes to active TENSIX cores in `[core_start, core_end]` — DRAM/ETH/PCIe cores in that range are silently dropped. With `RemoteProtocol::write_to_core_range` currently always returning `false`, every remote multicast hits this path; a caller that multicasts to DRAM cores on a remote chip will get a silent partial write rather than an error. Worth either (a) widening the lambda to cover all valid core types, or (b) at minimum a `log_warning` when the broadcast range contains non-TENSIX cores.

tests/api/test_device_io.cpp
```diff
@@ -316,11 +316,14 @@ class TestMulticastWriteFixture : public ::testing::TestWithParam<std::tuple<boo
 // Bystander verification is only performed for isolated-core multicasts.
 TEST_P(TestMulticastWriteFixture, TestMulticastWrite) {
     // TODO: sysmem_enabled parameter to be added in the following PR.
```
pjanevskiTT: Stale TODO: `// TODO: sysmem_enabled parameter to be added in the following PR.` — `sysmem_enabled` is now in the param tuple, so this comment can go.

device/tt_device/tt_device.cpp
```diff
@@ -457,56 +457,71 @@ tt_xy_pair TTDevice::get_arc_core() const { return arc_core; }
...
+        core_start.x,
+        core_start.y,
+        core_end.x,
+        core_end.y,
+        broadcast_success ? "succeeded" : "fallbacked to unicast");
+
+    if (broadcast_success) {
+        get_remote_communication()->wait_for_non_mmio_flush();
+        return;
+    }
```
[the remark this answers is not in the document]
broskoTT: The broadcast path (remote_protocol.cpp) excludes non-TENSIX rows/cols but not harvested TENSIX rows. That actually matches the local hardware multicast (get_pcie_interface()->noc_multicast_write), which also doesn't filter harvested rows — writing to a disabled core is harmless.

device/tt_device/tt_device.cpp
```diff
@@ -457,56 +457,71 @@ tt_xy_pair TTDevice::get_arc_core() const { return arc_core; }
...
+
+    log_info(
+        LogUMD,
+        "Remote multicast write to cores ({}, {}) - ({}, {}) {}",
+        core_start.x,
+        core_start.y,
+        core_end.x,
+        core_end.y,
+        broadcast_success ? "succeeded" : "fallbacked to unicast");
+
```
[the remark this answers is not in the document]
broskoTT: log_debug

tests/api/test_device_io.cpp
```diff
@@ -316,11 +316,14 @@ class TestMulticastWriteFixture : public ::testing::TestWithParam<std::tuple<boo
...
-    auto [use_noc0, full_grid] = GetParam();
+    auto [use_noc0, full_grid, sysmem_enabled] = GetParam();
 
-    std::unique_ptr<Cluster> cluster = std::make_unique<Cluster>(ClusterOptions{.num_host_mem_ch_per_mmio_device = 0});
+    std::unique_ptr<Cluster> cluster =
+        std::make_unique<Cluster>(ClusterOptions{.num_host_mem_ch_per_mmio_device = sysmem_enabled ? 1 : 0});
```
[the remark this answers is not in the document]
broskoTT: outdated

tests/api/test_device_io.cpp
```diff
@@ -316,11 +316,14 @@ class TestMulticastWriteFixture : public ::testing::TestWithParam<std::tuple<boo
...
 
-    constexpr uint64_t address = 0x100;
+    test_utils::safe_test_cluster_start(cluster.get());
+
```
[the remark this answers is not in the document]
broskoTT: no other test calls close_device. THis is a deeper problem though https://github.com/tenstorrent/tt-umd/issues/2520

device/tt_device/tt_device.cpp
broskoTT: multicast is TENSIX-only by contract (the test asserts non-TENSIX cores must stay unmodified), and both paths honor it — broadcast excludes non-TENSIX, fallback writes only active TENSIX.

device/tt_device/tt_device.cpp
```diff
@@ -494,7 +495,80 @@ tt_xy_pair TTDevice::get_arc_core() const { return arc_core; }
 void TTDevice::noc_multicast_write(
     const void *src, size_t size, tt_xy_pair core_start, tt_xy_pair core_end, uint64_t addr) {
     ZoneScopedC(tracy::Color::Orange);
-    get_pcie_interface()->noc_multicast_write(src, size, core_start, core_end, addr, get_selected_noc_id());
+    if (!is_remote_tt_device) {
+        get_pcie_interface()->noc_multicast_write(src, size, core_start, core_end, addr, get_selected_noc_id());
```
pjanevskiTT, line 498: is this going to throw if this is not pci device? Should we check for that

device/tt_device/tt_device.cpp
```diff
@@ -494,7 +495,80 @@ tt_xy_pair TTDevice::get_arc_core() const { return arc_core; }
...
+        return;
+    }
+
+    UMD_ASSERT(
+        size <= std::numeric_limits<uint32_t>::max(),
```
pjanevskiTT: I think we should remove this check? Seems like this limitation doesn't come from the device. We could add all these checks to other function as well, right? I would remove it

pjanevskiTT (approved):

device/tt_device/tt_device.cpp
broskoTT: get_pcie_interface() will throw the right exception

PR #2573, opened 2026-05-03
write_to_core_range for PCIProtocol

broskoTT:
### Issue
Related to #2521 

### Description
Unifies multicast to just call write_to_core_range in case of both local and remote chips.

### List of the changes
- Implement write_to_core_ranges to just call noc_multicast
- Adjust TTDevice::noc_multicast_write given that now you can always try calling write_to_core_range

### Testing
There's already an exhaustive multicast write test.

### API Changes
This PR has API changes, implements write_to_core_ranges for pci protocol

### DIFF
--- a/device/tt_device/tt_device.cpp
+++ b/device/tt_device/tt_device.cpp
@@ -537,25 +537,24 @@ tt_xy_pair TTDevice::get_arc_core() const { return is_selected_noc1() ? arc_core
 void TTDevice::noc_multicast_write(
     const void *src, size_t size, tt_xy_pair core_start, tt_xy_pair core_end, uint64_t addr) {
     ZoneScopedC(tracy::Color::Orange);
-    if (!is_remote_tt_device) {
-        get_pcie_interface()->noc_multicast_write(src, size, core_start, core_end, addr, get_selected_noc_id());
-        return;
-    }
-
-    bool broadcast_success =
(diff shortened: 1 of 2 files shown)


pjanevskiTT (approved):

device/tt_device/tt_device.cpp
```diff
@@ -457,24 +457,24 @@ tt_xy_pair TTDevice::get_arc_core() const { return arc_core; }
 
 void TTDevice::noc_multicast_write(void *src, size_t size, tt_xy_pair core_start, tt_xy_pair core_end, uint64_t addr) {
     ZoneScopedC(tracy::Color::Orange);
-    if (!is_remote_tt_device) {
-        get_pcie_interface()->noc_multicast_write(src, size, core_start, core_end, addr);
-        return;
-    }
-
-    bool broadcast_success = device_protocol_->write_to_core_range(src, core_start, core_end, addr, size);
+    bool multicast_success = device_protocol_->write_to_core_range(src, core_start, core_end, addr, size_u32);
```
pjanevskiTT: `device_protocol_->write_to_core_range(src, core_start, core_end, addr, size_u32)` — `size_u32` is undefined in this scope; the original argument was the `size_t` parameter named `size`. This will not compile. Either define a local `size_u32` (with a narrowing-check from `size_t`) or keep passing `size`.

PR #2580, opened 2026-05-05
TTSim: enable 4GB TLB allocation through BAR4 on Blackhole

pjanevskiTT:
### Issue
On Blackhole, the 4GB TLBs are mapped through BAR4 (not BAR0 like the 2MB TLBs). `SimulationTlbManager::get_tlb_address_from_index()` had a `UMD_THROW("4GB TLBs in Blackhole are not currently supported in simulation TLB allocation.")` for any 4GB index, so any caller that allocated a 4GB TLB window would crash the moment `TTSimTlbHandle` tried to compute its base address. The bookkeeping (8 slots starting at index 202) was already in place — only the address computation and the BAR4 base were missing.

Pairs with [ttsim-private#350](https://github.com/tenstorrent/ttsim-private/pull/350), which adds the matching BAR4 decode on the simulator side. Both halves are needed end-to-end.

### Description
Plumb a `bar4_base` through `SimulationTlbManager`. `TTSimTTDevice` reads it from PCI config registers `0x20`/`0x24` (mirroring how `bar0_base` is read from `0x10`/`0x14`) and passes it in. `RtlSimulationTTDevice` passes `0` since RTL sim does not use this path.

`get_tlb_address_from_index()` for a 4GB index now returns `bar4_base + (tlb_index - tlb_4gb_start_index_) * 4GB`, matching the layout the simulator expects (each 4GB slot gets its own contiguous 4GB region in BAR4, in TLB-index order).

The simulator side ([ttsim-private#350](https://github.com/tenstorrent/ttsim-private/pull/350)) decodes BAR4 reads/writes by computing `tlb_index = 202 + (offset >> 32)` and looking up the matching TLB config slot. With both halves in place, allocating and using a 4GB TLB window on Blackhole now round-trips through BAR4 end-to-end.

### List of the changes
- `device/api/umd/device/chip_helpers/simulation_tlb_manager.hpp`: add `bar4_base` ctor parameter and `bar4_base_` member; update doxygen on `get_tlb_address_from_index` to reflect that 4GB TLBs return a BAR4 address.
- `device/chip_helpers/simulation_tlb_manager.cpp`: replace the `UMD_THROW` for 4GB indices with `bar4_base_ + slot * tlb_4gb_size_`.
- `device/api/umd/device/tt_device/tt_sim_tt_device.hpp`, `device/tt_device/tt_sim_tt_device.cpp`: add `bar4_base` field, read from PCI config space, pass to `SimulationTlbManager`.
- `device/tt_device/rtl_simulation_tt_device.cpp`: pass `/*bar4_base=*/0`.
- `nanobind/py_api_tt_device.cpp`: expose `bar4_base` alongside `bar0_base`.
- `tests/api/test_ttsim_device_io.cpp`: add `TTSimDeviceIOFixture.FourGBTlbBar4PathRoundTrip` — allocates a 4GB TLB window, writes via `write_block_reconfigure` (which goes out over BAR4), verifies via `tile_rd_bytes` (direct sim API) and a TLB read-back. Skipped on non-Blackhole.

### Testing
- Locally ran the full `TTSimDeviceIOFixture` against `sim_bh/libttsim.so` built from [ttsim-private#350](https://github.com/tenstorrent/ttsim-private/pull/350): all 13 tests pass, including the new one.
- Existing tests are unaffected; the only behaviour change for current callers is that a 4GB allocation no longer throws.
- `pre-commit run --all-files` passes.

### API Changes
- `SimulationTlbManager` constructor gains a `bar4_base` parameter (positional, between `bar0_base` and `arch_impl`). Both call sites in this repo are updated.
- `TTSimTTDevice::bar4_base` is a new public field, exposed to Python via nanobind alongside `bar0_base`.

### DIFF
--- a/device/api/umd/device/tt_device/tt_sim_tt_device.hpp
+++ b/device/api/umd/device/tt_device/tt_sim_tt_device.hpp
@@ -85,6 +85,7 @@ public:
     TLBManager *get_tlb_manager() override;
 
     uint64_t bar0_base = 0;
+    uint64_t bar4_base = 0;
 
 protected:
     void retrain_dram_core(const uint32_t dram_channel) override;
--- a/device/tt_device/tt_sim_tt_device.cpp
+++ b/device/tt_device/tt_sim_tt_device.cpp
@@ -105,13 +105,13 @@ TTSimTTDevice::TTSimTTDevice(
     tlb_manager_ = std::make_unique<SimulationTlbManager>(
         this,
         bar0_base,
-        bar4_base,
         architecture_impl_.get(),
         [comm = communicator_.get()](
             SimulationTlbManager* mgr, int id, size_t sz, TlbMapping map, tlb_data cfg) -> std::unique_ptr<TlbWindow> {
             auto handle = TTSimTlbHandle::create(mgr, comm, id, sz, map);
             return std::make_unique<TTSimTlbWindow>(std::move(handle), comm, cfg);
-        });
+        },
+        bar4_base);
     cached_tlb_window_ = tlb_manager_->allocate_default_tlb_window();
 }
 
--- a/device/tt_device/tt_sim_tt_device.cpp
+++ b/device/tt_device/tt_sim_tt_device.cpp
@@ -87,7 +87,10 @@ TTSimTTDevice::TTSimTTDevice(
         bar0_base |= uint64_t(communicator_->pci_config_read32(0, 0x14)) << 32;
         bar0_base &= ~15ull;  // ignore attributes, just obtain the physical address
 
-        // Compute physical address of BAR4 from PCI config registers (Blackhole 4GB TLBs live here).
+        // BAR4 is a 64-bit memory BAR; its base address is split across two PCI config
+        // registers — 0x20 holds the low 32 bits, 0x24 holds the high 32 bits. The low 4 bits
+        // of the low register encode BAR attributes (memory vs IO, prefetchable, 64-bit width)
+        // and are masked off to leave the physical address.
         bar4_base = communicator_->pci_config_read32(0, 0x20);
         bar4_base |= uint64_t(communicator_->pci_config_read32(0, 0x24)) << 32;
         bar4_base &= ~15ull;
--- a/device/tt_device/tt_sim_tt_device.cpp
+++ b/device/tt_device/tt_sim_tt_device.cpp
@@ -87,6 +87,11 @@ TTSimTTDevice::TTSimTTDevice(
         bar0_base |= uint64_t(communicator_->pci_config_read32(0, 0x14)) << 32;
         bar0_base &= ~15ull;  // ignore attributes, just obtain the physical address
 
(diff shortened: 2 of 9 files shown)


device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -86,6 +86,11 @@ TTSimTTDevice::TTSimTTDevice(
         bar0_base |= uint64_t(communicator_->pci_config_read32(0, 0x14)) << 32;
         bar0_base &= ~15ull;  // ignore attributes, just obtain the physical address
 
+        // Compute physical address of BAR4 from PCI config registers (Blackhole 4GB TLBs live here).
```
broskoTT: This comment explains the "what" which is clear from the code.
"why" is obvious.
Explain the "how" in comment instead.

device/api/umd/device/tt_device/tt_sim_tt_device.hpp
```diff
@@ -83,6 +83,7 @@ class TTSimTTDevice : public TTDevice {
     TLBManager *get_tlb_manager() override;
 
     uint64_t bar0_base = 0;
+    uint64_t bar4_base = 0;
```
broskoTT, line 88: Why is this part of public api?

tests/api/test_ttsim_device_io.cpp
```diff
@@ -382,4 +386,42 @@ TEST_F(TTSimDeviceIOFixture, RepeatedWriteReadCycles) {
...
+// rather than BAR0, exercising SimulationTlbManager::get_tlb_address_from_index's
+// BAR4 branch and the simulator's BAR4 TLB-translation path.
+TEST_F(TTSimDeviceIOFixture, FourGBTlbBar4PathRoundTrip) {
+    const SocDescriptor& soc = tt_device->get_soc_descriptor();
+    if (soc.arch != tt::ARCH::BLACKHOLE) {
+        GTEST_SKIP() << "4GB TLBs only exist on Blackhole; skipping for arch " << tt::arch_to_str(soc.arch);
+    }
+
+    auto tensix_cores = soc.get_cores(CoreType::TENSIX);
+    const tt_xy_pair core = soc.translate_coord_to(tensix_cores.at(0), CoordSystem::TRANSLATED);
```
broskoTT, line 404: Is there a path to talk to DRAM tiles through these TLBs or that is yet to come

broskoTT (approved):

device/api/umd/device/tt_device/tt_sim_tt_device.hpp
pjanevskiTT: needed for tlb window handling, will remove it once allocator - manager decoupling happens

tests/api/test_ttsim_device_io.cpp
pjanevskiTT: it works, I added the test

device/tt_device/tt_sim_tt_device.cpp
pjanevskiTT: added the comment

Issue #1933, opened 2026-01-28
Make adaptive polling utility


nbuncicTT:
Implement a utility that can be reused around the codebase in the following manner:
- Busy wait for some amount of time checking a specific condition
- After the busy wait is done, the thread goes to sleep (yielding the CPU) and then checks the condition after each sleep has timed out.

ajovanovicTT:
Closed with #2582

PR #2582, opened 2026-05-05
Add adaptive polling utility

ajovanovicTT:
### Issue
#1933

### Description
Add adaptive polling utility and apply it in blackhole arc core start wait

### List of the changes
 - Template function poll_until in utils.hpp
 - Refactor in blackhole wait_arc_core_start to use it

### Testing
 - CI

### API Changes

### DIFF
--- a/device/common/utils.hpp
+++ b/device/common/utils.hpp
@@ -138,6 +138,13 @@ inline bool poll_until(
     const std::chrono::milliseconds timeout,
     const std::chrono::microseconds busy_poll_window,
     const std::chrono::microseconds poll_interval) {
+    // Ensure the predicate is callable.
+    static_assert(std::is_invocable_v<Predicate>, "poll_until: The predicate provided must be callable.");
+
+    // Enforce strict bool return type.
+    using ReturnType = std::invoke_result_t<Predicate>;
+    static_assert(std::is_same_v<ReturnType, bool>, "poll_until: Predicate must return 'bool'.");
+
     const auto start = std::chrono::steady_clock::now();
     while (!predicate()) {
         const auto elapsed = std::chrono::steady_clock::now() - start;
--- a/device/common/utils.hpp
+++ b/device/common/utils.hpp
@@ -13,6 +13,7 @@
 #include <optional>
 #include <stdexcept>
 #include <string>
+#include <thread>
 #include <tt-logger/tt-logger.hpp>
 #include <unordered_set>
 #include <vector>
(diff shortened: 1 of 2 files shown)


device/tt_device/blackhole_tt_device.cpp
```diff
@@ -183,51 +182,24 @@ ChipInfo BlackholeTTDevice::get_chip_info() {
...
-        }
-
-        // If past 200us, avoid busy-waiting. Request a 10us sleep (minimum) -
-        // actual duration will be longer due to OS scheduling and jitter.
-        // This prevents 100% CPU usage during longer hardware initialization.
-        std::this_thread::sleep_for(std::chrono::microseconds(10));
+    constexpr auto busy_poll_window = std::chrono::microseconds(1000);
+    constexpr auto poll_interval = std::chrono::microseconds(10);
+
+    const bool started = utils::poll_until(
```
nbuncicTT: nit: Rename this to something more descriptive, `started` is vague. e.g. `arc_core_started` or `arc_core_ready`.

device/tt_device/blackhole_tt_device.cpp
```diff
@@ -183,51 +182,24 @@ ChipInfo BlackholeTTDevice::get_chip_info() {
...
+        [&]() {
```
nbuncicTT: nit: capturing everything can be a landmine in some situations as somebody can add/modify stuff in this lambda accidentally. To make this more verbose, add named references in the capture, like `[this, &arc_boot_status, &arc_postcode]()`

device/common/utils.hpp
```diff
@@ -120,6 +121,36 @@ inline bool check_timeout(
...
+ * @param poll_interval Sleep duration between checks once past the busy window.
+ * @returns True if predicate became true before timeout, false otherwise.
+ *          Caller is responsible for handling the timeout case.
+ */
+template <typename Predicate>
+inline bool poll_until(
+    Predicate predicate,
+    const std::chrono::milliseconds timeout,
+    const std::chrono::microseconds busy_poll_window,
+    const std::chrono::microseconds poll_interval) {
```
nbuncicTT, line 140: In the body of the function, add a check which requires the lambda to return a bool. 

Something like:
```cpp
    // Ensure the predicate is actually callable
    static_assert(std::is_invocable_v<Predicate>, 
        "poll_until: The predicate provided must be callable.");

    // Enforce strict bool return type.
    using ReturnType = std::invoke_result_t<Predicate>;
    static_assert(std::is_same_v<ReturnType, bool>, 
        "poll_until: Predicate must return 'bool'. ");
```

nbuncicTT (approved):
Well done.

PR #2585, opened 2026-05-06
Refactor Chip/TTDevice constructors (part 2)

### LINKED ISSUE #1716: Move `SocDescriptor` to `TTDevice`

aleksamarkovicTT:
### Issue
Step towards #1163 
Closes #1716

### Description
Simplified constructors for Chip and other improvements. Remove SocDescriptor from Chip.

### List of the changes
- Removed `SocDescriptor` in `Chip`. Mock and Emule Chips have their own private copy.
- Removed `SocDescriptor` from all `Chip` constructors taking `TTDevice`.
- Refactored `RemoteChip` constructors, removed double creation of `RemoteCommuncation`.
- Reuse of single `SocArchDescriptor` for all TTDevices in TopologyDiscovery.
- `init_tt_device()` takes `SocArchDescriptor` instead of `soc_descriptor_path`. (will be moved to ctor in future PR)
- Throwing on accessing `ChipInfo` of uninitialized `TTDevice`.
- Added `UninitializedDeviceError` and test. Will be extended in future PR.
- Python bindings for `SocArchDescriptor`.

### Testing
CI, tested custom SocDesc path.

### API Changes
This PR has API changes.


device/tt_device/tt_device.cpp
```diff
@@ -422,6 +423,9 @@ double TTDevice::get_asic_temperature() { return get_firmware_info_provider()->g
 uint8_t TTDevice::get_asic_location() { return get_firmware_info_provider()->get_asic_location(); }
 
 ChipInfo TTDevice::get_chip_info() {
+    if (firmware_info_provider == nullptr) {
```
pjanevskiTT, line 426: The new throw-on-uninitialized behavior is a declared API/behavioral change but isn't covered by a test. Consider adding a test that constructs a `TTDevice` and calls `get_chip_info()` before `init_tt_device()`, asserting it throws `error::RuntimeError`. Without one, this contract can silently regress.

device/tt_device/tt_device.cpp
```diff
@@ -565,11 +569,11 @@ void TTDevice::dma_h2d_zero_copy(uint32_t dst, const void *src, size_t size) {
 
 const SocDescriptor &TTDevice::get_soc_descriptor() const { return soc_descriptor_.value(); }
```
pjanevskiTT: `get_chip_info()` now throws `UMD_THROW(error::RuntimeError, ...)` on an uninitialized device, but `get_soc_descriptor()` here still throws a raw `std::bad_optional_access` via `.value()` when `soc_descriptor_` is empty. Since this PR routes all chip soc-descriptor access through `tt_device_->get_soc_descriptor()`, an uninitialized device now surfaces an inconsistent, non-UMD exception type. Suggest guarding with `if (!soc_descriptor_.has_value()) UMD_THROW(...)` so "uninitialized access" semantics are uniform across both accessors.

device/chip/remote_chip.cpp
```diff
@@ -28,30 +28,13 @@ namespace tt::umd {
 
 static_assert(!std::is_abstract<RemoteChip>(), "RemoteChip must be non-abstract.");
 
-std::unique_ptr<RemoteChip> RemoteChip::create(
-    LocalChip* local_chip,
-    EthCoord target_eth_coord,
-    const std::set<uint32_t>& remote_transfer_eth_channels,
-    SocDescriptor soc_descriptor) {
+std::unique_ptr<RemoteChip> RemoteChip::create(std::unique_ptr<TTDevice> remote_tt_device, LocalChip* local_chip) {
```
pjanevskiTT, line 31: `create` doesn't validate `remote_tt_device != nullptr` before the ctor dereferences it (`remote_tt_device->get_chip_info()` / `get_arch()`). Other chips (e.g. `SimulationChip`) assert non-null `tt_device`. A `UMD_ASSERT(remote_tt_device != nullptr, ...)` here would turn a potential null deref into a clear error.

device/api/umd/device/topology/topology_discovery_wormhole.hpp
```diff
@@ -5,12 +5,11 @@
 #pragma once
 
 #include <cstdint>
-#include <string>
+#include <utility>
 
 #include "umd/device/topology/topology_discovery.hpp"
 #include "umd/device/tt_device/tt_device.hpp"
 #include "umd/device/types/arch.hpp"
```
pjanevskiTT, line 12: This PR drops `#include "umd/device/types/xy_pair.hpp"` (previously just below this line). The header still uses `CoreCoord`/`EthCoord`, which now arrive only transitively via `tt_device.hpp` — it compiles, but it's an IWYU regression. Prefer keeping an explicit include for whatever provides `CoreCoord`/`EthCoord`. (Placed on the `arch.hpp` include since the removed line isn't on the RIGHT side of the diff.)

nanobind/py_api_soc_descriptor.cpp
```diff
@@ -4,11 +4,15 @@
 
 #include <nanobind/nanobind.h>
 #include <nanobind/stl/optional.h>
+#include <nanobind/stl/pair.h>
```
pjanevskiTT: `stl/pair.h` (here), `stl/tuple.h` (line 10), and `stl/unordered_map.h` (line 11) are added, but the map/tuple-returning getters (`get_cores`, `get_dram_core_channel_map`, `get_ethernet_core_channel_map`) aren't bound below — so these casters are currently unused. Either drop them, or wire up the map getters (a one-line comment on why the maps are/aren't exposed would help future readers).

device/api/umd/device/topology/topology_discovery_wormhole.hpp
aleksamarkovicTT: Moved include to `topology_discovery_wormhole.cpp`

nanobind/py_api_soc_descriptor.cpp
aleksamarkovicTT: removed

device/api/umd/device/chip/remote_chip.hpp
```diff
@@ -53,6 +48,8 @@ class RemoteChip : public Chip {
     SysmemManager* get_sysmem_manager() override;
     TLBManager* get_tlb_manager() override;
 
+    const SocDescriptor& get_soc_descriptor() const override { return tt_device_->get_soc_descriptor(); }
```
pjanevskiTT, line 51: this seems to be the same as for LocalChip? Should we make this default implementation inside chip.cpp and mock and emule (and maybe other chips as well...) should override it if needed

nanobind/py_api_soc_descriptor.cpp
```diff
@@ -76,6 +78,53 @@ void bind_soc_descriptor(nb::module_ &m) {
         .def("__eq__", &CoreCoord::operator==, release_gil())
         .def("__lt__", &CoreCoord::operator<, release_gil());
 
+    nb::class_<SocArchDescriptor>(m, "SocArchDescriptor")
```
pjanevskiTT, line 81: maybe this could be a separate PR, but not a huge concern

pjanevskiTT (approved):

device/api/umd/device/chip/remote_chip.hpp
aleksamarkovicTT: I agree but there is no TTDevice pointer in chip (yet).

nanobind/py_api_soc_descriptor.cpp
aleksamarkovicTT: I wanted to have something since I had to change init_tt_device() to take socarchdescriptor.

PR #2588, opened 2026-05-07
SimulationTlbManager: dummy 4GB TLB for Quasar

pjanevskiTT:
### Issue

Stacked on #2538. Refs Quasar TLB enablement (#450 / #2317).

### Description

`Cluster::configure_tlb` / `SimulationTlbManager::allocate_tlb_window` both fail on Quasar today: the allocator has no pools for it, so any allocation request returns -1 and we throw "No available TLB of requested size." The cached-window path in the TTDevice ctor only worked because it bypassed the allocator with a hard-coded factory call inside `allocate_default_tlb_window`.

Quasar's real TLB topology isn't finalized, and the simulator's communicator handles all I/O regardless of TLB programming, so the allocation just needs to succeed and return a non-zero TLB. This PR unifies the bypass: any Quasar allocation hands back a dummy 4GB window directly via the factory, with no index/address/size lookup against the allocator.

### List of the changes

- `SimulationTlbManager::allocate_tlb_window`: short-circuit on Quasar, call the factory directly with `tlb_id=0` and `size=4GB`. The allocator is untouched for Quasar.
- `SimulationTlbManager::allocate_default_tlb_window`: drop the bespoke Quasar factory call; route Quasar through the regular `allocate_tlb_window` path now that it handles Quasar.
- `TTSimTlbHandle::configure`: early-return on Quasar so we don't try to write a TLB config register at the (zeroed) `tlb_reg_addr_`.
- Update the comment in `SimulationTlbAllocator::initialize_architecture_config` to point at the new bypass site.

### Testing

- `baremetal_tests --gtest_filter='SimulationTlbAllocator.*'` — 13/13 pass.

### API Changes

None.

### DIFF
--- a/device/api/umd/device/chip_helpers/simulation_tlb_manager.hpp
+++ b/device/api/umd/device/chip_helpers/simulation_tlb_manager.hpp
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <atomic>
 #include <cstddef>
 #include <cstdint>
 #include <functional>
@@ -55,6 +56,11 @@ public:
 private:
     SimulationTlbAllocator allocator_;
     TlbWindowFactory factory_;
+
+    // Monotonic TLB id handed out for architectures that bypass the allocator
+    // (Quasar). Always increases — never reused — so TlbHandle::get_tlb_id()
+    // is unique across the lifetime of the manager.
+    std::atomic<int> next_bypass_tlb_id_{0};
 };
 
 }  // namespace tt::umd
--- a/device/chip_helpers/simulation_tlb_manager.cpp
+++ b/device/chip_helpers/simulation_tlb_manager.cpp
@@ -26,6 +26,15 @@ SimulationTlbManager::SimulationTlbManager(
 std::unique_ptr<TlbWindow> SimulationTlbManager::allocate_tlb_window(
     tlb_data config, const TlbMapping mapping, const size_t tlb_size) {
     ZoneScopedC(tracy::Color::Cyan);
+
+    // Quasar's TLB topology isn't finalized: bypass the allocator and hand back
+    // a dummy 4GB window. The simulator's communicator handles real I/O, so the
+    // window doesn't need a real index, address, or size.
+    if (allocator_.get_architecture() == tt::ARCH::QUASAR) {
+        static constexpr size_t SIZE_4GB = 4ULL * 1024 * 1024 * 1024;
+        return factory_(&allocator_, /*tlb_id=*/0, SIZE_4GB, mapping, config);
+    }
+
     int tlb_index = allocator_.allocate_tlb_index(tlb_size);
     if (tlb_index == -1) {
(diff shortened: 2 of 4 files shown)


broskoTT (approved):

device/api/umd/device/chip_helpers/simulation_tlb_manager.hpp
```diff
@@ -55,6 +56,11 @@ class SimulationTlbManager : public TLBManager {
 private:
     SimulationTlbAllocator allocator_;
     TlbWindowFactory factory_;
+
+    // Monotonic TLB id handed out for architectures that bypass the allocator
+    // (Quasar). Always increases — never reused — so TlbHandle::get_tlb_id()
+    // is unique across the lifetime of the manager.
+    std::atomic<int> next_bypass_tlb_id_{0};
```
pjanevskiTT, line 63: `std::atomic<int>` (signed) on `next_bypass_tlb_id_` makes overflow technically UB. Realistically you'd never hit 2^31 allocations, but `std::atomic<uint32_t>` (or `uint64_t`) makes the wraparound well-defined and matches the "always increases — never reused" intent stated in the comment.

device/chip_helpers/simulation_tlb_manager.cpp
```diff
@@ -39,19 +50,14 @@ std::unique_ptr<TlbWindow> SimulationTlbManager::allocate_tlb_window(
...
-    static constexpr size_t SIZE_4GB = 4ULL * 1024 * 1024 * 1024;
     const tt::ARCH architecture = allocator_.get_architecture();
     switch (architecture) {
         case tt::ARCH::BLACKHOLE:
             return allocate_tlb_window({}, TlbMapping::WC, SIZE_2MB);
         case tt::ARCH::WORMHOLE_B0:
             return allocate_tlb_window({}, TlbMapping::WC, SIZE_16MB);
         case tt::ARCH::QUASAR:
-            return factory_(&allocator_, 0, SIZE_4GB, TlbMapping::WC, {});
+            return allocate_tlb_window({}, TlbMapping::WC, /*tlb_size=*/0);
```
pjanevskiTT, line 60: Calling `allocate_tlb_window({}, TlbMapping::WC, /*tlb_size=*/0)` for Quasar is fragile: the only thing keeping this from dividing by zero inside `RtlSimTlbHandle::configure` (`new_config.local_offset / tlb_size_`) is the QUASAR bypass at line 35 forcing `SIZE_4GB`. If the bypass is ever moved/removed, the `tlb_size=0` here turns into a latent UB. Consider passing `SIZE_4GB` explicitly to make the intent local, or have `allocate_default_tlb_window` early-return for Quasar.

device/api/umd/device/chip_helpers/simulation_tlb_manager.hpp
pjanevskiTT: it's because TLB manager uses int, we will remove it and then we can remove this as well

PR #2591, opened 2026-05-07
Remove tlb manager from Sysmem manager

broskoTT:
### Description
This is a simple search and replace to remove SysmemManager's dependancy on TLBManager.

### List of the changes
- Pass TTDevice instead of TLBManager to SysmemManager and SysmemBuffer
- use pci_device member instead of tt_device->get_pci_device()
- use tt_device member instead of tlb_manager->get_tt_device() 

### Testing
No testing, code builds

### API Changes
This PR has API changes on sysmem manager

### DIFF
--- a/device/api/umd/device/chip_helpers/sysmem_buffer.hpp
+++ b/device/api/umd/device/chip_helpers/sysmem_buffer.hpp
@@ -9,12 +9,12 @@
 #include <memory>
 #include <optional>
 
-#include "umd/device/chip_helpers/tlb_manager.hpp"
 #include "umd/device/pcie/tlb_window.hpp"
 #include "umd/device/types/xy_pair.hpp"
 
 namespace tt::umd {
-class TLBManager;
+class PCIDevice;
+class TTDevice;
 
 /**
  * SysmemBuffer class should represent the resource of the HOST memory that is visible to the device.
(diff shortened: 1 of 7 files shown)


device/chip_helpers/silicon_sysmem_manager.cpp
```diff
@@ -129,14 +128,13 @@ void SiliconSysmemManager::unpin_or_unmap_sysmem() {
     } else {
         for (int ch = 0; ch < hugepage_mapping_per_channel.size(); ch++) {
             auto &HugepageMapping = hugepage_mapping_per_channel[ch];
-            if (HugepageMapping.physical_address &&
-                tt_device_->get_pci_device()->is_mapping_buffer_to_noc_supported()) {
+            if (HugepageMapping.physical_address && pci_device_->is_mapping_buffer_to_noc_supported()) {
```
pjanevskiTT: nit: it's not you that have introduced this, but maybe we should rename HugepageMapping var to hugepage_mapping since that's how we name our vars

device/api/umd/device/chip_helpers/sysmem_buffer.hpp
```diff
@@ -118,7 +119,8 @@ class SysmemBuffer {
 
     TlbWindow* get_cached_tlb_window();
 
-    TLBManager* tlb_manager_;
+    PCIDevice* pci_device_;
```
pjanevskiTT, line 122: nit: can these vars be const?

pjanevskiTT: also, do we introduce pci_device member just because we use it a lot?

pjanevskiTT (approved):

device/api/umd/device/chip_helpers/sysmem_buffer.hpp
broskoTT: Yes but also, in some of the following PRs we'll remove the tt_device_ member and not this one

broskoTT: >nit: can these vars be const?

Not really, they use non const functions like pci_device_->map_buffer_to_noc

PR #2595, opened 2026-05-08
Blacklist some bh-gh runs to speed up merge queue

broskoTT:
### Issue
Merge queue is slow

### Description
bh-gh-07 is an only machine. Reduce its load by running it only on one job in merge queue. Skip it for PR as well, as it is rare for this machine only to fail.

### List of the changes
- Introduced infra for blacklisting based on build-type
- Added entry for bhgh07 for everything except merge-queue Release.

### Testing
I'll have to look at all the jobs once this gets into merge queue, to verify.
Update: I've looked at the jobs on merge queue and can verify they're generated properly

### API Changes

### DIFF
--- a/.github/workflows/build-and-run-all-tests.yml
+++ b/.github/workflows/build-and-run-all-tests.yml
@@ -88,9 +88,10 @@ jobs:
           echo "entries=$(echo "$CARDS" | jq -c .)" >> $GITHUB_OUTPUT
 
   # Central blacklist for matrix entries that should not run under certain conditions.
-  # Each entry is a pipe-delimited "triggering-event|card" pair.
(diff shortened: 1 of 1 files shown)


pjanevskiTT (approved):

.github/workflows/build-and-run-all-tests.yml
```diff
@@ -88,9 +88,10 @@ jobs:
           echo "entries=$(echo "$CARDS" | jq -c .)" >> $GITHUB_OUTPUT
```
pjanevskiTT, line 88: can we just open an issue to track all of these workarounds for merge queue?

broskoTT:
skipping merge queue because:
- this change doesn't affect tests
- this diff will speed up other stuff in merge queue

PR #2607, opened 2026-05-11
Replace TLB window memcpy with 8/4-byte streaming stores (no AVX)

pjanevskiTT:
### Issue
closing https://github.com/tenstorrent/tt-umd/issues/1265
closing #68
closing #791

### Description

Alternative to #2357 / #2455 that does not use AVX2/SSE intrinsics. Same
single-write-per-address contract, but the aligned middle uses plain 8-byte
and 4-byte volatile stores instead of `_mm256_stream_si256` / `_mm_stream_si32`.

#2357 was reverted in #2430 due to issues observed with the AVX path, and the
re-land attempt in #2455 has been hitting similar problems. Switching the
aligned middle to scalar 8/4-byte stores keeps the no-double-write property —
the actual correctness requirement on device memory mapped through a TLB
window — while removing the AVX/non-temporal codepath entirely.

#### Background

glibc's `memcpy` can emit overlapping stores, which corrupts data on device
memory mapped through TLB windows because each store is a separate PCIe write
transaction. The streaming copy routines added here issue each destination
address exactly once.

### List of the changes

- Add `device_memcpy.hpp` / `device_memcpy.cpp` with `streaming_memcpy_to_device`
  and `streaming_memcpy_from_device`. Phases:
  - Byte-wide volatile stores/loads to align to 4 bytes
  - One 4-byte volatile store/load if needed to reach 8-byte alignment
  - 8-byte volatile stores/loads for the bulk
  - Optional 4-byte tail
  - Byte-wide tail
- Replace `memcpy` in `write_block`/`read_block` for non-Wormhole architectures
  with the streaming functions.
- WH-specific `memcpy_to_device` / `memcpy_from_device`: replace the
  word-by-word aligned-middle loop with the streaming functions.
- `SiliconTlbWindow::memcpy_from_device` signature updated to take
  `const volatile void*` (private static, internal only).
- Add `DebugDoubleWrite` test in `test_cluster_bh.cpp` that checks the NIU
  posted-write counter to verify each `write_to_device` produces exactly one
  PCIe write.

### Testing

CI. The new `DebugDoubleWrite` test directly validates the no-double-write
property on Blackhole.

### API Changes

No public API changes.

### DIFF
--- a/tests/blackhole/test_cluster_bh.cpp
+++ b/tests/blackhole/test_cluster_bh.cpp
@@ -510,6 +510,46 @@ TEST(SiliconDriverBH, DISABLED_VirtualCoordinateBroadcast) {  // same problem as
     cluster.close_device();
 }
 
+TEST(SiliconDriverBH, DebugDoubleWrite) {
+    // NOC register that counts the number of posted write requests received by a core.
+    // Streaming stores go through write-combining BAR and generate posted PCIe writes.
+    constexpr uint64_t NIU_SLV_POSTED_WR_REQ_RECEIVED = 0xffb202e0;
+
+    Cluster cluster;
+    set_barrier_params(cluster);
+    test_utils::safe_test_cluster_start(&cluster);
+
+    auto chip_id = *cluster.get_target_mmio_device_ids().begin();
+    auto tensix_core = cluster.get_soc_descriptor(chip_id).get_cores(CoreType::TENSIX).at(0);
+
+    uint64_t addr = 0;
+    uint32_t data = 1;
+    uint32_t counter = 0;
+
+    uint32_t base_reg_val;
+    cluster.read_from_device_reg(&base_reg_val, chip_id, tensix_core, NIU_SLV_POSTED_WR_REQ_RECEIVED, sizeof(uint32_t));
+
+    for (uint32_t i = 0; i < 100000; i++) {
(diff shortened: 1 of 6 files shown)


tests/blackhole/test_cluster_bh.cpp
```diff
@@ -510,6 +510,46 @@ TEST(SiliconDriverBH, DISABLED_VirtualCoordinateBroadcast) {  // same problem as
...
+    auto tensix_core = cluster.get_soc_descriptor(chip_id).get_cores(CoreType::TENSIX).at(0);
+
+    uint64_t addr = 0;
+    uint32_t data = 1;
+    uint32_t counter = 0;
+
+    uint32_t base_reg_val;
+    cluster.read_from_device_reg(&base_reg_val, chip_id, tensix_core, NIU_SLV_POSTED_WR_REQ_RECEIVED, sizeof(uint32_t));
+
+    for (uint32_t i = 0; i < 100000; i++) {
```
broskoTT, line 532: If we need stress tests like this in our repo, that is fine. But we should have a separate target for stress tests. They can run maybe just on Main or sth like that.
How much does this test run? And does it reproduce the issue in close to 100%

pjanevskiTT:
closed since #2455 was merged

PR #2606, opened 2026-05-11
Add get_tray_id method to ClusterDescriptor

tt-vjovanovic:
### Issue
(No linked issue)

### Description
Introduce the `get_tray_id` method in the `ClusterDescriptor` class to retrieve the tray ID for UBB boards based on the chip ID.

All tools should start using these exported data instead of hardcoding it across the projects.

### List of the changes
- Added `get_tray_id` method to `ClusterDescriptor`.
- Exposed `get_tray_id` in the Python API.

### Testing
Manual testing performed to ensure the new method returns correct tray IDs for UBB boards.

### API Changes
This PR has API changes:
- [ ] (If breaking change) tt_metal approved PR pointing to this branch: link
- [ ] (If breaking change) tt_debuda approved PR pointing to this branch: link

PR #2608, opened 2026-05-11
Expose additional ClusterDescriptor accessors in the Python API

pjanevskiTT:
### Issue
(No linked issue)

### Description
Split out of #2606 per review feedback. Exposes additional `ClusterDescriptor` accessors in the Python API so tooling can consume PCI / ASIC location data from the cluster descriptor instead of hardcoding it across projects.

### List of the changes
- Added out-of-line C++ definition for `ClusterDescriptor::get_chip_to_bus_id` (declaration was already in the header).
- Exposed `get_chip_pci_bdfs`, `get_chip_to_bus_id`, `get_bus_id` and `get_asic_location` in the Python API via nanobind.

### Testing
- Local `cmake --build build` clean.

### API Changes
This PR has API changes:

### DIFF
--- a/nanobind/py_api_topology_discovery.cpp
+++ b/nanobind/py_api_topology_discovery.cpp
@@ -90,6 +90,29 @@ void bind_topology_discovery(nb::module_& m) {
             nb::arg("chip_id"),
             release_gil(),
             "Returns the tray id (1..4) for UBB boards, or None for non-UBB / unknown.")
+        .def(
+            "get_chip_pci_bdfs",
+            &ClusterDescriptor::get_chip_pci_bdfs,
+            release_gil(),
+            "Map of ChipId -> PCI BDF string (e.g. \"0000:41:00.0\"). "
+            "Only contains entries for MMIO-capable chips.")
+        .def(
+            "get_chip_to_bus_id",
+            &ClusterDescriptor::get_chip_to_bus_id,
+            release_gil(),
+            "Map of ChipId -> PCI bus id (uint16, e.g. 0x41).")
+        .def(
+            "get_bus_id",
+            &ClusterDescriptor::get_bus_id,
+            nb::arg("chip_id"),
+            release_gil(),
+            "Returns the PCI bus id (uint16) for the given chip, or 0 if unknown.")
+        .def(
+            "get_asic_location",
+            &ClusterDescriptor::get_asic_location,
+            nb::arg("chip_id"),
+            release_gil(),
+            "Returns the ASIC location index within the chip's board (uint8), or 0 if unknown.")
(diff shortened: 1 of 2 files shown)


tt-vjovanovic (approved):

nanobind/py_api_topology_discovery.cpp
```diff
@@ -84,6 +84,29 @@ void bind_topology_discovery(nb::module_& m) {
...
+            "get_chip_to_bus_id",
+            &ClusterDescriptor::get_chip_to_bus_id,
+            release_gil(),
+            "Map of ChipId -> PCI bus id (uint16, e.g. 0x41).")
+        .def(
+            "get_bus_id",
+            &ClusterDescriptor::get_bus_id,
+            nb::arg("chip_id"),
+            release_gil(),
+            "Returns the PCI bus id (uint16) for the given chip, or 0 if unknown.")
```
pjanevskiTT, line 109: The docstring says "or 0 if unknown" for `get_bus_id`, but `chip_to_bus_id` is populated for MMIO-capable chips on real hardware (bus 0 is rare but architecturally valid) and via `kMockBusIds` for mock descriptors. Python callers can't distinguish "no entry for this chip" from "this chip happens to be on bus 0." Consider returning `std::optional<uint16_t>` to Python (`nb::optional<uint16_t>`) or documenting the precondition that callers should consult `get_chip_to_bus_id().contains(...)` first.

nanobind/py_api_topology_discovery.cpp
```diff
@@ -84,6 +84,29 @@ void bind_topology_discovery(nb::module_& m) {
...
+        .def(
+            "get_asic_location",
+            &ClusterDescriptor::get_asic_location,
+            nb::arg("chip_id"),
+            release_gil(),
+            "Returns the ASIC location index within the chip's board (uint8), or 0 if unknown.")
```
pjanevskiTT, line 115: Same sentinel ambiguity for `get_asic_location`: ASIC location 0 is a valid location for the first ASIC on a board (this is what `asic_locations` stores). Returning 0 for "unknown" conflates a real value with the missing case. Worth documenting "use 0 only as 'either ASIC 0 or unknown'" or surfacing it as `optional`.

nanobind/py_api_topology_discovery.cpp
pjanevskiTT: will address this as follow up, since it need changes in other code in UMD as well

nanobind/py_api_topology_discovery.cpp
pjanevskiTT: same as above

PR #2611, opened 2026-05-11
AICLK not reaching limit mediation

broskoTT:
### Issue
https://tenstorrent.atlassian.net/browse/DCAMP-186
#2663 

### Description
Per suggestion from the ticket:
Update timeout period, and update warning message to be more informative.

### List of the changes
- Raise timeout from 100 to 200ms, since 100ms is exactly the period for telemetry update. This makes sure that if you do some event, and then read for 200ms you're guaranteed to get at least one update after the event.
- log warning if timeout period is less than the update telemetry period.
- Log the value for AICLK_ARB_MAX in case AICLK didn't reach the final value.

### Testing
I've commented out the line which increases the AICLK, and got these warning messages:

```
2026-05-11 18:08:28.886 | warning  |             UMD | Waiting for AICLK value to settle failed on timeout after 10. Expected to see 1350, last value observed 800. This can be due to possible overheating of the chip or other issues. ASIC temperature: 56.686187744140625 (chip.cpp:254)
2026-05-11 18:08:28.886 | warning  |             UMD | AICLK timeout (10 ms) is not larger than the telemetry update interval (100 ms); the observed AICLK may be a stale telemetry value. Consider increasing AICLK_TIMEOUT. (chip.cpp:268)
2026-05-11 18:08:28.886 | warning  |             UMD | AICLK is clamped by max-arbiter index 0 at 1350 MHz. (chip.cpp:280)
```

### API Changes

### DIFF
--- a/device/chip/chip.cpp
+++ b/device/chip/chip.cpp
@@ -251,33 +251,33 @@ void Chip::wait_for_aiclk_value(
         auto end = std::chrono::steady_clock::now();
         auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
         if (duration.count() > timeout_ms.count()) {
+            auto* telemetry = tt_device->get_arc_telemetry_reader();
+            std::string arb_max_info;
+            if (telemetry != nullptr && telemetry->is_entry_available(TelemetryTag::AICLK_ARB_MAX)) {
+                const uint32_t arb_max = telemetry->read_entry(TelemetryTag::AICLK_ARB_MAX);
+                const uint32_t arb_freq = arb_max & 0xFFFF;
+                const uint32_t arb_idx = (arb_max >> 16) & 0xFFFF;
+                arb_max_info = fmt::format(", AICLK clamped by max-arbiter index {} at {} MHz", arb_idx, arb_freq);
+            }
             log_warning(
                 LogUMD,
                 "Waiting for AICLK value to settle failed on timeout after {}. Expected to see {}, last value "
                 "observed {}. This can be due to possible overheating of the chip or other issues. ASIC temperature: "
-                "{}",
+                "{}{}",
                 timeout_ms.count(),
                 target_aiclk,
                 aiclk,
-                tt_device->get_asic_temperature());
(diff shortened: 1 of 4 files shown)


device/chip/chip.cpp
```diff
@@ -259,6 +260,26 @@ void Chip::wait_for_aiclk_value(
                 target_aiclk,
                 aiclk,
                 tt_device->get_asic_temperature());
+            auto* telemetry = tt_device->get_arc_telemetry_reader();
+            if (telemetry != nullptr) {
+                if (telemetry->is_entry_available(TelemetryTag::UPDATE_TELEM_SPEED)) {
```
pjanevskiTT: not for you to resolve, but maybe we can consider having this check everywhere in the code where we poll the telemetry

device/chip/chip.cpp
```diff
@@ -259,6 +260,26 @@ void Chip::wait_for_aiclk_value(
...
+                    if (timeout_ms.count() <= update_telem_speed_ms) {
+                        log_warning(
+                            LogUMD,
+                            "AICLK timeout ({} ms) is not larger than the telemetry update interval ({} ms); the "
+                            "observed AICLK may be a stale telemetry value. Consider increasing AICLK_TIMEOUT.",
+                            timeout_ms.count(),
+                            update_telem_speed_ms);
+                    }
+                }
+                if (telemetry->is_entry_available(TelemetryTag::AICLK_ARB_MAX)) {
```
pjanevskiTT: does it make sense to always have this check? Should we maybe log warning just if this is clamped to some lower value than 1350? I think this log might not be needed when max AICLK is the actual blackhole max AICLK

pjanevskiTT (approved):

device/chip/chip.cpp
broskoTT: You're right, the separate warning with this info can be confusing, when the max is already 1350. I'll maybe add it as the info, such as the temperature is currently printed out

device/chip/chip.cpp
broskoTT: yeah we actually don't have such cases besides this one. Some similar, but not like this

PR #2618, opened 2026-05-12
tests: port LoopbackStressSize + SimpleApiTest to tests/api/

pjanevskiTT:
### Issue
N/A — coverage migration from `tests/simulation/`.

### Description
Ports two tests from `tests/simulation/test_simulation_device.cpp`
(SimulationChip-level fixture, only runs as part of the `simulation_tests`
target driven by the RTL-sim consumer) into the regular `api_tests`:

- `LoopbackStressSize` — power-of-two size sweep on a TENSIX core (up to 1 MB) and a DRAM core (up to 1 GB). Runs on silicon and simulation.
- `SimpleApiTest` — write/read loopback plus a sequence of `assert_risc_reset`/`deassert_risc_reset` with `ALL_TENSIX`, `ALL_NEO_DMS`, `BRISC`, and a custom `DM0|DM1|DM7` bitmask. Sim-only via `GTEST_SKIP()` because silicon liveness checks would need an arch-specific RISC program; this only verifies the API accepts the masks.

Both tests use `make_cluster_for_test()` from #2635 to pick up the
simulator when `TT_UMD_SIMULATOR` is set. This PR is rebased on top of
#2635 and should land after it.

`tests/simulation/test_simulation_device.cpp` is unchanged in this PR —
its deletion is the follow-up PR stacked on top.

### List of the changes
- `tests/api/test_device_io.cpp`: add `TEST_F(TestDeviceIOFixture, LoopbackStressSize)`.
- `tests/api/test_risc_program.cpp`: add `TEST(TestRiscProgram, SimpleApiTest)`.

### Testing
- `cmake --build build --target api_tests` passes.
- `pre-commit run --files <changed files>` passes.

### API Changes
None.

### DIFF
--- a/tests/api/test_device_io.cpp
+++ b/tests/api/test_device_io.cpp
@@ -11,6 +11,7 @@
 #include <cstring>
 #include <filesystem>
 #include <iostream>
+#include <limits>
 #include <memory>
 #include <optional>
 #include <random>
@@ -948,17 +949,17 @@ TEST(TestDramMembar, StartDeviceDramMembarSubchannel) {
 }
 
 // Stress-size loopback: write/read increasing power-of-two payloads on a Tensix core
-// (up to 1 MB) and a DRAM core (up to 1 GB). The equivalent SimulationChip-level test
+// (up to 1 MB) and a DRAM core (up to 256 MB). The equivalent SimulationChip-level test
 // still lives in tests/simulation/ for the tt-umd-simulators consumer.
 TEST_F(TestDeviceIOFixture, LoopbackStressSize) {
     std::unique_ptr<Cluster> cluster = make_cluster_for_test();
 
-    constexpr uint64_t addr = 0x0;
-    std::random_device rd;
-    std::mt19937 gen(rd());
-    std::uniform_int_distribution<uint32_t> dis(0, 100);
+    const uint32_t seed = std::random_device{}();
+    GTEST_LOG_(INFO) << "LoopbackStressSize RNG seed = " << seed;
+    std::mt19937 gen(seed);
+    std::uniform_int_distribution<uint32_t> dis(0, std::numeric_limits<uint32_t>::max());
 
-    auto run_sweep = [&](ChipId chip_id, const CoreCoord& core, uint32_t max_shift) {
+    auto run_sweep = [&](ChipId chip_id, const CoreCoord& core, uint64_t addr, uint32_t max_shift) {
         for (uint32_t shift = 2; shift <= max_shift; ++shift) {
             const size_t size_bytes = size_t{1} << shift;
             std::vector<uint32_t> wdata(size_bytes / sizeof(uint32_t));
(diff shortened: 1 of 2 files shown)


tests/api/test_device_io.cpp
```diff
@@ -960,3 +950,43 @@ TEST(TestDramMembar, StartDeviceDramMembarSubchannel) {
...
     EXPECT_NO_THROW(cluster.start_device({.init_device = true, .dram_membar_subchannel = 1}));
 }
+
+// Stress-size loopback: write/read increasing power-of-two payloads on a Tensix core
+// (up to 1 MB) and a DRAM core (up to 1 GB). The equivalent SimulationChip-level test
+// still lives in tests/simulation/ for the tt-umd-simulators consumer.
+TEST_F(TestDeviceIOFixture, LoopbackStressSize) {
+    std::unique_ptr<Cluster> cluster = make_cluster();
+
+    constexpr uint64_t addr = 0x0;
```
pjanevskiTT: `addr = 0x0` is used for both the TENSIX and DRAM sweeps. The header in `test_api_common.hpp` documents `SAFE_IO_L1_ADDRESS = 0x1000` and notes that "Low addresses (e.g. 0x10) are reserved on Blackhole by ARC firmware (doppler throttle state), so tests must start at or above this address." The small-size iterations of the TENSIX sweep (4 B, 8 B, … up through 0x10) will land in that reserved range on Blackhole. Consider passing a per-sweep base address and using `SAFE_IO_L1_ADDRESS` for the TENSIX run (DRAM at 0x0 is fine).

tests/api/test_device_io.cpp
```diff
@@ -960,3 +950,43 @@ TEST(TestDramMembar, StartDeviceDramMembarSubchannel) {
...
+    std::random_device rd;
+    std::mt19937 gen(rd());
```
pjanevskiTT: Seeding from `std::random_device` makes a failed `ASSERT_EQ` impossible to reproduce — the failure message prints the size but not the data. Either use a fixed seed, or log the seed (e.g. via `RecordProperty` or a leading `GTEST_LOG_(INFO)`) so a CI failure can be replayed locally.

tests/api/test_device_io.cpp
```diff
@@ -960,3 +950,43 @@ TEST(TestDramMembar, StartDeviceDramMembarSubchannel) {
...
+    std::uniform_int_distribution<uint32_t> dis(0, 100);
```
pjanevskiTT: `uniform_int_distribution<uint32_t> dis(0, 100)` limits each word to 0..100, so the upper 25 bits are always zero. For a loopback stress test that misses the bugs this kind of test usually catches (endianness flips, stuck-at bits in upper bytes, sign-extension on intermediate paths). Suggest widening to the full `uint32_t` range (`dis(0, std::numeric_limits<uint32_t>::max())`).

tests/api/test_device_io.cpp
```diff
@@ -960,3 +950,43 @@ TEST(TestDramMembar, StartDeviceDramMembarSubchannel) {
...
+    for (auto chip_id : cluster->get_target_device_ids()) {
+        const SocDescriptor& soc_desc = cluster->get_soc_descriptor(chip_id);
+
+        const auto& tensix_cores = soc_desc.get_cores(CoreType::TENSIX);
+        ASSERT_FALSE(tensix_cores.empty());
+        run_sweep(chip_id, tensix_cores[0], 20);  // 2^20 = 1 MB
+
+        const auto& dram_cores = soc_desc.get_cores(CoreType::DRAM);
+        ASSERT_FALSE(dram_cores.empty());
+        run_sweep(chip_id, dram_cores[0], 30);  // 2^30 = 1 GB
```
pjanevskiTT: 2^30 = 1 GB per device read + 1 GB per device write, with `wdata` and `rdata` together holding ~2 GB of host memory at peak. On a multi-chip cluster this multiplies per chip, and over the simulator this iteration will dominate runtime. Is this within budget for CI runners, or should the DRAM sweep cap at something like 2^28 (256 MB)?

broskoTT (approved):
Just make sure the tests pass.
I don't think this is the issue that Aleksa was solving with heartbeat.
There is some other issue where we can easily break the eth core, I think I saw it in some of my tests as well..

Issue #1533, opened 2025-10-27
Add regression checks for UMD microbenchmarks


pjanevskiTT:
Tied to the issue #1067 

After setting expected perf for the tests, we can introduce the percentage for toleration of perf on certain tests and then add regression testing, especially for IO to/from device. Not sure how complicated this would be, but maybe it would also be good to pinpoint some crucial tests that can be gate for PRs if PRs regress the perf

broskoTT:
Regression summary is condensed.
Currently setup to run only open/close Cluster test on N150.
Still seeing much variation on Cluster construction with SocDescriptor.
Works good for Cluster construction without SocDescriptor.
Baseline setup manually by a local script

broskoTT:
Baseline filled up from a CI run. 
Numbers stabilised.
Turned on both Cluster Open tests (w/wo soc_desc).
Ran workflow from PR on bgd-lab-06.

ajovanovicTT:
Merged #2691. Will follow in next period how added workflow behaves in main.

PR #2619, opened 2026-05-12
Add CI regression checking workflow

ajovanovicTT:
### Issue
#1533 

### Description
Adds a CI workflow that compares each benchmark run against an in-repo baseline (median + tolerance per case/arch) and surfaces regressions on the run page. Pilot is soft-alert by default; a per-entry gate: true flag escalates a case to a hard job failure when needed.

### List of the changes
 - New workflow regression-check.yml — compares each run against baselines.yaml, writes a summary to the run page, uploads a regression-summary artifact.
 - Wired into build-and-run-all-benchmarks.yml to run alongside analyze_results.yml.
 - New baselines.yaml — median throughput + tolerance per (test, case, arch). Pilot scope: ClusterConstructor and TLB_Tensix_Unicast_v_Multicast on n150 / n300 / p150b. Supports gate: true for hard gating.
 - New summarize_regressions.py — renders the cross-arch table and per-cell detail sections; exits non-zero on gated breaches.
 - New refresh_baselines.py — local tool that recalibrates baselines.yaml from the last N successful main runs. Per case: median_throughput = median(samples) and tolerance_pct = max(5, ceil(3 · stdev_pct)) (three-sigma rule with a 5% floor); existing gate: true flags are preserved.

### Testing
 - CI
 - Manual checks of data gathered by refresh_baselines.py
 - Reviewing compared data

### API Changes

### DIFF
--- a/.github/workflows/build-and-run-all-benchmarks.yml
+++ b/.github/workflows/build-and-run-all-benchmarks.yml
@@ -81,3 +81,11 @@ jobs:
       ubuntu-docker-version: ${{ matrix.ubuntu-docker-version}}
       card: ${{ matrix.test-group.card}}
       baseline-run-id: ${{ inputs.baseline-run-id }}
+
+  # Regression check against in-repo baselines.yaml (median of N main runs +
+  # per-case tolerance). Runs in parallel with analyze-results, which answers a
+  # different question (drift vs. the single latest main run).
+  regression-check:
+    secrets: inherit
+    needs: run-benchmarks
+    uses: ./.github/workflows/regression-check.yml
--- a/.github/workflows/build-and-run-all-benchmarks.yml
+++ b/.github/workflows/build-and-run-all-benchmarks.yml
@@ -89,4 +89,8 @@ jobs:
     secrets: inherit
     needs: run-benchmarks
     if: ${{ !cancelled() && needs.run-benchmarks.result != 'skipped' }}
+    permissions:
+      contents: read
+      packages: read
+      actions: read
     uses: ./.github/workflows/regression-check.yml
--- a/.github/workflows/build-and-run-all-benchmarks.yml
+++ b/.github/workflows/build-and-run-all-benchmarks.yml
@@ -88,4 +88,5 @@ jobs:
   regression-check:
     secrets: inherit
     needs: run-benchmarks
+    if: ${{ !cancelled() && needs.run-benchmarks.result != 'skipped' }}
     uses: ./.github/workflows/regression-check.yml
--- a/.github/workflows/regression-check.yml
+++ b/.github/workflows/regression-check.yml
@@ -65,14 +65,22 @@ jobs:
       - name: Summarize regressions
         run: |
           . /tmp/venv/bin/activate
+          # Capture exit code without aborting — summarize_regressions.py exits 1
+          # when a gated case breaches, but we still want the step summary and
+          # log output published before the step fails.
+          set +e
           python3 tests/microbenchmark/tools/summarize_regressions.py \
             --current "${RESULTS_DIR}" \
             --baselines tests/microbenchmark/expected/baselines.yaml \
             --output "${SUMMARY_PATH}"
-          cat "${SUMMARY_PATH}" >> "${GITHUB_STEP_SUMMARY}"
-          # Also echo to job log so the table is greppable from the run page
-          echo "----- regression summary -----"
-          cat "${SUMMARY_PATH}"
+          rc=$?
+          set -e
+          if [ -f "${SUMMARY_PATH}" ]; then
+            cat "${SUMMARY_PATH}" >> "${GITHUB_STEP_SUMMARY}"
+            echo "----- regression summary -----"
+            cat "${SUMMARY_PATH}"
+          fi
+          exit "${rc}"
 
       - name: Upload regression summary artifact
         if: always()
--- a/.github/workflows/regression-check.yml
+++ b/.github/workflows/regression-check.yml
@@ -5,7 +5,7 @@
 # tolerances, and emits a coarse cross-arch summary plus per-(test, arch) detail
 # sections to $GITHUB_STEP_SUMMARY.
 #
-# Pilot: soft alerts only — the job always exits 0.
+# Pilot: soft by default, but can hard-fail when gate: true breaches.
 name: Regression check against in-repo baseline
 
 on:
@@ -22,8 +22,6 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: ghcr.io/${{ github.repository }}/tt-umd-ci-ubuntu-24.04:latest
-    permissions:
-      contents: read
 
     steps:
       - name: Checkout regression tooling
--- a/.github/workflows/regression-check.yml
+++ b/.github/workflows/regression-check.yml
@@ -20,6 +20,11 @@ jobs:
   regression-check:
     name: Regression summary vs baselines.yaml
     runs-on: ubuntu-latest
+    timeout-minutes: 15
+    permissions:
+      contents: read   # actions/checkout
+      packages: read   # pull ghcr.io container image
+      actions: read    # actions/download-artifact API access
     container:
       image: ghcr.io/${{ github.repository }}/tt-umd-ci-ubuntu-24.04:latest
 
@@ -82,7 +87,7 @@ jobs:
 
       - name: Upload regression summary artifact
         if: always()
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
         with:
           name: regression-summary
           path: ${{ env.SUMMARY_PATH }}
--- a/.github/workflows/regression-check.yml
+++ b/.github/workflows/regression-check.yml
@@ -35,8 +35,24 @@ jobs:
             tests/microbenchmark/tools/analyze_results.py
           sparse-checkout-cone-mode: false
 
+      # Container is Ubuntu 24.04 (PEP 668-enforced); install into a venv to
+      # match the pattern in analyze_results.yml.
+      - name: Install python3-venv
+        run: |
+          apt-get update -qq
+          apt-get install -y --no-install-recommends python3-venv
+
+      - name: Create and activate virtual environment
+        run: |
+          python3 -m venv /tmp/venv
+          . /tmp/venv/bin/activate
+          echo "PATH=$PATH" >> $GITHUB_ENV
+          echo "VIRTUAL_ENV=$VIRTUAL_ENV" >> $GITHUB_ENV
+
       - name: Install python dependencies
-        run: python3 -m pip install --no-cache-dir pandas psutil pyyaml tabulate
+        run: |
+          . /tmp/venv/bin/activate
+          python3 -m pip install --no-cache-dir pandas psutil pyyaml tabulate
 
       - name: Download all benchmark JSON artifacts
         uses: actions/download-artifact@v8
@@ -48,6 +64,7 @@ jobs:
 
       - name: Summarize regressions
         run: |
+          . /tmp/venv/bin/activate
           python3 tests/microbenchmark/tools/summarize_regressions.py \
             --current "${RESULTS_DIR}" \
             --baselines tests/microbenchmark/expected/baselines.yaml \
--- a/.github/workflows/regression-check.yml
+++ b/.github/workflows/regression-check.yml
@@ -1,0 +1,66 @@
+# Regression check against in-repo baselines.yaml.
+# Complementary to analyze_results.yml (which diffs current run vs the single
+# latest main run). This workflow compares against the median of N main runs
+# stored in tests/microbenchmark/expected/baselines.yaml, applies per-case
+# tolerances, and emits a coarse cross-arch summary plus per-(test, arch) detail
+# sections to $GITHUB_STEP_SUMMARY.
+#
+# Pilot: soft alerts only — the job always exits 0.
+name: Regression check against in-repo baseline
+
+on:
+  workflow_call:
+    # No inputs — downloads benchmark-json-* artifacts from the parent run.
+
+env:
+  RESULTS_DIR: /tmp/all-arch-results
+  SUMMARY_PATH: /tmp/regression_summary.md
+
+jobs:
+  regression-check:
+    name: Regression summary vs baselines.yaml
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/${{ github.repository }}/tt-umd-ci-ubuntu-24.04:latest
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout regression tooling
+        uses: actions/checkout@v6
+        with:
+          sparse-checkout: |
+            tests/microbenchmark/expected/baselines.yaml
+            tests/microbenchmark/tools/summarize_regressions.py
+            tests/microbenchmark/tools/analyze_results.py
+          sparse-checkout-cone-mode: false
+
+      - name: Install python dependencies
+        run: python3 -m pip install --no-cache-dir pandas psutil pyyaml tabulate
+
+      - name: Download all benchmark JSON artifacts
+        uses: actions/download-artifact@v8
+        with:
+          pattern: benchmark-json-*
+          path: ${{ env.RESULTS_DIR }}
+          merge-multiple: false
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Summarize regressions
+        run: |
+          python3 tests/microbenchmark/tools/summarize_regressions.py \
+            --current "${RESULTS_DIR}" \
+            --baselines tests/microbenchmark/expected/baselines.yaml \
+            --output "${SUMMARY_PATH}"
+          cat "${SUMMARY_PATH}" >> "${GITHUB_STEP_SUMMARY}"
+          # Also echo to job log so the table is greppable from the run page
+          echo "----- regression summary -----"
+          cat "${SUMMARY_PATH}"
+
+      - name: Upload regression summary artifact
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: regression-summary
+          path: ${{ env.SUMMARY_PATH }}
+          if-no-files-found: warn
--- a/tests/microbenchmark/expected/baselines.yaml
+++ b/tests/microbenchmark/expected/baselines.yaml
@@ -2,8 +2,14 @@
 # Calibrated from 30 successful main runs of build-and-run-all-benchmarks.yml.
 # To refresh: tests/microbenchmark/tools/refresh_baselines.py --runs 30
 #
-# Pilot scope: ClusterConstructor only. Other suites (TLB, PCIe DMA, IOMMU,
-# Ethernet I/O) will be onboarded once the pipeline is validated end-to-end.
+# Pilot scope: ClusterConstructor + TLB_Tensix_Unicast_v_Multicast. Other suites
+# (TLB_DRAM, PCIe DMA, IOMMU, Ethernet I/O) will be onboarded once the pipeline
+# is validated end-to-end.
+#
+# Note on per-arch case sets: TLB_Tensix_Unicast_v_Multicast varies its core
+# count by arch (72 cores on n150, 64 on n300, 120 on p150b), so each case
+# entry contains a single arch — the cell totals in the summary table will
+# differ per arch, which is expected.
 #
 # Schema:
 #   <bench_title>:
@@ -16,7 +22,7 @@
 metadata:
   calibrated_at: "2026-05-11"
   calibrated_from_runs: 30
-  notes: "Pilot — soft alerts only. ClusterConstructor only."
+  notes: "Pilot — soft alerts only. ClusterConstructor + TLB_Tensix_Unicast_v_Multicast."
 
 ClusterConstructor:
   default:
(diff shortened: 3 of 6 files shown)


tests/microbenchmark/expected/baselines.yaml
```diff
@@ -0,0 +1,217 @@
...
+# differ per arch, which is expected.
+#
+# Schema:
+#   <bench_title>:
+#     <case_name>:
+#       <arch_label>:
+#         median_throughput: float       # ops/s (batch / median_elapsed)
+#         tolerance_pct: float           # symmetric +-%; alert if |delta%| > this
+#         gate: bool (optional)          # Phase 2 hard gates; preserved on refresh
+
```
ajovanovicTT, line 26: For this pilot only ClusterConstructor and TLB_Tensix_Unicast_v_Multicast tests are included (TLB_Tensix_Unicast_v_Multicast is added since on current infra it shows the smallest variability over commits, so it has tightest tolerances). Generally, would expect that when we have stable infrastructure with dedicated runners, those tolerances become much tighter.

tests/microbenchmark/expected/baselines.yaml
```diff
@@ -0,0 +1,217 @@
...
+  "Unicast, 72 cores, 4096 bytes":
+    WH n150: { median_throughput: 2.865e+07, tolerance_pct: 5 }
+  "Unicast, 72 cores, 524288 bytes":
+    WH n150: { median_throughput: 3.112e+07, tolerance_pct: 5 }
+  "Unicast, 72 cores, 65536 bytes":
+    WH n150: { median_throughput: 3.102e+07, tolerance_pct: 5 }
+  "Unicast, 72 cores, 8 bytes":
+    WH n150: { median_throughput: 1.194e+05, tolerance_pct: 38 }
+  "Unicast, 72 cores, 8192 bytes":
+    WH n150: { median_throughput: 2.99e+07, tolerance_pct: 5 }
```
ajovanovicTT, line 223: After adding medians of other tests in baseline.yaml they automatically become included in comparison. Optional gate: true arg per test case make CI actually fail on that test case, not just report.

ajovanovicTT:
Example of how it looks now in the CI with gate true only on one test case for one arch : https://github.com/tenstorrent/tt-umd/actions/runs/25792176133. For now Gated breaches is separate section from Slowdown reports section, but in the future if we have stable environment and want to have gates on as many microbenchmark test cases possible, it might be too long summary so would aim to unify to one section (leave on example only on what gates ci failed), while this first general table would persist as is.

How it looks like when there is no regression detected: https://github.com/tenstorrent/tt-umd/actions/runs/25726849170


Example when there is speedup as well : https://github.com/tenstorrent/tt-umd/actions/runs/25795239593

.github/workflows/build-and-run-all-benchmarks.yml
```diff
@@ -81,3 +81,11 @@ jobs:
       ubuntu-docker-version: ${{ matrix.ubuntu-docker-version}}
       card: ${{ matrix.test-group.card}}
```
pjanevskiTT, line 82: **[from Claude — design discussion, needs your input]**

**Design concern: in-repo static baselines vs. rolling baseline computed live in CI**

Every recalibration of `baselines.yaml` requires a hand-reviewed PR — fine in principle, but `refresh_baselines.py` is a manual local tool, so in practice baselines will drift between human-initiated refreshes, and each refresh will then shift dozens of tolerances at once (the diff-line warnings already in `refresh_baselines.py` anticipate this exact friction).

A more robust alternative: store rolling stats as a CI-managed artifact (a `perf-history` orphan branch, a GitHub release asset, or an external blob), recomputed on every successful `main` run from the trailing N runs. The repo then keeps only the small human-curated bits — the schema and the `gate: true` flags — while the numbers live as a rolling window. Benefits:

- No PR required to keep baselines fresh; no drift.
- Naturally enables K-of-N debouncing across recent runs (greatly reduces false-positive rate vs. single-sample-vs-median comparisons).
- `baselines.yaml` stops growing into a thousands-of-entries diff that nobody can review meaningfully.

Not a blocker for the pilot scope, but worth deciding before phase-2 expansion of suites/archs makes the static-YAML approach hard to walk back.

pjanevskiTT, line 82: **[from Claude — design discussion, needs your input]**

**Design concern: surface improvements as improvements, not as visual breaches**

Confirmed that `classify()` returns `"UP"` for `Δ% > tolerance` and that `gated_breaches` correctly excludes `UP` — so a real speedup never fails CI even with `gate: true`. That part is right, and per the maintainer ask, we *do* want improvements tracked rather than silenced.

Remaining issue is purely presentation: the cell tag and the coarse table render `UP (n)` next to `DOWN (m)` with the same visual weight, so a reviewer skimming the summary sees red-coded counts on both sides and has to read the legend to know one is good. Two small changes preserve full bi-directional tracking:

1. Relabel `UP` → `IMP` in cell tags and add an "Improvements" sub-table heading symmetric with "Slowdowns" — the renderer already has the structure for it.
2. Optional: a CRIT-up tier at `2·tolerance` so dramatic improvements (likely measurement bugs or genuine wins worth attention) are called out — but still never gateable.

Worth adding a one-line code comment in `classify()` stating that the symmetric tolerance is intentional and that `UP` is intentionally non-gateable, so a future maintainer doesn't mistake the symmetry for a bug.

.github/workflows/build-and-run-all-benchmarks.yml
```diff
@@ -81,3 +81,11 @@ jobs:
...
       baseline-run-id: ${{ inputs.baseline-run-id }}
+
+  # Regression check against in-repo baselines.yaml (median of N main runs +
+  # per-case tolerance). Runs in parallel with analyze-results, which answers a
+  # different question (drift vs. the single latest main run).
+  regression-check:
```
pjanevskiTT, line 88: **[major]** The new `regression-check` job declares `needs: run-benchmarks` but has no `if:` clause. GitHub skips dependent jobs by default when any needed matrix shard fails, so a single flaky arch will silently skip the regression check entirely — defeating the gate. Add `if: ${{ !cancelled() && needs.run-benchmarks.result != 'skipped' }}` (or `always()`) so the check runs against whatever artifacts did upload.

.github/workflows/regression-check.yml
```diff
@@ -0,0 +1,89 @@
...
+          . /tmp/venv/bin/activate
+          echo "PATH=$PATH" >> $GITHUB_ENV
+          echo "VIRTUAL_ENV=$VIRTUAL_ENV" >> $GITHUB_ENV
+
+      - name: Install python dependencies
+        run: |
+          . /tmp/venv/bin/activate
+          python3 -m pip install --no-cache-dir pandas psutil pyyaml tabulate
+
+      - name: Download all benchmark JSON artifacts
```
pjanevskiTT, line 60: **[blocker]** `download-artifact` has no empty-result guard. If every `run-benchmarks` matrix shard fails (or the artifact naming drifts), `pattern: benchmark-json-*` matches zero artifacts, the action succeeds silently, and `summarize_regressions.py` renders every cell as "no result" and exits 0. A totally-broken benchmark run will look like "regression check passed". Either fail explicitly when nothing is downloaded (check the action's `artifact-ids` output), or have the python tool exit non-zero when collected results are empty.

.github/workflows/regression-check.yml
```diff
@@ -0,0 +1,89 @@
...
+
+on:
+  workflow_call:
+    # No inputs — downloads benchmark-json-* artifacts from the parent run.
+
+env:
+  RESULTS_DIR: /tmp/all-arch-results
+  SUMMARY_PATH: /tmp/regression_summary.md
+
+jobs:
```
pjanevskiTT, line 19: **[major]** No top-level or job-level `permissions:` block and no `timeout-minutes`. Since this workflow is reachable from a `pull_request` trigger (via `build-and-run-all-benchmarks.yml`), it inherits whatever the calling workflow / repo default grants — likely wider than needed. Add `permissions: { contents: read }` at job level and `timeout-minutes: 15` so a hung step can't burn a runner.

.github/workflows/regression-check.yml
```diff
@@ -0,0 +1,89 @@
...
+          if [ -f "${SUMMARY_PATH}" ]; then
+            cat "${SUMMARY_PATH}" >> "${GITHUB_STEP_SUMMARY}"
+            echo "----- regression summary -----"
+            cat "${SUMMARY_PATH}"
+          fi
+          exit "${rc}"
+
+      - name: Upload regression summary artifact
+        if: always()
+        uses: actions/upload-artifact@v4
```
pjanevskiTT: **[major]** Uses `actions/upload-artifact@v4`, but the rest of this repo (`run-benchmarks.yml`, `build-device.yml`, `release.yml`) standardized on `@v7`. Inconsistent action versions across a single repo are a maintenance footgun and confuse Dependabot bumps. Bump to `@v7` to match.

tests/microbenchmark/expected/baselines.yaml
```diff
@@ -0,0 +1,217 @@
...
+  "Unicast, 64 cores, 2 bytes":
+    WH n300: { median_throughput: 1.51e+04, tolerance_pct: 42 }
+  "Unicast, 64 cores, 2048 bytes":
+    WH n300: { median_throughput: 3.008e+07, tolerance_pct: 7 }
+  "Unicast, 64 cores, 262144 bytes":
+    WH n300: { median_throughput: 3.519e+07, tolerance_pct: 7 }
+  "Unicast, 64 cores, 32768 bytes":
+    WH n300: { median_throughput: 3.501e+07, tolerance_pct: 8 }
+  "Unicast, 64 cores, 4 bytes":
+    WH n300: { median_throughput: 7.456e+04, tolerance_pct: 93 }
```
pjanevskiTT, line 183: **[nit]** `tolerance_pct: 93` (and ≥50% elsewhere — lines 45, 47, 49, 57, 58, 185) effectively makes these rows un-gateable noise. Acknowledged that small-payload cases are dominated by shared-runner variance and dedicated hardware is pending — that's the right call. Worth deciding the policy now and documenting it: keep these rows as informational-only (and document that `gate: true` should never be set on a `tolerance_pct > 30` entry), or omit them from `baselines.yaml` until dedicated hardware lands. Otherwise reviewers will keep asking "why is tolerance 93%?" and rediscover the same context.

tests/microbenchmark/expected/baselines.yaml
```diff
@@ -0,0 +1,217 @@
...
+#
+# Schema:
+#   <bench_title>:
+#     <case_name>:
+#       <arch_label>:
+#         median_throughput: float       # ops/s (batch / median_elapsed)
+#         tolerance_pct: float           # symmetric +-%; alert if |delta%| > this
+#         gate: bool (optional)          # Phase 2 hard gates; preserved on refresh
+
+metadata:
```
pjanevskiTT, line 27: **[nit]** Two small additions to `metadata:` worth doing while the file is small:

1. Add `schema_version: 1`. When you add a field in phase 2 (`unit`, `samples`, `p95`), there'll be no way to detect old/new programmatically.
2. Add a `notes:` line stating that wide tolerances on small-payload cases reflect shared-runner heterogeneity, with tightening planned when dedicated boards land. Future maintainers will otherwise treat them as a calibration bug to be fixed.

tests/microbenchmark/tools/summarize_regressions.py
```diff
@@ -0,0 +1,429 @@
...
+from pathlib import Path
+
+import yaml
+
+# --- Arch label canonicalization -------------------------------------------------
+# Artifact names look like:
+#   benchmark-json-wormhole_b0-tt-ubuntu-2204-n150-viommu-stable-ubuntu-22.04
+# We canonicalize to short labels matching the keys in baselines.yaml.
+# Substring match keeps this robust to minor variations (e.g. -iter1 suffixes).
+ARCH_PATTERNS = [
```
pjanevskiTT, line 38: **[blocker]** `ARCH_PATTERNS` uses substring match with `all(n in name for n in needles)`. Substrings like `n150` can match in unrelated tokens, and there's no anchoring — an artifact whose name contains both `n150` and `n300` would silently pick the first label listed since the check returns on first match. Anchor to word boundaries (e.g. `re.search(r'(?:^|-)n150(?:-|$)', name)`) and order most-specific first.

tests/microbenchmark/tools/summarize_regressions.py
```diff
@@ -0,0 +1,429 @@
...
+                for case_name, case_entry in baselines[title].items()
+                if arch in case_entry
+            }
+            if not cases_with_baseline:
+                cell_state[(title, arch)] = (counts, breached, stable, "no baseline")
+                continue
+            if not arch_results:
+                cell_state[(title, arch)] = (counts, breached, stable, "no result")
+                continue
+            for case_name, baseline_entry in cases_with_baseline.items():
```
pjanevskiTT, line 308: **[major]** A case present in `baselines.yaml` but missing from this run's JSON is silently `continue`d and does **not** count as a breach — even for `gate: true` cases. A regression that crashes the benchmark binary (no JSON written, or only a subset of cases run) will pass the gate. At minimum, warn to stderr per missing gated case; ideally render a "MISSING" status that counts as a gated breach so a partial benchmark run can't be confused with a clean one.

tests/microbenchmark/tools/summarize_regressions.py
```diff
@@ -0,0 +1,429 @@
...
+                    "throughput": batch / med,
+                    "unit": unit,
+                }
+    return results
+
+
+# --- Classification --------------------------------------------------------------
+
+
+def classify(
```
pjanevskiTT, line 129: **[major]** `classify()` will `ZeroDivisionError` if `baseline_thr == 0`. Current-run results are guarded for `med == 0` / `batch == 0` above, but baseline values are not. A malformed/zero baseline crashes the entire summary mid-walk after some markdown is already built. Guard and skip-with-warn, matching the script's overall skip-and-continue posture.

tests/microbenchmark/tools/summarize_regressions.py
```diff
@@ -0,0 +1,429 @@
...
+    current_thr: float, baseline_thr: float, tolerance_pct: float
+) -> tuple[str, float]:
+    """Return (status, signed Δ%). Δ% = (current − baseline) / baseline × 100."""
+    delta_pct = (current_thr - baseline_thr) / baseline_thr * 100.0
+    if abs(delta_pct) <= tolerance_pct:
+        return "OK", delta_pct
+    if delta_pct > tolerance_pct:
```
pjanevskiTT, line 136: **[major]** Boundary gap in `classify()`: when `delta_pct == -2 * tolerance_pct` exactly, neither the `> tolerance_pct` nor the `< -2 * tolerance_pct` branch fires, so a clean 2× breach falls through and gets classified `DOWN`, not `CRIT`. Use `<=` on the CRIT branch (or document the half-open interval and align both edges).

tests/microbenchmark/tools/summarize_regressions.py
```diff
@@ -0,0 +1,429 @@
...
+        if not artifact_dir.name.startswith("benchmark-json-"):
+            continue
+        arch = arch_label_from_artifact(artifact_dir.name)
+        if arch is None:
+            print(
+                f"WARN: cannot derive arch label from artifact {artifact_dir.name}; skipping",
+                file=sys.stderr,
+            )
+            continue
+        for path in sorted(artifact_dir.glob("*.json")):
```
pjanevskiTT: **[major]** `json.load(f)` on a malformed/empty/truncated nanobench JSON raises and aborts the entire summary, losing all other arches and tests. Wrap in `try/except (json.JSONDecodeError, OSError)`, warn to stderr, and continue with the next file.

tests/microbenchmark/tools/summarize_regressions.py
```diff
@@ -0,0 +1,429 @@
...
+    if up:
+        parts.append(f"{up} {decorate_status('UP')}")
+    return ", ".join(parts) + f" (of {total})"
+
+
+# Width of the truncation for detail subtables — see plan §"Detail section rules".
+DETAIL_TRUNCATION_LIMIT = 10
+
+
+def render_detail_subtable(title: str, rows: list) -> str:
```
pjanevskiTT, line 174: **[major]** Markdown cells don't escape `|` in `case` / `title` strings. A future case name containing `|` (e.g. a parameterized variant or any human-edited case name) silently shatters the markdown table. Add `.replace("|", "\\|")` at all three row-format sites: this `render_detail_subtable`, the test-title column around L332, and the gated-breaches table around L358.

tests/microbenchmark/tools/summarize_regressions.py
```diff
@@ -0,0 +1,429 @@
...
+    cases that breached tolerance AND carry `gate: true` in baselines.yaml —
+    the caller exits non-zero when this list is non-empty.
+    """
+    meta = baselines.get("metadata") or {}
+    calibrated_at = meta.get("calibrated_at", "unknown date")
+    calibrated_from_runs = meta.get("calibrated_from_runs", "unknown")
+
+    # Tests come from baselines (so a missing-from-baseline test is visible but
+    # never alerts).
+    test_titles = [k for k in baselines if k != "metadata"]
```
pjanevskiTT, line 273: **[major]** `archs` is populated only from `baselines.yaml`, so an arch that ran in CI but has zero baseline entries (e.g. a newly-added board, or one whose label changed) is invisible in the summary — no row, no warning. Either union `current.keys()` into `archs` and render those columns as "no baseline", or at minimum print a stderr warning enumerating arches present in `current` but absent from `baselines`.

tests/microbenchmark/tools/refresh_baselines.py
```diff
@@ -0,0 +1,359 @@
...
+    stdev_thr = statistics.stdev(samples)
+    stdev_pct = stdev_thr / median_thr * 100 if median_thr else 0.0
+    tolerance_pct = max(TOLERANCE_FLOOR_PCT, math.ceil(3 * stdev_pct))
+    return median_thr, float(tolerance_pct), suffix
+
+
+# --- YAML rendering --------------------------------------------------------------
+
+
+def render_yaml(
```
pjanevskiTT, line 195: **[blocker]** YAML is emitted via hand-rolled string concatenation rather than a proper round-trip emitter. Any human-added comments, custom key order, or formatting in `baselines.yaml` is wiped on every refresh — including the `metadata.notes` field, which is hardcoded back to `"Pilot — soft alerts only."` on L219, so any future note added by hand gets clobbered. Use `ruamel.yaml` round-trip mode to preserve comments and merge into the existing structure, or at minimum carry forward the existing `metadata` block.

tests/microbenchmark/tools/refresh_baselines.py
```diff
@@ -0,0 +1,359 @@
...
+    out.write("\n")
+    out.write("metadata:\n")
+    out.write(f'  calibrated_at: "{date.today().isoformat()}"\n')
+    out.write(f"  calibrated_from_runs: {runs_used}\n")
+    out.write('  notes: "Pilot — soft alerts only."\n')
+    out.write("\n")
+
+    # Deterministic ordering: tests alphabetical, archs alphabetical within case.
+    all_titles = sorted({title for arch in samples for title in samples[arch]})
+    for title in all_titles:
```
pjanevskiTT, line 234: **[major]** Cases that exist in the **old** `baselines.yaml` but are absent from the new sample set (flaky tests that produced no data in any of the N runs, or were skipped) are silently dropped from output — including their `gate: true` flag. This violates the PR-stated invariant "existing `gate: true` flags are preserved." Either carry forward old entries with a `# stale, no recent samples` comment, or hard-fail when a previously-gated case has no samples in the refresh window.

tests/microbenchmark/tools/refresh_baselines.py
```diff
@@ -0,0 +1,359 @@
...
+        headers={
+            "Authorization": f"Bearer {token}",
+            "Accept": "application/vnd.github+json",
+        },
+    )
+    with _opener().open(req) as r:
+        return r.read() if raw else json.load(r)
+
+
+def list_recent_main_runs(repo: str, workflow: str, n: int, token: str) -> list[dict]:
```
pjanevskiTT, line 87: **[major]** Two GitHub-API issues in this call:

1. `per_page={n}` is unpaginated, but the API caps `per_page` at 100 — so `--runs 200` silently returns ≤100 runs without warning. Paginate via the `Link` header until N successful runs are collected.
2. The `status=success` filter is wrong field: the API uses `status` for in-progress state (queued/in_progress/completed) and `conclusion` for outcome (success/failure/...). The `status=success` alias is accepted in newer API versions but is undocumented and fragile. Use `status=completed&conclusion=success` or filter client-side on `r["conclusion"] == "success"`.

tests/microbenchmark/tools/refresh_baselines.py
```diff
@@ -0,0 +1,359 @@
...
+    if not runs:
+        sys.exit("No successful runs found.")
+    print(f"Got {len(runs)} runs. Downloading benchmark artifacts...", file=sys.stderr)
+    samples, _units = collect_samples(args.repo, runs, token)
+    if not samples:
+        sys.exit(
+            "No samples collected — workflow may not produce benchmark-json-* artifacts."
+        )
+
+    output_path = args.output.resolve()
```
pjanevskiTT, line 344: **[major]** No `--dry-run` and no diff preview before overwriting. The docstring says "review the diff and submit it as a PR," but the script unconditionally writes the file in place. Add `--dry-run` (print proposed contents to stdout, don't write) and ideally print a unified diff against the existing file before writing. Without this, a typo in `--repo` or `--runs` will overwrite the baselines and require a `git checkout` to recover.

tests/microbenchmark/tools/refresh_baselines.py
```diff
@@ -0,0 +1,359 @@
...
+
+# --- Tolerance derivation --------------------------------------------------------
+
+
+def derive_baseline_entry(samples: list[float]) -> tuple[float, float, str]:
+    """Return (median_throughput, tolerance_pct, comment).
+
+    `comment` is a free-text annotation (may be empty) for the YAML.
+    """
+    if len(samples) < MIN_SAMPLES_FOR_RELIABLE:
```
pjanevskiTT, line 176: **[major]** When `len(samples) < 2`, the function returns the median plus the 5% floor — but never surfaces the single-sample case in `diff_lines` or in stderr. A case that only appears in one of N runs (e.g. a new test added in the latest run) gets baked into `baselines.yaml` with 5% tolerance based on **one data point**, and downstream consumers can't distinguish it from a well-calibrated entry. Add `--min-samples` (default 3) and skip cases below it, or emit them with a clear `# calibrated from 1 sample` annotation.

tests/microbenchmark/expected/baselines.yaml
```diff
@@ -0,0 +1,217 @@
+# tests/microbenchmark/expected/baselines.yaml
+# Calibrated from 30 successful main runs of build-and-run-all-benchmarks.yml.
```
pjanevskiTT, line 2: we should discuss what is the proper strategy to get these results. Few things that are top of my mind

- Measure actual baselines on proper dedicated machines and put that. The problem with this is that we don't have dedicated machines in CI at the moment so those numbers may vary in CI and become useless
- Do something like Aleksa already did (median or smh) from CI runs, but that may not be true on dedicated machine setup so we have to update them later on again when we have dedicated machines

@broskoTT could we wire up bgd-lab-06 and bh-40 as our dedicated machines to run things in CI as well?

tests/microbenchmark/expected/baselines.yaml
```diff
@@ -0,0 +1,217 @@
...
+# To refresh: tests/microbenchmark/tools/refresh_baselines.py --runs 30
```
pjanevskiTT, line 3: look at claude comment if we could somehow run this remotely, not statically, also needs discussion

tests/microbenchmark/expected/baselines.yaml
```diff
@@ -0,0 +1,217 @@
...
+#         gate: bool (optional)          # Phase 2 hard gates; preserved on refresh
+
+metadata:
+  calibrated_at: "2026-05-11"
+  calibrated_from_runs: 30
+  notes: "Pilot — soft alerts only. ClusterConstructor + TLB_Tensix_Unicast_v_Multicast."
+
+ClusterConstructor:
+  default:
+    BH p150b: { median_throughput: 119.5, tolerance_pct: 45 }
```
pjanevskiTT, line 35: `median_throughput` doesn't seem as the right name for this field because the metric might be different than throughput. For example I would assume that this is time in ms for ClusterConstructor. So I would like something like

`  BH p150b: { median: 119.5, tolerance_pct: 45, unit: ms }`

tests/microbenchmark/expected/baselines.yaml
```diff
@@ -0,0 +1,217 @@
...
+    WH n150: { median_throughput: 138.3, tolerance_pct: 30 }
+    WH n300: { median_throughput: 41.16, tolerance_pct: 16 }
+  from sdesc:
+    BH p150b: { median_throughput: 104.1, tolerance_pct: 49 }
+    WH n150: { median_throughput: 161.9, tolerance_pct: 15 }
+    WH n300: { median_throughput: 36.76, tolerance_pct: 20 }
+
+TLB_Tensix_Unicast_v_Multicast:
+  "Multicast, 120 cores, 1 bytes":
```
pjanevskiTT, line 44: this is very nice, I like per transfer size benchmark without having to run separate tests for that

tests/microbenchmark/tools/refresh_baselines.py
```diff
@@ -0,0 +1,359 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
```
pjanevskiTT, line 1: won't review this in too much detail currently, but from skimming it seems quite readable to me, it matches the workflow imo

tests/microbenchmark/tools/summarize_regressions.py
```diff
@@ -0,0 +1,429 @@
...
+# --- Rendering -------------------------------------------------------------------
+
+# Hybrid emoji + ASCII labels. The Step Summary tab renders the emoji as a
+# colored glyph; the raw job log and `grep` still match the ASCII suffix. OK
+# stays plain — no need to call attention to passing cases.
+STATUS_DECORATION = {
+    "OK": "OK",
+    "UP": "🟢 UP",
+    "DOWN": "🟡 DOWN",
+    "CRIT": "🔴 CRIT",
```
pjanevskiTT: I would maybe consider not having both DOWN and CRIT. I would maybe just have UP and DOWN, no matter the pct of the failure. Then in CI view we could maybe sort it based on pct of the failure or something like that

pjanevskiTT (commented):
imo this is very nice work for v0 regression checking. In order to get this approved and checked in, I would like to hear from other UMD folks and also work on fixing few tests with smallest pct of variance as gate so we could actually have this infra in CI

tests/microbenchmark/tools/summarize_regressions.py
```diff
@@ -0,0 +1,429 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
```
pjanevskiTT, line 1: can we call this with local changes to compare it against the baselines? I think that would be useful. Seems like this file alone can't do it but maybe having another file that can run the benchmarks and compare local results to those checked in the repo would be nice

tests/microbenchmark/expected/baselines.yaml
ajovanovicTT: Documented on the top of yaml in the comment for now.

tests/microbenchmark/tools/summarize_regressions.py
ajovanovicTT: Would not expect n150 and n300 in one artifact name in current run-benchmarks configuration.

tests/microbenchmark/tools/summarize_regressions.py
ajovanovicTT: A 0 baseline_thr can only come from a manually-edited YAML where someone typed median_throughput: 0, in current configuration would not expect that ever happen, so would leave this as is now.

tests/microbenchmark/tools/summarize_regressions.py
ajovanovicTT: Would not expect | symbol in bench case name.

tests/microbenchmark/tools/summarize_regressions.py
ajovanovicTT: Will leave this for later consideration, for now would expect to update baselines when new infra is set up, for arches that tests are going to be run on.

tests/microbenchmark/tools/refresh_baselines.py
ajovanovicTT: If tests are dropped in on example last 30 runs, that would be intentional, so removing gates on them would be desirable. In general wouldn't expect many changes in benchmarks that will be included in perf reporting when everything is set up.

tests/microbenchmark/tools/refresh_baselines.py
ajovanovicTT: There is no conclusion query parameter on this endpoint, checked, only status exists. Leaving as is.

tests/microbenchmark/tools/refresh_baselines.py
ajovanovicTT: Added warnings in stderr when running script.

tests/microbenchmark/tools/summarize_regressions.py
ajovanovicTT: Removed CRIT, left only DOWN tag, and in detailed view breaches are already sorted to show biggest outliers first.

.github/workflows/build-and-run-all-benchmarks.yml
ajovanovicTT: Removed CRIT tag so this is outdated now.

ajovanovicTT:
Here is updated summary look : https://github.com/tenstorrent/tt-umd/actions/runs/25860065379
Added missing gated sections for cases that are in baseline but missing from current run.

tests/microbenchmark/expected/baselines.yaml
broskoTT: Yes, consider bgd-lab-06 05 and bh-40 our dedicated machines for now

Issue #2599, opened 2026-05-08
Remote IO failure in tests


broskoTT:
We've known about this for a while, and yes, we should add a retry to warm reset tests.

I think we lowered the heartbeat wait time a bit and now we're seeing failures when it shouldn't fail, e.g. I think the timeout is small:

https://github.com/tenstorrent/tt-umd/actions/runs/25544060460/job/74976670559

broskoTT:
https://github.com/tenstorrent/tt-umd/actions/runs/25561414615/job/75036965270?pr=2596

broskoTT:
This is a fail on warm reset specifically: https://github.com/tenstorrent/tt-umd/actions/runs/25557618693/job/75023196116

broskoTT:
Another failure on CreateDestroy: https://github.com/tenstorrent/tt-umd/actions/runs/25596274699/job/75142876317

broskoTT:
Another real fail: https://github.com/tenstorrent/tt-umd/actions/runs/25502637472/job/74840401916

broskoTT:
Another true fail: https://github.com/tenstorrent/tt-umd/actions/runs/25598392159/job/75148407990

broskoTT:
another probably true failure: https://github.com/tenstorrent/tt-umd/actions/runs/25660532115/job/75320386939

broskoTT:
Looks fixed now.

PR #2623, opened 2026-05-12
Warm reset with remote IO failure recovery

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/2599

### Description
Related to long standing issue that we know of, where erisc can be stuck after warm reset. This should fix our ci runs.

### List of the changes
- Implement WarmReset with recovery, which will run warm reset a couple of times
- Expose in nanobind
- Make all our CI tests use this new class to avoid CI issues
- Adapt the warm_reset tool to have the option to use this

### Testing
This is a test of 100 iterations of warm reset, hopefully shows up that this class does fix some issues: https://github.com/tenstorrent/tt-umd/actions/runs/25800112547
Note: For some reason, it seems exceptionally hard to repro that issue at the moment. I made a test case locally which returned heartbeat stuck on first try of topo discovery only. Confirmed locally that the attempt 2 was called inside python test. I expect it to work on our CI as well.

### API Changes
This PR has API changes:, introduces a new warm reset class

### DIFF
--- a/device/api/umd/device/warm_reset_with_recovery.hpp
+++ b/device/api/umd/device/warm_reset_with_recovery.hpp
@@ -1,0 +1,50 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <chrono>
+#include <string>
+#include <vector>
+
+#include "umd/device/utils/timeouts.hpp"
+
+namespace tt::umd {
+
+// Wraps WarmReset entry points with a follow-up TopologyDiscovery::discover() and retries
+// the whole sequence if discovery fails. This works around transient post-reset states
+// (e.g. an ETH core that never recovers its heartbeat) which can leave the board in a
+// state that only another reset can clear.
+//
+// Each method returns true if a (reset, discovery) attempt eventually succeeds, false
+// otherwise. If the underlying warm reset itself fails, the methods bail immediately
+// without retrying because such failures (no devices, ARM platform) do not recover by
+// retrying.
+class WarmResetWithRecovery {
+public:
+    static bool warm_reset(
+        int max_attempts = 3,
+        bool reset_m3 = false,
+        bool secondary_bus_reset = true,
+        std::chrono::milliseconds m3_delay = timeout::WARM_RESET_M3_TIMEOUT);
+
+    static bool warm_reset_chip_id(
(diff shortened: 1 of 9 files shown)


device/api/umd/device/warm_reset_with_recovery.hpp
```diff
@@ -0,0 +1,50 @@
...
+// Wraps WarmReset entry points with a follow-up TopologyDiscovery::discover() and retries
+// the whole sequence if discovery fails. This works around transient post-reset states
+// (e.g. an ETH core that never recovers its heartbeat) which can leave the board in a
+// state that only another reset can clear.
+//
+// Each method returns true if a (reset, discovery) attempt eventually succeeds, false
+// otherwise. If the underlying warm reset itself fails, the methods bail immediately
+// without retrying because such failures (no devices, ARM platform) do not recover by
+// retrying.
+class WarmResetWithRecovery {
```
pjanevskiTT, line 24: is this going to become default which clients should call? Just curious, seems like we should still expose both as we do right now

device/warm_reset_with_recovery.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+namespace tt::umd {
+
+namespace {
+
+bool reset_then_discover(int max_attempts, const std::function<bool()>& do_reset) {
+    max_attempts = std::max(max_attempts, 1);
+
+    for (int attempt = 1; attempt <= max_attempts; ++attempt) {
+        log_info(tt::LogUMD, "Warm reset attempt {} of {}.", attempt, max_attempts);
```
pjanevskiTT, line 23: maybe this should be lower level of logging but I don't have hard opinions on it. If we expose this to clients then I would lower it, otherwise it's fine

pjanevskiTT (approved):

device/api/umd/device/warm_reset_with_recovery.hpp
broskoTT: I didn't want to make such decisions. Both are available.
This one relies on TopologyDiscovery, whereas the other does not.
In all our CI, we are calling WarmResetWithRecovery so that the CI doesn't fail.

device/warm_reset_with_recovery.cpp
broskoTT: I find it really useful to track in our CI.
If you called something which does WarmReset with recovery, to me it seems normal to get a log such as this, to let you know which attempts and how many are running

PR #2624, opened 2026-05-13
device/soc_descriptor: fix off-by-one boundary check in serialize_dram_cores

Dev-X25874:
### Issue
No tracked issue — bug found via code audit. The inconsistency is self-evident by comparing line 83 with lines 61 and 93 in the same function.

### Description
`serialize_dram_cores` uses `>` instead of `>=` when checking whether a DRAM core coordinate falls outside the grid, causing a coordinate equal to exactly `grid_size.x` or `grid_size.y` to pass the outer filter while being silently dropped by the inner `<` guard. This results in an empty YAML sequence `[]` being emitted for that DRAM bank — a structurally valid but semantically wrong entry in the serialized descriptor.

Every other boundary comparison in the same function (lines 61 and 93) and across the file correctly uses `<` / `>=`. This was a plain `>` vs `>=` slip.

### List of the changes
- `device/soc_descriptor.cpp` line 83: changed `dram_core.x > grid_size.x` and `dram_core.y > grid_size.y` to `>=` to match the inner guard and the `write_coords` guard, ensuring coordinates at exactly `grid_size` are correctly excluded from serialization.

### Testing
No hardware-specific test required — the fix is verified by code inspection: the corrected condition is now identical in semantics to the inner `if (dram_core.x < grid_size.x && dram_core.y < grid_size.y)` guard on line 93, eliminating the window where `BeginSeq`/`EndSeq` could be emitted with no coordinate written between them. Existing CI covers DRAM descriptor serialization round-trips.

### API Changes


Dev-X25874:
Hey @broskoTT @pjanevskiTT @nbuncicTT @aleksamarkovicTT — just wanted to 
gently ping on this. The PR has been open for 5 days. Would really 
appreciate a review when you get a chance! 🙏

aleksamarkovicTT (approved):

aleksamarkovicTT:
Thank you for the fix.

Issue #2625, opened 2026-05-13
Restore TLB cleanup in RtlSimTlbHandle / TTSimTlbHandle destructors once Metal teardown order is fixed


pjanevskiTT:
### Background
`RtlSimTlbHandle::~RtlSimTlbHandle` previously called `free_tlb()`, which deallocates the TLB index via the `SimulationTlbAllocator*` member.

In some tt-metal teardown paths the `SimulationTlbAllocator` is destroyed before the outstanding `RtlSimTlbHandle` instances, so the destructor dereferences an already-destroyed allocator (use-after-free).

A temporary workaround was applied that defaults `~RtlSimTlbHandle` (no `free_tlb()` call), with a TODO comment explaining the situation.

The equivalent TTSim handle (`TTSimTlbHandle`, `device/pcie/tt_sim_tlb_handle.cpp`) currently uses `= default` for its destructor — same shape, same risk. Cleanup must be added there too once the ordering issue is fixed.

### Action items
- [ ] Fix the destruction order in tt-metal so that `RtlSimTlbHandle` / `TTSimTlbHandle` outlive the `SimulationTlbAllocator`.
- [ ] Restore the `free_tlb()` call in `RtlSimTlbHandle::~RtlSimTlbHandle()` and remove the TODO.
- [ ] Add the corresponding `free_tlb()` invocation in `TTSimTlbHandle::~TTSimTlbHandle()` (currently `= default`).

pjanevskiTT:
fixed by #2627

PR #2626, opened 2026-05-13
Temporarily default ~RtlSimTlbHandle to avoid use-after-free at teardown

pjanevskiTT:
### Issue
Tracking issue: #2625

### Description
`~RtlSimTlbHandle` calls `free_tlb()`, which dereferences the `SimulationTlbAllocator*` member. In tt-metal teardown paths the `SimulationTlbAllocator` can be destroyed before the outstanding `RtlSimTlbHandle` instances, so the destructor ends up calling into an already-destroyed allocator (use-after-free).

This PR is a temporary workaround: default the destructor and leave a TODO explaining why. The cleanup call should be restored once Metal's shutdown sequence is corrected.

The equivalent TTSim handle (`TTSimTlbHandle`) currently uses `= default` for its destructor and has the same risk; that path is tracked in the same follow-up issue (#2625).

### List of the changes
- `device/api/umd/device/pcie/rtl_sim_tlb_handle.hpp`: change `~RtlSimTlbHandle()` to `noexcept override = default` with a TODO comment explaining the workaround.
- `device/pcie/rtl_sim_tlb_handle.cpp`: remove the out-of-line destructor definition (now defaulted in the header).

### Testing

### API Changes

### DIFF
--- a/device/api/umd/device/pcie/rtl_sim_tlb_handle.hpp
+++ b/device/api/umd/device/pcie/rtl_sim_tlb_handle.hpp
@@ -27,7 +27,12 @@ public:
     static std::unique_ptr<RtlSimTlbHandle> create(
         SimulationTlbAllocator* allocator, int tlb_id, size_t size, TlbMapping mapping);
 
-    ~RtlSimTlbHandle() noexcept;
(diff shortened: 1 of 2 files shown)


pjanevskiTT:
closed in favor of #2627

Issue #1673, opened 2025-11-28
Investigate host crashes for bad noc reads


broskoTT:
Explained in https://tenstorrent.slack.com/archives/C07L16Z59AP/p1764342423798049

Reserve tensix wormhole machine. Pick single tensix core (e.g. 0,0 in tt-exalens) and read over noc address: 0xFFB11030. Read will timeout (will return 0xFFFFFF after ~700ms). If you continuing reading from that core you will crash host

Characterise the issue, and try to figure out why it happens, and figure out a solution.

broskoTT:
Nenad and Vuk discussed this in more depth, and @pjanevskiTT is working on an example PR to solve this.
This is a very hard to solve bug.
Discussion needed

broskoTT:
Sync pending next week with exalens crew

pjanevskiTT:
Current state explained in https://github.com/tenstorrent/tt-umd/blob/pjanevski/noc-hang-detection-doc/docs/NOC_HANG_DETECTION.md

pjanevskiTT:
Action items from the meeting (will become issues when we merge initial PR)

- [ ] Use template function instead of std::function
- [ ] For reading from device, try checking just on 0xFF bytes instead of using the timeout
- [ ] Investigate if timeout logic applies also for qb/lb configurations. https://github.com/tenstorrent/tt-exalens/issues/749
- [ ] Check PCIe register for errors (suggested by @nbuncicTT). Needs investigation whether this is something that we can use if just noc is hung, not pcie?

broskoTT:
Ran some claude minions to test perf impact of these changes

broskoTT:
Split PR in multiple smaller ones, pending review + another PR iteration, should be merged soon

nbuncicTT:
Fix merged into main, and unblocked work on recovery for exabox, for now.

PR #2629, opened 2026-05-13
Add per-op MMIO timeout for TLB-mapped device access

pjanevskiTT:
### Issue

#1673

### Description

`memcpy_to_device` / `memcpy_from_device` store/load against TLB-mapped device memory with no liveness check. The existing SIGBUS guard (`SiliconTlbWindow::execute_safe`) only recovers when the platform PCIe completion timeout fires (a mapping/hardware fault), so it misses two cases: **slow degradation** (ops complete, just far slower — the host grinds) and **posted-write pile-up** (writes succeed locally into a dead device before back-pressure stalls the CPU).

This PR adds a **per-op wall-clock budget** to the TLB-touching transfer paths: each transfer constructs an `OpTimeoutGuard` (from #2842) with the per-op budget, an optional liveness veto, and an overrun action that throws `DeviceTimeoutError` (from #2840). On overrun the veto decides:

- **no veto, or veto reports the NOC hung** → throw `DeviceTimeoutError`;
- **veto reports the NOC healthy** → continue; the next op gets a fresh budget.

The veto consumer is `HangDetector`, wired by `TTDevice::set_hang_detector` (called from the Wormhole/Blackhole device constructors). The hang check is stored on the `SiliconTlbWindow` (via `set_io_timeout_hang_check`), which binds the in-flight NOC into the per-op veto. On PCIe the liveness check runs from inside the timed-out transfer, so it reads through a dedicated, separately-locked probe window owned by the injected reader lambda — it never re-takes the `io_lock_` the stalled op holds and never recurses into the timed path. Only single-op stalls are caught; total transfer time is not bounded, and the check cannot preempt a CPU already stalled mid-access (SIGBUS still covers mapping invalidation, e.g. on reset).

The default budget is `timeout::MMIO_OP_TIMEOUT` (**10 ms**, headroom over the multi-ms single-op latency seen on contended/virtualized hosts) and is configurable at runtime via `MmioTimeoutConfig::set_op_timeout`, also exposed to Python as `tt_umd.MmioTimeoutConfig`. The budget bounds a single op (a bulk block, finer in the tails), not the whole transfer.

### List of the changes

- **`MmioTimeoutConfig`** — runtime-configurable per-op budget (default `timeout::MMIO_OP_TIMEOUT` = 10 ms), exposed to Python; non-positive values disable the check.
- **device_memcpy** — `memcpy_to_device` / `memcpy_from_device` and the `write16`/`write32`/`read16`/`read32` scalar helpers take an optional `std::function<bool()> on_timeout`; a `make_mmio_timeout_guard` helper builds the `OpTimeoutGuard` whose overrun action throws `DeviceTimeoutError`. Existing callers compile unchanged.
- **TlbWindow** — new `set_io_timeout_hang_check`; `SiliconTlbWindow` stores the hang check, builds the per-op veto from it, and rebuilds on config changes. `LocalChip` installs it on its cached WC/UC windows and `PcieProtocol` on its cached window.
- **PcieProtocol / TTDevice** — `PcieInterface::set_io_timeout_callback` forwards the hang check to the cached window (no side-channel member); `TTDevice::set_hang_detector` wires the callback and, on PCIe, hands the hang detector a dedicated, separately-locked probe window owned by the injected lambda (with a null-detector guard).
- **Tests** — `tests/api/test_hang_detection.cpp` exercises the timeout on a hung NOC; `test_tlb` / `test_tlb_manager` drive TLB windows with no veto wired, so they widen the budget via `MmioTimeoutConfig` for each test and restore the default afterwards.

### Testing

CI

### API Changes

- New optional `on_timeout` (`std::function<bool()>`) parameter on `memcpy_to_device` / `memcpy_from_device` and the `write16`/`write32`/`read16`/`read32` device scalar helpers — existing call sites compile unchanged.
- New `TlbWindow::set_io_timeout_hang_check` and `PcieInterface::set_io_timeout_callback`.
- New `tt::umd::MmioTimeoutConfig` (`set_op_timeout` / `get_op_timeout`), also exposed to Python as `tt_umd.MmioTimeoutConfig`; default `tt::umd::timeout::MMIO_OP_TIMEOUT` = 10 ms.

Note: uses `OpTimeoutGuard` from #2842 and `DeviceTimeoutError` from #2840, both already on `main`.

### DIFF
--- a/device/api/umd/device/chip/local_chip.hpp
+++ b/device/api/umd/device/chip/local_chip.hpp
@@ -107,10 +107,11 @@ private:
     TlbWindow* get_cached_wc_tlb_window();
     TlbWindow* get_cached_uc_tlb_window();
 
-    // Builds the per-op MMIO timeout veto for TLB-mapped transfers: on an overrun the configured hang
-    // check is consulted for the in-flight op's NOC. A hung NOC confirms the timeout (aborts with
-    // DeviceTimeoutError); a healthy NOC vetoes it so a slow-but-completing op is not a false positive.
-    std::function<bool()> make_io_timeout_callback();
+    // Builds the per-op MMIO timeout hang check installed on the cached TLB windows: given the in-flight
+    // op's NOC, returns whether that NOC is hung. SiliconTlbWindow turns this into the OpTimeoutGuard
+    // false-alarm veto (a hung NOC confirms the timeout; a healthy NOC lets a slow-but-completing op
+    // through). Reads through an independent, separately-locked window so it does not re-enter the TLB lock.
+    std::function<bool(NocId)> make_io_timeout_hang_check();
 
     std::unique_ptr<TlbWindow> cached_wc_tlb_window = nullptr;
     std::unique_ptr<TlbWindow> cached_uc_tlb_window = nullptr;
--- a/device/api/umd/device/chip/local_chip.hpp
+++ b/device/api/umd/device/chip/local_chip.hpp
@@ -6,6 +6,7 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <functional>
 #include <memory>
 #include <mutex>
 #include <optional>
@@ -105,6 +106,11 @@ private:
 
     TlbWindow* get_cached_wc_tlb_window();
     TlbWindow* get_cached_uc_tlb_window();
+
+    // Builds the per-op MMIO timeout veto for TLB-mapped transfers: on an overrun the configured hang
+    // check is consulted for the in-flight op's NOC. A hung NOC confirms the timeout (aborts with
+    // DeviceTimeoutError); a healthy NOC vetoes it so a slow-but-completing op is not a false positive.
+    std::function<bool()> make_io_timeout_callback();
 
     std::unique_ptr<TlbWindow> cached_wc_tlb_window = nullptr;
     std::unique_ptr<TlbWindow> cached_uc_tlb_window = nullptr;
--- a/device/api/umd/device/chip/local_chip.hpp
+++ b/device/api/umd/device/chip/local_chip.hpp
@@ -107,10 +107,6 @@ private:
     TlbWindow* get_cached_wc_tlb_window();
     TlbWindow* get_cached_uc_tlb_window();
 
-    // Builds the per-op MMIO timeout hang check installed on the cached TLB windows: given the in-flight
-    // op's NOC, returns whether that NOC is hung. SiliconTlbWindow turns this into the OpTimeoutGuard
-    // false-alarm veto (a hung NOC confirms the timeout; a healthy NOC lets a slow-but-completing op
-    // through). Reads through an independent, separately-locked window so it does not re-enter the TLB lock.
     std::function<bool(NocId)> make_io_timeout_hang_check();
 
     std::unique_ptr<TlbWindow> cached_wc_tlb_window = nullptr;
--- a/device/api/umd/device/pcie/silicon_tlb_window.hpp
+++ b/device/api/umd/device/pcie/silicon_tlb_window.hpp
@@ -30,8 +30,9 @@ public:
     uint16_t read16(uint64_t offset, const std::function<bool()>& on_timeout = {}) override;
     void write32(uint64_t offset, uint32_t value, const std::function<bool()>& on_timeout = {}) override;
     uint32_t read32(uint64_t offset, const std::function<bool()>& on_timeout = {}) override;
-    void write_register(uint64_t offset, const void* data, size_t size) override;
-    void read_register(uint64_t offset, void* data, size_t size) override;
+    void write_register(
+        uint64_t offset, const void* data, size_t size, const std::function<bool()>& on_timeout = {}) override;
+    void read_register(uint64_t offset, void* data, size_t size, const std::function<bool()>& on_timeout = {}) override;
     void write_block(
         uint64_t offset, const void* data, size_t size, const std::function<bool()>& on_timeout = {}) override;
     void read_block(uint64_t offset, void* data, size_t size, const std::function<bool()>& on_timeout = {}) override;
@@ -109,8 +110,9 @@ private:
     static void memcpy_to_device(
         void* dest, const void* src, std::size_t num_bytes, const std::function<bool()>& on_timeout);
 
-    void write_regs(volatile uint32_t* dest, const uint32_t* src, uint32_t word_len);
-    void read_regs(void* src_reg, uint32_t word_len, void* data);
+    void write_regs(
+        volatile uint32_t* dest, const uint32_t* src, uint32_t word_len, const std::function<bool()>& on_timeout = {});
+    void read_regs(void* src_reg, uint32_t word_len, void* data, const std::function<bool()>& on_timeout = {});
 
     template <typename Func, typename... Args>
     decltype(auto) execute_safe(Func&& func, Args&&... args);
--- a/device/api/umd/device/pcie/silicon_tlb_window.hpp
+++ b/device/api/umd/device/pcie/silicon_tlb_window.hpp
@@ -91,14 +91,18 @@ public:
     // When empty (the default), an overrun aborts outright.
     void set_io_timeout_hang_check(const std::function<bool(NocId)>& hang_check) override;
 
+    // Reconfigures the window and refreshes the cached timeout callback, since the NOC may have changed.
+    void configure(const tlb_data& new_config) override;
+
     static void set_sigbus_safe_handler(bool set_safe_handler);
 
 private:
-    // Builds the per-op timeout veto (an OpTimeoutGuard is_false_alarm callback) for the window's
-    // currently configured NOC: true means "healthy, false positive" so the overrun is ignored, false
-    // confirms it. Reads the NOC from the live TLB config, so it tracks whatever the last (re)configure
-    // selected. Returns an empty callback when no hang check is wired, which makes an overrun abort.
-    std::function<bool()> make_io_timeout_callback() const;
+    // Rebuilds io_timeout_callback_ (an OpTimeoutGuard is_false_alarm callback) for the window's currently
+    // configured NOC: true means "healthy, false positive" so the overrun is ignored, false confirms it.
+    // The NOC is read from the live TLB config, so this must be called whenever the config or hang check
+    // changes (construction, configure(), set_io_timeout_hang_check()). Leaves an empty callback when no
+    // hang check is wired, which makes an overrun abort.
+    void update_io_timeout_callback();
 
     // Custom device memcpy. This is only safe for memory-like regions on the device (Tensix L1, DRAM, ARC CSM).
     // Both routines assume that misaligned accesses are permitted on host memory.
@@ -120,6 +124,9 @@ private:
     decltype(auto) execute_safe(Func&& func, Args&&... args);
 
     std::function<bool(NocId)> hang_check_;
+    // Cached is_false_alarm callback bound to the currently configured NOC; refreshed by
+    // update_io_timeout_callback() so the IO paths need not rebuild it per call.
+    std::function<bool()> io_timeout_callback_;
 };
 
 }  // namespace tt::umd
--- a/device/api/umd/device/pcie/silicon_tlb_window.hpp
+++ b/device/api/umd/device/pcie/silicon_tlb_window.hpp
@@ -24,12 +24,7 @@ class TlbHandle;
  */
 class SiliconTlbWindow : public TlbWindow {
 public:
-    // hang_check (optional) wires the per-op MMIO timeout: on a single-op overrun it is consulted for the
-    // window's currently configured NOC. Returning true means that NOC is hung (the transfer aborts with
-    // DeviceTimeoutError); false means it is healthy (a slow-but-completing op is not a false positive).
-    // When empty, an overrun aborts outright. See make_io_timeout_callback().
-    SiliconTlbWindow(
-        std::unique_ptr<TlbHandle> handle, const tlb_data config = {}, std::function<bool(NocId)> hang_check = {});
+    SiliconTlbWindow(std::unique_ptr<TlbHandle> handle, const tlb_data config = {});
 
     // Implementation of memory access methods using direct pointer access.
     void write16(uint64_t offset, uint16_t value) override;
@@ -90,6 +85,10 @@ public:
         NocId noc_id,
         uint64_t ordering = tlb_data::Strict) override;
 
+    // Wires the per-op MMIO timeout: on a single-op overrun hang_check is consulted for the window's
+    // currently configured NOC. Returning true means that NOC is hung (the transfer aborts with
+    // DeviceTimeoutError); false means it is healthy (a slow-but-completing op is not a false positive).
+    // When empty (the default), an overrun aborts outright.
     void set_io_timeout_hang_check(const std::function<bool(NocId)>& hang_check) override;
 
     static void set_sigbus_safe_handler(bool set_safe_handler);
--- a/device/api/umd/device/pcie/silicon_tlb_window.hpp
+++ b/device/api/umd/device/pcie/silicon_tlb_window.hpp
@@ -6,6 +6,7 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <functional>
 #include <memory>
 
 #include "umd/device/pcie/tlb_handle.hpp"
@@ -25,30 +26,33 @@ public:
     SiliconTlbWindow(std::unique_ptr<TlbHandle> handle, const tlb_data config = {});
 
     // Implementation of memory access methods using direct pointer access.
-    void write16(uint64_t offset, uint16_t value) override;
-    uint16_t read16(uint64_t offset) override;
-    void write32(uint64_t offset, uint32_t value) override;
-    uint32_t read32(uint64_t offset) override;
+    void write16(uint64_t offset, uint16_t value, const std::function<bool()>& on_timeout = {}) override;
+    uint16_t read16(uint64_t offset, const std::function<bool()>& on_timeout = {}) override;
+    void write32(uint64_t offset, uint32_t value, const std::function<bool()>& on_timeout = {}) override;
+    uint32_t read32(uint64_t offset, const std::function<bool()>& on_timeout = {}) override;
     void write_register(uint64_t offset, const void* data, size_t size) override;
     void read_register(uint64_t offset, void* data, size_t size) override;
-    void write_block(uint64_t offset, const void* data, size_t size) override;
-    void read_block(uint64_t offset, void* data, size_t size) override;
+    void write_block(
+        uint64_t offset, const void* data, size_t size, const std::function<bool()>& on_timeout = {}) override;
+    void read_block(uint64_t offset, void* data, size_t size, const std::function<bool()>& on_timeout = {}) override;
 
-    void safe_write16(uint64_t offset, uint16_t value) override;
+    void safe_write16(uint64_t offset, uint16_t value, const std::function<bool()>& on_timeout = {}) override;
 
-    uint16_t safe_read16(uint64_t offset) override;
+    uint16_t safe_read16(uint64_t offset, const std::function<bool()>& on_timeout = {}) override;
 
-    void safe_write32(uint64_t offset, uint32_t value) override;
+    void safe_write32(uint64_t offset, uint32_t value, const std::function<bool()>& on_timeout = {}) override;
 
-    uint32_t safe_read32(uint64_t offset) override;
+    uint32_t safe_read32(uint64_t offset, const std::function<bool()>& on_timeout = {}) override;
 
     void safe_write_register(uint64_t offset, const void* data, size_t size) override;
 
     void safe_read_register(uint64_t offset, void* data, size_t size) override;
 
-    void safe_write_block(uint64_t offset, const void* data, size_t size) override;
+    void safe_write_block(
+        uint64_t offset, const void* data, size_t size, const std::function<bool()>& on_timeout = {}) override;
 
-    void safe_read_block(uint64_t offset, void* data, size_t size) override;
+    void safe_read_block(
+        uint64_t offset, void* data, size_t size, const std::function<bool()>& on_timeout = {}) override;
 
     void safe_write_block_reconfigure(
         const void* mem_ptr,
@@ -56,11 +60,17 @@ public:
         uint64_t addr,
         size_t size,
         NocId noc_id,
-        uint64_t ordering = tlb_data::Strict) override;
+        uint64_t ordering = tlb_data::Strict,
+        const std::function<bool()>& on_timeout = {}) override;
 
     void safe_read_block_reconfigure(
-        void* mem_ptr, tt_xy_pair core, uint64_t addr, size_t size, NocId noc_id, uint64_t ordering = tlb_data::Strict)
-        override;
+        void* mem_ptr,
+        tt_xy_pair core,
+        uint64_t addr,
+        size_t size,
+        NocId noc_id,
+        uint64_t ordering = tlb_data::Strict,
+        const std::function<bool()>& on_timeout = {}) override;
 
     void safe_read_register_reconfigure(
         void* mem_ptr, tt_xy_pair core, uint64_t addr, size_t size, NocId noc_id, uint64_t ordering = tlb_data::Strict)
@@ -81,7 +91,8 @@ public:
         tt_xy_pair core_end,
         uint64_t addr,
         NocId noc_id,
-        uint64_t ordering = tlb_data::Strict) override;
+        uint64_t ordering = tlb_data::Strict,
+        const std::function<bool()>& on_timeout = {}) override;
 
     static void set_sigbus_safe_handler(bool set_safe_handler);
 
@@ -93,8 +104,10 @@ private:
     // which glibc's memcpy may perform when unrolling. This affects from and to device.
     // 2. syseng#3487 WH GDDR5 controller has a bug when 1-byte writes are temporarily adjacent
     // to 2-byte writes. We avoid ever performing a 1-byte write to the device. This only affects to device.
-    static void memcpy_from_device(void* dest, const volatile void* src, std::size_t num_bytes);
-    static void memcpy_to_device(void* dest, const void* src, std::size_t num_bytes);
+    static void memcpy_from_device(
+        void* dest, const volatile void* src, std::size_t num_bytes, const std::function<bool()>& on_timeout);
+    static void memcpy_to_device(
+        void* dest, const void* src, std::size_t num_bytes, const std::function<bool()>& on_timeout);
 
     void write_regs(volatile uint32_t* dest, const uint32_t* src, uint32_t word_len);
     void read_regs(void* src_reg, uint32_t word_len, void* data);
--- a/device/api/umd/device/pcie/silicon_tlb_window.hpp
+++ b/device/api/umd/device/pcie/silicon_tlb_window.hpp
@@ -11,6 +11,7 @@
 
 #include "umd/device/pcie/tlb_handle.hpp"
 #include "umd/device/pcie/tlb_window.hpp"
+#include "umd/device/types/noc_id.hpp"
 #include "umd/device/types/tlb.hpp"
 #include "umd/device/types/xy_pair.hpp"
 
@@ -23,37 +24,38 @@ class TlbHandle;
  */
 class SiliconTlbWindow : public TlbWindow {
 public:
-    SiliconTlbWindow(std::unique_ptr<TlbHandle> handle, const tlb_data config = {});
+    // hang_check (optional) wires the per-op MMIO timeout: on a single-op overrun it is consulted for the
+    // window's currently configured NOC. Returning true means that NOC is hung (the transfer aborts with
+    // DeviceTimeoutError); false means it is healthy (a slow-but-completing op is not a false positive).
+    // When empty, an overrun aborts outright. See make_io_timeout_callback().
+    SiliconTlbWindow(
+        std::unique_ptr<TlbHandle> handle, const tlb_data config = {}, std::function<bool(NocId)> hang_check = {});
 
     // Implementation of memory access methods using direct pointer access.
-    void write16(uint64_t offset, uint16_t value, const std::function<bool()>& on_timeout = {}) override;
-    uint16_t read16(uint64_t offset, const std::function<bool()>& on_timeout = {}) override;
-    void write32(uint64_t offset, uint32_t value, const std::function<bool()>& on_timeout = {}) override;
-    uint32_t read32(uint64_t offset, const std::function<bool()>& on_timeout = {}) override;
-    void write_register(
-        uint64_t offset, const void* data, size_t size, const std::function<bool()>& on_timeout = {}) override;
-    void read_register(uint64_t offset, void* data, size_t size, const std::function<bool()>& on_timeout = {}) override;
-    void write_block(
-        uint64_t offset, const void* data, size_t size, const std::function<bool()>& on_timeout = {}) override;
-    void read_block(uint64_t offset, void* data, size_t size, const std::function<bool()>& on_timeout = {}) override;
+    void write16(uint64_t offset, uint16_t value) override;
+    uint16_t read16(uint64_t offset) override;
+    void write32(uint64_t offset, uint32_t value) override;
+    uint32_t read32(uint64_t offset) override;
+    void write_register(uint64_t offset, const void* data, size_t size) override;
+    void read_register(uint64_t offset, void* data, size_t size) override;
+    void write_block(uint64_t offset, const void* data, size_t size) override;
+    void read_block(uint64_t offset, void* data, size_t size) override;
 
-    void safe_write16(uint64_t offset, uint16_t value, const std::function<bool()>& on_timeout = {}) override;
+    void safe_write16(uint64_t offset, uint16_t value) override;
 
-    uint16_t safe_read16(uint64_t offset, const std::function<bool()>& on_timeout = {}) override;
+    uint16_t safe_read16(uint64_t offset) override;
 
-    void safe_write32(uint64_t offset, uint32_t value, const std::function<bool()>& on_timeout = {}) override;
+    void safe_write32(uint64_t offset, uint32_t value) override;
 
-    uint32_t safe_read32(uint64_t offset, const std::function<bool()>& on_timeout = {}) override;
+    uint32_t safe_read32(uint64_t offset) override;
 
     void safe_write_register(uint64_t offset, const void* data, size_t size) override;
 
     void safe_read_register(uint64_t offset, void* data, size_t size) override;
 
-    void safe_write_block(
-        uint64_t offset, const void* data, size_t size, const std::function<bool()>& on_timeout = {}) override;
+    void safe_write_block(uint64_t offset, const void* data, size_t size) override;
 
-    void safe_read_block(
-        uint64_t offset, void* data, size_t size, const std::function<bool()>& on_timeout = {}) override;
+    void safe_read_block(uint64_t offset, void* data, size_t size) override;
 
     void safe_write_block_reconfigure(
         const void* mem_ptr,
@@ -61,17 +63,11 @@ public:
         uint64_t addr,
         size_t size,
         NocId noc_id,
-        uint64_t ordering = tlb_data::Strict,
-        const std::function<bool()>& on_timeout = {}) override;
+        uint64_t ordering = tlb_data::Strict) override;
 
     void safe_read_block_reconfigure(
-        void* mem_ptr,
-        tt_xy_pair core,
-        uint64_t addr,
-        size_t size,
-        NocId noc_id,
-        uint64_t ordering = tlb_data::Strict,
-        const std::function<bool()>& on_timeout = {}) override;
+        void* mem_ptr, tt_xy_pair core, uint64_t addr, size_t size, NocId noc_id, uint64_t ordering = tlb_data::Strict)
+        override;
 
     void safe_read_register_reconfigure(
         void* mem_ptr, tt_xy_pair core, uint64_t addr, size_t size, NocId noc_id, uint64_t ordering = tlb_data::Strict)
@@ -92,12 +88,19 @@ public:
         tt_xy_pair core_end,
         uint64_t addr,
         NocId noc_id,
-        uint64_t ordering = tlb_data::Strict,
-        const std::function<bool()>& on_timeout = {}) override;
+        uint64_t ordering = tlb_data::Strict) override;
+
+    void set_io_timeout_hang_check(const std::function<bool(NocId)>& hang_check) override;
 
     static void set_sigbus_safe_handler(bool set_safe_handler);
 
 private:
+    // Builds the per-op timeout veto (an OpTimeoutGuard is_false_alarm callback) for the window's
+    // currently configured NOC: true means "healthy, false positive" so the overrun is ignored, false
+    // confirms it. Reads the NOC from the live TLB config, so it tracks whatever the last (re)configure
+    // selected. Returns an empty callback when no hang check is wired, which makes an overrun abort.
+    std::function<bool()> make_io_timeout_callback() const;
+
     // Custom device memcpy. This is only safe for memory-like regions on the device (Tensix L1, DRAM, ARC CSM).
     // Both routines assume that misaligned accesses are permitted on host memory.
     //
@@ -111,11 +114,13 @@ private:
         void* dest, const void* src, std::size_t num_bytes, const std::function<bool()>& on_timeout);
 
     void write_regs(
-        volatile uint32_t* dest, const uint32_t* src, uint32_t word_len, const std::function<bool()>& on_timeout = {});
-    void read_regs(void* src_reg, uint32_t word_len, void* data, const std::function<bool()>& on_timeout = {});
+        volatile uint32_t* dest, const uint32_t* src, uint32_t word_len, const std::function<bool()>& on_timeout);
+    void read_regs(void* src_reg, uint32_t word_len, void* data, const std::function<bool()>& on_timeout);
 
     template <typename Func, typename... Args>
     decltype(auto) execute_safe(Func&& func, Args&&... args);
+
+    std::function<bool(NocId)> hang_check_;
 };
 
 }  // namespace tt::umd
--- a/device/api/umd/device/tt_device/protocol/pcie_interface.hpp
+++ b/device/api/umd/device/tt_device/protocol/pcie_interface.hpp
@@ -49,9 +49,8 @@ public:
     virtual void bar_write32(uint32_t addr, uint32_t data) = 0;
     virtual uint32_t bar_read32(uint32_t addr) = 0;
 
-    // Hook for the timed MMIO path: invoked (with the in-flight op's NOC) when a single op exceeds its
-    // per-op budget. Returns true if that NOC is confirmed hung (abort the transfer with
-    // DeviceTimeoutError), false to treat the slow op as a false positive and continue.
+    // Sets the hang check invoked on an IO-op timeout: returns true if the in-flight NOC is hung (abort),
+    // false to continue.
     virtual void set_io_timeout_callback(const std::function<bool(NocId)>& hang_check) = 0;
 };
 
--- a/device/api/umd/device/tt_device/protocol/pcie_interface.hpp
+++ b/device/api/umd/device/tt_device/protocol/pcie_interface.hpp
@@ -7,6 +7,7 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <functional>
 
 #include "umd/device/types/noc_id.hpp"
 #include "umd/device/types/xy_pair.hpp"
@@ -47,6 +48,11 @@ public:
 
     virtual void bar_write32(uint32_t addr, uint32_t data) = 0;
     virtual uint32_t bar_read32(uint32_t addr) = 0;
+
+    // Hook for the timed MMIO path: invoked (with the in-flight op's NOC) when a single op exceeds its
+    // per-op budget. Returns true if that NOC is confirmed hung (abort the transfer with
+    // DeviceTimeoutError), false to treat the slow op as a false positive and continue.
+    virtual void set_io_timeout_callback(const std::function<bool(NocId)>& hang_check) = 0;
 };
 
 }  // namespace tt::umd
--- a/device/api/umd/device/tt_device/protocol/pcie_protocol.hpp
+++ b/device/api/umd/device/tt_device/protocol/pcie_protocol.hpp
@@ -99,9 +99,8 @@ private:
     std::unique_ptr<TlbWindow> cached_tlb_window_;
     std::unique_ptr<TlbWindow> cached_dma_tlb_window_;
 
-    // Hang check forwarded per-call to the cached window's timed memcpy path. The in-flight op's NOC is
-    // bound into a per-call lambda at each I/O call site, so the check probes the right NOC. Empty until a
-    // HangDetector is wired in (see TTDevice::set_hang_detector); while empty, an overrun throws.
+    // Hang check consulted on an IO-op timeout; empty until a HangDetector is wired in (see
+    // TTDevice::set_hang_detector).
     std::function<bool(NocId)> hang_check_;
 };
 
--- a/device/api/umd/device/tt_device/protocol/pcie_protocol.hpp
+++ b/device/api/umd/device/tt_device/protocol/pcie_protocol.hpp
@@ -7,6 +7,7 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <functional>
 #include <memory>
 #include <mutex>
 #include <optional>
@@ -50,6 +51,7 @@ public:
     int get_mmio_id() override;
 
     // PcieInterface.
+    void set_io_timeout_callback(const std::function<bool(NocId)>& hang_check) override;
     PCIDevice* get_pci_device() override;
     [[nodiscard]] bool dma_write_to_device(
         const void* src, size_t size, tt_xy_pair core, uint64_t addr, NocId noc_id) override;
@@ -96,6 +98,11 @@ private:
     std::mutex dma_mutex_;
     std::unique_ptr<TlbWindow> cached_tlb_window_;
     std::unique_ptr<TlbWindow> cached_dma_tlb_window_;
+
+    // Hang check forwarded per-call to the cached window's timed memcpy path. The in-flight op's NOC is
+    // bound into a per-call lambda at each I/O call site, so the check probes the right NOC. Empty until a
+    // HangDetector is wired in (see TTDevice::set_hang_detector); while empty, an overrun throws.
+    std::function<bool(NocId)> hang_check_;
 };
 
 }  // namespace tt::umd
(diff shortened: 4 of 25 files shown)


pjanevskiTT:
## Perf comparison vs latest main

CI run [`25851022729`](https://github.com/tenstorrent/tt-umd/actions/runs/25851022729) (this PR, head `664a2c43`) vs main run [`25837720472`](https://github.com/tenstorrent/tt-umd/actions/runs/25837720472) (sha `51d15f18`). Aggregate over all `test-all` jobs that succeeded on both.

**TL;DR: no real perf impact. Aggregate is −0.3 %; the per-runner and per-test spread you see below is dominated by runner / single-sample noise, not by anything this PR is doing.**

Pattern check: deltas swing both directions (some runners and tests are faster on the PR run, some slower), and the runners that diverge most are not the ones that exercise the timeout path hardest. `SysmemReadWrite` for example doesn't traverse `SiliconTlbWindow::write_block`/`read_block` at all and still shows ±tens of ms across runs — that's the noise floor.

### Per-runner IO-test totals

| Runner | main (s) | PR (s) | Δ |
|---|---:|---:|---:|
| BH p100a-viommu | 25.39 | 24.42 | −3.8 % |
| BH p150b-stable | 15.56 | 15.46 | −0.6 % |
| BH p150b-viommu | 20.03 | 19.33 | −3.5 % |
| WH 6u | 50.45 | 51.39 | +1.9 % |
| WH n150-stable | 6.88 | 6.68 | −2.9 % |
| WH n150-viommu | 12.55 | 13.24 | +5.5 % |
| WH n300-llmbox-stable | 22.36 | 21.98 | −1.7 % |
| WH n300-llmbox-viommu | 50.06 | 49.83 | −0.5 % |
| WH n300-stable | 8.55 | 6.93 | −19.0 % |
| WH n300-viommu | 15.47 | 17.43 | +12.7 % |
| **GRAND TOTAL** | **227.30** | **226.68** | **−0.3 %** |

### Per-test deltas (aggregated across runners)

| main (ms) | PR (ms) | Δ | Test |
|---:|---:|---:|---|
| 6000 | 5725 | −4.6 % | `SiliconDriverBH.MultiThreadedMemBar` |
| 5940 | 5669 | −4.6 % | `TestDeviceIOFixture.WriteDataReadReg` |
| 16690 | 15970 | −4.3 % | `TestDeviceIOFixture.RegReadWrite` |
| 15320 | 15744 | +2.8 % | `SiliconDriverWH.HarvestingRuntime` |
| 24979 | 25347 | +1.5 % | `SiliconDriverWH.UnalignedStaticTLB_RW` |
| 8069 | 7934 | −1.7 % | `ClusterReadWriteL1Test.ReadWriteL1/Silicon` |
| 7616 | 7729 | +1.5 % | `DynamicTLB_RW/DRAM` |
| 14884 | 14820 | −0.4 % | `DynamicTLB_RW/TENSIX` |
| 10701 | 10738 | +0.3 % | `SiliconDriverWH.MultiThreadedMemBar` |
| 18326 | 18339 | +0.1 % | `SiliconDriverWH.CreateDestroy` |
| 29868 | 29834 | −0.1 % | `SiliconDriverBH.CreateDestroy` |
| 583 | 647 | +11.0 % | `SiliconDriverWH.StaticTLB_RW` (~60 ms abs) |
| 621 | 664 | +6.9 % | `SiliconDriverWH.DynamicTLB_RW` (~65 ms abs) |

The double-digit-% movers are sub-100 ms tests where small absolute jitter shows up as large relative numbers.

pjanevskiTT:
On bh-40 dedicated machine slowdown is from 3m13.660s to 3m27.312s for API tests

device/pcie/device_memcpy.cpp
```diff
@@ -4,35 +4,177 @@
...
 namespace tt::umd {
 
-void memcpy_to_device(volatile void* dest, const void* src, std::size_t size) {
+namespace {
+
+// Hard-coded default per-op budget; overridable at process start via the env var
+// TT_UMD_MMIO_OP_TIMEOUT_MS. Applied to every TLB-touching store/load. Set to 30 ms
+// so post-reset reads (which can legitimately take >5 ms before the device settles)
+// don't trip the timeout on the happy path.
+constexpr std::chrono::milliseconds kDefaultMmioOpTimeout{100};
```
tt-vjovanovic: Note that after thorough testing @adjordjevic-TT set default timeout for MMIO in tt-exalens to `2ms` for reading/writing 4 bytes.

device/pcie/device_memcpy.cpp
```diff
@@ -4,35 +4,177 @@
...
+bool mmio_timing_enabled() {
+    static const bool value = [] {
+        const char* env = std::getenv("TT_UMD_MEMCPY_TIMING");
+        return env != nullptr && *env != '\0' && env[0] != '0';
+    }();
+    return value;
+}
+
+struct OpTiming {
+    std::int64_t delta_ns;
```
tt-vjovanovic: Use `std::chrono` for type and then you don't need to add type into name.

device/pcie/device_memcpy.hpp
```diff
@@ -5,9 +5,28 @@
...
+ * Typical use is a NOC hang check: probe the device cheaply and return
+ * "is this op stalled because the device is actually hung?".
+ *
+ * The implementation must not recurse into the same memcpy / I/O path that
+ * triggered it. A thread-local re-entrancy guard inside memcpy short-circuits
+ * one level of recursion (the inner timeout will simply throw without
+ * re-invoking the callback), but composability and lock-ordering across
+ * the I/O stack are the caller's responsibility.
+ */
+using MemcpyTimeoutFn = std::function<bool()>;
```
tt-vjovanovic: If you notice bigger perf impact consider using template function argument instead of `std::function<>` and avoid relying on optimizer being smart to inline everything. Mostly problem are heap allocations that happen because of using `std::function<>`.

device/pcie/device_memcpy.cpp
```diff
@@ -81,39 +233,87 @@ void memcpy_to_device(volatile void* dest, const void* src, std::size_t size) {
...
+        auto delta_ms = std::chrono::duration_cast<std::chrono::milliseconds>(delta).count();
+        auto budget_ms = op_timeout.count();
+        throw error::DeviceTimeoutError(
+            "memcpy_from_device per-op timeout: " + std::to_string(op_bytes) + "B load took " +
+            std::to_string(delta_ms) + " ms (budget=" + std::to_string(budget_ms) + " ms), " + std::to_string(size) +
+            " of " + std::to_string(original_size) + " bytes remaining.");
+    };
+
     // Phase 0: Align device source to 4 bytes using byte-wide volatile loads.
     while (size > 0 && (reinterpret_cast<std::uintptr_t>(s) % 4) != 0) {
```
tt-vjovanovic, line 286: Not sure if optimizer is smart enough, but you can calculate `%` only once instead of looping through it up to 3 times...

device/pcie/device_memcpy.cpp
tt-vjovanovic, line 414:
```suggestion
#else
    // TODO: Care about alignment, in phase 0 you would align to 8 bytes!

    // Speeding up non x86 code
    while (size >= 8) {
        auto t = std::chrono::steady_clock::now();
        std::uint64_t tmp = *reinterpret_cast<const volatile std::uint64_t*>(s);
        record_and_check(t, 8);
        std::memcpy(d, &tmp, sizeof(tmp));
        d += 8;
        s += 8;
        size -= 8;
    }#endif
```

device/pcie/device_memcpy.cpp
```diff
@@ -4,35 +4,177 @@
...
+struct OpTiming {
+    std::int64_t delta_ns;
+    std::uint32_t size_bytes;
+};
+
+// Re-entrancy guard: if a MemcpyTimeoutFn callback itself ends up issuing MMIO
+// that triggers a per-op timeout, the inner timeout must short-circuit (don't
+// invoke the callback recursively — that would loop). The inner record_and_check
+// sees the flag and treats the inner timeout as a confirmed abort.
+thread_local bool in_timeout_callback = false;
```
tt-vjovanovic: I don't think this is something you want to solve in your API function. User can provide/omit `on_timeout` callback. In other words, user code that issues "regular" read would have `on_timeout` set that when hit issues read that doesn't have `on_timeout` set...

device/pcie/device_memcpy.cpp
```diff
@@ -4,35 +4,177 @@
...
+            return;
+        }
+        // Per-op budget exceeded. Consult on_timeout if available and we're not
+        // already inside the callback (which would recurse).
+        if (on_timeout && !in_timeout_callback) {
+            ScopedTimeoutCallbackGuard guard;
+            if (!on_timeout()) {
+                return;  // false positive — device healthy, continue
+            }
+        }
```
tt-vjovanovic: When you remove `in_timeout_callback` guard, I think this should be changed to:
```
if (!on_timeout || !on_timeout()) {
    return;
}
```

device/pcie/device_memcpy.cpp
```diff
@@ -4,35 +4,177 @@
...
+        // Either no on_timeout provided, we're already inside one (skip recursion),
+        // or on_timeout confirmed the abort. Throw.
+        auto delta_ms = std::chrono::duration_cast<std::chrono::milliseconds>(delta).count();
+        auto budget_ms = op_timeout.count();
+        throw error::DeviceTimeoutError(
+            "memcpy_to_device per-op timeout: " + std::to_string(op_bytes) + "B store took " +
+            std::to_string(delta_ms) + " ms (budget=" + std::to_string(budget_ms) + " ms), " + std::to_string(size) +
+            " of " + std::to_string(original_size) + " bytes remaining.");
```
tt-vjovanovic: You probably want to save arguments in exception object and do string formatting in `what()` override...

device/pcie/device_memcpy.cpp
```diff
@@ -4,35 +4,177 @@
...
+    const auto op_timeout = mmio_op_timeout();
+    const bool timing_enabled = mmio_timing_enabled();
+    thread_local std::vector<OpTiming> timings;
+    if (timing_enabled) {
+        timings.clear();
+    }
+
+    // Called after every TLB-touching store. Single now() shared between
+    // timing record and the per-op budget check.
+    auto record_and_check = [&](std::chrono::steady_clock::time_point t_before, std::uint32_t op_bytes) {
```
tt-vjovanovic: To avoid code duplication (having this same lambda in two functions), consider creating simple class that can do this:
- in constructor you evaluate `op_timeout` and `timing_enabled`
- you have simple `record_and_check` function that does the same

device/pcie/device_memcpy.cpp
pjanevskiTT: measured locally and it doesn't seem to have any perf impact at the moment and I think the code is more readable this way

device/pcie/device_memcpy.cpp
pjanevskiTT: Tracked in #2690 — will land as a follow-up so this PR stays focused on timeout enforcement.

broskoTT:
@pjanevskiTT lets discuss this in greater detail so we can solve it the right way, please organise a brainstorming sync with UMD + VukJ at least, before continuing development on this

device/pcie/device_memcpy.cpp
pjanevskiTT: removed

pjanevskiTT:
**Benchmark: `std::function` vs templated timeout callback — Wormhole n150**

`api_tests` on a single Wormhole n150 (`wormhole_b0`), three branches on the same latest-`main` base, 3 runs each. gtest total ms (median); Δ% vs baseline. **Heavy-IO** = `test_device_io.cpp` IO tests; **Stress** = `DISABLED_LoopbackStressSize` (1 MB L1 + 256 MB DRAM sweep).

| Branch | Full suite | Δ% | Heavy-IO | Δ% | Stress | Δ% |
|---|---|---|---|---|---|---|
| baseline (`main`, no timeout) | 42060 ms | — | 2544 ms | — | 16022 ms | — |
| `std::function` (this PR) | 42157 ms | +0.23% | 2645 ms | +3.97% | 16066 ms | +0.27% |
| templated | 42163 ms | +0.24% | 2601 ms | +2.24% | 16038 ms | +0.10% |

Heavy-IO baseline runs spanned 2490–2617 ms (±2.5%). On every set, templated ≤ `std::function`: stress −0.17 pp, heavy-IO −1.73 pp, full ≈0.

tt-vjovanovic:
> The timeout callback's overhead is negligible. On the memcpy-bound stress test the `std::function` version costs +0.27% and templating trims it to +0.10% — a consistent but tiny win, since the 256 MB transfers are bandwidth-bound and dwarf per-op dispatch cost. The heavy-IO gap (+4.0% / +2.2%) is within run-to-run noise (baseline's own three runs span 127 ms there). The full suite is flat (+0.2%). Templating doesn't buy a meaningful speedup, so it isn't worth the added complexity.

I think we should go with `std::function<>` at the moment and switch to templates only if we need to (maybe make an issue not to forget that as an option).

pjanevskiTT:
**Benchmark: `std::function` vs templated timeout callback — Blackhole p150**

`api_tests` on a single Blackhole p150, three branches on the same latest-`main` base, 3 runs each. gtest total ms (median); Δ% vs baseline. **Heavy-IO** = `test_device_io.cpp` IO tests excluding `DMA2` (bandwidth-bound, thermal-drift outlier); **Stress** = `DISABLED_LoopbackStressSize` (1 MB L1 + 256 MB DRAM sweep).

| Branch | Full suite | Δ% | Heavy-IO (no DMA2) | Δ% | Stress | Δ% |
|---|---|---|---|---|---|---|
| baseline (`main`, no timeout) | 196453 ms | — | 5591 ms | — | 12374 ms | — |
| `std::function` (this PR) | 198065 ms | +0.82% | 5692 ms | +1.81% | 12546 ms | +1.39% |
| templated | 198040 ms | +0.81% | 5590 ms | −0.02% | 12525 ms | +1.22% |

On every set, templated ≤ `std::function`: heavy-IO −1.83 pp (templated returns to baseline), stress −0.17 pp, full ≈0.

device/api/umd/device/utils/error.hpp
```diff
@@ -42,4 +44,78 @@ class SigbusError : public std::runtime_error {
...
+ * piling up against a slow or dead NOC before SIGBUS could fire.
+ *
+ * Fields are stored as-is and the formatted what() string is built lazily
+ * on first call. The exception does NOT copy function_name and operation:
+ * callers MUST pass pointers with static storage duration (string literals).
+ * Passing a temporary's c_str() (e.g. std::string::c_str()) leaves dangling
+ * pointers inside the exception. The memcpy path passes literals
+ * ("memcpy_to_device" / "store" etc.).
+ */
+class DeviceTimeoutError : public std::exception {
```
aleksamarkovicTT: Adapt this to use UmdError, see errors in tt_device_error.hpp for examples.

device/pcie/device_memcpy.cpp
```diff
@@ -4,51 +4,170 @@
...
 
-void write16_to_device(volatile void* dest, std::uint16_t value) {
+namespace {
+
+// Hard-coded default per-op budget; overridable at process start via the env var
+// TT_UMD_MMIO_OP_TIMEOUT_MS. Applied once per 256-byte block in the bulk AVX2
+// phase and once per op in the 32 / 16 / 4-byte and byte-wide tail phases. Set to
+// 100 ms so post-reset reads (which can legitimately take tens of ms before the
+// device settles) don't trip the timeout on the happy path.
+constexpr std::chrono::milliseconds kDefaultMmioOpTimeout{100};
```
aleksamarkovicTT: This should be moved to timeout.hpp

device/pcie/device_memcpy.cpp
```diff
@@ -4,51 +4,170 @@
...
+std::chrono::milliseconds mmio_op_timeout() {
+    static const std::chrono::milliseconds value = [] {
+        const char* env = std::getenv("TT_UMD_MMIO_OP_TIMEOUT_MS");
+        if (env == nullptr || *env == '\0') {
+            return kDefaultMmioOpTimeout;
+        }
+        try {
+            return std::chrono::milliseconds(std::stoul(env));
+        } catch (...) {
+            return kDefaultMmioOpTimeout;
```
aleksamarkovicTT: Not sure this should be a silent failure, could mislead the user. I think it's better to fail fast or at least warn. Failing would be consistent to TT_VISIBLE_DEVICES.

device/api/umd/device/pcie/silicon_tlb_window.hpp
```diff
@@ -81,14 +83,21 @@ class SiliconTlbWindow : public TlbWindow {
     // which glibc's memcpy may perform when unrolling. This affects from and to device.
     // 2. syseng#3487 WH GDDR5 controller has a bug when 1-byte writes are temporarily adjacent
     // to 2-byte writes. We avoid ever performing a 1-byte write to the device. This only affects to device.
-    static void memcpy_from_device(void* dest, const volatile void* src, std::size_t num_bytes);
-    static void memcpy_to_device(void* dest, const void* src, std::size_t num_bytes);
+    static void memcpy_from_device(
+        void* dest, const volatile void* src, std::size_t num_bytes, const std::function<bool()>& on_timeout);
```
broskoTT, line 115: Seems a bit confusing to have fallback both defined here and in the constructor?

tests/api/test_hang_detection.cpp
```diff
@@ -164,6 +170,30 @@ TEST_P(NocHangDetectionTest, DISABLED_TestIsNocHungAPI) {
         << "is_noc_hung() still true after warm reset.";
 }
 
+// Verifies the per-op MMIO timeout (PR #2629): once a NOC is hung, reads complete only after
+// ~700 ms, which exceeds the default 100 ms per-op budget, so the memcpy path aborts with
+// DeviceTimeoutError instead of letting the stores/loads pile up and grind the host.
+TEST_P(NocHangDetectionTest, DISABLED_PerOpTimeoutThrowsOnHungNoc) {
```
broskoTT, line 183: Maybe a better option is to have these kinds of tests in dedicated file (and maybe this whole file with current contents is the right one) with these hang tests, which are too destructive to run on each unit test invocation?
Maybe we can define a "destructive unit tests" target? add this file, maybe also SPI tests? And in the future we can run them daily/weekly on our dedicated, jtag enabled machines

broskoTT (commented):
In general, I think this could've been a couple of smaller PRs. Even if you can only fully test at the tip of the PR stack, it could be much easier to review individual components added and how they're integrated between each other. Your "List of changes" is already one idea on how to draw these lines

device/api/umd/device/tt_device/protocol/device_protocol.hpp
```diff
@@ -41,6 +42,12 @@ class DeviceProtocol {
 
     // Returns the MMIO device ID, used to uniquely identify both the device and its associated protocol instance.
     virtual int get_mmio_id() = 0;
+
+    // Optional hook for the timed MMIO path: invoked (with the in-flight op's NOC) when a single op
+    // exceeds its per-op budget. Returns true if that NOC is confirmed hung (abort the transfer with
+    // DeviceTimeoutError), false to treat the slow op as a false positive and continue. Default is a
+    // no-op for protocols without a timed MMIO path.
+    virtual void set_io_timeout_callback(const std::function<bool(NocId)>& /*hang_check*/) {}
```
nbuncicTT: Device protocol shouldn't have an IO timeout callback as this is a PCIe specific behavior, hence this should be removed.

device/api/umd/device/tt_device/hang_detection/hang_detector.hpp
```diff
@@ -56,6 +67,11 @@ class HangDetector {
     PcieInterface* pcie_interface_;
     bool is_mmio_protocol_;
     architecture_implementation* arch_impl_;
+
+    // Dedicated NOC-probe window (PCIe only), created lazily on first probe and serialized by
+    // probe_lock_. Separate from the protocol's cached window so probes never contend io_lock_.
+    std::unique_ptr<TlbWindow> noc_probe_window_;
```
nbuncicTT: Hang detector shouldn't be aware of anything lower than the protocol level (Device/PCIe, etc.), so it shouldn't be aware of IoWindows/TlbWindows.

device/api/umd/device/tt_device/hang_detection/hang_detector.hpp
```diff
@@ -56,6 +67,11 @@ class HangDetector {
...
+    std::mutex probe_lock_;
```
nbuncicTT: Hang detector shouldn't lock, as TTDevice should provide a lock if it isn't on the higher level API.

device/tt_device/hang_detection/hang_detector.cpp
```diff
@@ -19,6 +24,31 @@ HangDetector::HangDetector(DeviceProtocol* protocol, architecture_implementation
     is_mmio_protocol_(!dynamic_cast<RemoteInterface*>(protocol)),
     arch_impl_(arch_impl) {}
 
+HangDetector::~HangDetector() = default;
+
+uint32_t HangDetector::read_noc_reg_via_probe_window(tt_xy_pair core, uint64_t addr, NocId noc) {
```
nbuncicTT: Instead of providing a full function with an implementation that is tightly coupled to TlbWindow and has a lock, I think we should provide an std::function which accepts these parameters and have two lambdas:
- The default lambda which was here before
- The new lambda with the newly allocated window

My take is that for this new lambda which should be safe for concurrent writes should have a shared_ptr<mutex> which is owned by TTDevice and passed into the lambda's capture.

That way, HangDetector is slightly changed by adding a new way of overriding the default behavior and containing the old one (i.e. one lambda for old, one lambda for new behavior).

There is no knowledge of TlbWindow or locks in HangDetector.

nbuncicTT (commented):
So in a nutshell:
1. `HangDetector` should have an `std::function<uint32_t(tt_xy_pair, uint64_t, NocId)>` member — its "read register" hook
2. Default lambda (set in constructor): wraps protocol_->read_from_device() — the existing behavior, works for JTAG/Remote, no lock needed, and if there's no available additional window (i.e. TLB allocation fails) we can roll back to the default lambda.
3. Override lambda (set via a setter): created by TTDevice, captures a dedicated SiliconTlbWindow and a shared_ptr<mutex> that TTDevice owns — handles the PCIe case with reconfigure + lock, fully self-contained
4. HangDetector calls whichever lambda is set. It has no idea what's behind it — no PcieInterface*, no noc_probe_window_, no probe_lock_.

Also, I'd like that this PR would be split into smaller ones, here's a proposal:
- Changes to HangDetector (for receiving a lambda etc.)
- Changes to TTDevice by setting the new lambda
- Changes to SiliconTlbWindow and memcpy to accept a callback with tests that verify this
- Addition of new exception

device/api/umd/device/pcie/silicon_tlb_window.hpp
pjanevskiTT: this is static function that should probablu move out of this class, but the point of it is that you can have multiple tlb windows with different behavior for fallback if you want

tests/api/test_hang_detection.cpp
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/2835

device/api/umd/device/tt_device/protocol/device_protocol.hpp
pjanevskiTT: addressed

device/pcie/device_memcpy.hpp
pjanevskiTT: settled on std function for now per analysis below, will address template as separate pr

device/tt_device/protocol/pcie_protocol.cpp
```diff
@@ -55,16 +55,26 @@ PcieProtocol::PcieProtocol(std::unique_ptr<PCIDevice> pci_device, bool use_safe_
...
-            pci_device_->get_architecture_implementation()->get_cached_tlb_size(), TlbMapping::UC));
+        // on_timeout consults the wired hang check for the NOC of the in-flight op. true (NOC hung)
+        // aborts with DeviceTimeoutError; false (healthy) continues. With no hang check wired yet, an
+        // overrun aborts outright, preserving the no-callback throw-by-default behavior.
+        auto on_timeout = [this]() -> bool { return hang_check_ ? hang_check_(in_flight_noc_) : true; };
+        cached_tlb_window_ = std::make_unique<SiliconTlbWindow>(
+            pci_device_->allocate_tlb(
+                pci_device_->get_architecture_implementation()->get_cached_tlb_size(), TlbMapping::UC),
+            tlb_data{},
+            on_timeout);
```
tt-vjovanovic: This adds lifetime dependency between `SiliconTlbWindow` and `PcieProtocol`. Consider forwarding on_timeout to read/write calls.

tt-vjovanovic: That will remove `in_flight_noc_` field as well.

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -500,7 +500,12 @@ class TTDevice {
 
     virtual uint32_t get_max_dram_retrain_attempts() const { return 0; }
 
-    void set_hang_detector(std::unique_ptr<HangDetector> hang_detector) { hang_detector_ = std::move(hang_detector); }
+    // Installs the hang detector and, on a timed MMIO path, wires it in two ways:
+    //  - routes a single-op memcpy overrun to a NOC liveness check on the in-flight op's NOC;
+    //  - on PCIe, overrides how that check reads NOC registers so it goes through a dedicated,
+    //    separately-locked window — safe to call from inside a timed-out memcpy (no io_lock_ re-take,
+    //    no timed-path recursion). See the definition in tt_device.cpp.
```
tt-vjovanovic: These comments are not easy to maintain.
> See the definition in tt_device.cpp.
Is the most problematic one as code moves.

device/api/umd/device/utils/error.hpp
```diff
@@ -42,4 +45,57 @@ class SigbusError : public std::runtime_error {
...
+
+/**
+ * @brief Error raised when a host-side MMIO operation exceeds its configured wall-clock budget.
+ *
+ * Distinct from SigbusError: SigbusError indicates the platform's PCIe completion timeout fired
+ * (a hardware-detected fault), whereas DeviceTimeoutError indicates a software budget elapsed —
+ * typically because writes were piling up against a slow or dead NOC before SIGBUS could fire.
+ * Keeping them separate lets callers branch on retry policy.
+ */
+struct DeviceTimeoutError : public UmdError<DeviceTimeoutData> {
```
tt-vjovanovic, line 72: Please expose this to Python code as this is critical exception for not killing host.

device/api/umd/device/utils/timeouts.hpp
```diff
@@ -9,6 +9,11 @@
 namespace tt::umd::timeout {
 inline constexpr auto NON_MMIO_RW_TIMEOUT = std::chrono::milliseconds(5'000);
 
+// Default per-op budget for a single host-side MMIO (TLB-mapped) transfer, overridable at process start
+// via the TT_UMD_MMIO_OP_TIMEOUT_MS env var. 100 ms so post-reset reads (which can legitimately take tens
+// of ms before the device settles) don't trip the timeout on the happy path.
+inline constexpr auto MMIO_OP_TIMEOUT = std::chrono::milliseconds(100);
```
tt-vjovanovic: Consider lowering default timeout based on this data: https://github.com/tenstorrent/tt-exalens/issues/749#issuecomment-3643006466

device/pcie/device_memcpy.cpp
```diff
@@ -4,51 +4,186 @@
...
+                throw std::out_of_range("TT_UMD_MMIO_OP_TIMEOUT_MS out of range");
+            }
+            return std::chrono::milliseconds(static_cast<rep>(ms));
+        } catch (...) {
+            UMD_THROW(
+                error::RuntimeError,
+                std::string("Invalid TT_UMD_MMIO_OP_TIMEOUT_MS: '") + env +
+                    "'. Expected a non-negative integer number of milliseconds.");
+        }
+    }();
```
tt-vjovanovic: Ideally, we would like to be able to change timeout value in runtime.

device/pcie/device_memcpy.cpp
```diff
@@ -4,51 +4,186 @@
...
+        if (env == nullptr || *env == '\0') {
+            return timeout::MMIO_OP_TIMEOUT;
+        }
+        try {
+            using rep = std::chrono::milliseconds::rep;
+            std::size_t pos = 0;
+            const unsigned long ms = std::stoul(env, &pos);
+            // std::stoul silently stops at the first non-digit and quietly wraps a leading '-', so
+            // validate that the whole string was consumed, that it wasn't negative, and that it fits
+            // in milliseconds — otherwise "100ms", "-1" or an overflowing value would disable or
```
tt-vjovanovic: Having constants in comments make them outdated easily.

device/tt_device/tt_device.cpp
```diff
@@ -50,6 +51,44 @@
...
+            return HANG_READ_VALUE;
+        }
+        return value;
+    }
+
+private:
+    PCIDevice *pci_device_;
+    std::mutex probe_lock_;
+    std::unique_ptr<TlbWindow> window_;
+};
```
tt-vjovanovic: In the future, we probably want to share TLB window with other processes. That probably means that we would need to have read/write in KMD.

tt-vjovanovic (approved):

device/pcie/device_memcpy.cpp
```diff
@@ -4,51 +4,186 @@
...
+namespace {
+
+// Resolves the per-op budget once: the TT_UMD_MMIO_OP_TIMEOUT_MS override if set, otherwise the default
+// from timeout::MMIO_OP_TIMEOUT. The budget is applied once per 256-byte block in the bulk AVX2 phase and
+// once per op in the 32 / 16 / 4-byte and byte-wide tail phases. A malformed override fails fast rather
+// than being silently ignored (consistent with TT_VISIBLE_DEVICES), so a typo can't quietly disable the
+// intended budget.
+std::chrono::milliseconds mmio_op_timeout() {
+    static const std::chrono::milliseconds value = [] {
+        const char* env = std::getenv("TT_UMD_MMIO_OP_TIMEOUT_MS");
```
nbuncicTT: Why are we setting the timeout from environmental variables?

device/pcie/device_memcpy.cpp
```diff
@@ -4,51 +4,186 @@
...
 #if defined(__x86_64__) || defined(_M_X64)
 #include <immintrin.h>
 #endif
 
 namespace tt::umd {
 
-void write16_to_device(volatile void* dest, std::uint16_t value) {
+namespace {
+
+// Resolves the per-op budget once: the TT_UMD_MMIO_OP_TIMEOUT_MS override if set, otherwise the default
```
nbuncicTT: I'd refrain from commenting on how the timeout duration is set and focus on the algorithm in the comment (and preferably is shorter).

device/pcie/device_memcpy.cpp
nbuncicTT: I see, we want runtime changes of this. So I'd say it's better to use functions to set it than to use environmental variables as they're hard to discover, can't be tested, and leak across processes. None of the other timeouts in `timeouts.hpp` have one - also the parsing is easier as we have strong types.

If we need runtime configurability while settling on the right default, a static setter/getter is a better fit:
```cpp
class MmioTimeoutConfig {
public:
    static void set_op_timeout(std::chrono::milliseconds ms);
    static std::chrono::milliseconds get_op_timeout();
private:
    static std::chrono::milliseconds op_timeout_; // default: timeout::MMIO_OP_TIMEOUT
};
```
`mmio_op_timeout()` calls `get_op_timeout()` instead of reading the env. Expose it to Python so the user sets it explicitly in their script. Easy to rip out once the value stabilizes, we just delete the class and use the constant.

device/pcie/device_memcpy.cpp
```diff
@@ -4,51 +4,186 @@
...
+        on_timeout_(on_timeout),
+        op_timeout_(mmio_op_timeout()) {}
+
+    void record_and_check(std::chrono::steady_clock::time_point t_before, std::uint32_t op_bytes) {
+        auto t_now = std::chrono::steady_clock::now();
+        auto delta = t_now - t_before;
+        if (delta < op_timeout_) {
+            return;
+        }
+        // Over budget. A wired callback gets to confirm the stall: returning false (false
```
nbuncicTT: Too big of a comment, it should be shorter and less verbose. Also, it uses the budget as a synonym for time elapsed which I find weird.

device/pcie/device_memcpy.cpp
```diff
@@ -4,51 +4,186 @@
...
+    const char* op_verb_;
+    std::size_t total_size_;
+    const std::size_t& remaining_;
+    const MemcpyTimeoutFn& on_timeout_;
+    std::chrono::milliseconds op_timeout_;
+};
+
+}  // namespace
+
+// The single-word scalar transfers below get the same per-op budget as the bulk memcpy paths:
```
nbuncicTT, line 44: Why are we mentioning SIGBUS and hard stalls here? We get SIGBUS once the mappings are invalidated, not when we timeout or get a hard stall.

Too much code duplication on all of the write/read 16/32 functions - use a template for the functions with the correct type.

Something like:
```cpp
template <typename Func>
auto timed_mmio_op(const char* name, const char* verb, std::size_t size, const MemcpyTimeoutFn& on_timeout, Func&& op) {
    std::size_t remaining = size;
    MmioOpTimer timer(name, verb, size, remaining, on_timeout);
    auto t = std::chrono::steady_clock::now();
    if constexpr (std::is_void_v<std::invoke_result_t<Func>>) {
        op();
        remaining = 0;
        timer.record_and_check(t, size);
    } else {
        auto result = op();
        remaining = 0;
        timer.record_and_check(t, size);
        return result;
    }
}
```

device/pcie/device_memcpy.cpp
```diff
@@ -4,51 +4,179 @@
...
+    // bulk loop checks the timeout once per 256-byte block (after all 8 unrolled AVX2
+    // stores), keeping the store pipeline intact within a block. The explicit volatile
+    // loops in Phases 0/4/5 still bound the ordering for the byte/word tails.
     // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast).
     auto* d_simd = const_cast<std::uint8_t*>(static_cast<const volatile std::uint8_t*>(d));
 
     // Phase 1: Bulk 256-byte blocks (8 x 32-byte AVX2 unaligned stores).
     // After this loop, 0 <= size < 256; remaining bytes are handled by Phases 2-5.
     while (size >= 256) {
+        // Loads are from host memory (s) — no TLB access, not instrumented.
```
nbuncicTT, line 116: My comment comes a bit late, but let's collapse these commands into a for loop and tell the compiler to unroll them manually:
```cpp
__m256i v[8];
__attribute__((unroll(8)))
for (int i = 0; i < 8; ++i) {
    v[i] = _mm256_loadu_si256(reinterpret_cast<const __m256i*>(s + i * 32));
}

__attribute__((unroll(8)))
for (int i = 0; i < 8; ++i) {
    _mm256_storeu_si256(reinterpret_cast<__m256i*>(d_simd + i * 32), v[i]);
}
```

if `__attribute__((unroll(8))) ` doesn't work on GCC (or clang), we can generate a macro which does this:
```
#if defined(__clang__)
#define UMD_UNROLL_8 _Pragma("unroll 8")
#elif defined(__GNUC__)
#define UMD_UNROLL_8 _Pragma("GCC unroll 8")
#endif
```
You can ignore this comment for this PR, but let's not forget to tidy up a bit in a separate PR.

device/pcie/device_memcpy.hpp
```diff
@@ -28,8 +51,18 @@ namespace tt::umd {
  * Handles arbitrary alignment and size — leading/trailing bytes smaller than a DWORD are
  * written as individual byte-wide PCIe transactions (the Blackhole PCIe controller supports
  * sub-DWORD writes natively, so no read-modify-write is required).
+ *
+ * Per-op budget: TLB-touching stores must complete within a hard-coded budget
```
nbuncicTT: Drop budget and concrete values from the comments. Make shorter comments if possible.

device/pcie/device_memcpy.hpp
```diff
@@ -6,9 +6,32 @@
...
+ * already held by that op, nor loop back into this same callback. Every TLB-mapped memcpy is
+ * timed, so the timer logic itself is not the hazard; lock re-entrancy and unbounded recursion
+ * are. There is no re-entrancy guard inside memcpy. Any device I/O the callback performs must go
+ * through an independent path with its own lock (and, if it is a memcpy, its own budget). A cheap
+ * BAR-based probe (e.g. HangDetector::is_pcie_hung) satisfies this, and so does a NOC read routed
+ * through a dedicated, separately-locked probe window (see NocProbeReader in tt_device.cpp).
+ * Probing through the stalled op's own cached window / I/O lock (e.g. is_noc_hung via
+ * PcieProtocol::read_from_device) does not.
+ */
+using MemcpyTimeoutFn = std::function<bool()>;
```
nbuncicTT: Do you really need this typedef? I mean in `SiliconTlbWindow` we've used `std::function<bool()>` without an alias? If we do, let's add an alias in `SiliconTlbWindow` also.

device/tt_device/tt_device.cpp
```diff
@@ -50,6 +51,44 @@
...
+// Reads NOC registers for the hang check through a dedicated TLB window guarded by its own lock.
+//
+// This is the override HangDetector uses on PCIe (injected via set_noc_reg_reader). It is deliberately
+// kept out of HangDetector so the detector stays unaware of windows and locks. Because it owns a window
+// and lock that are independent of the protocol's cached window / io_lock_, it is safe to invoke from
+// inside a timed-out memcpy: it neither re-takes io_lock_ (no deadlock) nor re-enters the in-flight op's
+// memcpy path (no recursion), even though its own read goes through the timed memcpy path. The window is
+// created lazily on first probe. The probe read carries its own per-op budget, so if it overruns it throws
+// DeviceTimeoutError, which we map to HANG_READ_VALUE so the hang check reports the NOC as hung.
+class NocProbeReader {
```
nbuncicTT: Why is this in an anonymous namespace?

nbuncicTT (commented):
Will submit more comments later

device/tt_device/tt_device.cpp
```diff
@@ -305,6 +344,27 @@ bool TTDevice::is_noc_hung(NocId noc, TTDevice::HangAction action) {
...
+
+    // Route a single-op memcpy overrun to a NOC liveness check on the in-flight op's NOC: a hung NOC
+    // aborts the transfer with DeviceTimeoutError; a healthy NOC lets it continue.
+    pcie_capabilities_->set_io_timeout_callback(
+        [this](NocId noc) -> bool { return is_noc_hung(noc, HangAction::RETURN); });
+
+    // The liveness check runs from inside a timed-out memcpy that holds io_lock_, so it must not read
+    // through the protocol's cached window. Override the detector's NOC read with a dedicated,
+    // separately-locked probe window (owned by the injected reader, not by HangDetector).
+    auto probe = std::make_shared<NocProbeReader>(pcie_capabilities_->get_pci_device());
```
nbuncicTT: I think `NocProbeReader` is probably overengineered. The lambda injected via set_noc_reg_reader (from PR #2799) can own both the window and the lock directly — no separate class needed:

```cpp
void TTDevice::set_hang_detector(std::unique_ptr<HangDetector> hang_detector) {
    hang_detector_ = std::move(hang_detector);

    if (pcie_capabilities_ == nullptr) {
        return;
    }

    pcie_capabilities_->set_io_timeout_callback(
        [this](NocId noc) -> bool { return is_noc_hung(noc, HangAction::RETURN); });

    auto w = std::shared_ptr<TlbWindow>(get_io_window({}, TlbMapping::UC));
    auto mtx = std::make_shared<std::mutex>();

    hang_detector_->set_noc_reg_reader(
        [w, mtx](tt_xy_pair core, uint64_t addr, NocId noc) -> uint32_t {
            std::lock_guard<std::mutex> lock(*mtx);
            uint32_t value = 0;
            try {
                w->read_block_reconfigure(&value, core, addr, sizeof(value), noc);
            } catch (const error::DeviceTimeoutError&) {
                return HANG_READ_VALUE;
            }
            return value;
        });
}
```
The window and lock live entirely inside the lambda's capture. TTDevice spawns them and hands them off — it never stores or touches them again. HangDetector sees only the std::function, no knowledge of windows or locks. NocProbeReader can be deleted.

device/tt_device/protocol/pcie_protocol.cpp
```diff
@@ -55,16 +55,26 @@ PcieProtocol::PcieProtocol(std::unique_ptr<PCIDevice> pci_device, bool use_safe_
...
     }
     return cached_tlb_window_.get();
 }
 
 void PcieProtocol::write_to_device(const void* mem_ptr, tt_xy_pair core, uint64_t addr, size_t size, NocId noc_id) {
     std::lock_guard<std::mutex> lock(io_lock_);
+    in_flight_noc_ = noc_id;
```
nbuncicTT: `in_flight_noc_` seems like an unnecessary side-channel. The timeout callback is `std::function<bool()>` with no parameters, so it can't receive the NOC directly and has to fish it out of a member variable set before each transfer. Instead, make the callback take the NOC as a parameter:


`using MemcpyTimeoutFn = std::function<bool(NocId)>;`

The NOC is already known at the call site, pass it through the timer/memcpy infrastructure to the callback invocation:


if (on_timeout_ && !on_timeout_(noc_id)) {
    return;
}
in_flight_noc_, set_io_timeout_callback, and the hang_check_ member on PcieProtocol all go away.

device/api/umd/device/utils/error.hpp
pjanevskiTT: did it in a separate PR https://github.com/tenstorrent/tt-umd/pull/2848

device/pcie/device_memcpy.cpp
pjanevskiTT: Tracked as a follow-up in #2849 — will address there to keep this PR focused on the timeout mechanism.

device/pcie/device_memcpy.cpp
pjanevskiTT: Tracked as a follow-up in #2849 — will address there to keep this PR focused on the timeout mechanism.

device/tt_device/tt_device.cpp
pjanevskiTT: removed this class

device/api/umd/device/chip/local_chip.hpp
```diff
@@ -106,6 +107,11 @@ class LocalChip : public Chip {
     TlbWindow* get_cached_wc_tlb_window();
     TlbWindow* get_cached_uc_tlb_window();
 
+    // Builds the per-op MMIO timeout veto for TLB-mapped transfers: on an overrun the configured hang
```
nbuncicTT: nit: No need for this comment here, as it's usage and purpose should be explained in the TTDevice callback. 

On a side note: This is just copy-paste code which we have because we currently don't have WC TLBs in TTDevice, and when we have them, we will delete this.

device/chip/local_chip.cpp
```diff
@@ -360,7 +361,16 @@ void LocalChip::read_from_device_reg(CoreCoord core, void* dest, uint64_t reg_sr
     TlbWindow* tlb_window = get_cached_uc_tlb_window();
     tlb_window->configure(config);
 
-    tlb_window->read_register(reg_src - tlb_window->get_base_address(), dest, size);
+    tlb_window->read_register(reg_src - tlb_window->get_base_address(), dest, size, make_io_timeout_callback());
+}
+
+std::function<bool()> LocalChip::make_io_timeout_callback() {
+    // Consulted only on a per-op MMIO overrun (see OpTimeoutGuard). Evaluate the NOC selection at call
```
nbuncicTT: nit: I'd omit this comment because it explains what the lambda is used for, and this is repeated around the codebase. 

These, and other comments add noise to somewhat simple functions. The functionality should be explained where this lambda is used (i.e. the TLB class, and I think it's already commented).

device/api/umd/device/tt_device/protocol/pcie_interface.hpp
```diff
@@ -47,6 +48,11 @@ class PcieInterface {
 
     virtual void bar_write32(uint32_t addr, uint32_t data) = 0;
     virtual uint32_t bar_read32(uint32_t addr) = 0;
+
+    // Hook for the timed MMIO path: invoked (with the in-flight op's NOC) when a single op exceeds its
+    // per-op budget. Returns true if that NOC is confirmed hung (abort the transfer with
```
nbuncicTT: nit: I'd omit the comment, because there is no context what's the "per-op budget". Maybe a single line comment which states that this is the callable invoked when a timeout elapses for IO's, but that's already in the name of the function..

device/api/umd/device/tt_device/protocol/pcie_protocol.hpp
```diff
@@ -96,6 +98,11 @@ class PcieProtocol : public DeviceProtocol, public PcieInterface {
     std::mutex dma_mutex_;
     std::unique_ptr<TlbWindow> cached_tlb_window_;
     std::unique_ptr<TlbWindow> cached_dma_tlb_window_;
+
+    // Hang check forwarded per-call to the cached window's timed memcpy path. The in-flight op's NOC is
```
nbuncicTT: nit: I'd make the comment a one-liner static that this is the callback called on an IO timeout

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -511,7 +511,8 @@ class TTDevice {
 
     virtual uint32_t get_max_dram_retrain_attempts() const { return 0; }
 
-    void set_hang_detector(std::unique_ptr<HangDetector> hang_detector) { hang_detector_ = std::move(hang_detector); }
+    // Installs the hang detector and, on PCIe, wires it to the per-op MMIO timeout path.
```
nbuncicTT: Why would the public facing API care about the per-op MMIO timeout path? I'd remove this comment.

device/api/umd/device/utils/mmio_timeout_config.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+#include "umd/device/utils/timeouts.hpp"
+
+namespace tt::umd {
+
+/**
+ * @brief Runtime-configurable per-op budget for host-side MMIO (TLB-mapped) transfers.
+ *
+ * The per-op MMIO timeout is still being tuned, so the budget is settable at runtime via this small
+ * setter/getter instead of a constant. A programmatic setter (rather than an env var) is discoverable,
```
nbuncicTT: nit: I really think the comment that suggests how  a setter is better than an env var should not be here, or how the callers use this directly or through python. Keep the comment about what it does to the op_timeout.

device/pcie/device_memcpy.cpp
```diff
@@ -4,51 +4,118 @@
...
 #include <immintrin.h>
 #endif
 
 namespace tt::umd {
 
-void write16_to_device(volatile void* dest, std::uint16_t value) {
+namespace {
+
+// Builds the per-op timeout guard for the MMIO transfer routines: the configured budget plus the
+// optional liveness veto, with an overrun action that throws DeviceTimeoutError. `bytes_remaining` is
```
nbuncicTT: nit: I think this should be phrased more simply, omitting words like optional liveness veto

device/tt_device/protocol/pcie_protocol.cpp
```diff
@@ -55,10 +55,14 @@ PcieProtocol::PcieProtocol(std::unique_ptr<PCIDevice> pci_device, bool use_safe_
...
+void PcieProtocol::set_io_timeout_callback(const std::function<bool(NocId)>& hang_check) { hang_check_ = hang_check; }
+
 TlbWindow* PcieProtocol::get_cached_tlb_window() {
     if (cached_tlb_window_ == nullptr) {
-        cached_tlb_window_ = std::make_unique<SiliconTlbWindow>(pci_device_->allocate_tlb(
-            pci_device_->get_architecture_implementation()->get_cached_tlb_size(), TlbMapping::UC));
+        cached_tlb_window_ = std::make_unique<SiliconTlbWindow>(
+            pci_device_->allocate_tlb(
+                pci_device_->get_architecture_implementation()->get_cached_tlb_size(), TlbMapping::UC),
+            tlb_data{});
```
nbuncicTT: Why is adding `tlb_data{}` needed here?

nbuncicTT (commented):
After looking at this PR a few times, I'd propose a simplification because the blast radius is too big, and the interface changes are too much for what should be an implementation detail with a settable parameter..

The timeout mechanism is good, but passing `on_timeout` through every `TlbWindow` virtual signature is too much interface churn for what's fundamentally an internal implementation detail. 22 files changed, all public signatures modified for a callback that's always the same function (`is_noc_hung`) and never a caller's decision.

Proposed simplification:

- `SiliconTlbWindow` constructor accepts an `std::function<bool(NocId)>` hang check callback (defaults to empty). It stores it and uses it internally, reads the NocId from its current TLB config, constructs the veto, passes it to `device_memcpy`. No signature changes on any `TlbWindow` method.
- PcieProtocol passes its stored `hang_check_` when lazily constructing its cached windows. The wiring path (`TTDevice::set_hang_detector` → `PcieInterface` → `PcieProtocol` → constructor) stays the same.
- `MmioTimeoutConfig::set_op_timeout` remains the only client-facing API.

Same safety, same behavior, but the blast radius drops from 22 files to ~4 and the public interface stays untouched.

device/pcie/silicon_tlb_window.cpp
```diff
@@ -66,27 +66,44 @@ struct ScopedJumpGuard {
...
-    TlbWindow(std::move(handle), config) {}
+SiliconTlbWindow::SiliconTlbWindow(
+    std::unique_ptr<TlbHandle> handle, const tlb_data config, std::function<bool(NocId)> hang_check) :
+    TlbWindow(std::move(handle), config), hang_check_(std::move(hang_check)) {}
+
+void SiliconTlbWindow::set_io_timeout_hang_check(const std::function<bool(NocId)> &hang_check) {
+    hang_check_ = hang_check;
+}
+
+std::function<bool()> SiliconTlbWindow::make_io_timeout_callback() const {
```
pjanevskiTT: is this function needed, why don't read/write16 and 32 use already stored hang check callback?

device/api/umd/device/chip/local_chip.hpp
```diff
@@ -106,6 +107,8 @@ class LocalChip : public Chip {
     TlbWindow* get_cached_wc_tlb_window();
     TlbWindow* get_cached_uc_tlb_window();
 
+    std::function<bool(NocId)> make_io_timeout_hang_check();
```
nbuncicTT, line 110: This is only needed as the WC window is currently managed by LocalChip, right? Once we move it out, this function will also be migrated?

nbuncicTT (approved):
Well done

device/api/umd/device/chip/local_chip.hpp
pjanevskiTT: yes

Issue #2531, opened 2026-04-28
Power mode with O_APPEND leads to NOC1 failures, needs investigation


nbuncicTT:
**Describe the bug**
Here is the issue that discovered the bug: https://github.com/tenstorrent/tt-umd/issues/2449

After bisecting it was found that this PR introduced the bug: https://github.com/tenstorrent/tt-umd/pull/2252.

After further investigation, in the UMD codebase, it was found that once opening the file descriptor of the character driver without `O_APPEND`, it doesn't have this behavior from the issue above.

Note: `O_APPEND` explanation - this flag makes the driver user a "power aware" user meaning that it doesn't automatically go out of low power mode when opening the file descriptor - it rather reacts to IOCTLs with set the power state explicitly.

**Stack trace**
/

**To reproduce**
Do the steps from issue above with a power aware KMD version and with the PR which made UMD aware of these changes.

**Expected behavior**
The same behavior as before (i.e. NOC1 doesn't fail on reading a register in the tensix if the power mode is high).

broskoTT:
As a workaround we turned off power awareness (which means we will always use old KMD API which opens device in high power mode by default always): Now the tests don't fail anymore.

@aleksamarkovicTT tried to repro this again with reverted change explaned in previous comment, and he still couldn't repro. He's suspecting this magically got resolved on its own.

aleksamarkovicTT:
Not able to repro using NOC1 exalens library tests on bh-40; FW 19.5, 19.8; KMD 2.7.0
Failures were originally ND to begin with, tests failing every 3rd run, but failing would always result in a NOC1 hang, killing all further runs or the host itself. Failure was a NOC1 read to 0xFFB121B0 TENSIX_SOFT_RESET_ADDR when setting up debug bus tests.

broskoTT:
Agreement was to run tests many times on BH on CI to verify it this is happening. If not happening, merge to main

aleksamarkovicTT:
Repro'd: https://github.com/tenstorrent/tt-umd/actions/runs/27537813359/job/81392449701?pr=2737

aleksamarkovicTT:
Note: Unlike we previously thought, exalens does not create a new context between tests suites. Discovery is only done at the beginning, and 2-3 more times at the end. The tests where this hang usually occurs is at quite a later point from discovery.

I can't repro on bh-40, we might need a different machine.

broskoTT:
Action items:
[...]

PR #2631, opened 2026-05-13
Modify log for low power mode

nbuncicTT:
### Issue
#2531 

### Description
This change is to clarify the handling of low power mode and update the related warning message, ensuring that the use is aware that the device remains in high power mode until a known issue is resolved.

### List of the changes
Power management logic improvements:

* Changed where log_warning is submitted.
* Added comment.

### Testing
CI

### API Changes
/


device/topology/topology_discovery.cpp
```diff
@@ -135,12 +135,14 @@ void TopologyDiscovery::get_connected_devices() {
...
+        if (options.low_power) {
             // Low power mode is temporarily disabled. See https://github.com/tenstorrent/tt-umd/issues/2531.
             log_warning(
                 LogUMD,
-                "Low power mode is disabled while UMD holds open file descriptors. The device will return to low power "
-                "mode once all file descriptors are closed.");
+                "Low power mode is not yet supported. The device will remain in high power mode while UMD holds open "
+                "file descriptors.");
+        } else {
+            // Setting power state is disabled until https://github.com/tenstorrent/tt-umd/issues/2531 is resolved.
```
pjanevskiTT: This comment is contradictory: it says "Setting power state is disabled until #2531 is resolved" but the very next line calls `set_power_state(true)`. The "disabled until #2531" reasoning applies to the *low power* request (handled in the `if` branch above), not to the high-power default path. Suggest either dropping this comment, or rewording it to clarify that what's gated on #2531 is the ability to honour `low_power = true` — the high-power state itself is still being set here.

nbuncicTT: It's not contradictory because the underlying API doesn't set the power state even after the call to `set_power_state` as it's effectively a no-op. I'll modify the comment to reflect this.

broskoTT (approved):

Issue #2494, opened 2026-04-23
[Quasar] Add Quasar coordinate manager


pjanevskiTT:
`device/coordinates/coordinate_manager.cpp:620` and `:684` have `case tt::ARCH::QUASAR: // TODO (#450): Add Quasar configuration`. Issue #450 is closed but these TODOs were not covered.

Add:
- `device/api/umd/device/coordinates/quasar_coordinate_manager.hpp`
- `device/coordinates/quasar_coordinate_manager.cpp`

Subclass of `CoordinateManager` implementing NOC0↔NOC1 translation for the Quasar grid. Wire into the factory switch in `coordinate_manager.cpp`.

Pattern: `blackhole_coordinate_manager.*`.

**Acceptance**: TODOs removed, coordinate translation round-trips in a unit test.

Part of #2490.

PR #2633, opened 2026-05-14
[Quasar] Add QuasarCoordinateManager (identity translated mapping)

pjanevskiTT:
### Issue
#2494

### Description
Quasar previously fell through to `BlackholeCoordinateManager` in both `CoordinateManager::create_coordinate_manager` factory overloads. The arch-enum-only path used `blackhole::*` core constants for Quasar; the cores-vector path instantiated a `BlackholeCoordinateManager` directly. In simulation, the YAML-supplied NOC0 cores do get wired in (via the cores-vector overload), but Blackhole's `fill_eth_noc0_translated_mapping` and `fill_pcie_noc0_translated_mapping` apply Blackhole-specific translated-coordinate offsets, and `assert_coordinate_manager_constructor` enforces Blackhole-specific ETH harvesting rules. Either can produce missing TRANSLATED entries or spurious assert failures on Quasar inputs.

This PR introduces `QuasarCoordinateManager` as its own subclass of `CoordinateManager`, wires it into both factory overloads, and pulls the arch-enum-only path onto `grendel::*` constants. The translated-coordinate fill overrides delegate to the base-class identity defaults, which is correct for current Quasar simulation usage (SOC descriptor YAML supplies NOC0 directly) and a known placeholder for silicon — tracked in #2494.

### List of the changes
- New `device/api/umd/device/coordinates/quasar_coordinate_manager.hpp`
- New `device/coordinates/quasar_coordinate_manager.cpp` (identity overrides for tensix / dram / eth / pcie / arc `fill_*_noc0_translated_mapping`)
- `device/coordinates/coordinate_manager.cpp`: split `QUASAR` out of the `BLACKHOLE` fallthrough in both `create_coordinate_manager` overloads; use `grendel::*` constants on the arch-enum-only path
- `device/CMakeLists.txt`: register new `.cpp` and the public `.hpp`

### Testing
- `cmake --build build` clean
- `pre-commit run --all-files` clean
- Manually traced `quasar_simulation_2x3.yaml` cores (TENSIX `[0-1, 1-1]`, DRAM `[1-0],[0-0]`, DISPATCH `[1-2]`) through the new manager and confirmed all get identity TRANSLATED entries via the base-class default helpers

### API Changes
New public header `umd/device/coordinates/quasar_coordinate_manager.hpp` exposing `QuasarCoordinateManager`. No changes to existing API surface.

### DIFF
--- a/tests/baremetal/test_core_coord_translation_quasar.cpp
+++ b/tests/baremetal/test_core_coord_translation_quasar.cpp
@@ -1,0 +1,230 @@
+// SPDX-FileCopyrightText: © 2024 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include <gtest/gtest.h>
+
+#include <cstddef>
+#include <cstdint>
+#include <memory>
+#include <set>
+#include <stdexcept>
+#include <vector>
+
+#include "umd/device/arch/grendel_implementation.hpp"
+#include "umd/device/coordinates/coordinate_manager.hpp"
+#include "umd/device/types/arch.hpp"
+#include "umd/device/types/cluster_descriptor_types.hpp"
+#include "umd/device/types/core_coordinates.hpp"
+#include "umd/device/types/xy_pair.hpp"
+#include "umd/device/utils/common.hpp"
+
+using namespace tt;
+using namespace tt::umd;
+
+// QuasarCoordinateManager currently maps every NOC0 coordinate to itself in
+// the TRANSLATED system (see #2494). These tests assert that identity mapping
+// for every core type the manager handles, plus the logical/NOC0/NOC1 and
+// harvesting plumbing inherited from CoordinateManager.
+
+namespace {
+
+// Checks that NOC0 -> TRANSLATED is an identity on (x, y). The placeholder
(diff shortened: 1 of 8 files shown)


tests/baremetal/test_core_coord_translation_quasar.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+namespace {
+
+// Checks that NOC0 -> TRANSLATED is an identity on (x, y). The placeholder
+// Quasar arch constants have overlapping NOC0 entries across core types
+// (e.g. {9, 2} is in both TENSIX_CORES_NOC0 and DRAM_CORES_NOC0), so the
+// looked-up TRANSLATED core_type is not guaranteed to match the requested
+// one — but the coordinate identity is what QuasarCoordinateManager actually
+// promises.
+void expect_identity_translated_for_cores(
```
broskoTT, line 38: We're again programming tests... You aren't just testing the inputs and outputs, you're locking in the implementation.
I think this is bad, I know it is already like that for WH and BH... We'll have to clean it up someday anyway

broskoTT (approved):

PR #2650, opened 2026-05-17
Hardcode earlier TTSIm version

broskoTT:
### Issue
Failing tests on newest runs: https://github.com/tenstorrent/tt-umd/actions/runs/25995615435/job/76409480020?pr=2567

### Description
Seems like there is some sort of a regression in the newest ttsim (or an improvement we didn't integrate)

### List of the changes
- Hardcode temporarily ttsim version to 1.5.8

### Testing
No testing, but here's a passing run on main branch with this version: https://github.com/tenstorrent/tt-umd/actions/runs/25938895541/job/76251280249

### API Changes

### DIFF
--- a/.github/workflows/run-ttsim-tests.yml
+++ b/.github/workflows/run-ttsim-tests.yml
@@ -71,6 +71,9 @@ jobs:
             TTSIM_VERSION="v1.0.0"
           fi
           echo "Extracted ttsim version: $TTSIM_VERSION"
+          # TODO: Temporarily hardcoded to v1.5.8 because the newest version has an issue and tests don't pass.
+          TTSIM_VERSION="v1.5.8"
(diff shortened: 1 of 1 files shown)


mcraigheadTT:
`[ RUN      ] TTSimDeviceIOFixture.WriteToDeviceReadByTileRdBytes`

This is using a deprecated API (has been deprecated for many months; was made officially Unsupported in preparation for upcoming ttsim OSS release, as this code's inclusion would have made the OSS-filtered code quite confusing for several reasons) and then hitting `[0] ERROR: UnsupportedFunctionality: libttsim_tile_rd_bytes`.

From the test name alone, I suspect that this test is simply no longer valid, at least on WH/BH, and should be disabled/removed, at least on those chips.

Let me know if you need anything else.

pjanevskiTT:
handled properly by #2654

PR #2662, opened 2026-05-18
Manual main runs

broskoTT:
### Issue
Optimizing merge queue

### Description
Delay jobs on main to reduce pressure on CI machines and make merge queue faster.

### List of the changes
- Add gatekeeper job which will run only on push to main. The key is it runs in specific environment. In the settings, I've currently setup the environment to have a delay of 12h. We can adjust in the future as needed.
- Make sure the status report at the end checks the ensure-docker job as well, just in case.
- Add explicit checks to all the jobs, so they don't get unintentionally skipped.

### Testing
Added pull_request to trigger just to check how it works:

image

### API Changes


.github/workflows/build-and-run-all-tests-multiple-builds.yml
```diff
@@ -39,11 +39,24 @@ permissions:
   packages: read
 
 jobs:
+  # Gatekeeper job allows us to define an environent which only affects pushes to main, but not other types of jobs.
```
nbuncicTT: nit: environment*

.github/workflows/build-and-run-all-tests-multiple-builds.yml
```diff
@@ -39,11 +39,24 @@ permissions:
...
+  # With this environment, we can control how the jobs run on main, for exammple delayed or waiting for approval.
```
nbuncicTT: nit: example

.github/workflows/build-and-run-all-tests-multiple-builds.yml
```diff
@@ -39,11 +39,24 @@ permissions:
...
+  gatekeeper:
```
nbuncicTT, line 44: Wondering if it makes sense to have one nightly test scheduled at a fixed time instead of doing this 12 hour delay - because if in a short period of time we get 10 PRs merged, you'll schedule 10 subsequent tests on main with a 12 hour delay - and that is more robust, but it's also CI consuming.

With all said, I think the merge queue is robust enough so my proposal would be to have one fixed job daily which takes the latest main. If you think this isn't good enough because of tracking (i.e. which change broke main) - I'm fine with this going into the CI.

nbuncicTT (approved):

.github/workflows/build-and-run-all-tests-multiple-builds.yml
broskoTT: Yeah main motivation was tracking...
Having one job during the night would definitely reduce the stress on CI machines though

PR #2691, opened 2026-05-20
Add CI perf regression detection

ajovanovicTT:
### Issue
#1533

### Description
Introduces ci regression detection, which for now due to unstable testing environment, summarizes regression only on n150 architecture (binding microbenchmark test runs for that arch to bgd-lab-06), on ClusterConstructor test. This pipeline now bound only to n150 arch should be the base, to which tests when stabilized can be added easily later.

### List of the changes
**New scripts** (`tests/microbenchmark/tools/`):
- `update_baseline.py` — consumes a downloaded benchmark-json-* CI artifact from a dedicated runner and writes the per-arch baseline YAML. Derives median_throughput = batch / median(elapsed) and tolerance_pct = max(5, ceil(MAPE_K * mape_pct)) from nanobench's median(elapsed) + medianAbsolutePercentError(elapsed). Preserves existing gate: true flags; cases present in the YAML but absent from the artifact keep their old values with a # no new result this calibration annotation.
- `summarize_regressions.py` — diffs CI's per-arch `benchmark-json-*` artifacts against the in-repo baselines, emits a cross-arch summary table + per-(test, arch) breach detail to `$GITHUB_STEP_SUMMARY`, exits 1 on any `gate: true` breach.
- `compare_to_baseline.py` — local wrapper around `summarize_regressions` for on-runner comparison.

**Baseline layout** (`tests/microbenchmark/expected/baselines/`):
- `n150.yaml` — calibrated from CI run on bgd-lab-06.
- `n300.yaml`, `p150.yaml` — placeholders (contents commented out) until those runners come online; loader skips them.

**Workflows:**
- `regression-check.yml` (new) — downloads `benchmark-json-*`, runs `summarize_regressions.py` against the baselines dir, publishes markdown to the Step Summary, and uploads it as `regression-summary`. Wired in as a `needs: run-benchmarks` downstream job.
- `build-and-run-all-benchmarks.yml` — n150 pinned to `bgd-lab-06`; n300/p150b keep their pool labels.

### Testing
CI
Manual testing

### API Changes


ajovanovicTT:
New PR, since logic was changed significantly since previously opened #2619, but regression-check summary examples look the same as in links in comments on that PR. There are examples of how summaries would look like when more tests are included in bottom comments on that PR (intentional slowdowns and speedups so it can be visualised).

ajovanovicTT:
Intentional close to boundary slowdown for example:
https://github.com/tenstorrent/tt-umd/actions/runs/26164288064

tests/microbenchmark/expected/baselines/wh_n150.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+metadata:
+  schema_version: 1
+  arch: "WH n150"
+  runner_hostname: "bgd-lab-06"
+  calibrated_at: "2026-05-22"
+  notes: "Updated from a CI benchmark JSON artifact via tests/microbenchmark/tools/update_baseline.py."
+
+ClusterConstructor:
+  default: { median_throughput: 248.7, tolerance_pct: 5 }
```
pjanevskiTT: similar comment to the one in the previous PR, I think median_throughput is not the adequate name for this. Because sometimes this might be time (like in this example) so I would go for median_value or something like that

pjanevskiTT: Is this in milliseconds, I would expect n150 constructor to work faster

tests/microbenchmark/tools/compare_to_baseline.py
```diff
@@ -0,0 +1,151 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
```
pjanevskiTT, line 1: can we come up with short readme in this tools directory on how can we use all these scripts and what are the usecases

tests/microbenchmark/expected/baselines/wh_n150.yaml
pjanevskiTT: since this was a comment on previous PR as well, make sure to check if other comments from old PR can be applied to this one

tests/microbenchmark/tools/summarize_regressions.py
```diff
@@ -0,0 +1,549 @@
...
+
+def render_detail_subtable(title: str, rows: list) -> str:
+    """Render a markdown table for one direction of breaches.
+
+    `rows` is a list of (case_name, current_thr, baseline_thr, delta_pct,
+    tolerance_pct, status, unit) tuples, presumed already sorted.
+    """
+    # Local import of analyze_results to avoid a top-level dependency cycle
+    # (analyze_results pulls in pandas, which isn't needed for the loading path).
+    sys.path.insert(0, str(Path(__file__).resolve().parent))
```
pjanevskiTT: The `sys.path.insert` + lazy `from analyze_results import format_throughput` mutates global state on every `render_detail_subtable` call. The rationale (dodging a pandas top-level dep) doesn't hold — `regression-check.yml` already pip-installs pandas, and `compare_to_baseline.py` will hit this import the moment any cell breaches. Net effect: a user running `compare_to_baseline.py` locally without pandas gets a confusing `ImportError` mid-render. Fix: move the import to module top-level, or extract `format_throughput` (14 lines) into a shared module so we can drop `analyze_results` from the sparse-checkout entirely.

.github/workflows/build-and-run-all-benchmarks.yml
```diff
@@ -81,3 +91,17 @@ jobs:
...
       baseline-run-id: ${{ inputs.baseline-run-id }}
+
+  # Regression check against in-repo per-arch baseline YAMLs (median calibrated
+  # on the dedicated runner + per-case tolerance). Runs in parallel with
+  # analyze-results, which answers a different question (drift vs. the single
+  # latest main run).
+  regression-check:
+    secrets: inherit
+    needs: run-benchmarks
+    if: ${{ !cancelled() && needs.run-benchmarks.result != 'skipped' }}
```
pjanevskiTT: `if: !cancelled() && needs.run-benchmarks.result != 'skipped'` lets the regression check run even when `run-benchmarks` *failed*. If a benchmark binary crashes mid-suite, partial artifacts upload, and the missing-gated table doesn't fail the job by design — so a silent crash produces green CI. Either tighten this to `needs.run-benchmarks.result == 'success'`, or promote missing-gated to a job failure in `summarize_regressions.py` when a gated case produces no result. The latter is more useful long-term.

tests/microbenchmark/tools/compare_to_baseline.py
```diff
@@ -0,0 +1,151 @@
...
+exporter actually writes JSON), then point this script at the resulting
+directory and your arch to get a markdown regression report against the
+in-repo per-arch baseline YAML for that arch.
+
+Example:
+
+    export UMD_MICROBENCHMARK_RESULTS_PATH=/tmp/umd-bench
+    mkdir -p "$UMD_MICROBENCHMARK_RESULTS_PATH"
+      ./build/test/umd/microbenchmark/umd_microbenchmark \\
+        --gtest_filter='MicrobenchmarkTLB.CompareMulticastandUnicast'
```
pjanevskiTT: The example references `MicrobenchmarkTLB.CompareMulticastandUnicast`, but the only shipped n150 baseline is `MicrobenchmarkOpenCluster.ClusterConstructor` (see `expected/baselines/wh_n150.yaml`). A user copy-pasting this gets a report full of "no baseline" cells. Swap the `--gtest_filter` to a case that exists in the shipped baseline.

tests/microbenchmark/tools/update_baseline.py
```diff
@@ -0,0 +1,375 @@
...
+#
+# mape_pct is nanobench's medianAbsolutePercentError(elapsed) for a single
+# run: the spread of epoch medians within that run, expressed as a percent of
+# the median. MAPE_K=1 sizes tolerance to equal that observed spread.
+#
+# Bump MAPE_K up if CI's run-to-run variance turns out larger than within-run
+# variance.
+
+TOLERANCE_FLOOR_PCT = 5
+MAPE_K = 1
```
pjanevskiTT, line 70: `MAPE_K = 1` and `TOLERANCE_FLOOR_PCT = 5` are reasonable for the first n150 calibration but hardcoded. Different hardware noise profiles on n300/p150b will likely want different values. Expose both as CLI flags (`--mape-k`, `--tolerance-floor`) before the next calibration cycle so each arch can be tuned without code edits.

tests/microbenchmark/tools/summarize_regressions.py
```diff
@@ -0,0 +1,549 @@
...
+            for case_name, entry in cases.items():
+                combined.setdefault(title, {}).setdefault(case_name, {})[arch] = entry
+    return combined
+
+
+# --- Classification --------------------------------------------------------------
+
+
+def classify(
+    current_thr: float, baseline_thr: float, tolerance_pct: float
```
pjanevskiTT, line 183: No unit tests for any of this tooling — `classify`, the MAPE math in `update_baseline.derive_entry`, gate preservation, the nan-regex preprocessing, baseline merging. The risk of silent regression in the regression detector itself is non-trivial as this expands beyond n150. A small pytest with synthetic fixtures (one baselines-dir + one artifact dir exercising OK / DOWN / UP / no-result / no-baseline / gated-breach) would lock the behavior down cheaply.

.github/workflows/regression-check.yml
```diff
@@ -0,0 +1,97 @@
...
+# tests/microbenchmark/tools/update_baseline.py — median(elapsed) +
+# MAPE-derived tolerance), applies per-case tolerances, and emits a coarse
+# cross-arch summary plus per-(test, arch) detail sections to
+# $GITHUB_STEP_SUMMARY.
+#
+# Pilot: soft by default, but can hard-fail when gate: true breaches.
+name: Regression check against in-repo baseline
+
+on:
+  workflow_call:
```
pjanevskiTT, line 15: `on: workflow_call` only — no `workflow_dispatch` fallback. That means this workflow can't be invoked standalone for debugging (e.g. to re-run the summary against artifacts from a prior run-id). Fine if intentional for scratch scope, but worth confirming.

tests/microbenchmark/tools/update_baseline.py
```diff
@@ -0,0 +1,375 @@
...
+def arch_slug(arch_label: str) -> str:
+    """Canonical filename slug for a given arch label.
+    "WH n150" -> "wh_n150", "BH p150b" -> "bh_p150b".
+    """
+    return arch_label.lower().replace(" ", "_")
+
+
+def _yaml_escape(s: str) -> str:
+    """Wrap a string in quotes if it contains characters that confuse YAML."""
+    if any(c in s for c in ":#,&*!|>'\"%@`{}[]\n") or s.strip() != s:
```
pjanevskiTT: The hand-rolled `_yaml_escape` covers `` :#,&*!|>'"%@`{}[] `` and leading/trailing whitespace, but misses:
- leading `-` / `?` (block-sequence / block-mapping indicators)
- names starting with digits, or `true`/`false`/`null`/`yes`/`no` literals

A future case named e.g. `1-byte transfer` would round-trip as an int key, not a string. Either expand the check or leave a comment listing which characters are intentionally not handled so the next person knows the contract.

tests/microbenchmark/tools/summarize_regressions.py
```diff
@@ -0,0 +1,549 @@
...
+_NAN_RE = re.compile(r"-?\bnan\b")
+
+import yaml
+
+# --- Arch label canonicalization -------------------------------------------------
+# Artifact names look like:
+#   benchmark-json-wormhole_b0-tt-ubuntu-2204-n150-viommu-stable-ubuntu-22.04
+# We canonicalize to short labels matching the per-arch YAML filenames.
+# Substring match keeps this robust to minor variations (e.g. -iter1 suffixes).
+ARCH_PATTERNS = [
```
pjanevskiTT: `arch_label_from_artifact` substring-matches "n150"/"n300"/"p150b" against the full artifact name. Robust today, but a future arch named e.g. `p150b-revB` would mis-route as `BH p150b`. Token-matching between `-` separators is safer; not urgent.

tests/microbenchmark/tools/update_baseline.py
```diff
@@ -0,0 +1,375 @@
...
+            print(f"WARN: skipping {path}: {e}", file=sys.stderr)
+            continue
+        for result in data.get("results", []):
+            case = result.get("name")
+            med = result.get("median(elapsed)")
+            batch = result.get("batch")
+            if case is None or med is None or batch is None or med == 0 or batch == 0:
+                continue
+            # nanobench stores MAPE as a fraction (0.03 == 3%); convert to %.
+            mape_pct = (result.get("medianAbsolutePercentError(elapsed)") or 0.0) * 100
```
pjanevskiTT: `result.get("medianAbsolutePercentError(elapsed)") or 0.0` conflates a legitimate `0.0` MAPE (a perfectly clean run) with a missing key. No observable bug today since the 5% floor catches both, but if the tolerance math ever stops flooring, this becomes a silent fallback. Use `val if val is not None else 0.0` for clarity.

.github/workflows/regression-check.yml
```diff
@@ -0,0 +1,97 @@
...
+      image: ghcr.io/${{ github.repository }}/tt-umd-ci-ubuntu-24.04:latest
+
+    steps:
+      - name: Checkout regression tooling
+        uses: actions/checkout@v6
+        with:
+          sparse-checkout: |
+            tests/microbenchmark/expected/baselines/
+            tests/microbenchmark/tools/summarize_regressions.py
+            tests/microbenchmark/tools/analyze_results.py
```
pjanevskiTT: This sparse-checkout pulls `analyze_results.py` solely so `summarize_regressions.py` can lazy-import `format_throughput` from it. That, in turn, forces installing pandas + psutil in this job's venv (line 61). If `format_throughput` (14 lines) were lifted into a tiny shared module — or just inlined — this job could drop both heavyweight deps and the lazy import discussed at `summarize_regressions.py:224`.

tests/microbenchmark/tools/update_baseline.py
```diff
@@ -0,0 +1,375 @@
...
+
+    if args.dry_run:
+        print(
+            f"--- dry-run: would write the following to {out_path} ---",
+            file=sys.stderr,
+        )
+        sys.stdout.write(yaml_text)
+    else:
+        args.baselines_dir.mkdir(parents=True, exist_ok=True)
+        out_path.write_text(yaml_text)
```
pjanevskiTT, line 376: `out_path.write_text(yaml_text)` is non-atomic — a SIGINT mid-write leaves a half-written YAML. The blast radius is small (calibration script run by a human who can re-run), but a `tempfile.NamedTemporaryFile` in the same dir + `os.replace` is one extra line for a clean abort guarantee. Optional.

tests/microbenchmark/tools/compare_to_baseline.py
```diff
@@ -0,0 +1,151 @@
...
+    p = argparse.ArgumentParser(description=__doc__)
+    p.add_argument(
+        "--results",
+        type=Path,
+        default=os.environ.get("UMD_MICROBENCHMARK_RESULTS_PATH"),
+        help=(
+            "Directory holding local <title>.json files. "
+            "Defaults to $UMD_MICROBENCHMARK_RESULTS_PATH."
+        ),
+    )
```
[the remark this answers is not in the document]
ajovanovicTT: argparse applies type= to string defaults — os.environ.get(...) returning a string is converted to Path.

tests/microbenchmark/tools/summarize_regressions.py
```diff
@@ -0,0 +1,549 @@
...
+        with open(yaml_path) as f:
+            data = yaml.safe_load(f) or {}
+        per_arch_meta = data.get("metadata") or {}
+        arch = per_arch_meta.get("arch")
+        if not arch:
+            print(
+                f"WARN: {yaml_path} has no metadata.arch field; skipping",
+                file=sys.stderr,
+            )
+            continue
```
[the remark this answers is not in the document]
ajovanovicTT: Ok to warn that architectures except n150 are skipped for now.

tests/microbenchmark/tools/update_baseline.py
```diff
@@ -0,0 +1,375 @@
...
+            mape_pct = (result.get("medianAbsolutePercentError(elapsed)") or 0.0) * 100
+            entries.setdefault(title, {})[case] = derive_entry(med, batch, mape_pct)
+    return entries
+
+
+def load_existing_arch_yaml(path: Path) -> dict:
+    """Load a per-arch YAML if it exists; return {} otherwise."""
+    if not path.exists():
+        return {}
+    with open(path) as f:
```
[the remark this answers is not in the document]
ajovanovicTT: Dev-only CLI tool, so operator runs it locally to recalibrate baselines from a CI artifact, with no untrusted input path.

tests/microbenchmark/expected/baselines/wh_n150.yaml
ajovanovicTT: Renaming it to median_value, it actually never shows time, its always some kind of a throughput (in this case its not miliseconds, but expresing clusters/s), so if named just median it wouldn't be adequate ('median' field in benchmark result is showing time).

tests/microbenchmark/tools/update_baseline.py
ajovanovicTT: Wouldn't do this for now. CLI flags don't survive in the repo. Whoever recalibrates next month has no record of what --mape-k 2.5 flag was used last time. When other arches are included, i would go towards moving this to per arch dictionary if different numbers are needed for other architectures. Definitely, I would first observe variability on other runners before making this changes.

tests/microbenchmark/tools/update_baseline.py
ajovanovicTT: "1-byte transfer" round-trips fine because the dash + letters disambiguate it from a number. I think this would be big generalization for now, since none of existing benchmark cases would face a problem now.

tests/microbenchmark/tools/summarize_regressions.py
ajovanovicTT: We will adjust this based on arches from few available dedicated runners.

tests/microbenchmark/tools/update_baseline.py
ajovanovicTT: Outcome is the same, user would just have to rerun the script again in both cases.

tests/microbenchmark/tools/compare_to_baseline.py
ajovanovicTT: Added readme, that reflects current state and tools usage.

tests/microbenchmark/tools/summarize_regressions.py
ajovanovicTT: Won't write test for this in the near future. This should be tested by causing the real regression, and checking if workflow shows the real regression.

.github/workflows/regression-check.yml
ajovanovicTT: If neccessary, the workflow could be manually rerun by rerunning run-benchmarks workflow. Don't see the use case of adding worflow_dispatch to this workflow.

tests/microbenchmark/tools/analyze_results.py
```diff
@@ -13,22 +13,8 @@
...
-    # Otherwise use binary units (KiB, MiB, GiB)
-    if throughput >= 1024**3:
-        return f"{throughput / (1024**3):.2f} GiB/s"
-    elif throughput >= 1024**2:
-        return f"{throughput / (1024**2):.2f} MiB/s"
-    elif throughput >= 1024:
-        return f"{throughput / 1024:.2f} KiB/s"
-    else:
-        return f"{throughput:.2f} B/s"
+sys.path.insert(0, str(Path(__file__).resolve().parent))
```
aleksamarkovicTT: Don't hack PYTHONPATH, instead, create an `__init__.py` file.

.github/workflows/regression-check.yml
```diff
@@ -0,0 +1,97 @@
...
+          # log output published before the step fails.
+          set +e
+          python3 tests/microbenchmark/tools/summarize_regressions.py \
+            --current "${RESULTS_DIR}" \
+            --baselines-dir tests/microbenchmark/expected/baselines \
+            --output "${SUMMARY_PATH}"
+          rc=$?
+          set -e
+          if [ -f "${SUMMARY_PATH}" ]; then
+            cat "${SUMMARY_PATH}" >> "${GITHUB_STEP_SUMMARY}"
```
broskoTT, line 64: What does this block actually do?

.github/workflows/regression-check.yml
```diff
@@ -0,0 +1,97 @@
...
+        run: |
+          python3 -m venv /tmp/venv
+          . /tmp/venv/bin/activate
+          echo "PATH=$PATH" >> $GITHUB_ENV
+          echo "VIRTUAL_ENV=$VIRTUAL_ENV" >> $GITHUB_ENV
+
+      - name: Install python dependencies
+        run: |
+          . /tmp/venv/bin/activate
+          python3 -m pip install --no-cache-dir pyyaml
```
broskoTT: You can run this job in our docker container, umd-ci.
Then, install in its python whatever you need (in docker-install-common for example).
Then you can skip creating and provisioning venv in this job altogether

broskoTT: Actualy you're already using container, yeah then no need to meddle with venv, container will be destroyed anyway after the job

broskoTT: But since you do need pyyaml, add it to the umd-ci docker

.github/workflows/regression-check.yml
```diff
@@ -0,0 +1,97 @@
...
+      - name: Summarize regressions
+        run: |
+          . /tmp/venv/bin/activate
+          # Capture exit code without aborting — summarize_regressions.py exits 1
+          # when a gated case breaches, but we still want the step summary and
+          # log output published before the step fails.
+          set +e
+          python3 tests/microbenchmark/tools/summarize_regressions.py \
+            --current "${RESULTS_DIR}" \
+            --baselines-dir tests/microbenchmark/expected/baselines \
```
broskoTT: So baselines will be committed in the repo? that's fine. Maybe drop one layer of dirs though, `expected/baselines`

.github/workflows/regression-check.yml
```diff
@@ -0,0 +1,97 @@
...
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Summarize regressions
+        run: |
+          . /tmp/venv/bin/activate
+          # Capture exit code without aborting — summarize_regressions.py exits 1
+          # when a gated case breaches, but we still want the step summary and
+          # log output published before the step fails.
+          set +e
+          python3 tests/microbenchmark/tools/summarize_regressions.py \
```
broskoTT, line 57: If the block below is not that important, and if you end up avoiding venv, this step might be a single command, just this one

.github/workflows/build-and-run-all-benchmarks.yml
```diff
@@ -45,10 +45,19 @@ jobs:
...
+        # labels until those dedicated runners come online. Only n150 has a
+        # baseline in tests/microbenchmark/expected/baselines/, so only n150 is
+        # subject to regression-check — the other archs' bench results land in
+        # artifacts but are silently ignored by summarize_regressions.
         test-group: [
-          {arch: wormhole_b0, card: tt-ubuntu-2204-n150-viommu-stable, timeout: 45},
-          {arch: wormhole_b0, card: tt-ubuntu-2204-n300-viommu-stable, timeout: 45},
-          {arch: blackhole, card: tt-ubuntu-2204-p150b-viommu-stable, timeout: 45},
+          {arch: wormhole_b0, card: tt-ubuntu-2204-n150-viommu-stable,
+           runner_label: bgd-lab-06, timeout: 45},
```
broskoTT: The name here can change.
I think we should use labels instead.
You can define something like n150-umd-perf. Then we can add that label to whichever machine we want to use as dedicated for n150 perf

.github/workflows/build-and-run-all-benchmarks.yml
```diff
@@ -45,10 +45,19 @@ jobs:
...
+          {arch: wormhole_b0, card: tt-ubuntu-2204-n300-viommu-stable,
```
broskoTT: Not sure you need both card and runner_label for all of these? They both just represent the string that the runners will request for the machine to run on

tests/microbenchmark/tools/summarize_regressions.py
```diff
@@ -0,0 +1,555 @@
...
+# Python's json parser rejects (it only accepts the capitalized `NaN`).
+_NAN_RE = re.compile(r"-?\bnan\b")
+
+# --- Arch label canonicalization -------------------------------------------------
+# Artifact names look like:
+#   benchmark-json-wormhole_b0-tt-ubuntu-2204-n150-viommu-stable-ubuntu-22.04
+# We canonicalize to short labels matching the per-arch YAML filenames.
+# Substring match keeps this robust to minor variations (e.g. -iter1 suffixes).
+ARCH_PATTERNS = [
+    ("WH n150", ["n150"]),
```
broskoTT: Why do you need WH BH prefix?
Seems like this could be simpler just with a card name

tests/microbenchmark/baselines/bh_p150b.yaml
```diff
@@ -0,0 +1,10 @@
+# metadata:
+#   schema_version: 1
+#   arch: "BH p150b"
```
broskoTT: `b` just explains cooling, I'd drop that suffix. All p150 should behave the same

tests/microbenchmark/tools/formatting.py
```diff
@@ -0,0 +1,26 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+# SPDX-License-Identifier: Apache-2.0
+
+"""Shared formatting helpers for microbenchmark tooling.
```
broskoTT: Maybe this could be renamed to a more general "utils" file?

tests/microbenchmark/tools/compare_to_baseline.py
```diff
@@ -0,0 +1,151 @@
...
+                )
+    return filtered
+
+
+def main() -> int:
+    p = argparse.ArgumentParser(description=__doc__)
+    p.add_argument(
+        "--results",
+        type=Path,
+        default=os.environ.get("UMD_MICROBENCHMARK_RESULTS_PATH"),
```
broskoTT, line 68: I'd prefer avoiding passing this through env variable, not even offer it as an option.
Seems like a clearer solution for it to always be an argument of the script. Please reconsider

tests/microbenchmark/tools/compare_to_baseline.py
```diff
@@ -0,0 +1,151 @@
...
+    export UMD_MICROBENCHMARK_RESULTS_PATH=/tmp/umd-bench
+    mkdir -p "$UMD_MICROBENCHMARK_RESULTS_PATH"
+    ./build/test/umd/microbenchmark/umd_microbenchmark \\
+        --gtest_filter='MicrobenchmarkOpenCluster.ClusterConstructor'
+
+    python3 tests/microbenchmark/tools/compare_to_baseline.py --arch 'WH n150'
+
+The summary is printed to stdout. Exit code is 1 if any `gate: true` case in
+the per-arch baseline YAML breached as DOWN (same gating rule as the CI
+workflow); 0 otherwise.
```
broskoTT, line 25: Can you also provide an example of how the format looks like of the reported results from umd_microbenchmark?
It's hard to follow logic below without it

tests/microbenchmark/tools/update_baseline.py
```diff
@@ -0,0 +1,385 @@
...
+
+
+def arch_slug(arch_label: str) -> str:
+    """Canonical filename slug for a given arch label.
+    "WH n150" -> "wh_n150", "BH p150b" -> "bh_p150b".
+    """
+    return arch_label.lower().replace(" ", "_")
+
+
+def _yaml_escape(s: str) -> str:
```
broskoTT: this could also go to utils that I mentioned in other comment

tests/microbenchmark/tools/update_baseline.py
```diff
@@ -0,0 +1,385 @@
...
+MAPE_K = 1
+
+# nanobench renders unset floating-point fields as bare `-nan`/`nan`, which
+# Python's json parser rejects (it only accepts the capitalized `NaN`).
+_NAN_RE = re.compile(r"-?\bnan\b")
+
+# Arch label -> (yaml filename, dedicated runner hostname). Used to find the
+# destination file under baselines/ and to fill in the runner_hostname
+# metadata field that records which machine produced the JSON output.
+ARCH_RUNNERS: dict[str, tuple[str, str]] = {
```
broskoTT: Can we avoid having this mapping altogether?

The driver should be able to define what is the card it is running on.
So if you run the script on the current machine, it figures out it is running on n150 for example, and it automatically updates just that file from the baseline files?
The machine name it ran on can be present as metadata in the resulting file, but I'd like to avoid having a tight coupling to a specific machine. What that means is that I think nowhere in this PR (except output baseline file) should bgd-lab-06 be referenced for example

broskoTT (commented):
Ping me for any confusing comments, and to look at this on a new iteration

.github/workflows/regression-check.yml
ajovanovicTT: That block writes to github action summary tab and logs from the file that the script was writing to. Script can return 1 on gated breaches, so set +e is used in that case to avoid finishing step, but actually write to summary in that case also.

tests/microbenchmark/tools/analyze_results.py
ajovanovicTT: Just removed this unnecessary syspath insert and since everything is in the same directory that is added to path by default, no init file is needed.

.github/workflows/regression-check.yml
ajovanovicTT: Removed unnecessary parent dir.

.github/workflows/regression-check.yml
ajovanovicTT: Synced offline on this, leaving block as is, just removed venv.

.github/workflows/build-and-run-all-benchmarks.yml
ajovanovicTT: Changed, assigned new n150-umd-perf label to bgd-lab-06 runner.

.github/workflows/build-and-run-all-benchmarks.yml
ajovanovicTT: Removed unecessary runner_label field.

tests/microbenchmark/tools/compare_to_baseline.py
ajovanovicTT: Leaving env var for now as we discussed offline, since bencmark results exporting (populating jsons and htmls) is currently relying on that variable, and this script just reuses path from that variable to find stored local run results to compare them with the baseline.

.github/workflows/regression-check.yml
ajovanovicTT: Removed venv.

tests/microbenchmark/tools/compare_to_baseline.py
ajovanovicTT: Provided json formats (baseline and current microbenchmark run) summary in top comment in summarize_regressions.py and how values from them are used, and since this is just wrapper script provided reference to that comment here. Please let me now if it would be better to add more detailed explanation.

tests/microbenchmark/tools/update_baseline.py
ajovanovicTT: Avoided this mapping and also made update_baseline.py script read arch from added arch field to host specs, so user doesnt have to provide --arch flag manually.

pjanevskiTT (approved):
Wait for other people that reviewed to sign off, I am fine with the changes

.github/workflows/build-and-run-all-benchmarks.yml
```diff
@@ -81,3 +88,17 @@ jobs:
       ubuntu-docker-version: ${{ matrix.ubuntu-docker-version}}
       card: ${{ matrix.test-group.card}}
       baseline-run-id: ${{ inputs.baseline-run-id }}
+
+  # Regression check against in-repo per-arch baseline YAMLs (median calibrated
+  # on the dedicated runner + per-case tolerance). Runs in parallel with
+  # analyze-results, which answers a different question (drift vs. the single
+  # latest main run).
```
broskoTT, line 93: nit: I know that claude always tries to give a lot of context through comments. And often, like here, I find it mistakenly placed. Please remove the parts from this comment which reference how baseline is calculated. Because if we change that algorithm, it seems wrong to have to change the comment in yaml which does regression checking.

.github/workflows/regression-check.yml
broskoTT: thanks for adjusting the comment

.github/workflows/run-benchmarks.yml
```diff
@@ -88,7 +88,8 @@ jobs:
         if: ${{ inputs.arch != 'baremetal' }}
         run: |
             python3 $GITHUB_WORKSPACE/tt-umd/tests/microbenchmark/tools/gather_host_specs.py \
-            --json --output $UMD_MICROBENCHMARK_RESULTS_PATH/machine_host_spec
+            --json --output $UMD_MICROBENCHMARK_RESULTS_PATH/machine_host_spec \
+            --card "${{ inputs.card }}"
```
broskoTT: Can we remove this `--card` argument? It goes along with what we talked about, if this script is ran on a machine which has specific arch, why pass it as an argument

tests/microbenchmark/tools/gather_host_specs.py
```diff
@@ -204,6 +204,10 @@ def gather_specs():
             "Distro": get_distro_pretty_name(),
             "Kernel": platform.release(),
             "Hostname": socket.gethostname(),
+            # CI runner's registered name; unlike Hostname above it isn't the ephemeral container ID. Absent outside CI.
+            "CI_Runner": os.environ.get("RUNNER_NAME", "unknown"),
+            # CI runner/board label (e.g. "n150-umd-perf"); arch is recovered from it downstream.
+            "Card": card,
```
broskoTT: Hmm... I think you can gather the arch and board type here easily. 
You could even call one of UMD's python method for it, like PCIDevice::enumerate_devices_info()[0].get_arch(), or there might be a shortcut for it...

Then if you still want to register the label requested, change the --card argument to --label-requested.
But passing arch and card name here seems unintuitive and error prone

tests/microbenchmark/tools/utils.py
```diff
@@ -0,0 +1,62 @@
...
+
+def load_nanobench_json(path):
+    """Read a nanobench JSON file, normalizing bare `nan`/`-nan` to `NaN`.
+
+    Raises OSError / json.JSONDecodeError on failure for the caller to handle.
+    """
+    return json.loads(_NAN_RE.sub("NaN", path.read_text()))
+
+
+def arch_label_from_string(s):
```
broskoTT: Please try to remove this function. As we talked previously, I'd prefer we don't have any kind of filtering and mapping between arch/board names

tests/microbenchmark/tools/README.md
```diff
@@ -0,0 +1,142 @@
...
+    --current /tmp/all-arch-results \
+    --baselines-dir tests/microbenchmark/baselines \
+    --output /tmp/summary.md
+```
+
+Exit code: `0` always, except when a `gate: true` case breaches DOWN — then
+exit `1` so CI fails. Non-gated breaches are soft alerts; they show up in
+the summary but don't fail the job.
+
+## `compare_to_baseline.py` — legacy local comparator
```
broskoTT: remove if legacy (will still be present in git logs, so don't worry about data loss)

tests/microbenchmark/tools/README.md
```diff
@@ -0,0 +1,142 @@
...
+python3 tests/microbenchmark/tools/update_baseline.py \
+    --from-results-dir /tmp/ci-bench
+
+# inspect the diff, commit the updated <arch>.yaml.
+git diff tests/microbenchmark/baselines/n150.yaml
+git add tests/microbenchmark/baselines/n150.yaml
+git commit -m "Recalibrate n150 baseline from CI run <run-id>"
+```
+
+The arch is inferred from the `Card` field in the artifact's
```
broskoTT: If you adjust from other comments, then remove this

broskoTT (approved):

tests/microbenchmark/tools/gather_host_specs.py
ajovanovicTT: run-benchmarks workflow now uses build with TT_UMD_BUILD_PYTHON=OFF so I think gathering board type here in that way would require turning it ON in build-tests.yml with import tt_umd in this script.

ajovanovicTT: Just installed tt-umd nanobind for this purpose in run-benchmarks.yml before gather host spec step.

tests/microbenchmark/tools/README.md
ajovanovicTT: Refactored comment so it matches current state more accurately.

PR #2694, opened 2026-05-21
Multichip TT-sim UMD support

Riddy21:
## Summary

Rebases the multichip simulation support from [nkapreTT/tt-umd nkapre/multichip](https://github.com/tenstorrent/tt-umd/compare/main...nkapreTT:tt-umd:nkapre/multichip) onto tenstorrent/tt-umd main.

This is the UMD half of the multichip CraqSim integration (see companion tt-metal PR for the other half).

### What this adds

- **Single libttsim dlopen + eth-MAC pre-pass**: enables multi-chip simulation via a single shared library with ethernet MAC address pre-negotiation
- **advance_clock for all chips**: `advance_clock()` ticks all simulated chips via `libttsim_clock_devices`
- **Serialized shared-dlopen I/O and static TLB writes**: prevents races in multi-chip sim paths
- **Mapped simulation sysmem buffers**: support for host memory buffers mapped into simulator address space
- **DRAM teleport via coordinate ABI**: routes DRAM teleport operations through coordinate-based addressing
- **Direct L1 and fabric endpoint hooks**: exposes low-level L1 access and fabric endpoint direction registration
- **Fabric node id registration**: registers fabric node identifiers in the simulator

### Commits

| SHA | Message |
|-----|---------|
| `53d248bc` | feat(sim): multichip via single libttsim dlopen + eth-MAC pre-pass |
| `1e457baf` | feat(sim/v3.5-#8): advance_clock ticks all chips via libttsim_clock_devices |
| `46883949` | feat(sim/v3.5-#9): use libttsim_clock_all_devices in advance_clock |
| `e892fd4b` | fix(sim/v3.5-#16): serialize shared-dlopen I/O and static TLB writes |
| `2d37532a` | feat(sim): support mapped simulation sysmem buffers |
| `32a04b8a` | feat(sim/v3.5-#21): route DRAM teleport through coordinate ABI |

### Notes for reviewers

- All changes are in simulation-specific code paths (`#ifdef TT_METAL_SIMULATOR` guards or `TTSimCommunicator`/`TTSimTTDevice` classes); no impact on hardware paths
- This branch is based on current tt-umd main (`0e00b4e9`)
- Companion PR: tenstorrent/tt-metal (will be linked when created)

### References
- Source branch: https://github.com/tenstorrent/tt-umd/compare/main...nkapreTT:tt-umd:nkapre/multichip
- CraqSim: tenstorrent/craq-sim (private)

### DIFF
--- a/device/api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
+++ b/device/api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
@@ -6,6 +6,10 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <mutex>
+#include <optional>
+#include <utility>
+#include <vector>
 
 #include "umd/device/chip_helpers/sysmem_manager.hpp"
 
@@ -24,6 +28,9 @@ public:
 
     void unpin_or_unmap_sysmem() override;
 
+    void write_to_sysmem(uint16_t channel, const void* src, uint64_t sysmem_dest, uint32_t size) override;
+    void read_from_sysmem(uint16_t channel, void* dest, uint64_t sysmem_src, uint32_t size) override;
+
     std::unique_ptr<SysmemBuffer> allocate_sysmem_buffer(
         size_t sysmem_buffer_size, const bool map_to_noc = false) override;
 
@@ -34,8 +41,21 @@ protected:
     bool init_sysmem(uint32_t num_host_mem_channels) override;
 
 private:
+    struct MappedBuffer {
+        uint64_t base = 0;
+        void* buffer = nullptr;
+        size_t size = 0;
+    };
+
+    std::optional<MappedBuffer> find_mapped_buffer(uint64_t base, uint32_t size);
+    void remove_mapped_buffer(uint64_t base);
+
     uint8_t* system_memory_ = nullptr;
     size_t system_memory_size_ = 0;
+    uint64_t next_mapped_buffer_base_ = 0;
+    std::vector<MappedBuffer> mapped_buffers_;
+    std::vector<std::pair<void*, size_t>> owned_allocations_;
+    std::mutex mapped_buffers_mutex_;
 };
 
 }  // namespace tt::umd
--- a/device/api/umd/device/cluster.hpp
+++ b/device/api/umd/device/cluster.hpp
@@ -84,7 +84,7 @@ struct ClusterOptions {
      * If not provided, the value is determined automatically by determining the max
      * amount of chips connected through one PCIe interface in the ClusterDescriptor.
      */
-    std::optional<uint32_t> num_host_mem_ch_per_mmio_device = 0;
+    std::optional<uint32_t> num_host_mem_ch_per_mmio_device = std::nullopt;
 
     /**
      * If set, this soc descriptor will be used to construct devices on this cluster. If not set, the default soc
@@ -142,7 +142,23 @@ public:
      * - Perform this for each of the chips connected to the system.
      * @param options See documentation of ClusterOptions for explanation of specific arguments.
      */
-    Cluster(ClusterOptions options = {});
+    Cluster() : Cluster(ClusterOptions{}) {}
+    explicit Cluster(ClusterOptions options);
+
+    // Factories that build ClusterOptions inside libtt-umd to avoid cross-DSO STL ABI issues.
+    static std::unique_ptr<Cluster> create_silicon_cluster(
+        std::optional<uint32_t> num_host_mem_ch_per_mmio_device = std::nullopt);
+    static std::unique_ptr<Cluster> create_single_chip_simulation_cluster(
+        uint32_t num_host_mem_ch_per_mmio_device, ChipId chip_id, const char* simulator_path);
+    static std::unique_ptr<Cluster> create_simulation_cluster_with_descriptor(
+        uint32_t num_host_mem_ch_per_mmio_device,
+        const char* simulator_path,
+        const char* sdesc_path,
+        ClusterDescriptor* cluster_descriptor);
+    static std::unique_ptr<Cluster> create_mock_cluster(
+        const char* sdesc_path, ClusterDescriptor* cluster_descriptor = nullptr);
+    static std::unique_ptr<Cluster> create_swemule_cluster(
+        const char* sdesc_path, ClusterDescriptor* cluster_descriptor = nullptr);
     /**
      * Closing the device. Should undo everything that was done in the constructor. Break created links, free memory,
      * leave the device in a state where it can be re-initialized.
--- a/device/api/umd/device/cluster.hpp
+++ b/device/api/umd/device/cluster.hpp
@@ -281,13 +281,7 @@ public:
     void configure_active_ethernet_cores_for_mmio_device(
         ChipId mmio_chip, const std::unordered_set<CoreCoord>& active_eth_cores_per_chip);
 
-#ifdef TT_UMD_BUILD_SIMULATION
-    // Passthroughs to libttsim_switch_register_fabric_* for wiring mock
-    // multichip fabric topologies under craq-sim v3.5.  Only available when
-    // the library is built with TT_UMD_BUILD_SIMULATION=ON.
     void register_sim_fabric_endpoint_direction(ChipId chip_id, uint32_t eth_tile_id, uint32_t direction);
-    void register_sim_fabric_node_id(ChipId chip_id, uint32_t mesh_id, uint32_t fabric_chip_id);
-#endif  // TT_UMD_BUILD_SIMULATION
 
     //---------- Start and stop the device and tensix cores.
 
@@ -613,6 +607,8 @@ public:
     void read_from_sysmem(void* mem_ptr, uint64_t addr, uint16_t channel, uint32_t size, ChipId src_device_id);
 
     void advance_device_execution(ChipId device_id);
+    void sim_arm_launch_watcher(ChipId chip_id, CoreCoord virtual_core, bool is_eth);
+    void register_sim_fabric_node_id(ChipId chip_id, uint32_t mesh_id, uint32_t fabric_chip_id);
 
     /**
      * Query number of memory channels on Host device allocated for a specific device during initialization.
--- a/device/api/umd/device/cluster.hpp
+++ b/device/api/umd/device/cluster.hpp
@@ -281,8 +281,13 @@ public:
     void configure_active_ethernet_cores_for_mmio_device(
         ChipId mmio_chip, const std::unordered_set<CoreCoord>& active_eth_cores_per_chip);
 
+#ifdef TT_UMD_BUILD_SIMULATION
+    // Passthroughs to libttsim_switch_register_fabric_* for wiring mock
+    // multichip fabric topologies under craq-sim v3.5.  Only available when
+    // the library is built with TT_UMD_BUILD_SIMULATION=ON.
     void register_sim_fabric_endpoint_direction(ChipId chip_id, uint32_t eth_tile_id, uint32_t direction);
     void register_sim_fabric_node_id(ChipId chip_id, uint32_t mesh_id, uint32_t fabric_chip_id);
+#endif  // TT_UMD_BUILD_SIMULATION
 
     //---------- Start and stop the device and tensix cores.
 
--- a/device/api/umd/device/cluster.hpp
+++ b/device/api/umd/device/cluster.hpp
@@ -143,6 +143,7 @@ public:
      * @param options See documentation of ClusterOptions for explanation of specific arguments.
      */
     Cluster() : Cluster(ClusterOptions{}) {}
+
     explicit Cluster(ClusterOptions options);
 
     // Factories that build ClusterOptions inside libtt-umd to avoid cross-DSO STL ABI issues.
--- a/device/api/umd/device/cluster.hpp
+++ b/device/api/umd/device/cluster.hpp
@@ -281,6 +281,9 @@ public:
     void configure_active_ethernet_cores_for_mmio_device(
         ChipId mmio_chip, const std::unordered_set<CoreCoord>& active_eth_cores_per_chip);
 
+    void register_sim_fabric_endpoint_direction(ChipId chip_id, uint32_t eth_tile_id, uint32_t direction);
+    void register_sim_fabric_node_id(ChipId chip_id, uint32_t mesh_id, uint32_t fabric_chip_id);
+
     //---------- Start and stop the device and tensix cores.
 
     /**
--- a/device/api/umd/device/simulation/tt_sim_communicator.hpp
+++ b/device/api/umd/device/simulation/tt_sim_communicator.hpp
@@ -183,6 +183,7 @@ private:
     // v3.5 multi-chip ABI. Resolved via dlsym; nullptr if .so is legacy single-chip.
     void *(*pfn_libttsim_create_device_by_id_)(uint32_t chip_id, int chip_x, int chip_y) = nullptr;
     void  (*pfn_libttsim_select_device_by_id_)(uint32_t chip_id) = nullptr;
+    void  (*pfn_libttsim_clock_devices_)(uint32_t n_clocks) = nullptr;
 
     // v3.5 commit #6 — eth-MAC wiring.
     void *dev_handle_ = nullptr;
--- a/device/api/umd/device/simulation/tt_sim_communicator.hpp
+++ b/device/api/umd/device/simulation/tt_sim_communicator.hpp
@@ -205,7 +205,7 @@ private:
 
     // v3.5 commit #6 — eth-MAC wiring.
     void *dev_handle_ = nullptr;
-    void (*pfn_libttsim_switch_reset_)(void) = nullptr;
+    void (*pfn_libttsim_switch_reset_)() = nullptr;
     void (*pfn_libttsim_switch_register_)(void *dev, uint32_t tile_id, uint64_t mac) = nullptr;
     void (*pfn_libttsim_configure_eth_link_virtual_)(void *dev, uint32_t tile_id, uint64_t local_mac) = nullptr;
     void (*pfn_libttsim_switch_register_peer_)(void *dev, uint32_t tile_id, void *peer_dev, uint32_t peer_tile_id) =
@@ -213,7 +213,7 @@ private:
     void (*pfn_libttsim_switch_register_fabric_node_id_)(void *dev, uint32_t mesh_id, uint32_t chip_id) = nullptr;
     void (*pfn_libttsim_switch_register_fabric_endpoint_direction_)(void *dev, uint32_t tile_id, uint32_t direction) =
         nullptr;
-    void (*pfn_libttsim_switch_drain_)(void) = nullptr;
+    void (*pfn_libttsim_switch_drain_)() = nullptr;
 
     // Stored callbacks for DMA memory operations.
     std::function<void(uint64_t, void *, uint32_t)> pci_dma_mem_rd_bytes_callback_;
--- a/device/api/umd/device/simulation/tt_sim_communicator.hpp
+++ b/device/api/umd/device/simulation/tt_sim_communicator.hpp
@@ -26,9 +26,17 @@ public:
      * Constructor for TTSimCommunicator.
      *
      * @param simulator_directory Path to the simulator binary/directory
-     * @param copy_sim_binary If true, copy the simulator binary to memory for security
+     * @param copy_sim_binary If true, copy the simulator binary to memory for
+     *   security AND the loaded .so doesn't support the v3.5 chip-id ABI. If
+     *   the .so exports libttsim_create_device_by_id, v3.5 shared-library
+     *   mode is auto-enabled at initialize() time, ignoring this flag.
+     * @param chip_id Logical chip ID (0..N-1) within the cluster. Only used
+     *   in v3.5 multichip mode (per docs/v3_5_umd_patch_sketch.md). Default 0
+     *   for legacy single-chip consumers.
      */
-    TTSimCommunicator(const std::filesystem::path &simulator_directory, bool copy_sim_binary = false);
+    TTSimCommunicator(const std::filesystem::path &simulator_directory,
+                      bool copy_sim_binary = false,
+                      uint32_t chip_id = 0);
 
     /**
      * Destructor that properly cleans up library handles and file descriptors.
@@ -120,6 +128,13 @@ public:
 
     void start_sim();
 
+    // v3.5 commit #6 — eth-MAC wiring.
+    void *get_dev_handle() const { return dev_handle_; }
+    void switch_reset();
+    void register_eth_endpoint(uint32_t eth_tile_id, uint64_t mac);
+    void switch_drain();
+    void register_peer(uint32_t eth_tile_id, void* peer_dev, uint32_t peer_tile_id);
+
 private:
     // Library management.
     void create_simulator_binary();
@@ -141,6 +156,17 @@ private:
     // Flag to indicate if binary should be copied to memory.
     bool copy_sim_binary_;
 
+    // v3.5 multi-chip mode: when v3.5 symbols are present in the .so, all
+    // TTSimCommunicators sharing the same simulator_directory_ share one
+    // dlopen (held via s_shared_handle_ + refcounted via s_shared_refcount_).
+    // Each communicator records its chip_id_ and calls
+    // libttsim_select_device_by_id(chip_id_) before every I/O.
+    bool v3_5_multichip_mode_ = false;
+    uint32_t chip_id_ = 0;
+    static void *s_shared_handle_;
+    static int s_shared_refcount_;
+    static std::mutex s_shared_init_mutex_;
+
     // Function pointers to simulator library functions.
     void (*pfn_libttsim_init_)() = nullptr;
     void (*pfn_libttsim_exit_)() = nullptr;
@@ -153,6 +179,18 @@ private:
     void (*pfn_libttsim_set_pci_dma_mem_callbacks_)(
         void (*pfn_pci_dma_mem_rd_bytes)(uint64_t paddr, void *p, uint32_t size),
         void (*pfn_pci_dma_mem_wr_bytes)(uint64_t paddr, const void *p, uint32_t size)) = nullptr;
+
+    // v3.5 multi-chip ABI. Resolved via dlsym; nullptr if .so is legacy single-chip.
+    void *(*pfn_libttsim_create_device_by_id_)(uint32_t chip_id, int chip_x, int chip_y) = nullptr;
+    void  (*pfn_libttsim_select_device_by_id_)(uint32_t chip_id) = nullptr;
+
+    // v3.5 commit #6 — eth-MAC wiring.
+    void *dev_handle_ = nullptr;
+    void (*pfn_libttsim_switch_reset_)(void) = nullptr;
+    void (*pfn_libttsim_switch_register_)(void *dev, uint32_t tile_id, uint64_t mac) = nullptr;
+    void (*pfn_libttsim_configure_eth_link_virtual_)(void *dev, uint32_t tile_id, uint64_t local_mac) = nullptr;
+    void (*pfn_libttsim_switch_register_peer_)(void *dev, uint32_t tile_id, void *peer_dev, uint32_t peer_tile_id) = nullptr;
+    void (*pfn_libttsim_switch_drain_)(void) = nullptr;
 
     // Stored callbacks for DMA memory operations.
     std::function<void(uint64_t, void *, uint32_t)> pci_dma_mem_rd_bytes_callback_;
--- a/device/api/umd/device/simulation/tt_sim_communicator.hpp
+++ b/device/api/umd/device/simulation/tt_sim_communicator.hpp
@@ -78,6 +78,14 @@ public:
     void tile_write_bytes(uint32_t x, uint32_t y, uint64_t addr, const void *data, uint32_t size);
 
     /**
+     * Fast simulator-only DRAM access. Returns false when the loaded simulator
+     * does not export the direct DRAM ABI, allowing callers to fall back to the
+     * normal tile/TLB path.
+     */
+    bool dram_read_bytes(uint32_t x, uint32_t y, uint64_t addr, void *data, uint32_t size);
+    bool dram_write_bytes(uint32_t x, uint32_t y, uint64_t addr, const void *data, uint32_t size);
+
+    /**
      * Read data from PCI memory.
      *
      * @param paddr Physical address
@@ -175,6 +183,10 @@ private:
     void (*pfn_libttsim_pci_mem_wr_bytes_)(uint64_t paddr, const void *p, uint32_t size) = nullptr;
     void (*pfn_libttsim_tile_rd_bytes_)(uint32_t x, uint32_t y, uint64_t addr, void *p, uint32_t size) = nullptr;
     void (*pfn_libttsim_tile_wr_bytes_)(uint32_t x, uint32_t y, uint64_t addr, const void *p, uint32_t size) = nullptr;
+    void (*pfn_libttsim_dram_rd_bytes_by_id_)(uint32_t chip_id, uint32_t dram_channel, uint64_t addr, void *p, uint32_t size) = nullptr;
+    void (*pfn_libttsim_dram_wr_bytes_by_id_)(uint32_t chip_id, uint32_t dram_channel, uint64_t addr, const void *p, uint32_t size) = nullptr;
+    void (*pfn_libttsim_dram_core_rd_bytes_by_id_)(uint32_t chip_id, uint32_t x, uint32_t y, uint64_t addr, void *p, uint32_t size) = nullptr;
+    void (*pfn_libttsim_dram_core_wr_bytes_by_id_)(uint32_t chip_id, uint32_t x, uint32_t y, uint64_t addr, const void *p, uint32_t size) = nullptr;
     void (*pfn_libttsim_clock_)(uint32_t n_clocks) = nullptr;
     void (*pfn_libttsim_set_pci_dma_mem_callbacks_)(
         void (*pfn_pci_dma_mem_rd_bytes)(uint64_t paddr, void *p, uint32_t size),
--- a/device/api/umd/device/simulation/tt_sim_communicator.hpp
+++ b/device/api/umd/device/simulation/tt_sim_communicator.hpp
@@ -204,8 +204,10 @@ private:
     static void pci_dma_mem_rd_bytes_wrapper(uint64_t paddr, void *p, uint32_t size);
     static void pci_dma_mem_wr_bytes_wrapper(uint64_t paddr, const void *p, uint32_t size);
 
-    // Thread safety.
-    mutable std::mutex device_lock_;
+    // Thread safety. In v3.5 shared-dlopen mode, libttsim_select_device_by_id()
+    // and the following libttsim I/O call must be serialized across all
+    // communicators because the active device selector is process-global.
+    static std::mutex device_lock_;
 };
 
 }  // namespace tt::umd
--- a/device/api/umd/device/simulation/tt_sim_communicator.hpp
+++ b/device/api/umd/device/simulation/tt_sim_communicator.hpp
@@ -175,6 +175,7 @@ private:
     uint32_t chip_id_ = 0;
     static void *s_shared_handle_;
     static int s_shared_refcount_;
+    static bool s_sim_initialized_;
     static std::mutex s_shared_init_mutex_;
 
     // Function pointers to simulator library functions.
--- a/device/api/umd/device/simulation/tt_sim_communicator.hpp
+++ b/device/api/umd/device/simulation/tt_sim_communicator.hpp
@@ -11,6 +11,10 @@
 #include <functional>
 #include <mutex>
 
+namespace tt {
+enum class ARCH;
+}
+
 namespace tt::umd {
 
 /**
@@ -85,6 +89,14 @@ public:
     bool dram_write_bytes(uint32_t x, uint32_t y, uint64_t addr, const void *data, uint32_t size);
 
     /**
+     * Fast simulator-only L1 access. Returns false when the loaded simulator
+     * does not export the direct L1 ABI, allowing callers to fall back to the
+     * normal TLB path.
+     */
+    bool l1_read_bytes(uint32_t x, uint32_t y, uint64_t addr, void *data, uint32_t size);
+    bool l1_write_bytes(uint32_t x, uint32_t y, uint64_t addr, const void *data, uint32_t size);
+
+    /**
      * Read data from PCI memory.
      *
      * @param paddr Physical address
@@ -145,6 +157,14 @@ public:
     void register_fabric_node_id(uint32_t mesh_id, uint32_t chip_id);
     void register_fabric_endpoint_direction(uint32_t eth_tile_id, uint32_t direction);
 
+    // Shared ttsimd: fetch/create a Device* for any cluster chip_id (idempotent).
+    void *get_or_create_device_handle(uint32_t chip_id, int chip_x, int chip_y);
+    void register_eth_endpoint_on_device(void *dev, uint32_t eth_tile_id, uint64_t mac);
+    void register_peer_on_devices(void *dev, uint32_t eth_tile_id, void *peer_dev, uint32_t peer_tile_id);
+
+    // v3.4-B launch_msg watcher: arm per-cycle GO polling for tensix/erisc dispatch.
+    void arm_launch_watcher_for_noc_core(uint32_t noc_x, uint32_t noc_y, bool is_eth, tt::ARCH arch);
+
 private:
     // Library management.
     void create_simulator_binary();
@@ -175,7 +195,6 @@ private:
     uint32_t chip_id_ = 0;
     static void *s_shared_handle_;
     static int s_shared_refcount_;
-    static bool s_sim_initialized_;
     static std::mutex s_shared_init_mutex_;
 
     // Function pointers to simulator library functions.
@@ -194,6 +213,10 @@ private:
         uint32_t chip_id, uint32_t x, uint32_t y, uint64_t addr, void *p, uint32_t size) = nullptr;
     void (*pfn_libttsim_dram_core_wr_bytes_by_id_)(
         uint32_t chip_id, uint32_t x, uint32_t y, uint64_t addr, const void *p, uint32_t size) = nullptr;
+    void (*pfn_libttsim_l1_rd_bytes_by_id_)(
+        uint32_t chip_id, uint32_t x, uint32_t y, uint64_t addr, void *p, uint32_t size) = nullptr;
+    void (*pfn_libttsim_l1_wr_bytes_by_id_)(
+        uint32_t chip_id, uint32_t x, uint32_t y, uint64_t addr, const void *p, uint32_t size) = nullptr;
     void (*pfn_libttsim_clock_)(uint32_t n_clocks) = nullptr;
     void (*pfn_libttsim_set_pci_dma_mem_callbacks_)(
         void (*pfn_pci_dma_mem_rd_bytes)(uint64_t paddr, void *p, uint32_t size),
@@ -206,15 +229,17 @@ private:
 
     // v3.5 commit #6 — eth-MAC wiring.
     void *dev_handle_ = nullptr;
-    void (*pfn_libttsim_switch_reset_)() = nullptr;
+    void (*pfn_libttsim_switch_reset_)(void) = nullptr;
     void (*pfn_libttsim_switch_register_)(void *dev, uint32_t tile_id, uint64_t mac) = nullptr;
     void (*pfn_libttsim_configure_eth_link_virtual_)(void *dev, uint32_t tile_id, uint64_t local_mac) = nullptr;
     void (*pfn_libttsim_switch_register_peer_)(void *dev, uint32_t tile_id, void *peer_dev, uint32_t peer_tile_id) =
         nullptr;
     void (*pfn_libttsim_switch_register_fabric_node_id_)(void *dev, uint32_t mesh_id, uint32_t chip_id) = nullptr;
-    void (*pfn_libttsim_switch_register_fabric_endpoint_direction_)(void *dev, uint32_t tile_id, uint32_t direction) =
-        nullptr;
-    void (*pfn_libttsim_switch_drain_)() = nullptr;
+    void (*pfn_libttsim_switch_register_fabric_endpoint_direction_)(
+        void *dev, uint32_t tile_id, uint32_t direction) = nullptr;
+    void (*pfn_libttsim_switch_drain_)(void) = nullptr;
+    void (*pfn_libttsim_tensix_arm_launch_watcher_)(void *dev, uint32_t tile_id) = nullptr;
+    void (*pfn_libttsim_erisc_arm_launch_watcher_)(void *dev, uint32_t tile_id) = nullptr;
 
     // Stored callbacks for DMA memory operations.
     std::function<void(uint64_t, void *, uint32_t)> pci_dma_mem_rd_bytes_callback_;
--- a/device/api/umd/device/simulation/tt_sim_communicator.hpp
+++ b/device/api/umd/device/simulation/tt_sim_communicator.hpp
@@ -183,7 +183,7 @@ private:
     // v3.5 multi-chip ABI. Resolved via dlsym; nullptr if .so is legacy single-chip.
     void *(*pfn_libttsim_create_device_by_id_)(uint32_t chip_id, int chip_x, int chip_y) = nullptr;
     void  (*pfn_libttsim_select_device_by_id_)(uint32_t chip_id) = nullptr;
-    void  (*pfn_libttsim_clock_devices_)(uint32_t n_clocks) = nullptr;
+    void  (*pfn_libttsim_clock_all_devices_)(uint32_t n_cycles) = nullptr;
 
     // v3.5 commit #6 — eth-MAC wiring.
     void *dev_handle_ = nullptr;
--- a/device/api/umd/device/simulation/tt_sim_communicator.hpp
+++ b/device/api/umd/device/simulation/tt_sim_communicator.hpp
@@ -142,6 +142,8 @@ public:
     void register_eth_endpoint(uint32_t eth_tile_id, uint64_t mac);
     void switch_drain();
     void register_peer(uint32_t eth_tile_id, void *peer_dev, uint32_t peer_tile_id);
+    void register_fabric_node_id(uint32_t mesh_id, uint32_t chip_id);
+    void register_fabric_endpoint_direction(uint32_t eth_tile_id, uint32_t direction);
 
 private:
     // Library management.
@@ -207,6 +209,9 @@ private:
     void (*pfn_libttsim_switch_register_)(void *dev, uint32_t tile_id, uint64_t mac) = nullptr;
     void (*pfn_libttsim_configure_eth_link_virtual_)(void *dev, uint32_t tile_id, uint64_t local_mac) = nullptr;
     void (*pfn_libttsim_switch_register_peer_)(void *dev, uint32_t tile_id, void *peer_dev, uint32_t peer_tile_id) =
+        nullptr;
+    void (*pfn_libttsim_switch_register_fabric_node_id_)(void *dev, uint32_t mesh_id, uint32_t chip_id) = nullptr;
+    void (*pfn_libttsim_switch_register_fabric_endpoint_direction_)(void *dev, uint32_t tile_id, uint32_t direction) =
         nullptr;
     void (*pfn_libttsim_switch_drain_)(void) = nullptr;
 
--- a/device/api/umd/device/simulation/tt_sim_communicator.hpp
+++ b/device/api/umd/device/simulation/tt_sim_communicator.hpp
@@ -34,9 +34,8 @@ public:
      *   in v3.5 multichip mode (per docs/v3_5_umd_patch_sketch.md). Default 0
      *   for legacy single-chip consumers.
      */
-    TTSimCommunicator(const std::filesystem::path &simulator_directory,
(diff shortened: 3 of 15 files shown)


Riddy21 (commented):
test

device/simulation/tt_sim_communicator.cpp
```diff
@@ -31,11 +31,34 @@
 
 namespace tt::umd {
 
-TTSimCommunicator::TTSimCommunicator(const std::filesystem::path &simulator_directory, bool copy_sim_binary) :
-    simulator_directory_(simulator_directory), copy_sim_binary_(copy_sim_binary) {}
+// v3.5 multi-chip shared-library state. When the loaded libttsim.so exports
```
Riddy21, line 40: **1.1 — Single shared `dlopen` (commit `53d248bc`)**

Previously each chip did its own memfd copy + `dlopen` of `libttsim.so`. That breaks multichip because each `.so` instance has separate process-global state (eth switch, device registry) and chips cannot talk to each other.

When the loaded `.so` exports `libttsim_create_device_by_id` + `libttsim_select_device_by_id`, all `TTSimCommunicator` instances in this process share one `dlopen` via `s_shared_handle_` (refcounted by `s_shared_refcount_`). Each communicator stores its `chip_id_` and calls `libttsim_select_device_by_id(chip_id_)` before every I/O.

**Backward compat:** if v3.5 symbols are absent, the legacy per-chip memfd path is unchanged.

**Not MP-specific:** this is per-process. In an `mpirun` job each rank has its own `s_shared_handle_`.

device/simulation/tt_sim_communicator.cpp
```diff
@@ -44,6 +67,68 @@ TTSimCommunicator::~TTSimCommunicator() {
 void TTSimCommunicator::initialize() {
     std::lock_guard<std::mutex> lock(device_lock_);
 
+    // Probe the .so for v3.5 multichip ABI symbols. We do this before
```
Riddy21, line 76: **v3.5 ABI probe**

Before committing to memfd-vs-direct dlopen, we temporarily `dlopen` the `.so` and `dlsym` for the v3.5 chip-id symbols. If present, we skip memfd entirely (multiple chips must share one dlopen handle).

The probe uses `RTLD_NOLOAD` first so we don't bump the refcount if the library is already loaded.

device/simulation/tt_sim_communicator.cpp
```diff
@@ -56,43 +141,106 @@ void TTSimCommunicator::initialize() {
 
 void TTSimCommunicator::start_sim() {
```
Riddy21, line 160: **Chip registration in shared simulator**

In v3.5 mode:
- `libttsim_init()` runs **once** (first communicator)
- Each subsequent communicator calls `libttsim_create_device_by_id(chip_id_, ...)` and stores the returned `dev_handle_`
- That `dev_handle_` is used later by the Cluster eth-MAC pre-pass for `register_eth_endpoint` / `register_peer`

device/simulation/tt_sim_communicator.cpp
```diff
@@ -31,11 +31,34 @@
...
+std::mutex TTSimCommunicator::s_shared_init_mutex_;
+std::mutex TTSimCommunicator::device_lock_;
+
+TTSimCommunicator::TTSimCommunicator(
+    const std::filesystem::path &simulator_directory, bool copy_sim_binary, uint32_t chip_id) :
+    simulator_directory_(simulator_directory), copy_sim_binary_(copy_sim_binary), chip_id_(chip_id) {}
 
 TTSimCommunicator::~TTSimCommunicator() {
-    if (libttsim_handle_) {
+    if (v3_5_multichip_mode_) {
```
Riddy21, line 54: **Refcounted teardown**

In v3.5 mode `shutdown()` is a no-op (calling `libttsim_exit` per chip would tear down the simulator for all other chips). Instead, the **last** communicator destructor decrements `s_shared_refcount_` to zero, then calls `libttsim_exit()` once and `dlclose`s the shared handle.

device/simulation/tt_sim_communicator.cpp
```diff
@@ -56,43 +141,106 @@ void TTSimCommunicator::initialize() {
...
+        // Defer libttsim_exit until the last communicator destructs (handled
+        // by refcount in the destructor's shared-handle close path). The
+        // simulator is process-global; calling exit per chip would be wrong.
+        return;
+    }
     pfn_libttsim_exit_();
 }
 
+// v3.5: in multi-chip mode, every I/O entry point first selects the right
```
Riddy21, line 191: **Per-I/O chip selection (v3.5-#16)**

`libttsim_select_device_by_id()` sets a **process-global** active device inside the shared `.so`. Every I/O entry point must call it under `device_lock_` (now **static**, not per-instance) so chip A's select+I/O pair cannot be interleaved with chip B's.

device/simulation/tt_sim_communicator.cpp
```diff
@@ -56,43 +141,106 @@ void TTSimCommunicator::initialize() {
...
     std::lock_guard<std::mutex> lock(device_lock_);
+    // pci_config_read32 reads compile-time constants; no chip selection needed.
+    // But we still select for consistency and future-proofing.
+    SELECT_CHIP_IF_NEEDED();
     return pfn_libttsim_pci_config_rd32_(bus_device_function, offset);
 }
 
 void TTSimCommunicator::advance_clock(uint32_t n_clocks) {
     std::lock_guard<std::mutex> lock(device_lock_);
+    if (v3_5_multichip_mode_ && pfn_libttsim_clock_all_devices_) {
```
Riddy21, line 270: **1.3 — `advance_clock` ticks all chips (commits `1e457baf`, `46883949`)**

In legacy single-chip mode, `advance_clock(n)` only ticks the chip currently selected. In multichip that deadlocks: chip A writes an eth packet and polls, but chip B never advances → Fabric Router Sync stuck at `0xa0b0c0d0`.

Fix: in v3.5 mode call `libttsim_clock_all_devices(n)` which walks the simulator's chip_id registry and drains the eth switch in sort order. (Commit #8 initially used the wrong `libttsim_clock_devices` ABI; #9 corrected it.)

device/simulation/tt_sim_communicator.cpp
```diff
@@ -44,6 +67,68 @@ TTSimCommunicator::~TTSimCommunicator() {
...
+        s_shared_refcount_++;
+        libttsim_handle_ = s_shared_handle_;
+        DLSYM_FUNCTION(libttsim_init)
+        DLSYM_FUNCTION(libttsim_exit)
+        DLSYM_FUNCTION(libttsim_pci_config_rd32)
+        DLSYM_FUNCTION(libttsim_pci_mem_rd_bytes)
+        DLSYM_FUNCTION(libttsim_pci_mem_wr_bytes)
+        DLSYM_FUNCTION(libttsim_tile_rd_bytes)
+        DLSYM_FUNCTION(libttsim_tile_wr_bytes)
+        pfn_libttsim_dram_rd_bytes_by_id_ =
```
Riddy21: **1.5 — Optional DRAM teleport ABI (commit `32a04b8a`)**

These four symbols are resolved with raw `dlsym` (not `DLSYM_FUNCTION`) because they are **optional** — older `libttsim` builds won't have them. Callers (`TTSimTTDevice::read/write_to_device`) fall back to the tile/TLB path when `dram_*_bytes` returns false.

device/simulation/tt_sim_communicator.cpp
```diff
@@ -196,6 +344,45 @@ void TTSimCommunicator::load_simulator_library(const std::filesystem::path &path
     DLSYM_FUNCTION(libttsim_set_pci_dma_mem_callbacks)
 }
 
+// v3.5 commit #6 — eth-MAC wiring methods. All no-ops in legacy single-chip mode.
```
Riddy21, line 541: **Eth-MAC wiring helpers (used by Cluster pre-pass)**

- `switch_reset()` — clears libttsim's process-global eth switch (call once)
- `register_eth_endpoint(tile, mac)` — sets `link_mode=Virtual`, writes link-up sentinel, registers MAC
- `register_peer(tile, peer_dev, peer_tile)` — bidirectional peer info for source-aware routing (BH BCAST/MCAST)

All no-ops when `v3_5_multichip_mode_` is false.

device/api/umd/device/simulation/tt_sim_communicator.hpp
```diff
@@ -26,9 +26,16 @@ class TTSimCommunicator final {
...
+     * @param copy_sim_binary If true, copy the simulator binary to memory for
+     *   security AND the loaded .so doesn't support the v3.5 chip-id ABI. If
+     *   the .so exports libttsim_create_device_by_id, v3.5 shared-library
+     *   mode is auto-enabled at initialize() time, ignoring this flag.
+     * @param chip_id Logical chip ID (0..N-1) within the cluster. Only used
+     *   in v3.5 multichip mode (per docs/v3_5_umd_patch_sketch.md). Default 0
+     *   for legacy single-chip consumers.
      */
-    TTSimCommunicator(const std::filesystem::path &simulator_directory, bool copy_sim_binary = false);
+    TTSimCommunicator(
```
Riddy21, line 41: **New `chip_id` parameter**

Each `TTSimChip` in a multichip cluster gets a distinct `chip_id` (0..N-1). The communicator passes it through to `libttsim_create_device_by_id` / `libttsim_select_device_by_id`. Default 0 preserves legacy single-chip callers.

device/api/umd/device/simulation/tt_sim_communicator.hpp
```diff
@@ -165,8 +221,10 @@ class TTSimCommunicator final {
     static void pci_dma_mem_rd_bytes_wrapper(uint64_t paddr, void *p, uint32_t size);
     static void pci_dma_mem_wr_bytes_wrapper(uint64_t paddr, const void *p, uint32_t size);
 
-    // Thread safety.
-    mutable std::mutex device_lock_;
+    // Thread safety. In v3.5 shared-dlopen mode, libttsim_select_device_by_id()
```
Riddy21, line 255: **v3.5-#16: static `device_lock_`**

Changed from per-instance `mutable std::mutex device_lock_` to `static std::mutex device_lock_` because the active device selector inside libttsim is process-global when using shared dlopen. Per-instance mutexes allowed chip A's `select(0)+I/O` to interleave with chip B's `select(1)+I/O`.

device/cluster.cpp
```diff
@@ -408,10 +410,126 @@ Cluster::Cluster(ClusterOptions options) {  // NOLINT(performance-unnecessary-va
                 std::move(tt_device)));
     }
 
+    // -------------------------------------------------------------------
+    // v3.5 commit #6 — eth-MAC wiring pre-pass.
```
Riddy21, line 421: **1.2 — Eth-MAC wiring pre-pass (commit `53d248bc`)**

Runs after all chips are constructed (each has a `dev_handle_`) but before `construct_cluster()`. Walks `cluster_desc->get_ethernet_connections()` from the mock cluster YAML and wires libttsim's process-global eth switch.

**Without this:** eth tiles boot with `link_mode=Down`, cross-chip fabric handshakes never converge.

**Gate:** only for `ChipType::SIMULATION` + `.so` simulator path (v3.5 shared-library mode).

device/cluster.cpp
```diff
@@ -408,10 +410,126 @@ Cluster::Cluster(ClusterOptions options) {  // NOLINT(performance-unnecessary-va
...
+                return nullptr;
+            }
+            auto* sim_chip = dynamic_cast<tt::umd::TTSimChip*>(it->second.get());
+            if (!sim_chip) {
+                return nullptr;
+            }
+            auto* sim_tt = dynamic_cast<tt::umd::TTSimTTDevice*>(sim_chip->get_tt_device());
+            return sim_tt ? sim_tt->get_communicator() : nullptr;
+        };
+        auto eth_sim_mac = [](ChipId cid, int chan) -> uint64_t {
```
Riddy21, line 439: **Synthetic MAC assignment**

Deterministic locally-administered MAC: `0x021E52000000 | (chip_id << 8) | channel`. Both endpoints can compute each other's MAC from the cluster descriptor alone — no hardware MACs in sim.

device/cluster.cpp
```diff
@@ -408,10 +410,126 @@ Cluster::Cluster(ClusterOptions options) {  // NOLINT(performance-unnecessary-va
...
+                    continue;
+                }
+                auto* ca = get_comm(chip_a);
+                auto* cb = get_comm(chip_b);
+                if (!ca || !cb) {
+                    continue;
+                }
+                uint64_t mac_a = eth_sim_mac(chip_a, chan_a);
+                uint64_t mac_b = eth_sim_mac(chip_b, chan_b);
+                ca->register_eth_endpoint(uint32_t(chan_a), mac_a);
```
Riddy21: **Per-link wiring**

For each undirected eth edge (visited once via `chip_a >= chip_b` skip):
1. `register_eth_endpoint` on both sides — virtual link + MAC in magic switch
2. `register_peer` in **both** directions — peer `Device*` + tile for source-aware routing

`if (!ca || !cb) continue` — in MP runs, a chip may belong to another MPI rank with no local communicator; that rank wires its own edges.

device/cluster.cpp
```diff
@@ -408,10 +410,126 @@ Cluster::Cluster(ClusterOptions options) {  // NOLINT(performance-unnecessary-va
...
+                    mac_a,
+                    chip_b,
+                    chan_b,
+                    mac_b);
+            }
+        }
+    }
+
     construct_cluster(options.num_host_mem_ch_per_mmio_device.value(), options.chip_type);
+    options_ = std::move(options);
```
Riddy21, line 536: **Constructor fix:** `options_` is now move-assigned at end of construction instead of copied at the start, avoiding a redundant copy of `ClusterOptions` mid-init.

device/cluster.cpp
```diff
@@ -408,10 +410,126 @@ Cluster::Cluster(ClusterOptions options) {  // NOLINT(performance-unnecessary-va
...
     log_info(LogUMD, "Cluster constructor completed.");
 }
 
+std::unique_ptr<Cluster> Cluster::create_silicon_cluster(std::optional<uint32_t> num_host_mem_ch_per_mmio_device) {
```
Riddy21, line 540: **1.7 — Cluster factories (commit `59de0b48`)**

Static factories build `ClusterOptions` **inside libtt-umd** so callers (e.g. tt-metal in a separate `.so`) pass only scalars and C-strings. Passing `ClusterOptions` by value across DSO boundaries corrupts the heap due to mismatched STL ABI for `std::string` / `std::optional` / `std::filesystem::path`.

device/api/umd/device/cluster.hpp
```diff
@@ -84,7 +84,7 @@ struct ClusterOptions {
      * If not provided, the value is determined automatically by determining the max
      * amount of chips connected through one PCIe interface in the ClusterDescriptor.
      */
-    std::optional<uint32_t> num_host_mem_ch_per_mmio_device = 0;
+    std::optional<uint32_t> num_host_mem_ch_per_mmio_device = std::nullopt;
```
Riddy21, line 87: **Default fix:** changed from `0` to `std::nullopt`.

With `0`, sim open forced zero host mem channels. `std::nullopt` triggers auto-detection from the cluster descriptor (max chips per MMIO), matching the docstring.

device/api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
```diff
@@ -34,8 +41,21 @@ class SimulationSysmemManager : public SysmemManager {
     bool init_sysmem(uint32_t num_host_mem_channels) override;
 
 private:
+    struct MappedBuffer {
```
Riddy21, line 44: **1.6 — Mapped simulation sysmem (commit `2d37532a`)**

`MappedBuffer` tracks host buffers assigned synthetic device-side sysmem addresses (`pcie_base_ + base`). When libttsim DMA callbacks fire with a paddr in this range, `write/read_from_sysmem` memcpy directly instead of going through the hugepage channel layout.

Previously `allocate_sysmem_buffer` / `map_sysmem_buffer` returned `nullptr`.

device/chip_helpers/simulation_sysmem_manager.cpp
```diff
@@ -76,14 +102,81 @@ void SimulationSysmemManager::unpin_or_unmap_sysmem() {
...
     size_t sysmem_buffer_size, const bool map_to_noc) {
-    return nullptr;
+    void* mapping =
+        mmap(nullptr, sysmem_buffer_size, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE | MAP_POPULATE, -1, 0);
+    UMD_ASSERT(mapping != MAP_FAILED, error::RuntimeError, "Simulation sysmem buffer mmap() failed");
+    owned_allocations_.push_back({mapping, sysmem_buffer_size});
+    return map_sysmem_buffer(mapping, sysmem_buffer_size, map_to_noc);
 }
 
 std::unique_ptr<SysmemBuffer> SimulationSysmemManager::map_sysmem_buffer(
```
Riddy21, line 162: **Synthetic IOVA assignment**

Bump-allocates addresses above the hugepage arena (`next_mapped_buffer_base_` starts at `system_memory_size_`). Creates a sim-only `SysmemBuffer` with `device_io_addr = pcie_base_ + mapped_base` and an `unmap_callback` that removes the entry from `mapped_buffers_` on destruction.

device/chip_helpers/sysmem_buffer.cpp
```diff
@@ -45,6 +46,23 @@ SysmemBuffer::SysmemBuffer(TTDevice* tt_device, void* buffer_va, size_t buffer_s
     TracyAllocN(buffer_va_, mapped_buffer_size_, "SysmemBuffer");
 }
 
+SysmemBuffer::SysmemBuffer(
```
Riddy21, line 49: **Simulation-only `SysmemBuffer` constructor**

No PCI/KMD involvement — just records a pre-assigned synthetic IOVA. Used by `SimulationSysmemManager::map_sysmem_buffer`. On destruction, invokes `unmap_callback_` instead of `pci_device_->unmap_for_dma`.

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -55,7 +79,13 @@ TTSimTTDevice::TTSimTTDevice(
     ChipId chip_id,
     bool copy_sim_binary,
     int num_host_mem_channels) :
-    communicator_(std::make_unique<TTSimCommunicator>(simulator_directory, copy_sim_binary)),
+    TTDevice(architecture_implementation::create(soc_descriptor.arch)),
```
Riddy21, line 105: **Chip ID wire-up + base class init**

- Passes `chip_id` to `TTSimCommunicator` for v3.5 shared-dlopen mode
- Calls `TTDevice(architecture_implementation::create(soc_descriptor.arch))` — TTSim doesn't go through PCI probe, so without this `arch` stays `Invalid` downstream

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -33,6 +35,28 @@ namespace tt::umd {
 
 static_assert(!std::is_abstract<TTSimTTDevice>(), "TTSimChip must be non-abstract.");
 
+namespace {
+
+bool sim_dram_teleport_enabled() {
```
Riddy21: **1.5 — DRAM teleport opt-in**

Gated by `TT_METAL_SIMULATOR_DRAM_TELEPORT=1`. When enabled and the target core is a translated DRAM core, uses `libttsim_dram_core_{rd,wr}_bytes_by_id` (bypasses tile/TLB). Falls back to tile path if the ABI isn't in the loaded `.so`.

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -144,8 +174,21 @@ std::unique_ptr<TlbWindow> TTSimTTDevice::get_io_window(tlb_data config, TlbMapp
 
 TTSimTTDevice::~TTSimTTDevice() { communicator_->shutdown(); }
 
+void TTSimTTDevice::start_device() {}
```
Riddy21, line 207: **Lifecycle hooks**

`close_device()` / destructor call `communicator_->shutdown()`. In v3.5 multichip mode `shutdown()` is intentionally a no-op — actual `libttsim_exit` runs when the last communicator's refcount hits zero.

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -155,6 +198,15 @@ void TTSimTTDevice::write_to_device(const void* mem_ptr, tt_xy_pair core, uint64
 
 void TTSimTTDevice::read_from_device(void* mem_ptr, tt_xy_pair core, uint64_t addr, size_t size) {
     std::lock_guard<std::recursive_mutex> lock(device_lock);
+    if (sim_dram_teleport_enabled()) {
+        if (is_translated_dram_core(get_soc_descriptor(), core)) {
+            if (!communicator_->dram_read_bytes(core.x, core.y, addr, mem_ptr, size)) {
+                communicator_->tile_read_bytes(core.x, core.y, addr, mem_ptr, size);
+            }
+            communicator_->advance_clock(10);
```
Riddy21, line 247: **Explicit clock advance (related)**

`TTSimTTDevice::advance_device_execution()` (unchanged in this diff) delegates to `communicator_->advance_clock(1)`. In v3.5 mode that calls `libttsim_clock_all_devices`, ticking all chips together.

The DRAM read path on this line also calls `advance_clock(10)` after dram teleport reads to keep peer chips advancing during polling loops.

Riddy21:
## Multichip sim review guide

Posted inline comments on each change area. Quick map:

| Commit | Area | Key file |
|--------|------|----------|
| `53d248bc` | Shared dlopen + eth-MAC pre-pass | `tt_sim_communicator.cpp`, `cluster.cpp` |
| `1e457baf` / `46883949` | `advance_clock` ticks all chips | `tt_sim_communicator.cpp` |
| `e892fd4b` | Static `device_lock_` serialization | `tt_sim_communicator.hpp` |
| `2d37532a` | Mapped simulation sysmem | `simulation_sysmem_manager.*`, `sysmem_buffer.cpp` |
| `32a04b8a` | DRAM teleport ABI | `tt_sim_communicator.cpp`, `tt_sim_tt_device.cpp` |
| `59de0b48` | Cluster factories (cross-DSO ABI) | `cluster.hpp`, `cluster.cpp` |

All changes are simulation-only (`TTSimCommunicator` / `TTSimTTDevice` / `ChipType::SIMULATION` paths).

device/simulation/tt_sim_communicator.cpp
Riddy21: Need to check this over, still not 100% on this

Riddy21:
## Build fix: `#ifdef TT_UMD_BUILD_SIMULATION` guards

Root cause of the linker failures:

`cluster.cpp` included `tt_sim_chip.hpp`, `tt_sim_tt_device.hpp`, and `tt_sim_communicator.hpp` unconditionally.  These headers declare `TTSimChip`, `TTSimTTDevice`, and `TTSimCommunicator` — whose implementations are only compiled when `TT_UMD_BUILD_SIMULATION=ON`.  The standard Release CI build doesn't set that flag, so the symbols were declared but never defined, giving:

```
lib/libtt-umd.so: undefined reference to `typeinfo for tt::umd::TTSimTTDevice'
lib/libtt-umd.so: undefined reference to `tt::umd::TTSimCommunicator::register_fabric_node_id(...)'
lib/libtt-umd.so: undefined reference to `tt::umd::TTSimCommunicator::switch_reset()'
... (7 symbols total)
```

**Fixes committed:**

| Commit | Change |
|--------|--------|
| `279ac559` | Wrap simulation includes + eth-MAC wiring block + `register_sim_fabric_*` methods in `cluster.cpp` with `#ifdef TT_UMD_BUILD_SIMULATION` |
| `279ac559` | Wrap `register_sim_fabric_*` declarations in `cluster.hpp` with `#ifdef TT_UMD_BUILD_SIMULATION` |
| `b8c1da02` | `clang-format` fix for `tt_sim_communicator.cpp` (pre-commit hook) |

The guards follow the existing pattern already used for the simulation device construction block at `cluster.cpp:349`.  Hardware builds are unaffected; simulation builds (with `TT_UMD_BUILD_SIMULATION=ON`) see no change in behaviour.

Riddy21:
## Code Review Notes — Multichip Simulation PR

### Issues Fixed (pushed as new commits)

1. **`s_init_called` never reset on teardown** (HIGH severity)
   - The function-local `static bool s_init_called` in `start_sim()` was never reset when the last `TTSimCommunicator` destructed. If the simulator was torn down and re-created within the same process, `libttsim_init_()` would be skipped.
   - **Fix**: Replaced with class-level static `s_sim_initialized_` that resets alongside `s_shared_handle_` in the destructor.
   - Commit: `1a87ec93`

2. **`sim_dram_teleport_enabled()` calls `std::getenv` on every I/O** (MEDIUM severity)
   - Performance: every `write_to_device`/`read_from_device` call was doing a `getenv` lookup.
   - **Fix**: Cached result in function-local static.
   - Commit: `b294f0cb`

### Issues Flagged — Not Auto-Fixed

3. **`callback_instance_` is process-global, only last chip's DMA callbacks work** (HIGH severity, pre-existing design flaw amplified by multi-chip)
   - `set_pcie_dma_mem_callbacks` sets the static `callback_instance_ = this`. In multi-chip mode, each chip overwrites the previous, so only the last-initialized chip receives correct DMA read/write callbacks. Earlier chips' sysmem will not be reached by the C-style callback wrappers.
   - **Why not fixed**: The `pfn_libttsim_set_pci_dma_mem_callbacks_` C API does not accept a context/user-data pointer, so there is no way to route callbacks back to the correct instance without either (a) a libttsim ABI change adding a context parameter, or (b) per-chip trampoline functions generated at runtime. Both require coordination with the libttsim side.
   - **Mitigation**: In practice, if the v3.5 simulator internally dispatches DMA to the currently-selected device, the callback may coincidentally invoke the correct sysmem path when the select+I/O pair is serialized under `device_lock_`. But this is fragile and worth fixing properly in a follow-up with a per-chip callback dispatch table.

### Review Verdict

The rest of the PR looks solid:
- **Thread safety**: Lock ordering (`device_lock_` → `s_shared_init_mutex_`) is consistent. All I/O paths serialize under `device_lock_`. `mapped_buffers_mutex_` properly protects the mapped buffer vector.
- **Memory safety**: No leaks or UAF found. `owned_allocations_` properly tracks mmap'd buffers and `unpin_or_unmap_sysmem` cleans up. `SysmemBuffer` destructor correctly invokes `unmap_callback_`.
- **ABI stability**: Simulation-only methods in `cluster.hpp` are properly guarded with `#ifdef TT_UMD_BUILD_SIMULATION`. Non-simulation builds see no ABI change.
- **dlsym patterns**: New manual `dlsym` calls for `register_fabric_*` use `reinterpret_cast` correctly. The existing `DLSYM_FUNCTION` macro uses C-style casts but is pre-existing and NOLINT'd.
- **Factory methods**: `create_simulation_cluster_with_descriptor` follows the same pattern as `create_mock_cluster`/`create_swemule_cluster`.
- **`options_ = std::move(options)` placement**: Safe — `construct_cluster()` takes its needed values by-value params, and the move happens after all reads of `options`.

device/api/umd/device/cluster.hpp
```diff
@@ -142,7 +142,24 @@ class Cluster {
...
      * @param options See documentation of ClusterOptions for explanation of specific arguments.
      */
-    Cluster(ClusterOptions options = {});
+    Cluster() : Cluster(ClusterOptions{}) {}
+
+    explicit Cluster(ClusterOptions options);
+
+    // Factories that build ClusterOptions inside libtt-umd to avoid cross-DSO STL ABI issues.
+    static std::unique_ptr<Cluster> create_silicon_cluster(
+        std::optional<uint32_t> num_host_mem_ch_per_mmio_device = std::nullopt);
```
pjanevskiTT, line 151: The factory still exposes `std::optional<uint32_t>` across the DSO boundary. The commit message claims these factories "avoid cross-DSO STL ABI issues" — but `std::optional<uint32_t>` is just as STL-exposed as the previous direct constructor. Either change to `const uint32_t* num_host_mem_ch_per_mmio_device = nullptr` (or a sentinel), or drop the factory and say the silicon path doesn't need the workaround.

device/api/umd/device/cluster.hpp
pjanevskiTT, line 87: Subtle behavior change: the default flips from `0` to `std::nullopt`. Previously `ClusterOptions{}` left `num_host_mem_ch_per_mmio_device` set to `0`; now it triggers the auto-calculate path in `construct_cluster`. Existing callers that pass `{}` will see different behavior than before — worth calling out in the PR description.

device/api/umd/device/cluster.hpp
```diff
@@ -142,7 +142,24 @@ class Cluster {
      * - Perform this for each of the chips connected to the system.
      * @param options See documentation of ClusterOptions for explanation of specific arguments.
      */
-    Cluster(ClusterOptions options = {});
+    Cluster() : Cluster(ClusterOptions{}) {}
+
+    explicit Cluster(ClusterOptions options);
```
pjanevskiTT, line 147: Making the `ClusterOptions` constructor `explicit` is a source-breaking change for any caller that wrote `Cluster c = some_options;` or relied on implicit conversion in argument-passing. Probably fine to do, but please call it out in the PR description as an API change.

device/cluster.cpp
```diff
@@ -408,10 +415,137 @@ Cluster::Cluster(ClusterOptions options) {  // NOLINT(performance-unnecessary-va
...
+            auto* sim_chip = dynamic_cast<tt::umd::TTSimChip*>(it->second.get());
+            if (!sim_chip) {
+                return nullptr;
+            }
+            auto* sim_tt = dynamic_cast<tt::umd::TTSimTTDevice*>(sim_chip->get_tt_device());
+            return sim_tt ? sim_tt->get_communicator() : nullptr;
+        };
+        // Deterministic per-(chip, channel) MAC: top byte is OUI, next byte
+        // encodes chip_id, bottom byte encodes channel.
+        auto eth_sim_mac = [](ChipId cid, int chan) -> uint64_t {
```
pjanevskiTT, line 439: The MAC encoding `(cid & 0xFF) << 8 | (chan & 0xFF)` caps at 256 chips × 256 channels. Wormhole/blackhole racks can exceed 256 chips per session in CI. At minimum add `UMD_ASSERT(cid <= 0xFF && chan <= 0xFF, ...)` so a future regression fails loudly instead of silently aliasing two distinct links to the same MAC. The remaining 24 bits of the OUI-local 48-bit MAC are unused — could be reallocated for headroom.

device/cluster.cpp
```diff
@@ -408,10 +415,137 @@ Cluster::Cluster(ClusterOptions options) {  // NOLINT(performance-unnecessary-va
...
+                    chip_b,
+                    chan_b,
+                    mac_b);
+            }
+        }
+    }
+#endif  // TT_UMD_BUILD_SIMULATION
+
     construct_cluster(options.num_host_mem_ch_per_mmio_device.value(), options.chip_type);
+    options_ = std::move(options);
```
pjanevskiTT, line 536: Moving `options_ = std::move(options);` from the top of the constructor to the end changes exception semantics: any `UMD_THROW` between lines 327 and 491 now destroys a partially-constructed `Cluster` whose `options_` is default-initialized rather than holding the user's options. There's currently no destructor or partial-state recovery that reads `options_`, so this is latent — but it changes the previous invariant. Either restore the early assignment (and drop the late one), or document why the move was needed.

device/chip_helpers/simulation_sysmem_manager.cpp
```diff
@@ -76,14 +102,81 @@ void SimulationSysmemManager::unpin_or_unmap_sysmem() {
...
+    SysmemManager::read_from_sysmem(channel, dest, sysmem_src, size);
+}
+
 std::unique_ptr<SysmemBuffer> SimulationSysmemManager::allocate_sysmem_buffer(
     size_t sysmem_buffer_size, const bool map_to_noc) {
-    return nullptr;
+    void* mapping =
+        mmap(nullptr, sysmem_buffer_size, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE | MAP_POPULATE, -1, 0);
+    UMD_ASSERT(mapping != MAP_FAILED, error::RuntimeError, "Simulation sysmem buffer mmap() failed");
+    owned_allocations_.push_back({mapping, sysmem_buffer_size});
```
pjanevskiTT, line 158: `owned_allocations_.push_back` is not guarded by `mapped_buffers_mutex_`, but `unpin_or_unmap_sysmem()` iterates and clears `owned_allocations_` while holding the lock, and `map_sysmem_buffer` (below) also pushes to `mapped_buffers_` under the same mutex. Concurrent `allocate_sysmem_buffer` + teardown (or two allocates) will race on `owned_allocations_`. Move this push into the same `std::lock_guard` block that `map_sysmem_buffer` uses — or rename the mutex to cover both vectors and document the invariant.

device/chip_helpers/simulation_sysmem_manager.cpp
```diff
@@ -76,14 +102,81 @@ void SimulationSysmemManager::unpin_or_unmap_sysmem() {
...
+        std::lock_guard<std::mutex> lock(mapped_buffers_mutex_);
+        mapped_base = align_up(next_mapped_buffer_base_, page_size);
+        next_mapped_buffer_base_ = mapped_base + mapped_size;
+        mapped_buffers_.push_back({mapped_base, buffer, sysmem_buffer_size});
+    }
+
+    const uint64_t device_io_addr = pcie_base_ + mapped_base;
+    std::optional<uint64_t> noc_addr = map_to_noc ? std::optional<uint64_t>(device_io_addr) : std::nullopt;
+    return std::make_unique<SysmemBuffer>(buffer, sysmem_buffer_size, device_io_addr, noc_addr, [this, mapped_base]() {
+        remove_mapped_buffer(mapped_base);
```
pjanevskiTT, line 178: The unmap callback captures `[this, mapped_base]` but `SysmemBuffer`'s lifetime is decoupled from `SimulationSysmemManager`'s. `~SysmemBuffer()` (`sysmem_buffer.cpp:181`) invokes this lambda; if the manager has already been destroyed (or `unpin_or_unmap_sysmem` has run), `this` dangles → UB. Two options:

1. Make `SimulationSysmemManager` hold a `std::shared_ptr<MappedBufferRegistry>` (or just `mutex` + `vector`) and capture a `weak_ptr` in the lambda; the unmap becomes a no-op after teardown.
2. Document loudly that all `SysmemBuffer`s returned from this manager must be destroyed before the manager itself.

Option 1 is more robust — `unpin_or_unmap_sysmem` already clears `mapped_buffers_`, so calling `remove_mapped_buffer` after teardown is racing with a no-op even on the happy path.

device/chip_helpers/simulation_sysmem_manager.cpp
```diff
@@ -6,7 +6,16 @@
...
+#include <sys/stat.h>  // for fstat
+#include <unistd.h>
+
+#include <algorithm>
+#include <cstddef>
+#include <cstdint>
+#include <cstring>
+#include <filesystem>
+#include <fstream>
+#include <iostream>
```
pjanevskiTT, line 18: nit: this `#include <iostream>` doesn't appear to be used anywhere in the file (no `std::cout`/`std::cerr` in the diff). Safe to drop.

device/api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
```diff
@@ -34,8 +41,21 @@ class SimulationSysmemManager : public SysmemManager {
...
+        uint64_t base = 0;
+        void* buffer = nullptr;
+        size_t size = 0;
+    };
+
+    std::optional<MappedBuffer> find_mapped_buffer(uint64_t base, uint32_t size);
```
pjanevskiTT, line 50: `find_mapped_buffer` requires the caller to hold `mapped_buffers_mutex_` (the two callers in `read_from_sysmem`/`write_to_sysmem` do), but the invariant isn't documented anywhere. Suggest renaming to `find_mapped_buffer_locked` or adding a `// caller must hold mapped_buffers_mutex_` comment so future callers don't accidentally bypass it.

Separately, this is O(N) per read/write; consider a sorted structure (e.g. `std::map<base, MappedBuffer>` + `upper_bound`) if many buffers end up mapped.

device/simulation/tt_sim_communicator.cpp
```diff
@@ -44,6 +69,74 @@ TTSimCommunicator::~TTSimCommunicator() {
 void TTSimCommunicator::initialize() {
     std::lock_guard<std::mutex> lock(device_lock_);
 
+    // Probe the .so for v3.5 multichip ABI symbols. We do this before
+    // committing to memfd-vs-direct dlopen because the shared-library path
+    // skips memfd entirely (multiple chips share one dlopen).
+    bool v3_5_supported = false;
+    {
+        // Temporarily dlopen just to check symbols; close after probe.
+        void *probe = dlopen(simulator_directory_.c_str(), RTLD_LAZY | RTLD_NOLOAD);
```
pjanevskiTT, line 82: This probe sequence (`dlopen(RTLD_NOLOAD)` → if-null fresh `dlopen(RTLD_LAZY)` → `dlclose` → real `dlopen` later in `initialize()`) can run the .so's global constructors and then immediately tear them down via `dlclose`. For a sim library with non-idempotent global state (file/socket opens in ctors, register allocation, etc.) this is dangerous. Safer alternatives:
- Check `s_shared_handle_` first; if already set, use it for the probe.
- If you must `dlopen` fresh, keep the handle on success and skip the second `dlopen` below — no double-init.

device/simulation/tt_sim_communicator.cpp
```diff
@@ -44,6 +69,74 @@ TTSimCommunicator::~TTSimCommunicator() {
...
+        DLSYM_FUNCTION(libttsim_set_pci_dma_mem_callbacks)
+        DLSYM_FUNCTION(libttsim_create_device_by_id)
+        DLSYM_FUNCTION(libttsim_select_device_by_id)
+        DLSYM_FUNCTION(libttsim_clock_all_devices)
+        DLSYM_FUNCTION(libttsim_switch_reset)
+        DLSYM_FUNCTION(libttsim_switch_register)
+        DLSYM_FUNCTION(libttsim_switch_drain)
+        DLSYM_FUNCTION(libttsim_configure_eth_link_virtual)
+        DLSYM_FUNCTION(libttsim_switch_register_peer)
+        pfn_libttsim_switch_register_fabric_node_id_ =
```
pjanevskiTT, line 138: The contract between "v3.5 mode" and "which symbols must exist" is inconsistent. `libttsim_switch_drain`, `libttsim_configure_eth_link_virtual`, and `libttsim_switch_register_peer` are required (`DLSYM_FUNCTION` throws on missing), while `libttsim_switch_register_fabric_node_id` and `_endpoint_direction` here are optional (`dlsym` + nullptr-check at call site). If the asymmetry is intentional, document the v3.5 ABI contract somewhere (header comment or a `docs/simulation.md`). Otherwise make them all optional with graceful degradation, or all required.

device/simulation/tt_sim_communicator.cpp
```diff
@@ -196,6 +351,61 @@ void TTSimCommunicator::load_simulator_library(const std::filesystem::path &path
...
+    // link-up sentinel + registers MAC. Falls back to switch_register if
+    // configure_eth_link_virtual is not exported.
+    if (pfn_libttsim_configure_eth_link_virtual_) {
+        pfn_libttsim_configure_eth_link_virtual_(dev_handle_, eth_tile_id, mac);
+    } else if (pfn_libttsim_switch_register_) {
+        pfn_libttsim_switch_register_(dev_handle_, eth_tile_id, mac);
+    }
+}
+
+void TTSimCommunicator::register_peer(uint32_t eth_tile_id, void *peer_dev, uint32_t peer_tile_id) {
```
pjanevskiTT, line 565: `register_peer` guards `this->dev_handle_` but not `peer_dev`. In the current code path the peer's `dev_handle_` is always populated before this is called (every `TTSimTTDevice` ctor calls `start_sim()`), but the contract isn't documented and a future caller could violate it. Worth adding `if (!peer_dev) return;` — passing a nullptr peer to `libttsim_switch_register_peer` is presumably a hard crash inside the .so.

device/api/umd/device/simulation/tt_sim_communicator.hpp
```diff
@@ -165,8 +227,10 @@ class TTSimCommunicator final {
...
+    // and the following libttsim I/O call must be serialized across all
+    // communicators because the active device selector is process-global.
+    static std::mutex device_lock_;
```
pjanevskiTT, line 258: Promoting `device_lock_` from per-instance to `static` is the right thing for v3.5 shared-dlopen mode (the chip selector is process-global), but it also globally serializes every libttsim call across communicators in **legacy** single-chip mode. Multiple independent legacy-mode TTSim chips in one process now contend for one lock for every read/write. If multi-legacy in one process is out of scope, please document that here; otherwise consider splitting the lock — per-instance for legacy mode, static for v3.5 mode.

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -144,8 +176,21 @@ std::unique_ptr<TlbWindow> TTSimTTDevice::get_io_window(tlb_data config, TlbMapp
 
 TTSimTTDevice::~TTSimTTDevice() { communicator_->shutdown(); }
 
+void TTSimTTDevice::start_device() {}
+
+void TTSimTTDevice::close_device() { communicator_->shutdown(); }
```
pjanevskiTT, line 209: `close_device()` calls `communicator_->shutdown()`, which in **legacy** mode invokes `libttsim_exit` immediately — so any subsequent I/O via this device crashes inside the .so. In **v3.5 multichip** mode `shutdown()` is a no-op (the destructor handles cleanup via the shared refcount), so I/O after `close_device()` keeps working until the last communicator destructs. This asymmetry is hidden from callers. Suggest adding a `closed_` flag (set here, checked in `read_from_device`/`write_to_device`) so the contract is uniform.

device/api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
pjanevskiTT, line 44: could we maybe implement this in a separate PR? Seems that we can split it, would be easier to review/integrate. Will still review this one, just as a comment

device/api/umd/device/simulation/tt_sim_communicator.hpp
```diff
@@ -26,9 +26,16 @@ class TTSimCommunicator final {
      * Constructor for TTSimCommunicator.
      *
      * @param simulator_directory Path to the simulator binary/directory
-     * @param copy_sim_binary If true, copy the simulator binary to memory for security
+     * @param copy_sim_binary If true, copy the simulator binary to memory for
+     *   security AND the loaded .so doesn't support the v3.5 chip-id ABI. If
+     *   the .so exports libttsim_create_device_by_id, v3.5 shared-library
+     *   mode is auto-enabled at initialize() time, ignoring this flag.
+     * @param chip_id Logical chip ID (0..N-1) within the cluster. Only used
+     *   in v3.5 multichip mode (per docs/v3_5_umd_patch_sketch.md). Default 0
```
pjanevskiTT, line 38: where is this docs/.... seems like claude generated for implemention. Can we change the comment?

device/api/umd/device/simulation/tt_sim_communicator.hpp
```diff
@@ -120,6 +135,16 @@ class TTSimCommunicator final {
 
     void start_sim();
 
+    // v3.5 commit #6 — eth-MAC wiring.
```
pjanevskiTT, line 150: comment doesn't help much in understanding the code imo

device/api/umd/device/simulation/tt_sim_communicator.hpp
```diff
@@ -141,6 +166,18 @@ class TTSimCommunicator final {
     // Flag to indicate if binary should be copied to memory.
     bool copy_sim_binary_;
 
+    // v3.5 multi-chip mode: when v3.5 symbols are present in the .so, all
```
pjanevskiTT, line 189: any docs somewhere on how this multichip support is implemented in the sim?

device/api/umd/device/simulation/tt_sim_communicator.hpp
```diff
@@ -149,11 +186,36 @@ class TTSimCommunicator final {
     void (*pfn_libttsim_pci_mem_wr_bytes_)(uint64_t paddr, const void *p, uint32_t size) = nullptr;
     void (*pfn_libttsim_tile_rd_bytes_)(uint32_t x, uint32_t y, uint64_t addr, void *p, uint32_t size) = nullptr;
     void (*pfn_libttsim_tile_wr_bytes_)(uint32_t x, uint32_t y, uint64_t addr, const void *p, uint32_t size) = nullptr;
+    void (*pfn_libttsim_dram_rd_bytes_by_id_)(
```
pjanevskiTT, line 208: also seems like something that can be split into a separate PR

device/api/umd/device/cluster.hpp
pjanevskiTT, line 87: again, I think this should be a separate PR. Does this affect silicon as well? If so, maybe we don't want to change this. Few reasons 
- UMD in general shouldn't allocate anything in the constructor, clients should use separate API or directly change this parameter
- UMD tests don't want to allocate anything, and we are calling with default options. Would be nice to update all of our tests if that is the case

device/api/umd/device/cluster.hpp
pjanevskiTT: I think this is fine, again I would consider making this a separate PR

device/chip_helpers/simulation_sysmem_manager.cpp
```diff
@@ -22,6 +31,14 @@ enum class ARCH;
 
 namespace tt::umd {
 
+namespace {
+
+constexpr uint64_t kHostMemChannelSize = 1ULL << 30;
```
pjanevskiTT, line 36: please reuse this constant if we have it hard coded anywhere in the repo already. If not, free to stay here

device/chip_helpers/simulation_sysmem_manager.cpp
```diff
@@ -76,14 +102,81 @@ void SimulationSysmemManager::unpin_or_unmap_sysmem() {
...
+    {
+        std::lock_guard<std::mutex> lock(mapped_buffers_mutex_);
+        auto mapped_buffer = find_mapped_buffer(base, size);
+        if (mapped_buffer.has_value()) {
+            std::memcpy(static_cast<uint8_t*>(mapped_buffer->buffer) + (base - mapped_buffer->base), src, size);
+            return;
+        }
+    }
+
+    SysmemManager::write_to_sysmem(channel, src, sysmem_dest, size);
```
pjanevskiTT, line 136: I am missing why do we need both the code above and this calling into the base class? They both do memcpy in the end, right?

third_party/CMakeLists.txt
```diff
@@ -17,15 +17,15 @@ include(${PROJECT_SOURCE_DIR}/cmake/ttexalens_private.cmake)
 ####################################################################################################################
 if(TT_UMD_BUILD_TESTS)
     CPMAddPackage(
-        NAME GTest
+        NAME googletest
```
pjanevskiTT: I think this should be dropped from the PR, https://github.com/tenstorrent/tt-umd/pull/2728 I guess this will solve the problems

pjanevskiTT (commented):
Finished part 1 of the review (tt_sim_communicator is next)
I think it would be nice to at least split this PR into 2

1. Multichip support
2. Sysmem buffer support
3. Misc things you added

device/simulation/tt_sim_communicator.cpp
```diff
@@ -44,6 +69,74 @@ TTSimCommunicator::~TTSimCommunicator() {
...
+    bool v3_5_supported = false;
+    {
+        // Temporarily dlopen just to check symbols; close after probe.
+        void *probe = dlopen(simulator_directory_.c_str(), RTLD_LAZY | RTLD_NOLOAD);
+        if (!probe) {
+            // RTLD_NOLOAD fails if not yet loaded — that's fine, do a fresh probe.
+            probe = dlopen(simulator_directory_.c_str(), RTLD_LAZY);
+        }
+        if (probe) {
+            v3_5_supported = dlsym(probe, "libttsim_create_device_by_id") != nullptr &&
```
pjanevskiTT, line 88: as a general comment, can we make all v3_5* variables and comments state multichip_*? Is there anything else in v3_5?

device/simulation/tt_sim_communicator.cpp
```diff
@@ -44,6 +69,74 @@ TTSimCommunicator::~TTSimCommunicator() {
...
+        if (!s_shared_handle_) {
+            s_shared_handle_ = dlopen(simulator_directory_.c_str(), RTLD_LAZY);
+            if (!s_shared_handle_) {
+                UMD_THROW(
+                    error::RuntimeError, fmt::format("Failed to dlopen simulator library (shared): {}", dlerror()));
+            }
+        }
+        s_shared_refcount_++;
+        libttsim_handle_ = s_shared_handle_;
+        DLSYM_FUNCTION(libttsim_init)
```
pjanevskiTT, line 109: we already have the code that calls DLSYM_FUNCTION on bunch of these functions. Please reuse it if you can

device/simulation/tt_sim_communicator.cpp
```diff
@@ -56,43 +149,105 @@ void TTSimCommunicator::initialize() {
...
+        // simulator is process-global; calling exit per chip would be wrong.
+        return;
+    }
     pfn_libttsim_exit_();
 }
 
+// v3.5: in multi-chip mode, every I/O entry point first selects the right
+// chip via libttsim_select_device_by_id(chip_id_) under the held device_lock_.
+// The shared libttsim's internal recursive_mutex provides defense-in-depth.
+#define SELECT_CHIP_IF_NEEDED()                                                \
```
pjanevskiTT, line 194: maybe make it just a simple helper function? Noticed any benefits of having this as #define ?

device/tt_device/tt_device.cpp
```diff
@@ -389,6 +389,8 @@ void TTDevice::wait_for_non_mmio_flush() {
     get_remote_interface()->get_remote_communication()->wait_for_non_mmio_flush();
 }
 
+void TTDevice::advance_device_execution() {}
```
pjanevskiTT, line 392: why reorder? Revert if there is no particular reason?

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -33,6 +35,30 @@ namespace tt::umd {
 
 static_assert(!std::is_abstract<TTSimTTDevice>(), "TTSimChip must be non-abstract.");
 
+namespace {
+
+bool sim_dram_teleport_enabled() {
+    // Cache the result since this is called on every device read/write.
+    static const bool enabled = [] {
+        const char* env = std::getenv("TT_METAL_SIMULATOR_DRAM_TELEPORT");
```
pjanevskiTT: can we have all these newly added env vars just have TT_ prefix? No metal, no umd. Especially if it's plumbed all the way down the stack, if it's not something really specific for the component?

pjanevskiTT (commented):
batch 2

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -33,6 +35,30 @@ namespace tt::umd {
...
+        if (env == nullptr) {
+            return false;
+        }
+        std::string_view value(env);
+        return value == "1" || value == "true" || value == "TRUE" || value == "on" || value == "ON";
+    }();
+    return enabled;
+}
+
+bool is_translated_dram_core(const SocDescriptor& soc_descriptor, tt_xy_pair core) {
```
pjanevskiTT, line 53: can we add this function to soc descriptor directly?

device/cluster.cpp
```diff
@@ -408,10 +415,137 @@ Cluster::Cluster(ClusterOptions options) {  // NOLINT(performance-unnecessary-va
...
+            return sim_tt ? sim_tt->get_communicator() : nullptr;
+        };
+        // Deterministic per-(chip, channel) MAC: top byte is OUI, next byte
+        // encodes chip_id, bottom byte encodes channel.
+        auto eth_sim_mac = [](ChipId cid, int chan) -> uint64_t {
+            return 0x021E52000000ULL | (uint64_t(cid & 0xFF) << 8) | uint64_t(chan & 0xFF);
+        };
+        // Reset the virtual switch once (singleton, process-global state).
+        if (!chips_.empty()) {
+            ChipId c0 = cluster_desc->get_chips_local_first(cluster_desc->get_all_chips()).front();
```
pjanevskiTT, line 468: can we rename all c0/c variables to chip* variables, makes it more readable

device/api/umd/device/cluster.hpp
```diff
@@ -142,7 +142,24 @@ class Cluster {
...
+
+    // Factories that build ClusterOptions inside libtt-umd to avoid cross-DSO STL ABI issues.
+    static std::unique_ptr<Cluster> create_silicon_cluster(
```
pjanevskiTT, line 150: I am not sure what is the benefit of all of these create functions. Seems like client still needs to have some logic which will decide which create function to call. Why not just populate ClusterOptions yourself at that point, that is what they are used for. I would probably remove these functions if there is not some case which I am missing

pjanevskiTT (commented):
Should be final batch of comments, happy to review split PRs with addressed comments

device/api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
```diff
@@ -6,6 +6,10 @@
```
pjanevskiTT, line 6: In general, for every split PR, I would prefer to have the tests, which don't have to be turned on right now, just as a way of showing how this stuff should be used. I guess for some of this stuff we can even turn on the tests immediately

pjanevskiTT:
closing this PR in favor of newly open 3 PRs that split this work, thanks @Riddy21

Issue #2679, opened 2026-05-19
Simulation support efforts


broskoTT:
This parent task should track many simulation efforts that we have, both for ttsim and for rtl sim, for all architectures.

@pjanevskiTT already did a huge amount of stuff on supporting various codepaths for Quasar, as well as supporting Fast Dispatch on both

## SimulationChip → LocalChip merge

Fold simulator backends into `LocalChip` so there is a single chip type for any `TTDevice`-backed device. The original plan (#2683–#2688) added a temporary `TTDevice::is_simulation()` capability flag and scattered gating through `LocalChip` before cleaning it up. The revised plan pushes backend variance directly into the `TTDevice` layer so `LocalChip`'s body becomes uniform across PCIe / JTAG / simulation — no capability flag is introduced. Prereq #2671 (TTSim+RTL merge) is in flight.

PR #2704, opened 2026-05-21
Move noc_multicast_write into TTDevice

### LINKED ISSUE #2699: [Refactor] Move noc_multicast_write into TTDevice virtual
Part of #2679. Supersedes part of #2686.

## Goal

`LocalChip::noc_multicast_write` is PCIe-shaped (uses the cached WC TLB window's multicast-reconfigure path); `SimulationChip` falls back to a per-core unicast loop. Push the variance into a `TTDevice` virtual so `LocalChip` delegates unconditionally.

## Changes

- `TTDevice` declares:
  ```cpp
  virtual void noc_multicast_write(
      void* src, size_t size, CoreCoord core_start, CoreCoord core_end, uint64_t addr);
  ```
- PCIe `TTDevice` override: the existing `LocalChip::noc_multicast_write` body moves in (uses the cached WC TLB window now owned by PCIe TTDevice from the TLB-infrastructure issue).
- Sim `TTDevice` subclasses: per-core unicast workaround with the BLACKHOLE x=8/9 skip.
- JTAG `TTDevice`: throws (matches today).
- `LocalChip::noc_multicast_write` becomes `tt_device_->noc_multicast_write(src, size, core_start, core_end, addr)` plus the Tensix-only type check.

## Verification

- `cmake --build build`.
- Full PCIe / JTAG / simulation test suites.
- `pre-commit run --all-files`.

pjanevskiTT:
### Issue
Closes #2699. Part of #2679.

### Description
Push `noc_multicast_write` variance from the chip layer into `TTDevice`, so `LocalChip` and `SimulationChip` both reduce to thin delegations.

This is a PR that is pushing everything into TTDevice layer so Chip class stays unchanged for every backend.

### List of the changes

- Override noc_multicast_write in sim classes
- Make SimulationChip call into TTDevice layer

### Testing

CI

### API Changes
/

### DIFF
--- a/device/chip/local_chip.cpp
+++ b/device/chip/local_chip.cpp
@@ -550,21 +550,6 @@ void LocalChip::noc_multicast_write(
     if (core_start.core_type != CoreType::TENSIX || core_end.core_type != CoreType::TENSIX) {
         UMD_THROW(error::RuntimeError, "noc_multicast_write is only supported for Tensix cores.");
     }
-
-    // Multicast write relies on PCIe-specific TLB operations; ensure the communication device is PCIe.
-    if (tt_device_->get_communication_device_type() != IODeviceType::PCIe) {
-        UMD_THROW(error::RuntimeError, "noc_multicast_write is only supported on PCIe devices.");
-    }
-
-    std::lock_guard<std::mutex> lock(wc_tlb_lock);
-
-    get_cached_wc_tlb_window()->noc_multicast_write_reconfigure(
-        src,
-        size,
-        get_soc_descriptor().translate_chip_coord_to_translated(core_start),
-        get_soc_descriptor().translate_chip_coord_to_translated(core_end),
-        addr,
-        get_selected_noc_id(),
-        tlb_data::Relaxed);
+    tt_device_->noc_multicast_write(src, size, core_start, core_end, addr);
 }
 }  // namespace tt::umd
--- a/device/tt_device/rtl_simulation_tt_device.cpp
+++ b/device/tt_device/rtl_simulation_tt_device.cpp
@@ -336,6 +336,21 @@ void RtlSimulationTTDevice::retrain_dram_core(const uint32_t dram_channel) {
     UMD_THROW(error::RuntimeError, "DRAM retraining is not supported in RTL simulation device.");
 }
 
+void RtlSimulationTTDevice::noc_multicast_write(
+    const void* src, size_t size, tt_xy_pair core_start, tt_xy_pair core_end, uint64_t addr) {
+    // RTL sim has no hardware multicast; fall back to per-core unicast.
+    // TODO: investigate proper multicast support for simulations so we can remove this workaround.
+    for (uint32_t x = core_start.x; x <= core_end.x; ++x) {
+        for (uint32_t y = core_start.y; y <= core_end.y; ++y) {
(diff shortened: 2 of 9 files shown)


device/tt_device/rtl_simulation_tt_device.cpp
```diff
@@ -334,6 +334,21 @@ void RtlSimulationTTDevice::retrain_dram_core(const uint32_t dram_channel) {
...
+    const void* src, size_t size, tt_xy_pair core_start, tt_xy_pair core_end, uint64_t addr) {
+    // RTL sim has no hardware multicast; fall back to per-core unicast.
+    // TODO: investigate proper multicast support for simulations so we can remove this workaround.
+    for (uint32_t x = core_start.x; x <= core_end.x; ++x) {
+        for (uint32_t y = core_start.y; y <= core_end.y; ++y) {
+            // BLACKHOLE: x==8 is ARC/L2CPU and x==9 is GDDR, not actual Tensix cores.
+            if (arch == tt::ARCH::BLACKHOLE && (x == 8 || x == 9)) {
+                continue;
+            }
+            write_to_device(src, tt_xy_pair{x, y}, addr, size);
```
broskoTT: Do you feel we could have a common tt device implementation for both simulations? Just a thought

broskoTT (approved):

device/chip/local_chip.cpp
```diff
@@ -552,21 +552,6 @@ void LocalChip::noc_multicast_write(
...
-
-    get_cached_wc_tlb_window()->noc_multicast_write_reconfigure(
-        src,
-        size,
-        get_soc_descriptor().translate_chip_coord_to_translated(core_start),
-        get_soc_descriptor().translate_chip_coord_to_translated(core_end),
-        addr,
-        get_selected_noc_id(),
-        tlb_data::Relaxed);
+    tt_device_->noc_multicast_write(src, size, core_start, core_end, addr);
```
nbuncicTT, line 553: Is this okay to do? Because here we are having the TLB mapped as WC, and the one in TTDevice is UC.

Do you think that once we have both WC and UC TLBs cached in TTDevice (coming soon) we should use WC for multicast?

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -390,6 +390,21 @@ void TTSimTTDevice::retrain_dram_core(const uint32_t dram_channel) {
     UMD_THROW(error::RuntimeError, "DRAM retraining is not supported in TTSim device.");
 }
 
+void TTSimTTDevice::noc_multicast_write(
```
nbuncicTT, line 393: This is completely the same as in `RtlSimulationTTDevice` - my suggestion would be to make a free helper function which is used by both classes for these kinds of situations (a sort of utility which can be re-used if there's a third similar impl. etc.).

You can solve this by setting a lambda/std::function or template for the `write_to_device` which each simulation class would set.

nbuncicTT (approved):

device/tt_device/rtl_simulation_tt_device.cpp
pjanevskiTT: https://github.com/tenstorrent/tt-umd/issues/2865

device/chip/local_chip.cpp
pjanevskiTT: yes we can change multicast to use WC after it's added

PR #2710, opened 2026-05-22
pcie: add portable SIMD memcpy paths for non-x86 (GCC/Clang)

lu-zero:
Add wide load/store paths to memcpy_to_device and memcpy_from_device using GCC vector extensions (__attribute__((vector_size))) and __builtin_memcpy.  On AArch64 this lowers to LDP/STP (NEON load/store pairs), issuing each device address exactly once.

For writes (to_device), Clang's __builtin_nontemporal_store produces STNP (non-temporal store pair) which bypasses the CPU cache so data goes straight to the PCIe write buffer.  GCC falls back to regular STP.  A 2 KB software prefetch hides source DRAM latency.

For reads (from_device), both compilers emit LDP/STP + PRFM prefetch.

Guard restructured from __has_builtin(__builtin_nontemporal_store) (Clang-only) to defined(__GNUC__) so both GCC and Clang get the wide path, with the non-temporal store conditional kept inside the block.

82 off-device correctness tests covering all internal phases (alignment, 32/16/4/1-byte tails, 256-byte bulk blocks) and several dst/src misalignment combinations.

### DIFF
--- a/device/pcie/device_memcpy.cpp
+++ b/device/pcie/device_memcpy.cpp
@@ -91,6 +91,64 @@ void memcpy_to_device(volatile void* dest, const void* src, std::size_t size) {
     }
 
     d = reinterpret_cast<volatile std::uint8_t*>(d_simd);
+
+#elif defined(__GNUC__)
+    // GCC/Clang non-x86: vector extensions + __builtin_memcpy (LDP/STP on AArch64).
+    // Clang gets non-temporal stores (STNP); GCC falls back to regular stores.
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast).
+    auto* d_wide = const_cast<std::uint8_t*>(static_cast<const volatile std::uint8_t*>(d));
+
+    typedef std::uint8_t __attribute__((vector_size(32), aligned(1))) v32;
+    typedef std::uint8_t __attribute__((vector_size(16), aligned(1))) v16;
+    constexpr std::size_t prefetch_distance = 2048;
+
+    // Phase 1: 256-byte blocks (8 x 32-byte stores).
+    while (size >= 256) {
+        for (int j = 0; j < 8; ++j) {
+            __builtin_prefetch(s + prefetch_distance, 0, 0);
+            v32 chunk;
+            __builtin_memcpy(&chunk, s, 32);
+#if __has_builtin(__builtin_nontemporal_store)
+            __builtin_nontemporal_store(chunk, reinterpret_cast<v32*>(d_wide));
+#else
+            __builtin_memcpy(d_wide, &chunk, 32);
+#endif
+            s += 32;
+            d_wide += 32;
+        }
+        size -= 256;
+    }
+
+    // Phase 2: Remaining 32-byte chunks.
+    while (size >= 32) {
+        __builtin_prefetch(s + prefetch_distance, 0, 0);
+        v32 chunk;
+        __builtin_memcpy(&chunk, s, 32);
+#if __has_builtin(__builtin_nontemporal_store)
+        __builtin_nontemporal_store(chunk, reinterpret_cast<v32*>(d_wide));
+#else
+        __builtin_memcpy(d_wide, &chunk, 32);
+#endif
+        s += 32;
+        d_wide += 32;
+        size -= 32;
+    }
+
+    // Phase 3: Remaining 16-byte chunk.
+    if (size >= 16) {
+        v16 chunk;
+        __builtin_memcpy(&chunk, s, 16);
+#if __has_builtin(__builtin_nontemporal_store)
+        __builtin_nontemporal_store(chunk, reinterpret_cast<v16*>(d_wide));
+#else
+        __builtin_memcpy(d_wide, &chunk, 16);
+#endif
+        s += 16;
+        d_wide += 16;
+        size -= 16;
+    }
+
+    d = reinterpret_cast<volatile std::uint8_t*>(d_wide);
 #endif
 
     // Phase 4: Remaining 4-byte chunks.
@@ -172,6 +230,51 @@ void memcpy_from_device(void* dest, const volatile void* src, std::size_t size) 
     }
 
     s = reinterpret_cast<const volatile std::uint8_t*>(s_simd);
+
+#elif defined(__GNUC__)
+    // GCC/Clang non-x86: vector extensions + __builtin_memcpy (LDP/STP on AArch64).
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast).
+    auto* s_wide = const_cast<std::uint8_t*>(static_cast<const volatile std::uint8_t*>(s));
+
+    typedef std::uint8_t __attribute__((vector_size(32), aligned(1))) v32;
+    typedef std::uint8_t __attribute__((vector_size(16), aligned(1))) v16;
+    constexpr std::size_t prefetch_distance = 2048;
+
+    // Phase 1: 256-byte blocks (8 x 32-byte loads).
+    while (size >= 256) {
+        for (int j = 0; j < 8; ++j) {
+            __builtin_prefetch(s_wide + prefetch_distance, 0, 0);
+            v32 chunk;
+            __builtin_memcpy(&chunk, s_wide, 32);
+            __builtin_memcpy(d, &chunk, 32);
+            s_wide += 32;
+            d += 32;
+        }
+        size -= 256;
+    }
+
+    // Phase 2: Remaining 32-byte chunks.
+    while (size >= 32) {
+        __builtin_prefetch(s_wide + prefetch_distance, 0, 0);
+        v32 chunk;
+        __builtin_memcpy(&chunk, s_wide, 32);
+        __builtin_memcpy(d, &chunk, 32);
+        s_wide += 32;
+        d += 32;
+        size -= 32;
+    }
+
+    // Phase 3: Remaining 16-byte chunk.
+    if (size >= 16) {
+        v16 chunk;
+        __builtin_memcpy(&chunk, s_wide, 16);
+        __builtin_memcpy(d, &chunk, 16);
+        s_wide += 16;
+        d += 16;
+        size -= 16;
+    }
+
+    s = reinterpret_cast<const volatile std::uint8_t*>(s_wide);
 #endif
 
     // Phase 4: Remaining 4-byte chunks.
--- a/device/pcie/device_memcpy.cpp
+++ b/device/pcie/device_memcpy.cpp
@@ -93,22 +93,44 @@ void memcpy_to_device(volatile void* dest, const void* src, std::size_t size) {
     d = reinterpret_cast<volatile std::uint8_t*>(d_simd);
 
 #elif defined(__GNUC__)
-    // GCC/Clang non-x86: vector extensions + __builtin_memcpy (LDP/STP on AArch64).
-    // Clang gets non-temporal stores (STNP); GCC falls back to regular stores.
+    // Like the x86 path, these wide stores are not volatile — compiler reordering is
+    // bounded by the Phase 0/4/5 volatile loops. On Clang, __builtin_nontemporal_store
+    // lowers to STNP (weakly ordered on AArch64), so a release fence is inserted after
+    // Phases 1-3 to ensure all NT stores retire before Phase 4/5 volatile writes or any
+    // subsequent caller doorbell write. GCC uses regular STP stores (strongly ordered).
+
+    // Further align destination to 16 bytes: STP Q / STNP Q to Device/UC memory (PCIe BAR)
+    // require 16-byte-aligned addresses; Phase 0 only guaranteed 4-byte alignment.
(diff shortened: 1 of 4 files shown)


device/pcie/device_memcpy.cpp
```diff
@@ -91,6 +91,64 @@ void memcpy_to_device(volatile void* dest, const void* src, std::size_t size) {
...
+    constexpr std::size_t prefetch_distance = 2048;
+
+    // Phase 1: 256-byte blocks (8 x 32-byte stores).
+    while (size >= 256) {
+        for (int j = 0; j < 8; ++j) {
+            __builtin_prefetch(s + prefetch_distance, 0, 0);
+            v32 chunk;
+            __builtin_memcpy(&chunk, s, 32);
+#if __has_builtin(__builtin_nontemporal_store)
+            __builtin_nontemporal_store(chunk, reinterpret_cast<v32*>(d_wide));
```
pjanevskiTT, line 134: Non-temporal store + no memory barrier = ordering hazard on AArch64.

STNP (what Clang's `__builtin_nontemporal_store` lowers to on AArch64 for a 32-byte vector) is weakly ordered relative to subsequent stores. The existing x86 path is safe because x86 TSO orders regular VMOVDQU stores with later volatile writes — the comment at line 43–45 banks on this implicitly ("Phases 0/4/5 bound the reordering window"). That argument *does not* carry to ARM: the Phase 4/5 volatile byte/dword loops bound compiler reordering but are not hardware memory barriers against non-temporal stores.

If a caller follows `memcpy_to_device(buf, data, N)` with a volatile write to a doorbell/control register, the doorbell can land on the device before the data writes. Same hazard inside Phase 2 (line 128).

Two safe options:
- Drop `__builtin_nontemporal_store` and use plain `__builtin_memcpy` on Clang too — loses the cache-bypass perf delta but matches the x86 contract.
- Emit a `__atomic_thread_fence(__ATOMIC_RELEASE)` (DMB ST on AArch64) at the end of the wide-store path, before Phase 4.

Either way, document the contract: "caller must fence before any subsequent write that depends on this data."

device/pcie/device_memcpy.cpp
```diff
@@ -91,6 +91,64 @@ void memcpy_to_device(volatile void* dest, const void* src, std::size_t size) {
...
 
     d = reinterpret_cast<volatile std::uint8_t*>(d_simd);
+
+#elif defined(__GNUC__)
+    // GCC/Clang non-x86: vector extensions + __builtin_memcpy (LDP/STP on AArch64).
+    // Clang gets non-temporal stores (STNP); GCC falls back to regular stores.
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast).
+    auto* d_wide = const_cast<std::uint8_t*>(static_cast<const volatile std::uint8_t*>(d));
+
+    typedef std::uint8_t __attribute__((vector_size(32), aligned(1))) v32;
```
pjanevskiTT, line 118: `aligned(1)` on the vector type may prevent STP/STNP emission.

The PR description claims "lowers to LDP/STP (NEON load/store pairs)" on AArch64, but `aligned(1)` tells the compiler the type has only byte alignment — and STP/STNP for Q registers requires 16-byte alignment (unaligned STP traps on some implementations). Phase 0 only aligns to 4 bytes, so the compiler cannot prove 16-byte alignment at the wide-path entry. With `aligned(1)`, GCC/Clang typically fall back to a pair of `STR Q` (or even `STR D` × 2) rather than `STP Q, Q`.

Worth verifying with `objdump -d` that STP/STNP is actually what gets emitted in your build. If the compiler is producing two separate STR instructions, the "each device address exactly once" claim and the perf delta both need re-checking. One way to force the pair: align the destination to 16 bytes in Phase 0 (extend the alignment loop above) and drop `aligned(1)`.

device/pcie/device_memcpy.cpp
```diff
@@ -91,6 +91,64 @@ void memcpy_to_device(volatile void* dest, const void* src, std::size_t size) {
     }
 
     d = reinterpret_cast<volatile std::uint8_t*>(d_simd);
+
+#elif defined(__GNUC__)
+    // GCC/Clang non-x86: vector extensions + __builtin_memcpy (LDP/STP on AArch64).
```
pjanevskiTT: The x86 path opens with an explicit reordering caveat (line 43–45) — "AVX2 intrinsics not themselves volatile, compiler is free to reorder ..., Phases 0/4/5 bound the window." The new GCC path is in the same boat (these are also non-volatile stores) but the comment doesn't reproduce the caveat. Either duplicate it or factor a single explainer above both blocks. As-is, a future reader of just the GCC branch won't realize the same constraint applies. See also the comment on line 112 re: hardware ordering.

device/pcie/device_memcpy.cpp
```diff
@@ -91,6 +91,64 @@ void memcpy_to_device(volatile void* dest, const void* src, std::size_t size) {
...
+    auto* d_wide = const_cast<std::uint8_t*>(static_cast<const volatile std::uint8_t*>(d));
+
+    typedef std::uint8_t __attribute__((vector_size(32), aligned(1))) v32;
+    typedef std::uint8_t __attribute__((vector_size(16), aligned(1))) v16;
+    constexpr std::size_t prefetch_distance = 2048;
+
+    // Phase 1: 256-byte blocks (8 x 32-byte stores).
+    while (size >= 256) {
+        for (int j = 0; j < 8; ++j) {
+            __builtin_prefetch(s + prefetch_distance, 0, 0);
```
pjanevskiTT, line 130: The prefetch is issued 8× per outer iteration (once per inner-loop iteration). Each prefetch targets a *different* address since `s` advances by 32 inside the loop, so they're not strictly redundant — but the natural pattern would be to prefetch once at the top of the outer block targeting `s + prefetch_distance` (or the next outer block's start). Compilers typically don't dedupe prefetches even when they're hints. Minor; consider hoisting if a perf re-run shows it matters.

device/pcie/device_memcpy.cpp
```diff
@@ -91,6 +91,64 @@ void memcpy_to_device(volatile void* dest, const void* src, std::size_t size) {
...
+
+#elif defined(__GNUC__)
+    // GCC/Clang non-x86: vector extensions + __builtin_memcpy (LDP/STP on AArch64).
+    // Clang gets non-temporal stores (STNP); GCC falls back to regular stores.
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast).
+    auto* d_wide = const_cast<std::uint8_t*>(static_cast<const volatile std::uint8_t*>(d));
+
+    typedef std::uint8_t __attribute__((vector_size(32), aligned(1))) v32;
+    typedef std::uint8_t __attribute__((vector_size(16), aligned(1))) v16;
+    constexpr std::size_t prefetch_distance = 2048;
```
pjanevskiTT, line 121: `prefetch_distance = 2048` is a magic number. The PR description explains the choice ("hides source DRAM latency") but the code itself doesn't. A brief comment — "tuned for typical AArch64 LLC-miss latency at PCIe DMA write rates; ~32 cache lines ahead" — would let the next person tune this confidently. Minor.

device/pcie/device_memcpy.cpp
lu-zero: I went with the latter.

device/pcie/device_memcpy.cpp
lu-zero: at least with clang 22 it seems doing the right thing.

lu-zero:
## AArch64 test results (Blackhole, SMMU enabled)

All TLB benchmarks pass with the portable memcpy paths on AArch64:

| Path | Transfer size | Throughput |
|---|---|---|
| Static TLB write | 4 KiB | 5.53 GB/s |
| Static TLB write | 8 KiB | 5.53 GB/s |
| Static TLB write | 1 MiB | 5.53 GB/s |
| Dynamic TLB write | 1 MiB | 5.52 GB/s |
| Static TLB read | 1 KiB–1 MiB | ~22 MB/s |
| Dynamic TLB read | 1 KiB–1 MiB | ~21 MB/s |

The read asymmetry (~22 MB/s vs ~5.5 GB/s for writes) is expected: MMIO reads stall the CPU waiting for a PCIe completion packet, while writes post to the CPU write-combining buffer and return immediately.

Note: without the `configure_tlb` fix from PR #2648, none of these benchmarks run on AArch64 (SIGBUS on the first TLB configuration).

lu-zero (commented):
All review comments are addressed in the current revision:

- **STNP ordering hazard**:  inserted after Phases 1–3 on Clang (the only path that emits STNP); GCC uses regular STP which is TSO-ordered.
- **`aligned(1)` and STP/STNP emission**: comment added confirming Clang 22 emits `stnp q` regardless of the alignment attribute (verified via objdump on AArch64).
- **Missing reordering caveat**: GCC/Clang block now opens with the same caveat as the x86 path.
- **Prefetch pattern / magic number**: per-chunk prefetch rationale documented inline; `prefetch_distance = 2048` comment explains the choice.
- **`__has_builtin` portability**: avoided entirely — `__builtin_nontemporal_store` is guarded with `#if defined(__clang__)` which is unambiguous on all supported compilers.

lu-zero (commented):
All review comments are addressed in the current revision:

- **STNP ordering hazard**: a release fence is inserted after Phases 1-3 on Clang (the only path that emits STNP); GCC uses regular STP which is strongly ordered.
- **`aligned(1)` and STP/STNP emission**: comment added confirming Clang 22 emits `stnp q` regardless of the alignment attribute (verified via objdump on AArch64).
- **Missing reordering caveat**: the GCC/Clang block now opens with the same caveat as the x86 path.
- **Prefetch pattern / magic number**: per-chunk prefetch rationale documented inline; `prefetch_distance = 2048` comment explains the tuning choice.
- **`__has_builtin` portability**: avoided entirely — `__builtin_nontemporal_store` is guarded with `#if defined(__clang__)` which is unambiguous on all supported compilers.

device/pcie/device_memcpy.cpp
```diff
@@ -91,6 +91,91 @@ void memcpy_to_device(volatile void* dest, const void* src, std::size_t size) {
...
+
+    // Phase 1: 256-byte blocks (8 x 32-byte stores).
+    // The prefetch is issued once per 32-byte chunk (not hoisted to the outer loop) so
+    // that after unrolling the compiler issues 8 prfm instructions covering 8 distinct
+    // cache lines (+2048..+2272 bytes ahead). Hoisting to one prefetch per 256-byte block
+    // would fetch only a single cache line ahead, leaving the other 7 uncovered.
+    while (size >= 256) {
+        for (int j = 0; j < 8; ++j) {
+            __builtin_prefetch(s + prefetch_distance, 0, 0);
```
pjanevskiTT, line 130: this seems like something we can introduce even in our x86 code, it's on us to investigate, no action for you

device/pcie/device_memcpy.cpp
```diff
@@ -91,6 +91,91 @@ void memcpy_to_device(volatile void* dest, const void* src, std::size_t size) {
...
+        size -= 256;
+    }
+
+    // Phase 2: Remaining 32-byte chunks.
+    while (size >= 32) {
+        __builtin_prefetch(s + prefetch_distance, 0, 0);
+        v32 chunk;
+        __builtin_memcpy(&chunk, s, 32);
+#if defined(__clang__)
+        __builtin_nontemporal_store(chunk, reinterpret_cast<v32*>(d_wide));
```
pjanevskiTT, line 150: we initially had a PR that had nontemporal stores on x86 as well, but it had some problems in multiprocess tests specifically. So since we don't have arm in the ci, could you please run all the tests from the CI in your local env so we confirm this works?

tests/api/test_device_memcpy.cpp
```diff
@@ -0,0 +1,164 @@
...
+namespace {
+
+// Pad on each side to catch over-reads / over-writes.
+constexpr size_t kGuardBytes = 64;
+constexpr uint8_t kGuardFill = 0xAB;
+constexpr uint8_t kSrcFill = 0x42;
+
+// Verify that memcpy_to_device writes exactly the right bytes and leaves
+// everything outside [dst_offset, dst_offset+size) untouched.
+void check_to_device(size_t size, size_t dst_offset, size_t src_offset) {
```
pjanevskiTT: check_to_device and check_from_device seem to share a lot of code. Maybe try to merge that code in functions or somehow parametrize a single check_* function with memcpy_to/from device. It's just a code health thing to address, won't block pr on this

pjanevskiTT (approved):
Looks good, would like confirmation that all tests from ci passed in your env, because of non temporal stores. Let us know when comments are addressed so we can merge

tests/api/test_device_memcpy.cpp
lu-zero: Done :)

device/pcie/device_memcpy.cpp
lu-zero: Everything seems working fine, only the warm reset ones have pre-existing timing problems.

device/pcie/device_memcpy.cpp
pjanevskiTT: #2770

PR #2711, opened 2026-05-22
sysmem: add 512MB hugepage fallback for aarch64

lu-zero:
Add a 512MB hugepage attempt between the existing 1GB and 2MB paths in mmap_with_hugepage_fallback.  On AArch64 with 64K base pages, 512MB PMD blocks map as a single SMMU TLB entry — the most efficient option for this platform (2 entries cover the 1GB IOMMU sysmem window).

Promote the silent regular-page fallback to log_warning with a hint to pre-allocate hugepages.  Fix the hardcoded "without hugepages" log message in pin_or_map_iommu to say "via IOMMU".

Measured on Blackhole AArch64 (PCIe 5.0 x16, SMMU3, 64K pages):
  64KB pages (baseline):   4704KB H2D = 993µs / 4.85 GB/s
  2MB hugepages:           4704KB H2D = 963µs / 5.00 GB/s  (+3%)
  512MB hugepages:         4704KB H2D = 866µs / 5.56 GB/s  (+15%)

### DIFF
--- a/device/chip_helpers/silicon_sysmem_manager.cpp
+++ b/device/chip_helpers/silicon_sysmem_manager.cpp
@@ -7,7 +7,7 @@
 #include "umd/device/chip_helpers/silicon_sysmem_manager.hpp"
 
 #include <fmt/format.h>
-#include <linux/mman.h>  // for MAP_HUGE_1GB, MAP_HUGE_2MB
+#include <linux/mman.h>  // for MAP_HUGE_1GB, MAP_HUGE_512MB, MAP_HUGE_2MB
 #include <sys/mman.h>    // for mmap, munmap
 #include <sys/stat.h>    // for fstat
 #include <unistd.h>
@@ -37,11 +37,14 @@
 
 namespace tt::umd {
 
-// Try to mmap with 1GB hugepages first, then 2MB hugepages, then regular pages.
-// This is a performance optimization: hugepages reduce page fault overhead during allocation.
-// All three options are functionally correct when IOMMU is enabled.
+// Try hugepage sizes from largest to smallest, then fall back to regular pages.
+// Larger hugepages are strictly preferred: each hugepage maps to one SMMU TLB entry
+// regardless of size, so 1GB (PUD/level-1 block) consumes half as many TLB entries as
+// 512MB (PMD/level-2 block) for the same total memory, reducing IOMMU mapping overhead.
+// 512MB is only meaningful on AArch64 with 64K base pages; it is not exposed on x86.
 static void *mmap_with_hugepage_fallback(size_t size) {
     constexpr size_t kHugepage1GiB = 1ULL << 30;
+    constexpr size_t kHugepage512MiB = 512ULL << 20;
     constexpr size_t kHugepage2MiB = 2ULL << 20;
 
     void *addr = MAP_FAILED;
@@ -57,6 +60,21 @@ static void *mmap_with_hugepage_fallback(size_t size) {
             0);
         if (addr != MAP_FAILED) {
             log_debug(LogUMD, "Allocated {:#x} bytes using 1GB hugepages.", size);
+            return addr;
+        }
+    }
+
+    // 512MB hugepages (PMD block on AArch64 with 64K base pages).
+    if (size >= kHugepage512MiB && (size % kHugepage512MiB) == 0) {
+        addr = mmap(
+            nullptr,
+            size,
+            PROT_READ | PROT_WRITE,
+            MAP_PRIVATE | MAP_ANONYMOUS | MAP_HUGETLB | MAP_HUGE_512MB | MAP_POPULATE,
+            -1,
+            0);
+        if (addr != MAP_FAILED) {
+            log_debug(LogUMD, "Allocated {:#x} bytes using 512MB hugepages.", size);
             return addr;
         }
     }
@@ -78,7 +96,14 @@ static void *mmap_with_hugepage_fallback(size_t size) {
 
     addr = mmap(nullptr, size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS | MAP_POPULATE, -1, 0);
     if (addr != MAP_FAILED) {
-        log_debug(LogUMD, "Allocated {:#x} bytes using regular pages.", size);
+        log_warning(
+            LogUMD,
+            "Sysmem ({:#x} bytes) using regular pages; pre-allocate hugepages for better DMA performance "
+            "(e.g. {} × 1GB or {} × 2MB; on AArch64 also {} × 512MB).",
+            size,
+            (size + kHugepage1GiB - 1) / kHugepage1GiB,
+            (size + kHugepage2MiB - 1) / kHugepage2MiB,
+            (size + kHugepage512MiB - 1) / kHugepage512MiB);
     }
     return addr;
 }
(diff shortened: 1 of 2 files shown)


pjanevskiTT (approved):

device/chip_helpers/silicon_sysmem_manager.cpp
```diff
@@ -37,12 +37,11 @@
...
 
-// Try to mmap with 1GB hugepages first, then 2MB hugepages, then regular pages.
-// This is a performance optimization: hugepages reduce page fault overhead during allocation.
-// All three options are functionally correct when IOMMU is enabled.
+// Try hugepage sizes from largest to smallest, then fall back to regular pages.
 static void *mmap_with_hugepage_fallback(size_t size) {
-    constexpr size_t kHugepage1GiB = 1ULL << 30;
-    constexpr size_t kHugepage2MiB = 2ULL << 20;
+    constexpr size_t kHugepage1GiB   = 1ULL << 30;
+    constexpr size_t kHugepage512MiB = 512ULL << 20;
```
pjanevskiTT, line 47: The aligned `=` signs across the three `constexpr` lines (42, 43, 44) will be collapsed by pre-commit's clang-format hook — the project's `.clang-format` inherits Google style without `AlignConsecutiveAssignments`, so single-space is required. Confirmed by running clang-format locally: both lines 42 and 44 get reformatted. Pre-commit will fail or auto-modify on next commit. Fix: drop the extra spaces, one space before `=`.

tests/microbenchmark/benchmarks/iommu/test_iommu.cpp
```diff
@@ -137,6 +137,57 @@ TEST(MicrobenchmarkIOMMU, MapHugepages2M) {
     test::utils::export_results(bench.title(), std::vector<ankerl::nanobench::Result>{map_result, unmap_result});
 }
 
+TEST(MicrobenchmarkIOMMU, MapHugepages512M) {
```
pjanevskiTT, line 150: The adjacent 2M test (line 86–88) and 1G test (line 191–193) both carry the same explanatory comment:

```
// Measure the time it takes to map hugepages using IOMMU.
// These should be different from regular buffers because it's guaranteed that hugepages are contiguous in memory.
// Since contiguous memory has less entries in the IOMMU page table, we expect the mapping to be faster.
```

The new 512M test is the only one without it, breaking the pattern. Copy the comment over for consistency.

device/chip_helpers/silicon_sysmem_manager.cpp
```diff
@@ -78,7 +92,14 @@ static void *mmap_with_hugepage_fallback(size_t size) {
 
     addr = mmap(nullptr, size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS | MAP_POPULATE, -1, 0);
     if (addr != MAP_FAILED) {
-        log_debug(LogUMD, "Allocated {:#x} bytes using regular pages.", size);
+        log_warning(
+            LogUMD,
+            "Sysmem ({:#x} bytes) allocated with regular pages; hugepage allocation failed. "
+            "Pre-allocate hugepages for better DMA performance "
+            "(e.g. {} × 512MB or {} × 2MB).",
```
pjanevskiTT: "Pre-allocate hugepages ... (e.g. N × 512MB or N × 2MB)" suggests a 512MB hugepage size to every platform that hits this fallback, but on x86_64 the kernel only supports 2MB (PMD) and 1GB (PUD) hugepages — there's no 512MB size at all. An x86 admin reading this warning will be confused trying to pre-allocate a non-existent page size. Two options:

- Include 1GB in the hint too: `(e.g. N × 1GB, N × 512MB on AArch64, or N × 2MB)`
- Or word it neutrally: "depending on platform support"

Not a correctness bug, just a usability nit on a message you only see when something has already gone wrong.

device/chip_helpers/silicon_sysmem_manager.cpp
```diff
@@ -61,6 +60,21 @@ static void *mmap_with_hugepage_fallback(size_t size) {
         }
     }
 
+    // 512MB hugepages (PMD block on AArch64 with 64K base pages).
```
pjanevskiTT, line 67: Order is 1GB → 512MB → 2MB. Correct on AArch64 with 64K base pages (1GB PUD blocks are even better than 512MB PMD blocks where available) and the PR description's "+15%" perf number was measured against the 512MB path, so the ordering is intentional. Worth confirming there's no expectation that 512MB is preferred over 1GB on AArch64 — if there *is* a reason to prefer 512MB (e.g. SMMU page-table walker behavior), swapping the order would matter. If 1GB-first is correct, ignore.

pjanevskiTT:
@lu-zero let us know when comments are addressed so we can merge this

lu-zero:
Now it should be good to go.

pjanevskiTT:
@lu-zero can you just resolve all the threads? You can either resolve it or comment why didn't you solve it (or solved it some other way)

tests/microbenchmark/benchmarks/iommu/test_iommu.cpp
```diff
@@ -137,6 +137,57 @@ TEST(MicrobenchmarkIOMMU, MapHugepages2M) {
...
+    std::vector<int> pci_device_ids = PCIDevice::enumerate_devices();
+    if (!PCIDevice(pci_device_ids.at(0)).is_iommu_enabled()) {
+        GTEST_SKIP() << "Skipping test since IOMMU is not enabled on the system.";
```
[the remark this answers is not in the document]
lu-zero: Solved for all the places that have the problem.

tests/microbenchmark/benchmarks/iommu/test_iommu.cpp
lu-zero: Done.

device/chip_helpers/silicon_sysmem_manager.cpp
lu-zero: I hope the new wording is ok.

device/chip_helpers/silicon_sysmem_manager.cpp
lu-zero: on aarch64 + 64k pages:

  hugepages-16777216kB  ← 16GiB
  hugepages-524288kB     ← 512MiB
  hugepages-2048kB          ← 2MiB
  
  And 16GiB is maybe a bit overkill.
  
  so 1GiB for less exotic system, 512MiB as failover (preferred on 64k) and so on.

lu-zero:
Done. Future hardware could leverage a 16GiB size?

lu-zero:
## AArch64 test results (Blackhole, SMMU enabled, 64K base pages)

### IOMMU map/unmap timing — regular pages vs 512 MiB hugepages

| Buffer type | Size | Map time | Unmap time |
|---|---|---|---|
| Regular pages (64K base) | 64 KiB | 0.003 ms | 0.001 ms |
| Regular pages (64K base) | 1 MiB | 0.008 ms | 0.002 ms |
| Regular pages (64K base) | 64 MiB | 0.119 ms | 0.079 ms |
| Regular pages (64K base) | 512 MiB | 0.860 ms | 0.609 ms |
| Regular pages (64K base) | 1 GiB | 1.472 ms | 1.222 ms |
| **512 MiB hugepage** | **512 MiB** | **11.984 ms** | **0.010 ms** |

Key observations:
- **Unmap is ~60× faster** with 512 MiB hugepages (0.010 ms vs 0.609 ms for the same 512 MiB buffer mapped with 64K regular pages). This is the critical path — `free_sysmem` is called on every buffer release.
- Map cost is higher for hugepages (11.984 ms vs 0.860 ms) because the kernel must find and pin a contiguous 512 MiB physical range and build a single SMMU block descriptor. This is paid once at sysmem initialization, not per transfer.
- On AArch64 with 64K base pages, a 512 MiB regular buffer requires 8192 IOMMU page table entries. A 512 MiB hugepage is a single PMD-level block descriptor — one entry regardless of size.
- On this system (AArch64, 64K pages): valid hugepage sizes are 2 MiB, 512 MiB, and 16 GiB. There are no 1 GiB hugepages (that's an x86/4K-page-base concept).

PR #2712, opened 2026-05-22
warm_reset: enable on AmpereOne, skip on other ARM platforms

lu-zero:
Remove the compile-time ARM guard that was preventing tt-smi -r from working on aarch64.  Warm reset works correctly on ARM (tested on P150A / Blackhole).

### DIFF
--- a/device/api/umd/device/warm_reset.hpp
+++ b/device/api/umd/device/warm_reset.hpp
@@ -41,6 +41,10 @@ public:
 
     static bool ubb_warm_reset(const std::chrono::milliseconds timeout_ms = timeout::UBB_WARM_RESET_TIMEOUT);
 
+    // Power-cycles devices by removing them from the PCIe bus and triggering a rescan.
+    // Requires CAP_SYS_ADMIN (root).  Returns false if unsupported or permission is denied.
+    static bool cold_reset(std::vector<int> pci_device_ids = {});
+
 private:
     static constexpr auto POST_RESET_WAIT = std::chrono::milliseconds(2'000);
     static constexpr auto UBB_POST_RESET_WAIT = std::chrono::milliseconds(30'000);
--- a/device/common/utils.hpp
+++ b/device/common/utils.hpp
@@ -10,6 +10,7 @@
 #include <array>
 #include <chrono>
 #include <cstring>
+#include <fstream>
 #include <optional>
 #include <stdexcept>
 #include <string>
@@ -276,6 +277,27 @@ public:
 constexpr bool is_arm_platform() {
 #if defined(__aarch64__) || defined(__arm__)
     return true;
+#else
+    return false;
+#endif
+}
+
+// Check whether the CPU is an Ampere custom core (AmpereOne family, implementer 0xc0).
+// Returns false for Ampere Altra (which uses ARM Neoverse N1 reference cores, implementer 0x41)
+// and for non-ARM platforms.  Reads /proc/cpuinfo once.
+inline bool is_ampereone() {
+#if defined(__aarch64__) || defined(__arm__)
+    static const bool result = [] {
+        std::ifstream f("/proc/cpuinfo");
+        std::string line;
+        while (std::getline(f, line)) {
+            if (line.find("CPU implementer") != std::string::npos && line.find("0xc0") != std::string::npos) {
+                return true;
+            }
+        }
+        return false;
+    }();
+    return result;
 #else
     return false;
 #endif
--- a/device/warm_reset.cpp
+++ b/device/warm_reset.cpp
@@ -20,6 +20,7 @@
 #include <cstring>
 #include <exception>
 #include <filesystem>
+#include <fstream>
 #include <functional>
 #include <map>
 #include <memory>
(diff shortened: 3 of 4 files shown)


lu-zero:
It is probably overly conservative, but better safer than sorry.

device/common/utils.hpp
```diff
@@ -281,6 +282,27 @@ constexpr bool is_arm_platform() {
...
+// Check whether the CPU is an Ampere custom core (AmpereOne family, implementer 0xc0).
+// Returns false for Ampere Altra (which uses ARM Neoverse N1 reference cores, implementer 0x41)
+// and for non-ARM platforms.  Reads /proc/cpuinfo once.
+inline bool is_ampereone() {
+#if defined(__aarch64__) || defined(__arm__)
+    static const bool result = [] {
+        std::ifstream f("/proc/cpuinfo");
+        std::string line;
+        while (std::getline(f, line)) {
+            if (line.find("CPU implementer") != std::string::npos && line.find("0xc0") != std::string::npos) {
```
pjanevskiTT, line 243: nit: consider having "CPU implementer" and "0xc0" into constants, since we are probably going to reuse it at some point

pjanevskiTT: as a more general comment for UMD team (not a topic for you), we should consider making a util that can query multiple CPU attributes

pjanevskiTT (approved):

broskoTT (approved):
Not to happy to enable something which we aren't testing in CI, but more unhappy to block somebody's environment

lu-zero:
While at it, I prepared a cold reset that works potentially on more systems.

tests/api/test_warm_reset.cpp
```diff
@@ -402,9 +402,6 @@ TEST(WarmResetTest, GalaxyWarmResetScratch) {
 }
 
 TEST(WarmResetTest, ClusterWarmReset) {
-    if constexpr (utils::is_arm_platform()) {
```
nbuncicTT, line 335: We don't test this in CI, and we don't have an ARM platform we can test.

I understand if you use these tests for your CI, but let's then modify this `if` for AmpereOne:
- If it's not AmpereOne, do the check for the ARM platform and skip the test
- If it is AmpereOne, omit the check for ARM platform

device/warm_reset.cpp
```diff
@@ -178,6 +183,62 @@ int wait_for_pci_bdf_to_reappear(
     return interface_id;
 }
 
+bool WarmReset::cold_reset(std::vector<int> pci_device_ids) {
```
nbuncicTT, line 187: The first part of the comment is for UMD to take this into account:
- Should we leave the defined static interfaces (warm_reset, warm_reset_chip_id, cold_reset, etc.) as they are
- Modify their static behavior in compile-time dependent on the platform we use (i.e. x86, ARM, RISCV, etc.)
- Add additional behavior injection (via polymorphism, templates, std::function etc.) for clients to define their own reset approach

Because this `cold_reset` is being done by setting a character in the filesystem, and is not the standard way we do `cold_reset`, nor do we define it for our hardware.

@lu-zero I have a couple of questions:
- Do you use tt-kmd as your kernel driver modified for ARM? Or there's a separate kernel driver you use (completely new, or forked version)? 
- If you do use tt-kmd, how does it handle cold reset: Does it go through the driver's reset path, or does it also fall back to the sysfs remove/rescan approach?

lu-zero: I use tt-kmd as is and it seems working. I directly went through the cold-reset via sysfs I wasn't aware of this feature of tt-kmd. It worked w/out changes so I had no reason to look into it.

lu-zero:
I rebased again and I'm using it quite a bit in the past days. if it could land would be great.

lu-zero:
@nbuncicTT rebased again, do we have a chance to land it?

device/api/umd/device/warm_reset.hpp
```diff
@@ -41,6 +41,10 @@ class WarmReset {
 
     static bool ubb_warm_reset(const std::chrono::milliseconds timeout_ms = timeout::UBB_WARM_RESET_TIMEOUT);
 
+    // Power-cycles devices by removing them from the PCIe bus and triggering a rescan.
+    // Requires CAP_SYS_ADMIN (root).  Returns false if unsupported or permission is denied.
+    static bool cold_reset(std::vector<int> pci_device_ids = {});
```
nbuncicTT: Please rename the name from `cold_reset` to `warm_reset_sysfs`.

device/warm_reset.cpp
```diff
@@ -179,6 +184,62 @@ int wait_for_pci_bdf_to_reappear(
...
+        pci_device_ids = PCIDevice::enumerate_devices();
+    }
+    if (pci_device_ids.empty()) {
+        log_info(LogUMD, "No PCI devices found.");
+        return false;
+    }
+
+    auto pci_devices_info = PCIDevice::enumerate_devices_info();
+    std::vector<std::string> bdfs;
+    for (auto id : pci_device_ids) {
```
nbuncicTT, line 198: The content of this for can be changed into:
```cpp
for (int id : pci_device_ids) {
    if (auto it = pci_devices_info.find(id); it != pci_devices_info.end()) {
        bdfs.push_back(it->second.pci_bdf);             // Reads directly from iterator
    }
}
```
This would make sense because then there is a single lookup.

nbuncicTT (commented):
@lu-zero please rename the function as said in the comments and address the rest of the comments, once that is done, I'll look into the PR again and approve it.

I think these are all minor changes and we should land it soon.

Note: Once you have addressed the comments, please resolve them, that way we know the PR is ready for review. If we want something more out of that topic, we can always unresolve it.

lu-zero:
Everything should be addressed now.

nbuncicTT (approved):

Issue #2727, opened 2026-05-27
UMD bump blocked


broskoTT:
**Describe the bug**
https://github.com/tenstorrent/tt-metal/pull/44725 bump is blocked

**Stack trace**
Seems like CMake configuration issue

**To reproduce**

**Expected behavior**
Galaxy unit tests pass

broskoTT:
Unblocked

PR #2188, opened 2026-02-27
Fix local googletest

RossComputerGuy:
### Issue
(Link to Github issue(s))

### Description
Fixes this error when building with `CPM_LOCAL_PACKAGES_ONLY`
```
tt-umd> CMake Error at cmake/CPM.cmake:728 (message):
tt-umd>   CPM: googletest not found via find_package(googletest 1.13.0)
tt-umd> Call Stack (most recent call first):
tt-umd>   third_party/CMakeLists.txt:19 (CPMAddPackage)
tt-umd>
tt-umd>
tt-umd> -- CPM: Adding package googletest@1.13.0 (v1.13.0)
tt-umd> CMake Warning at /nix/store/jzkw4x0b77fv2il53vpp6fylc09a6p75-cmake-4.1.2/share/cmake-4.1/Modules/FetchContent.cmake:2111 (message):
tt-umd>   FETCHCONTENT_FULLY_DISCONNECTED is set to true, which requires the source
tt-umd>   directory for dependency googletest to already be populated.  This
tt-umd>   generally means it must not be set to true the first time CMake is run in a
tt-umd>   build directory.  The following source directory should already be
tt-umd>   populated, but it doesn't exist:
tt-umd>
tt-umd>     /build/source/build/_deps/googletest-src
tt-umd>
tt-umd>   Policy CMP0170 controls enforcement of this requirement.
tt-umd> Call Stack (most recent call first):
tt-umd>   /nix/store/jzkw4x0b77fv2il53vpp6fylc09a6p75-cmake-4.1.2/share/cmake-4.1/Modules/FetchContent.cmake:2384 (__FetchContent_Populate)
tt-umd>   cmake/CPM.cmake:1114 (FetchContent_MakeAvailable)
tt-umd>   cmake/CPM.cmake:895 (cpm_fetch_package)
tt-umd>   third_party/CMakeLists.txt:19 (CPMAddPackage)
tt-umd>
tt-umd>
tt-umd> CMake Error at third_party/CMakeLists.txt:29 (target_compile_options):
tt-umd>   Cannot specify compile options for target "gtest" which is not built by
tt-umd>   this project.
```

And this as well:
```
tt-umd> CMake Error at tests/CMakeLists.txt:2 (target_link_libraries):
tt-umd>   The link interface of target "test_common" contains:
tt-umd>
tt-umd>     spdlog::spdlog_header_only
tt-umd>
tt-umd>   but the target was not found.  Possible reasons include:
tt-umd>
[...]

PR #2728, opened 2026-05-27
Fix GTest component name

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/2727

### Description
Aftermath of #2188 
Seems like the google test packages aren't matched now between metal and umd.
We might consume UMD in metal better in the future, but for now this is the only fix.

### List of the changes
- Changed the workflow which builds tt_metal to catch this failure
- Fixed the naming to match the one in metal.

### Testing
Failing run after added additional check, on first commit only on this PR: https://github.com/tenstorrent/tt-umd/actions/runs/26519304422/job/78105479689?pr=2728
Passing after the fix: https://github.com/tenstorrent/tt-umd/actions/runs/26519589569/job/78106522121

### API Changes
This PR has (no) API changes.

### DIFF
--- a/.github/workflows/run-ttsim-tests.yml
+++ b/.github/workflows/run-ttsim-tests.yml
@@ -106,7 +106,7 @@ jobs:
           echo "TT_METAL_HOME=$(pwd)" >> $GITHUB_ENV
           echo "TT_METAL_RUNTIME_ROOT=$(pwd)" >> $GITHUB_ENV
           echo "PYTHONPATH=$(pwd)" >> $GITHUB_ENV
-          ./build_metal.sh --build-all --enable-ccache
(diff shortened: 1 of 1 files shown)


pjanevskiTT (approved):

broskoTT:
@RossComputerGuy FYI

broskoTT:
Ah this was actually already fixed in metal by https://github.com/tenstorrent/tt-metal/pull/45167

RossComputerGuy:
> Seems like the google test packages aren't matched now between metal and umd.
> We might consume UMD in metal better in the future, but for now this is the only fix.

Cool, great. Glad to see downstream repos syncing up with the name. That'll make packaging with Nix easier. Btw, I should be able to work on the other PR tonight and hopefully we could merge that one soon.

PR #2729, opened 2026-05-27
feat(sim): core multichip support via shared dlopen, eth-MAC wiring, DRAM teleport

Riddy21:
## Summary

Split from #2694. Core multichip simulation support:

- **Shared dlopen**: v3.5 libttsim ABI with process-global `s_shared_handle_` and per-chip `libttsim_select_device`. Probe reuses existing handle; flags/refcount deferred until all dlsym calls succeed.
- **Eth-MAC wiring pre-pass**: Deterministic MAC assignment and bidirectional peer registration in Cluster constructor for virtual switch routing. Includes MAC overflow assert (`cid <= 0xFF && chan <= 0xFF`).
- **DRAM teleport**: Coordinate-based `libttsim_dram_core_{rd,wr}_bytes_by_id` with `TT_UMD_SIM_DRAM_TELEPORT` env var gating.
- **Closed-guard pattern**: `mark_closed()`/`is_closed()` on communicator; `close_device()` on TTSimTTDevice.
- **Fabric registration**: `register_sim_fabric_endpoint_direction` and `register_sim_fabric_node_id` passthroughs under `#ifdef TT_UMD_BUILD_SIMULATION`.

### Review fixes applied (from #2694)
1. Replaced class docstring with concise comment
2. Added multichip model documentation block
3. Documented `v3_5_multichip_mode_` purpose
4. Documented `device_lock_` legacy serialization limitation
5. Fixed probe to reuse `s_shared_handle_` instead of double-dlopen
6. Deferred `v3_5_multichip_mode_` and `s_shared_refcount_++` until after ALL dlsym calls
7. Added ABI tiering comment block (required vs optional symbols)
8. Added `!peer_dev` guard in `register_peer`
9. Added MAC overflow assert in eth-MAC wiring
10. Moved `options_ = std::move(options)` after `construct_cluster()`
11. Documented `callback_instance_` as known limitation
12. Added `closed_` member with `mark_closed()`/`is_closed()`
13. Removed `// NOLINT(performance-unnecessary-value-param)` annotation

## Test plan
- [ ] Build with `TT_UMD_BUILD_SIMULATION=ON`
- [ ] Verify single-chip simulation still works
- [ ] Verify multichip simulation with v3.5 libttsim
- [ ] Verify DRAM teleport with `TT_UMD_SIM_DRAM_TELEPORT=1`

### DIFF
--- a/tests/api/test_sim_multichip.cpp
+++ b/tests/api/test_sim_multichip.cpp
@@ -39,7 +39,7 @@ class IsCoreOfTypeTest : public ::testing::TestWithParam<ARCH> {};
 
 TEST_P(IsCoreOfTypeTest, DramCoreIsIdentifiedCorrectly) {
     const ARCH arch = GetParam();
-    const std::string sdesc_path = get_soc_descriptor_path(arch);
+    const std::string sdesc_path = test_utils::get_soc_descriptor_path(arch);
     SocDescriptor soc(std::make_shared<SocArchDescriptor>(sdesc_path));
 
     // Every core returned by get_cores(DRAM, TRANSLATED) must satisfy is_core_of_type.
@@ -55,7 +55,7 @@ TEST_P(IsCoreOfTypeTest, DramCoreIsIdentifiedCorrectly) {
 
 TEST_P(IsCoreOfTypeTest, TensixCoreIsNotDram) {
     const ARCH arch = GetParam();
-    const std::string sdesc_path = get_soc_descriptor_path(arch);
+    const std::string sdesc_path = test_utils::get_soc_descriptor_path(arch);
     SocDescriptor soc(std::make_shared<SocArchDescriptor>(sdesc_path));
 
     auto tensix_cores = soc.get_cores(CoreType::TENSIX, CoordSystem::TRANSLATED);
@@ -72,7 +72,7 @@ TEST_P(IsCoreOfTypeTest, TensixCoreIsNotDram) {
 
 TEST_P(IsCoreOfTypeTest, EthCoreIsIdentifiedCorrectly) {
     const ARCH arch = GetParam();
-    const std::string sdesc_path = get_soc_descriptor_path(arch);
+    const std::string sdesc_path = test_utils::get_soc_descriptor_path(arch);
     SocDescriptor soc(std::make_shared<SocArchDescriptor>(sdesc_path));
 
     auto eth_cores = soc.get_cores(CoreType::ETH, CoordSystem::TRANSLATED);
--- a/tests/api/test_sim_multichip.cpp
+++ b/tests/api/test_sim_multichip.cpp
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: (c) 2025 Tenstorrent Inc.
+// SPDX-FileCopyrightText: © 2026 Tenstorrent Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
 
@@ -114,10 +114,13 @@ TEST_P(IsCoreOfTypeTest, GarbageXYIsNotAnyKnownType) {
     EXPECT_FALSE(soc.is_core_of_type(garbage, CoreType::PCIE, CoordSystem::TRANSLATED));
 }
 
+// QUASAR is included because is_core_of_type's DRAM-teleport path in
+// tt_sim_tt_device.cpp special-cases it, and a quasar_simulation_1x1.yaml
+// descriptor is available in the repo.
 INSTANTIATE_TEST_SUITE_P(
     AllArchs,
     IsCoreOfTypeTest,
-    ::testing::Values(ARCH::WORMHOLE_B0, ARCH::BLACKHOLE),
+    ::testing::Values(ARCH::WORMHOLE_B0, ARCH::BLACKHOLE, ARCH::QUASAR),
     [](const ::testing::TestParamInfo<ARCH>& info) { return arch_to_str(info.param); });
 
 // ---------------------------------------------------------------------------
@@ -139,24 +142,6 @@ protected:
     const char* simulator_path_ = nullptr;
 };
 
-// Verify that two communicators targeting different chip IDs can be
-// constructed from the same simulator directory. In multichip mode they
-// share a single dlopen handle (refcount = 2).
-TEST_F(TTSimCommunicatorTest, TwoCommunicatorsShareHandle) {
-    // chip_id 0 and 1 -- does not need a real multi-chip sim binary;
-    // the constructor merely stores chip_id, actual multichip detection
-    // happens in initialize() which we skip here to avoid needing a
-    // live simulator.
-    TTSimCommunicator comm_0(simulator_path_, /* copy_sim_binary= */ false, /* chip_id= */ 0);
-    TTSimCommunicator comm_1(simulator_path_, /* copy_sim_binary= */ false, /* chip_id= */ 1);
-
-    // Both communicators exist without throwing.
-    // If the .so supports multichip ABI, initialize() would set them up
-    // with shared dlopen. Without calling initialize() we just verify
-    // construction is safe.
-    SUCCEED();
-}
-
 // Verify that TTSimTTDevice::create produces a non-null device.
 TEST_F(TTSimCommunicatorTest, CreateSimDevice) {
     auto device = TTSimTTDevice::create(simulator_path_);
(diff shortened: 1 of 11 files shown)


tests/api/test_sim_multichip.cpp
```diff
@@ -0,0 +1,239 @@
...
+    // happens in initialize() which we skip here to avoid needing a
+    // live simulator.
+    TTSimCommunicator comm_0(simulator_path_, /* copy_sim_binary= */ false, /* chip_id= */ 0);
+    TTSimCommunicator comm_1(simulator_path_, /* copy_sim_binary= */ false, /* chip_id= */ 1);
+
+    // Both communicators exist without throwing.
+    // If the .so supports multichip ABI, initialize() would set them up
+    // with shared dlopen. Without calling initialize() we just verify
+    // construction is safe.
+    SUCCEED();
```
pjanevskiTT: This test never calls `initialize()`, but the shared-dlopen handle and `s_shared_refcount_` logic lives entirely in `initialize()` and the destructor. The constructor only stores `simulator_directory_`/`copy_sim_binary_`/`chip_id_` — no dlopen, no probe; `multichip_mode_` stays `false`, and the destructor takes the `else if (libttsim_handle_)` branch with a null handle (no-op). So "they share a single dlopen handle (refcount = 2)" is never exercised — this would pass even if the entire shared-handle path were deleted. Either drive `initialize()` and assert on observable state, or drop the test.

tests/api/test_sim_multichip.cpp
```diff
@@ -0,0 +1,239 @@
...
+// simulator path but different chip_ids does not crash on destruction.
+// This exercises the shared dlopen refcount (2 -> 1 -> 0).
+TEST_F(TTSimCommunicatorTest, SharedDlopenRefcount) {
+    {
+        TTSimCommunicator comm_a(simulator_path_, /* copy_sim_binary= */ false, /* chip_id= */ 0);
+        TTSimCommunicator comm_b(simulator_path_, /* copy_sim_binary= */ false, /* chip_id= */ 1);
+        // Both alive; shared refcount should be 2.
+    }
+    // Both destroyed; refcount should have gone 2 -> 1 -> 0 without crash.
+    SUCCEED();
```
pjanevskiTT: Same issue as `TwoCommunicatorsShareHandle`: `s_shared_refcount_` never leaves 0 because `initialize()` is never called, so the "2 → 1 → 0" the comment describes never happens. The bare `SUCCEED()` validates only that a member-init constructor doesn't throw.

tests/api/test_sim_multichip.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+    device->close_device();
+
+    // A second close should not crash.
+    EXPECT_NO_THROW(device->close_device());
+}
+
+// Two TTSimTTDevice instances writing and reading back independently.
+// This implicitly exercises select_chip_if_needed for each I/O call.
+TEST_F(TTSimCommunicatorTest, TwoDevicesIndependentIO) {
```
pjanevskiTT, line 186: `TTSimTTDevice::create()` hardcodes `chip_id = 0` (the `make_unique<TTSimTTDevice>(..., 0, ...)` call in `tt_sim_tt_device.cpp`), so both `dev_0` and `dev_1` get `chip_id_ == 0`. In real multichip mode `select_chip_if_needed()` then selects the *same* simulated chip for both, so writing `pattern_0` then `pattern_1` to the same core/addr makes `readback_0 == pattern_1` and the `EXPECT_EQ(readback_0, pattern_0)` at line 228 would fail. The test can only pass in legacy separate-dlopen mode — the mode where no chip selection happens — which is the opposite of what the name/comment claim. To actually test independence the two devices need distinct `chip_id`s (a `create`/ctor path that accepts chip_id).

tests/api/test_sim_multichip.cpp
```diff
@@ -0,0 +1,239 @@
...
+// Create a TTSimTTDevice, close it, then verify that close_device() can be
+// called again without crashing (idempotency via closed_ flag).
+TEST_F(TTSimCommunicatorTest, IOAfterCloseIsSafe) {
+    auto device = TTSimTTDevice::create(simulator_path_);
+    ASSERT_NE(device, nullptr);
+
+    device->close_device();
+
+    // A second close should not crash.
+    EXPECT_NO_THROW(device->close_device());
```
pjanevskiTT, line 180: The "idempotency via closed_ flag" claim isn't accurate. `close_device()` calls `mark_closed()` then `shutdown()`, and `shutdown()` checks only `multichip_mode_`, never `closed_`/`is_closed()`. So in single-chip mode the second `close_device()` calls `pfn_libttsim_exit_()` a second time (and `~TTSimTTDevice` calls `shutdown()` a third time). The `closed_` flag only guards `write_to_device`/`read_from_device`. So the idempotency this observes doesn't come from the mechanism named — and it may be masking a real double-`libttsim_exit`. Consider asserting the actual guard (a write after close is a no-op) and/or making `shutdown()` honor `closed_`.

tests/api/test_sim_multichip.cpp
```diff
@@ -0,0 +1,239 @@
...
+    EXPECT_FALSE(soc.is_core_of_type(garbage, CoreType::TENSIX, CoordSystem::TRANSLATED));
+    EXPECT_FALSE(soc.is_core_of_type(garbage, CoreType::ETH, CoordSystem::TRANSLATED));
+    EXPECT_FALSE(soc.is_core_of_type(garbage, CoreType::ARC, CoordSystem::TRANSLATED));
+    EXPECT_FALSE(soc.is_core_of_type(garbage, CoreType::PCIE, CoordSystem::TRANSLATED));
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    AllArchs,
+    IsCoreOfTypeTest,
+    ::testing::Values(ARCH::WORMHOLE_B0, ARCH::BLACKHOLE),
```
pjanevskiTT: `is_core_of_type` was extracted specifically for the DRAM-teleport path in `write_to_device`/`read_from_device`, and that production code special-cases QUASAR repeatedly. QUASAR is the most simulation-relevant arch and a descriptor is available (`fetch_local_files.hpp`, `quasar_simulation_1x1.yaml`), yet the params only cover WORMHOLE_B0 and BLACKHOLE. Consider adding `ARCH::QUASAR`.

tests/api/test_sim_multichip.cpp
```diff
@@ -0,0 +1,239 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent Inc.
```
pjanevskiTT: Nit: tests/api uniformly uses `©` (not `(c)`) and new 2026 files use `© 2026`. This says `(c) 2025`.

pjanevskiTT (approved):

tests/api/test_sim_multichip.cpp
Riddy21: Fixed in 53dd8ae — copyright updated to `© 2026`.

tests/api/test_sim_multichip.cpp
Riddy21: Added `ARCH::QUASAR` to the `INSTANTIATE_TEST_SUITE_P` call. Added a comment explaining why — the DRAM-teleport path in `tt_sim_tt_device.cpp` special-cases QUASAR, and `quasar_simulation_1x1.yaml` is available in the repo. Default `ChipInfo{}` is sufficient for QUASAR (no harvesting constraints).

tests/api/test_sim_multichip.cpp
Riddy21: Dropped both `TwoCommunicatorsShareHandle` and `SharedDlopenRefcount` — they never called `initialize()` so the refcount path was never exercised and both tests passed even if the entire shared-handle path were deleted. The only meaningful communicator tests now are `CreateSimDevice`, `WriteAfterCloseIsNoOp`, and `TwoDevicesIndependentIO`.

tests/api/test_sim_multichip.cpp
Riddy21: Same as above — `SharedDlopenRefcount` dropped alongside `TwoCommunicatorsShareHandle`.

tests/api/test_sim_multichip.cpp
Riddy21: Fixed on two fronts: (1) `shutdown()` now checks `closed_` at the top and returns early, so neither the second `close_device()` call nor the destructor re-invokes `pfn_libttsim_exit_`. (2) The test is now `WriteAfterCloseIsNoOp` — it asserts that `write_to_device` after `close_device()` is a no-op (the actual `closed_` guard), that a second `close_device()` does not crash, and that the destructor (which also calls `shutdown()`) is safe.

tests/api/test_sim_multichip.cpp
Riddy21: Fixed: added a `create(path, chip_id, copy_sim_binary)` overload to `TTSimTTDevice` and updated `TwoDevicesIndependentIO` to use it with `chip_id=0` and `chip_id=1`. The original `create()` now delegates to the new overload with `chip_id=0`.

tests/api/test_sim_multichip.cpp
Riddy21: Fixed in 53dd8ae — copyright updated to `© 2026`.

tests/api/test_sim_multichip.cpp
Riddy21: Added `ARCH::QUASAR` to `INSTANTIATE_TEST_SUITE_P`. Added a comment explaining why: the DRAM-teleport path in `tt_sim_tt_device.cpp` special-cases QUASAR, and `quasar_simulation_1x1.yaml` is available. Default `ChipInfo{}` is sufficient for QUASAR (no harvesting constraints like BH).

tests/api/test_sim_multichip.cpp
Riddy21: Dropped both `TwoCommunicatorsShareHandle` and `SharedDlopenRefcount` — neither called `initialize()` so the shared-handle refcount path was never exercised and the tests passed even if the entire path were deleted. The communicator tests are now `CreateSimDevice`, `WriteAfterCloseIsNoOp`, and `TwoDevicesIndependentIO`.

tests/api/test_sim_multichip.cpp
Riddy21: Dropped alongside `TwoCommunicatorsShareHandle` for the same reason.

tests/api/test_sim_multichip.cpp
Riddy21: Fixed on two fronts: (1) `TTSimCommunicator::shutdown()` now returns early when `closed_` is set, preventing the destructor from re-invoking `pfn_libttsim_exit_`. (2) The test is now `WriteAfterCloseIsNoOp` — it asserts that `write_to_device` after `close_device()` is a no-op (exercising the actual `closed_` guard), and that a second `close_device()` plus the destructor are safe.

tests/api/test_sim_multichip.cpp
Riddy21: Fixed: added `TTSimTTDevice::create(path, chip_id, copy_sim_binary)` overload and updated `TwoDevicesIndependentIO` to use it with `chip_id=0` and `chip_id=1`. The original `create(path, num_host_mem_channels, copy_sim_binary)` now delegates to the new overload with `chip_id=0`.

Riddy21:
Fixed a compile error: `ChipId` is an alias for `int`, so `create(path, ChipId, bool)` and `create(path, int num_host_mem_channels, bool)` have identical signatures — causing a *class member cannot be redeclared* error on all toolchains. Renamed the multichip factory to `create_for_chip()` to disambiguate. Tests updated accordingly.

PR #2730, opened 2026-05-27
feat(sim): support mapped simulation sysmem buffers

Riddy21:
## Summary

Split from #2694. Adds mapped buffer management to `SimulationSysmemManager` for simulation environments that bypass the hugepage channel layout. Adds a simulation-only `SysmemBuffer` constructor with an unmap callback.

- **SimulationSysmemManager**: `allocate_sysmem_buffer` / `map_sysmem_buffer` with synthetic IOVAs, `write_to_sysmem` / `read_from_sysmem` overrides with two-path lookup (mapped buffers first, then base class hugepage path).
- **SysmemBuffer**: New constructor `(void* buffer_va, size_t, uint64_t device_io_addr, optional<uint64_t> noc_addr, function<void()> unmap_callback)` for simulation use. Destructor invokes callback instead of PCI unmap when present.

### Review fixes applied (from #2694)
1. Removed `#include <iostream>`
2. Added `constexpr uint64_t kHostMemChannelSize` with descriptive comment
3. Added comments explaining two-path logic in write/read_to_sysmem
4. Fixed data race: `owned_allocations_.push_back` now under `std::lock_guard`
5. Added TODO for dangling `this` capture in unmap callback
6. Renamed `find_mapped_buffer` -> `find_mapped_buffer_locked` with caller-must-hold-mutex comment
7. Added `TracyAllocN` to simulation constructor (fixes alloc/free mismatch with destructor's `TracyFreeN`)

## Test plan
- [ ] Build with `TT_UMD_BUILD_SIMULATION=ON`
- [ ] Verify simulation sysmem buffer allocation and mapped read/write
- [ ] Verify Tracy profiling shows balanced alloc/free

### DIFF
--- a/device/chip_helpers/simulation_sysmem_manager.cpp
+++ b/device/chip_helpers/simulation_sysmem_manager.cpp
@@ -6,23 +6,18 @@
 
 #include <fmt/format.h>
 #include <sys/mman.h>  // for mmap, munmap
-#include <sys/stat.h>  // for fstat
 #include <unistd.h>
 
 #include <algorithm>
 #include <cstddef>
 #include <cstdint>
 #include <cstring>
-#include <filesystem>
-#include <fstream>
 #include <memory>
-#include <string>
 #include <vector>
 
 #include "hugepage.hpp"
 #include "tracy.hpp"
 #include "umd/device/chip_helpers/sysmem_buffer.hpp"
-#include "umd/device/types/cluster_types.hpp"
 #include "umd/device/utils/error.hpp"
 
 namespace tt {
@@ -32,10 +27,6 @@ enum class ARCH;
 namespace tt::umd {
 
 namespace {
-
-// Channel stride matches the hugepage region size used by SysmemManager
-// and the pci_dma_{read,write}_bytes address layout.
-constexpr uint64_t kHostMemChannelSize = HUGEPAGE_REGION_SIZE;
 
 uint64_t align_up(uint64_t value, uint64_t alignment) { return (value + alignment - 1) & ~(alignment - 1); }
 
@@ -72,7 +63,12 @@ bool SimulationSysmemManager::init_sysmem(uint32_t num_host_mem_channels) {
     UMD_ASSERT(system_memory_ != MAP_FAILED, error::RuntimeError, "system_memory mmap() failed");
     madvise(system_memory_, total_size, MADV_HUGEPAGE);
     system_memory_size_ = total_size;
-    registry_->next_base = total_size;
+
+    // The mapped-buffer arena starts immediately after the hugepage region so
+    // that device IO addresses assigned to mapped buffers (pcie_base_ + arena
+    // offset) never alias a hugepage channel address (pcie_base_ + channel *
+    // 1 GB, for channel in [0, num_channels)).
+    registry_->next_arena_offset = total_size;
 
     for (int i = 0; i < num_host_mem_channels; i++) {
         size_t channel_size = (i == 3 && num_host_mem_channels == 4) ? (768 * (1ULL << 20)) : (1ULL << 30);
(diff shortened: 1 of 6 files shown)


device/chip_helpers/simulation_sysmem_manager.cpp
```diff
@@ -76,14 +105,105 @@ void SimulationSysmemManager::unpin_or_unmap_sysmem() {
...
+    for (const auto& mapped_buffer : registry_->buffers) {
+        if (base >= mapped_buffer.base && base + size <= mapped_buffer.base + mapped_buffer.size) {
+            return mapped_buffer;
+        }
+    }
+    return std::nullopt;
+}
+
+void SimulationSysmemManager::write_to_sysmem(uint16_t channel, const void* src, uint64_t sysmem_dest, uint32_t size) {
+    // Two distinct backing stores exist in simulation:
```
pjanevskiTT: ok I see why we had to override this. But still few comments for this

1. We are doing override but this is not behaving the same way as regular write_to_sysmem. sysmem_dest should be the offset inside channel in the system memory. I know channel thing in sysmem doesn't make a lot of sense, but I would say this is not behaving the same way as in regular SysmemManager class
2. That being said, I propose few alternatives
- Why not use write_to_sysmem directly inside SysmemBuffer?
- If that becomes really hard, I would like to see a comment explaining that the behavior is different for simulation sysmem in the .hpp file. We are going to refactor this part soon so we will follow up on how to properly use new structures

tests/api/test_simulation_sysmem_manager.cpp
```diff
@@ -91,3 +94,136 @@ TEST(ApiSimulationSysmemManager, TestFourChannels) {
...
+    // Read it back.
+    std::vector<uint8_t> readback(pattern.size(), 0);
+    sysmem->read_from_sysmem(channel, readback.data(), offset_in_channel, readback.size());
+    EXPECT_EQ(pattern, readback);
+
+    // Also verify the data is visible through the buffer VA directly.
+    EXPECT_EQ(0, std::memcmp(buffer->get_buffer_va(), pattern.data(), pattern.size()));
+}
+
+TEST(ApiSimulationSysmemManager, MultipleMappedBuffersAreIndependent) {
```
pjanevskiTT: these tests are all for wormhole, should we parametrize it to support BH as well?

pjanevskiTT (commented):
I would just like to converge to a solution for read/write to sysmem. Otherwise this seems good

pjanevskiTT (approved):

device/chip_helpers/simulation_sysmem_manager.cpp
Riddy21: Good catch — addressed in the latest commit. The root issue was that  /  were being overridden with different semantics than the base class (which treats  as a within-channel offset, not an absolute address). This also masked a latent bug: the old channel decomposition skipped subtracting , so it would produce wrong channel/offset values on WH (pcie_base = 0x800000000) and BH.

The fix moves the mapped-buffer lookup one level up to , where the raw  is still available. Two new public methods —  /  — check the registry by device IO address and memcpy on a hit. On a miss,  is subtracted before decomposing into  and forwarding to the unoverridden base-class methods. The overrides are removed.

Riddy21:
Fixed in the latest commit (ac4659da).

**Root cause of ttsim CI failure:** craq-sim's `translate_pci_dma_addr()` *subtracts* `pcie_base` before invoking the DMA callbacks, so `paddr` in `pci_dma_{read,write}_bytes` is already an offset from `pcie_base`, not an absolute address. The previous refactor incorrectly asserted `paddr >= pcie_base` and subtracted it a second time.

**Fix:** When looking up the mapped-buffer registry (keyed by absolute `device_io_addr = pcie_base + arena_offset`), we now add `pcie_base` back to `paddr`. The hugepage channel decomposition uses `paddr` directly (it's already the offset).

**Test fixes (Pavle's second comment):**
- Rewrote `WriteReadThroughMappedBuffer` to call `write_mapped_buffer`/`read_mapped_buffer` directly — these are the methods that `pci_dma_*_bytes` delegates to, and they accept the absolute `device_io_addr`
- Parametrized all mapped-buffer tests for both `WORMHOLE_B0` and `BLACKHOLE`

PR #2731, opened 2026-05-27
feat: add Cluster factory methods and fix ClusterOptions defaults

Riddy21:
### PR Category
Bug fix / Cleanup

### Summary

Split from #2694. Two targeted changes to `Cluster` and `ClusterOptions` for better ABI stability and correctness.

**1. `num_host_mem_ch_per_mmio_device` default: `0` → `std::nullopt`**

Previously the field defaulted to `0`, suppressing auto-detection of host memory channels. With `std::nullopt`, callers that don't set this field get the correct auto-calculate path in `construct_cluster` (reads max chips-per-MMIO from the cluster descriptor). The old `0` default was a latent bug — callers that want zero channels must now set it explicitly.

**2. `explicit Cluster(ClusterOptions)` + split default ctor**

`Cluster(ClusterOptions options = {})` is replaced by:
- `Cluster() : Cluster(ClusterOptions{}) {}` — explicit default ctor, no behavior change
- `explicit Cluster(ClusterOptions options)` — prevents unintended implicit conversions from `ClusterOptions` at call sites (cross-DSO ABI hazard)

Any caller using implicit conversion (`Cluster c = some_options;`) must change to `Cluster c(some_options);`.

**3. New test: `test_cluster_options.cpp`**

Verifies `ClusterOptions{}` defaults (`num_host_mem_ch_per_mmio_device == std::nullopt`, `chip_type == SILICON`, empty target devices) and that explicit construction compiles.

### ⚠️ Breaking behavior change: `ClusterOptions{}` default

Callers that used `ClusterOptions{}` without setting `num_host_mem_ch_per_mmio_device` and relied on the `0` value (no host memory channels) will now hit the auto-calculate path. To preserve old behavior, explicitly set `options.num_host_mem_ch_per_mmio_device = 0`.

**tt-metal audit** (`tt_metal/llrt/tt_cluster.cpp`, `tt_metal/distributed/pcie_core_writer.cpp`):

| File | Path | Sets field explicitly? | Impact |
|------|------|------------------------|--------|
| `tt_cluster.cpp:392` | Silicon | ✅ `= std::nullopt` | No change |
| `tt_cluster.cpp:400` | Simulator (mock desc) | ✅ `= 1` | No change |
| `tt_cluster.cpp:408` | Simulator (no desc) | ✅ `= 1` | No change |
| `tt_cluster.cpp:421` | Mock cluster | ❌ not set | Was `0`, now auto-calculate |
| `tt_cluster.cpp:431` | Emule cluster | ❌ not set | Was `0`, now auto-calculate |
| `pcie_core_writer.cpp:21` | PCIe core writer | ❌ not set | Was `0`, now auto-calculate (correct) |

Mock/Emule paths in tt-metal `main` should add `num_host_mem_ch_per_mmio_device = 0` before consuming this UMD change.

### ⚠️ API change: `explicit Cluster(ClusterOptions)`

Source-breaking for implicit conversions:
```cpp
// Before (compiled):
Cluster c = some_options;
foo(some_options);  // if foo takes Cluster

// After (must change to):
Cluster c(some_options);
foo(Cluster(some_options));
```

All tt-metal callers use `std::make_unique<Cluster>(ClusterOptions{...})` — no impact.

### Notes for reviewers
- The `test_cluster_options.cpp` `ExplicitConstructorCompiles` test catches exceptions on construction (Cluster may throw if no real hardware present in CI) — `SUCCEED()` in the catch is intentional.
- Factory methods (`create_silicon_cluster`, etc.) are part of #2694 and not included here.

### Test plan
- [x] New `ClusterOptions` unit tests in `test_cluster_options.cpp`
- [ ] Existing API test suite passes
- [ ] Audit Mock/Emule call sites in tt-metal `main` for `num_host_mem_ch_per_mmio_device = 0` follow-up

### DIFF
--- a/device/api/umd/device/cluster.hpp
+++ b/device/api/umd/device/cluster.hpp
@@ -84,7 +84,7 @@ struct ClusterOptions {
      * If not provided, the value is determined automatically by determining the max
      * amount of chips connected through one PCIe interface in the ClusterDescriptor.
      */
-    std::optional<uint32_t> num_host_mem_ch_per_mmio_device = 0;
+    std::optional<uint32_t> num_host_mem_ch_per_mmio_device = std::nullopt;
 
     /**
      * If set, this soc descriptor will be used to construct devices on this cluster. If not set, the default soc
@@ -142,7 +142,11 @@ public:
      * - Perform this for each of the chips connected to the system.
      * @param options See documentation of ClusterOptions for explanation of specific arguments.
      */
-    Cluster(ClusterOptions options = {});
+    Cluster() : Cluster(ClusterOptions{}) {}
+
+    // Explicit to prevent implicit conversion from ClusterOptions at call sites.
+    explicit Cluster(ClusterOptions options);
+
     /**
      * Closing the device. Should undo everything that was done in the constructor. Break created links, free memory,
      * leave the device in a state where it can be re-initialized.
--- a/tests/api/test_cluster_options.cpp
+++ b/tests/api/test_cluster_options.cpp
@@ -5,38 +5,35 @@
 #include <gtest/gtest.h>
 
 #include <optional>
+#include <type_traits>
 
 #include "umd/device/cluster.hpp"
 #include "umd/device/types/cluster_types.hpp"
 
 using namespace tt::umd;
 
-TEST(ClusterOptions, DefaultHasNulloptHostMemChannels) {
+// Verifies the documented defaults of a freshly constructed ClusterOptions{}.
+TEST(ClusterOptions, Defaults) {
     ClusterOptions opts{};
+
+    // Host memory channels default to nullopt so the auto-calculate path runs in construct_cluster.
     EXPECT_EQ(opts.num_host_mem_ch_per_mmio_device, std::nullopt);
-}
-
-TEST(ClusterOptions, DefaultIsSilicon) {
-    ClusterOptions opts{};
     EXPECT_EQ(opts.chip_type, ChipType::SILICON);
-}
-
-TEST(ClusterOptions, DefaultHasEmptyTargetDevices) {
-    ClusterOptions opts{};
     EXPECT_TRUE(opts.target_devices.empty());
+
+    // An explicit 0 is distinct from nullopt and must be preserved.
+    ClusterOptions explicit_zero{.num_host_mem_ch_per_mmio_device = 0};
+    ASSERT_TRUE(explicit_zero.num_host_mem_ch_per_mmio_device.has_value());
+    EXPECT_EQ(explicit_zero.num_host_mem_ch_per_mmio_device.value(), 0u);
 }
 
-TEST(ClusterOptions, ExplicitZeroHostMemChannelsIsNotNullopt) {
-    ClusterOptions opts{.num_host_mem_ch_per_mmio_device = 0};
-    ASSERT_TRUE(opts.num_host_mem_ch_per_mmio_device.has_value());
-    EXPECT_EQ(opts.num_host_mem_ch_per_mmio_device.value(), 0u);
-}
+// Cluster(ClusterOptions) must be explicit to prevent implicit conversions at call sites.
+TEST(ClusterTest, ConstructorIsExplicit) {
+    // Explicit construction from ClusterOptions must be allowed.
+    EXPECT_TRUE((std::is_constructible_v<Cluster, ClusterOptions>))
+        << "Cluster should be explicitly constructible from ClusterOptions.";
 
-TEST(ClusterOptions, ExplicitConstructorCompiles) {
-    try {
-        Cluster c(ClusterOptions{});
-        (void)c;
-    } catch (const std::exception&) {
-        SUCCEED();
-    }
+    // Implicit conversion from ClusterOptions must be blocked.
+    EXPECT_FALSE((std::is_convertible_v<ClusterOptions, Cluster>))
+        << "API Violation: Cluster(ClusterOptions) must be explicit to prevent implicit conversion!";
 }
--- a/tests/api/test_cluster_options.cpp
+++ b/tests/api/test_cluster_options.cpp
@@ -1,0 +1,42 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include <gtest/gtest.h>
+
+#include <optional>
+
+#include "umd/device/cluster.hpp"
+#include "umd/device/types/cluster_types.hpp"
+
+using namespace tt::umd;
+
+TEST(ClusterOptions, DefaultHasNulloptHostMemChannels) {
+    ClusterOptions opts{};
+    EXPECT_EQ(opts.num_host_mem_ch_per_mmio_device, std::nullopt);
+}
+
+TEST(ClusterOptions, DefaultIsSilicon) {
+    ClusterOptions opts{};
+    EXPECT_EQ(opts.chip_type, ChipType::SILICON);
+}
+
+TEST(ClusterOptions, DefaultHasEmptyTargetDevices) {
+    ClusterOptions opts{};
+    EXPECT_TRUE(opts.target_devices.empty());
+}
+
+TEST(ClusterOptions, ExplicitZeroHostMemChannelsIsNotNullopt) {
+    ClusterOptions opts{.num_host_mem_ch_per_mmio_device = 0};
+    ASSERT_TRUE(opts.num_host_mem_ch_per_mmio_device.has_value());
+    EXPECT_EQ(opts.num_host_mem_ch_per_mmio_device.value(), 0u);
+}
+
+TEST(ClusterOptions, ExplicitConstructorCompiles) {
+    try {
+        Cluster c(ClusterOptions{});
+        (void)c;
+    } catch (const std::exception&) {
+        SUCCEED();
+    }
+}
--- a/tests/api/CMakeLists.txt
+++ b/tests/api/CMakeLists.txt
@@ -9,6 +9,7 @@ set(API_TESTS_SRCS
     test_sysmem_manager.cpp
     test_simulation_sysmem_manager.cpp
     test_sim_multichip.cpp
+    test_cluster_options.cpp
     test_tt_device.cpp
     test_spi_tt_device.cpp
     test_noc.cpp


Riddy21:
## tt-metal validation — behavior change impact

Audited all 6 `ClusterOptions` call sites in [tenstorrent/tt-metal](https://github.com/tenstorrent/tt-metal) (`tt_metal/llrt/tt_cluster.cpp` and `tt_metal/distributed/pcie_core_writer.cpp`).

### `explicit Cluster(ClusterOptions)` — ✅ No breakage
All tt-metal callers construct `ClusterOptions` inline (aggregate brace-init) and pass directly to the `Cluster` constructor via `std::make_unique`. No implicit conversions exist anywhere. This change is safe.

### `num_host_mem_ch_per_mmio_device` default: `0` → `std::nullopt` — ⚠️ 3 affected call sites

| File | Path | Explicit value set? | Impact |
|------|------|---------------------|--------|
| `tt_cluster.cpp:392` | Silicon | ✅ `.num_host_mem_ch_per_mmio_device = std::nullopt` | No change |
| `tt_cluster.cpp:400` | Simulator (mock) | ✅ `.num_host_mem_ch_per_mmio_device = 1` | No change |
| `tt_cluster.cpp:408` | Simulator (no mock) | ✅ `.num_host_mem_ch_per_mmio_device = 1` | No change |
| `tt_cluster.cpp:421` | Mock cluster | ❌ not set | **Was `0`, now auto-calculate** |
| `tt_cluster.cpp:431` | Emule cluster | ❌ not set | **Was `0`, now auto-calculate** |
| `pcie_core_writer.cpp:21` | PCIe core writer | ❌ not set | **Was `0`, now auto-calculate** |

**Affected callers:**

**Mock/Emule:** These create clusters backed by a mock descriptor with no real PCIe hardware. Previously they got 0 host-memory channels (safe for no-hardware paths). With `nullopt`, they'll auto-calculate from the descriptor — which may attempt to allocate hugepage channels that don't exist on a mock device. Recommend fixing these by explicitly setting `0` or leaving `nullopt` only if `construct_cluster` handles non-MMIO chips gracefully.

**PCIe core writer (`pcie_core_writer.cpp:21`):** This is a real silicon path. Auto-calculate should be correct behavior here — it will allocate the right number of channels based on the cluster. The old `0` default was likely a latent bug (zero host channels on a silicon device).

**Proposed fix for tt-metal (companion change needed):**
```cpp
// tt_cluster.cpp:421 — Mock path: explicit 0 to preserve old behavior
device_driver = std::make_unique<tt::umd::Cluster>(tt::umd::ClusterOptions{
    .chip_type = tt::umd::ChipType::MOCK,
    .num_host_mem_ch_per_mmio_device = 0,  // Mock has no real host memory channels
    .sdesc_path = sdesc_path,
    .cluster_descriptor = mock_cluster_desc.get(),
});

// tt_cluster.cpp:431 — Emule path: same
device_driver = std::make_unique<tt::umd::Cluster>(tt::umd::ClusterOptions{
    .chip_type = tt::umd::ChipType::SWEMULE,
    .num_host_mem_ch_per_mmio_device = 0,  // Emule has no real host memory channels
    .sdesc_path = sdesc_path,
    .cluster_descriptor = mock_cluster_desc.get(),
});
```

`pcie_core_writer.cpp:21` should be left as-is (auto-calculate is correct there).

cc @pjanevskiTT

Riddy21:
**Update:** The companion tt-metal PR (tenstorrent/tt-metal#44889, branch `ridvan/nkapre-multichip-metal-v2`) already uses the factory methods (`create_mock_cluster`, `create_swemule_cluster`, `create_silicon_cluster`, etc.) and does not use bare `ClusterOptions{}` at all. The factories control `num_host_mem_ch_per_mmio_device` internally so the default change has no impact on the companion PR.

The main branch of tt-metal (`tt_cluster.cpp` on `main`) does still use bare `ClusterOptions{}` for Mock/Emule paths without setting `num_host_mem_ch_per_mmio_device` — that would need a follow-up fix before this UMD change merges into UMD main and is consumed by tt-metal main. The explicit `0` should be added to those two call sites at merge time.

Riddy21:
RE: factory functions vs raw ClusterOptions

Pavle, good point. I checked tt-metal's usage and it currently constructs `ClusterOptions` directly everywhere (`tt_cluster.cpp`, `pcie_core_writer.cpp`) — nobody calls the factory methods yet.

The original motivation was to avoid cross-DSO STL ABI issues when passing `ClusterOptions` (which contains `std::string`, `std::optional`, `std::filesystem::path`) by value across `.so` boundaries. In practice, tt-metal statically links `libtt-umd` so the DSO boundary concern doesn't apply today.

I think the factories still provide a marginal convenience benefit (less boilerplate for common patterns), but I agree with you that they add surface area without a clear current need. The `explicit Cluster(ClusterOptions)` constructor + the default `Cluster()` constructor already cover the use cases cleanly.

**Action taken:** Keeping the factories for now since they're already written and don't add maintenance burden, but I've added a comment in the header clarifying they're optional convenience wrappers. If you feel strongly they should go, happy to remove them in a follow-up — just let me know.

device/api/umd/device/cluster.hpp
```diff
@@ -142,7 +142,29 @@ class Cluster {
...
+
+    // Explicit to prevent implicit conversion from ClusterOptions at call sites.
+    // The default Cluster() delegates to this constructor with a default-constructed
+    // ClusterOptions so that callers without options don't need to name the type.
+    explicit Cluster(ClusterOptions options);
+
+    // Optional convenience factories that build ClusterOptions internally.
+    // Callers may also construct ClusterOptions directly and pass it to the
+    // explicit Cluster(ClusterOptions) constructor — these are equivalent.
+    static std::unique_ptr<Cluster> create_silicon_cluster(
```
pjanevskiTT: https://github.com/tenstorrent/tt-umd/pull/2694#discussion_r3312028718

device/api/umd/device/cluster.hpp
```diff
@@ -84,7 +84,7 @@ struct ClusterOptions {
      * If not provided, the value is determined automatically by determining the max
      * amount of chips connected through one PCIe interface in the ClusterDescriptor.
      */
-    std::optional<uint32_t> num_host_mem_ch_per_mmio_device = 0;
+    std::optional<uint32_t> num_host_mem_ch_per_mmio_device = std::nullopt;
```
pjanevskiTT, line 87: https://github.com/tenstorrent/tt-umd/pull/2694#discussion_r3311653592

pjanevskiTT (commented):
I think we don't need these factories in general, do you see any benefit from it? I left 2 comments from previous PR so we can use those for discussion

device/api/umd/device/cluster.hpp
```diff
@@ -84,7 +84,7 @@ struct ClusterOptions {
      * If not provided, the value is determined automatically by determining the max
      * amount of chips connected through one PCIe interface in the ClusterDescriptor.
      */
-    std::optional<uint32_t> num_host_mem_ch_per_mmio_device = 0;
```
pjanevskiTT, line 87: ok I haven't seen the general comments for factories. I would like to get input from other UMD team members as well on this. @nbuncicTT @broskoTT @aleksamarkovicTT. https://github.com/tenstorrent/tt-umd/pull/2731#issuecomment-4556299396

device/api/umd/device/cluster.hpp
pjanevskiTT: For this, I would just like to update all the call sites in UMD where we don't call into factories (if we keep them) to use num_host_... = 0 or use the factory

device/api/umd/device/cluster.hpp
aleksamarkovicTT: I left it to be 0 because updating all call sites in UMD was a lot more work. I don't have a preference on the default value. @broskoTT preferred nullopt (automatic no. of channels) as a default if I remember.

device/api/umd/device/cluster.hpp
nbuncicTT: As @pjanevskiTT said, there is no apparent need to have multiple factories for populating a data structure - can you answer @Riddy21 why are you adding these helpers?

device/api/umd/device/cluster.hpp
```diff
@@ -142,7 +142,29 @@ class Cluster {
...
+    // Optional convenience factories that build ClusterOptions internally.
+    // Callers may also construct ClusterOptions directly and pass it to the
+    // explicit Cluster(ClusterOptions) constructor — these are equivalent.
+    static std::unique_ptr<Cluster> create_silicon_cluster(
+        std::optional<uint32_t> num_host_mem_ch_per_mmio_device = std::nullopt);
+    static std::unique_ptr<Cluster> create_single_chip_simulation_cluster(
+        uint32_t num_host_mem_ch_per_mmio_device, ChipId chip_id, const char* simulator_path);
+    static std::unique_ptr<Cluster> create_simulation_cluster_with_descriptor(
+        uint32_t num_host_mem_ch_per_mmio_device,
+        const char* simulator_path,
```
nbuncicTT: Why are you trying to achieve ABI stability by using `char*` instead of a `std::string` on the `Cluster` layer? I understand that today we have sprinkled out creation of different types of backends for Chips/TTDevice's, but where are you hitting a problem with Cluster - and where is the necessity for these factory functions arise?

device/api/umd/device/cluster.hpp
```diff
@@ -142,7 +142,29 @@ class Cluster {
      * - Perform this for each of the chips connected to the system.
      * @param options See documentation of ClusterOptions for explanation of specific arguments.
      */
-    Cluster(ClusterOptions options = {});
+    Cluster() : Cluster(ClusterOptions{}) {}
+
+    // Explicit to prevent implicit conversion from ClusterOptions at call sites.
+    // The default Cluster() delegates to this constructor with a default-constructed
+    // ClusterOptions so that callers without options don't need to name the type.
+    explicit Cluster(ClusterOptions options);
```
nbuncicTT, line 148: I'd prefer to have this explicit to omit implicit conversions - nice.

device/api/umd/device/cluster.hpp
nbuncicTT: In general I like this being `std::nullopt` if the type is `std::optional<unit32_t>` - because if we don't have a valid value that needs to be populated and 0 is a valid value, this needs to happen explicitly.

nbuncicTT (commented):
UMD is going in a direction where the `Cluster`/`ClusterOptions`, `Chip` classes won't change anymore because of silicon/emulation/simulation - they will be independent of the backend (silicon/emulation/simulation). And the backend behavior will be defined by `TTDevice`.

That being said, I think this PR is just adding code for the sake of convenience for filling a data structure.

I think the explicit constructor and the `std::nullopt` default values are an improvement, but this needs to be verified that it doesn't affect clients (the subtle behavior change).

device/api/umd/device/cluster.hpp
Riddy21: Agreed — dropped all five `Cluster::create_*` factories in the latest push (`b50cc36`).

Audited tt-metal: **zero** call sites for these APIs; multichip/TTSim still uses inline `ClusterOptions` in `tt_cluster.cpp::open_driver`. No ABI benefit for our stack since tt-metal statically links libtt-umd. Callers should keep using `ClusterOptions` + `explicit Cluster(ClusterOptions)`.

device/api/umd/device/cluster.hpp
Riddy21: Keeping `std::nullopt` as the default. I'll audit UMD internal `ClusterOptions{}` sites and set `num_host_mem_ch_per_mmio_device = 0` explicitly wherever the old implicit-zero behavior is still required.

tt-metal follow-up (separate PR): explicit `= 0` on MOCK and SWEMULE in `tt_cluster.cpp::open_driver` before this lands on UMD main. Silicon and SIMULATION paths already set the field explicitly.

device/api/umd/device/cluster.hpp
Riddy21: Update after team feedback: factories are removed in `b50cc36`. This PR now only has the explicit ctor, `nullopt` default, post-`construct_cluster` `options_` move, and `test_cluster_options.cpp` for the default semantics.

Happy to keep it that way unless we hear a concrete caller that needs a named helper.

Riddy21: Staying with `nullopt` as the default (auto-detect when unset). Call sites that need zero host-memory channels will set `num_host_mem_ch_per_mmio_device = 0` explicitly — tests in `test_cluster_options.cpp` lock in that `0` and `nullopt` stay distinguishable.

device/api/umd/device/cluster.hpp
Riddy21: Fair point — there was no live need: tt-metal never called these APIs, and the `const char*` ABI argument doesn't apply while tt-metal statically links UMD.

Dropped the factories; sim/mock/emule paths in tt-metal stay as inline `ClusterOptions`, which matches the Cluster/TTDevice direction you described.

device/api/umd/device/cluster.hpp
Riddy21: No concrete cross-DSO problem on `Cluster` today — tt-metal links UMD statically and constructs `ClusterOptions` in-process. The `const char*` paths were speculative; removed with the factories.

If we need a stable boundary later, it belongs at the `TTDevice`/backend layer, not on `Cluster`.

device/api/umd/device/cluster.hpp
Riddy21: Keeping this — prevents accidental implicit `ClusterOptions` → `Cluster` conversions as we add more sim/mock option shapes.

device/api/umd/device/cluster.hpp
Riddy21: Agreed — `0` is a valid channel count, so "unspecified" should be `nullopt`. `test_cluster_options.cpp` checks that `ClusterOptions{}` defaults to `nullopt` and that explicit `.num_host_mem_ch_per_mmio_device = 0` remains distinct.

Riddy21 (commented):
@pjanevskiTT @nbuncicTT — Re your review summaries: agreed on dropping the factories; removed in `b50cc36`. This PR now keeps only the explicit ctor, `nullopt` default, `options_` after auto-detect, and `test_cluster_options.cpp`.

@nbuncicTT — Client check: audited tt-metal (no factory callers). MOCK/SWEMULE in `tt_cluster.cpp` need explicit `num_host_mem_ch_per_mmio_device = 0` in a follow-up before UMD main picks this up; silicon/sim paths already set the field.

device/api/umd/device/cluster.hpp
Riddy21: This is done. Audited all UMD call sites in the current branch — every `ClusterOptions{}` construction that previously relied on the implicit `= 0` default now explicitly sets `num_host_mem_ch_per_mmio_device` to the appropriate value (0, 1, or a helper like `get_num_host_ch_for_test()`). The one site that intentionally wants auto-detection (`test_sysmem_manager.cpp`) sets `std::nullopt` explicitly. No UMD call site silently relies on the new `nullopt` default for what used to be zero-channel behaviour.

pjanevskiTT (approved):

tests/api/test_cluster_options.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include <gtest/gtest.h>
+
+#include <optional>
+
+#include "umd/device/cluster.hpp"
+#include "umd/device/types/cluster_types.hpp"
+
+using namespace tt::umd;
+
+TEST(ClusterOptions, DefaultHasNulloptHostMemChannels) {
```
nbuncicTT: For my taste, these tests are too granular, you could put all of them into one test, and verify the default behavior, and for the explicit construction compilation - I would have a different test for that. The changes would reduce the LOC in general, and from my point of view, enhance readability.

Something like this would be the test:
```cpp
#include <type_traits>

TEST(ClusterTest, ConstructorIsExplicit) {
    // 1. Verify that explicit construction works
    EXPECT_TRUE((std::is_constructible_v<Cluster, ClusterOptions>))
        << "Cluster should be explicitly constructible from ClusterOptions.";

    // 2. Verify that implicit conversion is blocked
    EXPECT_FALSE((std::is_convertible_v<ClusterOptions, Cluster>))
        << "API Violation: Cluster(ClusterOptions) must be explicit to prevent implicit conversion!";
}
```

nbuncicTT (approved):
Thanks for the contribution!

tests/api/test_cluster_options.cpp
Riddy21: Done in `bc32a9b`. Consolidated the granular tests into two:

- **`ClusterOptions.Defaults`** — single test covering the default-behavior checks (`num_host_mem_ch_per_mmio_device == std::nullopt`, `chip_type == SILICON`, empty `target_devices`, plus that an explicit `0` is preserved as distinct from `nullopt`).
- **`ClusterTest.ConstructorIsExplicit`** — a separate compile-time test using type traits, as you suggested:
```cpp
EXPECT_TRUE((std::is_constructible_v<Cluster, ClusterOptions>));
EXPECT_FALSE((std::is_convertible_v<ClusterOptions, Cluster>));
```
This also replaces the old runtime `ExplicitConstructorCompiles` test (which built a `Cluster` and swallowed exceptions) — the trait check actually verifies *explicitness*, not just that construction compiles. Net −21/+18 LOC.

Built `api_tests` and ran both — 2 passed.

PR #2732, opened 2026-05-27
Allowing bytes, memoryview and bytearray as write arguments in nanobind

tt-vjovanovic:
Add a `PyBufferView` class to facilitate the use of buffer protocol objects in device methods, allowing for more flexible data handling without unnecessary copies.


nanobind/py_api_tt_device.cpp
```diff
@@ -575,21 +620,22 @@ void bind_tt_device(nb::module_ &m) {
...
                 -> void {
                 if (noc_id != 0) {
                     UMD_THROW(error::RuntimeError, "noc_id must be 0.");
                 }
+                PyBufferView buffer(data);
                 tt_xy_pair core = {core_x, core_y};
-                const char *data_ptr = data.c_str();
-                size_t data_size = data.size();
-                self.write_to_device(data_ptr, core, addr, data_size);
+                self.write_to_device(buffer.data(), core, addr, buffer.size());
```
pjanevskiTT: The sibling `noc_write` overload without `noc_id` (around line 298) releases the GIL across `write_to_device`, but this overload still holds it. The asymmetry is pre-existing, but the PR is already rewriting this lambda for buffer-protocol support — it's the natural place to also wrap the call:

```cpp
{
    nb::gil_scoped_release release;
    self.write_to_device(buffer.data(), core, addr, buffer.size());
}
```

Holding the GIL across a blocking device write serializes all Python threads behind every NOC transfer, which undercuts the "more flexible data handling" goal of the PR.

nanobind/py_api_tt_device.cpp
```diff
@@ -609,18 +655,19 @@ void bind_tt_device(nb::module_ &m) {
...
         .def(
             "noc_broadcast",
-            [](TTDevice &self, uint32_t noc_id, uint64_t addr, const nb::bytes &data) -> void {
+            [](TTDevice &self, uint32_t noc_id, uint64_t addr, nb::handle data) -> void {
                 if (noc_id != 0) {
                     UMD_THROW(error::RuntimeError, "noc_id must be 0.");
                 }
-                std::vector<uint8_t> buffer(data.c_str(), data.c_str() + data.size());
+                PyBufferView buffer(data);
                 self.noc_multicast_write(buffer.data(), buffer.size(), addr);
```
pjanevskiTT: Same as the `noc_id` overload of `noc_write` above: the sibling `noc_broadcast` without `noc_id` (lines 313–316) now releases the GIL around `noc_multicast_write`, but this overload doesn't. Wrap `self.noc_multicast_write(...)` in a `nb::gil_scoped_release` block to match.

nanobind/py_api_tt_device.cpp
```diff
@@ -40,6 +40,37 @@ namespace nb = nanobind;
...
+            throw nb::python_error();
+        }
+    }
+
+    ~PyBufferView() { PyBuffer_Release(&buffer_); }
+
+    PyBufferView(const PyBufferView &) = delete;
+    PyBufferView &operator=(const PyBufferView &) = delete;
+
+    void *data() const { return buffer_.buf; }
```
pjanevskiTT, line 75: Returning non-const `void*` is a footgun. When the wrapped object is an immutable `bytes`, `buffer_.readonly == 1` and writing through this pointer would corrupt the Python object (UB). All current call sites only read, so this is latent — but the API surface invites a future bug.

Either change the signature to `const void *data() const` and `const_cast` at the few call sites whose downstream UMD function takes `void*` but doesn't mutate (making the intent visible at each call), or assert `!buffer_.readonly` somewhere central if you ever expect to write.

nanobind/py_api_tt_device.cpp
```diff
@@ -40,6 +40,37 @@ namespace nb = nanobind;
...
+// supports the buffer protocol — bytes, bytearray, memoryview, numpy arrays —
+// instead of forcing callers to materialize an extra bytes copy first.
+// Acquire/release must happen with the GIL held; callers are expected to keep
+// this object alive across any nb::gil_scoped_release block that uses .data().
+//
+// The pointer is exposed as void* to match the underlying Py_buffer::buf field
+// and to pass cleanly into legacy UMD APIs that take void* (and don't actually
+// mutate the data). PyBUF_SIMPLE alone is read-only at the protocol level —
+// callers must not actually write through this pointer.
+class PyBufferView {
```
pjanevskiTT, line 62: The PR broadens the input contract from `nb::bytes`/`nb::bytearray` to "any buffer-protocol object" and the `nb::sig` strings now advertise `bytes | bytearray | memoryview`, but `nanobind/tests/test_py_tt_device.py` and `test_py_spi_tt_device.py` still only exercise the `bytes` path. Buffer-protocol lifetime/contiguity bugs are notoriously silent under trivial tests. Suggest adding at least:

- a `memoryview` round-trip through `noc_write` / `noc_read`,
- a `bytearray` write,
- and a sliced/non-contiguous memoryview (e.g. `memoryview(b"abcdef")[::2]`) — this should raise `BufferError` rather than silently corrupt, because `PyBUF_SIMPLE` rejects non-contiguous views. That test pins down the contract.

nanobind/tests/test_py_tt_device.py
```diff
@@ -62,6 +62,36 @@ def test_low_level_tt_device(self):
             )
             dev.noc_write(tensix_core.x, tensix_core.y, 0x200, original_data)  # Restore
 
+            # Test noc_write with buffer-protocol inputs other than bytes
+            # (bytearray and memoryview) and verify the read-back matches.
+            try:
+                bytearray_data = bytearray((b ^ 0x5A) for b in original_data)
+                dev.noc_write(tensix_core.x, tensix_core.y, 0x200, bytearray_data)
+                self.assertEqual(
+                    dev.noc_read(tensix_core.x, tensix_core.y, 0x200, 16),
```
pjanevskiTT: nit: can we replace 16 with len(bytearray_data) or something similar?

pjanevskiTT (approved):

PR #2733, opened 2026-05-28
pcie: fix DMA buffer allocation on AArch64 with 64K base pages

lu-zero:
## Summary

On AArch64 systems with 64K base pages (e.g. Blackhole development systems), `sysconf(_SC_PAGESIZE)` returns 65536. The PCIe DMA buffer completion page was hardcoded as `+0x1000` (4 KiB), making the total allocation size non-page-aligned (e.g. 512K + 4K = 528384, which is not a multiple of 65536). `map_for_dma()` enforces page-alignment and threw an exception, which was silently caught — causing the DMA buffer to **never allocate** on AArch64, and every single DMA transfer to silently fall back to the TLB/memcpy path.

- Replace all three hardcoded `0x1000` completion-page offsets in `try_allocate_pcie_dma_buffer_iommu`, `try_allocate_pcie_dma_buffer_no_iommu`, and the destructor `munmap` with `sysconf(_SC_PAGESIZE)`
- No behaviour change on x86 (4K pages): `sysconf(_SC_PAGESIZE)` returns 4096 = `0x1000`

## Verification

Confirmed on AArch64 Blackhole with IOMMU enabled:
- Before: "DMA buffer was not allocated for PCI device 0" warning on every transfer, ~70 MB/s (TLB fallback)
- After: DMA engine fires correctly, peaks at ~7.5 GB/s H2D on Tensix

## Test plan

- [x] Run `MicrobenchmarkPCIeDMA.TensixSweepSizes` on AArch64 — no "DMA buffer was not allocated" warnings, throughput matches DMA engine bandwidth
- [ ] Run `MicrobenchmarkPCIeDMA.TensixSweepSizes` on x86 — no regression (`0x1000 == sysconf(_SC_PAGESIZE)` on 4K-page systems)

### DIFF
--- a/tests/microbenchmark/benchmarks/pcie_dma/test_pcie_dma.cpp
+++ b/tests/microbenchmark/benchmarks/pcie_dma/test_pcie_dma.cpp
@@ -208,9 +208,6 @@ TEST(MicrobenchmarkPCIeDMA, EthernetSweepSizes) {
 // to and from the device.
 TEST(MicrobenchmarkPCIeDMA, DRAMZeroCopy) {
     std::vector<int> pci_device_ids = PCIDevice::enumerate_devices();
-    if (pci_device_ids.empty()) {
-        GTEST_SKIP() << "No Tenstorrent PCI devices found.";
-    }
     if (!PCIDevice(pci_device_ids.at(0)).is_iommu_enabled()) {
         GTEST_SKIP() << "Skipping test since IOMMU is not enabled on the system.";
     }
@@ -245,9 +242,6 @@ TEST(MicrobenchmarkPCIeDMA, DRAMZeroCopy) {
 // This test measures bandwidth of IO using PCIe DMA engine without overhead of copying data into DMA buffer.
 TEST(MicrobenchmarkPCIeDMA, TensixZeroCopy) {
     std::vector<int> pci_device_ids = PCIDevice::enumerate_devices();
-    if (pci_device_ids.empty()) {
-        GTEST_SKIP() << "No Tenstorrent PCI devices found.";
-    }
     if (!PCIDevice(pci_device_ids.at(0)).is_iommu_enabled()) {
         GTEST_SKIP() << "Skipping test since IOMMU is not enabled on the system.";
     }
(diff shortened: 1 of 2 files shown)


lu-zero:
## AArch64 test results (Blackhole, SMMU enabled, 64K base pages)

All H2D DMA benchmarks now run correctly on AArch64 Blackhole:

| Test | Transfer size | Throughput |
|---|---|---|
| DMA write (bounce buffer, Tensix) | 256 KiB | 7.25 GB/s |
| DMA write (bounce buffer, Tensix) | 512 KiB | 7.69 GB/s |
| DMA write (bounce buffer, Tensix) | 1 MiB | 7.74 GB/s |
| DMA write (bounce buffer, DRAM) | 1 MiB | 7.74 GB/s |
| DMA ZeroCopy write (DRAM) | 1 MiB | 36.57 GB/s |
| DMA ZeroCopy write (Tensix) | 1 MiB | 22.74 GB/s |
| DMA MapBuffer ZeroCopy (Tensix) | 1 MiB | 24.77 GB/s |

tests/microbenchmark/benchmarks/pcie_dma/test_pcie_dma.cpp
```diff
@@ -242,6 +245,9 @@ TEST(MicrobenchmarkPCIeDMA, DRAMZeroCopy) {
 // This test measures bandwidth of IO using PCIe DMA engine without overhead of copying data into DMA buffer.
 TEST(MicrobenchmarkPCIeDMA, TensixZeroCopy) {
     std::vector<int> pci_device_ids = PCIDevice::enumerate_devices();
+    if (pci_device_ids.empty()) {
```
pjanevskiTT: we removed these checks from the tests that need a card in the system since we are running it only on the hosts that have a TT card in CI? Can you remove this? I agree that this test will silently fail if you run it on a system without a card but I think we need to have more robust way of handling this than having this check in every test. Can you remove it from all benchmarks?

pjanevskiTT (approved):

tests/microbenchmark/benchmarks/pcie_dma/test_pcie_dma.cpp
lu-zero: copilot flagged that and I wasn't sure if it was intentional or not since looks like other slipped in meanwhile.

lu-zero:
Done — removed the `enumerate_devices().empty()` guard from `DRAMZeroCopy`, `TensixZeroCopy`, and `TensixMapBufferZeroCopy`. The IOMMU check stays since that's a genuine capability condition.

pjanevskiTT:
ok enabled auto merge, will follow if CI is green

Issue #2450, opened 2026-04-10
Fix ARM build issues and add CI support


broskoTT:
We need to be able to build our SW on ARM

broskoTT:
+ add manylinux arm support
+ potentially add riscv support

broskoTT:
Added CI ARM support including python wheels

PR #2765, opened 2026-06-01
ttsim ARM test

broskoTT:
### Issue
#2450 

### Description
(Add freeform description.)

### List of the changes

### Testing
(Comment on CI testing or Manual testing steps that were performed to prove this change does what you intend it to do.)

### API Changes
This PR has (no) API changes.


broskoTT:
This is actually hard to do...
Since we're viewing the build and run all tests job as running on a single docker variant, since if it works on one it probably works on others.
Now running ttsim would break that pattern trying to run on multiple dockers

PR #2767, opened 2026-06-01
Riscv build

broskoTT:
### Issue
#2450 

### Description
Edit: Chose option 2

This PR (currently draft) showcases two ways we can add RISCV builds:

## Option 1 - marked "temp1" in the PR
Running on x86 machine with all the tools needed for cross compiling. This approach is faster, but there's a lot of caveats with cross compiling, you have to ifelse on specific tools, you have to add cross compiler configuration cmake, cmake option. You also can't even try to load the built *.so because it is not built for the platform that it was compiled on, so a lot of options are skipped.

## Option 2 - marked "temp2" in the PR
Docker actually supports running cross platform. But it involves QEMU, so for this option we need a slightly different path when starting the job, as we can't just put the desired riscv container in "container" field, but have to manually setup QEMU first and then run docker container, hence the dedicated build-riscv-emulated job which is outside of standard matrix. It also has problems that some of the stuff is not available for riscv, so not sure if it can build all things, but it handles python stuff and nanobind stub at least. The big downside is due to emulation this is supposed to be much slower.
From the CI:
Option 1 - ~3min
Option 2 - ~12min
Mind that 12min is usually faster than completing silicon tests, so it might be fine.

## Conclusion
If Option 2's slowness is not a problem, the solution looks much cleaner. And will be very easy to port to a real riscv machine once we get our hands on that.

### List of the changes
- Add dockerfile for riscv image
- Add a step to run emulated riscv instruction set on x86 machine

### Testing
CI on this branch:
https://github.com/tenstorrent/tt-umd/actions/runs/29283695178/job/86930939760?pr=2767

### API Changes
This PR has no API changes.


.github/workflows/build-device.yml
```diff
@@ -312,6 +325,60 @@ jobs:
...
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build All
+        shell: bash
+        run: |
+          IMAGE_TAG="${{ needs.ensure-riscv-image.outputs.image-tag }}"
+          IMAGE="ghcr.io/${{ github.repository }}/tt-umd-ci-ubuntu-24.04-riscv:${IMAGE_TAG}"
```
pjanevskiTT: This job runs on the self-hosted x86 runner `tt-ubuntu-2204-large-stable`, but pulls from `ghcr.io` directly. The existing `build` job's convention (lines 140-146) is that self-hosted x86 jobs go through the Harbor pull-through mirror (`harbor.ci.tenstorrent.net/ghcr.io`) and only the GitHub-hosted arm runner pulls ghcr directly. Consider routing this through Harbor for consistency and pull-through caching, or confirm the direct pull is intentional (e.g. Harbor doesn't mirror the riscv64 manifest). If it stays on direct ghcr, the explicit `docker login` step is correct and necessary.

.github/workflows/build-device.yml
```diff
@@ -312,6 +325,60 @@ jobs:
...
+            bash -c '
+              set -euo pipefail
+              git config --global --add safe.directory "*"
+              cmake -B ${{ env.BUILD_OUTPUT_DIR }} -G Ninja \
+                -DTT_UMD_BUILD_TESTS=OFF \
+                -DTT_UMD_BUILD_SIMULATION=OFF \
+                -DTT_UMD_BUILD_PIP=OFF \
+                -DTT_UMD_BUILD_PYTHON=ON \
+                -DTT_UMD_BUILD_EXAMPLES=OFF \
+                -DTT_UMD_ENABLE_CLANG_TIDY=ON \
```
pjanevskiTT: clang-tidy runs here under QEMU emulation, which is dramatically slower and pushes this job toward the 60-min timeout. Its findings are architecture-independent and already covered by the native x86 builds. Consider `OFF` for the emulated build to cut wall-clock significantly, unless you specifically want to exercise this flag path on riscv64.

pjanevskiTT (approved):
lgtm, although I don't consider myself expert here, would like to get another pair of eyes on this before merging

.github/workflows/build-device.yml
```diff
@@ -84,11 +84,24 @@ jobs:
     with:
```
pjanevskiTT, line 102: if you didn't add it already, I would like to have a comment somewhere explaining which options we chose and why, just to be aware that there is a faster option for this if we ever care about it

.github/workflows/build-device.yml
aleksamarkovicTT: I agree, this is a developer-facing tool, no need to have it in artifact-producing CI.

broskoTT: removed

.github/workflows/build-device.yml
broskoTT: added

.github/workflows/build-device.yml
```diff
@@ -102,6 +102,15 @@ jobs:
...
       docker-version: manylinux-aarch64
 
+  ensure-riscv-image:
```
broskoTT, line 105: don't forget the PR gate

broskoTT: Verified it doesn't execute anymore on PR

Issue #2237, opened 2026-03-06
[FW release] Investigate FW release 19.7.0


aleksamarkovicTT:
> Here's the release candidate 19.7.0-rc1,
https://github.com/tenstorrent/tt-system-firmware/releases/tag/v19.7.0-rc1
https://github.com/tenstorrent/tt-zephyr-platforms/blob/main/doc/release/release-notes-19.7.md

> Here's the release candidate 19.7.0-rc2,
https://github.com/tenstorrent/tt-system-firmware/releases/tag/v19.7.0-rc2
https://github.com/tenstorrent/tt-zephyr-platforms/blob/main/doc/release/release-notes-19.7.md
Changes from -rc1:
Add release note on vf curve
Tune TDP throttler with new constants
Wake the GDDR PHY before attempting GDDR reset

>Here's the latest release candidate 19.7.0-rc3,
https://github.com/tenstorrent/tt-system-firmware/releases/tag/v19.7.0-rc3
https://github.com/tenstorrent/tt-zephyr-platforms/blob/main/doc/release/release-notes-19.7.md
Changes from -rc2:
Add release notes.
Report GDDR BIST status in telemetry

broskoTT:
Should we create the newer ones as well? And then close them all with a PR which updates the latest supported fw version?

PR #2768, opened 2026-06-02
Retrain DRAM cores on BIST failure

pjanevskiTT:
### Issue

#2199

### Description

As of FW 19.7.0 the GDDR `BIST` status is reported in the upper 16 bits of the `GDDR_STATUS` telemetry tag (`[31:16]`), while the GDDR training status stays in the lower 16 bits (`[15:0]`). A channel can train successfully yet fail `BIST`, in which case it should be retrained rather than reported as healthy. This change folds a `BIST` failure into the per-channel `DramTrainingStatus`, so the existing `TTDevice::wait_dram_channel_training` retrain loop issues a directed retrain on `BIST` failure as well, not just on a training failure.

### List of the changes

- Decode the `BIST` half (`[31:16]`) of `GDDR_STATUS` alongside the training half (`[15:0]`), which share the same per-channel 2-bit layout.
- Report a channel as `SUCCESS` only when both training and `BIST` pass, and as `FAIL` if either fails (which triggers the existing directed retrain).
- Gate `BIST` inspection on Blackhole (the only architecture with GDDR `BIST` status) and FW `>= 19.7.0`, so other architectures and older firmware (which leaves `[31:16]` zero) keep the current training-only behavior.

### Testing

CI

### API Changes

None

### DIFF
--- a/device/firmware/firmware_info_provider.cpp
+++ b/device/firmware/firmware_info_provider.cpp
@@ -523,10 +523,10 @@ std::vector<DramTrainingStatus> FirmwareInfoProvider::get_dram_training_status(u
     bool is_legacy_wormhole =
         tt_device->get_arch() == ARCH::WORMHOLE_B0 && firmware_version <= FirmwareBundleVersion(18, 3, 0);
 
-    // BIST status is reported in the upper 16 bits of GDDR_STATUS starting with FW 19.7.0. Older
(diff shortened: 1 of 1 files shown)


aleksamarkovicTT (approved):

aleksamarkovicTT:
Related #2237

Issue #2788, opened 2026-06-08
Expose ETH heartbeat and link status


nbuncicTT:
**Description**
The ethernet heartbeat and link status are currently being read from an L1 location via direct reads. Syseng has exposed this metric via firmware, hence reading it from there is a needed telemetry feature. 

**Scope**
- Expose ETH heartbeat and link status from the WH/BH available versions (19.9+) via UMD
- Add tests to verify this behavior
- Have backwards compatibility for ETH retrain which was never supported for BH and is deprecated for WH after 19.9 FW version

**Timeline**
/

**Is your feature request related to a problem? Please describe.**
tt-telemetry and other tools need to read via L1 the value which is already collected by the on-board firmware itself.

**Definition of Done**
- Exposed APIs which return the expected value for the expected FW version
- Tests that verify this behavior

PR #2789, opened 2026-06-08
Expose ETH hearbeat and link status

nbuncicTT:
### Issue
#2788 

### Description
This pull request adds support for reporting per-link Ethernet link and heartbeat status, while deprecating retrain status for newer FW versions.

### List of the changes
* Added a new method `get_eth_link_status()` to `FirmwareInfoProvider` that returns the per-link Ethernet link status as a vector of bools, available on firmware 19.9+ for both Wormhole and Blackhole.
* Updated the firmware feature mapping logic to distinguish between heartbeat, retrain, and link status, and to map them appropriately based on firmware version and architecture. This includes new static feature map constructors for Wormhole and Blackhole 19.9+.
* Refactored the handling of Ethernet status features, replacing the old `ETH_LIVE_STATUS` feature with three distinct features: `ETH_HEARTBEAT_STATUS`, `ETH_RETRAIN_STATUS`, and `ETH_LINK_STATUS`, and updated their usage throughout the codebase.

### Testing

* Added and updated unit tests to verify the correct availability and behavior of Ethernet heartbeat, retrain, and link status methods for different architectures and firmware versions, including a new test for `get_eth_link_status()`.

These changes ensure that Ethernet link status is accurately reported and accessible through the API for supported firmware versions, while maintaining backward compatibility and test coverage.

Manual

### API Changes
Added one API (get_eth_link_status).

### DIFF
--- a/device/firmware/firmware_info_provider.cpp
+++ b/device/firmware/firmware_info_provider.cpp
@@ -131,7 +131,7 @@ FirmwareInfoProvider::FirmwareInfoProvider(TTDevice* tt_device) :
         {FirmwareFeature::BOARD_ID_HIGH, {WormholeTag::BOARD_ID_HIGH, LinearTransform{}}},
         {FirmwareFeature::BOARD_ID_LOW, {WormholeTag::BOARD_ID_LOW, LinearTransform{}}},
         {FirmwareFeature::ASIC_LOCATION, {FixedValue{0}, LinearTransform{}}},
-        {FirmwareFeature::ASIC_TEMPERATURE,  {WormholeTag::ASIC_TEMPERATURE,  LinearTransform{0, 0xFFFF,     1.0 / 16.0,    0.0}}},
+        {FirmwareFeature::ASIC_TEMPERATURE,  {WormholeTag::ASIC_TEMPERATURE,  LinearTransform{0, 0xFFFF, 1.0 / 16.0, 0.0}}},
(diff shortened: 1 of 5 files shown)


broskoTT (approved):

device/firmware/firmware_info_provider.cpp
```diff
@@ -625,18 +651,26 @@ std::optional<uint32_t> FirmwareInfoProvider::get_therm_trip_count() const {
...
-    if (!is_feature_available(FirmwareFeature::ETH_LIVE_STATUS)) {
+    if (!is_feature_available(FirmwareFeature::ETH_HEARTBEAT_STATUS)) {
         return std::nullopt;
     }
-    uint32_t data = read_raw_telemetry(firmware_feature_map.at(FirmwareFeature::ETH_LIVE_STATUS).key);
+    uint32_t data = read_raw_telemetry(firmware_feature_map.at(FirmwareFeature::ETH_HEARTBEAT_STATUS).key);
     return parse_eth_status_bitmask(static_cast<uint16_t>(data & 0xFFFF));
 }
 
+std::optional<std::vector<bool>> FirmwareInfoProvider::get_eth_link_status() const {
```
aleksamarkovicTT, line 661: This could be further improved by returning CoreCoords, either with an outer-class utility or adding a SocDescriptor reference to FirmwareInfoProvider. `vector<bool>` is certainly not the best interface.

nbuncicTT: I agree, but for now I would omit this - because we don't have SocDescriptor dependency on FirmwareInfoProvider. 

But I welcome this as an API change in the spec (because interfaces are decoupled, but are concrete implementations can depend on returning CoreCoord).

Let's brainstorm, but for now, I won't address it. 

Thanks for the suggestion!

Issue #2782, opened 2026-06-05
Address truncated in Arc Telemetry Init


micahwhiteTT:
**Describe the bug**
[tag_table_address](https://github.com/tenstorrent/tt-umd/blob/e85ef9571b972873f2f1f46d4d6175d8061ac373/device/arc/arc_telemetry_reader.cpp#L55) is uint32, should be uint64.

PR #2790, opened 2026-06-08
Fix truncation of telemetry_table_addr to uint32_t

mcraigheadTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/2782

### Description
Address math was truncating a uint64_t down to a uint32_t.

### List of the changes
Change local variable to uint64_t so value doesn't get truncated.

### Testing
@micahwhiteTT has tested this locally

### API Changes
This PR has (no) API changes.


aleksamarkovicTT (approved):
Thank you. I'm considering solving all of these with `-Wconversion`.

PR #2791, opened 2026-06-09
SWEmuleChip: Skip RISC-reset Broadcast + Fix Duplicate DRAM Backing

arminaleTT:
### Issue
_None — supports the tt-emule software-emulation backend (`ChipType::SWEMULE`)._

### Description
Two `SWEmuleChip` fixes for tt-emule's software-emulation backend, both gated on
`ChipType::SWEMULE` (no-ops for silicon/simulation).

### List of the changes
- **Skip the RISC-reset register broadcast for SWEMULE chips.** #2552 removed the
  per-chip reset dispatch (which `SWEmuleChip` overrode as a no-op) and made
  `assert/deassert_risc_reset` unconditionally broadcast to the TENSIX soft-reset
  register. Emule cores have no register backing, so that write lands out of L1 and
  aborts device bring-up. Early-return for SWEMULE in both reset entry points.
- **One backing per DRAM channel.** A DRAM channel is fronted by several NOC endpoint
  coords that all address the same physical bank, so `get_core` now aliases them to a
  single `Core`. DRAM cores are also classified in both NOC0 and TRANSLATED coords (on
  Blackhole `TRANSLATED != NOC0` for DRAM). Without this, a `noc=1` access reads a
  different mmap than the `noc=0`/host write — needed for faithful per-kernel
  `noc_index` in the emule runtime.

### Testing
tt-emule regressions (slow dispatch): **WH C++ 39/0, BH C++ 19/0**.

### API Changes
This PR has no API changes.

### Dependency chain
Merge bottom-up: **this PR** → tenstorrent/tt-metal#46492 (bumps the umd pin to this branch) → tenstorrent/tt-emule#120 (pins that tt-metal).

### DIFF
--- a/device/chip/sw_emule_chip.cpp
+++ b/device/chip/sw_emule_chip.cpp
@@ -91,8 +91,8 @@ tt_emule::Core* SWEmuleChip::get_core(tt_xy_pair core_xy) {
             return chit->second;  // share the channel's existing backing
         }
         tt_emule::CoreCoord dram_coord{core_xy.x, core_xy.y};
-        auto dram_core =
-            std::make_unique<tt_emule::Core>(dram_coord, tt_emule::CoreRole::DRAM, static_cast<size_t>(dram_bank_size_));
+        auto dram_core = std::make_unique<tt_emule::Core>(
+            dram_coord, tt_emule::CoreRole::DRAM, static_cast<size_t>(dram_bank_size_));
         tt_emule::Core* raw_ptr = dram_core.get();
         dram_channel_core_[channel] = raw_ptr;
         cores_[core_xy] = std::move(dram_core);  // first coord of the channel owns it
--- a/device/chip/sw_emule_chip.cpp
+++ b/device/chip/sw_emule_chip.cpp
@@ -39,11 +39,21 @@ SWEmuleChip::SWEmuleChip(const SocDescriptor& soc_descriptor) :
     // misses for translated coords on Blackhole (where TRANSLATED != NOC0 for
     // DRAM cores) and routes large DRAM writes into a 1.5 MB worker slot —
     // OOB-aborts as soon as the offset exceeds the worker L1 size.
+    // Register each DRAM subchannel core in BOTH coord systems it can reach emule
+    // in: NOC0 (get_dram_cores default; the runner's NOC0 core_map registration and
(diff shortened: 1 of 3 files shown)


device/cluster.cpp
```diff
@@ -589,6 +589,12 @@ std::set<ChipId> Cluster::get_target_mmio_device_ids() { return local_chip_ids_;
 std::set<ChipId> Cluster::get_target_remote_device_ids() { return remote_chip_ids_; }
 
 void Cluster::assert_risc_reset() {
+    // SW-emule cores have no register backing, so the soft-reset register broadcast would
```
broskoTT, line 592: You could also guard reads/writes in SWEmuleChip to ignore if it is in register space? Then you wouldn't need this early return

device/chip/sw_emule_chip.cpp
```diff
@@ -68,13 +66,28 @@ tt_emule::Core* SWEmuleChip::get_core(tt_xy_pair core_xy) {
         return it->second.get();
     }
 
-    // Lazy-create with appropriate role and size.
+    // One backing per DRAM channel: a channel's several NOC endpoint coords all address
+    // the same physical bank, so alias them to one Core (individual mmap, not pooled) —
+    // else a noc=1 read sees a different mmap than the noc=0/host write.
+    if (is_dram_core(core_xy)) {
+        uint32_t channel = dram_core_to_channel_.at(core_xy);
```
broskoTT: If your channels are standard channels/subchannels as used from tt_metal, you should be able to just use LOGICAL coordinates, using translate_to in soc_descriptor, then the x coord is channel and y is subchannel. Don't block the merge on this comment though

broskoTT (approved):

device/cluster.cpp
arminaleTT: I considered that, but we want to be able to detect accidental writes outside the L1 space by ops. If we no-op register space reads/writes, we lose a bit of that diagnostic capability

device/chip/sw_emule_chip.cpp
arminaleTT: Thanks for the pointer. I will look into this

arminaleTT: I have fixed this so we no longer track two sets of coordinates

Issue #2792, opened 2026-06-09
Simulation server: owner/host creates the server at a fixed socket path with direct access


pjanevskiTT:
Part of #2677. Design: `docs/SIMULATION_SERVER_PROCESS.md` §3, §5–§7, §9.

A UMD process becomes the **owner/host**: brings up the backend, owns a **single cluster's** state, and uses the **direct in-process backing** for its own access (no socket — max perf). It binds a UNIX socket at a **fixed path** so a client can connect later, but does not serve requests yet.

## Scope
- Communicator seam (§9): `Cluster → SimulationChip → TTDevice` reach the backend through a backing; wire today's in-process path (`tt_sim_communicator` / RTL communicator) as the **direct backing**. Behavior-preserving.
- open-or-attach (create side): connect-probe fixed path → if refused, create + become owner, then `bind()` + `listen()` (socket *is* the entry). Stub accept loop, no request handling.
- Stale reclaim / create race: refused → `unlink` → `bind`; `bind()` arbitrates simultaneous openers.
- Owner destroys the server on exit / explicit stop: tear down backend, destroy state, remove socket. No ownership transfer.

## Out of scope
Wire protocol & serving (#2793, #2794); client attach (#2795); discovery (§4, future); client accounting / linger.

## Acceptance
- Host created; existing sim workloads run via the direct path unchanged (sim tests pass).
- Socket connectable while host is up, gone after exit; stale socket reclaimed on restart.

pjanevskiTT:
**Follow-up (deferred from #2800): stale-reclaim probe treats any `connect()` failure as stale.**

In `SimulationSocket`, the stale-socket reclaim path relies on `is_live()`, which probes the bound path with a throwaway `connect()` and currently treats **any** non-success as "not live → reclaim (`unlink` + `bind`)". It does not distinguish:
- `ECONNREFUSED` / `ENOENT` — a genuine crashed-owner corpse, safe to reclaim, vs.
- `EACCES` and other errors — e.g. a *live* socket owned by another uid that we simply can't probe.

This is acceptable today **only** under the trusted-dir, single-simulator-per-chip assumption (predictable world-writable path, one kind of owner). It's now documented as an explicit assumption in `is_live()`, not fixed.

[...]

Issue #2677, opened 2026-05-19
Simulation server process


broskoTT:
Brainstormed with @tt-vjovanovic 
There is a need for running simulator more persistently as a server.
This would unlock easier usage by multiple processes (think about tt-triage scenario specifically).
This would also behave more like the silicon device.

We brainstormed how would that look like for rtl/emu. For ttsim we could also offer this, but it would have a perf penalty.

One way to surface these devices is by unix sockets in some folder, each file would represent a simulation device.
It is debatable who manages the lifetime of these objects.

pjanevskiTT:
closed by multiple PRs for now, will reopen if new features are needed

PR #2798, opened 2026-06-09
Simulation server: design doc + per-chip SimulationSocket (first slice)

pjanevskiTT:
### Issue
Part of #2677. First implementation slice of #2792.

### Description
Reworks the simulation-server design discussion around a **host-owns-the-sim** model: one UMD process (the host) owns the simulator and serves other UMD clients over a UNIX socket, while keeping a direct in-process fast path for its own access. The first iteration is scoped to a single server at a fixed location (multi-device discovery deferred).

Adds the first implementation slice: `SimulationSocket`, which surfaces a simulated device on disk as a per-chip UNIX socket ("the card"). The sim TTDevices create it at the end of bring-up and tear it down on teardown; the in-process fast path is unchanged.

### List of the changes
- docs: rework `docs/SIMULATION_SERVER_PROCESS.md` — drop clean-slate reset; host-owns-sim model; UNIX-socket exposure with a direct in-process fast path; scope the first iteration to a single server at a fixed location; mark discovery (§4) as future.
- Add `SimulationSocket`: binds a per-chip UNIX socket (`<tmp>/tt-umd-sim-<uid>-<chip_id>.sock`), reclaims a stale socket, refuses when a live owner holds the path, drains connections with a stub accept loop, and closes/unlinks on destruction.
- Wire `SimulationSocket` into `TTSimTTDevice` and `RtlSimulationTTDevice` (created at end of bring-up, destroyed on teardown).
- Tests: `test_simulation_socket.cpp` — 6 unit tests + 1 simulator-gated integration test.

### Testing
- `SimulationSocket` unit tests (6) + `SimulationSocketIntegration` (requires `TT_UMD_SIMULATOR`) pass against the TTSim `libttsim.so`.
- Full TTSim device-IO suite passes — confirms the in-process hot path is unchanged.
- RTL wiring is compile-verified (no RTL simulator available locally; exercised in CI).

### API Changes
- New public header `umd/device/simulation/simulation_socket.hpp` (class `SimulationSocket`).
- No changes to existing public APIs.

### DIFF
--- a/docs/SIMULATION_SERVER_PROCESS.md
+++ b/docs/SIMULATION_SERVER_PROCESS.md
@@ -5,7 +5,7 @@
 This document frames the problem and lays out the design decisions we need to make to run
 the simulator **persistently, as a server process** — decoupled from the lifetime of any
 single UMD client. It is meant to drive a design discussion, not to prescribe a final
-implementation. Each major section ends with the **decision(s)** the room needs to settle.
+implementation.
 
 ---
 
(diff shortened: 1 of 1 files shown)


pjanevskiTT:
## Hot-path performance check (TTSim)

Compared the TTSim tests on this branch (per-chip `SimulationSocket`) vs `main`, to confirm the owner's direct in-process fast path isn't slowed by the added socket. Release build, qsr `libttsim.so`, min of 3 runs.

| Workload | main | branch | delta | per-device |
|---|---|---|---|---|
| Full TTSim suite (14 tests × 100 = 1400 construct+IO cycles) | 2.99 s | 3.23 s | +0.24 s | **+0.17 ms** |
| Construct/destruct only (`CreateSimDevice` × 500) | 0.934 s | 1.010 s | +0.076 s | **+0.15 ms** |

**Conclusion: the I/O hot path is not slowed.** The per-cycle delta is essentially identical whether or not the cycle does heavy I/O (0.17 ms vs 0.15 ms — within noise), so the added cost is entirely the **per-device socket lifecycle** (bind + spawn accept thread + join + unlink at construction/teardown), not device reads/writes. Direct in-process I/O is untouched.

Net: **~0.15 ms one-time setup per device construction** (~8% of the already-tiny ~1.9 ms construct cost), dominated by the accept-thread create/join. Negligible for real workloads where a device is constructed once and lives for the whole run.

If we want to drive even that to zero, the accept thread could be started lazily (only when serving lands in a later issue), so the owner-only path pays just the cheap `bind`+`listen`.

pjanevskiTT:
Splitting this up: the implementation now lives in its own PR (#2800), and the design discussion stays on branch `pjanevski/sim-server-design-doc` (no PR). Closing this one.

PR #2799, opened 2026-06-09
Make HangDetector NOC register read injectable

pjanevskiTT:
### Issue

#1673

### Description

Structural prep extracted from #2629, addressing review feedback that HangDetector should stay unaware of anything below the protocol level (no TlbWindow, no locks).

HangDetector now reads its NOC liveness register through an injectable `NocRegReader`, defaulting to the current protocol-based read. A higher layer can override it via `set_noc_reg_reader()` with a reader that owns its own probe window and lock, keeping that knowledge out of HangDetector. The override consumer lands later with the per-op MMIO timeout work. Behavior is unchanged.

### List of the changes

- Add `NocRegReader` and a `set_noc_reg_reader()` override hook to `HangDetector`; default reader reads via `DeviceProtocol`.
- Route `BlackholeHangDetector` / `WormholeHangDetector` NOC reads through a new protected `read_noc_reg()`; drop their now-unused `device_protocol.hpp` include.

### Testing

CI

### API Changes

- New `tt::umd::NocRegReader` typedef and `HangDetector::set_noc_reg_reader()` method.

### DIFF
--- a/device/api/umd/device/tt_device/hang_detection/hang_detector.hpp
+++ b/device/api/umd/device/tt_device/hang_detection/hang_detector.hpp
@@ -20,14 +20,9 @@ class PcieInterface;
 enum class NocId : uint8_t;
 
 /**
- * Reads a single 32-bit NOC register at (core, addr) over the given NOC and returns its value.
- *
- * This is the injection seam for the NOC liveness probe. By default HangDetector reads through its
- * DeviceProtocol, which is all the stock hang check needs. A higher layer (e.g. TTDevice) can swap in
- * its own reader via set_noc_reg_reader() — for instance one that probes through a dedicated, separately
- * locked window so the check stays safe to call from inside a timed-out memcpy. Crucially, the window and
- * any lock live in the injected reader, NOT in HangDetector: HangDetector stays unaware of anything below
- * the protocol level (no TlbWindow, no locks).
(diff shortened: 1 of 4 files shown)


device/api/umd/device/tt_device/hang_detection/hang_detector.hpp
```diff
@@ -6,17 +6,31 @@
...
+#include "umd/device/types/xy_pair.hpp"
 
 namespace tt::umd {
 
 class DeviceProtocol;
 class PcieInterface;
 enum class NocId : uint8_t;
 
+/**
+ * Reads a single 32-bit NOC register at (core, addr) over the given NOC and returns its value.
```
nbuncicTT: nit: I think that this comment it to poetic, it should be more technical and concise.

nbuncicTT: It also implies "how" to use this lambda, which should be a thing that the user defines, so the memcpy, timeout, locks and other stuff are implementation details clients worry about - I'd refrain from adding this kind of comment.

nbuncicTT (approved):
Well done!

pjanevskiTT:
sems like merge queue is failing for the same reason as main, not related to this change, will force merge

PR #2800, opened 2026-06-09
Simulation: per-chip SimulationSocket

pjanevskiTT:
### Issue
First slice of #2792 (part of #2677).

### Description
Adds `SimulationSocket` — a per-chip UNIX domain socket that surfaces a simulated device on disk ("the card"). Standalone and independently tested; nothing consumes it yet — the discovery that drives the host-vs-client decision is the stacked PR.

- Binds a per-chip socket; reclaims a stale socket left by a dead owner.
- `try_create` is the non-throwing host-vs-client arbiter (bind/reclaim → host; a live owner → `nullptr`).
- Accept-and-drop (liveness only); request serving lands in a later PR.
- Closes and removes the socket file on destruction.

### List of the changes
- `simulation_socket.{hpp,cpp}` + `test_simulation_socket.cpp` + CMake.

### Testing
- CI

### API Changes
- None — `simulation_socket.hpp` is an internal header (not part of the public API file set).


device/simulation/simulation_socket.cpp
```diff
@@ -0,0 +1,201 @@
...
+    }
+    if (shutdown_pipe_[0] >= 0) {
+        ::close(shutdown_pipe_[0]);
+        shutdown_pipe_[0] = -1;
+    }
+    if (shutdown_pipe_[1] >= 0) {
+        ::close(shutdown_pipe_[1]);
+        shutdown_pipe_[1] = -1;
+    }
+    std::filesystem::remove(socket_path_);
```
pjanevskiTT: The destructor removes the socket file unconditionally, but in the `try_create` client path the object never bound — `bind_and_listen()` returned `false` because a *live* owner holds the path, `try_create` returns `nullptr`, and destroying that never-bound object then deletes **the live owner's** socket file. That defeats the host-vs-client contract this slice establishes: after a client gets `nullptr`, the host's socket is gone from disk. Suggest a `bound_` flag set at the end of a successful `bind_and_listen()` and gating the `remove()` (and ideally all teardown) on it. The throwing ctor is unaffected since a throw from the ctor body skips `~SimulationSocket`.

tests/baremetal/test_simulation_socket.cpp
```diff
@@ -0,0 +1,131 @@
...
+    std::filesystem::remove(path);
+}
+
+TEST(SimulationSocket, TryCreateReturnsNullWhenLiveHostExists) {
+    const std::filesystem::path path = unique_socket_path(5);
+    std::filesystem::remove(path);
+
+    auto host = SimulationSocket::try_create(path);
+    ASSERT_NE(host, nullptr);
+    EXPECT_EQ(SimulationSocket::try_create(path), nullptr);  // live host -> null, no throw
```
pjanevskiTT: This exercises the bug flagged in the destructor but doesn't catch it: after the second `try_create` returns `nullptr`, the destructor of the throwaway object has already removed `path`, so the live `host` is silently de-listed. Please add `EXPECT_TRUE(std::filesystem::exists(path));` and `EXPECT_TRUE(can_connect(path));` after this line — they should pass once the destructor is ownership-gated, and fail today.

device/simulation/simulation_socket.cpp
```diff
@@ -0,0 +1,201 @@
...
+        UMD_THROW(
+            error::RuntimeError, fmt::format("Failed to create simulation server socket: {}", std::strerror(errno)));
+    }
+
+    sockaddr_un addr = make_unix_addr(socket_path_);
+    if (::bind(listen_fd_, reinterpret_cast<sockaddr*>(&addr), sizeof(addr)) != 0) {
+        if (errno == EADDRINUSE) {
+            // Something already holds the path. A live owner means we must attach as a client
+            // instead (return false); a stale leftover (crashed owner) is reclaimed.
+            if (is_live(socket_path_)) {
```
pjanevskiTT: The reclaim sequence `is_live() → remove() → re-bind()` is a TOCTOU window: two contenders both seeing the stale file can both `remove()` and `bind()`, or a new legitimate owner binding in the gap can have its file removed. Likely tolerable for the single-simulator-per-chip assumption, but it's currently neither documented nor tested. Worth a brief "best-effort reclaim, assumes a single contender per path" comment so the stacked discovery PR inherits the constraint explicitly.

device/simulation/simulation_socket.cpp
```diff
@@ -0,0 +1,201 @@
...
+        }
+    }
+}
+
+std::filesystem::path SimulationSocket::default_socket_path(ChipId chip_id) {
+    std::filesystem::path dir = std::filesystem::temp_directory_path();
+    if (const char* env = std::getenv("TT_UMD_SIM_SOCKET_DIR")) {
+        dir = std::filesystem::path(env);
+    }
+    return dir / fmt::format("tt-umd-sim-{}-{}.sock", ::geteuid(), chip_id);
```
pjanevskiTT: The path is fully predictable in a world-writable temp dir. Putting the euid in the *name* doesn't isolate it — another uid can pre-`bind()` this path, making `is_live()` succeed so we silently attach to a foreign socket as a "client." `$TT_UMD_SIM_SOCKET_DIR` is also taken verbatim. Consider a per-uid `0700` subdirectory, or document that the socket dir is assumed trusted/per-user.

device/api/umd/device/simulation/simulation_socket.hpp
```diff
@@ -0,0 +1,69 @@
...
+//
+// Binding the socket is the act of surfacing the device; a client proves liveness
+// by connecting to it. On construction it reclaims a stale socket left by a dead
+// owner. Whether a live owner already holds the path is the host-vs-client arbiter:
+// the throwing constructor refuses in that case, while try_create() returns nullptr so
+// the caller (SimulationTopologyDiscovery) can attach as a client instead. On
+// destruction it closes the socket and removes the file.
+//
+// It does not yet handle client requests: connections are accepted and dropped. See
+// docs/SIMULATION_SERVER_PROCESS.md.
```
pjanevskiTT: `docs/SIMULATION_SERVER_PROCESS.md` doesn't exist in this PR. Either add it or drop the reference until the stacked PR lands.

device/simulation/simulation_socket.cpp
```diff
@@ -0,0 +1,201 @@
...
+    return addr;
+}
+
+// Returns true if a listener is currently reachable at path.
+bool is_live(const std::filesystem::path& path) {
+    int fd = ::socket(AF_UNIX, SOCK_STREAM, 0);
+    if (fd < 0) {
+        return false;
+    }
+    sockaddr_un addr = make_unix_addr(path);
```
pjanevskiTT: `is_live()` calls `make_unix_addr`, which `UMD_THROW`s on an over-long path. It's only reached after `bind_and_listen` already validated the same path, so it can't throw in practice — but a predicate-named helper that can throw is a latent footgun. Consider returning `false` on the over-long case, or documenting the precondition.

device/CMakeLists.txt
```diff
@@ -136,6 +136,9 @@ target_sources(
         tt_kmd_lib/tt_kmd_lib.c
         arc/wormhole_spi_tt_device.cpp
         arc/blackhole_spi_tt_device.cpp
+        # Pure POSIX (UNIX socket + filesystem); no simulation deps, so always built. This
```
nbuncicTT: In general, the PR looks good from a prototype point of view, but one thing though: we already use ASIO for Unix domain sockets in `warm_reset.cpp` (`WarmResetCommunication::Monitor` is essentially the same pattern: background thread, acceptor, async accept loop, shutdown via io_context::stop()). This PR reimplements all of that with raw POSIX (socket/bind/poll/pipe/close/errno), which is a fair amount of manual fd lifecycle and error handling that ASIO gives us for free.

I think you should port this to asio::local::stream_protocol. It should roughly halve the implementation, as the shutdown pipe goes away (io_context::stop()), the accept loop becomes async_accept, sockaddr_un filling becomes endpoint(path), and fd cleanup is RAII. The stale-reclaim logic stays the same, just with ASIO's bind/connect instead of raw syscalls.

Let's sync offline to see if this makes sense.

pjanevskiTT: makes sense, moved SimulationScoket to use ASIO

device/simulation/simulation_socket.cpp
```diff
@@ -0,0 +1,220 @@
...
+    // filesystem_error would call std::terminate. Cleanup is best-effort.
+    std::error_code ec;
+    std::filesystem::remove(socket_path_, ec);
+}
+
+std::filesystem::path SimulationSocket::default_socket_path(ChipId chip_id) {
+    std::filesystem::path dir = std::filesystem::temp_directory_path();
+    if (const char* env = std::getenv("TT_UMD_SIM_SOCKET_DIR")) {
+        dir = std::filesystem::path(env);
+    }
```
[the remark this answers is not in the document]
pjanevskiTT: Dropped the TT_UMD_SIM_SOCKET_DIR override entirely — default_socket_path() now always uses the system temp dir, so the relative-path footgun no longer applies.

device/simulation/simulation_socket.cpp
```diff
@@ -0,0 +1,220 @@
...
+        // is_live() -> remove() -> re-bind() sequence is a TOCTOU window, so two
+        // processes racing to reclaim the same stale socket (or a new owner binding in
+        // the gap) is undefined. The socket dir is assumed trusted (see
+        // default_socket_path), which is what makes this acceptable. Note a stale socket
+        // owned by another user may not be removable under a sticky dir (e.g. /tmp).
+        if (is_live(socket_path_)) {
+            return false;
+        }
+        std::filesystem::remove(socket_path_);
+        impl_->acceptor.bind(endpoint, ec);
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed: the reclaim path now refuses to unlink a non-socket object squatting the path and surfaces remove() errors instead of ignoring them, with a regression test (RefusesToReclaimNonSocketFile).

device/CMakeLists.txt
```diff
@@ -136,6 +136,9 @@ target_sources(
...
+        # lets its unit tests run in the no-card baremetal CI lane (TT_UMD_BUILD_SIMULATION=OFF).
+        simulation/simulation_socket.cpp
```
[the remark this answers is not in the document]
pjanevskiTT: Updated the PR description: this header is intentionally internal (not added to the public FILE_SET); no public API is introduced.

device/CMakeLists.txt
nbuncicTT: nit: The CMake comments still say "Pure POSIX" etc., but now that it uses ASIO those should be updated. The reasoning (no simulation deps, always built) still holds though.

tests/baremetal/test_simulation_socket.cpp
```diff
@@ -0,0 +1,149 @@
...
+    std::strncpy(addr.sun_path, path.c_str(), sizeof(addr.sun_path) - 1);
+    ASSERT_EQ(::bind(fd, reinterpret_cast<sockaddr*>(&addr), sizeof(addr)), 0);
+    ::close(fd);
+    ASSERT_TRUE(std::filesystem::exists(path));
+}
+
+}  // namespace
+
+TEST(SimulationSocket, ExposesConnectableSocket) {
+    const std::filesystem::path path = unique_socket_path(1);
```
nbuncicTT: Consider using a `TEST_F` fixture for the repeated setup/teardown pattern. Every test does `unique_socket_path(N)` + remove at the start and end. Something like:
```cpp
class SimulationSocketTest : public ::testing::Test {
protected:
    void SetUp() override {
        path_ = unique_socket_path();
        std::filesystem::remove(path_);
    }
    void TearDown() override {
        std::filesystem::remove(path_);
    }
    std::filesystem::path path_;
};
```
Also, the incrementing tag parameter doesn't serve a purpose because the path is removed after every test, these aren't multithreaded/multiprocess tests, and the PID already handles parallel test binaries. `unique_socket_path` can just use a fixed name.

tests/baremetal/test_simulation_socket.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+    SimulationSocket server(path);
+
+    EXPECT_TRUE(std::filesystem::exists(path));
+    EXPECT_TRUE(can_connect(path));
+
+    std::filesystem::remove(path);
+}
+
+TEST(SimulationSocket, RemovesSocketOnDestruction) {
```
nbuncicTT: With the new fixture, this test would become something like:
```cpp
TEST_F(SimulationSocketTest, RemovesSocketOnDestruction) {
    { SimulationSocket server(path_); }

    EXPECT_FALSE(std::filesystem::exists(path_));
    EXPECT_FALSE(can_connect(path_));
}
```

device/simulation/simulation_socket.hpp
```diff
@@ -0,0 +1,77 @@
...
+#pragma once
+
+#include <filesystem>
+#include <memory>
+
+#include "umd/device/types/cluster_descriptor_types.hpp"
+
+namespace tt::umd {
+
+// Owns the per-chip UNIX domain socket that exposes one simulation device on disk
```
nbuncicTT: nit: I'd simplify the comment to this:
```cpp
// Manages a per-chip UNIX domain socket that represents a simulated device on disk.
// One socket per simulated chip.
//
// The socket acts as a presence indicator: if a process can bind the path, it becomes
// the host; if a live host already exists, try_create() returns nullptr so the caller
// can connect as a client instead. Stale sockets left by crashed owners are
// automatically reclaimed.
```
Do you think it makes sense?

device/simulation/simulation_socket.cpp
```diff
@@ -0,0 +1,235 @@
...
+                "Simulation server socket path is too long ({} >= {}): {}",
+                path.string().size(),
+                sizeof(sockaddr_un::sun_path),
+                path.string()));
+    }
+    return stream_protocol::endpoint(path.string());
+}
+
+// Returns true if a listener is currently reachable at path.
+bool is_live(const std::filesystem::path& path) {
```
nbuncicTT: `is_live()` creates a throwaway `asio::io_context` just to do one synchronous connect. bind_and_listen() already has `impl_->io` available - consider making this a private method that reuses it, same pattern as `get_connected_listeners` in `warm_reset.cpp`.

device/simulation/simulation_socket.cpp
```diff
@@ -0,0 +1,235 @@
...
+struct SimulationSocket::Impl {
+    asio::io_context io;
+    stream_protocol::acceptor acceptor{io};
+    std::thread io_thread;
+};
+
+SimulationSocket::SimulationSocket(const std::filesystem::path& socket_path, DeferBind) :
+    socket_path_(socket_path), impl_(std::make_unique<Impl>()) {}
+
+SimulationSocket::SimulationSocket(const std::filesystem::path& socket_path) :
```
nbuncicTT, line 47: Let's simplify this by not having two pairs of constructors (one for throwing and one for not throwing).

Instead, let's have one private constructor that doesn't throw, and a separate static create function that creates the object and invokes `try_create()'.

If `try_create()` returns nullptr, the new create function throws.

The usage would change:
```cpp
// Before:
SimulationSocket server(path);  // throws if can't bind

// After:
auto server = SimulationSocket::create(path);  // throws if can't bind
```

and the public factory would look like:
```cpp
static std::unique_ptr<SimulationSocket> create(const std::filesystem::path& path) {
    auto s = try_create(path);
    if (!s) UMD_THROW(...);
    return s;
}
```

nbuncicTT (approved):

Issue #2785, opened 2026-06-07
viommu perf regression


broskoTT:
**Describe the bug**
It seems like we regressed viommu perf.
Seems like all our main CI runs now fail. This is due to main only having 6u glx with viommu as well as bh loudbox in viommu, which turn out to be the slowest ones.

By inspecting these two jobs:
https://github.com/tenstorrent/tt-umd/actions/runs/26879706923
https://github.com/tenstorrent/tt-umd/actions/runs/26942419407
Which correspond to this PR https://github.com/tenstorrent/tt-umd/pull/2731

Slowness can be seen in test runs:
llmbox: 7m32s -> 19m49s
n300: 6m6s -> 8m31s
n150: 3m56s -> 4m50s

**Stack trace**
Not applicable

**To reproduce**
can be seen on current main on CI

**Expected behavior**
perf reverted to previous state.

broskoTT:
And looking at the PR the cause is clear, the default for number of host memory channels  in ClusterOptions changed.

broskoTT:
Linked as the cause of a failed CI run: https://github.com/tenstorrent/tt-umd/actions/runs/26955471204

broskoTT:
Linked as the cause of a failed CI run: https://github.com/tenstorrent/tt-umd/actions/runs/26967848721

broskoTT:
Linked as the cause of a failed CI run: https://github.com/tenstorrent/tt-umd/actions/runs/26962113011

broskoTT:
Linked as the cause of a failed CI run: https://github.com/tenstorrent/tt-umd/actions/runs/26949512369

broskoTT:
Linked as the cause of a failed CI run: https://github.com/tenstorrent/tt-umd/actions/runs/26942419407

broskoTT:
Linked as the cause of a failed CI run: https://github.com/tenstorrent/tt-umd/actions/runs/27075652542

aleksamarkovicTT:
Cause: https://github.com/tenstorrent/tt-umd/pull/2731

broskoTT:
Linked as the cause of a failed CI run: https://github.com/tenstorrent/tt-umd/actions/runs/27074330588

broskoTT:
Linked as the cause of a failed CI run: https://github.com/tenstorrent/tt-umd/actions/runs/27071979418

broskoTT:
[...]

PR #2806, opened 2026-06-09
Use Cluster as pointer in tests

broskoTT:
### Issue
#2785 

### Description
As a follow up to the last fix to tests I did, stop using Cluster as reference in our tests.

### List of the changes
- Switch all cluster usages in our tests to pointers

### Testing
No testing, code builds.
No functional changes

### API Changes
This PR has no API changes.

### DIFF
--- a/tests/api/test_device_io.cpp
+++ b/tests/api/test_device_io.cpp
@@ -787,11 +787,11 @@ TEST(TestDeviceIO, SmnReadWriteRoundTrip) {
  * dma_read_from_device is used. On other architectures, the standard read_from_device
  * path is used instead.
  */
-void read_data_based_on_architecture(Cluster& cluster, CoreCoord core, void* mem_ptr, uint64_t address, size_t size) {
-    if (cluster.get_tt_device(0)->get_arch() == tt::ARCH::WORMHOLE_B0) {
-        cluster.dma_read_from_device(mem_ptr, size, 0, core, address);
+void read_data_based_on_architecture(Cluster* cluster, CoreCoord core, void* mem_ptr, uint64_t address, size_t size) {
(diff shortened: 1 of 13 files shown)


tests/api/test_device_io.cpp
```diff
@@ -787,11 +787,11 @@ TEST(TestDeviceIO, SmnReadWriteRoundTrip) {
  * dma_read_from_device is used. On other architectures, the standard read_from_device
  * path is used instead.
  */
-void read_data_based_on_architecture(Cluster& cluster, CoreCoord core, void* mem_ptr, uint64_t address, size_t size) {
-    if (cluster.get_tt_device(0)->get_arch() == tt::ARCH::WORMHOLE_B0) {
-        cluster.dma_read_from_device(mem_ptr, size, 0, core, address);
+void read_data_based_on_architecture(Cluster* cluster, CoreCoord core, void* mem_ptr, uint64_t address, size_t size) {
```
nbuncicTT, line 790: Why are we switching to pointers instead of references?

Usually, it's better to use a reference than a pointer because the object that's being pointed to is immutable (as opposed to the freedom of changing a non-owning pointer).

And if that's the case, why isn't the pointer `const`?

nbuncicTT (commented):
I'm okay with the test having to use `make_unique` (as it's the way were using this today) - but why do APIs now accept raw pointers? I'm not against the change, but let's just sync and see why is this change wanted.

tests/api/test_device_io.cpp
broskoTT: You're right. I'll close this PR, I don't want to spend more time on this.
Changing the other way around would be an XL pr

broskoTT: And this one is already heavily outdated

PR #2808, opened 2026-06-10
Support wh_x2 (N300) multichip ttsim

mcraigheadTT:
### Issue
Part of #2679. Tracked by #2832.

### Description
UMD's ttsim path assumed a `libttsim.so` exposes a single MMIO chip, so a multichip simulator like the wh_x2 (N300) only ever surfaced chip 0. This PR enumerates the full N300 topology — the MMIO gateway (chip 0) and the remote chip (chip 1) over simulated ethernet — so device IO runs across both.

Most changes follow from a *simulated* remote chip not being like a silicon one:

- **No ARC.** A silicon remote chip is brought up over ARC during topology discovery (which gives it its `SocDescriptor`/chip info and lets it service power/eth-queue/reset messages). The simulated one has none of that, so its `SocDescriptor` is supplied to the remote `TTDevice` at creation and ARC-only steps are skipped in simulation mode. Silicon remote-over-eth is unchanged.
- **Clock.** `advance_device_execution()` on a remote `TTDevice` forwards to its gateway device, so polling the remote chip still advances the simulator clock.
- **Topology.** The mock descriptor marks every chip MMIO with no eth links, so a remote chip would never appear. UMD now discovers an optional `cluster_descriptor.yaml` beside the `.so` (like `soc_descriptor.yaml`); without it, behaviour is unchanged.

### List of the changes
- Enumerate non-MMIO simulation chips as remote chips routed through the MMIO gateway.
- Provide the `SocDescriptor` to the remote sim `TTDevice` at creation; `RemoteChip::get_soc_descriptor()` delegates to the device.
- Skip ARC-only steps (power state, ethernet-queue enable, close, reset) for simulated remote chips.
- Forward `advance_device_execution()` from a remote `TTDevice` to its gateway device.
- Discover an optional `cluster_descriptor.yaml` from the simulator directory.
- Extract simulation remote-chip construction into a `Cluster` helper.
- CI: exercise the wh_x2 (N300) ttsim.

### Testing
CI

### API Changes
`TTDevice::create` gains an optional `SocDescriptor` (defaulted; silicon callers unaffected). `RemoteChip` gains `create_for_simulation`. New `SimulationChip::get_cluster_descriptor_path_from_simulator_path`. Needs careful UMD review to confirm no existing assumptions are broken.

### DIFF
--- a/.github/workflows/run-ttsim-tests.yml
+++ b/.github/workflows/run-ttsim-tests.yml
@@ -97,6 +97,14 @@ jobs:
           fileName: "libttsim_qsr.so"
           out-file-path: /tmp/ttsim-qsr
 
+      - name: Download ttsim binary (wh_x2)
+        uses: robinraju/release-downloader@28fc21f50d76778e7023361aa1f863e717d3d56f  # v1.13
+        with:
+          repository: tenstorrent/ttsim
+          tag: ${{ steps.get-ttsim-version.outputs.ttsim-version }}
+          fileName: "libttsim_wh_x2.so"
+          out-file-path: /tmp/ttsim-wh-x2
+
       - name: Build tt-metal
         run: |
           cd tt-metal
@@ -112,13 +120,16 @@ jobs:
       - name: Prepare sim paths
         run: |
           echo "Preparing sim paths"
-          mkdir -p sim_wh sim_bh sim_qsr
+          mkdir -p sim_wh sim_bh sim_qsr sim_wh_x2
           mv /tmp/ttsim-wh/libttsim_wh.so sim_wh/libttsim.so
           mv /tmp/ttsim-bh/libttsim_bh.so sim_bh/libttsim.so
           mv /tmp/ttsim-qsr/libttsim_qsr.so sim_qsr/libttsim.so
+          mv /tmp/ttsim-wh-x2/libttsim_wh_x2.so sim_wh_x2/libttsim.so
           cp $TT_METAL_HOME/tt_metal/soc_descriptors/wormhole_b0_80_arch.yaml sim_wh/soc_descriptor.yaml
           cp $TT_METAL_HOME/tt_metal/soc_descriptors/blackhole_140_arch.yaml sim_bh/soc_descriptor.yaml
           cp tt-metal/tt_metal/third_party/umd/tests/soc_descs/quasar_32_arch.yaml sim_qsr/soc_descriptor.yaml
+          # wh_x2 is two Wormhole chips; reuse the per-chip Wormhole soc descriptor.
+          cp $TT_METAL_HOME/tt_metal/soc_descriptors/wormhole_b0_80_arch.yaml sim_wh_x2/soc_descriptor.yaml
 
       - name: Build UMD api_tests with simulation support
         run: |
@@ -187,6 +198,25 @@ jobs:
         timeout-minutes: 1
         env:
           TT_UMD_SIMULATOR: ${{ github.workspace }}/sim_qsr/libttsim.so
+        run: |
+          tt-metal/tt_metal/third_party/umd/build/test/umd/api/api_tests --gtest_filter="TTSimDeviceIOFixture.*"
+
+      # wh_x2 is the N300 multichip simulator (chip 0 MMIO gateway + chip 1 remote over
+      # simulated eth). The N300 cluster descriptor makes the cluster enumerate both chips,
+      # so the standard device-IO fixtures run across the gateway and the remote chip.
+      - name: Run UMD TestDeviceIOFixture on tt-sim wormhole x2
+        timeout-minutes: 1
+        env:
+          TT_UMD_SIMULATOR: ${{ github.workspace }}/sim_wh_x2/libttsim.so
+          TT_UMD_CLUSTER_DESCRIPTOR: wormhole_N300.yaml
+        run: |
+          tt-metal/tt_metal/third_party/umd/build/test/umd/api/api_tests --gtest_filter="TestDeviceIOFixture.*"
+
+      - name: Run UMD TTSimDeviceIOFixture on tt-sim wormhole x2
+        timeout-minutes: 1
+        env:
+          TT_UMD_SIMULATOR: ${{ github.workspace }}/sim_wh_x2/libttsim.so
+          TT_UMD_CLUSTER_DESCRIPTOR: wormhole_N300.yaml
         run: |
           tt-metal/tt_metal/third_party/umd/build/test/umd/api/api_tests --gtest_filter="TTSimDeviceIOFixture.*"
 
--- a/.github/workflows/run-ttsim-tests.yml
+++ b/.github/workflows/run-ttsim-tests.yml
@@ -130,6 +130,10 @@ jobs:
           cp tt-metal/tt_metal/third_party/umd/tests/soc_descs/quasar_32_arch.yaml sim_qsr/soc_descriptor.yaml
           # wh_x2 is two Wormhole chips; reuse the per-chip Wormhole soc descriptor.
           cp $TT_METAL_HOME/tt_metal/soc_descriptors/wormhole_b0_80_arch.yaml sim_wh_x2/soc_descriptor.yaml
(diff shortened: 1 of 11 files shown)


.github/workflows/run-ttsim-tests.yml
```diff
@@ -97,6 +97,14 @@ jobs:
           fileName: "libttsim_qsr.so"
           out-file-path: /tmp/ttsim-qsr
 
+      - name: Download ttsim binary (wh_x2)
```
broskoTT, line 100: Should parametrize this whole file probably. Not for this PR though

device/api/umd/device/chip/remote_chip.hpp
```diff
@@ -37,7 +37,9 @@ class RemoteChip : public Chip {
      * @param local_chip The local chip to be used for communication to this remote chip.
      * @return A unique pointer to the created RemoteChip instance.
      */
-    static std::unique_ptr<RemoteChip> create(std::unique_ptr<TTDevice> remote_tt_device, LocalChip* local_chip);
+    static std::unique_ptr<RemoteChip> create(std::unique_ptr<TTDevice> remote_tt_device, Chip* local_chip);
+    static std::unique_ptr<RemoteChip> create_for_simulation(
```
broskoTT, line 41: Well... if we have LocalChip and SimulationChip, then we should probably have RemoteChip and RemoteSimulationChip?

I guess going forward we shouldn't have neither Simulation nor RemoteSimulation Chips....

broskoTT (commented):
A lot of 
throughout the PR...
Let's sync offline

device/api/umd/device/chip/remote_chip.hpp
pjanevskiTT: we don't need to have RemoteSimulationChip. The only thing that is different than remote chip is whether this goes through init process with ARC and everything. Since that is not simulated in TTSim atm, we need to skip it. I am happy to introduce this through a flag or something else.

> I guess going forward we shouldn't have neither Simulation nor RemoteSimulation Chips....

this is blocked by base API work. I pushed everything to TTDevice layer which was needed apart from sysmem and TLBs (IO windows). Don't want to do double work if we are going to reimplement that part anyway

device/chip/remote_chip.cpp
```diff
@@ -28,27 +30,50 @@ namespace tt::umd {
...
+    tt_device_ = std::move(remote_tt_device);
+}
+
 bool RemoteChip::is_mmio_capable() const { return false; }
 
 void RemoteChip::start_device(uint32_t dram_membar_subchannel) {}
 
 void RemoteChip::close_device() {
     ZoneScopedC(tracy::Color::DarkRed);
+#ifdef TT_UMD_BUILD_SIMULATION
```
broskoTT, line 72: In "create_for_simulation" constructor you can set a private member "is_simulation" if that is needed. And than check that variable here, without ifdefs.

device/chip/remote_chip.cpp
```diff
@@ -97,7 +122,14 @@ void RemoteChip::dram_membar(const std::unordered_set<uint32_t>& channels, uint3
     wait_for_non_mmio_flush();
 }
 
-void RemoteChip::deassert_risc_resets() { local_chip_->deassert_risc_resets(); }
+void RemoteChip::deassert_risc_resets() {
+#ifdef TT_UMD_BUILD_SIMULATION
```
broskoTT, line 126: Same as above

device/cluster.cpp
```diff
@@ -220,6 +250,9 @@ std::unique_ptr<Chip> Cluster::construct_chip_from_cluster(
     }
     if (chip_type == ChipType::SIMULATION) {
 #ifdef TT_UMD_BUILD_SIMULATION
+        if (simulator_directory.extension() == ".so" && !cluster_desc->is_chip_mmio_capable(chip_id)) {
```
broskoTT, line 253: nit: is_chip_remote

device/simulation/simulation_chip.cpp
```diff
@@ -36,6 +36,12 @@ std::string SimulationChip::get_soc_descriptor_path_from_simulator_path(const st
                                                  : (simulator_path / "soc_descriptor.yaml");
 }
 
+std::string SimulationChip::get_cluster_descriptor_path_from_simulator_path(
```
broskoTT, line 39: If we're looking to remove SimulationChip class altogether in the near future, let's not go in a direction against that. Consider putting this function in Chip or Cluster classes

device/cluster.cpp
```diff
@@ -367,6 +398,22 @@ Cluster::Cluster(ClusterOptions options) {
         case ChipType::SWEMULE:
         case ChipType::SIMULATION: {
             if (options.cluster_descriptor == nullptr) {
+#ifdef TT_UMD_BUILD_SIMULATION
```
broskoTT, line 401: You won't need ifdef here if "get_cluster_descriptor_path_from_simulator_path" is moved out of SimulationChip

.github/workflows/run-ttsim-tests.yml
broskoTT: But could be done as a preparation for this PR if you're motivated to improve this workflow :)

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -91,9 +91,12 @@ class TTDevice {
         bool use_safe_api = false,
         const std::shared_ptr<SocArchDescriptor> &soc_arch_descriptor = nullptr);
 
+    // A remote TTDevice is normally initialized over ARC (init_tt_device), which constructs its SocDescriptor.
+    // Simulated remote chips have no ARC to probe, so the caller may supply the full descriptor via soc_descriptor.
```
broskoTT, line 95: I don't think this is good to have longer term like this.
I understand it is not trivial to provide currently a mocked ARC communicator for ttsim which would provide info for soc descriptor.
Please mark the argument somehow that it is there temporarily, and will be removed once ttsim gets arc mock

device/cluster.cpp
```diff
@@ -198,6 +198,36 @@ void Cluster::construct_cluster(const uint32_t& num_host_mem_ch_per_mmio_device,
     }
 }
 
+#ifdef TT_UMD_BUILD_SIMULATION
```
broskoTT, line 201: What doesn't compile if you omit this ifdef?

device/cluster.cpp
broskoTT: Look at other places too

broskoTT (approved):
As discussed, I think having smaller PRs will contribute to:
- quicker time-to-review
- better review overall = better code quality
- easier potential issue bisecting
- untwisting impact and priorities of changes: Each change will clearly warrant its own impact, motivation for change is clearer from the scoped PR itself (description, test modifications, etc)

Citing your own list of changes:
- Adding cluster descriptor readout for ttsim. Along with new tests
- device->set_soc_descriptor(*soc_descriptor) on creation
- Changing RemoteChip to accept Chip instead of LocalChip
- (the main one, maybe even this exact PR) Add n300 ttsim variant + advance execution for remote chip + n300 ttsim tests + skip close device + skip deassert + create_simulation_remote_chip where needed

mcraigheadTT:
I definitely agree this has gotten pretty hairy. This was originally a draft PR that I had coded up on my end to go along with the ttsim multichip changes; if there is a different way that your team would like to handle the staging of these changes, I'm completely fine with that as long as we get to an endpoint we are all happy with in a reasonable timeline. My name is on the PR but as @pjanevskiTT has been doing most of the heavy lifting here, I think it would be better for UMD team to make the final decision of if/when to merge this and what fixes/cleanups are "must have" vs. "can come in a subsequent PR".

pjanevskiTT:
Superseded by the split stack (per @broskoTT's review), rebased onto current `main`:
1. #2952 — `RemoteChip` accepts `Chip*` base
2. #2953 — optional `SocDescriptor` at remote `TTDevice::create`
3. #2954 — discover ttsim cluster descriptor
4. #2955 — N300 multichip integration + CI

Closing in favour of those.

PR #2811, opened 2026-06-11
Simulation: SimulationConnector (host-vs-client decision)

pjanevskiTT:
### Issue
Part of #2792. Builds on `SimulationSocket` (#2800), now merged to `main`.

### Description
Adds `SimulationConnector` — the entry point that scans the per-chip socket path and
decides, per device, host vs client. Host path: no live socket → bring up a host backend
that binds + exposes its socket. The backend is selected from the simulator path
(`.so` → `TTSimTTDevice`, directory → `RtlSimulationTTDevice`), mirroring the simulation
device factory. The client/attach branch lands in a later PR; for now a live socket is
reported as an error.

The sim TTDevice ctors no longer auto-bind; the connector creates the `SimulationSocket`
and hands ownership to the device via `adopt_socket`. `num_host_mem_channels` is threaded
through `create` → `create_for_chip` → ctor.

(Renamed from `SimulationTopologyDiscovery`: it does not inherit from or match the
`TopologyDiscovery` interface, so the shared name was misleading.)

### List of the changes
- `simulation_connector.{hpp,cpp}` + `test_simulation_connector.cpp`.
- `tt_sim_tt_device` / `rtl_simulation_tt_device`: `socket_` member + `adopt_socket`
  (ctor no longer auto-binds); `num_host_mem_channels` threaded through
  `create`/`create_for_chip`.
- `simulation_socket.hpp`: doc comment updated for the rename.
- CMake.

### Testing
- `SimulationConnector.CreatesHostDeviceAndExposesSocket` host-path test + full TTSim
  device-IO suite pass against the qsr `libttsim.so`.
- `SimulationConnector.ThrowsWhenLiveHostAlreadyExists` covers the live-socket → throw
  branch with no simulator, so simulator-less CI exercises the core host-vs-client decision.
- pre-commit clean.

### API Changes
- New public header `umd/device/simulation/simulation_connector.hpp`
  (`SimulationConnector`, `SimulationConnectorOptions`).
- `TTSimTTDevice::adopt_socket` / `RtlSimulationTTDevice::adopt_socket`.
- `TTSimTTDevice::create_for_chip` gains a `num_host_mem_channels` parameter.

### DIFF
--- a/device/api/umd/device/simulation/simulation_connector.hpp
+++ b/device/api/umd/device/simulation/simulation_connector.hpp
@@ -15,7 +15,7 @@ namespace tt::umd {
 class TTDevice;
 class ClusterDescriptor;
 
-struct SimulationTopologyDiscoveryOptions {
+struct SimulationConnectorOptions {
     // The simulator build to host: a TTSim .so (when the path ends in .so) or an RTL simulator
     // directory. The backend is selected from the path, mirroring the simulation device factory.
     std::filesystem::path simulator_directory;
@@ -33,9 +33,9 @@ struct SimulationTopologyDiscoveryOptions {
 //
 // Socket-first: binding the socket is the host-vs-client arbiter. The client/attach branch lands
 // in a later PR; for now a live socket is reported as an error.
-class SimulationTopologyDiscovery {
+class SimulationConnector {
 public:
-    static std::map<ChipId, std::unique_ptr<TTDevice>> discover(const SimulationTopologyDiscoveryOptions& options);
+    static std::map<ChipId, std::unique_ptr<TTDevice>> discover(const SimulationConnectorOptions& options);
 };
 
 }  // namespace tt::umd
--- a/device/api/umd/device/simulation/simulation_topology_discovery.hpp
+++ b/device/api/umd/device/simulation/simulation_topology_discovery.hpp
@@ -16,7 +16,8 @@ class TTDevice;
 class ClusterDescriptor;
 
 struct SimulationTopologyDiscoveryOptions {
-    // The simulator build to host (the TTSim .so, or an RTL sim directory).
+    // The simulator build to host: a TTSim .so (when the path ends in .so) or an RTL simulator
+    // directory. The backend is selected from the path, mirroring the simulation device factory.
     std::filesystem::path simulator_directory;
     // Connectivity/topology used to configure the simulator on the host path. Optional; when
     // attaching to a live host the topology comes from the host instead.
--- a/device/api/umd/device/simulation/simulation_topology_discovery.hpp
+++ b/device/api/umd/device/simulation/simulation_topology_discovery.hpp
@@ -1,0 +1,40 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <filesystem>
+#include <map>
+#include <memory>
+
+#include "umd/device/types/cluster_descriptor_types.hpp"
+
+namespace tt::umd {
+
+class TTDevice;
+class ClusterDescriptor;
+
+struct SimulationTopologyDiscoveryOptions {
+    // The simulator build to host (the TTSim .so, or an RTL sim directory).
+    std::filesystem::path simulator_directory;
+    // Connectivity/topology used to configure the simulator on the host path. Optional; when
+    // attaching to a live host the topology comes from the host instead.
+    std::shared_ptr<ClusterDescriptor> cluster_descriptor;
+    int num_host_mem_channels = 0;
+};
+
+// Entry point for opening simulated devices, mirroring silicon TopologyDiscovery. It scans the
+// well-known socket folder and decides, per device, whether to be the host or a client:
+//   - no live socket  -> create a host device (the direct in-process backend / hot path) which
+//                        binds + serves its socket;
+//   - live socket     -> create a socket-backed client device that attaches to the host.
+//
+// Socket-first: binding the socket is the host-vs-client arbiter. The client/attach branch lands
+// in a later PR; for now a live socket is reported as an error.
+class SimulationTopologyDiscovery {
+public:
+    static std::map<ChipId, std::unique_ptr<TTDevice>> discover(const SimulationTopologyDiscoveryOptions& options);
+};
+
+}  // namespace tt::umd
--- a/device/api/umd/device/tt_device/tt_sim_tt_device.hpp
+++ b/device/api/umd/device/tt_device/tt_sim_tt_device.hpp
@@ -10,7 +10,6 @@
 #include <filesystem>
 #include <memory>
 #include <mutex>
-#include <utility>
 
 #include "umd/device/chip_helpers/simulation_sysmem_manager.hpp"
 #include "umd/device/chip_helpers/simulation_tlb_allocator.hpp"
@@ -49,7 +48,10 @@ public:
     // Named distinctly from create() because ChipId is an alias for int, which
     // would otherwise produce a duplicate signature.
     static std::unique_ptr<TTSimTTDevice> create_for_chip(
-        const std::filesystem::path &simulator_directory, ChipId chip_id, bool copy_sim_binary = false);
+        const std::filesystem::path &simulator_directory,
+        ChipId chip_id,
+        int num_host_mem_channels = 0,
+        bool copy_sim_binary = false);
 
     void read_from_device(void *mem_ptr, tt_xy_pair core, uint64_t addr, size_t size) override;
     void write_to_device(const void *mem_ptr, tt_xy_pair core, uint64_t addr, size_t size) override;
--- a/device/api/umd/device/tt_device/tt_sim_tt_device.hpp
+++ b/device/api/umd/device/tt_device/tt_sim_tt_device.hpp
@@ -10,6 +10,7 @@
 #include <filesystem>
 #include <memory>
 #include <mutex>
+#include <utility>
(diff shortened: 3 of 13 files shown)


device/simulation/simulation_topology_discovery.cpp
```diff
@@ -0,0 +1,46 @@
...
+            fmt::format(
+                "Attaching to an existing simulation host is not implemented yet; a live socket "
+                "already exists at: {}",
+                socket_path.string()));
+    }
+
+    // We are the host: bring up the backend (the direct in-process hot path) and hand it the
+    // socket to own and expose.
+    std::unique_ptr<TTSimTTDevice> device =
+        TTSimTTDevice::create(options.simulator_directory, options.num_host_mem_channels);
```
pjanevskiTT: `options.num_host_mem_channels` is silently dropped here. `TTSimTTDevice::create()` ignores its `num_host_mem_channels` parameter — it forwards to `create_for_chip()`, which hardcodes `0` into the ctor. So setting this option has no effect, which is a misleading trap for the next PR's caller. Either thread the value through `create()` → `create_for_chip()` → ctor, or drop the argument here and add a `// TODO: num_host_mem_channels not yet honored by TTSimTTDevice::create` so the dead wiring is documented rather than misleading.

tests/api/test_simulation_topology_discovery.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+#include "umd/device/simulation/simulation_socket.hpp"
+#include "umd/device/simulation/simulation_topology_discovery.hpp"
+#include "umd/device/tt_device/tt_device.hpp"
+
+using namespace tt::umd;
+
+// Integration: with no live host, discovery creates a host device that binds + exposes its
+// socket, and tears it down with the device. Requires TT_UMD_SIMULATOR.
+TEST(SimulationTopologyDiscovery, CreatesHostDeviceAndExposesSocket) {
```
pjanevskiTT: This is the only test, and it `GTEST_SKIP`s without `TT_UMD_SIMULATOR`, so simulator-less CI executes zero coverage of this PR — including the live-socket → throw branch, which is the core host-vs-client decision. That branch needs no simulator: hold the path live with `SimulationSocket::try_create(SimulationSocket::default_socket_path(0))`, then `EXPECT_THROW(SimulationTopologyDiscovery::discover(options), std::runtime_error)` — `discover` throws before constructing any device, so no simulator is required.

device/api/umd/device/tt_device/tt_sim_tt_device.hpp
```diff
@@ -10,6 +10,7 @@
 #include <filesystem>
 #include <memory>
 #include <mutex>
+#include <utility>
```
pjanevskiTT: `#include <utility>` is unused in this header (`std::move` appears only in the .cpp, and `std::unique_ptr` comes from `<memory>`). The rtl_simulation_tt_device.hpp change correctly didn't add it — drop for consistency.

device/api/umd/device/simulation/simulation_topology_discovery.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+// Entry point for opening simulated devices, mirroring silicon TopologyDiscovery. It scans the
+// well-known socket folder and decides, per device, whether to be the host or a client:
+//   - no live socket  -> create a host device (the direct in-process backend / hot path) which
+//                        binds + serves its socket;
+//   - live socket     -> create a socket-backed client device that attaches to the host.
+//
+// Socket-first: binding the socket is the host-vs-client arbiter. The client/attach branch lands
+// in a later PR; for now a live socket is reported as an error.
+class SimulationTopologyDiscovery {
```
aleksamarkovicTT: Since this class does not inherit from or match the interface of TopologyDiscovery, I think it should have a different name, like SimulationConnector or similar, to avoid confusion.

device/api/umd/device/simulation/simulation_connector.hpp
```diff
@@ -0,0 +1,40 @@
...
+#include "umd/device/types/cluster_descriptor_types.hpp"
+
+namespace tt::umd {
+
+class TTDevice;
+class ClusterDescriptor;
+
+struct SimulationTopologyDiscoveryOptions {
+    // The simulator build to host (the TTSim .so, or an RTL sim directory).
+    std::filesystem::path simulator_directory;
```
aleksamarkovicTT, line 21: I don't like that the type of the simulator is magically inferred from the path. Prefer to pass a simulator type enum, or similar.

broskoTT: Please sync offline on this and come to an agreement

pjanevskiTT: this is an orthogonal problem to the one the PR is solving. We have been inferring simulator type from path for a while now and I am open to suggestions on how to solve this in a better way, but definitely won't do it in this PR

device/api/umd/device/simulation/simulation_topology_discovery.hpp
pjanevskiTT: I think this is the way base API will define topology discovery. If not, I am happy to change it to whatever is needed to match that

pjanevskiTT: @nbuncicTT

tests/api/test_simulation_topology_discovery.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+    {
+        auto devices = SimulationTopologyDiscovery::discover(options);
+        ASSERT_EQ(devices.size(), 1u);
+        ASSERT_NE(devices.at(0), nullptr);
+        EXPECT_TRUE(std::filesystem::exists(socket));  // host exposed it
+    }
+
+    EXPECT_FALSE(std::filesystem::exists(socket));  // torn down with the device
+}
```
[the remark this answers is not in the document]
pjanevskiTT: Reworked: the test no longer unconditionally removes the shared path. It probes with try_create (skips if a live host already holds it) and releases via RAII, so a concurrent host is never clobbered.

device/simulation/simulation_topology_discovery.cpp
pjanevskiTT: Fixed: num_host_mem_channels is now threaded through TTSimTTDevice::create -> create_for_chip -> constructor instead of being dropped.

device/api/umd/device/tt_device/tt_sim_tt_device.hpp
pjanevskiTT: Removed the unused <utility> include from the header (std::move lives only in the .cpp, which includes <utility> itself).

tests/api/test_simulation_topology_discovery.cpp
pjanevskiTT: Added SimulationTopologyDiscovery.ThrowsWhenLiveHostAlreadyExists, which exercises the live-socket -> throw branch with no TT_UMD_SIMULATOR, so simulator-less CI now covers it.

device/api/umd/device/simulation/simulation_topology_discovery.hpp
```diff
@@ -0,0 +1,40 @@
...
+class ClusterDescriptor;
+
+struct SimulationTopologyDiscoveryOptions {
+    // The simulator build to host (the TTSim .so, or an RTL sim directory).
+    std::filesystem::path simulator_directory;
+    // Connectivity/topology used to configure the simulator on the host path. Optional; when
+    // attaching to a live host the topology comes from the host instead.
+    std::shared_ptr<ClusterDescriptor> cluster_descriptor;
+    int num_host_mem_channels = 0;
+};
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed: discover() now dispatches the host backend by simulator path (.so -> TTSimTTDevice, directory -> RtlSimulationTTDevice), mirroring the simulation device factory, so the option comment is accurate and RTL is supported.

device/api/umd/device/simulation/simulation_topology_discovery.hpp
aleksamarkovicTT: The pair returning ClusterDescriptor is missing, the options are differernt - but I understand that inheritance is problematic because of the abstract methods. It probably won't be used in place of regular discovery, so better to give it a different name.

device/api/umd/device/simulation/simulation_socket.hpp
```diff
@@ -0,0 +1,72 @@
...
+#include <atomic>
+#include <filesystem>
+#include <memory>
+#include <thread>
+
+#include "umd/device/types/cluster_descriptor_types.hpp"
+
+namespace tt::umd {
+
+// Owns the per-chip UNIX domain socket that exposes one simulation device on disk
```
aleksamarkovicTT: I think it might be worth porting this to use asio instead of using the low-level socket interface directly. i'm sure it would be less code, more portable and easier to understand because of the library's C++ interface. I know the work has already been done, but the PR feels like it's cutting a few layers of abstraction.

aleksamarkovicTT (approved):

device/api/umd/device/simulation/simulation_socket.hpp
pjanevskiTT: this was a bad rebase, no socket in this PR

device/api/umd/device/simulation/simulation_topology_discovery.hpp
pjanevskiTT: renamed to SimulationConnector

PR #2817, opened 2026-06-12
More CoreCoord methods in TTDevice

aleksamarkovicTT:
### Issue
#2836

### Description
Adapt more methods in TTDevice to use `CoreCoord`. Fix potential NOC mismatches when accessing ARC core inside TTDevice.

### List of the changes
- Persist ARC core coordinates for NOC0 and NOC1.
- `get_arc_core()` returns NOC0 or NOC1 coordinate depending on currently selected NOC.
- Added CoreCoord variants for `assert_risc_reset()`, `deassert_risc_reset()`, `dma_read_from_device()`, `dma_write_to_device()`, `read_eth_core_training_status()`.

### Testing
CI

### API Changes
This PR has new methods in TTDevice.

### DIFF
--- a/device/api/umd/device/tt_device/tt_device.hpp
+++ b/device/api/umd/device/tt_device/tt_device.hpp
@@ -458,6 +458,7 @@ public:
      * @return Training status
      */
     virtual EthTrainingStatus read_eth_core_training_status(tt_xy_pair eth_core) = 0;
+    EthTrainingStatus read_eth_core_training_status(CoreCoord eth_core);
 
     const SocDescriptor &get_soc_descriptor() const;
 
--- a/device/api/umd/device/tt_device/tt_device.hpp
+++ b/device/api/umd/device/tt_device/tt_device.hpp
@@ -490,7 +490,8 @@ protected:
 
     bool is_remote_tt_device = false;
 
-    tt_xy_pair arc_core;
+    CoreCoord arc_core_noc0;
+    CoreCoord arc_core_noc1;
 
     void construct_soc_descriptor(const std::shared_ptr<SocArchDescriptor> &soc_arch_descriptor);
     void set_soc_descriptor(const SocDescriptor &soc_descriptor);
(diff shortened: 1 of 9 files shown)


device/tt_device/blackhole_tt_device.cpp
```diff
@@ -43,7 +44,8 @@ BlackholeTTDevice::BlackholeTTDevice(
     const std::shared_ptr<SocArchDescriptor> &soc_arch_descriptor,
     bool use_safe_api) :
     TTDevice(std::move(pci_device), std::make_unique<blackhole_implementation>(), soc_arch_descriptor, use_safe_api) {
-    arc_core = blackhole::get_arc_core(BlackholeTTDevice::get_noc_translation_enabled(), is_selected_noc1());
+    arc_core_noc0 = blackhole::get_arc_core(BlackholeTTDevice::get_noc_translation_enabled(), /*use_noc1=*/false);
```
pjanevskiTT: are this constructor and the one below completely the same? Should we remove code duplication by having helper function

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -504,7 +510,8 @@ class TTDevice {
 
     bool is_remote_tt_device = false;
 
-    tt_xy_pair arc_core;
+    CoreCoord arc_core_noc0;
```
pjanevskiTT: nit: not sure what is the right path for storing these, should we do just one or both? I would maybe prefer having only one and doing translation whenever we need NOC1, but NOC1 is not inferior to NOC0 in any sense so there is no reason to not do this. Just wanted to put it out there, fine with this

pjanevskiTT (approved):

device/api/umd/device/tt_device/tt_device.hpp
aleksamarkovicTT: I think the coordinate should come from SocArchDescriptor, we should a method. Translation might not always be available, so I think the coordinate should be persisted statically for all NOCs. Gonna keep it like this but will add a followup.

device/tt_device/blackhole_tt_device.cpp
aleksamarkovicTT: done

PR #2825, opened 2026-06-15
feat(sim): P300 multi-MMIO TTSim — BDF enumeration + outbound-iATU host-DMA routing

pjanevskiTT:
### Issue
/

### Description
Adds support for driving a multi-MMIO multi-chip TTSim — a simulator config where more than one chip is MMIO-capable, such as the Blackhole P300 (two MMIO Blackhole chips). Until now the TTSim path assumed a single MMIO device, so a P300 cluster could not be enumerated or driven.

The harder part is host DMA: with one shared simulator image serving both chips, a sysmem DMA callback fires process-globally, so we need to land each chip's DMA in that chip's own host memory. We model this exactly the way silicon does, rather than inventing a sim-only side channel: each chip's sysmem sits at a distinct host base, UMD programs the chip's outbound iATU (over BAR2) to map its NOC sysmem window onto that base, and a DMA is then routed to the owning chip purely by the host address it carries — no per-chip tag crosses the bus.

Doing it this way means no new ttsim ABI is required: the device dimension already exists in the stable ABI (the BDF arg to `pci_config_rd32` and the physical address to `pci_mem_*`), so each chip is simply addressed by its PCI device and its per-device BAR window.

### List of the changes
- Shared-dlopen BDF mode in `TTSimCommunicator`: chips addressed by per-device BAR; init and DMA callbacks run once on the first chip.
- Outbound-iATU host-DMA routing: per-chip host base, iATU programmed via BAR2, `dma_route` routes by host-address window. Programmed uniformly for single- and multi-chip.
- `RemoteChip`/cluster plumbing to construct simulation chips, threading `num_chips` through.
- Robustness fixes in the routing path (teardown unregister, `callback_instance_` race, BDF/iATU range asserts).
- CI: bh_x2 (P300) lane; `SysmemReadWrite` covers all MMIO chips.

### Testing
Verified locally against a ttsim built from the multichip branch: single-chip WH/BH and bh_x2 P300 pass (both chips enumerate, P300 eth topology parses, per-chip DMA routes correctly). Requires a ttsim modeling the BAR2 outbound iATU (stable release).

### API Changes
None public. `TTSimCommunicator` / `TTSimTTDevice` gain a `num_chips` parameter (default 1).

### DIFF
--- a/device/simulation/tt_sim_communicator.cpp
+++ b/device/simulation/tt_sim_communicator.cpp
@@ -179,9 +179,7 @@ void TTSimCommunicator::initialize() {
         shared_bdf_mode_ = true;
         s_shared_refcount_++;
         log_info(
-            tt::LogEmulationDriver,
-            "TTSim BDF multichip mode (chip_id={}, shared dlopen, per-device BARs)",
-            chip_id_);
+            tt::LogEmulationDriver, "TTSim BDF multichip mode (chip_id={}, shared dlopen, per-device BARs)", chip_id_);
         return;
     }
 
@@ -301,7 +299,7 @@ uint32_t TTSimCommunicator::pci_config_read32(uint32_t bus_device_function, uint
     // Callers pass bus_device_function 0 ("this device"); we fill in the device field.
     uint32_t bdf = bus_device_function;
     if (shared_bdf_mode_) {
-        bdf |= (chip_id_ << 3); // BDF: function[2:0], device[7:3], bus[15:8]
+        bdf |= (chip_id_ << 3);  // BDF: function[2:0], device[7:3], bus[15:8]
     }
     select_chip_if_needed();  // no-op outside the multichip-ABI mode
     return pfn_libttsim_pci_config_rd32_(bdf, offset);
@@ -323,17 +321,27 @@ void TTSimCommunicator::advance_clock(uint32_t n_clocks) {
 }
 
 TTSimCommunicator *TTSimCommunicator::callback_instance_ = nullptr;
-std::array<TTSimCommunicator *, TTSimCommunicator::kMaxDmaDevices> TTSimCommunicator::dma_instances_{};
+std::array<TTSimCommunicator::DmaHostRange, TTSimCommunicator::kMaxDmaDevices> TTSimCommunicator::dma_ranges_{};
+std::size_t TTSimCommunicator::dma_range_count_ = 0;
+std::mutex TTSimCommunicator::dma_ranges_mutex_;
 
-// Decode the originating MMIO device from a sim-tagged sysmem DMA paddr and return its
-// communicator plus the within-device offset. The sim adds device * kPerDevicePaddrStride
-// (see libttsim_pci_dma_mem_{rd,wr}_bytes); strip it and look the device up. Falls back to
-// callback_instance_ for the single-device / legacy path (device 0, no registered entry).
+// Route a sysmem DMA to the owning MMIO chip by host address: the address the chip emits already went
+// through its outbound iATU (UMD-programmed target = that chip's distinct host base), so we just find
+// the registered host window [host_base, host_base+size) that contains it -- exactly as a host routes a
+// DMA by which pinned region the physical address falls in. Rebase paddr to the within-window offset (so
+// the per-chip callback sees an offset relative to its own host base, like the single-device case) and
+// return the owning communicator. Falls back to callback_instance_ when no window matches (single
+// device, or an address outside any registered region).
 TTSimCommunicator *TTSimCommunicator::dma_route(uint64_t &paddr) {
-    uint32_t device = static_cast<uint32_t>(paddr / kPerDevicePaddrStride);
-    paddr -= static_cast<uint64_t>(device) * kPerDevicePaddrStride;
-    TTSimCommunicator *inst = (device < kMaxDmaDevices) ? dma_instances_[device] : nullptr;
-    return inst ? inst : callback_instance_;
+    std::lock_guard<std::mutex> lock(dma_ranges_mutex_);
+    for (std::size_t i = 0; i < dma_range_count_; i++) {
+        const DmaHostRange &r = dma_ranges_[i];
+        if (paddr >= r.host_base && paddr - r.host_base < r.host_size) {
+            paddr -= r.host_base;
+            return r.inst;
+        }
+    }
+    return callback_instance_;
 }
 
 void TTSimCommunicator::pci_dma_mem_rd_bytes_wrapper(uint64_t paddr, void *p, uint32_t size) {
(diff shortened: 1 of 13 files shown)


device/simulation/tt_sim_communicator.cpp
```diff
@@ -273,26 +321,77 @@ void TTSimCommunicator::advance_clock(uint32_t n_clocks) {
...
+        bool updated = false;
+        for (std::size_t i = 0; i < dma_range_count_; i++) {
+            if (dma_ranges_[i].inst == this) {  // re-registration: update in place
+                dma_ranges_[i] = {host_base, host_size, this};
+                updated = true;
+                break;
+            }
+        }
+        if (!updated && dma_range_count_ < kMaxDmaDevices) {
+            dma_ranges_[dma_range_count_++] = {host_base, host_size, this};
```
pjanevskiTT: `dma_ranges_` and `callback_instance_` are never cleared when a communicator is destroyed — the destructor (line 51) only touches the shared-handle refcount. If any chip's `TTSimCommunicator` is destroyed before the others (or during staggered teardown) while the shared simulator can still emit a DMA, `dma_route()` can return a dangling `inst` (use-after-free), and `dma_range_count_` never shrinks. Suggest removing this instance's entry — and clearing `callback_instance_` if it points to `this` — in the destructor under `dma_ranges_mutex_`.

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -154,6 +157,20 @@ TTSimTTDevice::TTSimTTDevice(
                 tt::arch_to_str(arch));
             break;
     }
+
+    // Program this chip's outbound iATU exactly as UMD does on silicon (LocalChip::init_pcie_iatus):
+    // one region per host-mem channel, mapping the NOC sysmem window onto this chip's distinct host
+    // base. The simulator honors the iATU at DMA egress, so each chip's DMA lands in its own host
+    // window by address alone -- no per-chip tag. Use a uniform 1 GiB region stride so region `ch`
+    // maps NOC-window offset ch*1GiB -> this chip's host base + ch*1GiB (matches SimulationSysmemManager).
+    if (arch == tt::ARCH::WORMHOLE_B0 || arch == tt::ARCH::BLACKHOLE) {
```
pjanevskiTT, line 173: This iATU-programming block now runs for every WH/BH device, including the single-chip legacy path (`num_chips == 1`), and `SimulationSysmemManager` now bases the hugepage `physical_address` at `host_base_` (0 for chip 0) instead of `pcie_base_`. The PR states single-chip behaviour is unchanged — please confirm the legacy single-chip sims simply ignore these BAR2 iATU register writes, and that no other consumer of `get_hugepage_mapping().physical_address` relied on the old `pcie_base`-relative value.

device/simulation/tt_sim_communicator.cpp
```diff
@@ -251,10 +294,15 @@ void TTSimCommunicator::pci_mem_write_bytes(uint64_t paddr, const void *data, ui
...
-    // pci_config_read32 reads compile-time constants; no chip selection needed.
-    // But we still select for consistency and future-proofing.
-    select_chip_if_needed();
-    return pfn_libttsim_pci_config_rd32_(bus_device_function, offset);
+    // In BDF mode there is no select_device_by_id: this chip's PCI device is named by
+    // its BDF (device field = chip_id), so each chip reads its own per-device BAR bases.
+    // Callers pass bus_device_function 0 ("this device"); we fill in the device field.
+    uint32_t bdf = bus_device_function;
+    if (shared_bdf_mode_) {
+        bdf |= (chip_id_ << 3);  // BDF: function[2:0], device[7:3], bus[15:8]
```
pjanevskiTT: The device field is 5 bits (`[7:3]`), so `chip_id >= 32` silently overflows into the bus field. A 32-chip cluster (ids 0..31) sits exactly at the limit. Consider an `assert(chip_id_ < 32)` so this fails loudly rather than misrouting.

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -347,10 +364,40 @@ void TTSimTTDevice::dma_multicast_write(
...
+    bar2_base &= ~15ull;  // strip BAR type/attribute bits, leaving the physical address
+    const uint64_t base = uint64_t(region) * region_size;  // region offset within the NOC sysmem window
+    const uint64_t limit = base + region_size - 1;
+    const uint64_t iatu = bar2_base + iatu_bar2_off + uint64_t(region) * 0x200;
+    auto wr = [&](uint64_t off, uint32_t val) { communicator_->pci_mem_write_bytes(iatu + off, &val, sizeof(val)); };
+    wr(0x04, 0);                       // region_ctrl_2: disable while (re)programming the region
+    wr(0x00, 0);                       // region_ctrl_1
+    wr(0x08, uint32_t(base));          // base lo
+    wr(0x0c, uint32_t(base >> 32));    // base hi
+    wr(0x10, uint32_t(limit));         // limit (low 32b; upper bits share base hi)
```
pjanevskiTT, line 403: `limit` is written as a `uint32_t`; this is only correct while `base + region_size - 1 < 4 GiB`. With `kChannelStride = 1 GiB` and up to 4 channels the top region ends exactly at `0xFFFFFFFF`, so any future increase in stride or channel count would silently truncate the limit. Consider writing the high bits (sharing `base hi`) or asserting the range.

device/tt_device/tt_sim_tt_device.cpp
pjanevskiTT: Analysis (single-chip behaviour is preserved):

**1. Do legacy single-chip sims ignore the BAR2 iATU writes?** It doesn't matter either way for chip 0, because `host_base_ == 0`:
- `set_pcie_dma_mem_callbacks` is always called with `host_size = kPerChipHostStride`, so chip 0 registers the window `[0, 64 GiB)`. In `dma_route()` the rebase `paddr -= host_base_` is therefore a **no-op**, leaving exactly the hugepage-space offset the legacy `translate_pci_dma_addr` path produced.
- The iATU target is `physical_address = host_base_ + ch*1GiB = ch*1GiB` for chip 0. So whether the legacy sim honors the iATU writes (maps to host base 0) or ignores them (keeps subtracting `pcie_base` itself), the offset that reaches `pci_dma_{read,write}_bytes` is the same `ch*1GiB + within`, and `channel = paddr / 1GiB` resolves correctly.

**2. Did any other consumer rely on the old `pcie_base`-relative `physical_address`?** No. Within the simulation path the only consumer of `get_hugepage_mapping().physical_address` is the iATU programming loop in `TTSimTTDevice` (the line this comment is on). The mapped-buffer registry and `map_sysmem_buffer` compute their device IO addresses from `pcie_base_` **independently**, so the `physical_address` rebasing to `host_base_` doesn't touch them.

So the change is safe for `num_chips == 1`: the per-chip host base collapses to 0 and the routing degenerates to the previous offset semantics. The single-chip `TestDeviceIOFixture` / `TTSimDeviceIOFixture` lanes in the ttsim CI workflow exercise exactly this path and are the regression gate.

pjanevskiTT: Correction to my earlier reply: single-chip behaviour was **not** preserved — this was a real regression, now fixed.

The outbound-iATU-via-BAR2 programming runs at device init for every WH/BH device, including the single-chip legacy path (`num_chips == 1`). The single-chip libttsim builds do not model the BAR2 outbound iATU and throw `UnimplementedFunctionality: pci_mem_wr_cur: bar2: paddr=0x...1004` on the first register write — which broke the previously-passing single-chip wh/bh sim tests (reproduced locally against the v1.8.3 release binaries).

Fix (commit `174f8ca9`): gate the iATU programming on `num_chips > 1`. A single-chip sim has one host window at base 0, so DMA routing degenerates to a no-op and the legacy path already works; only the genuine multi-MMIO case programs the iATU.

Verified locally:
- single-chip `wh` and `bh` (v1.8.3 release): `SysmemReadWrite` passes again.
- bh_x2 P300, built from the `mcraighead/multichip` ttsim branch (which models the BH iATU at BAR2 0x1000 + per-device BAR stride): full `TestDeviceIOFixture` passes — both chips enumerate via BDF, the P300 eth topology parses, and per-chip host DMA routes correctly. The pinned ttsim *release* doesn't model the iATU yet, so the bh_x2 CI steps are marked `continue-on-error` until it ships.

pjanevskiTT: Superseding my earlier "gate on `num_chips > 1`" reply: that gate was removed (`d987cc16`). The final design programs the outbound iATU **uniformly** for single- and multi-chip — single-chip is just `host_base = 0`, an identity mapping — so the init path is identical regardless of `num_chips`.

This intentionally depends on the stable ttsim modeling the BAR2 outbound iATU (verified locally against a ttsim built from the multichip branch); we no longer assume the legacy sim ignores the writes, and the PR no longer claims single-chip behaviour is unchanged.

Re: `physical_address` — its only consumer is the iATU programming; the mapped-buffer registry uses `pcie_base_` independently. Resolving.

pjanevskiTT:
split in two other PRs, will close this one

Issue #2526, opened 2026-04-27
UMD: Improve ARC init failure diagnostics


smohammadiTT:
## Problem                                                                                                              
                                                                                                                          
  When `wait_arc_core_start()` times out, the thrown `ArcStartupError` doesn't                                            
  tell you which physical chip failed or what GDDR state it was in.                                                       
                                                                                                                          
  ## Asks
                                                                                                                          
  **1. Add PCI identity to `ArcStartupData`** (`tt_device_error.hpp`):
  ```cpp
  // Add to ArcStartupData:
  uint32_t pci_interface_id = 0;   // /dev/tenstorrent/N  — get_device_num()
  uint32_t pci_bus = 0;            // from PciDeviceInfo                                                                  
  uint32_t pci_device = 0;
  std::string board_type;          // e.g. "GALAXY"       — get_board_type()                                              
  uint32_t asic_location = 0;      // UBB X, ASIC Y       — get_asic_location()          ```                                 
   
  2. Capture GDDR_STATUS at timeout and include in error message:                                                         
  GDDR_STATUS: bit[2n]=training_complete, bit[2n+1]=training_error per channel n
  0x55555555 = all 8 channels passed                                                                                      
                                                            
  Suggested output:
  ARC failed to initialize on PCI 0000:41:00.0 (/dev/tenstorrent/4) [UBB 2 ASIC 7, GALAXY]
  SCRATCH_RAM[2]: 0x00000004  POST: 0x...                                                 
  GDDR_STATUS:    0x55555551  (channel 0 failed training)                                                                 
   
  Call stack                                                                                                              
                                                            
  Post-reset path (300 s timeout):                                                                                        
  TopologyDiscovery::discover()                         [topology_discovery.cpp:153]
    └─ TTDevice::init_tt_device(ARC_LONG_POST_RESET_TIMEOUT)  [tt_device.cpp:104]   
         └─ probe_arc()                                        [tt_device.cpp:114]
         └─ BlackholeTTDevice::wait_arc_core_start(300s)       [blackhole_tt_device.cpp:178]                              
              [polls SCRATCH_RAM_2 & 0x7 == 0x5]                                            
              → UMD_THROW(ArcStartupError)                     [blackhole_tt_device.cpp:227]                              
                                                            
  Warm reset path:                                                                                                        
  WarmReset::reset_and_wait()                               
[...]

PR #2829, opened 2026-06-15
Display ARC SMC Init status in ArcStartupError

aleksamarkovicTT:
### Issue
#2526 

### Description
Display ARC SMC Init status in ArcStartupError. Implemented in CMFW 19.11, ARC firmware now provides flags that give more details when ARC startup failed. These are now surfaced in ArcStartupError.

### List of the changes
- Added SMC init status to ArcStartupError.
- Added interpretation for bits of ERROR_STATUS0.

### Testing
CI

### API Changes
This PR has (no) API changes.

### DIFF
--- a/device/api/umd/device/types/blackhole_arc.hpp
+++ b/device/api/umd/device/types/blackhole_arc.hpp
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <cstdint>
+#include <string>
 #include <string_view>
 #include <unordered_map>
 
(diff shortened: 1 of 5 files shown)


device/api/umd/device/types/blackhole_arc.hpp
```diff
@@ -53,4 +56,44 @@ enum BlackholeArcMessageQueueIndex : uint8_t {
     APPLICATION = 3,
 };
 
+// SMC init stages.
```
pjanevskiTT: comment doesn't help much in understanding the code, I think it's readable enough even without it

pjanevskiTT (approved):

PR #2840, opened 2026-06-17
Add DeviceTimeoutError exception type

pjanevskiTT:
### Issue

#1673

### Description

Adds `DeviceTimeoutError`, a structured exception for host-side MMIO operations that overrun a wall-clock budget. It is the dependency leaf of the per-op MMIO timeout work (#2629) — split out into its own PR so the timeout mechanism can stack on top of it.

`DeviceTimeoutError` is distinct from `SigbusError`: `SigbusError` signals the platform PCIe completion timeout (a hardware-detected fault), whereas `DeviceTimeoutError` signals that a software per-op budget elapsed — typically because writes were piling up against a slow or dead NOC before SIGBUS could fire. Keeping them as separate types lets callers branch on retry policy.

There are no throwers yet; this PR only introduces the type and its metadata.

### List of the changes

- New `tt::umd::error::DeviceTimeoutData` struct carrying the timing/progress context of a timed-out transfer (function name, operation, op bytes, elapsed time, configured budget, bytes remaining, total bytes).
- New `tt::umd::error::DeviceTimeoutError : public UmdError<DeviceTimeoutData>`, with a formatted message derived from that metadata.

### Testing

CI

### API Changes

- New exception type `tt::umd::error::DeviceTimeoutError`.
- New metadata struct `tt::umd::error::DeviceTimeoutData`.

### DIFF
--- a/device/api/umd/device/utils/error.hpp
+++ b/device/api/umd/device/utils/error.hpp
@@ -4,6 +4,9 @@
 
 #pragma once
 
+#include <chrono>
+#include <cstddef>
+#include <cstdint>
 #include <stdexcept>
 #include <string>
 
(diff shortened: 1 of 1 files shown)


device/api/umd/device/utils/error.hpp
```diff
@@ -42,4 +45,57 @@ class SigbusError : public std::runtime_error {
...
+    DeviceTimeoutError(
+        const std::string& function_name,
+        const std::string& operation,
+        std::uint32_t op_bytes,
+        std::chrono::nanoseconds delta,
+        std::chrono::milliseconds budget,
+        std::size_t bytes_remaining,
+        std::size_t total_bytes) :
+        UmdError<DeviceTimeoutData>(
+            format_message(function_name, operation, op_bytes, delta, budget, bytes_remaining, total_bytes),
```
aleksamarkovicTT: If you prefer to use fmt, you can move the implementation of this method to error.cpp and include fmt there. No action required.

aleksamarkovicTT (approved):

device/api/umd/device/utils/error.hpp
```diff
@@ -42,4 +45,57 @@ class SigbusError : public std::runtime_error {
...
 };
 
+/**
+ * @brief Structured metadata for a host-side MMIO operation that overran its budget.
+ *
+ * Carries the timing and progress context of the timed-out transfer. function_name and
+ * operation are owned std::strings, so callers may pass temporaries safely.
+ */
+struct DeviceTimeoutData {
+    std::string function_name;            ///< Name of the IO routine that timed out (e.g. "memcpy_to_device").
```
aleksamarkovicTT: Can you clarify why this would be useful to persist in this struct? There is a stack trace available when throwing an exception of this type. 

Instead, I see that there's no data identifying which device has the timeout. Consider moving this to tt_device_error.hpp and inheriting from TTDeviceData.

pjanevskiTT: Two parts here.

**On device identity:** agreed it's valuable, but I'd keep it out of this type for a layering reason. `DeviceTimeoutError` is thrown from the raw memcpy free functions in `device/pcie/device_memcpy.cpp` (`memcpy_to_device(volatile void*, const void*, size_t, ...)`) — there's no `TTDevice`/`chip_id`/`arch` in scope there, just pointers. `TTDeviceData`'s constructor requires a `const TTDevice&`, so inheriting it would mean plumbing a device handle down into the memcpy primitives. That's the same reason its sibling `SigbusError` lives in `error.hpp` rather than `tt_device_error.hpp` — everything under `device/pcie/` includes `utils/error.hpp` and nothing includes `tt_device_error.hpp` (which would drag in arch/xy_pair/noc_id/cluster types). I'd rather attach device identity at a higher layer that catches and enriches the timeout (the protocol/TTDevice level, which has the device) as a follow-up, than inherit `TTDeviceData` at a throw site that can't construct it.

**On persisting the timing fields:** the stack trace gives the code path but not the runtime values (actual elapsed vs configured budget, bytes remaining/total) — those are what a consumer branches on for telemetry/retry, which is what the structured `UmdError<Data>` payload is for. The consumer lands in the stacked timeout PR (#2629). That said, if some of these (`op_bytes`/`bytes_remaining`/`total_bytes`) turn out unused there, I'm happy to trim to just `budget`/`delta`.

aleksamarkovicTT: I asked about `function_name`. Timing fields and others make sense here.

I agree about TTDevice or even PCI device ID not being available here. Sure, we can catch and throw a richer error higher up, I think it will be very useful to have a device ID.

pjanevskiTT: makes sense to remove it, it's in stack trace

PR #2842, opened 2026-06-17
Add generic OpTimeoutGuard timing primitive

pjanevskiTT:
### Issue

#1673

### Description

Adds one small, standalone, header-only timing primitive in `utils`, generic and unit-testable without hardware:

- `OpTimeoutGuard` — times a single operation against a fixed per-op wall-clock budget and reacts to overruns. `record_and_check()` takes the timestamp sampled immediately before the op began; if the elapsed time exceeds the budget the guard reacts, otherwise it does nothing. A budget of `0` disables the check entirely (matching the `utils::check_timeout` convention).

Two caller-supplied behaviors layer on top of the bare budget check:

- an optional **false-alarm veto** (`is_false_alarm`), consulted only on an apparent overrun: returning `true` means "false positive, the op is healthy" and the overrun is ignored; returning `false` (or having no callback) confirms it.
- an **overrun action** (`on_overrun`), invoked on a confirmed overrun as `on_overrun(delta, args...)` — the measured delta followed by whatever extra arguments the caller forwarded to `record_and_check()`.

The action and the per-call arguments are deliberately generic so the guard stays reusable: the domain-specific reaction (e.g. throwing a typed timeout error with operation metadata) lives at the call site, not in the guard. The elapsed delta is captured the moment the overrun is detected, before the (possibly slow) false-alarm probe runs, so the reported delta reflects the op's time and not the probe's.

This is the timing building block for the per-op MMIO timeout work (#2629), which consumes it from its own memcpy adapter — split into its own PR so it can be reviewed and tested in isolation.

### List of the changes

- New `tt::umd::OpTimeoutGuard` in `device/api/umd/device/utils/op_timeout_guard.hpp` — header-only, `template <typename OnOverrun>`; `void record_and_check(std::chrono::steady_clock::time_point start, Args&&... args)` composing the budget check, the optional veto, and the generic overrun action.
- Unit tests in `tests/misc/test_op_timeout_guard.cpp` covering under-budget, zero-budget-disables, over-budget-fires, false-alarm suppression, confirmed overrun, action-throws-propagates, and delta/arg forwarding.
- Wired the new test into `tests/misc/CMakeLists.txt`.

### Testing

CI

### API Changes

- New `tt::umd::OpTimeoutGuard` (`utils/op_timeout_guard.hpp`).


device/api/umd/device/utils/op_budget_timer.hpp
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * @brief Times individual operations against a fixed per-op wall-clock budget.
+ *
+ * Construct once with a budget, then call record_and_check() for each op, passing the timestamp
+ * sampled immediately before the op began. It returns whether that op overran the budget — nothing
+ * more. The timer is stateless across ops and knows nothing about why the op is timed or what an
+ * overrun means; the caller decides how to react (e.g. probe for liveness, then throw). This keeps
+ * the timer a reusable, hardware-free, unit-testable primitive.
+ */
+class OpBudgetTimer {
```
nbuncicTT: I think this whole class can be deleted actually. We already have a timeout utility - `utils::check_timeout`. 

The replacement of this timer would be trivial in `OpTimeoutGuard::record_and_check`: 

`if (!timer_.record_and_check(start))` would become `if (!utils::check_timeout(start, budget_))`

The semantics of the function differ a bit, (>= vs >, and for 0 it has different behavior), but this shouldn't be a problem.

device/api/umd/device/utils/op_timeout_guard.hpp
```diff
@@ -0,0 +1,66 @@
...
+        // Over budget: an optional liveness probe gets to veto the reaction by declaring the op healthy.
+        if (on_timeout_ && !on_timeout_()) {
+            return;
+        }
+        on_overrun_(delta, std::forward<Args>(args)...);
+    }
+
+private:
+    OpBudgetTimer timer_;
+    std::function<bool()> on_timeout_;
```
nbuncicTT: Does it make sense to rename this to is_false_alarm_?

Then the if a few lines up would be:
```cpp
if (is_false_alarm_ && is_false_alarm_()) {
    return;
}
```
Because is_false_alarm returns a bool, and then we know the semantic meaning of the returned bool from the name..

tests/misc/test_op_budget_timer.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+#include <chrono>
+
+#include "umd/device/utils/op_budget_timer.hpp"
+
+using namespace tt::umd;
+using namespace std::chrono_literals;
+
+// An op that completes well within its budget is not over budget.
+TEST(OpBudgetTimer, UnderBudgetReturnsFalse) {
```
nbuncicTT: These tests can be omitted if `check_timeout` is used.

nbuncicTT (approved):

tests/misc/test_op_budget_timer.cpp
pjanevskiTT: changed the tests to test the other component

Issue #2749, opened 2026-06-01
Extract tt-kmd-lib


broskoTT:
More details in https://tenstorrent.atlassian.net/wiki/spaces/LLK/pages/2283405852/Repo+restructuring
- Change all ioctl calls in pci_device to call something in tt_kmd_lib
- pci_device shouldn't include ioctl.h anymore
- extract tt_kmd_lib to a separate component in the root of the repo. It should be consumed as PRIVATE in tt_umd package

broskoTT:
Started evaluating changes needed and started removing direct ioctl calls from PCIDevice

broskoTT:
@aleksamarkovicTT please ping me on slack when this is done, I want to communicate further

nbuncicTT:
Two PRs done, progress slowed down due to higher priority work.

nbuncicTT:
4 PRs out, outstanding 5th PR which creates a static lib out of the KMD lib.

nbuncicTT:
Pending: 2/5 PRs merged, 3 are awaiting review.

PR #2844, opened 2026-06-17
Add reset function to tt-kmd-lib

aleksamarkovicTT:
### Issue
Starts #2749

### Description
Add reset functionality to tt-kmd-lib, removing the need for a direct ioctl call in PciDevice.

### List of the changes
- Introduce `tt_device_reset()` to tt-kmd-lib.
- Replace ioctl call in `PciDevice` with kmd lib function.
- Rename method to `send_reset_ioctl_to_devices()`, move loop.

### Testing
CI

### API Changes
This PR has API changes.

### DIFF
--- a/device/api/umd/device/pcie/pci_device.hpp
+++ b/device/api/umd/device/pcie/pci_device.hpp
@@ -295,7 +295,14 @@ public:
     static uint8_t read_command_byte(const int pci_device_num);
 
     /**
-     * Reset device via ioctl.
+     * @brief Resets the specified Tenstorrent PCIe devices.
+     *
+     * @param pci_target_devices A set of PCI device identifiers to be reset.
+     *                          Each identifier uniquely identifies a device on the PCI bus.
+     * @param flag The type of reset operation to perform on the target devices.
+     *
+     * @note This is a blocking operation that may take time to complete depending
+     *       on the number of devices and the reset type.
      */
     static void reset_devices(const std::unordered_set<int> &pci_target_devices, TenstorrentResetDevice flag);
 
--- a/device/api/umd/device/pcie/pci_device.hpp
+++ b/device/api/umd/device/pcie/pci_device.hpp
@@ -297,7 +297,7 @@ public:
     /**
      * Reset device via ioctl.
      */
-    static void reset_device_ioctl(const std::unordered_set<int> &pci_target_devices, TenstorrentResetDevice flag);
+    static void reset_devices(const std::unordered_set<int> &pci_target_devices, TenstorrentResetDevice flag);
 
     /**
      * Temporary function which allows us to support both ways of mapping buffers during the transition period.
--- a/device/api/umd/device/pcie/pci_device.hpp
+++ b/device/api/umd/device/pcie/pci_device.hpp
@@ -297,14 +297,16 @@ public:
     /**
      * @brief Resets the specified Tenstorrent PCIe devices.
      *
-     * @param pci_target_devices A set of PCI device identifiers to be reset.
+     * @param pci_target_devices A set of PCI device identifiers to be reset. (/dev/tenstorrent/N)
      *                          Each identifier uniquely identifies a device on the PCI bus.
      * @param flag The type of reset operation to perform on the target devices.
+     * @param ignore_failures Ignore any failures when sending reset ioctls.
      *
      * @note This is a blocking operation that may take time to complete depending
      *       on the number of devices and the reset type.
      */
-    static void reset_devices(const std::unordered_set<int> &pci_target_devices, TenstorrentResetDevice flag);
+    static void send_reset_ioctl_to_devices(
+        const std::unordered_set<int> &pci_target_devices, TenstorrentResetDevice flag, bool ignore_failures = true);
 
     /**
      * Temporary function which allows us to support both ways of mapping buffers during the transition period.
--- a/device/pcie/pci_device.cpp
+++ b/device/pcie/pci_device.cpp
@@ -197,21 +197,21 @@ PciDeviceInfo PCIDevice::read_device_info(int fd) {
         get_physical_slot_for_pcie_bdf(pci_bdf)};
 }
 
-static void reset_device_ioctl(int device_id, uint32_t flags) {
+static void send_reset_ioctl(int device_id, uint32_t flags) {
     log_debug(tt::LogUMD, "Attempting to reset PCI device ID {} with flags {}", device_id, flags);
     std::string device_path = fmt::format("/dev/tenstorrent/{}", device_id);
     tt_device_t *dev = nullptr;
     int err = tt_device_open(device_path.c_str(), &dev, O_RDWR | O_CLOEXEC | O_APPEND);
     if (err != 0) {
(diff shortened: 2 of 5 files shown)


device/pcie/pci_device.cpp
```diff
@@ -197,44 +197,21 @@ PciDeviceInfo PCIDevice::read_device_info(int fd) {
...
+        log_error(LogUMD, "Failed to open device {} on path {} with error: {}", device_id, device_path, strerror(-err));
+        return;
+    }
 
-        close(fd);
+    err = tt_device_reset(dev, flags);
+    if (err != 0) {
+        UMD_THROW(
+            error::RuntimeError,
+            fmt::format("Reset failed on device {} with flags {}: {}", device_id, flags, strerror(-err)));
```
joelsmithTT: The error handling is inconsistent: fail to open logs and returns; fail to reset throws (but is caught and prints the next function up).

device/pcie/pci_device.cpp
```diff
@@ -919,8 +896,21 @@ void PCIDevice::configure_tlb(const uint32_t tlb_index, const tlb_data &tlb_conf
...
+        // to check if the device is in the pci_target_devices set.
+        if (!pci_target_devices.empty() && pci_target_devices.find(n) == pci_target_devices.end()) {
+            continue;
+        }
+        try {
+            reset_device_ioctl(n, static_cast<uint32_t>(flag));
+        } catch (const std::exception &e) {
+            log_error(tt::LogUMD, "Reset failed: {}", e.what());
+        } catch (...) {
+            log_error(tt::LogUMD, "Reset failed with unknown error.");
```
joelsmithTT: Do we expect `reset_device_ioctl` to throw something that is not derived from `std::exception`?

device/pcie/pci_device.cpp
```diff
@@ -919,8 +896,21 @@ void PCIDevice::configure_tlb(const uint32_t tlb_index, const tlb_data &tlb_conf
         tlb_config.static_vc);
 }
 
-void PCIDevice::reset_device_ioctl(const std::unordered_set<int> &pci_target_devices, TenstorrentResetDevice flag) {
-    umd::reset_device_ioctl(pci_target_devices, static_cast<uint32_t>(flag));
+void PCIDevice::reset_devices(const std::unordered_set<int> &pci_target_devices, TenstorrentResetDevice flag) {
```
joelsmithTT: This function doesn't really do what it says on the tin (it's just a looped wrapper around the ioctl, i.e., it does not perform the full reset sequence -- that looks like it still lives in warm_reset.cpp).  It's up to you if you want to change it; "reset_devices" is not an unreasonable name here.  But I point this out because if someone looks at this file in isolation (without the context of warm_reset.cpp) it may be misleading.

device/warm_reset.cpp
```diff
@@ -204,13 +204,13 @@ bool WarmReset::warm_reset_arch_agnostic(
...
 
     if (reset_m3) {
-        PCIDevice::reset_device_ioctl(pci_device_id_set, TenstorrentResetDevice::ASIC_DMC_RESET);
+        PCIDevice::reset_devices(pci_device_id_set, TenstorrentResetDevice::ASIC_DMC_RESET);
     } else {
-        PCIDevice::reset_device_ioctl(pci_device_id_set, TenstorrentResetDevice::ASIC_RESET);
+        PCIDevice::reset_devices(pci_device_id_set, TenstorrentResetDevice::ASIC_RESET);
     }
 
     // Calculate post-reset wait time: use provided M3 timeout if M3 reset, otherwise scale based on device count
```
joelsmithTT, line 217: Out of scope for this PR; I'm pointing it out as a future opportunity -- I'd like to see the notion of "post-reset wait time" eliminated from the system.

UMD can poll the config space for the reset marker bit (see [this code](https://github.com/tenstorrent/tt-kmd/blob/0ab170d79e9196ca2468c86f429295c8fae77ade/pcie.c#L140-L158) in the KMD) to tell when (if) the reset is complete.

So:
* ASIC_RESET
* Poll reset marker bit until it's clear or we time out
* POST_RESET

device/api/umd/device/pcie/pci_device.hpp
```diff
@@ -295,9 +295,16 @@ class PCIDevice {
     static uint8_t read_command_byte(const int pci_device_num);
 
     /**
-     * Reset device via ioctl.
+     * @brief Resets the specified Tenstorrent PCIe devices.
+     *
+     * @param pci_target_devices A set of PCI device identifiers to be reset.
+     *                          Each identifier uniquely identifies a device on the PCI bus.
```
joelsmithTT: It's ambiguous (to me, at least, since my familiarity with this code has eroded over time) what "PCI device identifiers" means in this context.  Is it the N in /dev/tenstorrent/N, or something else?

joelsmithTT (approved):
Overall looks reasonable; left a couple remarks for your consideration.

device/api/umd/device/pcie/pci_device.hpp
aleksamarkovicTT: It's "N in /dev/tenstorrent/N". Is there any other interpretation, not counting BDF?

device/warm_reset.cpp
aleksamarkovicTT: Is polling the config space available through KMD or does it have to go through sysfs?

device/pcie/pci_device.cpp
aleksamarkovicTT: I'll rename it to send reset ioctl or something.

device/warm_reset.cpp
joelsmithTT: sysfs

device/pcie/pci_device.cpp
aleksamarkovicTT: I don't think so.

device/api/umd/device/pcie/pci_device.hpp
```diff
@@ -295,9 +295,18 @@ class PCIDevice {
...
+     * @param pci_target_devices A set of PCI device identifiers to be reset. (/dev/tenstorrent/N)
+     *                          Each identifier uniquely identifies a device on the PCI bus.
+     * @param flag The type of reset operation to perform on the target devices.
+     * @param ignore_failures Ignore any failures when sending reset ioctls.
+     *
+     * @note This is a blocking operation that may take time to complete depending
+     *       on the number of devices and the reset type.
      */
-    static void reset_device_ioctl(const std::unordered_set<int> &pci_target_devices, TenstorrentResetDevice flag);
+    static void send_reset_ioctl_to_devices(
```
nbuncicTT, line 308: nit: Why are we renaming the API? I mean IOCTL is self-explanatory?

nbuncicTT (approved):
Well done

device/api/umd/device/pcie/pci_device.hpp
aleksamarkovicTT: I originally wanted a method that would send one ioctl to one device, instead of looping, but due to the policy of ignoring pretty much any kind of failures with these ioctls, I more-or-less retained the structure, but kept this name as it made more sense to have "devices" in plural as it is taking an unoredered set.

PR #2854, opened 2026-06-19
Use tt_kmd_lib for setting power state and KMD API version

aleksamarkovicTT:
### Issue
#2749 

### Description
Use tt_kmd_lib for setting power state and KMD API version

### List of the changes
- Added `tt_device_set_power_state()`
- Changed `PCIDevice::set_power_state()` to use kmd-lib.
- Changed `PCIDevice` ctor to use kmd-lib for KMD API version.

### Testing
CI

### API Changes
This PR has API changes in tt_kmd_lib.

### DIFF
--- a/device/api/umd/device/tt_kmd_lib/tt_kmd_lib.h
+++ b/device/api/umd/device/tt_kmd_lib/tt_kmd_lib.h
@@ -388,6 +388,31 @@ int tt_tlb_map(tt_device_t* dev, tt_tlb_t* tlb, tt_noc_addr_config_t* config);
  */
 int tt_tlb_map_unicast(tt_device_t* dev, tt_tlb_t* tlb, uint8_t x, uint8_t y, uint64_t addr);
 
+/**
+ * @brief Power flags for use with tt_device_set_power_state().
+ */
+#define TT_POWER_FLAG_MAX_AI_CLK (1U << 0)       /**< 1=Max AI Clock,  0=Min AI Clock */
+#define TT_POWER_FLAG_MRISC_PHY_WAKEUP (1U << 1) /**< 1=PHY Wakeup,    0=PHY Powerdown */
+#define TT_POWER_FLAG_TENSIX_ENABLE (1U << 2)    /**< 1=Enable Tensix, 0=Clock Gate Tensix */
+#define TT_POWER_FLAG_L2CPU_ENABLE (1U << 3)     /**< 1=Enable L2CPU,  0=Clock Gate L2CPU */
+
+/**
+ * @brief Set per-client power state requirements for this device file descriptor.
+ *
+ * Declares the power requirements of this file descriptor to the driver.
+ * The driver aggregates requirements across all open file descriptors,
+ * applying bitwise OR for power_flags. When the file descriptor is closed,
+ * its contribution is removed.
+ *
+ * Opening the device with O_APPEND opts out of legacy mode, allowing the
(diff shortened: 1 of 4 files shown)


device/api/umd/device/tt_kmd_lib/tt_kmd_lib.h
```diff
@@ -388,6 +388,31 @@ int tt_tlb_map(tt_device_t* dev, tt_tlb_t* tlb, tt_noc_addr_config_t* config);
  */
 int tt_tlb_map_unicast(tt_device_t* dev, tt_tlb_t* tlb, uint8_t x, uint8_t y, uint64_t addr);
 
+/**
+ * @brief Power flags for use with tt_device_set_power_state().
+ */
+#define TT_POWER_FLAG_MAX_AI_CLK (1U << 0)       /**< 1=Max AI Clock,  0=Min AI Clock */
```
broskoTT: Is this copied somewhere from KMD repo? Can we reference that file (as a link in the comment, not actual git mechanism to fetch it)?

device/pcie/pci_device.cpp
```diff
@@ -1074,20 +1071,13 @@ void PCIDevice::set_power_state(bool busy) {
...
-    power_state.validity = TT_POWER_VALIDITY(4, 0);
-
-    if (busy) {
-        power_state.power_flags =
-            TT_POWER_FLAG_MRISC_PHY_WAKEUP | TT_POWER_FLAG_TENSIX_ENABLE | TT_POWER_FLAG_L2CPU_ENABLE;
-    } else {
-        power_state.power_flags = 0;
-    }
+    uint16_t power_flags =
+        busy ? TT_POWER_FLAG_MRISC_PHY_WAKEUP | TT_POWER_FLAG_TENSIX_ENABLE | TT_POWER_FLAG_L2CPU_ENABLE : 0;
```
broskoTT: Make this a constant outside of the function?

broskoTT (approved):

device/pcie/pci_device.cpp
nbuncicTT: I'd add that it should be a named type safe static constexpr constant that's derived from the macro's used here with.

device/tt_kmd_lib/tt_kmd_lib.c
```diff
@@ -514,3 +514,17 @@ int tt_tlb_map_unicast(tt_device_t* dev, tt_tlb_t* tlb, uint8_t x, uint8_t y, ui
 
     return 0;
 }
+
+int tt_device_set_power_state(tt_device_t* dev, uint16_t power_flags) {
```
nbuncicTT, line 518: I like this pattern where we wrap the system calls and the boiler-plate code and return errno from those wrapped functions, maybe it makes sense to do it for mmap and other sys calls. What do you think @aleksamarkovicTT?

No action needed in this PR.

nbuncicTT (approved):

device/tt_kmd_lib/tt_kmd_lib.c
aleksamarkovicTT: I agree, but I would not commit to it at this moment.

device/api/umd/device/tt_kmd_lib/tt_kmd_lib.h
aleksamarkovicTT: This is a copy from ioctl.h, also present in this repo. I'll restructure this as a flag enum.

PR #2856, opened 2026-06-19
Modify ETH_LIVE_STATUS-based APIs to be precise and add harvesting utility for BH

### LINKED ISSUE #2960: Make ETH status APIs core-precise and add harvesting filter for Blackhole
**Description**

The ETH status APIs (`get_eth_heartbeat_status`, `get_eth_retrain_status`, `get_eth_link_status`) return a flat `vector<bool>` indexed by channel, with no indication of which core each entry maps to. The mapping is arch-specific, the vector is always 16 entries (wrong on Blackhole's 14), and harvested cores can't be excluded. We want each status paired with its ETH core's NOC0 coordinate, sized correctly per arch, plus a helper to filter harvested cores.

**Scope**

- Return `optional<vector<pair<CoreCoord, bool>>>` from the three ETH status APIs (NOC0 coords).
- Map bits to cores via arch-specific `ETH_CORES_NOC0` (16 WH / 14 BH).
- Add `filter_harvested_eth_status()` in `eth_status_utils.hpp`.
- Update doc comments; test coord type/system, per-arch sizes, and the filter.

**Timeline**

No hard deadline, but this is an API-breaking change — better to land before more consumers build on the old shape.

**Is your feature request related to a problem?**

The status vectors don't say which core each entry is, the mapping differs per arch, Blackhole gets phantom entries, and harvested cores can't be dropped.

**Definition of Done**

- Three APIs return `optional<vector<pair<CoreCoord, bool>>>` with NOC0 ETH coords, correct per-arch sizes, `nullopt` preserved for unsupported cases.
- `filter_harvested_eth_status()` excludes harvested cores.
- Tests cover coords, sizes, and the filter; CI passes.

nbuncicTT:
### Issue
#2960 

### Description
Makes the ETH status APIs core-precise: each status is now paired with its ETH core's NOC0 coord, sized correctly per arch (16 WH / 14 BH), plus a helper to filter harvested cores.

### List of the changes
- Three ETH status APIs now return `optional<vector<pair<CoreCoord, bool>>>` instead of `vector<bool>`.
- `parse_eth_status_bitmask()` maps bits to cores via arch-specific `ETH_CORES_NOC0`.
- Added `eth_status_utils.hpp` with `filter_harvested_eth_status()`.
- Updated doc comments and tests; added `PrintEthStatus` test.

### Testing
CI runs updated `test_firmware_info_provider`: checks coord type/system, per-arch sizes, and the harvesting filter.

### API Changes
This PR has API changes.


aleksamarkovicTT (approved):
Consider merging eth_status_utils.hpp into firmware_utils.hpp

device/firmware/firmware_info_provider.cpp
```diff
@@ -671,32 +672,36 @@ std::optional<uint32_t> FirmwareInfoProvider::get_therm_trip_count() const {
...
 }
 
-/* static */ std::vector<bool> FirmwareInfoProvider::parse_eth_status_bitmask(uint16_t bitmask) {
-    static constexpr uint32_t max_eth_links = 16;
-    std::vector<bool> statuses(max_eth_links);
-    for (uint32_t link = 0; link < max_eth_links; ++link) {
-        statuses[link] = static_cast<bool>(bitmask & (1u << link));
+std::vector<std::pair<CoreCoord, bool>> FirmwareInfoProvider::parse_eth_status_bitmask(uint16_t bitmask) const {
+    const auto& eth_cores_noc0 =
+        (tt_device->get_arch() == tt::ARCH::BLACKHOLE) ? blackhole::ETH_CORES_NOC0 : wormhole::ETH_CORES_NOC0;
```
pjanevskiTT: can we pull this out of architecture implemenation if it's available somehow in this class? Through TTDevice if possible?

tests/api/test_firmware_info_provider.cpp
```diff
@@ -612,7 +613,12 @@ TEST_F(TestFirmwareInfoProvider, EthHeartbeatStatus) {
             (arch == tt::ARCH::BLACKHOLE && fw_version >= FirmwareBundleVersion(19, 9, 0))) {
             EXPECT_TRUE(heartbeats.has_value());
             if (heartbeats.has_value()) {
-                EXPECT_EQ(heartbeats.value().size(), 16u);
+                size_t expected_size = (arch == tt::ARCH::BLACKHOLE) ? 14u : 16u;
```
pjanevskiTT, line 616: can we maybe also pull these constants out of arch implementation as well?

pjanevskiTT: ok these are the tests, hard coding constants is fine

pjanevskiTT (approved):

device/firmware/firmware_info_provider.cpp
pjanevskiTT: This ternary maps every non-Blackhole arch to `wormhole::ETH_CORES_NOC0`, so QUASAR (and any future arch) would silently get the Wormhole core list. A `grendel::ETH_CORES_NOC0` already exists separately. Even if ETH status is `NotAvailable` on those archs today, consider making the selection explicit (switch on arch) or asserting on the unsupported case, so a wrong-but-plausible coordinate list can't be produced later.

device/firmware/firmware_info_provider.cpp
```diff
@@ -671,32 +672,36 @@ std::optional<uint32_t> FirmwareInfoProvider::get_therm_trip_count() const {
...
+
+    std::vector<std::pair<CoreCoord, bool>> statuses;
+    statuses.reserve(eth_cores_noc0.size());
+    for (uint32_t channel = 0; channel < eth_cores_noc0.size(); ++channel) {
+        CoreCoord noc0_coord(eth_cores_noc0[channel], CoreType::ETH, CoordSystem::NOC0);
+        statuses.emplace_back(noc0_coord, static_cast<bool>(bitmask & (1u << channel)));
```
pjanevskiTT, line 699: Worth confirming the harvesting contract here, since it's the headline of this PR: this maps bit `channel` -> physical `ETH_CORES_NOC0[channel]` for all 14/16 cores, leaving harvesting to the post-filter. That's correct only if the firmware bitmask is indexed by *physical* channel (harvested channels simply have their bit clear) rather than packing bits for active channels only. If firmware ever packs/shifts bits on a harvested part, the bit<->core mapping would be off by the harvested count. Can you confirm the firmware reports by physical channel index on BH?

tests/api/test_firmware_info_provider.cpp
```diff
@@ -655,8 +666,61 @@ TEST_F(TestFirmwareInfoProvider, EthLinkStatus) {
...
+                for (const auto& [coord, status] : links.value()) {
+                    EXPECT_EQ(coord.core_type, CoreType::ETH);
+                    EXPECT_EQ(coord.coord_system, CoordSystem::NOC0);
+                }
             }
         }
     }
 }
+
+TEST_F(TestFirmwareInfoProvider, PrintEthStatus) {
```
pjanevskiTT: This test has no assertions -- it only logs and will always pass, adding output noise on every CI run. If it's intended as a manual diagnostic, consider prefixing it `DISABLED_` (like other diagnostic-only tests) or gating it behind a manual flag; otherwise add `EXPECT_*` checks so it actually verifies something.

device/api/umd/device/firmware/eth_status_utils.hpp
```diff
@@ -0,0 +1,33 @@
...
+ * for ETH cores that have been harvested.
+ */
+inline std::vector<std::pair<CoreCoord, bool>> filter_harvested_eth_status(
+    const std::vector<std::pair<CoreCoord, bool>>& statuses, const SocDescriptor& soc_desc) {
+    auto harvested = soc_desc.get_harvested_cores(CoreType::ETH, CoordSystem::NOC0);
+
+    std::vector<std::pair<CoreCoord, bool>> filtered;
+    filtered.reserve(statuses.size());
+    std::copy_if(statuses.begin(), statuses.end(), std::back_inserter(filtered), [&](const auto& entry) {
+        return std::find(harvested.begin(), harvested.end(), entry.first) == harvested.end();
```
pjanevskiTT: Minor: `get_harvested_cores` returns a vector and this does a linear `std::find` per entry (O(n*m)). Fine for <=16 ETH cores, but a `std::unordered_set`/`std::set` of harvested coords would read cleaner and scale trivially.

device/firmware/firmware_info_provider.cpp
nbuncicTT: Confirmed with the firmware source (`tt-system-firmware`,` lib/tenstorrent/bh_arc/eth.c`). `GetEthHeartbeatStatus`/`GetEthLinkStatus` loop over all `MAX_ETH_INSTANCES` and do `status |= BIT(eth_inst)` only for enabled tiles. So the bitmask is indexed by physical channel and harvested/disabled instances just leave their bit at 0 (no packing/shifting). `telemetry.c` then packs it as `(link << 16) | heartbeat`. So bit N → ETH_CORES_NOC0[N] with a post-filter for harvested cores is correct. 

I'll add a comment documenting this.

device/firmware/firmware_info_provider.cpp
nbuncicTT: The firmware bitmask is indexed by physical channel, and ETH_CORES_NOC0 holds the physical coords in that same channel order — which is exactly what we need to keep bit N → core N aligned, then filter harvested after. 

I'll add a comment on this in the API implementation.

Issue #2159, opened 2026-02-24
HealthState returned per chip


broskoTT:
We've talked about this for quite some time.
Exalens wants to run topodiscovery and then get the health state of all the chips, such as arc/dram/eth status, eth fw hash check, etc

btrzynadlowski-tt:
Will this be exposed as an interface for telemetry to poll? I'm a bit confused about this because it seems this is needed for UMD to initialize, right? This is not something that dynamically updates?

aleksamarkovicTT:
This interface is aimed more at exalens and will only contain diagnostic information about exception that you cannot recover from without resetting/reflashing. Examples:
- CMFW mismatch
- ETH FW mismatch
- ETH core hang
- ARC startup failure
etc.

Not something that you will need to poll necessarily, more of a side effect of UMD's device initialization to report metadata on these situations in an accessible way instead of logging strings.

PR #2866, opened 2026-06-22
Return device health related errors in TopologyDiscovery

aleksamarkovicTT:
### Issue
#2159 

### Description
Return device health related errors that occured in TopologyDiscovery. The errors are accessible through `ClusterDescriptor::get_health_errors()`.

### List of the changes
- Added device errors variant and `get_health_errors()` in `ClusterDescriptor`.
- Capture and persist errors when calling `init_tt_device()` in TopologyDiscovery.
- Added `HangDetectionTest.TopologyDiscoveryRecordsNocHangHealthError`, disabled by default.
- Fix `PcieHangError` not using `PcieHangData` struct for error metadata.

### Testing
Added disabled test that verifies health errors are returned.

### API Changes
This PR has API changes.


tests/api/test_hang_detection.cpp
```diff
@@ -170,3 +173,50 @@ INSTANTIATE_TEST_SUITE_P(
...
+        }
+    }
+    EXPECT_TRUE(found_noc_hang) << "Expected a NocHangError in cluster descriptor health errors.";
+
+    // Release discovery's device handles, then warm reset to recover the hung device so the host
+    // and subsequent tests are left in a good state.
+    cluster_desc.reset();
+    devices.clear();
+    WarmResetWithRecovery::warm_reset();
+    init_device(pci_device_id);
```
pjanevskiTT, line 221: why do we need this line?

device/api/umd/device/topology/topology_discovery.hpp
```diff
@@ -149,6 +149,8 @@ class TopologyDiscovery {
     // It's required to know which chip should be used for remote communication.
     std::map<uint64_t, uint64_t> remote_asic_id_to_mmio_device_id;
 
+    std::map<uint64_t, std::vector<ClusterDescriptor::DeviceHealthError>> health_errors;
```
pjanevskiTT, line 152: this stores vector of errors, but seems to me that you only store at most one error per device. Maybe this is not true, just what I got from reading the code

pjanevskiTT (approved):

device/api/umd/device/topology/topology_discovery.hpp
aleksamarkovicTT: It's true for this PR, but with #2900 and future PRs there will be multiple errors.

tests/api/test_hang_detection.cpp
aleksamarkovicTT: It calls init_tt_device(), which does a hang check again. If, for some reason, the NOC was still hung, it would fail the test.

aleksamarkovicTT: Have in mind it's init_device from the test suite.

PR #2869, opened 2026-06-22
ci: run smoke matrix on PR, full sweep on merge_group

blozano-tt:
## Problem

The `build-device.yml` workflow runs a full 10-leg build matrix on every PR push/rebase, costing ~10 builds × ~10 min = ~100 CPU-runner-minutes per trigger. A single engineer rebasing two branches spins up 20 parallel jobs.

The repo already has a merge queue (`merge_group` trigger, ruleset 739206). The full sweep should run there — not on every push.

## Change

- **PR / `pull_request` event**: runs a 2-leg smoke matrix
  - `ubuntu-24.04` + `clang-20`
  - `fedora-39` + `gcc-13`
- **Merge queue / `merge_group`, `push: main`, `workflow_dispatch`**: runs the full 10-leg matrix (unchanged)

Adds a `ci-complete` aggregator job that reports a single pass/fail status — intended to replace the per-matrix required-check entries in branch protection ruleset 739206 in a follow-up.

## Savings

~80 CPU-runner-minutes saved per PR push/rebase (10 legs → 2 legs on PR).

## Follow-up required

Branch protection ruleset 739206 still pins per-matrix job names as required checks. A separate admin step is needed to replace those with the `ci-complete` aggregator. Instructions in the ruleset edit section of the linked Slack thread.

## Implementation notes

- Event-based matrix selection is done via a small `generate-matrix` job that emits two JSON arrays (`build-matrix`, `test-matrix`); `build` and `test-packages` consume them with `fromJSON`. This avoids duplicating the matrix YAML. The full 10-leg `build-matrix` is reproduced verbatim from the previous inline matrix (no entries added/removed/changed).
- `test-packages` is also event-aware so its smoke legs match the `PLATFORM_SUFFIX` (distro + compiler) of the packages the smoke `build` legs actually produce (`ubuntu-24.04-clang-20`, `fedora-39-gcc-13`); otherwise its artifact downloads would 404 on PRs and the aggregator would always fail.
- The `on:` triggers, reusable `docker-image.yml` calls, manylinux wheel job, and all step contents are unchanged.
- The `ci-complete` job uses `if: always()` plus an explicit failing step on `contains(needs.*.result, 'failure'/'cancelled')`, so a skipped/failed dependency produces a real failed status (a single gated success step alone would report success when skipped).

### DIFF
--- a/.github/workflows/build-device.yml
+++ b/.github/workflows/build-device.yml
@@ -47,15 +47,56 @@ jobs:
       - id: set-matrix
         shell: bash
         run: |
+          # The matrices are authored as readable multi-line JSON heredocs and
+          # compacted to a single line with `jq -c` before being written to
+          # GITHUB_OUTPUT. This keeps every YAML line within yamllint's 120-char
+          # limit while emitting the same compact JSON the build job consumes.
           if [ "${{ github.event_name }}" = "pull_request" ]; then
             echo "Selecting 2-leg smoke matrix for pull_request event."
-            echo 'build-matrix={"image":[{"distro":"ubuntu-24.04","compiler-c":"clang-20","compiler-cpp":"clang++-20","package":"deb"},{"distro":"fedora-39","compiler-c":"gcc-13","compiler-cpp":"g++-13","package":"rpm"}]}' >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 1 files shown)


broskoTT:
Compute is usually cheap, that's why we haven't optimized this out of PR gate.
The other job which builds then runs tests on actual silicon is always slower, and runs in parallel to this one, so this won't speed up anything.
The jobs in this matrix run for 5-6mins, not 10mins. But in total then end up taking ~80mins of compute machine time.
We can build the whole matrix for merge gate only, makes sense.

I'll take another look

broskoTT:
Deprecated by https://github.com/tenstorrent/tt-umd/pull/2942

PR #2870, opened 2026-06-22
ci: switch build-device.yml to free GitHub-hosted runners

blozano-tt:
## Change

Migrates `build-device.yml` from Tenstorrent self-hosted runners to free GitHub-hosted runners, and removes the internal Harbor mirror dependency that came with them.

### Before → After

| Location | Before | After |
|---|---|---|
| `build` job `runs-on` | `tt-ubuntu-2204-large-stable` | `ubuntu-22.04` |
| `build-manylinux-wheels` `runs-on` | `tt-ubuntu-2204-large-stable` | `ubuntu-22.04` |
| `build` job `container.image` | Ternary: Harbor mirror for x86, ghcr.io for ARM | `ghcr.io` directly (unconditional) |

ARM builds (`ubuntu-24.04-arm`) and `test-packages` were already on GitHub-hosted runners — no change needed there.

## Why

GitHub-hosted runners are free for public repos. Removing the self-hosted runner dependency eliminates the per-runner cost concern for PR builds and removes a hard dependency on internal Tenstorrent infrastructure (Harbor pull-through cache, `tt-ubuntu-2204-large-stable` runner fleet).

Alternative to the smoke-matrix approach in #2869 — different tradeoff (free runners vs. fewer legs).

## Notes

- No matrix entries changed, no step contents changed, no image tags or digests changed.
- The Harbor pull-through cache (`harbor.ci.tenstorrent.net/ghcr.io`) is no longer referenced anywhere in this file.
- `build-device.yml` is the only file changed (+4/−12).


broskoTT:
I also changed more jobs and increased timeouts a bit

broskoTT (approved):
I've put a review, but somebody else should also review since this PR now has my changes

nbuncicTT (approved):

PR #2900, opened 2026-06-23
Add CMFWMismatchError and UnsupportedCMFWError to TopologyDiscovery health errors

aleksamarkovicTT:
### Issue
#2159 

### Description
Add CMFWMismatchError and UnsupportedCMFWError to TopologyDiscovery health errors.

### List of the changes
- Add CMFWMismatchError and UnsupportedCMFWError to TopologyDiscovery health errors.

### Testing
CI

### API Changes
This PR has API changes.


device/api/umd/device/topology/topology_discovery.hpp
```diff
@@ -134,6 +134,9 @@ class TopologyDiscovery {
     std::map<uint64_t, std::unique_ptr<TTDevice>> devices;
     std::unordered_map<uint64_t, ChipId> asic_id_to_chip_id;
 
+    // Caches the result of get_asic_id() per device to avoid repeated queries.
+    std::unordered_map<TTDevice*, uint64_t> asic_id_cache;
```
pjanevskiTT: can this work be separate PR from exceptions?

device/topology/topology_discovery.cpp
```diff
@@ -533,38 +534,45 @@ uint64_t TopologyDiscovery::get_asic_id(TTDevice* tt_device) {
...
     // Using board ID should happen only for unconnected boards (N150, P150).
+    auto cache_it = asic_id_cache.find(tt_device);
+    if (cache_it != asic_id_cache.end()) {
+        return cache_it->second;
+    }
+
     const SocDescriptor& soc_desc = tt_device->get_soc_descriptor();
     std::vector<CoreCoord> eth_cores = soc_desc.get_cores(CoreType::ETH);
 
+    uint64_t asic_id = get_unconnected_device_id(tt_device);
```
pjanevskiTT: this code was previously executing conditionally and now we always get this ID. Seems like we would still like to have it execute just if no ETH cores are trained. Can you change the code to match that behavior but still store asic id to cache. I see multiple ways to do this, you can choose how to do it

pjanevskiTT (approved):

device/topology/topology_discovery.cpp
aleksamarkovicTT: > Seems like we would still like to have it execute just if no ETH cores are trained.
Why? I don't think the code in `get_unconnected_device_id()` is expensive, it always reads from ARC telemetry.

The code is a bit peculiar, get_local_asic_id() should work even if the ETH cores are not trained (by the time this is called, they are), as this information is given to ETH by ARC on startup.

device/api/umd/device/topology/topology_discovery.hpp
aleksamarkovicTT: Sure

PR #2902, opened 2026-06-23
Simulation: add SimulationClient, the slim client-facing host handle

pjanevskiTT:
### Issue
Part of #2792. Stacked on #2940 (renames `SimulationSocket` → `SimulationServerSocket`); builds on the merged `SimulationConnector`/socket work.

### Description
Adds `SimulationClient` — the slim, always-built client-facing handle to a simulation host, reached over its per-chip UNIX socket. `attach()` connects, `detach()` closes; no request protocol yet. Grows operation-by-operation as the server work lands, at which point `TTSimTTDevice`'s client-mode dispatch is rebound onto these calls. The asio transport is held behind a pImpl so the public header stays asio-free (mirroring `SimulationServerSocket`).

### Testing
`SimulationClientTest` (3 tests) passes in the no-card baremetal lane; build + pre-commit clean.

### API Changes
New public header `umd/device/simulation/simulation_client.hpp` (`SimulationClient`).

### DIFF
--- a/device/CMakeLists.txt
+++ b/device/CMakeLists.txt
@@ -141,6 +141,7 @@ target_sources(
         # so always built. This lets its unit tests run in the no-card baremetal CI lane
         # (TT_UMD_BUILD_SIMULATION=OFF).
         simulation/simulation_server_socket.cpp
+        simulation/simulation_client.cpp
 )
 
 if(TT_UMD_BUILD_SIMULATION)
@@ -252,6 +253,7 @@ if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.23)
                 api/umd/device/types/risc_type.hpp
                 api/umd/device/simulation/simulation_chip.hpp
                 api/umd/device/simulation/simulation_connector.hpp
+                api/umd/device/simulation/simulation_client.hpp
                 api/umd/device/simulation/simulation_host.hpp
                 api/umd/device/simulation/tt_sim_communicator.hpp
                 api/umd/device/simulation/rtl_sim_communicator.hpp
--- a/device/api/umd/device/simulation/simulation_client.hpp
+++ b/device/api/umd/device/simulation/simulation_client.hpp
@@ -1,0 +1,44 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <filesystem>
+#include <memory>
+
+namespace tt::umd {
+
+// The client-facing handle to a simulation host ("the card"), reached over its per-chip UNIX
+// socket. Deliberately slim for now: just the session handshake -- attach() connects, detach()
+// closes. It grows operation-by-operation as the server work lands (device description, memory
+// access, TLB, sysmem, run/reset, ...), at which point TTSimTTDevice's client-mode dispatch is
+// rebound from its throwing stubs onto these calls. No simulation build needed (asio is a core
+// dependency), mirroring SimulationSocket on the host side.
+//
+// The asio transport is held behind a pImpl so this header stays free of asio (a private
+// dependency of the library) -- see Impl in the .cpp.
+class SimulationClient {
+public:
+    explicit SimulationClient(std::filesystem::path socket_path);
+    ~SimulationClient();
+
+    SimulationClient(const SimulationClient&) = delete;
+    SimulationClient& operator=(const SimulationClient&) = delete;
+
+    // Connects to the host socket; throws if no live host is reachable at the path. Idempotent
+    // once connected.
+    void attach();
+    // Closes the connection; idempotent.
+    void detach();
+
+private:
+    std::filesystem::path socket_path_;
+
+    // Holds the asio transport (io_context + connected socket). Defined in the .cpp so asio
+    // stays out of this header; owns the connection fd (RAII).
+    struct Impl;
+    std::unique_ptr<Impl> impl_;
+};
+
+}  // namespace tt::umd
--- a/device/simulation/simulation_client.cpp
+++ b/device/simulation/simulation_client.cpp
@@ -1,0 +1,77 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "umd/device/simulation/simulation_client.hpp"
+
+#include <fmt/format.h>
+#include <sys/un.h>  // sockaddr_un::sun_path, for the path-length guard
+
+#include <asio.hpp>
+#include <system_error>
+
+#include "umd/device/utils/error.hpp"
+
+namespace tt::umd {
+
+namespace {
+
+using stream_protocol = asio::local::stream_protocol;
+
+// Builds an endpoint for a pathname socket, throwing if the path is too long for sockaddr_un.
+// asio's endpoint(path) would itself throw on overflow, but with a less actionable message.
+// Named distinctly from SimulationServerSocket's equivalent helper: both files are always built
+// and share a unity-build translation unit, where two anonymous-namespace make_endpoint()s
+// would collide.
+stream_protocol::endpoint make_client_endpoint(const std::filesystem::path& path) {
+    if (path.string().size() >= sizeof(sockaddr_un::sun_path)) {
+        UMD_THROW(
+            error::RuntimeError,
+            fmt::format(
+                "Simulation host socket path is too long ({} >= {}): {}",
+                path.string().size(),
+                sizeof(sockaddr_un::sun_path),
+                path.string()));
+    }
+    return stream_protocol::endpoint(path.string());
+}
+
+}  // namespace
+
+// The asio transport, kept out of the header (see SimulationClient::Impl forward declaration).
+struct SimulationClient::Impl {
+    asio::io_context io;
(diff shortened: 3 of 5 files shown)


tests/baremetal/test_simulation_server_api.cpp
```diff
@@ -0,0 +1,54 @@
...
+    return std::filesystem::temp_directory_path() /
+           ("tt-umd-sim-server-api-test-" + std::to_string(::getpid()) + "-" + std::to_string(tag) + ".sock");
+}
+
+}  // namespace
+
+// attach() connects to a live host's socket; detach() closes and is idempotent.
+TEST(SimulationServerApi, AttachesToLiveHostThenDetaches) {
+    const std::filesystem::path path = unique_socket_path(1);
+    SimulationSocket server(path);  // Binds + listens => a live host at path.
```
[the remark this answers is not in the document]
pjanevskiTT: Done — `unique_socket_path()` now `std::filesystem::remove(path)`s before returning, so bind/connect start from a clean path even after a crashed run with PID reuse, mirroring the SimulationSocket tests.

tests/baremetal/test_simulation_server_api.cpp
```diff
@@ -0,0 +1,54 @@
...
+    EXPECT_NO_THROW(api.detach());
+    EXPECT_NO_THROW(api.detach());  // Idempotent.
+}
+
+// With no host bound at the path, attach() fails loudly rather than silently producing a
+// half-connected client.
+TEST(SimulationServerApi, AttachThrowsWhenNoHost) {
+    const std::filesystem::path path = unique_socket_path(2);
+
+    SimulationServerApi api(path);
```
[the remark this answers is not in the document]
pjanevskiTT: Done — the helper removes any pre-existing socket file at the path before each test uses it.

tests/baremetal/test_simulation_server_api.cpp
```diff
@@ -0,0 +1,54 @@
...
+    EXPECT_THROW(api.attach(), std::exception);
+}
+
+// The destructor releases the connection even if detach() was never called.
+TEST(SimulationServerApi, DestructorDetaches) {
+    const std::filesystem::path path = unique_socket_path(3);
+    SimulationSocket server(path);
```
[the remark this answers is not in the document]
pjanevskiTT: Done — covered by the `std::filesystem::remove(path)` now done in `unique_socket_path()` for every case.

device/api/umd/device/simulation/simulation_server_api.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+namespace tt::umd {
+
+// The client-facing handle to a simulation host ("the card"), reached over its per-chip UNIX
+// socket. Deliberately slim for now: just the session handshake -- attach() connects, detach()
+// closes. It grows operation-by-operation as the server work lands (device description, memory
+// access, TLB, sysmem, run/reset, ...), at which point TTSimTTDevice's client-mode dispatch is
+// rebound from its throwing stubs onto these calls. Pure POSIX (no simulation build needed),
+// mirroring SimulationSocket on the host side.
+class SimulationServerApi {
```
nbuncicTT: Do you think it makes sense to rename this to `SimluationClient`, as this does not act a server, but a client of the server that it should connect to?

device/api/umd/device/simulation/simulation_server_api.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+    // Connects to the host socket; throws if no live host is reachable at the path. Idempotent
+    // once connected.
+    void attach();
+    // Closes the connection; idempotent.
+    void detach();
+
+private:
+    std::filesystem::path socket_path_;
+    int fd_ = -1;
```
nbuncicTT: Instead of relying on raw file descriptors, let's use ASIO, so the proposal would be:
```cpp
asio::local::stream_protocol::socket socket_;
```

device/simulation/simulation_server_api.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "umd/device/utils/error.hpp"
+
+namespace tt::umd {
+
+SimulationServerApi::SimulationServerApi(std::filesystem::path socket_path) : socket_path_(std::move(socket_path)) {}
+
+SimulationServerApi::~SimulationServerApi() { detach(); }
+
+void SimulationServerApi::attach() {
+    if (fd_ >= 0) {
```
nbuncicTT: In ASIO this would be a query to:
```cpp
socket_.is_open()
```

device/simulation/simulation_server_api.cpp
```diff
@@ -0,0 +1,67 @@
...
+        return;  // Already attached.
+    }
+
+    sockaddr_un addr{};
```
nbuncicTT: The rest of the code code in the function could be replaced by:
```cpp
    if (socket_path_.string().size() >= max_path_size) {
        UMD_THROW(
            error::RuntimeError,
            fmt::format("Simulation host socket path is too long ({} >= {}): {}",
                        max_path_size,
                        socket_path_.string().size(), 
                        socket_path_.string()));
    }

    asio::local::stream_protocol::endpoint endpoint(socket_path_.string());

    asio::error_code ec;
    socket_.connect(endpoint, ec);

    if (ec) {
        UMD_THROW(
            error::RuntimeError,
            fmt::format("Failed to connect to simulation host socket at {}: {}", 
                        socket_path_.string(), 
                        ec.message()));
    }
```

You'd just have to define `max_path_size` (in the .cpp file) as:
```cpp
#include <sys/un.h>
static constexpr std::size_t max_path_size = sizeof(sockaddr_un::sun_path);
```

device/simulation/simulation_server_api.cpp
```diff
@@ -0,0 +1,67 @@
...
+            error::RuntimeError,
+            fmt::format(
+                "Failed to connect to simulation host socket at {}: {}", socket_path_.string(), std::strerror(err)));
+    }
+
+    fd_ = fd;
+}
+
+void SimulationServerApi::detach() {
+    if (fd_ >= 0) {
```
nbuncicTT: This would also become:
```cpp
void SimulationServerApi::detach() {
    if (!socket_.is_open()) {
        return;
    }

    asio::error_code ec;
    // ec suppresses exceptions; intentionally ignored to ensure close() runs even if shutdown() fails
    socket_.shutdown(asio::socket_base::shutdown_both, ec);
    // ec suppresses exceptions; guarantees non-throwing file descriptor release
    socket_.close(ec);
}
```

tests/baremetal/test_simulation_server_api.cpp
```diff
@@ -0,0 +1,59 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#include <gtest/gtest.h>
+#include <unistd.h>
+
+#include <filesystem>
+#include <string>
+
+#include "umd/device/simulation/simulation_server_api.hpp"
+#include "umd/device/simulation/simulation_socket.hpp"
```
nbuncicTT: Maybe the merged PR for the entry point of the server where the naming says `SimulationSocket` should be renamed to `SimulationServerSocket`, as it is the socket to the server (and this is the client itself in this PR).

tests/baremetal/test_simulation_server_api.cpp
```diff
@@ -0,0 +1,59 @@
...
+    // connect() start from a clean path, mirroring the SimulationSocket tests.
+    std::filesystem::remove(path);
+    return path;
+}
+
+}  // namespace
+
+// attach() connects to a live host's socket; detach() closes and is idempotent.
+TEST(SimulationServerApi, AttachesToLiveHostThenDetaches) {
+    const std::filesystem::path path = unique_socket_path(1);
```
nbuncicTT: Copy-pasting my prior comment from https://github.com/tenstorrent/tt-umd/pull/2800 on the tests here:
Consider using a TEST_F fixture for the repeated setup/teardown pattern. Every test does unique_socket_path(N) at the start. Something like:
```cpp
class SimulationSocketTest : public ::testing::Test {
protected:
    void SetUp() override {
        path_ = unique_socket_path();
        std::filesystem::remove(path_);
    }
    void set_path(std::filesystem::path new_path) {
        std::filesystem::remove(path_);
        path_ = std::move(new_path);
    }
    void TearDown() override {
        std::filesystem::remove(path_);
    }
    std::filesystem::path path_ = 
        std::filesystem::temp_directory_path() / 
        ("tt-umd-sim-client-test-" + std::to_string(::getpid()) + ".sock");
};
```
Also, the incrementing tag parameter doesn't serve a purpose because the path is removed after every test, these aren't multithreaded/multiprocess tests, and the PID already handles parallel test binaries. unique_socket_path can just use a fixed name.

tests/baremetal/test_simulation_server_api.cpp
```diff
@@ -0,0 +1,59 @@
...
+    SimulationSocket server(path);  // Binds + listens => a live host at path.
+
+    SimulationServerApi api(path);
```
nbuncicTT: if `SimulationServerApi` becomes `SimulationClient` I'd rename variable from `api` to `client`.

nbuncicTT (approved):
Look if you can re-use the test fixture from https://github.com/tenstorrent/tt-umd/pull/2800 once you use ASIO here.

PR #2904, opened 2026-06-23
Introduce EthFirmwareHeartbeatError, UnexpectedRoutingFirmwareConfigError and EthFirmwareMismatchError

aleksamarkovicTT:
### Issue
#2159 

### Description
Introduce EthFirmwareHeartbeatError, UnexpectedRoutingFirmwareConfigError and EthFirmwareMismatchError.

### List of the changes
- Introduce Introduce `EthFirmwareHeartbeatError`, `UnexpectedRoutingFirmwareConfigError` and `EthFirmwareMismatchError`.
- Some private TopologyDiscovery methods now take asic_id in order to easily map state that needs to be persisted per device.
- Merged `verify_eth_core_fw_version()` from two arch specific methods into one universal method.
- Related Python bindings.

### Testing
CI

### API Changes
This PR has API changes.


device/api/umd/device/topology/topology_discovery.hpp
```diff
@@ -82,7 +82,7 @@ class TopologyDiscovery {
 
     virtual bool is_eth_port_disabled(TTDevice* tt_device, CoreCoord eth_core) { return false; }
 
-    virtual bool eth_heartbeat_running(TTDevice* tt_device, CoreCoord eth_core);
+    virtual bool eth_heartbeat_running(TTDevice* tt_device, uint64_t asic_id, CoreCoord eth_core);
```
pjanevskiTT, line 85: are we passing this new argument just so we can form an exception? If so, I don't think it's needed. Can we pull out asic ID from TTDevice somehow? Or is that not the same ID that we would get in chip info?

pjanevskiTT: alternatively, this function can return bool if we should throw the error and then we can throw the error at call site, if we want to avoid passing asic_id? If these are not options for some reason it's fine to stay like this

pjanevskiTT (approved):

device/api/umd/device/topology/topology_discovery.hpp
aleksamarkovicTT: It's not for forming an exception, it's for associating an error with a TTDevice. This is currently the least invasive way to do it, but in the future, changes will be made to associate a TTDevice with asic ID more easily. Returning a bool is not ideal, as these methods could handle mutliple/different errors.

PR #2910, opened 2026-06-24
Exclusive / non-blocking device-open policy (#2841)

### LINKED ISSUE #2841: Adopt exclusive / non-blocking device-open policy to match KMD O_EXCL support
**Description**

[tt-kmd#241](https://github.com/tenstorrent/tt-kmd/pull/241) adds `O_EXCL` exclusive-open semantics to the `/dev/tenstorrent/N` chardev so that tools like tt-flash can hold the device exclusively across a flash+reset. UMD currently opens the device with a fixed `O_RDWR | O_CLOEXEC` and has no way to (a) request exclusive ownership, or (b) coexist gracefully when another process holds the device exclusively. We should update UMD's open policy to use the new KMD contract.

KMD contract from #241:

| open flags | behavior |
|---|---|
| `O_RDWR` | `-EBUSY` while another fd holds `O_EXCL`; otherwise unchanged |
| `O_RDWR \| O_EXCL` | blocks (interruptible) until the device is idle, then takes exclusive ownership |
| `O_RDWR \| O_EXCL \| O_NONBLOCK` | `-EAGAIN` if not immediately acquirable |

**Scope**

Update `open_pci_device` in `device/pcie/pci_device.cpp` to a new policy:

1. Add an opt-in `exclusive` parameter (default `false`) that adds `O_EXCL` to the open flags.
2. Attempt the open non-blocking first (`O_RDWR | O_CLOEXEC [| O_EXCL] | O_NONBLOCK`); stay silent on the success fast path.
3. If the attempt fails with `EAGAIN` or `EBUSY`, treat both identically: log a warning that the device is in use and we are waiting, then re-open **without** `O_NONBLOCK` (blocking, interruptible) and wait.
4. For any other `errno`, fail fast (throw) — no retry.

**Timeline**

Tied to tt-flash's need for exclusive flash+reset; gated on tt-kmd#241 landing.

Without this, UMD cannot request exclusive ownership of a device, and a normal UMD open during a tt-flash run fails with an opaque error instead of a clear "device held exclusively" message.

**Definition of Done**

- `open_pci_device` implements the non-blocking-then-blocking policy above with the `exclusive` opt-in.
- `EAGAIN`/`EBUSY` are handled identically (log + wait); all other errors fail fast.
- A clear, user-facing log message is emitted only when the open actually has to wait.

pjanevskiTT:
### Issue
Closes #2841. [tt-kmd#241](https://github.com/tenstorrent/tt-kmd/pull/241) adds `O_EXCL` semantics to `/dev/tenstorrent/N` so a tool can hold a device across a flash + reset. UMD opened with a fixed `O_RDWR | O_CLOEXEC` and could neither request exclusive ownership nor coexist with a process holding one.

### Description
`open_pci_device` follows the new KMD contract: try non-blocking first and stay silent on success; on `EAGAIN`/`EBUSY` log one warning and re-open blocking (interruptible) to wait for ownership; any other `errno`, including a failed blocking re-open, throws instead of returning `-1`. An opt-in `exclusive` flag adds `O_EXCL` and is plumbed through topology-discovery options, following the existing `use_safe_api` pattern. Every new parameter defaults to today's behaviour.

### List of the changes
- `pci_device.cpp`: the non-blocking-then-blocking policy plus an `exclusive` parameter; `O_NONBLOCK` is cleared on success so both paths return identically-flagged descriptors; `read_command_byte` wraps the open to keep its device-number context against the new throw contract.
- `TopologyDiscoveryOptions::open_devices_exclusively`, forwarded through `TTDevice::create` down to `PCIDevice`.
- Python bindings for both.

### Testing
CI.

### API Changes
Additive, all defaulted: `TopologyDiscoveryOptions::open_devices_exclusively`, a trailing `bool exclusive = false` on `TTDevice::create` and `PCIDevice(int)`, and the corresponding Python arguments.

### DIFF
--- a/device/api/umd/device/pcie/pci_device.hpp
+++ b/device/api/umd/device/pcie/pci_device.hpp
@@ -162,8 +162,12 @@ public:
      * sysfs, and maps device memory region(s) into the process address space.
      *
      * @param pci_device_number     N in /dev/tenstorrent/N
+     * @param exclusive             When true, open the device with O_EXCL to take exclusive
+     *                              ownership (waiting, interruptibly, until the device is idle).
+     *                              When false (default), open shared but wait if another process
+     *                              currently holds the device exclusively.
      */
-    PCIDevice(int pci_device_number);
+    PCIDevice(int pci_device_number, bool exclusive = false);
 
     /**
      * PCIDevice destructor.
--- a/device/api/umd/device/topology/topology_discovery_options.hpp
+++ b/device/api/umd/device/topology/topology_discovery_options.hpp
@@ -95,5 +95,13 @@ struct TopologyDiscoveryOptions {
      * Defaults to false.
      */
     bool use_safe_api = false;
+
+    /**
+     * @brief If true, devices are opened exclusively (O_EXCL), taking exclusive ownership and
+     * waiting (interruptibly) until any current owner releases the device. Matches the KMD
+     * exclusive-open contract (tt-kmd#241) used by tools such as tt-flash for flash + reset.
+     * Defaults to false.
+     */
+    bool open_devices_exclusively = false;
 };
 }  // namespace tt::umd
--- a/device/api/umd/device/tt_device/blackhole_tt_device.hpp
+++ b/device/api/umd/device/tt_device/blackhole_tt_device.hpp
@@ -80,7 +80,8 @@ private:
         int device_number,
         IODeviceType device_type,
         bool use_safe_api,
-        const std::shared_ptr<SocArchDescriptor> &soc_arch_descriptor);
+        const std::shared_ptr<SocArchDescriptor> &soc_arch_descriptor,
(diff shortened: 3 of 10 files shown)


pjanevskiTT:
Draft pending local testing with the new KMD: the `O_EXCL` / `EBUSY` / `EAGAIN` behavior this relies on only exists once [tt-kmd#241](https://github.com/tenstorrent/tt-kmd/pull/241) is loaded, so this needs to be validated on hardware against that KMD before it leaves draft.

nbuncicTT (commented):
This is not a feature UMD should be aware of on the top most level. 

Only low level utilities (like KMD lib and a few classes) that are needed for porting `tt-flash` from luwen to UMD should be aware, and certainly not `TTDevice`, `TopologyDiscovery`, etc.

Issue #2865, opened 2026-06-22
Investigate a common TTDevice implementation for RTL and tt_sim simulations


pjanevskiTT:
Raised in [#2704 (review)](https://github.com/tenstorrent/tt-umd/pull/2704#discussion_r3401862953).

`RtlSimulationTTDevice` and `TTSimTTDevice` currently duplicate some logic (e.g. the per-core unicast fallback for `noc_multicast_write`). We should investigate whether it makes sense to factor the shared behavior into a common base/implementation that both simulation devices reuse — and whether the two are similar enough for that to be worthwhile.

pjanevskiTT:
While adding client-mode (socket attach/detach + lifecycle dispatch) to both simulation devices in #2868, `TTSimTTDevice` and `RtlSimulationTTDevice` ended up duplicating a fair amount. Concrete candidates for a shared simulation-device base:

- the client-mode machinery: `server_api_`, the `init_`/`teardown_` lifecycle dispatch, `create_client(...)`, the client constructor, and the best-effort destructor teardown;
- `adopt_socket()` + the `socket_` member;
- the SoC-descriptor-from-simulator-directory construction used in the factories;
- the pile of identical "not supported" throwers (`dma_*`, `arc_*`, `get_clock`, and the `noc_multicast_write` unicast fallback that originally raised this issue);
- `read`/`write` share the same dispatch shape — only the backend communicator type and the host-mode bodies differ.

What stays per-class: the concrete communicator type (`TTSimCommunicator` vs `RtlSimCommunicator`) and the host-mode read/write/reset bodies that call backend-specific communicator methods.

So a common base could host the client-mode + lifecycle + socket + throwers, with the two backends overriding only their host-mode I/O.

pjanevskiTT:
Adding another duplication candidate from #2868: both `RtlSimulationTTDevice` and `TTSimTTDevice` now carry identical **client-mode machinery** for attaching to a live host over `SimulationClient`:
- the `setup_`/`teardown_` lifecycle function-pair (bound in the constructor) and the `client_` member,
- the `if (client_) throw ...` guards in `read_from_device`/`write_to_device`/`get_io_window`,
- the best-effort (try/catch) `teardown_` in the destructor.

[...]

PR #2914, opened 2026-06-24
Move common simulation device methods into SimulationTTDevice

pjanevskiTT:
### Issue
Part of #2865

### Description
Second step toward a common implementation for the RTL and TTSim simulation `TTDevice` backends, stacked on #2913. Moves the behavior that is identical across `TTSimTTDevice` and `RtlSimulationTTDevice` up into the `SimulationTTDevice` base. Backend-divergent behavior stays in the derived classes and is unified in follow-up PRs.

### List of the changes
- Move into `SimulationTTDevice`: the DMA and ARC APB/CSM throw-stubs, `get_clock`, `get_min_clock_freq`, `dma_multicast_write`, `retrain_dram_core`, both `noc_multicast_write` overloads (grid form delegates to `multicast_write_via_unicast`), `get_noc_translation_enabled`, the `get_sysmem_manager`/`get_tlb_allocator` accessors, and `adopt_socket` together with the `socket_` member.
- Define the base constructor and destructor out-of-line so the forward-declared `SimulationServerSocket` only needs to be complete in the implementation file.
- Unify the "not supported" error messages and remove the now-redundant overrides from the derived classes.

### Testing
CI

### API Changes
None.

### DIFF
--- a/device/CMakeLists.txt
+++ b/device/CMakeLists.txt
@@ -151,6 +151,7 @@ if(TT_UMD_BUILD_SIMULATION)
             simulation/simulation_host.cpp
             simulation/tt_sim_communicator.cpp
             simulation/rtl_sim_communicator.cpp
+            tt_device/simulation_tt_device.cpp
             tt_device/rtl_simulation_tt_device.cpp
             tt_device/tt_sim_tt_device.cpp
             chip_helpers/simulation_tlb_allocator.cpp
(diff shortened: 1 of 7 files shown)


nbuncicTT (approved):
As discussed offline, for now it's ok, just take in mind that SimulationTTDevice should maybe override the common interface from TTDevice, and other components should depend on the non-TTDevice APIs from the newly added class, so new simulation classes can use SimulationTTDevice without the burden of the socket stuff, and other classes can re-use the socket stuff without the burden of TTDevice simulation behavior.

No action needed, just a comment and well done.

ajovanovicTT (approved):

PR #2916, opened 2026-06-24
Unify simulation get_io_window behind a create_tlb_window hook

pjanevskiTT:
### Issue
Part of #2865

### Description
Third step toward a common implementation for the RTL and TTSim simulation `TTDevice` backends, stacked on #2914. This PR unifies the TLB window setup. The `get_io_window` logic (TLB index allocation, size resolution, QUASAR bypass) and the constructor TLB helpers move into `SimulationTTDevice`; the only backend-specific part — building the concrete `TlbHandle`/`TlbWindow` — is delegated to a new `create_tlb_window` hook.

### List of the changes
- Move `get_io_window` into `SimulationTTDevice`, delegating handle/window construction to a new pure-virtual `create_tlb_window` hook.
- Add `init_tlb_allocator(bar0_base)` and `setup_cached_tlb_window()` helpers to the base; derived constructors call these instead of duplicating the allocator setup and the per-arch cached-window switch.
- Implement `create_tlb_window` in `TTSimTTDevice` (passes the communicator to the handle) and `RtlSimulationTTDevice`.

### Testing
CI

### API Changes
None.

### DIFF
--- a/device/api/umd/device/tt_device/simulation_tt_device.hpp
+++ b/device/api/umd/device/tt_device/simulation_tt_device.hpp
@@ -14,6 +14,7 @@
 #include "umd/device/chip_helpers/simulation_tlb_allocator.hpp"
 #include "umd/device/pcie/tlb_window.hpp"
 #include "umd/device/tt_device/tt_device.hpp"
+#include "umd/device/types/tlb.hpp"
 #include "umd/device/types/xy_pair.hpp"
 
 namespace tt::umd {
@@ -23,13 +24,10 @@ class SimulationTlbAllocator;
 class SimulationServerSocket;
 class TlbWindow;
 
-// Common base class for the simulation TTDevice backends (TTSimTTDevice and RtlSimulationTTDevice).
-// It is introduced as an intermediary in the class hierarchy and owns the state that is shared by
-// both backends. The behavior that operates on this state still lives in the derived classes for now
-// and will be migrated into this base incrementally.
+// Common base class for the simulation TTDevice backends. It sits as an intermediary in the class
+// hierarchy and owns the state shared by the derived simulation devices.
 //
-// The backend communicator is intentionally NOT owned here: TTSimCommunicator and RtlSimCommunicator
-// are unrelated `final` classes with no common base, so each derived device keeps its own
+// The backend communicator is intentionally not owned here; each derived device keeps its own
 // concretely-typed communicator.
 class SimulationTTDevice : public TTDevice {
 public:
@@ -62,6 +60,8 @@ public:
 
     SimulationTlbAllocator* get_tlb_allocator() { return tlb_allocator_.get(); }
 
+    std::unique_ptr<TlbWindow> get_io_window(tlb_data config, TlbMapping mapping, size_t size) override;
+
 protected:
     SimulationTTDevice(
         const std::filesystem::path& simulator_directory, std::unique_ptr<SimulationSysmemManager> sysmem_manager);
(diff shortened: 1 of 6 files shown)


ajovanovicTT (approved):

ajovanovicTT:
Nice, looked at this one and previous two related PR's (had nothing specific to comment).

device/api/umd/device/tt_device/simulation_tt_device.hpp
```diff
@@ -0,0 +1,95 @@
...
+#include "umd/device/types/xy_pair.hpp"
+
+namespace tt::umd {
+
+class SimulationSysmemManager;
+class SimulationTlbAllocator;
+class SimulationServerSocket;
+class TlbWindow;
+
+// Common base class for the simulation TTDevice backends (TTSimTTDevice and RtlSimulationTTDevice).
```
nbuncicTT: nit: omit the usage of names of concrete classes as these are hard to maintain (the concrete derived classes can be erased, added, etc.)

Same goes for stuff like "both backends".

Also, I'd omit the part that this isn't fully implemented, as it adds the diff in commits which isn't particularly useful.

device/api/umd/device/tt_device/simulation_tt_device.hpp
```diff
@@ -0,0 +1,95 @@
...
+// It is introduced as an intermediary in the class hierarchy and owns the state that is shared by
+// both backends. The behavior that operates on this state still lives in the derived classes for now
+// and will be migrated into this base incrementally.
+//
+// The backend communicator is intentionally NOT owned here: TTSimCommunicator and RtlSimCommunicator
```
nbuncicTT: nit: Omit implementation details for the communicator, etc.

Keep comments only about the way the class should behave (i.e. what it should be, not how it is used in concrete use-cases).

device/api/umd/device/tt_device/simulation_tt_device.hpp
```diff
@@ -0,0 +1,95 @@
...
+    void init_tlb_allocator(uint64_t bar0_base);
+    // Allocate the cached default TLB window for the current arch. Must be invoked from the derived
+    // constructor (after the communicator exists), since it dispatches to the virtual get_io_window.
+    void setup_cached_tlb_window();
+
+    // Construct the backend-specific TlbHandle + TlbWindow for an already-allocated TLB index.
+    virtual std::unique_ptr<TlbWindow> create_tlb_window(
+        int tlb_index, size_t size, TlbMapping mapping, tlb_data config) = 0;
+
+    std::recursive_mutex device_lock;
```
nbuncicTT, line 86: Omit protected data if you can, the flaws are outlined here: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c133-avoid-protected-data.

Instead, move it to private and add protected getters.

nbuncicTT (approved):

device/api/umd/device/tt_device/simulation_tt_device.hpp
pjanevskiTT: Agreed on C.133. These members (`device_lock`, `sysmem_manager_`, `tlb_allocator_`, `cached_tlb_window_`, `socket_`, `simulator_directory_`) predate this PR -- they landed in #2913/#2914, and this PR only touches the `get_io_window` unification. To keep the diff focused, I'll move them to private with protected getters in a dedicated follow-up rather than expanding this PR into already-merged code.

PR #2917, opened 2026-06-24
Unify simulation read/write_from_device behind I/O hooks

pjanevskiTT:
### Issue
Part of #2865

### Description
Final step toward a common implementation for the RTL and TTSim simulation `TTDevice` backends, stacked on #2916. Unifies `read_from_device` / `write_to_device`: the shared skeleton moves into `SimulationTTDevice`, with backend specifics expressed through a small set of virtual hooks.

### List of the changes
- Move the `read_from_device` / `write_to_device` skeleton into `SimulationTTDevice`: closed-check, lock, coordinate translation, special-case fast path, cached-window-or-tile dispatch, and post-read clock. The `CoreCoord` is translated once via `translate_chip_coord_to_translated` and the translated coordinate is handed to the hooks.
- Add hooks: pure-virtual `tile_read_bytes` / `tile_write_bytes`, plus defaulted `is_device_closed`, `handle_special_read` / `handle_special_write`, `should_use_cached_tlb_window`, and `after_read`.
- `RtlSimulationTTDevice` overrides the special-case hooks for `SYSTEM_NOC`; `TTSimTTDevice` overrides them for DRAM teleport and also overrides the closed-check, the QUASAR cached-window exclusion, and the post-read auto-clock.

### Testing
CI

### API Changes
None.

### DIFF
--- a/device/api/umd/device/tt_device/simulation_tt_device.hpp
+++ b/device/api/umd/device/tt_device/simulation_tt_device.hpp
@@ -14,6 +14,7 @@
 #include "umd/device/chip_helpers/simulation_tlb_allocator.hpp"
 #include "umd/device/pcie/tlb_window.hpp"
 #include "umd/device/tt_device/tt_device.hpp"
+#include "umd/device/types/core_coordinates.hpp"
 #include "umd/device/types/tlb.hpp"
 #include "umd/device/types/xy_pair.hpp"
 
@@ -37,6 +38,9 @@ public:
     void adopt_socket(std::unique_ptr<SimulationServerSocket> socket);
 
     // --- TTDevice overrides whose behavior is identical across both simulation backends ---
+    void read_from_device(void* mem_ptr, CoreCoord core, uint64_t addr, size_t size) override;
+    void write_to_device(const void* mem_ptr, CoreCoord core, uint64_t addr, size_t size) override;
+
     void dma_d2h(void* dst, uint32_t src, size_t size) override;
     void dma_d2h_zero_copy(void* dst, uint32_t src, size_t size) override;
     void dma_h2d(uint32_t dst, const void* src, size_t size) override;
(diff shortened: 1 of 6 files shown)


device/api/umd/device/tt_device/simulation_tt_device.hpp
```diff
@@ -81,6 +85,29 @@ class SimulationTTDevice : public TTDevice {
     virtual std::unique_ptr<TlbWindow> create_tlb_window(
         int tlb_index, size_t size, TlbMapping mapping, tlb_data config) = 0;
 
+    // --- read/write_from_device hooks ---
+
+    // Direct tile (NOC unicast) access through the backend communicator.
+    virtual void tile_read_bytes(tt_xy_pair core, uint64_t addr, void* mem_ptr, size_t size) = 0;
+    virtual void tile_write_bytes(tt_xy_pair core, uint64_t addr, const void* mem_ptr, size_t size) = 0;
+
+    // Returns true when the device is closed and I/O should be silently skipped. Default: never.
```
nbuncicTT: Unnecessary comment, as the API name is self explanatory - a query that states if the device is closed. If the device is closed, IO isn't possible, and the default behavior is also seen in the body of this inline function - omit it.

device/api/umd/device/tt_device/simulation_tt_device.hpp
```diff
@@ -81,6 +85,29 @@ class SimulationTTDevice : public TTDevice {
...
+    virtual bool is_device_closed() { return false; }
+
+    // Backend-specific fast paths handled entirely outside the cached-TLB / tile path. Return true if
+    // the access was fully handled (and nothing further should run). Default: not handled.
```
nbuncicTT: nit: Omit the `Default: <states what default does>` comments - they're hard to maintain and don't clarify what the function does.

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -154,51 +154,52 @@ void TTSimTTDevice::close_device() {
...
+        return true;
     }
+    communicator_->tile_write_bytes(core.x, core.y, addr, mem_ptr, size);
+    return true;
 }
 
-void TTSimTTDevice::read_from_device(void* mem_ptr, CoreCoord core, uint64_t addr, size_t size) {
-    if (communicator_->is_closed()) {
-        return;
+bool TTSimTTDevice::handle_special_read(void* mem_ptr, tt_xy_pair core, uint64_t addr, size_t size) {
```
nbuncicTT: This is another special case which is controlled by an environment variable which adds a side-effect to DRAM writes in TTSim (or it's variant). Very hard to reason about.

I agree in general that it makes sense to cut this out in a separate function like this one, I would just not make this a virtual behavior.

I would do the following:
- Add a private non-virtual TTSimTTDevice, something named like `special_dram_read`
- Move the behavior from `handle_special_read` function to `special_dram_read`
- Invoke the function `handle_special_read`

That way, you have an extensible special read, because I have a feeling that there will be more hacks/special read/writes in the future.. Also, the special read is more self-explanatory because we don't know the intention of the special read, and by it being a special dram read, we know it's tied to that..

nbuncicTT: This goes for both special read/write for TTSim. For RtlSimulation I'd make private helpers for SMN because it's what they're doing - later this can be also extended.

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -154,51 +154,52 @@ void TTSimTTDevice::close_device() {
...
-        cached_tlb_window_->read_block_reconfigure(mem_ptr, translated_core, addr, size, get_selected_noc_id());
-    } else {
-        communicator_->tile_read_bytes(translated_core.x, translated_core.y, addr, mem_ptr, size);
+    if (!communicator_->dram_read_bytes(core.x, core.y, addr, mem_ptr, size)) {
+        communicator_->tile_read_bytes(core.x, core.y, addr, mem_ptr, size);
     }
-    // Ideally we would not auto-clock on reads at all, but some clocking is required to avoid hangs
-    // in the absence of an API reliably called from all spin loops polling the device
-    communicator_->advance_clock(1);
+    communicator_->advance_clock(10);
```
nbuncicTT, line 268: If you opt out to add the new function for special DRAM reads, please add a comment that it has a side effect of advancing the clock for 10 cycles.

We now assume it's 1 clock, but now it's 10 and for some other special function it'll be something else, so users are aware of the side effects.

nbuncicTT (approved):

Issue #2859, opened 2026-06-21
Add WC-based API in TTDevice


nbuncicTT:
**Description**

TTDevice currently maps all TLB windows as uncacheable (UC), including `write_device`/`read_device` which do not require UC. The interface already supports write-combine (WC), but it is not utilized. This change switches `write_device`/`read_device` to use WC mappings. `write_regs`/`read_regs` will remain UC and enforce alignment guarantees — **there will be no UC API path without alignment enforcement**. This will affect upstream users who depend on the current uniform UC behavior.

**Scope**

- Switch `write_device`/`read_device` TLB mappings from UC to WC
- Ensure `read_regs`/`write_regs` remain UC with enforced alignment guarantees
- Identify and document impact on upstream consumers
- Validate correctness of WC behavior on supported architectures

**Timeline**

No hard deadline. Part of the UMD Base API migration epic.

**Is your feature request related to a problem? Please describe.**

All TLB windows are currently mapped UC, leaving WC performance gains on the table for bulk device memory access. Register access needs to stay UC for alignment correctness, but there is no reason device memory access can't benefit from WC.

**Definition of Done**

- `write_device`/`read_device` use WC-mapped TLB windows
- `write_regs`/`read_regs` use UC-mapped TLB windows with alignment guarantees enforced
- Upstream impact is documented
- No regressions in existing tests

broskoTT:
Done, but broke the CI, will be resolved as part of #3247

PR #2920, opened 2026-06-25
Use a WC-mapped TLB window for device memory access in PcieProtocol

pjanevskiTT:
### Issue
#2859

### Description
`PcieProtocol` mapped its single cached TLB window UC and used it for both device memory and register access. This adds a separate WC-mapped window for the device memory path (`read_from_device`/`write_to_device` and multicast) while register access (`read_from_device_reg`/`write_to_device_reg`) keeps the UC window with Strict ordering and alignment enforcement. Only the CPU-side mapping changes; NOC ordering is unchanged. Depends on #2919, which moved register accesses off the memory API.

### List of the changes
- Split `PcieProtocol`'s cached TLB window into a WC window (device memory + multicast) and a UC window (register access).

### Testing
CI

### API Changes
None.


device/tt_device/protocol/pcie_protocol.cpp
```diff
@@ -57,19 +57,31 @@ PcieProtocol::~PcieProtocol() = default;
...
-    // The cached window may already exist if I/O ran before the hang detector was wired; keep it in sync.
-    if (cached_tlb_window_ != nullptr) {
-        cached_tlb_window_->set_io_timeout_hang_check(hang_check_);
+    // The cached windows may already exist if I/O ran before the hang detector was wired; keep them in sync.
+    if (cached_wc_tlb_window_ != nullptr) {
+        cached_wc_tlb_window_->set_io_timeout_hang_check(hang_check_);
     }
+    if (cached_uc_tlb_window_ != nullptr) {
+        cached_uc_tlb_window_->set_io_timeout_hang_check(hang_check_);
+    }
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed in c0d2f46d — `set_io_timeout_callback` now takes `io_lock_`, serializing it against the IO paths that read `hang_check_` and the cached windows under the same lock.

device/api/umd/device/tt_device/protocol/pcie_protocol.hpp
```diff
@@ -69,7 +69,10 @@ class PcieProtocol : public DeviceProtocol, public PcieInterface {
     uint32_t bar_read32(uint32_t addr) override;
 
 private:
-    TlbWindow* get_cached_tlb_window();
+    // WC-mapped window for device memory access (read_from_device/write_to_device, multicast).
+    TlbWindow* get_cached_wc_tlb_window();
```
nbuncicTT, line 82: So just a heads up which needs a bit of alignment - we are changing the default behavior of write/read to/from device from UC to WC and we are adding a new API with UC and alignment. We can maybe see potential issues arising from this, so can you do the following:
- How does this affect the upper workload layer? Evaluate if Chip based classes invoke the right TTDevice API
- Inform tt-exalens of this new API and the modified API of the old one
- Add this TTDevice API to nanobind so clients can use it

If you think this is too much for this PR, we can make a separate issue for this and a few more PRs.

nbuncicTT (approved):

ajovanovicTT (approved):

device/api/umd/device/tt_device/protocol/pcie_protocol.hpp
broskoTT: Agreed on all three, and it is too much for this PR. Tracked in #3219.

On the workload layer specifically, the answer is better than expected: on PCIe, `LocalChip` never reaches the protocol's cached windows. It owns its own WC data window and UC register window and only falls back to `tt_device_->...` when the device is not PCIe, so this PR does not change Chip-layer behavior at all, and its data path was already WC. Two call sites are wrong by contract though: the non-PCIe fallbacks in `LocalChip::write_to_device_reg`/`read_from_device_reg` call `write_to_device`/`read_from_device` to service a register access. Harmless today since `JtagProtocol` and `RemoteProtocol` forward ctrl to data after validation. `RemoteChip` gets it right.

nanobind is the sharpest gap: `noc_read32`/`noc_write32`/`noc_read`/`noc_write` all route to the data API and there is no ctrl binding, so Python callers poking registers move onto the WC window with nothing aligned to switch to.

For tt-exalens the open question is alignment: it reads byte-granular at arbitrary addresses, and the ctrl API enforces 4-byte alignment and multiple-of-4 size. #2859 says there will be no UC path without alignment enforcement, so either those accesses accept WC or they need something we do not offer yet. That needs settling with them, not assuming.

Separately, and not blocking this PR: the WC data window needs a host store fence before the TLB config write. `configure()` is a UC store through the UC BAR0 mapping, which drains the WC buffers on x86 as a side effect, but that is not an ordering guarantee and does not hold on aarch64 where the window is Normal-NC and the config space is Device memory. The existing `TLBManager`/`LocalChip` WC windows already have the same hole, so this PR widens it rather than creating it. Fixing it in a follow-up.

PR #2924, opened 2026-06-25
AICLK target tolerance

### LINKED ISSUE #2663: AICLK timeout warning can trigger in operational range on BH Galaxy
Related incident: https://tenstorrent.atlassian.net/browse/DCAMP-186

The incident in DCAMP-186 shows repeated AICLK timeout warnings on bh-glx-120-a09u08 during validation. In the reported case, the warning was:

Waiting for AICLK value to settle failed on timeout after 100ms
Expected to see 1350, last value observed 1337

At the same time, the reported ASIC temperature was normal, so this does not look like a straightforward overheating case.

The UMD side needs guidance on what should be considered an operational range where this warning should not fire. Today the warning is concerning to users even when the observed value appears close to the target.

Requested guidance:
- @jgrowdenTT: please confirm whether we should read back AICLK_ARB_MAX and print the limiting arbiter/frequency when this condition hits.
- @jgrowdenTT: please also confirm whether we should wait UPDATE_TELEM_SPEED plus a few extra ms before deciding timeout, since 100ms matches the telemetry update interval.
- @afongTT: please provide FW guidance on how “operational range” should be defined for this warning path.
- Original reporter: Gustavo Sarabando

Context from the Jira thread:
- The initial incident report came from Gustavo Sarabando.
- James suggested reading back AICLK_ARB_MAX and also extending the wait beyond the telemetry update interval.
- The Jira thread explicitly asked to move follow-up into a tt-umd issue and link back to DCAMP-186.
- Bojan also noted the original timeout discussion is related to tt-umd issue #1409.

Suggested outcome for this issue:
1. Define the acceptable operational range for AICLK mismatch before warning.
2. Update UMD logging/behavior accordingly.
3. If needed, improve diagnostic output by printing the limiting arbiter/frequency.
4. Revisit timeout/wait logic relative to telemetry refresh timing.

broskoTT:
### Issue
closes https://github.com/tenstorrent/tt-umd/issues/2663

### Description
Don't wait and send warnings if AICLK is close to the target.

### List of the changes
- Add a generic is_within_percentage
- Check if value is within AICLK_TOLERANCE_PERCENT, and don't wait any longer if it is.

### Testing
No testing

### API Changes
This PR has no API changes.

### DIFF
--- a/device/chip/chip.cpp
+++ b/device/chip/chip.cpp
@@ -25,10 +25,16 @@
 #include "umd/device/types/core_coordinates.hpp"
 #include "umd/device/types/telemetry.hpp"
 #include "umd/device/types/xy_pair.hpp"
+#include "umd/device/utils/common.hpp"
 #include "umd/device/utils/error.hpp"
 #include "umd/device/utils/timeouts.hpp"
 
 namespace tt::umd {
+
+namespace {
+// AICLK rarely settles on the exact target; accept any value within this percentage of the target.
+constexpr double AICLK_TOLERANCE_PERCENT = 5.0;
+}  // namespace
 
 Chip::Chip(tt::ARCH arch) { set_default_params(arch); }
 
@@ -224,6 +230,15 @@ void Chip::wait_for_aiclk_value(
     }
     uint32_t aiclk = tt_device->get_clock();
     while (aiclk != target_aiclk) {
+        if (is_within_percentage(aiclk, target_aiclk, AICLK_TOLERANCE_PERCENT)) {
+            log_warning(
+                LogUMD,
+                "AICLK settled at {} MHz, within {}% of the requested {} MHz but not an exact match. Proceeding.",
+                aiclk,
+                AICLK_TOLERANCE_PERCENT,
+                target_aiclk);
+            return;
+        }
         auto end = std::chrono::steady_clock::now();
         auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
         if (duration.count() > timeout_ms.count()) {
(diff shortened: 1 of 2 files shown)


device/chip/chip.cpp
```diff
@@ -25,11 +25,17 @@
 #include "umd/device/types/core_coordinates.hpp"
 #include "umd/device/types/telemetry.hpp"
 #include "umd/device/types/xy_pair.hpp"
+#include "umd/device/utils/common.hpp"
 #include "umd/device/utils/error.hpp"
 #include "umd/device/utils/timeouts.hpp"
 
 namespace tt::umd {
 
+namespace {
```
aleksamarkovicTT: Is there a need for this to be out of the tt::umd namespace, given it's not visible outside of the file?

device/chip/chip.cpp
```diff
@@ -224,6 +230,15 @@ void Chip::wait_for_aiclk_value(
     }
     uint32_t aiclk = tt_device->get_clock();
     while (aiclk != target_aiclk) {
+        if (is_within_percentage(aiclk, target_aiclk, AICLK_TOLERANCE_PERCENT)) {
+            log_warning(
```
aleksamarkovicTT, line 232: Should this be a warning or info?

broskoTT: I think warning is fine. It would be a warning today, the optimization here is actually not to wait `timeout_ms` if it is within 5%

aleksamarkovicTT: I see.

aleksamarkovicTT (approved):
Resolve the nit comments

PR #2936, opened 2026-06-26
Rename and move set_power_state logic to set_clock_state in TTDevice

### LINKED ISSUE #2862: Add set_clock_state API to TTDevice
**Description**

`Chip` currently has a `set_power_state` method that manages clock state. This change renames it to `set_clock_state`, migrates the behavior from `Chip` down into `TTDevice`, and exposes it as a proper base API.

**Scope**

- Add `set_clock_state` API on `TTDevice`
- Migrate clock management behavior from `Chip::set_power_state` into `TTDevice`
- Rename usage from `set_power_state` to `set_clock_state` in `Chip`
- Implement for supported device backends

**Timeline**

Clock management lives in `Chip` under the misleading name `set_power_state`. It belongs in the device layer with a name that reflects what it actually does.

**Definition of Done**

- `set_clock_state` is available on `TTDevice`
- Clock management logic removed from `Chip`

ajovanovicTT:
### Issue
#2862

### Description
Moves clock-state management from Chip into the TTDevice base API, renaming set_power_state(DevicePowerState) to set_clock_state. Chip::set_clock_state now delegates to TTDevice.

### List of the changes
 - Added TTDevice::set_clock_state (no-op default) + moved wait_for_aiclk_value into TTDevice.
 - Implemented set_clock_state for Wormhole and Blackhole backends.
 - Chip::set_power_state becomes non-virtual set_clock_state delegating to TTDevice; removed migrated logic.
 - Dropped no-op overrides from Simulation/Mock/SWEmule chips.
 - Updated callers (Cluster, LocalChip, RemoteChip).

### Testing
CI.

### API Changes
This PR has API changes.


device/cluster.cpp
```diff
@@ -995,7 +995,7 @@ void Cluster::broadcast_tensix_risc_reset_to_cluster(uint32_t reg_value) {
 
 void Cluster::set_power_state(DevicePowerState device_state) {
     for (auto& [_, chip] : chips_) {
```
pjanevskiTT, line 997: Naming inconsistency: `Cluster::set_power_state(DevicePowerState)` now delegates to `Chip::set_clock_state`. If #2862 is about standardizing on "clock state", consider renaming this Cluster method too in a follow-up (would touch `test_cluster.cpp` and `test_pcie_dma.cpp`). Non-blocking.

device/tt_device/tt_device.cpp
```diff
@@ -242,6 +242,59 @@ void TTDevice::set_power_state(bool busy) {
...
+        target_aiclk = get_min_clock_freq();
+    }
+    uint32_t aiclk = get_clock();
+    while (aiclk != target_aiclk) {
+        auto end = std::chrono::steady_clock::now();
+        auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
+        if (duration.count() > timeout_ms.count()) {
+            auto *telemetry = get_arc_telemetry_reader();
+            std::string arb_max_info;
+            if (telemetry != nullptr && telemetry->is_entry_available(TelemetryTag::AICLK_ARB_MAX)) {
```
pjanevskiTT: `TelemetryTag` is only pulled in transitively here (via `arc/arc_telemetry_reader.hpp`). Now that this logic lives in `tt_device.cpp`, a direct `#include "umd/device/types/telemetry.hpp"` would be more robust (IWYU). Optional.

device/tt_device/blackhole_tt_device.cpp
```diff
@@ -230,6 +230,21 @@ uint32_t BlackholeTTDevice::get_clock() {
...
+        exit_code = get_arc_messenger()->send_message((uint32_t)blackhole::ArcMessageType::AICLK_GO_BUSY);
+    } else {
+        exit_code = get_arc_messenger()->send_message((uint32_t)blackhole::ArcMessageType::AICLK_GO_LONG_IDLE);
+    }
+    UMD_ASSERT(
+        exit_code == 0,
+        error::RuntimeError,
+        fmt::format("Failed to set clock state to {} with exit code: {}", (int)state, exit_code));
+    wait_for_aiclk_value(state);
+}
```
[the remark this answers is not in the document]
ajovanovicTT: Since this PR only moves logic to tt_device would leave this as it was before, with later exploration if SHORT_IDLE could be sent to bh fw.

device/cluster.cpp
ajovanovicTT: Would open separate PR for this.

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -364,6 +365,17 @@ class TTDevice {
      */
     virtual void set_power_state(bool busy);
 
+    /**
```
pjanevskiTT, line 364: @nbuncicTT do we have any preferred style of comments for base API?

device/tt_device/blackhole_tt_device.cpp
```diff
@@ -230,6 +230,21 @@ uint32_t BlackholeTTDevice::get_clock() {
 
 uint32_t BlackholeTTDevice::get_min_clock_freq() { return blackhole::AICLK_IDLE_VAL; }
 
+void BlackholeTTDevice::set_clock_state(DevicePowerState state) {
+    ZoneScoped;
+    int exit_code = 0;
+    if (state == DevicePowerState::BUSY) {
```
pjanevskiTT: as we talked offline, can you open an issue to address short idle on blackhole? I don't know if that exists. But this is good for just moving current functionality to TTDevice layer, and this can be a follow up issue not related to base API work

device/tt_device/tt_device.cpp
```diff
@@ -242,6 +243,59 @@ void TTDevice::set_power_state(bool busy) {
     get_pci_device()->set_power_state(busy);
 }
 
+void TTDevice::set_clock_state(DevicePowerState /*state*/) {
```
pjanevskiTT, line 252: do we want empty function here or should we make this abstract and then simulators can explicitly override with empty impl? Note that TTSim should get this supported in the near future

pjanevskiTT (approved):

device/tt_device/blackhole_tt_device.cpp
ajovanovicTT: Searched over tt-system-firmware, only BUSY and LONG_IDLE exist there.

device/tt_device/tt_device.cpp
ajovanovicTT: As discussed offline, I would leave this empty implementation in base class for now, and than later if needed it can be made abstract with empty overrides in all child classes.

device/api/umd/device/chip/chip.hpp
```diff
@@ -108,7 +108,9 @@ class Chip {
     */
     virtual void deassert_risc_reset(const RiscType selected_riscs, bool staggered_start);
 
-    virtual void set_power_state(DevicePowerState state);
+    // Set the device clock state. Delegates to the underlying TTDevice, which drives AICLK via ARC
```
nbuncicTT: nit: Redundant comment, in the signature of the function it's stated what it does, no need to get into implementation details, hence remove it.

device/api/umd/device/tt_device/tt_device.hpp
nbuncicTT: Yes, we prefer doxygen style comments, but the problem is that the ones used for the docs are too complex, so I won't nitpick on these above the APIs themselves for now.

device/tt_device/tt_device.cpp
```diff
@@ -242,6 +243,59 @@ void TTDevice::set_power_state(bool busy) {
...
+    // No-op by default. Backends with a controllable clock (Wormhole, Blackhole) override this to
+    // drive AICLK via ARC; backends without one (e.g. simulation) keep the no-op.
+}
+
+void TTDevice::wait_for_aiclk_value(DevicePowerState power_state, const std::chrono::milliseconds timeout_ms) {
+    auto start = std::chrono::steady_clock::now();
+    uint32_t target_aiclk = 0;
+    if (power_state == DevicePowerState::BUSY) {
```
nbuncicTT: What happens if the `DevicePowerState` is `SHORT_IDLE`? We go to 0Hz AICLK? Does this make sense? 

Handle this enum with a switch case and define explicitly what happens for `SHORT_IDLE`.

device/tt_device/tt_device.cpp
```diff
@@ -242,6 +243,59 @@ void TTDevice::set_power_state(bool busy) {
...
+        target_aiclk = get_max_clock_freq();
+    } else if (power_state == DevicePowerState::LONG_IDLE) {
+        target_aiclk = get_min_clock_freq();
+    }
+    uint32_t aiclk = get_clock();
+    while (aiclk != target_aiclk) {
+        auto end = std::chrono::steady_clock::now();
+        auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
+        if (duration.count() > timeout_ms.count()) {
```
nbuncicTT: For polling/waiting until some event use the `check_timeout`/`poll_until` utilities.

device/tt_device/tt_device.cpp
```diff
@@ -242,6 +243,59 @@ void TTDevice::set_power_state(bool busy) {
     get_pci_device()->set_power_state(busy);
 }
 
+void TTDevice::set_clock_state(DevicePowerState /*state*/) {
+    // No-op by default. Backends with a controllable clock (Wormhole, Blackhole) override this to
+    // drive AICLK via ARC; backends without one (e.g. simulation) keep the no-op.
+}
+
+void TTDevice::wait_for_aiclk_value(DevicePowerState power_state, const std::chrono::milliseconds timeout_ms) {
```
nbuncicTT, line 257: Try to make minimal functions which re-use utilities we have, so take this example as a reference (modify if needed):
```cpp
void TTDevice::wait_for_aiclk_value(DevicePowerState power_state, const std::chrono::milliseconds timeout_ms) {
    uint32_t target_aiclk = 0;
    switch (power_state) {
        case DevicePowerState::BUSY:
            target_aiclk = get_max_clock_freq();
            break;
        case DevicePowerState::LONG_IDLE:
            target_aiclk = get_min_clock_freq();
            break;
        case DevicePowerState::SHORT_IDLE:
            // No defined target frequency — nothing to wait for.
            return;
    }

    const bool settled = utils::poll_until(
        [&] { return get_clock() == target_aiclk; },
        timeout_ms,
        std::chrono::microseconds(500),   // busy-poll window
        std::chrono::microseconds(100));   // sleep interval after busy window

    if (!settled) {
        log_aiclk_timeout_warning(target_aiclk, timeout_ms);
    }
}
```

And write a helper function which can look like this:

```cpp
void TTDevice::log_aiclk_timeout_warning(uint32_t target_aiclk, std::chrono::milliseconds timeout_ms) {
    uint32_t aiclk = get_clock();

    std::string arb_max_info;
    auto* telemetry = get_arc_telemetry_reader();
    if (telemetry != nullptr && telemetry->is_entry_available(TelemetryTag::AICLK_ARB_MAX)) {
        const uint32_t arb_max = telemetry->read_entry(TelemetryTag::AICLK_ARB_MAX);
        arb_max_info = fmt::format(
            ", AICLK clamped by max-arbiter index {} at {} MHz", (arb_max >> 16) & 0xFFFF, arb_max & 0xFFFF);
    }

    log_warning(
        LogUMD,
        "AICLK failed to settle after {} ms. Expected {}, observed {}. ASIC temperature: {}{}",
        timeout_ms.count(),
        target_aiclk,
        aiclk,
        get_asic_temperature(),
        arb_max_info);

    if (telemetry != nullptr && telemetry->is_entry_available(TelemetryTag::UPDATE_TELEM_SPEED)) {
        const uint32_t update_interval_ms = telemetry->read_entry(TelemetryTag::UPDATE_TELEM_SPEED);
        if (timeout_ms.count() <= update_interval_ms) {
            log_warning(
                LogUMD,
                "AICLK timeout ({} ms) <= telemetry update interval ({} ms); "
                "observed value may be stale. Consider increasing AICLK_TIMEOUT.",
                timeout_ms.count(),
                update_interval_ms);
        }
    }
```
etc. 

This way it's more easy to reason about the function itself.

device/tt_device/tt_device.cpp
```diff
@@ -242,6 +243,59 @@ void TTDevice::set_power_state(bool busy) {
...
+        auto end = std::chrono::steady_clock::now();
+        auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
+        if (duration.count() > timeout_ms.count()) {
+            auto *telemetry = get_arc_telemetry_reader();
+            std::string arb_max_info;
+            if (telemetry != nullptr && telemetry->is_entry_available(TelemetryTag::AICLK_ARB_MAX)) {
+                const uint32_t arb_max = telemetry->read_entry(TelemetryTag::AICLK_ARB_MAX);
+                const uint32_t arb_freq = arb_max & 0xFFFF;
+                const uint32_t arb_idx = (arb_max >> 16) & 0xFFFF;
+                arb_max_info = fmt::format(", AICLK clamped by max-arbiter index {} at {} MHz", arb_idx, arb_freq);
```
nbuncicTT: It's not a good practice to create a string every iteration, look how to mitigate this in the example written on the PR.

device/tt_device/wormhole_tt_device.cpp
```diff
@@ -111,6 +111,45 @@ uint32_t WormholeTTDevice::get_clock() {
...
+    uint32_t msg = get_power_state_arc_msg(state);
+    int exit_code = get_arc_messenger()->send_message(msg, {0, 0});
+    UMD_ASSERT(
+        exit_code == 0,
+        error::RuntimeError,
+        fmt::format("Failed to set clock state to {} with exit code: {}", (int)state, exit_code));
+
+    // SHORT_IDLE has no defined target AICLK, so waiting for the clock to settle would needlessly
+    // block until timeout. Skip the wait for that state.
+    if (state == DevicePowerState::SHORT_IDLE) {
```
nbuncicTT: Prefer early returns in these scenarios:
```cpp
if (state == DevicePowerState::SHORT_IDLE) {
    log_warning(LogUMD, "Skipping AICLK settle wait for SHORT_IDLE clock state.");
    return;
}
wait_for_aiclk_value(state);
```

nbuncicTT (approved):
Nicely done

device/tt_device/tt_device.cpp
ajovanovicTT: Before I moved this logic here, in wait_for_aiclk_value SHORT_IDLE would always hit timeout, so I added skip for that case in set_clock_state. Now with the latest commit, as you proposed, added this missing case to this method which just early returns, so that SHORT_IDLE message is being sent but not waited for aiclk to settle (message is sent for WH only, on BH SHORT_IDLE doesn't exist so LONG_IDLE is sent as it was before).

device/tt_device/tt_device.cpp
ajovanovicTT: Thanks for the suggestion, I checked this solution and applied it with slight modification.

device/tt_device/tt_device.cpp
pjanevskiTT: SHORT_IDLE is just going to IDLE state (same as LONG IDLE) but just for a few ms (not sure how much). I don't think it's useful, I didn't see anyone using it

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -360,6 +361,17 @@ class TTDevice {
...
+     * Set the device clock (AICLK) state by sending the corresponding power-state request to ARC
```
nbuncicTT: In the description of the comments use the notion of request to device instead of ARC. TTDevice is an abstraction and doesn't couple to concrete IP implementations (i.e. BH/WH have ARC, Grendel doesn't, etc.)

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -360,6 +361,17 @@ class TTDevice {
...
+     * and waiting for the clock to settle at the expected frequency.
+     *
+     * The default implementation is a no-op for backends without a controllable clock (e.g.
```
nbuncicTT: This no-op doesn't make sense as a comment. This is an implementation detail, hence omit it.

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -511,6 +523,11 @@ class TTDevice {
     void multicast_write_via_unicast(
         const void *src, size_t size, tt_xy_pair core_start, tt_xy_pair core_end, uint64_t addr);
 
+    // Polls AICLK until it reaches the frequency expected for `power_state`, or logs a warning and
```
nbuncicTT, line 523: Again, mentioning ARC in comments doesn't make sense, and going into impl. details also doesn't make sense for APIs. 

Just state that it polls the AICLK until it reaches the desired frequency or times out.

device/chip/chip.cpp
```diff
@@ -195,83 +167,9 @@ void Chip::advance_device_execution() {
...
-                        "observed AICLK may be a stale telemetry value. Consider increasing AICLK_TIMEOUT.",
-                        timeout_ms.count(),
-                        update_telem_speed_ms);
-                }
-            }
-            return;
-        }
-        aiclk = tt_device->get_clock();
+void Chip::set_clock_state(DevicePowerState state) {
+    if (auto* tt_device = get_tt_device()) {
```
nbuncicTT: Omit the `nullptr` check as the `TTDevice` must exist for a `Chip` class.

nbuncicTT (approved):

device/chip/chip.cpp
ajovanovicTT: Had to return nullptr check at the end due to failing tests that use MockChip which doesn't have tt_device.

PR #2946, opened 2026-06-29
Fix verify_routing_firmware_state() check condition

aleksamarkovicTT:
### Issue
/

### Description
Revert verify_routing_firmware_state() check condition to before introducing the new error type. This should fix the tt-metal bump. Instead of checking for 0, we check for not 1, as t3k has value 2 for multimesh mode.

### Testing
Visible only on t3k.

### API Changes
This PR has no API changes.

### DIFF
--- a/device/topology/topology_discovery_wormhole.cpp
+++ b/device/topology/topology_discovery_wormhole.cpp
@@ -174,7 +174,9 @@ void TopologyDiscoveryWormhole::verify_routing_firmware_state(
     uint32_t routing_firmware_disabled;
     tt_device->read_from_device(
         &routing_firmware_disabled, eth_core, EthAddresses::ROUTING_FIRMWARE_STATE, sizeof(uint32_t));
-    if (is_running_on_6u != (routing_firmware_disabled != 0)) {
+    bool unexpected_config =
+        (is_running_on_6u && routing_firmware_disabled == 0) || (!is_running_on_6u && routing_firmware_disabled == 1);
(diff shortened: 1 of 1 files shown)


device/topology/topology_discovery_wormhole.cpp
```diff
@@ -174,7 +174,9 @@ void TopologyDiscoveryWormhole::verify_routing_firmware_state(
     uint32_t routing_firmware_disabled;
     tt_device->read_from_device(
         &routing_firmware_disabled, eth_core, EthAddresses::ROUTING_FIRMWARE_STATE, sizeof(uint32_t));
-    if (is_running_on_6u != (routing_firmware_disabled != 0)) {
+    bool unexpected_config =
+        (is_running_on_6u && routing_firmware_disabled == 0) || (!is_running_on_6u && routing_firmware_disabled == 1);
```
broskoTT, line 178: what is the difference in behavior here?
Is it when routing_firmware_disabled is something outside 0 or 1?

broskoTT: If you don't know, please post both links (failing and succeeding) in the "Testing" section

aleksamarkovicTT: T3K has this value set to 2 instead of 0.
```
[7:0] == 2 => T3K Multimesh mode. Only do routing to R chip.
```
This is a small correction to fix the check on t3k.

broskoTT (approved):

Issue #2832, opened 2026-06-15
[TTSim] Multichip ttsim support: N300 (wh_x2)


pjanevskiTT:
Part of #2679.

Support the wh_x2 (N300) ttsim in UMD: one MMIO gateway (chip 0) plus a remote chip (chip 1) reached over simulated ethernet. UMD must enumerate both chips from a multichip `libttsim.so`, route device IO to the remote chip through the gateway, and skip ARC-only steps for the simulated remote chip (it has no ARC to service).

Implemented in #2808.

broskoTT:
Pending review

nbuncicTT:
Pending PR review.

pjanevskiTT:
closed by #2955

PR #2952, opened 2026-06-30
RemoteChip: accept Chip* base instead of LocalChip*

pjanevskiTT:
### Issue
Part of #2679, tracked by #2832. Split out of #2808.

### Description
Widen `RemoteChip`'s gateway pointer from `LocalChip*` to `Chip*` so a `RemoteChip` can sit behind any `Chip` (e.g. a `SimulationChip` gateway), not only a `LocalChip`. Behaviour-preserving: all uses go through `Chip` virtuals and existing callers pass a `LocalChip*` which is-a `Chip*`.

### List of the changes
- `RemoteChip::create`, its constructor, and the `local_chip_` member take `Chip*` instead of `LocalChip*`.

### Testing
CI

### API Changes
`RemoteChip::create` parameter type widened from `LocalChip*` to `Chip*`.

### DIFF
--- a/device/chip/remote_chip.cpp
+++ b/device/chip/remote_chip.cpp
@@ -28,15 +28,15 @@ namespace tt::umd {
 
 static_assert(!std::is_abstract<RemoteChip>(), "RemoteChip must be non-abstract.");
 
-std::unique_ptr<RemoteChip> RemoteChip::create(std::unique_ptr<TTDevice> remote_tt_device, LocalChip* local_chip) {
+std::unique_ptr<RemoteChip> RemoteChip::create(std::unique_ptr<TTDevice> remote_tt_device, Chip* local_chip) {
     ZoneScopedC(tracy::Color::DarkGreen);
     UMD_ASSERT(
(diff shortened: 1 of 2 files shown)


device/chip/remote_chip.cpp
```diff
@@ -28,15 +28,15 @@ namespace tt::umd {
 
 static_assert(!std::is_abstract<RemoteChip>(), "RemoteChip must be non-abstract.");
 
-std::unique_ptr<RemoteChip> RemoteChip::create(std::unique_ptr<TTDevice> remote_tt_device, LocalChip* local_chip) {
+std::unique_ptr<RemoteChip> RemoteChip::create(std::unique_ptr<TTDevice> remote_tt_device, Chip* local_chip) {
```
aleksamarkovicTT, line 31: The dependency on Chip should be removed entirely in the future. It is only used for calling the localchip's version of deassert_risc_resets(), which I don't think is necessary. @broskoTT told me he doesn't remember the reason why remotechip calls the localchip method.

aleksamarkovicTT (approved):

PR #2953, opened 2026-06-30
TTDevice: allow supplying SocDescriptor at remote create()

pjanevskiTT:
### Issue
Part of #2679, tracked by #2832. Split out of #2808; builds on #2952.

### Description
Add an optional, defaulted `SocDescriptor` to the remote `TTDevice::create()` overload. A remote `TTDevice` is normally initialized over ARC, which constructs its `SocDescriptor`; simulated remote chips have no ARC to probe, so the caller can supply the descriptor directly. Silicon callers are unaffected. The parameter is temporary and should be removed once ttsim provides a mocked ARC.

### List of the changes
- `TTDevice::create` (remote overload) gains an optional `SocDescriptor`, applied via `set_soc_descriptor` when set.

### Testing
CI

### API Changes
`TTDevice::create` (remote overload) gains a defaulted optional `SocDescriptor` parameter; existing callers unaffected.

### DIFF
--- a/device/api/umd/device/tt_device/tt_device.hpp
+++ b/device/api/umd/device/tt_device/tt_device.hpp
@@ -91,9 +91,14 @@ public:
         bool use_safe_api = false,
         const std::shared_ptr<SocArchDescriptor> &soc_arch_descriptor = nullptr);
 
+    // A remote TTDevice is normally initialized over ARC (init_tt_device), which constructs its SocDescriptor.
+    // Simulated remote chips have no ARC to probe, so the caller may supply the full descriptor via soc_descriptor.
+    // TODO: temporary - the soc_descriptor parameter should be removed once ttsim provides a mocked ARC that can
+    // serve the SocDescriptor the same way silicon does.
     static std::unique_ptr<TTDevice> create(
         std::unique_ptr<RemoteCommunication> remote_communication,
-        const std::shared_ptr<SocArchDescriptor> &soc_arch_descriptor = nullptr);
+        const std::shared_ptr<SocArchDescriptor> &soc_arch_descriptor = nullptr,
+        std::optional<SocDescriptor> soc_descriptor = std::nullopt);
 
     virtual ~TTDevice() = default;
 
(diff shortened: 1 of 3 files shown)


device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -91,9 +91,14 @@ class TTDevice {
...
         const std::shared_ptr<SocArchDescriptor> &soc_arch_descriptor = nullptr);
 
+    // A remote TTDevice is normally initialized over ARC (init_tt_device), which constructs its SocDescriptor.
+    // Simulated remote chips have no ARC to probe, so the caller may supply the full descriptor via soc_descriptor.
+    // TODO: temporary - the soc_descriptor parameter should be removed once ttsim provides a mocked ARC that can
+    // serve the SocDescriptor the same way silicon does.
     static std::unique_ptr<TTDevice> create(
         std::unique_ptr<RemoteCommunication> remote_communication,
-        const std::shared_ptr<SocArchDescriptor> &soc_arch_descriptor = nullptr);
+        const std::shared_ptr<SocArchDescriptor> &soc_arch_descriptor = nullptr,
```
aleksamarkovicTT: I don't like that we have to do this change. Can you please make this path non-callable outside of TTSimTTDevice, make the constructor private but a friend of TTSimTTDevice - separate this flow from Silicon as much as possible.

pjanevskiTT: Separated the simulation flow from silicon. The optional `soc_descriptor` param is off the public `create()` — it's back to its original signature. Simulation now goes through a dedicated `TTDevice::create_simulation_remote()` factory that's compiled in only under `TT_UMD_BUILD_SIMULATION` (both the declaration and the `WormholeTTDevice` friend grant are guarded), so in a silicon-only build the path doesn't exist at all.

One note: I didn't make it a friend of `TTSimTTDevice` specifically, because the actual caller of this path is `Cluster::create_simulation_remote_chip()` (also `#ifdef TT_UMD_BUILD_SIMULATION`), not `TTSimTTDevice`. The ifdef guard gives the separation you're after without tying it to a class that doesn't own the call. Let me know if you'd still prefer a tighter friend/private constructor scoping.

aleksamarkovicTT (approved):

device/api/umd/device/tt_device/tt_device.hpp
aleksamarkovicTT: Looks good, thanks.

Issue #1288, opened 2025-09-10
Cluster descriptor support for simulator


pjanevskiTT:
At the moment UMD assumes that all chips during simulation will be MMIO mapped, or LocalChip when it comes to UMD abstractions. This is going to change with future tt-sim work for supporting multi device
We need to support combination of local and remote chips properly for simulation, basically the same that we do for silicon when someones passes in custom cluster descriptor on silicon
From the top of my head, there shouldn't be another class to support this, we should just connect SimulationDevice to RemoteChip class and I would expect things to work out of the box, if we read the cluster descriptor properly

nbuncicTT:
Pending review.

PR #2954, opened 2026-06-30
Cluster: discover ttsim cluster descriptor from simulator directory

pjanevskiTT:
### Issue
Part of #2679, tracked by #2832. Split out of #2808; builds on #2953.
#1288 

### Description
Discover an optional `cluster_descriptor.yaml` beside a ttsim `.so` (mirroring the `soc_descriptor.yaml` convention) and build a constrained `ClusterDescriptor` from it. When the file is present, the simulator's `soc_descriptor.yaml` is also resolved so each chip's `SocDescriptor` is built from simulator-provided data rather than a default arch descriptor. Behaviour is unchanged when the file is absent.

### List of the changes
- Add `SimulationChip::get_cluster_descriptor_path_from_simulator_path`.
- `Cluster` reads an optional `cluster_descriptor.yaml` beside the simulator `.so`, resolving the sim's `soc_descriptor.yaml` for per-chip `SocDescriptor`s when it does.

### Testing
CI

### API Changes
New `SimulationChip::get_cluster_descriptor_path_from_simulator_path`.

### DIFF
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -367,6 +367,30 @@ Cluster::Cluster(ClusterOptions options) {
         case ChipType::SWEMULE:
         case ChipType::SIMULATION: {
             if (options.cluster_descriptor == nullptr) {
+#ifdef TT_UMD_BUILD_SIMULATION
+                // A ttsim .so may ship a cluster_descriptor.yaml beside it describing a real (possibly multichip)
+                // topology, mirroring the soc_descriptor.yaml convention. When present, use it; otherwise fall
+                // through to the mock single-chip descriptor below (unchanged behaviour).
+                const bool is_ttsim_build =
+                    (options.chip_type == ChipType::SIMULATION && options.simulator_directory.extension() == ".so");
+                const std::string cluster_desc_path =
+                    is_ttsim_build
+                        ? SimulationChip::get_cluster_descriptor_path_from_simulator_path(options.simulator_directory)
+                        : "";
+
+                if (is_ttsim_build && std::filesystem::exists(cluster_desc_path)) {
+                    // Resolve the simulator's soc_descriptor.yaml as well, so the per-chip SocDescriptors built
+                    // below are populated from the simulator-provided data instead of a default arch descriptor
+                    // (matches the fallback path's sdesc_path resolution).
+                    if (options.sdesc_path.empty()) {
+                        options.sdesc_path =
+                            SimulationChip::get_soc_descriptor_path_from_simulator_path(options.simulator_directory);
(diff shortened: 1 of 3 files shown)


aleksamarkovicTT (approved):
resolve the sdesc_path addressed issue

device/cluster.cpp
```diff
@@ -367,6 +367,22 @@ Cluster::Cluster(ClusterOptions options) {
         case ChipType::SWEMULE:
         case ChipType::SIMULATION: {
             if (options.cluster_descriptor == nullptr) {
+#ifdef TT_UMD_BUILD_SIMULATION
```
nbuncicTT, line 370: nit: This can be simplified:
```cpp
                const bool is_sim_build = (options.chip_type == ChipType::SIMULATION && options.simulator_directory.extension() == ".so");
                const std::string path = is_sim_build ? SimulationChip::get_cluster_descriptor_path_from_simulator_path(options.simulator_directory) : "";

                if (is_sim_build && std::filesystem::exists(path)) {
                    cluster_desc = ClusterDescriptor::create_constrained_cluster_descriptor(
                        ClusterDescriptor::create_from_yaml(path).get(), options.target_devices);
                    break; 
                }
                
#endif
```

nbuncicTT (approved):

device/cluster.cpp
```diff
@@ -367,6 +367,22 @@ Cluster::Cluster(ClusterOptions options) {
...
+                    std::string cluster_desc_path =
+                        SimulationChip::get_cluster_descriptor_path_from_simulator_path(options.simulator_directory);
+                    if (std::filesystem::exists(cluster_desc_path)) {
+                        std::unique_ptr<ClusterDescriptor> full_cluster_desc =
+                            ClusterDescriptor::create_from_yaml(cluster_desc_path);
+                        cluster_desc = ClusterDescriptor::create_constrained_cluster_descriptor(
+                            full_cluster_desc.get(), options.target_devices);
+                        break;
+                    }
+                }
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed. The cluster-descriptor branch now resolves `options.sdesc_path` to the simulator's `soc_descriptor.yaml` (when empty) before breaking, matching the fallback path, so per-chip `SocDescriptor`s are built from simulator-provided data rather than a default arch descriptor.

device/cluster.cpp
pjanevskiTT: Done — flattened the block, hoisting `is_ttsim_build` and `cluster_desc_path` out of the `if`.

Issue #2778, opened 2026-06-05
FanSpeed tests failing on 19.10


broskoTT:
**Describe the bug**
Something changed for Fan speed reporting in new FW, and we haven't aligned to it

**Stack trace**
Happens on CI on machines which have FW updated https://github.com/tenstorrent/tt-umd/actions/runs/26944707809/job/79499084077

broskoTT:
- Also ping on syseng-tt-firmware-releases to check if this UMD unit test failed when they were testing the new release. Check their process

aleksamarkovicTT:
Changes since CMFW 19.10:
- Speeds and RPM of 2 fans reported instead of one, 31:16 for left, 15:0 for right.
- Speed is raw "tach reading", on WH see https://yyz-gitlab.local.tenstorrent.com/tenstorrent/syseng/-/commit/2cced6a58c2a89b94323b615df968eaa0bcb0e15 for RPM conversion.
- On BH, all values are "requested" values, not actual values of RPM/speed.

broskoTT:
Linked as the cause of a failed CI run: https://github.com/tenstorrent/tt-umd/actions/runs/27004482920

broskoTT:
Linked as the cause of a failed CI run: https://github.com/tenstorrent/tt-umd/actions/runs/26912911727

broskoTT:
Linked as the cause of a failed CI run: https://github.com/tenstorrent/tt-umd/actions/runs/27007094672

broskoTT:
Linked as the cause of a failed CI run: https://github.com/tenstorrent/tt-umd/actions/runs/26944707809

broskoTT:
Linked as the cause of a failed CI run: https://github.com/tenstorrent/tt-umd/actions/runs/26943658169

broskoTT:
Linked as the cause of a failed CI run: https://github.com/tenstorrent/tt-umd/actions/runs/26942492593

broskoTT:
Linked as the cause of a failed CI run: https://github.com/tenstorrent/tt-umd/actions/runs/26879706560

broskoTT:
Linked as the cause of a failed CI run: https://github.com/tenstorrent/tt-umd/actions/runs/26879382341

broskoTT:
Linked as the cause of a failed CI run: https://github.com/tenstorrent/tt-umd/actions/runs/27013098529

broskoTT:
[...]

PR #2957, opened 2026-06-30
Update fan telemetry for 19.10

aleksamarkovicTT:
### Issue
#2778 

### Description
Update fan speed and fan RPM telemetry methods in FirmwareInfoProvider

### List of the changes
- `get_fan_speed()` and `get_fan_rpm()` return `std::vector<std::optional<uint32_t>>`, always size 2, nullopt when information is not available.
- Provide RPM for two fans on WH.
- Updated tests. 

### Testing
CI

### API Changes
This PR has API changes.


device/firmware/firmware_info_provider.cpp
```diff
@@ -429,24 +429,39 @@ std::optional<uint32_t> FirmwareInfoProvider::get_arcclk() const {
     return read_scalar<uint32_t>(FirmwareFeature::ARCCLK);
 }
 
-std::optional<uint32_t> FirmwareInfoProvider::get_fan_speed() const {
+std::vector<std::optional<uint32_t>> FirmwareInfoProvider::get_fan_speed() const {
+    std::vector<std::optional<uint32_t>> fan_speeds(2, std::nullopt);
```
nbuncicTT: nit: Omit the magic number here, and make a named constant that says something like `MAX_NUM_OF_FANS`.

Also, add a comment over the new constant that this should be made a syseng/firmware issue that we should read the number of fans from telemetry and use that for this API instead of relying on hardcoded values.

nbuncicTT (approved):
Well done

Issue #2895, opened 2026-06-23
Report tt-metal telemetry region address through UMD


jbaumanTT:
in SYS-4149 syseng is adding a 1kB memory region for use by telemetry in metal. Once that lands, I'd like to have a way to grab the address of that region from UMD so metal can write to it.

pjanevskiTT:
proper link to syseng issue https://tenstorrent.atlassian.net/browse/SYS-4149?issueKey=SYS-4149&subProduct=jira-software

PR #2965, opened 2026-07-01
Add Blackhole runtime telemetry buffer accessors

anashTT:
### Issue
Relates to tenstorrent/tt-metal#43293

### Description

This branch adds accessors to read the Blackhole runtime telemetry buffer's address and size from ARC scratch RAM, exposing where firmware publishes runtime telemetry data.

### List of the changes

- Added get_runtime_telemetry_buffer_address()/get_runtime_telemetry_buffer_size() virtuals to architecture_implementation (default std::nullopt), overridden in blackhole_implementation to return SCRATCH_RAM_22/SCRATCH_RAM_23 locations for firmware ≥ 19.12.0.
- Added the corresponding Firmware Info Provider methods

### Testing

Added test to TestFirmwareInfoProvider

### API Changes

New public methods FirmwareInfoProvider get_runtime_telemetry_buffer_address() and get_runtime_telemetry_buffer_size(), plus two new virtual methods on architecture_implementation (with Blackhole overrides).


device/tt_device/tt_device.cpp
```diff
@@ -560,6 +476,28 @@ FirmwareInfoProvider *TTDevice::get_firmware_info_provider() const {
 
 FirmwareBundleVersion TTDevice::get_firmware_version() { return get_firmware_info_provider()->get_firmware_version(); }
 
+std::optional<uint32_t> TTDevice::get_runtime_telemetry_buffer_address() {
```
aleksamarkovicTT: I don't think TTDevice is the place for this method. I think the best place would be the future DeviceFirmware class, defined in the base API spec, but that isn't implemented yet. Perhaps FirmwareInfoProvider would be a better place, or just a static firmware utility in firmware_utils.hpp.

tests/api/test_tt_device.cpp
```diff
@@ -376,3 +376,43 @@ TEST(ApiTTDeviceTest, UninitializedError) {
...
+            EXPECT_FALSE(size.has_value());
+            EXPECT_FALSE(address.has_value());
+            tt_device->set_power_state(false);
+            continue;
+        }
+
+        EXPECT_TRUE(size.has_value());
+        EXPECT_TRUE(address.has_value());
+
+        if (size.has_value() && address.has_value()) {
```
pjanevskiTT: not sure how much this part of the test makes sense, it is doing the same logic as ttdevice basically. Maybe decoding some data from the telemetry and figuring out if the value is as expected would be better, but if that is not possible, I would just remove this if

device/tt_device/tt_device.cpp
pjanevskiTT: +1 on this, moving this to firmware infor provider makes more sense

pjanevskiTT (approved):
No major changes needed for approve, so I will leave it but please resolve these comments and other that other team members leave on the pr

aleksamarkovicTT:
Related: #2895

tests/api/test_tt_device.cpp
anashTT: Fair enough, probably best to simply remove that section.

device/tt_device/tt_device.cpp
anashTT: Moved to Firmware Information Provider

Issue #2793, opened 2026-06-09
Simulation server: define the UMD ↔ host API (protocol + capability surface)


pjanevskiTT:
Part of #2677. Design: `docs/SIMULATION_SERVER_PROCESS.md` §8.

The contract clients use to talk to the host over the UNIX socket. Defines the protocol and operation set; per-capability specifics (incl. how TLB/sysmem split host vs client) are pinned down here.

## Scope
- Request/response framing over the UNIX socket (decide: reuse `nng` vs raw `AF_UNIX`).
- Capability surface: session (attach/detach); device description/identity (single-cluster topology + memory layout); memory access (read/write/multicast); TLB windows; sysmem (runtime, per-session); run/reset; execution/time; management (shutdown).
- Cross-cutting: per-client sessions; per-op atomicity (no cross-client transactions); request/response; backend-agnostic; host is **not** a client of its own API.
- Decide host-side state vs client-side handles for TLB/sysmem.

## Acceptance
- Documented protocol + shared header/IDL both serving and client build against.
- Review confirms it covers every device op UMD performs against silicon today.

broskoTT:
WIP

nbuncicTT:
First PR for API definition is merged, two more upcoming for basic functionality.

nbuncicTT:
Pending review

pjanevskiTT:
closed by multiple PRs

PR #2967, opened 2026-07-02
Simulation: UMD↔host wire protocol — messages + codec

pjanevskiTT:
### Issue
Part of #2677 (simulation server process). First slice of #2793 (define the UMD ↔ host API: protocol + capability surface). Builds on the merged SimulationClient / SimulationServerSocket / SimulationConnector work.

### Description
Adds the UMD ↔ simulation-host wire protocol: the request/response messages a client exchanges with a host ("the card") over its per-chip UNIX socket, plus a thin C++ codec. This is the contract both sides compile against; it wires neither side yet (the socket-backed client and the host serving layer are follow-ups).

Scope is intentionally narrow: one message pair for device-memory access — the bulk of the traffic — discriminated by a READ/WRITE command. The op set grows one message at a time as the server work lands (session/attach, TLB, sysmem, run/reset, …).

- FlatBuffers stays an implementation detail: generated types live in the `wire` namespace and only the .cpp includes the generated header; callers see plain structs.
- Stream framing is a transport concern, deliberately left out — it lands with the socket send/recv helpers.

### List of the changes
- New `simulation_server_protocol.{fbs,hpp,cpp}` — wire schema, public structs (`SimulationServerCommand/Request/Response`), and the FlatBuffers codec.
- New `tests/api/test_simulation_server_protocol.cpp` — 6 round-trip / malformed tests.
- `device/CMakeLists.txt` (multi-.fbs codegen + codec source), `tests/api/CMakeLists.txt` (register test). Both simulation-gated.

### Testing
`SimulationServerProtocol` (6 tests) pass; build + pre-commit clean.

### API Changes
New public header `umd/device/simulation/simulation_server_protocol.hpp`.

### DIFF
--- a/device/api/umd/device/simulation/simulation_server_protocol.hpp
+++ b/device/api/umd/device/simulation/simulation_server_protocol.hpp
@@ -1,0 +1,62 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+#include <vector>
+
+namespace tt::umd {
+
+// The one place that defines the UMD <-> simulation host wire protocol: the messages a UMD
+// client exchanges with a simulation host ("the card") over its per-chip UNIX socket, plus the
+// helpers that (de)serialize them. See simulation_server_protocol.fbs for the schema.
+//
+// Scope today is device-memory access -- the bulk of the traffic. It grows one message at a time
+// as the server work lands (session/attach, TLB windows, sysmem, run/reset, ...), at which point
+// the socket-backed client and the host serving layer marshal their operations through here.
+//
+// FlatBuffers stays an implementation detail (it lives in the .cpp); callers see only these plain
+// structs and free functions. Stream framing (length-prefixing messages on the socket) is a
+// transport concern and lives with the socket send/recv helpers, not here.
+
+// Direction of a device-memory access; mirrors wire::SimulationServerCommand in the schema.
+enum class SimulationServerCommand : int8_t {
+    Read = 0,
+    Write = 1,
+};
+
+// A device-memory access request addressed by (core x/y, address, size).
+struct SimulationServerRequest {
+    SimulationServerCommand command = SimulationServerCommand::Read;
+    uint32_t x = 0;
+    uint32_t y = 0;
+    uint64_t address = 0;
+    uint32_t size = 0;
+    // Bytes to write (length == size) for Write; empty for Read.
+    std::vector<uint8_t> data;
+};
+
+// The response to a SimulationServerRequest.
+struct SimulationServerResponse {
+    // 0 on success; nonzero signals a host-side failure.
+    int32_t status = 0;
+    // Bytes read (length == size on success) for Read; empty for Write.
+    std::vector<uint8_t> data;
+};
+
+// Serialize a message to a FlatBuffers payload.
+std::vector<uint8_t> encode(const SimulationServerRequest& request);
+std::vector<uint8_t> encode(const SimulationServerResponse& response);
+
+// Parse a message back from a FlatBuffers payload. The buffer is verified against the schema
+// before any field is read (it comes off a socket, so it may be malformed or truncated); a bad
+// or empty buffer throws error::RuntimeError rather than risking an out-of-bounds read.
+SimulationServerRequest decode_request(const uint8_t* data, size_t size);
+SimulationServerResponse decode_response(const uint8_t* data, size_t size);
+SimulationServerRequest decode_request(const std::vector<uint8_t>& bytes);
+SimulationServerResponse decode_response(const std::vector<uint8_t>& bytes);
+
+}  // namespace tt::umd
--- a/device/simulation/simulation_server_protocol.cpp
+++ b/device/simulation/simulation_server_protocol.cpp
@@ -1,0 +1,90 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "umd/device/simulation/simulation_server_protocol.hpp"
+
+#include <flatbuffers/flatbuffers.h>
+#include <fmt/format.h>
+
+#include "simulation_server_protocol_generated.h"
+#include "umd/device/utils/error.hpp"
+
+namespace tt::umd {
+
+namespace {
+
+// Verifies a FlatBuffers payload against schema table T (the buffer is untrusted socket input, so
+// this must run before any field access) and returns the parsed root. Throws on an empty, null,
+// or malformed/truncated buffer.
+template <typename T>
+const T* verify_and_get_root(const uint8_t* data, size_t size, const char* what) {
+    flatbuffers::Verifier verifier(data, size);
+    if (data == nullptr || !verifier.VerifyBuffer<T>(nullptr)) {
+        UMD_THROW(
+            error::RuntimeError,
+            fmt::format("Failed to verify simulation server protocol {} message ({} bytes)", what, size));
+    }
+    return flatbuffers::GetRoot<T>(data);
+}
+
+std::vector<uint8_t> to_bytes(const flatbuffers::FlatBufferBuilder& builder) {
+    return std::vector<uint8_t>(builder.GetBufferPointer(), builder.GetBufferPointer() + builder.GetSize());
+}
+
+}  // namespace
+
+std::vector<uint8_t> encode(const SimulationServerRequest& request) {
+    flatbuffers::FlatBufferBuilder builder;
+    auto data = builder.CreateVector(request.data);
+    builder.Finish(wire::CreateSimulationServerRequest(
+        builder,
+        static_cast<wire::SimulationServerCommand>(request.command),
+        request.x,
+        request.y,
+        request.address,
+        request.size,
+        data));
+    return to_bytes(builder);
+}
+
+std::vector<uint8_t> encode(const SimulationServerResponse& response) {
+    flatbuffers::FlatBufferBuilder builder;
+    auto data = builder.CreateVector(response.data);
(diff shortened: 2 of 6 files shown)


device/api/umd/device/simulation/simulation_server_protocol.hpp
```diff
@@ -0,0 +1,62 @@
...
+//
+// FlatBuffers stays an implementation detail (it lives in the .cpp); callers see only these plain
+// structs and free functions. Stream framing (length-prefixing messages on the socket) is a
+// transport concern and lives with the socket send/recv helpers, not here.
+
+// Direction of a device-memory access; mirrors wire::SimulationServerCommand in the schema.
+enum class SimulationServerCommand : int8_t {
+    Read = 0,
+    Write = 1,
+};
```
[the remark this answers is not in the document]
pjanevskiTT: Adopted `uint8_t` here — thanks. One correction on the rationale: in FlatBuffers `byte` is the **signed** 8-bit type (`int8_t`); `ubyte` is the unsigned one. So the original `int8_t` here actually matched the schema's `byte`, and switching only this side to `uint8_t` would have *introduced* a signedness mismatch against the generated `wire::SimulationServerCommand`. To keep both sides consistently unsigned I also changed the schema enum to `ubyte`, and added a `static_assert` bridging `tt::umd::SimulationServerCommand` and `wire::SimulationServerCommand` so any future drift is a compile error.

device/simulation/simulation_server_protocol.cpp
```diff
@@ -0,0 +1,90 @@
...
+    flatbuffers::FlatBufferBuilder builder;
+    auto data = builder.CreateVector(response.data);
+    builder.Finish(wire::CreateSimulationServerResponse(builder, response.status, data));
+    return to_bytes(builder);
+}
+
+SimulationServerRequest decode_request(const uint8_t* data, size_t size) {
+    const auto* fb = verify_and_get_root<wire::SimulationServerRequest>(data, size, "request");
+    SimulationServerRequest request;
+    request.command = static_cast<SimulationServerCommand>(fb->command());
```
nbuncicTT, line 61: Validate the command enum on decode using the generated MIN/MAX so an out-of-range command errors instead of casting to garbage.
```cpp
const auto cmd = fb->command();
if (flatbuffers::IsOutRange(cmd, wire::SimulationServerCommand_MIN, wire::SimulationServerCommand_MAX)) {
    UMD_THROW(error::RuntimeError, fmt::format("Unknown simulation server command {}", static_cast<int>(cmd)));
}
request.command = static_cast<SimulationServerCommand>(cmd);
```

device/api/umd/device/simulation/simulation_server_protocol.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+// A device-memory access request addressed by (core x/y, address, size).
+struct SimulationServerRequest {
```
nbuncicTT, line 32: nit: Reorder the request struct largest-first. The snippet bellow would drop 8 bytes of padding. Members are set by name so callers are unaffected.
```cpp
struct SimulationServerRequest {
    // Bytes to write (length == size) for Write; empty for Read.
    std::vector<uint8_t> data;
    uint64_t address = 0;
    uint32_t x = 0;
    uint32_t y = 0;
    uint32_t size = 0;
    SimulationServerCommand command = SimulationServerCommand::Read;
};
```

device/simulation/simulation_server_protocol.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+namespace {
+
+// Verifies a FlatBuffers payload against schema table T (the buffer is untrusted socket input, so
+// this must run before any field access) and returns the parsed root. Throws on an empty, null,
+// or malformed/truncated buffer.
+template <typename T>
+const T* verify_and_get_root(const uint8_t* data, size_t size, const char* what) {
+    flatbuffers::Verifier verifier(data, size);
+    if (data == nullptr || !verifier.VerifyBuffer<T>(nullptr)) {
```
nbuncicTT: nit: First check for `nullptr`, then create the `Verifier`:
```cpp
if (data == nullptr) {
    UMD_THROW(error::RuntimeError, fmt::format("Simulation server protocol {} message is null", what));
}
flatbuffers::Verifier verifier(data, size);
if (!verifier.VerifyBuffer<T>(nullptr)) {
// etc.
```

device/simulation/simulation_server_protocol.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+SimulationServerRequest decode_request(const uint8_t* data, size_t size) {
+    const auto* fb = verify_and_get_root<wire::SimulationServerRequest>(data, size, "request");
+    SimulationServerRequest request;
+    request.command = static_cast<SimulationServerCommand>(fb->command());
+    request.x = fb->x();
+    request.y = fb->y();
+    request.address = fb->address();
+    request.size = fb->size();
+    if (fb->data() != nullptr) {
```
nbuncicTT, line 66: nit: Maybe add an additional check because size and data are independent fields, so the user can leave them inconsistent. Enforce the write check (length == size) before returning:
```cpp
    if (request.command == SimulationServerCommand::Write && request.size != request.data.size()) {
        UMD_THROW(error::RuntimeError,
            fmt::format("Write payload length ({}) != size field ({})", request.data.size(), request.size));
    }
```

nbuncicTT (approved):

device/simulation/simulation_server_protocol.cpp
```diff
@@ -0,0 +1,90 @@
...
+template <typename T>
+const T* verify_and_get_root(const uint8_t* data, size_t size, const char* what) {
+    flatbuffers::Verifier verifier(data, size);
+    if (data == nullptr || !verifier.VerifyBuffer<T>(nullptr)) {
+        UMD_THROW(
+            error::RuntimeError,
+            fmt::format("Failed to verify simulation server protocol {} message ({} bytes)", what, size));
+    }
+    return flatbuffers::GetRoot<T>(data);
+}
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed — `verify_and_get_root()` now rejects `data == nullptr || size == 0` before constructing the `Verifier`. https://github.com/tenstorrent/tt-umd/commit/d15f024d

device/simulation/simulation_server_protocol.cpp
```diff
@@ -0,0 +1,90 @@
...
+    builder.Finish(wire::CreateSimulationServerRequest(
+        builder,
+        static_cast<wire::SimulationServerCommand>(request.command),
+        request.x,
+        request.y,
+        request.address,
+        request.size,
+        data));
+    return to_bytes(builder);
+}
```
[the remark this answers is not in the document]
pjanevskiTT: Addressed — `encode()` now runs `validate()` first, so a protocol-invalid request (e.g. Write with `size != data.size()`, or Read with a non-empty payload) throws instead of being serialized. https://github.com/tenstorrent/tt-umd/commit/d15f024d

device/simulation/simulation_server_protocol.cpp
```diff
@@ -0,0 +1,90 @@
...
+    request.command = static_cast<SimulationServerCommand>(fb->command());
+    request.x = fb->x();
+    request.y = fb->y();
+    request.address = fb->address();
+    request.size = fb->size();
+    if (fb->data() != nullptr) {
+        request.data.assign(fb->data()->begin(), fb->data()->end());
+    }
+    return request;
+}
```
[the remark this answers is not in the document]
pjanevskiTT: Addressed — `decode_request()` runs `validate()` after parsing: the command must be a known enum value, Read must carry no payload, and Write must have `size == data.size()`. Malformed requests are rejected rather than propagated upward. https://github.com/tenstorrent/tt-umd/commit/d15f024d

device/simulation/simulation_server_protocol.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+SimulationServerResponse decode_response(const uint8_t* data, size_t size) {
+    const auto* fb = verify_and_get_root<wire::SimulationServerResponse>(data, size, "response");
+    SimulationServerResponse response;
+    response.status = fb->status();
+    if (fb->data() != nullptr) {
+        response.data.assign(fb->data()->begin(), fb->data()->end());
+    }
+    return response;
+}
```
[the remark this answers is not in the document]
pjanevskiTT: Addressed — `validate()` enforces that an error response (nonzero `status`) carries no payload, on both encode and decode. https://github.com/tenstorrent/tt-umd/commit/d15f024d

device/simulation/simulation_server_protocol.cpp
pjanevskiTT: Out-of-range commands are rejected in the codec: both `decode_request()` and `encode()` run through `validate()`, which throws on any command that isn't Read/Write. I kept the single `validate()` guard rather than an inline `IsOutRange` so the same check covers encode and decode. https://github.com/tenstorrent/tt-umd/commit/d15f024d

device/api/umd/device/simulation/simulation_server_protocol.hpp
pjanevskiTT: Done — reordered largest-first (`data`, `address`, `x`, `y`, `size`, `command`); the padding is gone. https://github.com/tenstorrent/tt-umd/commit/f830a93b

device/simulation/simulation_server_protocol.cpp
pjanevskiTT: Fixed — the `nullptr`/empty check now runs before the `Verifier` is constructed. https://github.com/tenstorrent/tt-umd/commit/d15f024d

device/simulation/simulation_server_protocol.cpp
pjanevskiTT: Enforced in `validate()`: a Write whose `size` != `data.size()` throws, on both encode and decode. https://github.com/tenstorrent/tt-umd/commit/d15f024d

PR #2968, opened 2026-07-02
Move destructive hang tests into a dedicated test target

### LINKED ISSUE #2835: Move destructive hang tests into a dedicated test target
Split out from PR #2629 review ([discussion](https://github.com/tenstorrent/tt-umd/pull/2629#discussion_r3375564546)).

`tests/api/test_hang_detection.cpp` contains hang/timeout tests that are destructive — e.g. the per-op MMIO timeout test added in #2629 deliberately stalls device I/O — and are too expensive/risky to run on every CI pass alongside the normal API tests.

We should segregate these destructive hang tests into a dedicated test file/target (and/or gate them behind a separate CI job or gtest label) so they run on demand rather than on every pass. As @broskoTT suggested, `test_hang_detection.cpp` itself may be the right home for the segregated set.

Not addressed in #2629; tracking as a follow-up.

ajovanovicTT:
### Issue
#2835

### Description
Move the destructive NOC hang tests into their own test target so they no longer run on every CI pass.

### List of the changes
 - Added a hang_detection_tests target in tests/hang_detection, built from test_hang_detection.cpp
 - Removed test_hang_detection.cpp from api_tests and added the new target to the umd_tests aggregate
 - Moved the NOC hang warm reset test from test_warm_reset.cpp into the new target
 - Dropped the DISABLED prefixes, since the new target is never run automatically
 - Added a manual workflow_dispatch CI job that builds and runs the tests on a chosen card

### Testing
CI

### API Changes
This PR has no API changes.


.github/workflows/run-hang-detection-tests.yml
```diff
@@ -0,0 +1,135 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      card:
+        description: 'Card/runner to run the destructive hang tests on'
+        required: true
+        type: choice
+        default: tt-ubuntu-2204-n150-stable
+        options:
```
pjanevskiTT, line 13: I would add galaxies as well to these options. We care the most about the behavior of reset, NOC hang etc on those machines

tests/hang_detection/test_hang_detection.cpp
```diff
@@ -266,3 +267,74 @@ TEST_F(HangDetectionTest, DISABLED_TopologyDiscoveryRecordsNocHangHealthError) {
     WarmResetWithRecovery::warm_reset();
     init_device(pci_device_id);
 }
+
+TEST(WarmResetTest, TTDeviceWarmResetAfterNocHang) {
```
pjanevskiTT: maybe we can have this test in a separate file. It's testing warm reset more than NOC hang like the other tests

tests/hang_detection/test_hang_detection.cpp
```diff
@@ -9,13 +9,15 @@
 #include <cstdint>
```
pjanevskiTT, line 9: just make sure that all tests in this file are trying to recover the system after running each test, maybe we can put warm reset in the fixture functions if all these tests are in one fixture

.github/workflows/run-hang-detection-tests.yml
```diff
@@ -0,0 +1,135 @@
+# Destructive NOC hang / MMIO-timeout tests, run ON DEMAND only.
```
pjanevskiTT, line 1: reminder to open the issue to create a workflow to run these tests

pjanevskiTT (approved):

.github/workflows/run-hang-detection-tests.yml
ajovanovicTT: Thanks for reminder, already opened issue to wire as a nightly in GHA.

.github/workflows/run-hang-detection-tests.yml
```diff
@@ -0,0 +1,135 @@
...
+    env:
+      CC: clang-20
+      CXX: clang++-20
+
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          submodules: recursive
+
+      - name: Build hang_detection_tests
```
aleksamarkovicTT: Instead of another copy of a CMake invocation with different params, can you reuse build-tests.yml, parametrize workflow_call with a list of targets and pass hang_detection_tests.

PR #2969, opened 2026-07-02
Simulation: framed message transport over the host socket

pjanevskiTT:
### Issue
Part of #2677 (simulation server process). The transport half of the UMD ↔ host socket path; pairs with the wire protocol from #2967 (merged to `main`).

### Description
Adds `send_framed()` / `recv_framed()`: move one length-prefixed message (a 4-byte little-endian length, then the payload) over a connected UNIX-domain stream socket. A `SOCK_STREAM` socket has no message boundaries, so this framing is what lets the protocol messages travel intact between a UMD client and a simulation host.

- I/O goes through `asio::read`/`asio::write` on the connected `asio::local::stream_protocol::socket` — exact-N transfers, and asio sets `MSG_NOSIGNAL` on Linux, so writing to a closed peer surfaces as a thrown error rather than raising `SIGPIPE`. The socket must be in blocking mode (its owner clears any non-blocking flag asio leaves after connect/accept).
- `recv_framed` rejects an oversized length prefix (`MAX_FRAME_PAYLOAD_SIZE`, 1 GiB) before allocating, so a garbled/untrusted peer can't drive a multi-GB allocation.
- Payloads are opaque — encoding/decoding is the protocol layer's job.

### List of the changes
- New `device/simulation/simulation_server_transport.{hpp,cpp}`.
- New `tests/baremetal/test_simulation_server_transport.cpp` — 8 tests (single / empty / back-to-back / >1 MB payloads, peer-close, truncated, oversized-frame, SIGPIPE-safe send), using `asio::local::connect_pair`.
- `device/CMakeLists.txt` (always-built source), `tests/baremetal/CMakeLists.txt` (register test). Always built — asio is a core dependency; no simulation deps.

### Testing
`SimulationServerTransport` (8 tests) pass in the no-card baremetal lane; build + pre-commit clean.

### API Changes
None (internal `simulation/` header, not part of the public UMD API).

### DIFF
--- a/device/CMakeLists.txt
+++ b/device/CMakeLists.txt
@@ -148,6 +148,9 @@ target_sources(
         # (TT_UMD_BUILD_SIMULATION=OFF).
         simulation/simulation_server_socket.cpp
         simulation/simulation_client.cpp
+        # Length-prefixed message framing over the socket (uses asio, a core dependency; no
+        # FlatBuffers), so it is always built and unit-tested in the same no-card baremetal lane.
+        simulation/simulation_server_transport.cpp
 )
 
 if(TT_UMD_BUILD_SIMULATION)
--- a/device/simulation/simulation_server_transport.cpp
+++ b/device/simulation/simulation_server_transport.cpp
@@ -1,0 +1,100 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "simulation/simulation_server_transport.hpp"
+
+#include <fmt/format.h>
+
+#include <array>
+#include <asio.hpp>
+#include <system_error>
+
+#include "umd/device/utils/error.hpp"
+
+namespace tt::umd {
+
+namespace {
+
+using stream_protocol = asio::local::stream_protocol;
+
+constexpr size_t FRAME_HEADER_SIZE = sizeof(uint32_t);
+
+// Upper bound on a single framed payload. The length prefix comes off an untrusted socket, so a
+// garbled or malicious peer could otherwise advertise up to 4 GiB and make recv_framed() attempt a
+// huge allocation. Cap it well below that at a size no legitimate device-memory transfer reaches;
+// anything larger is rejected before allocating.
+constexpr uint32_t MAX_FRAME_PAYLOAD_SIZE = 1u << 30;  // 1 GiB
+
+// asio::write/asio::read transfer exactly n bytes or set ec (short transfer on a closed peer, etc.);
+// convert any failure to a RuntimeError. asio sets MSG_NOSIGNAL on the send, so a closed peer is an
+// error here, never a SIGPIPE.
+void write_exact(stream_protocol::socket& socket, const uint8_t* data, size_t n, const char* what) {
+    std::error_code ec;
+    asio::write(socket, asio::buffer(data, n), ec);
+    if (ec) {
+        UMD_THROW(
+            error::RuntimeError, fmt::format("Simulation server transport failed to write {}: {}", what, ec.message()));
+    }
+}
+
+void read_exact(stream_protocol::socket& socket, uint8_t* data, size_t n, const char* what) {
+    std::error_code ec;
+    asio::read(socket, asio::buffer(data, n), ec);
+    if (ec) {
+        UMD_THROW(
+            error::RuntimeError, fmt::format("Simulation server transport failed to read {}: {}", what, ec.message()));
+    }
+}
+
+}  // namespace
+
+void send_framed(stream_protocol::socket& socket, const std::vector<uint8_t>& payload) {
+    // Bounded so the length always fits the uint32_t prefix and matches what recv_framed() will
+    // accept; anything larger is a programming error rather than a truncation.
+    UMD_ASSERT(
+        payload.size() <= MAX_FRAME_PAYLOAD_SIZE,
+        error::RuntimeError,
+        fmt::format(
+            "Simulation server transport payload too large to frame ({} bytes, max {})",
+            payload.size(),
+            MAX_FRAME_PAYLOAD_SIZE));
+    const uint32_t length = static_cast<uint32_t>(payload.size());
+    const std::array<uint8_t, FRAME_HEADER_SIZE> header = {
+        static_cast<uint8_t>(length & 0xFF),
+        static_cast<uint8_t>((length >> 8) & 0xFF),
+        static_cast<uint8_t>((length >> 16) & 0xFF),
+        static_cast<uint8_t>((length >> 24) & 0xFF),
+    };
+    // Write the header and payload separately so a large payload is not copied into a combined
+    // frame buffer first.
+    write_exact(socket, header.data(), header.size(), "frame header");
+    if (!payload.empty()) {
+        write_exact(socket, payload.data(), payload.size(), "frame payload");
+    }
+}
+
+std::vector<uint8_t> recv_framed(stream_protocol::socket& socket) {
+    std::array<uint8_t, FRAME_HEADER_SIZE> header;
+    read_exact(socket, header.data(), header.size(), "frame header");
+    const uint32_t length = static_cast<uint32_t>(header[0]) | (static_cast<uint32_t>(header[1]) << 8) |
+                            (static_cast<uint32_t>(header[2]) << 16) | (static_cast<uint32_t>(header[3]) << 24);
+
+    // Reject an implausible length from the untrusted prefix before allocating, so a bad peer
+    // can't drive a multi-GB allocation (or std::bad_alloc).
+    UMD_ASSERT(
+        length <= MAX_FRAME_PAYLOAD_SIZE,
+        error::RuntimeError,
+        fmt::format(
+            "Simulation server transport received an oversized frame ({} bytes, max {})",
+            length,
+            MAX_FRAME_PAYLOAD_SIZE));
+
+    std::vector<uint8_t> payload(length);
+    if (length > 0) {
+        read_exact(socket, payload.data(), length, "frame payload");
+    }
+    return payload;
+}
+
+}  // namespace tt::umd
--- a/device/simulation/simulation_server_transport.hpp
+++ b/device/simulation/simulation_server_transport.hpp
@@ -1,0 +1,35 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <asio/local/stream_protocol.hpp>
+#include <cstdint>
+#include <vector>
+
+namespace tt::umd {
+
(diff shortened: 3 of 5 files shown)


device/simulation/simulation_server_transport.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include <cstring>
+#include <limits>
+
+#include "umd/device/utils/error.hpp"
+
+namespace tt::umd {
+
+namespace {
+
+constexpr size_t kFrameHeaderSize = sizeof(uint32_t);
```
nbuncicTT: nit: why do we have `kFrameHeaderSize` as the name, what does k symbolize? Why isn't it just `frameHeaderSize`? or `headerFrameSize` and since this is a constant, this should be in caps and snake case: `FRAME_HEADER_SIZE`

tests/baremetal/test_simulation_server_transport.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+#include "simulation/simulation_server_transport.hpp"
+
+using namespace tt::umd;
+
+namespace {
+
+// A connected UNIX SOCK_STREAM pair; both ends are closed on destruction. Mirrors the framed
+// client/host connection without needing a bound socket path.
+struct SocketPair {
```
nbuncicTT: Use `gtest`'s harness for these cases instead of introducing separate classes, then you can define what happens on `SetUp`, `TearDown`, etc.

device/simulation/simulation_server_transport.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+// Writes exactly n bytes, looping over partial writes and retrying on EINTR. Throws on a socket
+// error or if the peer is gone.
+void write_all(int fd, const uint8_t* data, size_t n) {
```
nbuncicTT: Instead of relying on raw fd's and writing boiler-plate code, why not use asio's `asio::write`?

device/simulation/simulation_server_transport.cpp
```diff
@@ -0,0 +1,107 @@
...
+                    "Simulation server transport write failed after {} of {} bytes: {}", written, n, strerror(errno)));
+        }
+        written += static_cast<size_t>(w);
+    }
+}
+
+// Reads exactly n bytes, looping over partial reads and retrying on EINTR. A read of 0 means the
+// peer closed the connection; that is an error here because the caller is mid-message (or expects
+// one). Throws on a socket error or a closed peer.
+void read_exact(int fd, uint8_t* data, size_t n) {
```
nbuncicTT: Same goes here, instead of relying on raw fd's and writing boiler-plate code, why not use asio's `asio::read`?

nbuncicTT (commented):
`umd_asio_lib` is linked into `tt-umd` **unconditionally** (only FlatBuffers is sim-gated, and framing never touches it), and both endpoints are already asio sockets so every real caller reaches around asio via `native_handle()` to use this raw-fd API, and there's no non-asio fd consumer on this path.

Since this is an **internal** header, depending on asio is fine; our public header is already decoupled via pimpl for the server side. 

Proposal:

- Take `asio::local::stream_protocol::socket&` instead of `int fd`.
- Replace `read_exact`/`write_all` with `asio::read`/`asio::write` using the **`error_code` overload + `UMD_THROW`** pattern already in `simulation_server_socket.cpp` — `transfer_all()` handles partial I/O + `EINTR`, EOF is `asio::error::eof`, and the overload returns `bytes_transferred` so we keep the "N of M bytes" messages.
- Keep the size guard + LE codec

device/simulation/simulation_server_transport.cpp
```diff
@@ -0,0 +1,107 @@
...
+    read_exact(fd, header, kFrameHeaderSize);
+    const uint32_t length = static_cast<uint32_t>(header[0]) | (static_cast<uint32_t>(header[1]) << 8) |
+                            (static_cast<uint32_t>(header[2]) << 16) | (static_cast<uint32_t>(header[3]) << 24);
+
+    std::vector<uint8_t> payload(length);
+    if (length > 0) {
+        read_exact(fd, payload.data(), length);
+    }
+    return payload;
+}
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed — `recv_framed` now rejects a length above `MAX_FRAME_PAYLOAD_SIZE` (1 GiB) before allocating, so an untrusted prefix can't drive a multi-GB allocation.

tests/baremetal/test_simulation_server_transport.cpp
```diff
@@ -0,0 +1,108 @@
...
+    SocketPair sp;
+    const uint8_t header[4] = {10, 0, 0, 0};  // claims a 10-byte payload
+    ASSERT_EQ(::write(sp.fds[0], header, sizeof(header)), static_cast<ssize_t>(sizeof(header)));
+    const uint8_t partial[3] = {1, 2, 3};  // but only 3 bytes follow
+    ASSERT_EQ(::write(sp.fds[0], partial, sizeof(partial)), static_cast<ssize_t>(sizeof(partial)));
+    ::close(sp.fds[0]);
+    sp.fds[0] = -1;
+
+    EXPECT_THROW(recv_framed(sp.fds[1]), std::exception);
+}
```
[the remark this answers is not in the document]
pjanevskiTT: Added `RecvThrowsOnOversizedFrame` (header advertises ~4 GiB → throws at the length check, before allocating).

device/simulation/simulation_server_transport.cpp
```diff
@@ -0,0 +1,107 @@
...
+// Writes exactly n bytes, looping over partial writes and retrying on EINTR. Throws on a socket
+// error or if the peer is gone.
+void write_all(int fd, const uint8_t* data, size_t n) {
+    size_t written = 0;
+    while (written < n) {
+        const ssize_t w = ::write(fd, data + written, n - written);
+        if (w < 0) {
+            if (errno == EINTR) {
+                continue;
+            }
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed — `send_framed` now uses `::send(..., MSG_NOSIGNAL)` so a closed peer yields `EPIPE` (a thrown error) instead of `SIGPIPE`. Moved the fix into this PR since `send_framed` is introduced here.

device/simulation/simulation_server_transport.cpp
pjanevskiTT: Renamed to `FRAME_HEADER_SIZE` (SCREAMING_SNAKE_CASE, matching the repo convention for constants).

tests/baremetal/test_simulation_server_transport.cpp
pjanevskiTT: Converted to a gtest fixture — `SimulationServerTransportTest` with `SetUp`/`TearDown` owning the socketpair.

ajovanovicTT (commented):
Nice, please just address Nenad's comments to use asio::read/write instead of raw fd read/write, otherwise I’m OK with the changes.

device/simulation/simulation_server_transport.cpp
pjanevskiTT: Done — switched to `asio::read`/`asio::write`; the transport now takes an `asio::local::stream_protocol::socket&` instead of a raw fd. Confirmed asio sets `MSG_NOSIGNAL` on Linux (`socket_ops.ipp`), so this keeps the SIGPIPE-safety from the other comment.

device/simulation/simulation_server_transport.cpp
pjanevskiTT: Done — switched to `asio::read`/`asio::write`; the transport now takes an `asio::local::stream_protocol::socket&` instead of a raw fd. Confirmed asio sets `MSG_NOSIGNAL` on Linux (`socket_ops.ipp`), so this keeps the SIGPIPE-safety from the other comment.

device/simulation/simulation_server_transport.hpp
```diff
@@ -0,0 +1,35 @@
...
+// owner clears any non-blocking flag asio leaves after connect/accept). Payloads are opaque here;
+// encoding/decoding is the protocol layer's job.
+//
+// Every failure (peer closed mid-message, oversized frame, socket error) throws error::RuntimeError.
+
+// Length-prefix and write one whole message to the socket.
+void send_framed(asio::local::stream_protocol::socket& socket, const std::vector<uint8_t>& payload);
+
+// Read exactly one length-prefixed message from the socket, blocking until it is complete.
+std::vector<uint8_t> recv_framed(asio::local::stream_protocol::socket& socket);
```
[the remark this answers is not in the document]
pjanevskiTT: The header doc was rewritten for the asio interface -- it no longer claims asio-free or references native_handle(). Taking an `asio::local::stream_protocol::socket&` is intentional now (asio handles exact-N transfers and MSG_NOSIGNAL).

device/CMakeLists.txt
```diff
@@ -148,6 +148,9 @@ target_sources(
         # (TT_UMD_BUILD_SIMULATION=OFF).
         simulation/simulation_server_socket.cpp
         simulation/simulation_client.cpp
+        # Pure POSIX stream framing (length-prefixed message send/recv); no asio or FlatBuffers,
+        # so always built and unit-tested in the same no-card baremetal lane.
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed -- updated the comment: the framing uses asio (a core dependency), no FlatBuffers; still always built and no-card-testable.

device/simulation/simulation_server_transport.cpp
```diff
@@ -0,0 +1,99 @@
...
+    // accept; anything larger is a programming error rather than a truncation.
+    UMD_ASSERT(
+        payload.size() <= MAX_FRAME_PAYLOAD_SIZE,
+        error::RuntimeError,
+        fmt::format(
+            "Simulation server transport payload too large to frame ({} bytes, max {})",
+            payload.size(),
+            MAX_FRAME_PAYLOAD_SIZE));
+    const uint32_t length = static_cast<uint32_t>(payload.size());
+    const uint8_t header[FRAME_HEADER_SIZE] = {
```
nbuncicTT: Use `std::array` instead of C-style arrays.

device/simulation/simulation_server_transport.cpp
```diff
@@ -0,0 +1,99 @@
...
+        static_cast<uint8_t>(length & 0xFF),
+        static_cast<uint8_t>((length >> 8) & 0xFF),
+        static_cast<uint8_t>((length >> 16) & 0xFF),
+        static_cast<uint8_t>((length >> 24) & 0xFF),
+    };
+    // Write the header and payload separately so a large payload is not copied into a combined
+    // frame buffer first.
+    write_exact(socket, header, FRAME_HEADER_SIZE, "frame header");
```
nbuncicTT: nit: use `header.size()` instead of `FRAME_HEADER_SIZE`, that way the code always passes the correct size (for the header being passed in)

nbuncicTT (approved):

Issue #2603, opened 2026-05-11
Setup TTSim perf tests


pjanevskiTT:
Setup TTSim microbenchmark tests. We would like to do this in order to avoid regressions on how long does it take for UMD to start for TTSim. There have been PRs in a past that have regressed this, so we would like to be able to take care of that

Basic tests we care about at the moment would be 

- UMD startup (creating Cluster)
- Basic write_to_device
- Basic read_from_device

PR #2972, opened 2026-07-02
Refactor simulation cluster options handling in tests

ajovanovicTT:
### Issue
Followup from #2603

### Description
Extract the simulator ClusterOptions setup into shared test helpers so it can be reused across test suites, following up on a review comment.

### List of the changes
 - Added shared sim ClusterOptions helpers to device_test_utils.hpp
 - Reused them in make_default_test_cluster, test_simulator.cpp, and test_device_io.cpp

### Testing
CI.

### API Changes
This PR has no API changes.


tests/test_utils/device_test_utils.hpp
```diff
@@ -40,6 +40,22 @@ inline bool has_remote_chips() {
...
+    options.target_devices = {0};
+    options.simulator_directory = simulator_directory;
+}
+
+inline ClusterOptions get_default_sim_cluster_options(
+    const std::filesystem::path& simulator_directory,
+    std::optional<uint32_t> num_host_mem_ch_per_mmio_device = std::nullopt) {
+    ClusterOptions options;
+    apply_default_sim_cluster_options(options, simulator_directory);
+    options.num_host_mem_ch_per_mmio_device = num_host_mem_ch_per_mmio_device;
```
pjanevskiTT: why is this not in apply function as well?

pjanevskiTT (approved):

tests/test_utils/device_test_utils.hpp
ajovanovicTT: It is due to its usage in make_default_test_cluster where it would overwrite num_host_mem_ch_per_mmio_device with 0. Tried to find some better and more general solution to avoid adding two functions but at the end, this looks more elegant than trying to unify them into one function.

ajovanovicTT: Pushed commit which introduces only one method, so please let me know if you think previous approach was better.

tests/microbenchmark/benchmarks/simulator/test_simulator.cpp
pjanevskiTT, line 1: this looks like a bad rebase now

pjanevskiTT (approved):

Issue #2970, opened 2026-07-02
Introduce compile artifact caching


aleksamarkovicTT:
We should integrate a compile artifact caching utility, such as SCCache, to our repo.
Caching tools:
- CCache - older, used in tt-metal, C/C++ only, harder to integrate
- SCCache - modern, more versatile, multiple languages. cloud-oriented, easier to integrate, more storage backends.

I prefer SCCache, I've seen it handle much larger repos than tt-metal, less hassle than CCache.

When to cache:
- CI, when building UMD.
  - build-tests.yaml (Release, ASAN, TSAN, Ubuntu 22, Tracy.. lots of combinations)
  - build-device.yaml (Release, Ubuntu 20-24, Fedora.. also lots of combinations) 
- Developer environment
  - Using local storage, ideally in `/localdev/`.

Currently, we should only cache code from `main`, otherwise the cache size will grow pretty fast, hard to estimate.

Where to cache:
- GHA Cache
  - 10 GB available for free, not sure if we have more available, figure out w/ org admin.
  - Easiest to integrate.
  - Not available on dev machines (investigate).
  - Auto-cleanup of unused artifacts in 7 days.
- Redis
  - Hosted by us, only available on workers inside VPN, 50 GB taken up completely by Metal.
  - Getting decomissioned soon.
- Garage S3
  - Hosted by us, only available on workers inside VPN, currently-used by tt-metal.
  - Will be available for dev machines.
  - @blozano-tt volunteered to provide our own bucket.
- AWS S3
  - Hosted in cloud, see company policy on cloud costs.
  - Can be used by any worker, any dev machine.
  - Best but priciest solution.
- /localdev
  - Only for dev machines. 
  - Fast, lots of storage, but local to one machine.

Tasklist:

- [x] Integrate in `build-tests.yml`
- [x] Integrate in `build-device.yml` (might not be a good idea)
- [ ] Integrate on dev machines using local storage.
- [ ] Compare GHA to S3 bucket once it becomes available.
- [ ] Multi-level cache local and S3 on dev machines, once it becomes available.

PR #2976, opened 2026-07-03
Introduce sccache for build-tests.yml using GHA Cache

aleksamarkovicTT:
### Issue
Begins #2970 

### Description
Introduce sccache for build-tests.yml using GHA Cache.
- SCCache is obtained by [sccache-action](https://github.com/marketplace/actions/sccache-action), seamlessly downloads it in the environment. 
- `use-sccache` input is used to easily gate sccache in situations where we need to disable it.
- `SCCACHE_BASEDIRS` is used to cut the absolute paths used by SCCache and make the cache shareable with other environments, like /localdev which situate tt-umd on a different absolute path.
- `SCCACHE_GHA_ENABLED` enabled GHA Cache as the storage provider.
- Changing `SCCACHE_GHA_VERSION` easily purges the existing cache. 1 is a sentinel value I picked at random.
- `SCCACHE_GHA_RW_MODE` makes the cache read-only when the workflow is not being run on main. PRs will fill the cache pretty quickly with artifacts that probably won't be used very often.
- SCCache is wired in with CMake cache options.


.github/workflows/build-tests.yml
```diff
@@ -115,6 +125,18 @@ jobs:
...
+        uses: mozilla-actions/sccache-action@v0.0.10
+
+      - name: Configure sccache to access GHA cache
+        if: ${{ inputs.use-sccache }}
+        uses: actions/github-script@v7
+        with:
+          script: |
+            core.exportVariable('ACTIONS_RESULTS_URL', process.env.ACTIONS_RESULTS_URL || '');
+            core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env.ACTIONS_RUNTIME_TOKEN || '');
+
```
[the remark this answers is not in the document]
aleksamarkovicTT: No need for this, it's observed that caching works.

.github/workflows/build-tests.yml
```diff
@@ -105,6 +109,12 @@ jobs:
       # CI runners may lack invariant TSC support (required by Tracy for high-res timestamps).
       # This skips Tracy's runtime TSC check; harmless when Tracy is disabled.
       TRACY_NO_INVARIANT_CHECK: 1
+      # Env. vars related to SCCache:
+      SCCACHE_GHA_ENABLED: ${{ inputs.use-sccache }}
+      SCCACHE_GHA_VERSION: 1
+      # Only the main branch may write to the sccache; all other branches read only.
+      SCCACHE_GHA_RW_MODE: ${{ github.ref == 'refs/heads/main' && 'READ_WRITE' || 'READ_ONLY' }}
+      SCCACHE_BASEDIRS: ${{ github.workspace }}
```
nbuncicTT: `SCCACHE_BASEDIRS: ${{ github.workspace }}` won't work in a container job because `github.workspace` resolves to the host path (`/home/runner/work/...`) but the code compiles at the in-container path (`/__w/...`), so the prefix never matches and paths are silently never stripped ([actions/runner#2058](https://github.com/actions/runner/issues/2058)); either set it via `echo "SCCACHE_BASEDIRS=$GITHUB_WORKSPACE" >> "$GITHUB_ENV"` in a step, or drop it since CI-internal caching works without it.

nbuncicTT (approved):

.github/workflows/build-tests.yml
aleksamarkovicTT: The problem is that build-device.yml builds UMD at a different path and would result in different cache entries, so it can't be dropped. I'll set it another step.

ajovanovicTT (approved):

PR #2979, opened 2026-07-05
Allow overriding x86 -march for Ivy Bridge hosts

nkapreTT:
## Summary
- add TT_X86_MARCH CMake cache variable for x86_64 host builds
- keep the existing default at x86-64-v3 so modern hosts retain current optimizations
- allow older hosts to configure with -DTT_X86_MARCH=ivybridge

## Motivation
Ivy Bridge hosts support AVX but not the full x86-64-v3 baseline, so prebuilt/default host code can fault with illegal instructions. This keeps the optimized default while making source builds possible on older compatible hosts.


aleksamarkovicTT (approved):
Makes sense, thanks for the PR. 
Optionally, you can validate the cache variable, but I don't think there's a need.

PR #2983, opened 2026-07-06
Add regression checks for TTSim microbenchmark tests

### LINKED ISSUE #2604: Add regressions checks for TTSim perf tests
Depends on completing #2603. After adding these test we would like to add regression checks to these tests. This should be the ideal first group of tests to test our regression workflow. We currently have the problem of not having a dedicated perf machine, which makes it harder for to do this for silicon, since the results that we get are not that reliable. Since TTSim tests run on more powerful CPU machines less difference between runs is expected.

ajovanovicTT:
### Issue
#2604

### Description
Add the UMD simulator microbenchmarks to the perf regression pipeline. The ttsim workflow now runs the simulator startup and basic IO benchmarks, labels and uploads results per arch, and checks them against per-arch baselines.

### List of the changes
 - Add a board-type override to gather_host_specs so simulator runs are labeled sim-wh, sim-bh, sim-qsr instead of unknown for simulator runners
 - In run-ttsim-tests, build umd_microbenchmark and run MicrobenchmarkSim per arch, uploading benchmark-json-sim artifacts
 - Add a regression-check job to run-ttsim-tests that compares the sim results against the sim baselines
 - Add the calibrated simulator baselines
 - Scope the regression summary to archs present in the current run, so hardware and ttsim runs no longer show empty columns for each other

### Testing
CI.

### API Changes
This PR has no API changes.


tests/microbenchmark/baselines/sim-qsr.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+metadata:
+  schema_version: 1
+  arch: "sim-qsr"
+  runner_hostname: "tt-ubuntu-2204-large-stable-gg6km-runner-8fdp9"
+  calibrated_at: "2026-07-06"
+  notes: "Updated from a CI benchmark JSON artifact via tests/microbenchmark/tools/update_baseline.py."
+
+ClusterConstructorSim:
+  default: { median_value: 181.5, tolerance_pct: 5 }
```
pjanevskiTT: should we also have these benchmarks for all the board types? Simulator is soon going to introduce multiple topologies. Is this scheme supporting this as well?

pjanevskiTT (commented):
If we can easily support multiple topologies for sim as well, I am happy to approve this PR

tests/microbenchmark/baselines/sim-wh.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+metadata:
+  schema_version: 1
+  arch: "sim-wh"
+  runner_hostname: "tt-ubuntu-2204-large-stable-gg6km-runner-8fdp9"
+  calibrated_at: "2026-07-06"
+  notes: "Updated from a CI benchmark JSON artifact via tests/microbenchmark/tools/update_baseline.py."
+
+ClusterConstructorSim:
+  default: { median_value: 198.5, tolerance_pct: 5 }
```
pjanevskiTT: None of the sim cases set `gate: true`, and `summarize_regressions.main()` only exits non-zero on gated breaches — so a real sim regression will render in the Step Summary but keep the job green. This matches the ungated hardware n150 pilot, so it may be intentional, but since the PR is titled "regression checks," worth confirming: should the tight cases here (and the two `DeviceIOSim` cases at 5%) get `gate: true` so they actually fail CI? Same applies to `sim-bh.yaml` / `sim-qsr.yaml`.

tests/microbenchmark/tools/summarize_regressions.py
```diff
@@ -322,7 +322,8 @@ def render_summary(current: dict, baselines: dict) -> tuple[str, list, list]:
     for title in test_titles:
         for case_entry in baselines[title].values():
             archs.update(case_entry.keys())
-    arch_order = sorted(archs)
+    # Only archs present in this run; the benchmark and ttsim workflows each cover a different subset.
+    arch_order = sorted(a for a in archs if a in current)
```
pjanevskiTT, line 324: This correctly drops the empty cross-workflow columns, but it's shared with the hardware path (via `compare_to_baseline.py`). Side effect: an arch whose detected `BoardType` ever drifts from its baseline label now disappears from the table entirely, instead of showing as `— (no result)`. Can't happen for sim (label is hardcoded via `--board-type`), and it's low risk for hardware — just flagging that a detection/label mismatch is now silent rather than visible.

tests/microbenchmark/tools/gather_host_specs.py
```diff
@@ -374,10 +374,20 @@ def main():
...
+            "Override the auto-detected board type, e.g. 'sim-wh' for a simulator run. "
+            "When omitted, the board type is auto-detected via tt_umd."
+        ),
+    )
     args = parser.parse_args()
 
     # 1. Gather the specs using manual methods
-    specs = gather_specs(board_type=detect_board_type())
+    board_type = args.board_type if args.board_type is not None else detect_board_type()
+    specs = gather_specs(board_type=board_type)
```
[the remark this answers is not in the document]
ajovanovicTT: Not the real problem since this script is only used from CI, where that arg is configured properly.

.github/workflows/run-ttsim-tests.yml
```diff
@@ -168,6 +168,68 @@ jobs:
...
+        run: |
+          python3 -m ensurepip --upgrade
+          python3 -m pip install --no-cache-dir pandas psutil
+
+      - name: Run UMD simulator microbenchmarks (wormhole)
+        timeout-minutes: 2
+        env:
+          TT_UMD_SIMULATOR: ${{ github.workspace }}/sim_wh/libttsim.so
+          UMD_MICROBENCHMARK_RESULTS_PATH: /tmp/umd_bench_sim_wh
+        run: |
```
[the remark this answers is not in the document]
ajovanovicTT: Job runs inside container so this is not a problem.

.github/workflows/run-ttsim-tests.yml
```diff
@@ -168,6 +168,68 @@ jobs:
...
+          name: benchmark-json-sim-wh
+          path: /tmp/umd_bench_sim_wh/*.json
+          if-no-files-found: error
+
+      - name: Run UMD simulator microbenchmarks (blackhole)
+        timeout-minutes: 2
+        env:
+          TT_UMD_SIMULATOR: ${{ github.workspace }}/sim_bh/libttsim.so
+          UMD_MICROBENCHMARK_RESULTS_PATH: /tmp/umd_bench_sim_bh
+        run: |
```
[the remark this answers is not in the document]
ajovanovicTT: Same as in the conversation above.

.github/workflows/run-ttsim-tests.yml
```diff
@@ -168,6 +168,68 @@ jobs:
...
+          name: benchmark-json-sim-bh
+          path: /tmp/umd_bench_sim_bh/*.json
+          if-no-files-found: error
+
+      - name: Run UMD simulator microbenchmarks (quasar)
+        timeout-minutes: 2
+        env:
+          TT_UMD_SIMULATOR: ${{ github.workspace }}/sim_qsr/libttsim.so
+          UMD_MICROBENCHMARK_RESULTS_PATH: /tmp/umd_bench_sim_qsr
+        run: |
```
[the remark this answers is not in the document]
ajovanovicTT: Same as in the conversation above.

tests/microbenchmark/baselines/sim-wh.yaml
ajovanovicTT: Will leave this non-gating for now and look at how it behaves in the CI when merged.

tests/microbenchmark/tools/summarize_regressions.py
ajovanovicTT: compare_to_baseline.py is local only and never in CI, so it is not the concern. The change reaches hardware directly via regression-check.yml running summarize_regressions.py. That path is low risk (pinned archs, stable label strings, and the unknown detection WARN), so keeping as is for now.

tests/microbenchmark/baselines/sim-qsr.yaml
ajovanovicTT: Yes, the scheme keys on the board-type label, so a new sim topology is just a new label, on example: sim-n300, a baselines/sim-n300.yaml, and a run step passing --board-type sim-n300 , so no changes to gather, summarize, or regression-check. The explicit --board-type handles this.

pjanevskiTT (approved):

PR #2984, opened 2026-07-06
Simulation: serve client requests on the host socket

pjanevskiTT:
### Issue
Part of #2677 (simulation server process). Wires the wire protocol (#2967 / #2793) and framed transport (#2969) into `SimulationServerSocket` so the host can serve client requests, not just indicate presence. Stacked on #2969.

### Description
`SimulationServerSocket` gains **`serve(RequestHandler)`** (`std::function<std::vector<uint8_t>(const std::vector<uint8_t>&)>`), which starts the accept loop; each accepted connection is served on its own thread running `recv_framed → handler → send_framed` until the peer disconnects.

**Serving is deliberately split from binding.** `try_create`/`create` only *bind* (claim the path, connectable for liveness); `serve()` is called later, once the owner's backend — which the handler dispatches into — is ready. The backend does not exist at `try_create` time, so a construction-time handler would have nothing to dispatch into; the host device binds its handler in `adopt_socket()` once the backend is up (see the follow-up host-wiring PR).

Design notes:
- **Protocol-agnostic:** the socket moves opaque payloads and calls the handler; decoding, dispatch, and encoding live in the handler (simulation-gated code). This keeps `SimulationServerSocket` always-built and FlatBuffers-free.
- **Per-connection blocking threads** reuse the #2969 transport and support multiple clients. The handler is installed *before* the accept loop starts, so no connection thread races on it.
- **Teardown** marks stopping (so a racing accept starts no new session), `shutdown()`s every live connection so its serving thread unblocks from `recv_framed`, joins the threads, then closes the acceptor.

Also folds in a small transport hardening (its own commit): `send_framed` uses `::send(..., MSG_NOSIGNAL)` so writing to a closed peer fails with `EPIPE` (a thrown error) rather than raising `SIGPIPE` — reachable now that the serving loop writes to remote clients.

### List of the changes
- `device/simulation/simulation_server_socket.{hpp,cpp}` — `serve(handler)`; accept loop moved out of `bind_and_listen` into `serve()`; `try_create`/`create` are path-only.
- `device/simulation/simulation_server_transport.cpp` — `send_framed` uses `MSG_NOSIGNAL`.
- `tests/baremetal/test_simulation_server_socket.cpp` — serving tests (serve-through-handler, sequential, multiple clients, clean teardown) + `ServesAfterDeferredServe` (presence-only → serve()).
- `tests/baremetal/test_simulation_server_transport.cpp` — send-to-closed-peer throws (SIGPIPE regression).

### Testing
`SimulationServerSocketTest` (14) and `SimulationServerTransport` (7) pass in the no-card baremetal lane; build + pre-commit clean.

### API Changes
None (internal `simulation/` header). `SimulationServerSocket` gains `serve()`; the factories are path-only.

### DIFF
--- a/device/simulation/simulation_server_socket.cpp
+++ b/device/simulation/simulation_server_socket.cpp
@@ -8,10 +8,14 @@
 #include <sys/un.h>  // sockaddr_un::sun_path, for the path-length guard
 
 #include <asio.hpp>
+#include <atomic>
 #include <memory>
+#include <mutex>
 #include <system_error>
 #include <thread>
+#include <vector>
 
+#include "simulation/simulation_server_transport.hpp"
 #include "umd/device/utils/error.hpp"
 
 namespace tt::umd {
@@ -42,6 +46,20 @@ struct SimulationServerSocket::Impl {
     asio::io_context io;
     stream_protocol::acceptor acceptor{io};
     std::thread io_thread;
+
+    // Per-connection serving state, held only while a handler is set. Finished connections (peer
+    // disconnected, flagged by the serving thread) are reaped so this doesn't grow unbounded across
+    // connect/disconnect cycles. Guarded: registered/reaped on the io_thread, torn down on the
+    // destructor thread.
+    struct Connection {
+        std::shared_ptr<stream_protocol::socket> socket;
+        std::thread thread;
+        std::shared_ptr<std::atomic<bool>> finished;
+    };
+
+    std::mutex connections_mutex;
+    std::vector<Connection> connections;
+    bool stopping = false;
 };
 
 SimulationServerSocket::SimulationServerSocket(const std::filesystem::path& socket_path) :
@@ -52,6 +70,8 @@ std::unique_ptr<SimulationServerSocket> SimulationServerSocket::try_create(const
     if (!socket->bind_and_listen()) {
         return nullptr;
     }
+    // Bound and connectable now (presence/liveness); serving begins when serve() is called with a
+    // handler bound to the owner's backend.
     return socket;
 }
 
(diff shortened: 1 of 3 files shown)


device/simulation/simulation_server_socket.hpp
```diff
@@ -55,18 +78,21 @@ class SimulationServerSocket {
...
 
-    // Re-arms the async accept; connections carry no requests yet (owner-only), so each
-    // accepted socket is dropped immediately.
+    // Re-arms the async accept; each accepted connection is handed to a serving thread. Only ever
+    // runs while serving (serve() started it), so request_handler_ is always set here.
     void do_accept();
 
     std::filesystem::path socket_path_;
+    // Opaque request/reply handler; empty means presence-only (accept and drop).
+    RequestHandler request_handler_;
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed: the handler comment now says an empty handler means the accept loop never starts (serve() starts it), so the socket is presence-only and accepts nothing -- matching the deferred-serve design.

device/simulation/simulation_server_socket.cpp
```diff
@@ -64,14 +78,57 @@ std::unique_ptr<SimulationServerSocket> SimulationServerSocket::create(const std
...
             return;
         }
+
+        // Serve the connection on its own thread until the peer disconnects. do_accept only runs
+        // while serving, so request_handler_ is always set here.
+        // asio leaves accepted sockets non-blocking (for async I/O); the serving thread does
+        // synchronous, blocking transport reads/writes on it, so clear that flag first.
+        std::error_code block_ec;
+        sock->native_non_blocking(false, block_ec);
+
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed: if native_non_blocking(false) fails in do_accept, the connection is dropped and the accept loop re-armed, instead of spawning a serving thread that can't do blocking I/O.

device/simulation/simulation_server_socket.cpp
```diff
@@ -42,6 +45,15 @@ struct SimulationServerSocket::Impl {
...
     std::thread io_thread;
+
+    // Serving state (only used when a request handler is set). Each accepted connection keeps its
+    // socket alive here (RAII owns the fd) alongside the thread serving it, so teardown can reach
+    // every live connection to unblock and join it. Guarded because connections are registered
+    // from the io_thread (accept) and torn down from the destructor thread.
+    std::mutex connections_mutex;
+    std::vector<std::shared_ptr<stream_protocol::socket>> connection_sockets;
+    std::vector<std::thread> connection_threads;
+    bool stopping = false;
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed: finished connections are now reaped. Each serving thread sets a `finished` flag on exit; on every accept the finished entries are joined and erased, so the bookkeeping no longer grows across connect/disconnect cycles.

device/CMakeLists.txt
```diff
@@ -148,6 +148,9 @@ target_sources(
         # (TT_UMD_BUILD_SIMULATION=OFF).
         simulation/simulation_server_socket.cpp
         simulation/simulation_client.cpp
+        # Length-prefixed message framing over the socket (uses asio, a core dependency; no
```
nbuncicTT, line 151: We don't usually add comments for the `.cpp` files in `CMakeLists.txt`, we rather have them in the respective files. Let's omit this comment.

device/simulation/simulation_server_socket.cpp
```diff
@@ -42,6 +46,21 @@ struct SimulationServerSocket::Impl {
     asio::io_context io;
     stream_protocol::acceptor acceptor{io};
     std::thread io_thread;
+
+    // Serving state (only used when a request handler is set). Each accepted connection keeps its
+    // socket alive (RAII owns the fd), the thread serving it, and a flag the thread sets when it
```
nbuncicTT: nit: RAII owns the fd - this is obvious by the fact that the shared_ptr is used to govern the socket, omit it (also goes for the thread and atomic bool).

device/simulation/simulation_server_transport.cpp
```diff
@@ -0,0 +1,100 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent Inc.
```
nbuncicTT, line 1: Please do a proper rebase, as this is already on main

nbuncicTT (approved):

device/CMakeLists.txt
pjanevskiTT: Resolved by the rebase onto main: #2969 (transport) is merged, so its CMake entry and comment are now inherited from main and no longer part of this PR's diff.

device/simulation/simulation_server_socket.cpp
pjanevskiTT: Fixed: trimmed the comment to the essential (reap rationale + guarding), dropping the obvious RAII/per-member details.

device/simulation/simulation_server_transport.cpp
pjanevskiTT: Done: rebased onto main. #2969 (transport) is merged there, so simulation_server_transport.cpp is no longer part of this PR's diff.

PR #2985, opened 2026-07-06
Simulation: model outbound-iATU host-DMA routing for TTSim

pjanevskiTT:
### Issue
Part of #2679, tracked by #2832. Split out of #2825; builds on #2954.

### Description
Model the TTSim host-DMA path the way silicon does: each simulation chip's sysmem sits at a distinct host base, UMD programs the chip's outbound iATU (via BAR2) to map its NOC sysmem window onto that base, and a sysmem DMA is routed to the owning chip purely by the host address it carries. Programmed uniformly for single- and multi-chip — a single-chip sim has host base 0, so the mapping is an identity and routing degenerates to a no-op. This lands the DMA/iATU model ahead of the multi-MMIO (P300) driver that will rely on it.

### List of the changes
- `SimulationSysmemManager`: per-chip host base (`chip_id * kPerChipHostStride`), used as the hugepage physical address (= outbound-iATU region target).
- `TTSimTTDevice`: program the outbound iATU via BAR2 (one region per host-mem channel) during backend bring-up; pass the chip's host window to the DMA callbacks.
- `TTSimCommunicator`: route sysmem DMA to the owning chip by host-address window (`dma_route`), and unregister the window at teardown.

### Testing
CI

### API Changes
Additive and default-compatible: `SimulationSysmemManager` gains an optional `chip_id` ctor arg; `TTSimCommunicator::set_pcie_dma_mem_callbacks` gains optional `host_base`/`host_size`. `TTSimTTDevice` overrides the existing `configure_iatu_region` base virtual.

### DIFF
--- a/device/api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
+++ b/device/api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
@@ -21,8 +21,20 @@ namespace tt::umd {
 
 class SimulationSysmemManager : public SysmemManager {
 public:
-    SimulationSysmemManager(uint32_t num_host_mem_channels, tt::ARCH arch);
+    SimulationSysmemManager(uint32_t num_host_mem_channels, tt::ARCH arch, uint32_t chip_id = 0);
     ~SimulationSysmemManager() override;
+
+    // Per-chip host base for this chip's sysmem. On silicon each chip's pinned host memory lives at a
+    // distinct host physical address; UMD programs that address into the chip's outbound iATU as the
+    // region target, and the chip's DMA resolves there purely by address (no per-chip "magic"). We
+    // model that by giving chip N a distinct host base (N * PER_CHIP_HOST_STRIDE) and using it as the
+    // hugepage physical_address (= iATU target). The simulator's host-side DMA router (dma_route) then
+    // routes by which chip's [host_base, host_base + PER_CHIP_HOST_STRIDE) window contains the address.
+    static constexpr uint64_t PER_CHIP_HOST_STRIDE = 1ULL << 36;  // 64 GiB; >> per-chip sysmem, non-overlapping
+
+    uint64_t get_host_base() const { return host_base_; }
+
+    uint64_t get_host_region_size() const { return PER_CHIP_HOST_STRIDE; }
 
     bool pin_or_map_sysmem_to_device() override;
 
@@ -76,6 +88,7 @@ private:
 
     uint8_t* system_memory_ = nullptr;
     size_t system_memory_size_ = 0;
+    uint64_t host_base_ = 0;  // this chip's distinct host-physical base (= chip_id * PER_CHIP_HOST_STRIDE)
     std::vector<std::pair<void*, size_t>> owned_allocations_;
     std::shared_ptr<MappedBufferRegistry> registry_;
 };
--- a/device/api/umd/device/simulation/tt_sim_communicator.hpp
+++ b/device/api/umd/device/simulation/tt_sim_communicator.hpp
@@ -6,6 +6,7 @@
 
 #include <sys/types.h>
 
+#include <array>
 #include <cstdint>
 #include <filesystem>
 #include <functional>
@@ -126,7 +127,9 @@ public:
      */
     void set_pcie_dma_mem_callbacks(
         std::function<void(uint64_t, void *, uint32_t)> pfn_pci_dma_mem_rd_bytes,
-        std::function<void(uint64_t, const void *, uint32_t)> pfn_pci_dma_mem_wr_bytes);
+        std::function<void(uint64_t, const void *, uint32_t)> pfn_pci_dma_mem_wr_bytes,
+        uint64_t host_base = 0,
+        uint64_t host_size = 0);
 
     void start_sim();
 
@@ -238,14 +241,31 @@ private:
     std::function<void(uint64_t, void *, uint32_t)> pci_dma_mem_rd_bytes_callback_;
     std::function<void(uint64_t, const void *, uint32_t)> pci_dma_mem_wr_bytes_callback_;
 
-    // Known limitation: callback_instance_ is process-global; in multichip mode,
-    // only the last chip to call set_pcie_dma_mem_callbacks receives correct DMA
-    // callbacks.  Fix requires libttsim ABI change to support per-chip context pointer.
+    // Host-side DMA routing by address range. Each MMIO chip's outbound iATU (programmed by UMD via
+    // BAR2, honored by the sim) maps the chip's NOC sysmem window onto that chip's distinct host base.
+    // So a sysmem DMA arrives here carrying a real host address; we find the owning chip by which
+    // registered window [host_base, host_base+size) contains it -- exactly as a host/OS routes a DMA by
+    // which pinned region the address falls in. No chip id crosses the bus; the address alone targets.
+    // callback_instance_ stays as the single-device / unregistered fallback.
+    struct DmaHostRange {
+        uint64_t host_base = 0;
+        uint64_t host_size = 0;
+        TTSimCommunicator *inst = nullptr;
+    };
+
+    static constexpr std::size_t MAX_DMA_DEVICES = 32;
     static TTSimCommunicator *callback_instance_;
+    static std::array<DmaHostRange, MAX_DMA_DEVICES> dma_ranges_;
+    static std::size_t dma_range_count_;
+    static std::mutex dma_ranges_mutex_;
 
     // Static wrapper functions for C-style callbacks.
-    static void pci_dma_mem_rd_bytes_wrapper(uint64_t paddr, void *p, uint32_t size);
-    static void pci_dma_mem_wr_bytes_wrapper(uint64_t paddr, const void *p, uint32_t size);
+    static void pci_dma_mem_rd_bytes_wrapper(uint64_t physical_address, void *p, uint32_t size);
+    static void pci_dma_mem_wr_bytes_wrapper(uint64_t physical_address, const void *p, uint32_t size);
+    // Find the chip whose registered host window contains physical_address, rebase physical_address to
+    // the within-window offset, and return its communicator. Falls back to callback_instance_ if no
+    // window matches.
+    static TTSimCommunicator *dma_route(uint64_t &physical_address);
 
     // Thread safety. In multichip shared-dlopen mode, libttsim_select_device_by_id()
     // and the following libttsim I/O call must be serialized across all
--- a/device/chip_helpers/simulation_sysmem_manager.cpp
+++ b/device/chip_helpers/simulation_sysmem_manager.cpp
@@ -32,8 +32,12 @@ uint64_t align_up(uint64_t value, uint64_t alignment) { return (value + alignmen
 
 }  // namespace
 
-SimulationSysmemManager::SimulationSysmemManager(uint32_t num_host_mem_channels, tt::ARCH arch) {
+SimulationSysmemManager::SimulationSysmemManager(uint32_t num_host_mem_channels, tt::ARCH arch, uint32_t chip_id) {
+    // pcie_base_ is the chip-side NOC sysmem-window base (fixed per arch; used for buffers' NOC addresses).
+    // host_base_ is this chip's distinct host-physical base -- what UMD programs as the outbound-iATU
+    // region target, so each chip's DMA lands in its own host window with no per-chip tag at egress.
     pcie_base_ = get_pcie_base_for_arch(arch);
+    host_base_ = static_cast<uint64_t>(chip_id) * PER_CHIP_HOST_STRIDE;
     registry_ = std::make_shared<MappedBufferRegistry>();
     SimulationSysmemManager::init_sysmem(num_host_mem_channels);
 }
@@ -72,8 +76,11 @@ bool SimulationSysmemManager::init_sysmem(uint32_t num_host_mem_channels) {
 
     for (int i = 0; i < num_host_mem_channels; i++) {
         size_t channel_size = (i == 3 && num_host_mem_channels == 4) ? (768 * (1ULL << 20)) : (1ULL << 30);
+        // physical_address is this chip's host base for the channel -- what UMD programs as the outbound
+        // iATU region target (host_base_ is per-chip distinct). The chip-side NOC sysmem-window address is
+        // separate (get_pcie_base_addr_from_device), so the iATU maps NOC-window offset -> this host base.
         hugepage_mapping_per_channel.push_back(
-            {system_memory_ + i * (1ULL << 30), channel_size, pcie_base_ + i * (1ULL << 30)});
+            {system_memory_ + i * (1ULL << 30), channel_size, host_base_ + i * (1ULL << 30)});
     }
 
     return true;
--- a/device/simulation/tt_sim_communicator.cpp
+++ b/device/simulation/tt_sim_communicator.cpp
@@ -46,6 +46,22 @@ TTSimCommunicator::TTSimCommunicator(
     simulator_directory_(simulator_directory), copy_sim_binary_(copy_sim_binary), chip_id_(chip_id) {}
 
 TTSimCommunicator::~TTSimCommunicator() {
+    // Unregister from the process-global DMA routing tables first. The shared simulator may still be
+    // alive (other chips not yet destructed) and could emit a DMA; leaving a stale entry/callback
+    // pointing at this freed communicator would route into freed memory (use-after-free).
+    {
+        std::lock_guard<std::mutex> ranges_lock(dma_ranges_mutex_);
+        for (std::size_t i = 0; i < dma_range_count_; i++) {
+            if (dma_ranges_[i].inst == this) {
+                dma_ranges_[i] = dma_ranges_[dma_range_count_ - 1];  // compact: move last into the hole
+                dma_ranges_[--dma_range_count_] = {};
+                break;
+            }
+        }
+        if (callback_instance_ == this) {
+            callback_instance_ = nullptr;
+        }
+    }
     if (multichip_mode_) {
         // Shared dlopen -- decrement refcount. When last communicator
         // destructs, call libttsim_exit (which the deferred shutdown()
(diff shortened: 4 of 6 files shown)


device/api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
```diff
@@ -21,9 +21,21 @@ namespace tt::umd {
...
+    SimulationSysmemManager(uint32_t num_host_mem_channels, tt::ARCH arch, uint32_t chip_id = 0);
     ~SimulationSysmemManager() override;
 
+    // Per-chip host base for this chip's sysmem. On silicon each chip's pinned host memory lives at a
+    // distinct host physical address; UMD programs that address into the chip's outbound iATU as the
+    // region target, and the chip's DMA resolves there purely by address (no per-chip "magic"). We
+    // model that by giving chip N a distinct host base (N * kPerChipHostStride) and using it as the
+    // hugepage physical_address (= iATU target). The simulator's host-side DMA router (dma_route) then
+    // routes by which chip's [host_base, host_base + kPerChipHostStride) window contains the address.
+    static constexpr uint64_t kPerChipHostStride = 1ULL << 36;  // 64 GiB; >> per-chip sysmem, non-overlapping
```
nbuncicTT: nit: make this constant in upper snake case.

device/simulation/tt_sim_communicator.cpp
```diff
@@ -273,26 +289,74 @@ void TTSimCommunicator::advance_clock(uint32_t n_clocks) {
...
+        if (host_size != 0) {
+            bool updated = false;
+            for (std::size_t i = 0; i < dma_range_count_; i++) {
+                if (dma_ranges_[i].inst == this) {  // re-registration: update in place
+                    dma_ranges_[i] = {host_base, host_size, this};
+                    updated = true;
+                    break;
+                }
+            }
+            if (!updated && dma_range_count_ < kMaxDmaDevices) {
```
pjanevskiTT: When `dma_range_count_` reaches `kMaxDmaDevices`, a new chip's window is silently dropped (the `if` fails, nothing registers). `dma_route` then falls back to `callback_instance_` and delivers the un-rebased host address to the wrong chip's sysmem — silent memory corruption with no diagnostic. Add an `else` that `UMD_ASSERT`/throws on overflow so exceeding the table is a hard error.

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -160,6 +164,26 @@ void TTSimTTDevice::initialize_backend() {
...
+    //
+    // Programmed uniformly for single- and multi-chip: a single-chip sim simply has host_base 0, so
+    // the mapping is an identity and routing degenerates to a no-op -- the init path is identical
+    // regardless of num_chips. Requires a libttsim that models the BAR2 outbound iATU (WH, and BH as
+    // of the multichip work), which is the behaviour of the stable simulator release.
+    if (arch == tt::ARCH::WORMHOLE_B0 || arch == tt::ARCH::BLACKHOLE) {
+        size_t nch = sysmem_manager_->get_num_host_mem_channels();
+        for (size_t ch = 0; ch < nch; ch++) {
+            HugepageMapping m = sysmem_manager_->get_hugepage_mapping(ch);
+            configure_iatu_region(ch, m.physical_address, m.mapping_size);
```
pjanevskiTT, line 184: These iATU regions only cover the channel grid `[0, num_channels·1GiB)`, but mapped sysmem buffers get arena IOVAs at offsets `>= total_size` (≥4 GiB), outside every region. If the sim passes out-of-region outbound addresses through untranslated, a mapped-buffer DMA arrives without this chip's `host_base` added, so `dma_route` matches chip 0's window and rebases against the wrong registry — misrouted to chip 0. Can you confirm the sim's behavior for out-of-region addresses in multichip? If it's pass-through, extend iATU coverage to the arena or route the arena explicitly.

device/simulation/tt_sim_communicator.cpp
```diff
@@ -273,26 +289,74 @@ void TTSimCommunicator::advance_clock(uint32_t n_clocks) {
...
+TTSimCommunicator *TTSimCommunicator::dma_route(uint64_t &paddr) {
+    std::lock_guard<std::mutex> lock(dma_ranges_mutex_);
+    for (std::size_t i = 0; i < dma_range_count_; i++) {
+        const DmaHostRange &r = dma_ranges_[i];
+        if (paddr >= r.host_base && paddr - r.host_base < r.host_size) {
+            paddr -= r.host_base;
+            return r.inst;
+        }
+    }
+    return callback_instance_;
```
pjanevskiTT, line 328: The no-window-match fallback returns `callback_instance_` but leaves `paddr` un-rebased. Harmless single-chip (`host_base==0`), but in multichip an unmatched address is delivered absolute to the last-registered chip and decomposes into an out-of-range channel index. Consider logging/asserting on an unmatched address when more than one window is registered.

device/simulation/tt_sim_communicator.cpp
```diff
@@ -273,26 +289,74 @@ void TTSimCommunicator::advance_clock(uint32_t n_clocks) {
...
+}
 
 void TTSimCommunicator::pci_dma_mem_rd_bytes_wrapper(uint64_t paddr, void *p, uint32_t size) {
-    if (callback_instance_ && callback_instance_->pci_dma_mem_rd_bytes_callback_) {
-        callback_instance_->pci_dma_mem_rd_bytes_callback_(paddr, p, size);
+    TTSimCommunicator *inst = dma_route(paddr);
+    if (inst && inst->pci_dma_mem_rd_bytes_callback_) {
+        inst->pci_dma_mem_rd_bytes_callback_(paddr, p, size);
```
pjanevskiTT: `dma_route` releases `dma_ranges_mutex_` and returns `inst`, then the callback is dereferenced outside the lock. If a DMA callback and `~TTSimCommunicator` (which unregisters and frees) can run on different threads, this is a use-after-free the teardown-unregister doesn't close. If DMA only runs on the chip's owning thread, worth a comment asserting that invariant; otherwise keep the instance alive across the call.

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -343,10 +367,51 @@ ChipInfo TTSimTTDevice::get_chip_info() {
     return chip_info;
 }
 
+void TTSimTTDevice::configure_iatu_region(size_t region, uint64_t target, size_t region_size) {
+    // Configure the outbound iATU the silicon way: iATU register writes via BAR2 (BH writes these
+    // directly on real HW at ATU_OFFSET_IN_BH_BAR2=0x1000; WH models its iATU regs at 0x1200). We issue
+    // the same register sequence through the simulator's BAR2 MMIO path; the sim decodes it into the
+    // modeled outbound iATU and honors it at DMA egress. `target` is this chip's host base for the
+    // region (per-chip distinct); `base` is the region's offset within the chip's NOC sysmem window.
+    const uint64_t iatu_bar2_off = (arch == tt::ARCH::BLACKHOLE) ? 0x1000 : 0x1200;
```
pjanevskiTT, line 380: The BH iATU offset `0x1000` (and the `0x200` region stride / register offsets) duplicate `BlackholeTTDevice::configure_iatu_region` and its `ATU_OFFSET_IN_BH_BAR2` constant. Referencing the existing named constant keeps the sim in lockstep if the silicon layout changes.

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -343,10 +367,51 @@ ChipInfo TTSimTTDevice::get_chip_info() {
...
+    const uint64_t iatu = bar2_base + iatu_bar2_off + uint64_t(region) * 0x200;
+    auto wr = [&](uint64_t off, uint32_t val) { communicator_->pci_mem_write_bytes(iatu + off, &val, sizeof(val)); };
+    wr(0x04, 0);                       // region_ctrl_2: disable while (re)programming the region
+    wr(0x00, 0);                       // region_ctrl_1
+    wr(0x08, uint32_t(base));          // base lo
+    wr(0x0c, uint32_t(base >> 32));    // base hi
+    wr(0x10, uint32_t(limit));         // limit (low 32b; upper bits share base hi)
+    wr(0x14, uint32_t(target));        // target lo
+    wr(0x18, uint32_t(target >> 32));  // target hi
+    wr(0x04, 1u << 31);                // region_ctrl_2 = REGION_EN, written last so the sim validates a complete region
```
pjanevskiTT: Unlike the real BH sequence, this omits explicit writes of `limit_hi` (0x1c) and `region_ctrl_3` (0x20), relying on the sim zero-initializing them. Since a region is disabled/re-enabled on reprogram, a stale `limit_hi` from a prior config could apply. Writing them to 0 explicitly (as `BlackholeTTDevice` does) removes that dependence.

device/api/umd/device/simulation/tt_sim_communicator.hpp
```diff
@@ -238,14 +241,30 @@ class TTSimCommunicator final {
...
+    // registered window [host_base, host_base+size) contains it -- exactly as a host/OS routes a DMA by
+    // which pinned region the address falls in. No chip id crosses the bus; the address alone targets.
+    // callback_instance_ stays as the single-device / unregistered fallback.
+    struct DmaHostRange {
+        uint64_t host_base = 0;
+        uint64_t host_size = 0;
+        TTSimCommunicator *inst = nullptr;
+    };
+
+    static constexpr std::size_t kMaxDmaDevices = 32;
```
nbuncicTT: nit: Make this upper snake case, do so for all constants here and in future.

device/chip_helpers/simulation_sysmem_manager.cpp
```diff
@@ -32,8 +32,12 @@ uint64_t align_up(uint64_t value, uint64_t alignment) { return (value + alignmen
 
 }  // namespace
 
-SimulationSysmemManager::SimulationSysmemManager(uint32_t num_host_mem_channels, tt::ARCH arch) {
+SimulationSysmemManager::SimulationSysmemManager(uint32_t num_host_mem_channels, tt::ARCH arch, uint32_t chip_id) {
+    // pcie_base_ is the chip-side NOC sysmem-window base (fixed per arch; used for buffers' NOC addresses).
+    // host_base_ is this chip's distinct host-physical base -- what UMD programs as the outbound-iATU
+    // region target, so each chip's DMA lands in its own host window with no per-chip tag at egress.
     pcie_base_ = get_pcie_base_for_arch(arch);
+    host_base_ = uint64_t(chip_id) * kPerChipHostStride;
```
nbuncicTT: nit: use static_cast here, it's idiomatic C++

device/simulation/tt_sim_communicator.cpp
```diff
@@ -273,26 +289,74 @@ void TTSimCommunicator::advance_clock(uint32_t n_clocks) {
...
+// through its outbound iATU (UMD-programmed target = that chip's distinct host base), so we just find
+// the registered host window [host_base, host_base+size) that contains it -- exactly as a host routes a
+// DMA by which pinned region the physical address falls in. Rebase paddr to the within-window offset (so
+// the per-chip callback sees an offset relative to its own host base, like the single-device case) and
+// return the owning communicator. Falls back to callback_instance_ when no window matches (single
+// device, or an address outside any registered region).
+TTSimCommunicator *TTSimCommunicator::dma_route(uint64_t &paddr) {
+    std::lock_guard<std::mutex> lock(dma_ranges_mutex_);
+    for (std::size_t i = 0; i < dma_range_count_; i++) {
+        const DmaHostRange &r = dma_ranges_[i];
```
nbuncicTT: nit: use a more descriptive name than `r`

device/simulation/tt_sim_communicator.cpp
```diff
@@ -273,26 +289,74 @@ void TTSimCommunicator::advance_clock(uint32_t n_clocks) {
...
+        if (paddr >= r.host_base && paddr - r.host_base < r.host_size) {
```
nbuncicTT: nit: we use `physical_address` in `pci_device.hpp`, so use this instead of `paddr`.

nbuncicTT: don't rely on operator precedence, use parentheses.

device/simulation/tt_sim_communicator.cpp
```diff
@@ -273,26 +289,74 @@ void TTSimCommunicator::advance_clock(uint32_t n_clocks) {
...
+            return r.inst;
+        }
+    }
+    return callback_instance_;
+}
 
 void TTSimCommunicator::pci_dma_mem_rd_bytes_wrapper(uint64_t paddr, void *p, uint32_t size) {
-    if (callback_instance_ && callback_instance_->pci_dma_mem_rd_bytes_callback_) {
-        callback_instance_->pci_dma_mem_rd_bytes_callback_(paddr, p, size);
+    TTSimCommunicator *inst = dma_route(paddr);
```
nbuncicTT: Rename inst to `dma_callback` or `dma_cb`. `inst` is too vague.

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -343,10 +367,51 @@ ChipInfo TTSimTTDevice::get_chip_info() {
...
+    // region (per-chip distinct); `base` is the region's offset within the chip's NOC sysmem window.
+    const uint64_t iatu_bar2_off = (arch == tt::ARCH::BLACKHOLE) ? 0x1000 : 0x1200;
+    uint64_t bar2_base = communicator_->pci_config_read32(0, 0x18);
+    bar2_base |= uint64_t(communicator_->pci_config_read32(0, 0x1C)) << 32;
+    bar2_base &= ~15ull;  // strip BAR type/attribute bits, leaving the physical address
+    // Channels sit on a fixed 1 GiB NOC-window grid (matching SimulationSysmemManager's placement),
+    // independent of region_size. region_size is the channel's actual mapping size and only bounds the
+    // limit: keeping base on the grid means a sub-1-GiB channel (WH channel 3 = 768 MiB) still starts at
+    // the right NOC offset (region * 1 GiB) while mapping only its backed range.
+    constexpr uint64_t kChannelStride = 1ULL << 30;
```
nbuncicTT: nit: upper snake case.

nbuncicTT (approved):

device/simulation/tt_sim_communicator.cpp
```diff
@@ -273,26 +289,74 @@ void TTSimCommunicator::advance_clock(uint32_t n_clocks) {
...
+                if (dma_ranges_[i].inst == this) {  // re-registration: update in place
+                    dma_ranges_[i] = {host_base, host_size, this};
+                    updated = true;
+                    break;
+                }
+            }
+            if (!updated && dma_range_count_ < kMaxDmaDevices) {
+                dma_ranges_[dma_range_count_++] = {host_base, host_size, this};
+            }
+        }
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed in `3f995bca`. Overflowing the range table is now a hard `UMD_ASSERT` instead of silently dropping the window, and calling `set_pcie_dma_mem_callbacks(..., host_size==0)` now removes any window previously registered for the instance, so a stale range can no longer persist.

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -430,10 +451,52 @@ void TTSimTTDevice::dma_multicast_write(
...
+    wr(0x14, uint32_t(target));        // target lo
+    wr(0x18, uint32_t(target >> 32));  // target hi
+    wr(0x04, 1u << 31);                // region_ctrl_2 = REGION_EN, written last so the sim validates a complete region
+}
+
 void TTSimTTDevice::initialize_sysmem_functions() {
+    // Register this chip's distinct host window [host_base, host_base+size) so the simulator's host-side
+    // DMA router (dma_route) routes each chip's DMA by address range -- the chip emits a host address
+    // (via its outbound iATU), and the host finds the owning chip by which window contains it.
+    auto* sim_mgr = static_cast<SimulationSysmemManager*>(sysmem_manager_.get());
```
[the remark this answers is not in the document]
pjanevskiTT: Keeping `static_cast` here: a `TTSimTTDevice` always constructs its `sysmem_manager_` as a `SimulationSysmemManager` (see the base-ctor call), so the type is guaranteed by construction. `static_cast` is also the idiom preferred elsewhere in this review.

device/api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
pjanevskiTT: Fixed in `3f995bca` — renamed to `PER_CHIP_HOST_STRIDE`.

device/simulation/tt_sim_communicator.cpp
pjanevskiTT: Fixed in `3f995bca` — overflowing `MAX_DMA_DEVICES` now throws via `UMD_ASSERT` instead of silently falling back to `callback_instance_`.

device/tt_device/tt_sim_tt_device.cpp
pjanevskiTT: Addressed in `3f995bca`. Assuming the simulator passes out-of-region outbound addresses through, `dma_route` now hard-fails (`UMD_ASSERT`) when an address matches no registered window and more than one window is registered, rather than misrouting to chip 0. Extending iATU coverage to the mapped-buffer arena (or routing it explicitly) is the real fix and is deferred.

device/simulation/tt_sim_communicator.cpp
pjanevskiTT: Fixed in `3f995bca` — the no-match path now throws in multichip (more than one window registered) instead of returning `callback_instance_` with an un-rebased address. Single-device keeps the identity fallback.

device/simulation/tt_sim_communicator.cpp
pjanevskiTT: Addressed in `3f995bca` — documented the invariant: DMA callbacks fire synchronously only from the owning chip's own thread while it drives the simulator under `device_lock_`, so a chip cannot be issuing a DMA and running its own `~TTSimCommunicator` (which unregisters under the same mutex) at the same time.

device/tt_device/tt_sim_tt_device.cpp
pjanevskiTT: Leaving the magic numbers with the explanatory comment for now: `ATU_OFFSET_IN_BH_BAR2` is private to `BlackholeTTDevice` and there is no named WH equivalent or stride constant, so true lockstep would mean exposing new shared constants across the device API — not worth it for this PR.

device/tt_device/tt_sim_tt_device.cpp
pjanevskiTT: Fixed in `3f995bca` — now writes `limit_hi` (0x1c) and `region_ctrl_3` (0x20) explicitly to 0, matching `BlackholeTTDevice::configure_iatu_region`.

device/api/umd/device/simulation/tt_sim_communicator.hpp
pjanevskiTT: Fixed in `3f995bca` — renamed to `MAX_DMA_DEVICES`.

device/chip_helpers/simulation_sysmem_manager.cpp
pjanevskiTT: Fixed in `3f995bca` — `static_cast<uint64_t>(chip_id)`.

device/simulation/tt_sim_communicator.cpp
pjanevskiTT: Fixed in `3f995bca` — renamed to `range`.

device/simulation/tt_sim_communicator.cpp
pjanevskiTT: Fixed in `3f995bca` — renamed `paddr` to `physical_address` and added parentheses around the offset comparison.

device/simulation/tt_sim_communicator.cpp
pjanevskiTT: Fixed in `3f995bca` — renamed to `dma_cb`.

device/tt_device/tt_sim_tt_device.cpp
pjanevskiTT: Fixed in `3f995bca` — renamed to `CHANNEL_STRIDE`.

device/tt_device/tt_sim_tt_device.cpp
pjanevskiTT: Follow-up: the explicit `limit_hi` (0x1c) / `region_ctrl_3` (0x20) writes were reverted in `9340b0e7`. The deployed ttsim iATU model does not implement those register offsets and throws `UnimplementedFunctionality: iatu_reg_field: bar2: iATU reg offset=0x1c` (CI: run-ttsim-tests WH step). Omitting them is safe here -- the region top is asserted `< 4 GiB` (limit_hi is always 0) and regions are programmed once at init, not reprogrammed, so there is no stale-high-bits hazard.

PR #2986, opened 2026-07-06
Simulation: SimulationClient::transact (framed request/response)

pjanevskiTT:
### Issue
Part of #2677 (simulation server process). Grows the client transport (#2902) from attach/detach to a full request/response turn, so it can carry device operations. Stacked on #2984.

### Description
Adds `SimulationClient::transact(request) -> reply`: sends one framed request over the connected socket and blocks for the framed reply, via the shared transport (#2969). `attach()` clears any non-blocking flag asio leaves on the socket so the transport's synchronous reads/writes block as intended.

Kept **protocol-agnostic** (opaque payloads in/out), like `SimulationServerSocket`: device operations are encoded/decoded *above* this handle, in the simulation-gated device that owns it, so `SimulationClient` stays always-built and FlatBuffers-free. This is also what lets the client-mode device dispatch be rebound onto `transact()` in a follow-up.

Pairs with the serving loop (#2984): `client.transact` ↔ server `recv_framed → handler → send_framed`, both over framed opaque payloads.

### List of the changes
- `device/api/umd/device/simulation/simulation_client.hpp`, `device/simulation/simulation_client.cpp` — `transact()` (framed request/response over the socket via the shared transport); blocking-mode set on `attach()`.
- `tests/baremetal/test_simulation_client.cpp` — 4 end-to-end tests (real client ↔ serving socket): fixed-reply round-trip, sequential requests, transact-before-attach throws, transact-after-host-gone throws.

### Testing
`SimulationClientTest` (7, 4 new) pass in the no-card baremetal lane; build + pre-commit clean.

### API Changes
`SimulationClient` gains a `transact()` method (existing public header).

### DIFF
--- a/device/api/umd/device/simulation/simulation_client.hpp
+++ b/device/api/umd/device/simulation/simulation_client.hpp
@@ -4,20 +4,18 @@
 
 #pragma once
 
+#include <cstdint>
 #include <filesystem>
 #include <memory>
+#include <vector>
 
 namespace tt::umd {
 
-// The client-facing handle to a simulation host ("the card"), reached over its per-chip UNIX
-// socket. Deliberately slim for now: just the session handshake -- attach() connects, detach()
-// closes. It grows operation-by-operation as the server work lands (device description, memory
-// access, TLB, sysmem, run/reset, ...), at which point TTSimTTDevice's client-mode dispatch is
-// rebound from its throwing stubs onto these calls. No simulation build needed (asio is a core
-// dependency), mirroring SimulationSocket on the host side.
+// Client-facing handle to a simulation host ("the card") over its per-chip UNIX socket. transact()
+// moves opaque request/reply payloads; encoding lives in the simulation device that owns this
+// handle, so this always-built transport carries no protocol/FlatBuffers dependency.
 //
-// The asio transport is held behind a pImpl so this header stays free of asio (a private
-// dependency of the library) -- see Impl in the .cpp.
+// asio is held behind a pImpl so this header stays asio-free (see Impl in the .cpp).
 class SimulationClient {
 public:
     explicit SimulationClient(std::filesystem::path socket_path);
@@ -31,6 +29,11 @@ public:
     void attach();
     // Closes the connection; idempotent.
     void detach();
+
+    // Sends one length-prefixed request to the host and blocks for the length-prefixed reply,
+    // returning its opaque payload. Payloads are encoded/decoded by the caller (the protocol
+    // layer). Throws if not attached, or if the host disconnects or errors mid-exchange.
+    std::vector<uint8_t> transact(const std::vector<uint8_t>& request);
 
 private:
     std::filesystem::path socket_path_;
--- a/device/simulation/simulation_client.cpp
+++ b/device/simulation/simulation_client.cpp
@@ -10,6 +10,7 @@
 #include <asio.hpp>
 #include <system_error>
 
+#include "simulation/simulation_server_transport.hpp"
 #include "umd/device/utils/error.hpp"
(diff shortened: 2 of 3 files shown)


device/simulation/simulation_client.cpp
```diff
@@ -61,6 +62,9 @@ void SimulationClient::attach() {
             error::RuntimeError,
             fmt::format("Failed to connect to simulation host socket at {}: {}", socket_path_.string(), ec.message()));
     }
+    // transact() does blocking reads/writes on the raw fd via the transport helpers, so make sure
+    // asio has not left the socket non-blocking.
+    impl_->socket.native_non_blocking(false, ec);
 }
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed: attach() now checks the native_non_blocking(false) error_code -- on failure it closes the socket and throws, so a half-attached client is never reported as connected.

device/api/umd/device/simulation/simulation_client.hpp
```diff
@@ -4,17 +4,23 @@
...
 // The client-facing handle to a simulation host ("the card"), reached over its per-chip UNIX
-// socket. Deliberately slim for now: just the session handshake -- attach() connects, detach()
-// closes. It grows operation-by-operation as the server work lands (device description, memory
-// access, TLB, sysmem, run/reset, ...), at which point TTSimTTDevice's client-mode dispatch is
-// rebound from its throwing stubs onto these calls. No simulation build needed (asio is a core
-// dependency), mirroring SimulationSocket on the host side.
+// socket. attach() connects, detach() closes, and transact() runs one request/response turn.
+//
+// It stays deliberately protocol-agnostic: transact() moves opaque request/reply payloads,
+// mirroring the server socket. The device operations (read/write memory, TLB, ...) are encoded
```
nbuncicTT: nit: Try to keep comments shorter, because this way, they are unmaintainable, and just give too much context, we need concise comments which would explain what the class does. And per API/method comments should be written above the functions themselves. Also, phrases like "deliberately protocol-agnostic" are too poetic, keep it simple and drop the deliberately.

Don't know if this make sense to add to your CLAUDE.md as a guide for comments, as this is a recurring pattern I'm seeing.

device/simulation/simulation_client.cpp
```diff
@@ -61,6 +62,20 @@ void SimulationClient::attach() {
             error::RuntimeError,
             fmt::format("Failed to connect to simulation host socket at {}: {}", socket_path_.string(), ec.message()));
     }
+    // transact() does blocking reads/writes on the raw fd via the transport helpers, so make sure
```
nbuncicTT: nit: You purposefully use PIMPL to hide the implementation details, and then say what the implementation details of transact() are, like using raw fd's. Let's omit these kinds of comments.

nbuncicTT: I've now double checked, transact() uses ASIO and not raw fd's so aside from the comment being misplaced, it's not correct. Please remove.

nbuncicTT (approved):

device/api/umd/device/simulation/simulation_client.hpp
pjanevskiTT: Fixed: shortened the class comment, dropped 'deliberately' and the backstory; per-method comments stay above the functions. (Also noted the concise-comment guidance for future work.)

device/simulation/simulation_client.cpp
pjanevskiTT: Fixed: dropped the transact() implementation-detail comment (raw fds) that the pImpl hides.

PR #2988, opened 2026-07-06
Simulation: BDF multi-MMIO TTSim driver + P300 (bh_x2) CI

pjanevskiTT:
### Issue
Part of #2679, tracked by #2832. Split out of #2825; builds on #2985.

### Description
Drive a multi-MMIO multi-chip TTSim (e.g. the Blackhole P300 — two MMIO chips). With no multichip ABI but a >1-chip cluster, all chips share one dlopen and are addressed by PCI device (BDF) + per-device BAR window — the host-enumeration model — so no new ttsim ABI is needed. Combined with the iATU host-DMA routing (#2985), each chip's DMA lands in its own host window by address alone.

### List of the changes
- `TTSimCommunicator`: shared-dlopen BDF mode for a >1-chip cluster addressed by PCI device; BDF device-field fill-in in `pci_config_read32`; shared init/refcount/teardown.
- Thread `num_chips` through `TTSimTTDevice` and `create_simulation_tt_device`.
- CI: add the bh_x2 (P300) lane; `SysmemReadWrite` exercises every MMIO chip with a chip-distinct pattern to catch DMA misrouting.

### Testing
CI

### API Changes
Additive and default-compatible: `TTSimCommunicator` and `TTSimTTDevice` gain a `num_chips` parameter (default 1).

### DIFF
--- a/device/api/umd/device/simulation/tt_sim_communicator.hpp
+++ b/device/api/umd/device/simulation/tt_sim_communicator.hpp
@@ -31,7 +31,10 @@ public:
      *   for legacy single-chip consumers.
      */
     TTSimCommunicator(
-        const std::filesystem::path &simulator_directory, bool copy_sim_binary = false, uint32_t chip_id = 0);
+        const std::filesystem::path &simulator_directory,
+        bool copy_sim_binary = false,
+        uint32_t chip_id = 0,
+        uint32_t num_chips = 1);
 
     /**
      * Destructor that properly cleans up library handles and file descriptors.
@@ -157,6 +160,12 @@ private:
     void close_simulator_binary();
     void load_simulator_library(const std::filesystem::path &path);
 
+    // Error-path cleanup for the shared-dlopen init paths. Called with s_shared_init_mutex_ held when
+    // symbol resolution throws before this communicator commits (bumps s_shared_refcount_). If this call
+    // performed the fresh dlopen (refcount still 0), releases s_shared_handle_ so it does not leak and a
+    // retry starts clean; if a prior communicator already owns it (refcount > 0), leaves it alone.
+    void release_uncommitted_shared_handle();
+
     // In multichip mode, selects this communicator's chip before an I/O call.
     void select_chip_if_needed();
 
@@ -193,7 +202,15 @@ private:
     // True when the loaded .so supports the multichip ABI and this
     // communicator is using the shared dlopen path.
     bool multichip_mode_ = false;
+    // True when the .so has NO multichip ABI but the cluster has >1 chip: all communicators share one
+    // dlopen and each chip is addressed by its PCI device (BDF) rather than libttsim_select_device_by_id.
+    bool shared_bdf_mode_ = false;
+
+    // Both modes use the single shared dlopen (s_shared_handle_) + refcount.
+    bool uses_shared_handle() const { return multichip_mode_ || shared_bdf_mode_; }
+
     uint32_t chip_id_ = 0;
+    uint32_t num_chips_ = 1;
     static void *s_shared_handle_;
     static int s_shared_refcount_;
     static bool s_sim_initialized_;
--- a/device/simulation/tt_sim_communicator.cpp
+++ b/device/simulation/tt_sim_communicator.cpp
@@ -14,10 +14,12 @@
 
 #include <cerrno>
 #include <cstring>
+#include <filesystem>
 #include <string>
 #include <tt-logger/tt-logger.hpp>
 #include <utility>
 
+#include "umd/device/simulation/simulation_chip.hpp"
 #include "umd/device/utils/error.hpp"
 
 // NOLINTBEGIN.
@@ -42,8 +44,11 @@ std::mutex TTSimCommunicator::s_shared_init_mutex_;
 std::mutex TTSimCommunicator::device_lock_;
 
 TTSimCommunicator::TTSimCommunicator(
-    const std::filesystem::path &simulator_directory, bool copy_sim_binary, uint32_t chip_id) :
-    simulator_directory_(simulator_directory), copy_sim_binary_(copy_sim_binary), chip_id_(chip_id) {}
+    const std::filesystem::path &simulator_directory, bool copy_sim_binary, uint32_t chip_id, uint32_t num_chips) :
+    simulator_directory_(simulator_directory),
+    copy_sim_binary_(copy_sim_binary),
+    chip_id_(chip_id),
+    num_chips_(num_chips) {}
 
 TTSimCommunicator::~TTSimCommunicator() {
     // Unregister from the process-global DMA routing tables first. The shared simulator may still be
@@ -62,7 +67,7 @@ TTSimCommunicator::~TTSimCommunicator() {
             callback_instance_ = nullptr;
         }
     }
-    if (multichip_mode_) {
+    if (uses_shared_handle()) {
         // Shared dlopen -- decrement refcount. When last communicator
         // destructs, call libttsim_exit (which the deferred shutdown()
         // path didn't call), then dlclose. The simulator is process-global,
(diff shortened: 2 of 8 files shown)


device/simulation/tt_sim_communicator.cpp
```diff
@@ -166,6 +176,36 @@ void TTSimCommunicator::initialize() {
...
 
+    // No multichip ABI, but a multi-chip cluster: share a single dlopen so all chips
+    // live in one libttsim process image (one shared per-chip state array), and address
+    // each chip by its PCI device (BDF) + per-device BAR window. This is the BDF
+    // host-enumeration path (docs/multichip/ARCHITECTURE.md) -- no select_device_by_id,
+    // no virtual eth switch (the simulator routes inter-chip eth internally).
+    if (num_chips_ > 1) {
+        std::lock_guard<std::mutex> init_lock(s_shared_init_mutex_);
+        if (!s_shared_handle_) {
+            s_shared_handle_ = dlopen(simulator_directory_.c_str(), RTLD_LAZY);
```
pjanevskiTT, line 215: If a `DLSYM_FUNCTION` throws after `s_shared_handle_ = dlopen(...)` but before `s_shared_refcount_` is bumped, the destructor's `else if (libttsim_handle_)` branch sees `== s_shared_handle_` and skips `dlclose`, so the handle leaks with refcount 0. Mirrors the existing probe path, but consider clearing `s_shared_handle_` on the throw so a retry starts clean.

device/api/umd/device/simulation/tt_sim_communicator.hpp
```diff
@@ -193,7 +196,18 @@ class TTSimCommunicator final {
     // True when the loaded .so supports the multichip ABI and this
     // communicator is using the shared dlopen path.
     bool multichip_mode_ = false;
+    // True when the .so has NO multichip ABI but the cluster has >1 chip: all
```
nbuncicTT: Too detailed comment, make it concise, up to one or two sentences. If needed, write a .md on simulators how this works instead of leaving unmaintainable comments.

tests/api/test_device_io.cpp
```diff
@@ -568,101 +563,116 @@ TEST_F(TestDeviceIOFixture, SysmemReadWrite) {
...
-            uint64_t aligned_offset = (test_offset / ALIGNMENT) * ALIGNMENT;
-            uint32_t value = 0xffffffff;
-            std::memcpy(&value, &sysmem[aligned_offset], sizeof(uint32_t));
-            EXPECT_EQ(value, 0);
+            // Write test - zero out the sysmem at the various offsets.
+            for (uint64_t test_offset : test_offsets) {
+                uint64_t aligned_offset = (test_offset / ALIGNMENT) * ALIGNMENT;
+                uint64_t device_offset = aligned_offset + channel * ONE_GIG;
+                uint64_t noc_addr = base_address + device_offset;
+                uint32_t value = 0;
```
nbuncicTT, line 668: Can we put a more distinct pattern to check against, like `0xDEAD'BEEF`?

tests/api/test_device_io.cpp
```diff
@@ -568,101 +563,116 @@ TEST_F(TestDeviceIOFixture, SysmemReadWrite) {
         test_utils::safe_test_cluster_start(cluster.get());
     }
 
-    for (uint32_t channel = 0; channel < channels_to_test; channel++) {
-        uint8_t* sysmem = static_cast<uint8_t*>(cluster->host_dma_address(mmio_chip_id, 0, channel));
+    // Exercise every MMIO chip's sysmem, not just chip 0. On a multi-MMIO simulator (e.g. P300 / bh_x2)
+    // this is what actually validates the host-DMA routing model: each chip's DMA must land in *its own*
+    // host window. Each chip fills its host buffer with a chip-distinct pattern, so a misrouted DMA
+    // (landing in another chip's window, or aliasing host_base 0) would surface as a readback mismatch.
+    const std::set<ChipId> mmio_chips = cluster->get_target_mmio_device_ids();
```
nbuncicTT: nit: use `auto` or `auto const` here so the test is independent of the API.

tests/api/test_device_io.cpp
```diff
@@ -568,101 +563,116 @@ TEST_F(TestDeviceIOFixture, SysmemReadWrite) {
...
+    for (const ChipId mmio_chip_id : mmio_chips) {
+        const auto pci_cores = cluster->get_soc_descriptor(mmio_chip_id).get_cores(CoreType::PCIE);
+        const auto pcie_core = pci_cores.at(0);
+        const auto base_address = cluster->get_pcie_base_addr_from_device(mmio_chip_id);
```
nbuncicTT, line 574: nit: rename to `get_sysmem_window_noc_base` and `sysmem_window_noc_base_` (the data it's getter returns).

device/api/umd/device/simulation/tt_sim_communicator.hpp
```diff
@@ -193,7 +196,18 @@ class TTSimCommunicator final {
...
+    // model in docs/multichip/ARCHITECTURE.md. No libttsim_select_device_by_id; the
+    // device is carried in the PCI config BDF and in the per-device BAR physical address.
+    bool shared_bdf_mode_ = false;
+
+    // Both modes use the single shared dlopen (s_shared_handle_) + refcount.
+    bool uses_shared_handle() const { return multichip_mode_ || shared_bdf_mode_; }
+
     uint32_t chip_id_ = 0;
+    uint32_t num_chips_ = 1;
     static void *s_shared_handle_;
```
nbuncicTT, line 214: Follow-up: replace the manual shared-handle refcount with std::shared_ptr<void>

The shared `libttsim` `dlopen` is currently tracked by hand — `s_shared_handle_` (raw void*) + `s_shared_refcount_` (int), bumped in both acquire paths (multichip-ABI and BDF) and decremented in the dtor, which runs a 3-step teardown (`libttsim_exit` → `dlclose` → reset `s_sim_initialized_`) at zero. This PR also adds the `libttsim_handle_ != s_shared_handle_` dtor guard to avoid a double-close when the ABI probe aliases the handle before anyone increments the count.

We can replace all of it with a static `std::shared_ptr<void>` whose custom deleter does the teardown: acquire = take a copy, release = drop it, "last one out" = the control block. That removes the manual refcount, the aliasing guard (double-close becomes impossible once the probe holds a copy too), and the leak-on-init-throw the guard currently trades for.

This is a follow-up, not a trivial swap — the current PR's locking is correct as-is; these are constraints for doing the refactor safely:

`s_shared_init_mutex_` stays — `shared_ptr` makes the refcount atomic, not the check-then-dlopen, and concurrent mutation of the same s_shared_handle_ object is still a race.
The deleter mutates shared static state (`s_sim_initialized_`) and fires whenever the last copy drops — it must take `s_shared_init_mutex_` internally.
Keep `libttsim_handle_` as a raw non-owning view for the dlsym sites; add one owning `shared_ptr` member rather than converting it.
Guard the dlopen result before wrapping — a shared_ptr on nullptr still runs its deleter.
Watch static-destruction ordering — a stray lingering copy shifts libttsim_exit/dlclose to process exit.

Make this a follow-up issue.

nbuncicTT (approved):

device/simulation/tt_sim_communicator.cpp
pjanevskiTT: Fixed in `81729f97`. Both shared-dlopen init paths now wrap symbol resolution in try/catch and, on a throw before the refcount is bumped, release `s_shared_handle_` when this call performed the fresh `dlopen` (refcount still 0) so it doesn't leak and a retry starts clean.

device/api/umd/device/simulation/tt_sim_communicator.hpp
pjanevskiTT: Fixed in `81729f97` — shortened to two lines.

tests/api/test_device_io.cpp
pjanevskiTT: Fixed in `81729f97` — sim now fills the first page with a chip-keyed 32-bit sentinel (`0xDEAD'BEEF + chip_id`), so a misroute is obvious on readback.

tests/api/test_device_io.cpp
pjanevskiTT: Fixed in `81729f97` — `const auto mmio_chips`.

tests/api/test_device_io.cpp
pjanevskiTT: `get_pcie_base_addr_from_device` is a pre-existing public `Cluster` API used across several call sites, not introduced here, so I've kept the rename out of this P300 CI PR and tracked it separately: #3068.

device/api/umd/device/simulation/tt_sim_communicator.hpp
pjanevskiTT: Filed as a follow-up: #3069. Leaving the manual refcount/locking as-is for this PR per your note that it's correct as written.

tests/api/test_device_io.cpp
```diff
@@ -568,103 +563,116 @@ TEST_F(TestDeviceIOFixture, SysmemReadWrite) {
...
-                              << "\n  - After adding the parameter, reboot the HOST (not just the VM)"
-                              << "\n  - Check host dmesg for IO page faults"
-                              << "\n  - Failure at offset >= 255MB strongly indicates dma_entry_limit issue";
+            if (is_simulation_test()) {
+                for (size_t i = 0; i < ONE_GIG; i++) {
+                    sysmem[i] = static_cast<uint8_t>((i + chip_pattern_bias) % 256);
                 }
+            } else {
+                test_utils::fill_with_random_bytes(sysmem, ONE_GIG);
+            }
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed in `81729f97` — sim only reads offset 0x0, so it now fills a single page instead of the byte-at-a-time 1 GiB loop.

tests/api/test_device_io.cpp
```diff
@@ -568,103 +563,116 @@ TEST_F(TestDeviceIOFixture, SysmemReadWrite) {
...
+                    (ONE_GIG / 2) - 0x1000,
+                    (ONE_GIG / 2) - 0x0004,
+                    (ONE_GIG / 2),
+                    (ONE_GIG / 2) + 0x0004,
+                    (ONE_GIG / 2) + 0x1000,
+                    (ONE_GIG - 0x1000),
+                    (ONE_GIG - 0x0004),
+                };
+                for (size_t i = 0; i < 8192; ++i) {
+                    uint64_t address = random_address_between(0, ONE_GIG);
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed in `81729f97` — capped the random upper bound to `ONE_GIG - sizeof(uint32_t)` so the inclusive distribution can no longer produce an out-of-bounds aligned read.

PR #2996, opened 2026-07-08
Fix false-positive NOC hangs from the per-op MMIO timeout

### LINKED ISSUE #2981: False-positive NOC hangs reported by NOC I/O timeout feature
Example: https://github.com/tenstorrent/tt-umd/actions/runs/28734116949/job/85205878500

pjanevskiTT:
### Issue
Closes #2981

### Description
The per-op MMIO timeout (#2629) reported false-positive NOC hangs on healthy devices whenever the budget was set low (e.g. exalens' 2 ms). Two mechanisms were at play. First, a TLB window with no hang detector wired left an empty timeout callback, so any overrun aborted with `DeviceTimeoutError` on time alone — nothing distinguished a slow-but-healthy op from a genuinely hung one. Second, the hang-detector's own NOC-liveness probe reads through exactly such a no-hang-check window; when the budget sat below the probe read's latency the probe itself timed out and was interpreted as a hang, so a healthy device threw `NocHangError` during device init.

The fix defaults a no-hang-check window to treating every overrun as a false alarm, so a bare per-op timeout never aborts on time alone and the probe read completes and is judged by value (`HANG_READ_VALUE`) rather than by elapsed time. With the probe no longer time-sensitive, the default budget no longer needs headroom for it, so it is lowered back to 2 ms, and the probe reader no longer swallows a `DeviceTimeoutError` as a hang.

### List of the changes
- `SiliconTlbWindow`: with no hang detector wired, default the I/O-timeout callback to always-false-alarm (was empty) so an overrun is not aborted on time alone.
- Lower the default per-op MMIO budget from 10 ms to 2 ms.
- Hang-detector probe reader: drop the `catch` that returned `HANG_READ_VALUE` on `DeviceTimeoutError`; let an unexpected probe timeout surface instead of being reported as a hang.
- Update the related doc comments.

### Testing
CI

### API Changes
This PR has no API changes.

### DIFF
--- a/.github/workflows/build-and-run-all-tests.yml
+++ b/.github/workflows/build-and-run-all-tests.yml
@@ -251,8 +251,7 @@ jobs:
         test-group: [
           {runs-on: tt-ubuntu-2204-p150b-stable},
           {runs-on: tt-ubuntu-2204-n150-stable},
-          # TODO(#2981): Re-enable once exalens tests NOC0 hang bug is resolved.
-          # {runs-on: tt-ubuntu-2204-n300-stable},
+          {runs-on: tt-ubuntu-2204-n300-stable},
         ]
         ubuntu-docker-version: [
           'ubuntu-22.04',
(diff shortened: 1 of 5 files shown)


aleksamarkovicTT (approved):
Please also re-enable n300 exalens tests

ajovanovicTT (approved):

PR #3008, opened 2026-07-10
Simulation: serve device memory over the socket (host handler)

pjanevskiTT:
### Issue
Part of #2677 (simulation server process). Connects the serving socket (#2984) to the host device's backend so remote clients can read/write device memory. Stacked on #2986. Builds on the merged SimulationTTDevice I/O hooks (#2917).

### Description
`SimulationTTDevice::handle_request(bytes) → bytes` is the host-side seam:
- `decode_request` → build a `CoreCoord` in LITERAL (the client already translated the coordinate, so the shared skeleton must not translate again) → run through the base `read_from_device`/`write_to_device`, so device memory goes through the ttsim device's **TLB window**, identical to the host's own I/O (the #2917 hooks) → `encode` the reply.
- A failed op returns a nonzero status rather than dropping the connection, mirroring how a silicon access error is reported.

`adopt_socket()` installs this handler via `SimulationServerSocket::serve()` once the backend is up. Safe against teardown: the derived destructors `socket_.reset()` (joining the serving threads) before tearing the backend down, so no handler runs against a half-destroyed device.

Coordinate translation is **client-side** (stateless, via the SoC descriptor); the host passes the received coordinate through verbatim. The socket/client layers stay protocol-agnostic; this is where the protocol is (de)serialized on the host.

### List of the changes
- `device/api/umd/device/tt_device/simulation_tt_device.hpp`, `device/tt_device/simulation_tt_device.cpp` — `handle_request`; `adopt_socket` starts serving.
- `tests/api/test_simulation_connector.cpp` — CI-gated e2e (`HostServesClientMemoryOverSocket`): in-process client reads what the host wrote and writes what the host reads back, over the socket.

### Testing
Build + pre-commit clean; no-card sim suites pass. `HostServesClientMemoryOverSocket` runs in the simulation CI lane (`TT_UMD_SIMULATOR`).

### API Changes
None (internal). `SimulationTTDevice` gains a private `handle_request`; `adopt_socket` now serves.

### DIFF
--- a/device/tt_device/simulation_tt_device.cpp
+++ b/device/tt_device/simulation_tt_device.cpp
@@ -4,6 +4,8 @@
 
 #include "umd/device/tt_device/simulation_tt_device.hpp"
 
+#include <fmt/format.h>
+
 #include <tt-logger/tt-logger.hpp>
 
 #include "noc_access.hpp"
@@ -11,6 +13,7 @@
 #include "umd/device/arch/architecture_implementation.hpp"
 #include "umd/device/chip_helpers/simulation_tlb_allocator.hpp"
 #include "umd/device/pcie/tlb_window.hpp"
+#include "umd/device/simulation/simulation_server_protocol.hpp"
 #include "umd/device/soc_descriptor.hpp"
 #include "umd/device/types/arch.hpp"
 #include "umd/device/types/core_coordinates.hpp"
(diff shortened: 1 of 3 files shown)


device/tt_device/simulation_tt_device.cpp
```diff
@@ -28,7 +31,53 @@ SimulationTTDevice::SimulationTTDevice(
...
+    const SimulationServerRequest request = decode_request(request_bytes);
+
+    // The client already translated the coordinate (translation is stateless and client-side), so
+    // pass it through verbatim as LITERAL -- the shared read/write skeleton must not translate
+    // again. CoreCoord defaults to CoreType::UNSPECIFIED + CoordSystem::LITERAL.
+    const CoreCoord core{request.x, request.y};
+
+    SimulationServerResponse response;
+    try {
+        switch (request.command) {
```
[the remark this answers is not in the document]
pjanevskiTT: The decode throw is already contained: handle_request runs on the per-connection serving thread inside a try/catch (see do_accept), so a malformed request drops that one connection rather than crashing the thread or host. Moving the decode inside handle_request's try wouldn't help either -- the reply message type depends on the command we failed to decode, so there's no typed error response we could form for undecodable bytes. Dropping the connection on garbage is the intended behavior; leaving as-is.

device/tt_device/simulation_tt_device.cpp
```diff
@@ -28,7 +31,53 @@ SimulationTTDevice::SimulationTTDevice(
 
 SimulationTTDevice::~SimulationTTDevice() = default;
 
-void SimulationTTDevice::adopt_socket(std::unique_ptr<SimulationServerSocket> socket) { socket_ = std::move(socket); }
+void SimulationTTDevice::adopt_socket(std::unique_ptr<SimulationServerSocket> socket) {
+    socket_ = std::move(socket);
+    // Begin serving remote clients now that the backend is up. Capturing `this` is safe: the
```
nbuncicTT: Omit the impl. detail of capturing this which is safe.

device/tt_device/simulation_tt_device.cpp
```diff
@@ -28,7 +31,53 @@ SimulationTTDevice::SimulationTTDevice(
...
+
+    SimulationServerResponse response;
+    try {
+        switch (request.command) {
+            case SimulationServerCommand::Read:
+                response.data.resize(request.size);
+                read_from_device(response.data.data(), core, request.address, request.size);
+                break;
+            case SimulationServerCommand::Write:
+                UMD_ASSERT(
```
nbuncicTT, line 56: nit: Prefer to use `if(condition) { UMD_THROW ...}` instead of `UMD_ASSERT`, as in C++ asserts are compile time check, and in C they're runtime checks which are only used with debug flags included (@aleksamarkovicTT to confirm). Hence to be uniform across the codebase, let's use `UMD_THROW`/

nbuncicTT (approved):

device/tt_device/simulation_tt_device.cpp
pjanevskiTT: Fixed: removed the 'capturing this is safe' implementation detail from the adopt_socket comment.

device/tt_device/simulation_tt_device.cpp
pjanevskiTT: UMD_ASSERT here isn't a C `assert` -- in this codebase it's a runtime check that throws error::RuntimeError (equivalent to UMD_THROW), used pervasively across UMD for exactly this kind of input validation, so it is not compiled out in release. Happy to switch to `if (...) UMD_THROW(...)` if that's the agreed convention -- leaving this open pending @aleksamarkovicTT's confirmation so we settle it once for the codebase rather than diverging a single call site.

PR #3014, opened 2026-07-13
Simulation: client-mode device memory I/O over the socket

pjanevskiTT:
### Issue
Part of #2677 (simulation server process). Rebinds the client-mode device's `read_from_device`/`write_to_device` onto the socket, completing the remote read/write loop (the host side is #3008). Stacked on the `client_` hoist (#3013).

### Description
In client mode (`client_` set, no local backend), the base `SimulationTTDevice`'s `read_from_device`/`write_to_device` no longer throw. They now:
- translate the `CoreCoord` **client-side** (stateless, via the SoC descriptor),
- `encode` a `SimulationServerRequest`,
- send it over `client_->transact` (framed request/response on the socket),
- `decode_response`, throw on a nonzero status, and apply the reply.

The host (#3008) receives the already-translated coordinate and passes it through verbatim (LITERAL), so the effective backend coordinate matches a direct host access. The client branch runs before the host-only checks (`is_device_closed`, backend hooks) and holds `device_lock`, so concurrent callers don't interleave request/response on the single socket.

**Scope:** read/write only. Client-mode `get_io_window` (TLB windows over the socket) still throws — that op isn't in the protocol yet.

### List of the changes
- `SimulationTTDevice::read_from_device`/`write_to_device` — dispatch on a named `client_mode()` to `client_read`/`client_write` (socket) or `host_read`/`host_write` (local backend); each path is self-contained and takes `device_lock` itself (client always; host after its `is_device_closed()` gate).
- `tests/api/test_simulation_connector.cpp` — CI-gated `ClientDeviceReadsAndWritesOverSocket` (two in-process opens: host + client device; client writes → host reads it, host writes → client reads it). Also corrects the now-outdated `ThrowsWhen…` test's comment/name (client attach is implemented; it throws on the invalid simulator directory).

### Testing
Build + pre-commit clean; no-card sim suites pass. The e2e runs in the simulation CI lane (`TT_UMD_SIMULATOR`).

> **Note (rebase):** #3013 (the `client_` hoist) merged to `main` without the small encapsulation/naming cleanup that was reviewed on it, so that cleanup is folded in here instead: a protected base client-mode constructor + `attach_client()`/`detach_client()` (so `client_` isn't a bare protected member the derived classes poke), and the "hoisted"→"pulled up" comment wording.

### API Changes
None (internal).

### DIFF
--- a/device/api/umd/device/tt_device/rtl_simulation_tt_device.hpp
+++ b/device/api/umd/device/tt_device/rtl_simulation_tt_device.hpp
@@ -86,8 +86,8 @@ private:
 
     // setup_ runs at construction, teardown_ at destruction -- the one real host-vs-client
     // difference today: host mode drives the in-process RTL backend (communicator_), client mode
-    // drives the remote host (client_->attach()/detach()). Note client_ is owned by the
-    // SimulationTTDevice base (hoisted there), not declared in this class.
+    // drives the remote host (attach_client()/detach_client()). Note client_ is owned by the
+    // SimulationTTDevice base (pulled up there), not declared in this class.
     std::function<void()> setup_;
     std::function<void()> teardown_;
 
--- a/device/api/umd/device/tt_device/rtl_simulation_tt_device.hpp
+++ b/device/api/umd/device/tt_device/rtl_simulation_tt_device.hpp
@@ -44,10 +44,11 @@ public:
     static std::unique_ptr<RtlSimulationTTDevice> create(
         const std::filesystem::path& simulator_directory, int num_host_mem_channels = 0);
 
-    // Builds a client-mode device that attaches to a live host (see the .cpp for how the SoC
-    // descriptor is sourced); discovery uses this when a host already owns the socket.
+    // Builds a client-mode device that attaches to a live host and sources its SoC descriptor from
+    // the host over the socket (see the .cpp); discovery uses this when a host already owns the
+    // socket.
     static std::unique_ptr<RtlSimulationTTDevice> create_client(
-        const std::filesystem::path& simulator_directory, ChipId chip_id, std::unique_ptr<SimulationClient> client);
+        ChipId chip_id, std::unique_ptr<SimulationClient> client);
 
     void wait_arc_core_start(const std::chrono::milliseconds timeout_ms = timeout::ARC_STARTUP_TIMEOUT) override;
     std::chrono::milliseconds wait_eth_core_training(
@@ -60,6 +61,8 @@ public:
     RtlSimCommunicator* get_communicator() { return communicator_.get(); }
 
 protected:
+    SimulationBackendType backend_type() const override { return SimulationBackendType::Rtl; }
+
     std::unique_ptr<TlbWindow> create_tlb_window(
         int tlb_index, size_t size, TlbMapping mapping, tlb_data config) override;
     void tile_read_bytes(tt_xy_pair core, uint64_t addr, void* mem_ptr, size_t size) override;
--- a/device/api/umd/device/tt_device/simulation_tt_device.hpp
+++ b/device/api/umd/device/tt_device/simulation_tt_device.hpp
@@ -86,8 +86,14 @@ protected:
     void retrain_dram_core(const uint32_t dram_channel) override;
 
     // Client-mode constructor: the device does not own a local simulator, so it has no simulator
-    // directory or sysmem manager -- those live on the remote host reached over the socket.
-    SimulationTTDevice() = default;
+    // directory or sysmem manager -- those live on the remote host reached over the socket. Takes
+    // the client here so client_ is initialized through the base, not written by each derived ctor.
+    explicit SimulationTTDevice(std::unique_ptr<SimulationClient> client);
+
+    // Attach to / detach from the remote host in client mode. Both derived devices drive their
+    // client-mode lifecycle (setup_/teardown_) through these rather than touching client_ directly.
+    void attach_client();
+    void detach_client();
 
     // Build tlb_allocator_ once the backend knows its BAR0 base (0 for RTL, PCI-probed for TTSim).
     void init_tlb_allocator(uint64_t bar0_base);
@@ -137,7 +143,7 @@ protected:
     std::unique_ptr<SimulationServerSocket> socket_;
 
     // Set only in client mode: the remote host this device talks to, instead of owning a local
-    // backend. Hoisted here from the derived devices (both held an identical member) since it is
+    // backend. Pulled up here from the derived devices (both held an identical member) since it is
     // the client-mode counterpart of the shared lifecycle; a follow-up wires read_from_device /
     // write_to_device to dispatch through it. Null in host/local mode.
     std::unique_ptr<SimulationClient> client_;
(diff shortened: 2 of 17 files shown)


device/tt_device/simulation_tt_device.cpp
```diff
@@ -121,6 +130,47 @@ void SimulationTTDevice::read_from_device(void* mem_ptr, CoreCoord core, uint64_
...
+    request.size = static_cast<uint32_t>(size);
+    const auto* bytes = static_cast<const uint8_t*>(mem_ptr);
+    request.data.assign(bytes, bytes + size);
+
+    const SimulationServerResponse response = decode_response(client_->transact(encode(request)));
+    UMD_ASSERT(
+        response.status == 0,
+        error::RuntimeError,
+        fmt::format("Remote simulation host failed the write (status {})", response.status));
+}
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed: client_write now asserts `size` fits uint32 before casting to the protocol's size field.

device/tt_device/simulation_tt_device.cpp
```diff
@@ -121,6 +130,47 @@ void SimulationTTDevice::read_from_device(void* mem_ptr, CoreCoord core, uint64_
...
+    UMD_ASSERT(
+        response.status == 0,
+        error::RuntimeError,
+        fmt::format("Remote simulation host failed the read (status {})", response.status));
+    UMD_ASSERT(
+        response.data.size() == size,
+        error::RuntimeError,
+        fmt::format("Remote read returned {} bytes, expected {}", response.data.size(), size));
+    std::memcpy(mem_ptr, response.data.data(), size);
+}
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed: client_read now asserts `size` fits uint32, and the memcpy is guarded on `size > 0`.

tests/api/test_simulation_connector.cpp
```diff
@@ -52,18 +52,19 @@ TEST(SimulationConnector, CreatesHostDeviceAndExposesSocket) {
...
     const std::filesystem::path socket = SimulationServerSocket::default_socket_path(0);
     auto host = SimulationServerSocket::try_create(socket);
     if (host == nullptr) {
         GTEST_SKIP() << "A live simulation host already holds " << socket << "; skipping.";
     }
 
     SimulationConnectorOptions options;
-    options.simulator_directory = "unused";  // discover() throws before it is touched.
+    options.simulator_directory = "unused";  // client creation reads the SoC descriptor from here and throws.
     EXPECT_THROW(SimulationConnector::discover(options), std::exception);
```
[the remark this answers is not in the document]
pjanevskiTT: Addressed downstream in #3020, which reworked this test: the throw-path test no longer relies on a relative "unused" path -- it uses an unmistakably-invalid "/nonexistent/..." path.

device/api/umd/device/tt_device/rtl_simulation_tt_device.hpp
```diff
@@ -86,13 +86,11 @@ class RtlSimulationTTDevice : public SimulationTTDevice {
 
     // setup_ runs at construction, teardown_ at destruction -- the one real host-vs-client
     // difference today: host mode drives the in-process RTL backend (communicator_), client mode
-    // drives the remote host (client_->attach()/detach()).
+    // drives the remote host (client_->attach()/detach()). Note client_ is owned by the
+    // SimulationTTDevice base (hoisted there), not declared in this class.
```
nbuncicTT: nit: replace hoisted with a more common, technical term

device/api/umd/device/tt_device/simulation_tt_device.hpp
```diff
@@ -135,6 +136,12 @@ class SimulationTTDevice : public TTDevice {
     // The host keeps its own direct in-process fast path; the socket is for remote clients.
     std::unique_ptr<SimulationServerSocket> socket_;
 
+    // Set only in client mode: the remote host this device talks to, instead of owning a local
+    // backend. Hoisted here from the derived devices (both held an identical member) since it is
+    // the client-mode counterpart of the shared lifecycle; a follow-up wires read_from_device /
+    // write_to_device to dispatch through it. Null in host/local mode.
+    std::unique_ptr<SimulationClient> client_;
```
nbuncicTT, line 154: So this class violates the rule where the data shouldn't be protected, and this is an issue because you can't control invariants and ordering over it via an API.

Do the following:
- Give the base a protected client-mode constructor taking `std::unique_ptr<SimulationClient>` and initialize `client_` in the base's init list. Derived ctors become `: SimulationTTDevice(std::move(client))`. This removes the derived write, and simultaneously eliminates the init-list and body the `std::move()` in the body of the ctor.
- Move the attach/detach into base methods (e.g. `attach_client()`/`detach_client()`), since both derived classes do exactly `client_->attach()` / `client_->detach()`. It removes the derived reads. So it would become:
```cpp
// SimulationTTDevice (base)
protected:
    void attach_client()  { client_->attach(); }
    void detach_client()  { if (client_) client_->detach(); }

// Derived becomes:
setup_    = [this] { attach_client(); };
teardown_ = [this] { detach_client(); };
```

device/tt_device/simulation_tt_device.cpp
```diff
@@ -79,10 +83,14 @@ std::vector<uint8_t> SimulationTTDevice::handle_request(const std::vector<uint8_
...
 void SimulationTTDevice::write_to_device(
     const void* mem_ptr, CoreCoord core, uint64_t addr, size_t size, NocId noc_id) {
-    // Client-mode devices run no local backend (tlb_allocator_/communicator_ are never built), so
-    // device I/O is unavailable. Fail loudly instead of dereferencing a null communicator_.
-    UMD_ASSERT(
-        tlb_allocator_ != nullptr, error::RuntimeError, "Client-mode simulation device I/O is not available yet.");
+    // Client mode: no local backend -- marshal the write to the remote host over the socket. The
+    // client branch comes first because the host-mode checks below (is_device_closed, the backend
+    // hooks) assume a local backend that a client does not have.
+    if (client_) {
```
nbuncicTT: Make the client-vs-host dispatch explicit. Right now the device's role is inferred from `if (client_)`, which is a raw null-check buried mid-function and the reader has to know that "client_ non-null equals client mode" to follow it. Both roles are also disjoint code sharing nothing, so they read better as named branches over named helpers:

```cpp
void SimulationTTDevice::write_to_device(const void* mem_ptr, CoreCoord core, uint64_t addr, size_t size, NocId noc_id) {
    if (client_mode()) {
        client_write(core, addr, mem_ptr, size);
    } else {
        host_write(core, addr, mem_ptr, size);
    }
}
```
with:

- `bool client_mode() const { return client_ != nullptr; }` this way the intent is named.
- `client_write`/`client_read` already exists; extract the current fall-through body into host_write (and the same for read) so each path is self-contained.

Two things to preserve in the move:

- Locking: today the client branch takes `device_lock` itself while the host branch takes it lower down — decide whether `write_to_device` locks once up front, or each helper locks; don't drop the lock on either path.
- `is_device_closed()`: it currently guards only the host path. Keep it inside host_write (or lift it and make the client-mode behavior on a closed device a conscious choice, per the earlier asymmetry note).
Same treatment for `read_from_device`.

nbuncicTT (approved):

device/api/umd/device/tt_device/rtl_simulation_tt_device.hpp
pjanevskiTT: Replaced "hoisted"/"Hoisted" with "pulled up" (Fowler's Pull Up Field) across the base and both derived headers. #3013 merged before this landed, so the wording fix is folded into this PR.

device/api/umd/device/tt_device/simulation_tt_device.hpp
pjanevskiTT: Done. The base now has a protected client-mode constructor `SimulationTTDevice(std::unique_ptr<SimulationClient>)` that initializes `client_` in the base init-list, plus `attach_client()`/`detach_client()`. Both derived ctors are now `: SimulationTTDevice(std::move(client))`, and their `setup_`/`teardown_` call `attach_client()`/`detach_client()` instead of touching `client_` directly. Folded into this PR since #3013 merged before it landed.

device/tt_device/simulation_tt_device.cpp
pjanevskiTT: Done. `write_to_device`/`read_from_device` now dispatch on a named `client_mode()` to `client_read`/`client_write` (socket) or `host_read`/`host_write` (local backend). Each path is self-contained and takes `device_lock` itself — the client path always, the host path after its `is_device_closed()` gate — so the lock is not dropped on either, and `is_device_closed()` stays inside the host path.

PR #3018, opened 2026-07-14
Add repeat and card filter for gh actions

nbuncicTT:
### Issue
/

### Description
This pull request introduces enhanced configurability and improved developer ergonomics for CI test workflows. It adds new options to control which test cards to run and how many times to repeat each, and provides a helper script to make running and reproducing CI workflows easier.

#### Workflow Input and Matrix Enhancements

* Added `repeat-count` and `cards-filter` inputs to both `build-and-run-all-tests.yml` and `build-and-run-all-tests-multiple-builds.yml` workflows, allowing users to specify how many times to repeat each test card and/or to run only a selected subset of cards. 

* Extended the test matrix generation logic to handle the new repeat and filter options, including validation to prevent exceeding GitHub's 256-job matrix cap. The matrix now repeats each card as specified and only includes selected cards if a filter is provided.

#### Job Parameterization

* Updated all test jobs in both workflows to pass the new `repeat-count` and `cards-filter` parameters through to the reusable workflow, ensuring the selected configuration is respected throughout the CI run.

##### Developer Tooling

* Added a new script `scripts/repro-ci.sh` that enables developers to easily dispatch CI runs from the command line with custom repeat and card selection options, including validation and a dry-run mode. It also supports listing valid build types and cards, making it easier to reproduce or customize CI runs locally.

### Testing
CI + Manual

### API Changes
/


aleksamarkovicTT (approved):

.github/scripts/repro-ci.py
```diff
@@ -0,0 +1,215 @@
+#!/usr/bin/env python3
```
pjanevskiTT, line 1: sorry for a late review, I would maybe prefer we move this to internal repo but not a strong opinion. Otherwise changes look good

PR #3019, opened 2026-07-14
Simulation: serve device identity (GET_DEVICE_INFO) over the socket

pjanevskiTT:
### Issue
Part of #2677 (simulation server process). A client-mode device currently reads its SoC descriptor from a local simulator build. This PR lets the host serve its device identity over the socket so a client can build a matching `SocDescriptor` with no local build — the prerequisite for deciding the host/client role purely from `simulator_path` (next PR). Stacked on #3014.

### Description
Adds one request command (`GET_DEVICE_INFO`) and one response message (`SimulationServerDeviceInfo`) to the wire protocol, carrying the arch, the **backend kind (TTSim/RTL)**, the full SoC-descriptor YAML text, `noc_translation_enabled`, and all five harvesting masks (tensix/dram/eth/l2cpu/pcie).

The served backend kind lets a client instantiate the matching device class — the prerequisite for the next PR, which decides the host/client role purely from `simulator_path` (a client's path is a socket *directory* that no longer carries the `.so` hint, so the class comes off the wire instead).

A small always-paired bridge translates between a `SocDescriptor` and that message:
- `describe_device()` (host): reads the descriptor's source YAML and captures its arch + masks.
- `build_soc_descriptor()` (client): materializes the served YAML to a temp file (the only way to construct `SocArchDescriptor`), applies the masks via `ChipInfo`, and asserts the YAML-derived arch matches the served arch.

The host answers `GET_DEVICE_INFO` from `get_soc_descriptor()` (read-only; no `device_lock`, before the read/write skeleton). `create_client()` (TTSim + RTL) now fetches the identity over the attached client and builds its descriptor from the wire instead of from `get_soc_descriptor_path_from_simulator_path()`; it no longer takes a `simulator_directory`.

Coordinate model is unchanged: the client still translates coordinates client-side, and the faithful descriptor round-trip is what keeps that translation correct.

### List of the changes
- `simulation_server_protocol.{fbs,hpp,cpp}` — `GET_DEVICE_INFO` command + `SimulationServerDeviceInfo` message (incl. `backend_kind`) + codec.
- `SimulationTTDevice::backend_kind()` — pure virtual returning the device's simulator kind (TTSim/RTL), overridden per device; the host fills the served `backend_kind` from it.
- `simulation_device_identity.{hpp,cpp}` (new) — `describe_device` / `build_soc_descriptor` / `fetch_device_info`.
- `simulation_tt_device.cpp` — `handle_request` serves `GET_DEVICE_INFO`.
- `tt_sim_tt_device` / `rtl_simulation_tt_device` — `create_client` sourced from the wire; drops `simulator_directory`.
- `simulation_connector.cpp` — updated `create_client` call sites.
- Tests: protocol round-trip + `SimulationDeviceIdentity` describe→rebuild fidelity (incl. masks); CI-gated e2e extended to fetch device info.

### Testing
Build + pre-commit clean; no-card sim unit suites pass (`SimulationServerProtocol.*` + `SimulationDeviceIdentity.*`, 9/9). The e2e runs in the simulation CI lane (`TT_UMD_SIMULATOR`).

### API Changes
`TTSimTTDevice::create_client` / `RtlSimulationTTDevice::create_client` drop the `simulator_directory` parameter (internal factories; the client now builds its descriptor from the wire). Otherwise internal.

### DIFF
--- a/device/api/umd/device/simulation/simulation_device_identity.hpp
+++ b/device/api/umd/device/simulation/simulation_device_identity.hpp
@@ -1,0 +1,26 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include "umd/device/simulation/simulation_client.hpp"
+#include "umd/device/simulation/simulation_server_protocol.hpp"
+#include "umd/device/soc_descriptor.hpp"
+
+namespace tt::umd {
+
+// Bridges a SocDescriptor and the GET_DEVICE_INFO wire message, so a client can rebuild the host's
+// device identity without a local simulator build.
+
+// Host side: package this device's SocDescriptor into a wire message, so a client can receive it
+// and build its own matching SocDescriptor. backend_type records which simulator the host runs.
+SimulationServerDeviceInfo describe_device(const SocDescriptor& soc_descriptor, SimulationBackendType backend_type);
+
+// Client side: build a SocDescriptor from a wire message served by the host.
+SocDescriptor build_soc_descriptor(const SimulationServerDeviceInfo& device_info);
+
+// Client side: request the host's device identity over the socket.
+SimulationServerDeviceInfo fetch_device_info_from_host(SimulationClient& client);
+
+}  // namespace tt::umd
--- a/device/api/umd/device/simulation/simulation_server_protocol.hpp
+++ b/device/api/umd/device/simulation/simulation_server_protocol.hpp
@@ -6,6 +6,7 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <string>
 #include <vector>
 
 namespace tt::umd {
@@ -26,6 +27,14 @@ namespace tt::umd {
 enum class SimulationServerCommand : int8_t {
     Read = 0,
     Write = 1,
+    GetDeviceInfo = 2,
+};
+
+// Which simulator the host runs; mirrors wire::SimulationBackendType. Served as part of the device
+// identity so a client can build the matching device class without a local simulator build.
+enum class SimulationBackendType : int8_t {
+    TTSim = 0,
+    Rtl = 1,
 };
 
 // A device-memory access request addressed by (core x/y, address, size).
@@ -47,16 +56,40 @@ struct SimulationServerResponse {
     std::vector<uint8_t> data;
 };
 
+// Device identity a host serves in reply to a GetDeviceInfo request; mirrors
+// wire::SimulationServerDeviceInfo. Everything a client needs to build a matching SocDescriptor.
+struct SimulationServerDeviceInfo {
+    // 0 on success; nonzero signals a host-side failure (e.g. the host had no YAML to serve).
+    int32_t status = 0;
+    // tt::ARCH value of the served device.
+    int32_t arch = 0;
+    // Which simulator the host runs, so the client instantiates the matching device class.
+    SimulationBackendType backend_type = SimulationBackendType::TTSim;
+    // Full text of the host's SoC descriptor YAML file, so the client can build a matching one.
+    std::string soc_descriptor_yaml;
+    // Whether the host applies NOC translation.
+    bool noc_translation_enabled = false;
+    // Per-chip harvesting masks (logical-coordinate bitmasks; see HarvestingMasks).
+    uint32_t tensix_harvesting_mask = 0;
+    uint32_t dram_harvesting_mask = 0;
+    uint32_t eth_harvesting_mask = 0;
+    uint32_t l2cpu_harvesting_mask = 0;
+    uint32_t pcie_harvesting_mask = 0;
+};
+
 // Serialize a message to a FlatBuffers payload.
 std::vector<uint8_t> encode(const SimulationServerRequest& request);
 std::vector<uint8_t> encode(const SimulationServerResponse& response);
+std::vector<uint8_t> encode(const SimulationServerDeviceInfo& device_info);
 
 // Parse a message back from a FlatBuffers payload. The buffer is verified against the schema
 // before any field is read (it comes off a socket, so it may be malformed or truncated); a bad
 // or empty buffer throws error::RuntimeError rather than risking an out-of-bounds read.
 SimulationServerRequest decode_request(const uint8_t* data, size_t size);
 SimulationServerResponse decode_response(const uint8_t* data, size_t size);
+SimulationServerDeviceInfo decode_device_info(const uint8_t* data, size_t size);
 SimulationServerRequest decode_request(const std::vector<uint8_t>& bytes);
 SimulationServerResponse decode_response(const std::vector<uint8_t>& bytes);
+SimulationServerDeviceInfo decode_device_info(const std::vector<uint8_t>& bytes);
 
 }  // namespace tt::umd
--- a/device/api/umd/device/tt_device/rtl_simulation_tt_device.hpp
+++ b/device/api/umd/device/tt_device/rtl_simulation_tt_device.hpp
@@ -44,10 +44,11 @@ public:
     static std::unique_ptr<RtlSimulationTTDevice> create(
         const std::filesystem::path& simulator_directory, int num_host_mem_channels = 0);
 
-    // Builds a client-mode device that attaches to a live host (see the .cpp for how the SoC
-    // descriptor is sourced); discovery uses this when a host already owns the socket.
+    // Builds a client-mode device that attaches to a live host and sources its SoC descriptor from
+    // the host over the socket (see the .cpp); discovery uses this when a host already owns the
+    // socket.
     static std::unique_ptr<RtlSimulationTTDevice> create_client(
-        const std::filesystem::path& simulator_directory, ChipId chip_id, std::unique_ptr<SimulationClient> client);
+        ChipId chip_id, std::unique_ptr<SimulationClient> client);
 
     void wait_arc_core_start(const std::chrono::milliseconds timeout_ms = timeout::ARC_STARTUP_TIMEOUT) override;
     std::chrono::milliseconds wait_eth_core_training(
@@ -60,6 +61,8 @@ public:
     RtlSimCommunicator* get_communicator() { return communicator_.get(); }
 
 protected:
+    SimulationBackendType backend_type() const override { return SimulationBackendType::Rtl; }
+
     std::unique_ptr<TlbWindow> create_tlb_window(
         int tlb_index, size_t size, TlbMapping mapping, tlb_data config) override;
     void tile_read_bytes(tt_xy_pair core, uint64_t addr, void* mem_ptr, size_t size) override;
--- a/device/simulation/simulation_device_identity.cpp
+++ b/device/simulation/simulation_device_identity.cpp
@@ -1,0 +1,128 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "umd/device/simulation/simulation_device_identity.hpp"
+
+#include <fmt/format.h>
+#include <unistd.h>  // write, close (temp-file materialization via mkstemps)
+
+#include <cerrno>
+#include <cstdlib>  // mkstemps
+#include <cstring>  // strerror, strlen
+#include <filesystem>
+#include <fstream>
+#include <memory>
+#include <sstream>
+#include <vector>
+
+#include "umd/device/simulation/simulation_client.hpp"
+#include "umd/device/soc_arch_descriptor.hpp"
+#include "umd/device/utils/error.hpp"
+
+namespace tt::umd {
+
+namespace {
+
+// A temp file that removes itself on destruction (best-effort), so callers don't leak it on a
+// throw between creation and cleanup.
+struct ScopedTempFile {
+    std::filesystem::path path;
+
+    ~ScopedTempFile() {
+        std::error_code ec;
+        std::filesystem::remove(path, ec);
+    }
+};
+
+// Materialize `contents` to a unique temp file ending in `suffix` (mkstemps avoids name races).
+ScopedTempFile write_temp_file(const std::string& contents, const char* suffix) {
+    const std::string name_template =
+        (std::filesystem::temp_directory_path() / (std::string("tt-umd-sim-socdesc-XXXXXX") + suffix)).string();
+    std::vector<char> path_buf(name_template.begin(), name_template.end());
+    path_buf.push_back('\0');
+    const int fd = ::mkstemps(path_buf.data(), static_cast<int>(std::strlen(suffix)));
+    UMD_ASSERT(
+        fd >= 0,
+        error::RuntimeError,
+        fmt::format("Failed to create a temp SoC descriptor file: {}", std::strerror(errno)));
+    const ssize_t written = ::write(fd, contents.data(), contents.size());
+    ::close(fd);
+    UMD_ASSERT(
+        written == static_cast<ssize_t>(contents.size()),
+        error::RuntimeError,
+        fmt::format("Failed to write temp file {}", path_buf.data()));
+    return ScopedTempFile{std::filesystem::path(path_buf.data())};
+}
+
+}  // namespace
+
+SimulationServerDeviceInfo describe_device(const SocDescriptor& soc_descriptor, SimulationBackendType backend_type) {
+    const std::string& yaml_path = soc_descriptor.get_arch_descriptor().get_device_descriptor_file_path();
+    UMD_ASSERT(
+        !yaml_path.empty(),
+        error::RuntimeError,
+        "Cannot serve device info: the SoC descriptor was not built from a YAML file.");
+
+    std::ifstream in(yaml_path);
+    UMD_ASSERT(in.good(), error::RuntimeError, fmt::format("Failed to open SoC descriptor YAML at {}", yaml_path));
+    std::stringstream buffer;
+    buffer << in.rdbuf();
+
+    SimulationServerDeviceInfo info;
+    info.status = 0;
+    info.arch = static_cast<int32_t>(soc_descriptor.arch);
+    info.backend_type = backend_type;
+    info.soc_descriptor_yaml = buffer.str();
+    info.noc_translation_enabled = soc_descriptor.noc_translation_enabled;
+    info.tensix_harvesting_mask = soc_descriptor.harvesting_masks.tensix_harvesting_mask;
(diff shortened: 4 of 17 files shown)


device/simulation/simulation_device_identity.cpp
```diff
@@ -0,0 +1,101 @@
...
+    chip_info.harvesting_masks.dram_harvesting_mask = device_info.dram_harvesting_mask;
+    chip_info.harvesting_masks.eth_harvesting_mask = device_info.eth_harvesting_mask;
+    chip_info.harvesting_masks.l2cpu_harvesting_mask = device_info.l2cpu_harvesting_mask;
+    chip_info.harvesting_masks.pcie_harvesting_mask = device_info.pcie_harvesting_mask;
+
+    SocDescriptor soc_descriptor(std::make_shared<SocArchDescriptor>(tmp.string()), chip_info);
+
+    std::error_code ec;
+    std::filesystem::remove(tmp, ec);  // best-effort cleanup
+
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed: build_soc_descriptor now materializes the YAML with mkstemps (unique filename), writes via the returned fd, and removes it -- no more predictable name.

device/api/umd/device/simulation/simulation_device_identity.hpp
```diff
@@ -0,0 +1,29 @@
...
+// backend_kind records which simulator the host runs so the client builds the matching device class.
+SimulationServerDeviceInfo describe_device(const SocDescriptor& soc_descriptor, SimulationBackendKind backend_kind);
+
+// Client side: rebuild a SocDescriptor from a served message. Materializes the YAML text to a
+// temporary file because SocArchDescriptor only constructs from a path.
+SocDescriptor build_soc_descriptor(const SimulationServerDeviceInfo& device_info);
+
+// Client side: request the host's device identity over an already-attached client. Attaches the
+// client if it is not open yet. Throws if the host reports a nonzero status.
+SimulationServerDeviceInfo fetch_device_info(SimulationClient& client);
```
nbuncicTT: nit: rename to `fetch_device_info_from_host`

device/api/umd/device/simulation/simulation_device_identity.hpp
```diff
@@ -0,0 +1,29 @@
...
+// Host side: capture a live SocDescriptor as a wire message. Reads the descriptor's source YAML
+// file (SocArchDescriptor::get_device_descriptor_file_path()); throws if that path is empty.
+// backend_kind records which simulator the host runs so the client builds the matching device class.
+SimulationServerDeviceInfo describe_device(const SocDescriptor& soc_descriptor, SimulationBackendKind backend_kind);
+
+// Client side: rebuild a SocDescriptor from a served message. Materializes the YAML text to a
+// temporary file because SocArchDescriptor only constructs from a path.
+SocDescriptor build_soc_descriptor(const SimulationServerDeviceInfo& device_info);
+
+// Client side: request the host's device identity over an already-attached client. Attaches the
```
nbuncicTT: Comment is confusing, says to request over an already-attached client and then says attaches the client if it is not open yet. Omit these non-important details.

device/api/umd/device/simulation/simulation_device_identity.hpp
```diff
@@ -0,0 +1,29 @@
...
+#include "umd/device/soc_descriptor.hpp"
+
+namespace tt::umd {
+
+// Bridges a SocDescriptor and the GET_DEVICE_INFO wire message so a client can reconstruct the
+// host's device identity without reading a local simulator build.
+
+// Host side: capture a live SocDescriptor as a wire message. Reads the descriptor's source YAML
+// file (SocArchDescriptor::get_device_descriptor_file_path()); throws if that path is empty.
+// backend_kind records which simulator the host runs so the client builds the matching device class.
```
nbuncicTT: Let's keep the comments simple, they're trying to detail everything but too tightly coupled to the implementation (which always changes over time). Do something like:
```cpp
// Bridges a SocDescriptor and the GET_DEVICE_INFO wire message, so a client can rebuild the host's
// device identity without a local simulator build.

// Host side: package this device's SocDescriptor into a wire message, so a client can receive it
// and build its own matching SocDescriptor.
```

device/api/umd/device/simulation/simulation_device_identity.hpp
```diff
@@ -0,0 +1,29 @@
...
+SimulationServerDeviceInfo describe_device(const SocDescriptor& soc_descriptor, SimulationBackendKind backend_kind);
+
+// Client side: rebuild a SocDescriptor from a served message. Materializes the YAML text to a
```
nbuncicTT: Make it simpler:
```cpp
// Client side: build a SocDescriptor from a wire message served by the host.
```

device/api/umd/device/simulation/simulation_device_identity.hpp
nbuncicTT: Comment should be simpler:
```cpp
// Client side: fetch the host's device identity over the socket. Throws if the host reports failure.
```

device/api/umd/device/simulation/simulation_server_protocol.hpp
```diff
@@ -26,6 +27,14 @@ namespace tt::umd {
 enum class SimulationServerCommand : int8_t {
     Read = 0,
     Write = 1,
+    GetDeviceInfo = 2,
+};
+
+// Which simulator the host runs; mirrors wire::SimulationBackendKind. Served as part of the device
+// identity so a client can build the matching device class without a local simulator build.
+enum class SimulationBackendKind : int8_t {
+    TTSim = 0,
```
nbuncicTT, line 36: Match the casing as in the rest of the repo which upper-snake case.

nbuncicTT:
```cpp
enum class SimulationBackendKind : int8_t {
    TTSIM = 0,
    RTL = 1,
};
```

nbuncicTT (commented):
Continuing tomorrow

device/api/umd/device/simulation/simulation_server_protocol.hpp
```diff
@@ -26,6 +27,14 @@ namespace tt::umd {
 enum class SimulationServerCommand : int8_t {
     Read = 0,
     Write = 1,
+    GetDeviceInfo = 2,
```
nbuncicTT, line 30: nit: Match the casing as in the rest of the repo which upper-snake case.

device/api/umd/device/tt_device/rtl_simulation_tt_device.hpp
```diff
@@ -60,6 +61,8 @@ class RtlSimulationTTDevice : public SimulationTTDevice {
     RtlSimCommunicator* get_communicator() { return communicator_.get(); }
 
 protected:
+    SimulationBackendKind backend_kind() const override { return SimulationBackendKind::Rtl; }
```
nbuncicTT: nit: Please rename from kind to type, so it's `SimulationBackendType` and `backend_type`. Kind is a loosely coupled classification, and here we have a concrete type of backend.

device/simulation/simulation_device_identity.cpp
```diff
@@ -0,0 +1,114 @@
...
+    info.pcie_harvesting_mask = soc_descriptor.harvesting_masks.pcie_harvesting_mask;
+    return info;
+}
+
+SocDescriptor build_soc_descriptor(const SimulationServerDeviceInfo& device_info) {
+    // SocArchDescriptor constructs only from a path, so materialize the served YAML to a *unique*
+    // temp file (mkstemps, not a predictable name -- avoids collisions/races between concurrent
+    // clients), build from it, then remove it. The SoC descriptor reads the YAML eagerly at
+    // construction, so the file is not needed afterwards (a client runs no backend, never re-reads).
+    constexpr const char* suffix = ".yaml";
```
nbuncicTT: Extract the temp-file materialization into a helper. The `mkstemps`/`write`/`close` block is a
self-contained "write a string to a temp file" op that clutters `build_soc_descriptor` and
currently leaks the file if anything after it throws (bad YAML, arch mismatch) and `remove` only
runs on the success path. Pulling it out both declutters and gives it RAII cleanup:

```cpp
// anonymous namespace in the .cpp
namespace {

// A temp file that removes itself on destruction (best-effort).
struct ScopedTempFile {
    std::filesystem::path path;
    ~ScopedTempFile() {
        std::error_code ec;
        std::filesystem::remove(path, ec);
    }
};

// Materialize `contents` to a unique temp file ending in `suffix` (mkstemps avoids name races).
ScopedTempFile write_temp_file(const std::string& contents, const char* suffix) {
    const std::string name_template =
        (std::filesystem::temp_directory_path() / (std::string("tt-umd-sim-socdesc-XXXXXX") + suffix)).string();
    std::vector<char> path_buf(name_template.begin(), name_template.end());
    path_buf.push_back('\0');
    const int fd = ::mkstemps(path_buf.data(), static_cast<int>(std::strlen(suffix)));
    UMD_ASSERT(
        fd >= 0, error::RuntimeError,
        fmt::format("Failed to create a temp SoC descriptor file: {}", std::strerror(errno)));
    const ssize_t written = ::write(fd, contents.data(), contents.size());
    ::close(fd);
    UMD_ASSERT(
        written == static_cast<ssize_t>(contents.size()), error::RuntimeError,
        fmt::format("Failed to write temp SoC descriptor file {}", path_buf.data()));
    return ScopedTempFile{std::filesystem::path(path_buf.data())};
}

}  // namespace

File: tests/api/test_simulation_server_protocol.cpp [hunk 10]
```diff
@@ -93,3 +93,39 @@ TEST(SimulationServerProtocol, DecodeTruncatedBufferThrows) {
...
+    info.backend_kind = SimulationBackendKind::Rtl;
+    info.soc_descriptor_yaml = "arch: wormhole_b0\ngrid: [10, 12]\n";
+    info.noc_translation_enabled = false;
+    info.tensix_harvesting_mask = 0x5;
+    info.dram_harvesting_mask = 0x0;
+    info.eth_harvesting_mask = 0x120;
+    info.l2cpu_harvesting_mask = 0x0;
+    info.pcie_harvesting_mask = 0x3;
+
+    const SimulationServerDeviceInfo decoded = decode_device_info(encode(info));
```
nbuncicTT, line 111: From my point of view, I'd rather have a symmetric pair of APIs instead of this asymmetry (which you have due to overloading rules in C++).

Currently we have:
- encode overloaded by type
- decode_device_info, decode_request, decode_reponse

I'd rather have encode_device_info, encode_request, etc.

Everything else stays the same. But your call.

tests/api/test_simulation_server_protocol.cpp
```diff
@@ -93,3 +93,39 @@ TEST(SimulationServerProtocol, DecodeTruncatedBufferThrows) {
 
     EXPECT_THROW(decode_request(encoded), std::exception);
 }
+
+// GET_DEVICE_INFO round-trips arch, YAML text, and harvesting masks.
+TEST(SimulationServerProtocol, DeviceInfoRoundTrip) {
```
nbuncicTT: `DeviceInfoSerializationRoundTrip` would be more precise for a name, as the round trip is for serializing/de-serializing data (and not a round trip through the sockets).

tests/api/test_simulation_connector.cpp
```diff
@@ -131,6 +131,16 @@ TEST(SimulationConnector, HostServesClientMemoryOverSocket) {
     std::vector<uint8_t> readback(pattern2.size());
     host->read_from_device(readback.data(), tensix, addr2, pattern2.size());
     EXPECT_EQ(readback, pattern2);
+
+    // The host answers GET_DEVICE_INFO with its arch and SoC-descriptor YAML.
```
nbuncicTT: Can this be moved out of this test into a separate one where we verify the round trip (serialization/de-serialization of the data through the socket)? And check every field?

Something like:
```cpp
TEST(SimulationConnector, HostServesDeviceInfoOverSocket) {
    const char* simulator_path = std::getenv("TT_UMD_SIMULATOR");
    if (simulator_path == nullptr) {
        GTEST_SKIP() << "TT_UMD_SIMULATOR is not set.";
    }

    const std::filesystem::path socket = SimulationServerSocket::default_socket_path(0);
    {
        auto probe = SimulationServerSocket::try_create(socket);
        if (probe == nullptr) {
            GTEST_SKIP() << "A live simulation host already holds " << socket << "; skipping.";
        }
    }

    SimulationConnectorOptions options;
    options.simulator_directory = simulator_path;
    auto devices = SimulationConnector::discover(options);
    ASSERT_EQ(devices.size(), 1u);
    TTDevice* host = devices.at(0).get();
    ASSERT_NE(host, nullptr);
    const SocDescriptor& soc = host->get_soc_descriptor();

    SimulationClient client(socket);
    client.attach();

    SimulationServerRequest info_request;
    info_request.command = SimulationServerCommand::GetDeviceInfo;
    const SimulationServerDeviceInfo info = decode_device_info(client.transact(encode(info_request)));

    // The served identity must match the host's live descriptor, field for field.
    EXPECT_EQ(info.status, 0);
    EXPECT_EQ(info.arch, static_cast<int32_t>(soc.arch));
    const bool is_ttsim = std::filesystem::path(simulator_path).extension() == ".so";
    EXPECT_EQ(info.backend_kind, is_ttsim ? SimulationBackendKind::TTSim : SimulationBackendKind::Rtl);
    EXPECT_FALSE(info.soc_descriptor_yaml.empty());
    EXPECT_EQ(info.noc_translation_enabled, soc.noc_translation_enabled);
    EXPECT_EQ(info.tensix_harvesting_mask, soc.harvesting_masks.tensix_harvesting_mask);
    EXPECT_EQ(info.dram_harvesting_mask, soc.harvesting_masks.dram_harvesting_mask);
    EXPECT_EQ(info.eth_harvesting_mask, soc.harvesting_masks.eth_harvesting_mask);
    EXPECT_EQ(info.l2cpu_harvesting_mask, soc.harvesting_masks.l2cpu_harvesting_mask);
    EXPECT_EQ(info.pcie_harvesting_mask, soc.harvesting_masks.pcie_harvesting_mask);
}
```
but there's also boiler-plate, so consider adding this:
```cpp
class SimulationHostTest : public ::testing::Test {
protected:
    void SetUp() override {
        const char* env = std::getenv("TT_UMD_SIMULATOR");
        if (env == nullptr) {
            GTEST_SKIP() << "TT_UMD_SIMULATOR is not set.";
        }
        simulator_path = env;

        // Probe that the socket is free, then release it so the test's discover() can bind it.
        socket = SimulationServerSocket::default_socket_path(0);
        auto probe = SimulationServerSocket::try_create(socket);
        if (probe == nullptr) {
            GTEST_SKIP() << "A live simulation host already holds " << socket << "; skipping.";
        }
    }

    std::string simulator_path;
    std::filesystem::path socket;
};
```
And use the `TEST_F` to reduce boiler-plate code.

nbuncicTT (approved):

device/api/umd/device/simulation/simulation_device_identity.hpp
pjanevskiTT: Renamed to `fetch_device_info_from_host`.

device/api/umd/device/simulation/simulation_device_identity.hpp
pjanevskiTT: Simplified -- the comment now just says it requests the host's device identity over the socket.

device/api/umd/device/simulation/simulation_device_identity.hpp
pjanevskiTT: Simplified the bridge and host-side comments along the lines you suggested, so they're less tied to the implementation.

device/api/umd/device/simulation/simulation_device_identity.hpp
pjanevskiTT: Simplified to "Client side: build a SocDescriptor from a wire message served by the host."

device/api/umd/device/simulation/simulation_server_protocol.hpp
pjanevskiTT: Holding off here: `Read`/`Write` are already merged (#2967) in PascalCase, so renaming only the newer values would make `SimulationServerCommand` inconsistent, and converting the whole enum to UPPER_SNAKE would touch already-merged code. I'd rather do a full-enum casing pass as its own cleanup than fold it into this PR -- leaving this for your call.

device/api/umd/device/simulation/simulation_server_protocol.hpp
pjanevskiTT: Holding off here: `Read`/`Write` are already merged (#2967) in PascalCase, so renaming only the newer values would make `SimulationServerCommand` inconsistent, and converting the whole enum to UPPER_SNAKE would touch already-merged code. I'd rather do a full-enum casing pass as its own cleanup than fold it into this PR -- leaving this for your call.

device/api/umd/device/tt_device/rtl_simulation_tt_device.hpp
pjanevskiTT: Renamed `SimulationBackendKind`->`SimulationBackendType` and `backend_kind`/`backend_kind()`->`backend_type`/`backend_type()` across the stack (incl. the `.fbs`).

device/simulation/simulation_device_identity.cpp
pjanevskiTT: Extracted into `write_temp_file()` returning a self-removing `ScopedTempFile` (anonymous namespace). That also fixes the leak you noted -- the file is now removed even if `SocArchDescriptor` construction or the arch check throws.

tests/api/test_simulation_server_protocol.cpp
pjanevskiTT: Leaving this as-is for now: the overloaded `encode` + named `decode_*` asymmetry is minor and switching touches every call site. Noting the preference -- happy to change if you feel strongly.

tests/api/test_simulation_server_protocol.cpp
pjanevskiTT: Renamed to `DeviceInfoSerializationRoundTrip`.

tests/api/test_simulation_connector.cpp
pjanevskiTT: Split the device-info assertions into a dedicated `HostServesDeviceInfoOverSocket` test that checks every field. Adapted the setup to the current per-server-directory API (`allocate_server_directory()`) rather than the `default_socket_path(0)` + probe in the snippet, since #3020/#3051 replaced that pattern. Left out the shared fixture to keep churn down.

PR #3020, opened 2026-07-14
Simulation: decide host/client role from simulator_path

pjanevskiTT:
### Issue
Part of #2677 (simulation server process). Until now a UMD process decided whether it was the simulation *host* or a *client* by a runtime socket bind-race (first to bind the per-chip socket wins). This PR makes the role a deterministic function of `simulator_path`. Stacked on #3019.

### Description
`SimulationConnector::discover()` now classifies `simulator_directory` and picks the role from what it points at — no bind-race:

- a `.so` file → **host** running the TTSim backend;
- a directory holding per-chip simulation sockets (`tt-umd-sim-<id>.sock`) → **client**: one device per socket in it, each attaching to a live host;
- any other directory → **host** running the RTL backend.

Client-vs-RTL is disambiguated by directory contents: a directory is a client target iff it contains AF_UNIX sockets matching the naming convention.

The **client path is multi-chip**: it enumerates the per-chip sockets and builds one client device per socket, with chip ids parsed from the socket names (so they match the host's). The connector performs the single `GET_DEVICE_INFO` fetch — it needs the served **backend kind** (from #3019) to instantiate the matching device class — and passes the fetched identity into `create_client()`, which no longer fetches it itself.

The **host path** claims its socket with `create()`: a live owner for the same chip is now a hard error rather than a silent demotion to client (role is decided by the path, so two hosts for one chip is a real conflict).

### List of the changes
- `SimulationConnector::discover()` — rewritten to a path classifier (`.so` / socket-dir / RTL-dir) with a multi-chip client path; host claims the socket via `create()`.
- `create_client()` (TTSim + RTL) — takes the connector-fetched `SimulationServerDeviceInfo` instead of fetching `GET_DEVICE_INFO` itself.
- `SimulationServerSocket::chip_id_from_socket_path()` — inverse of `default_socket_path()`'s naming.
- `SimulationServerSocket::sockets_in_directory()` — the per-chip sockets in a directory, keyed by chip id (AF_UNIX + name-matched).
- Review hardening: `chip_id_from_socket_path` parses via `std::from_chars`/`std::string_view` and rejects leading-zero names so `-0`/`-00` can't collide; `classify()` returns the sockets it enumerated so `discover()` doesn't re-scan (closing a TOCTOU that would otherwise return zero devices silently) and skips a dead socket per-chip with a warning instead of aborting the whole attach; the internal path-kind enum uses UPPER_SNAKE.
- Tests: `chip_id_from_socket_path` + `sockets_in_directory` unit-tested in the no-card baremetal lane; e2e connector tests updated to the path-based model (client discovered by pointing at the socket directory; obsolete "live-socket ⇒ client" test replaced with a path-classification test).

### Testing
Build + pre-commit clean; no-card lanes pass (179 baremetal, incl. the new socket-naming/enumeration tests; api protocol + identity + the path-classification test). The e2e connector tests run in the simulation CI lane (`TT_UMD_SIMULATOR`).

### API Changes
`TTSimTTDevice::create_client` / `RtlSimulationTTDevice::create_client` take an extra `const SimulationServerDeviceInfo&` (the connector fetches device identity once and passes it in). `SimulationServerSocket` gains `chip_id_from_socket_path()` and `sockets_in_directory()`. All internal to the simulation subsystem.

### DIFF
--- a/device/simulation/simulation_connector.cpp
+++ b/device/simulation/simulation_connector.cpp
@@ -4,45 +4,70 @@
 
 #include "umd/device/simulation/simulation_connector.hpp"
 
+#include <fmt/format.h>
+
+#include <filesystem>
+#include <map>
 #include <memory>
+#include <system_error>
+#include <tt-logger/tt-logger.hpp>
 #include <utility>
 
 #include "simulation/simulation_server_socket.hpp"
 #include "umd/device/simulation/simulation_client.hpp"
+#include "umd/device/simulation/simulation_device_identity.hpp"
+#include "umd/device/simulation/simulation_server_protocol.hpp"
 #include "umd/device/tt_device/rtl_simulation_tt_device.hpp"
 #include "umd/device/tt_device/tt_sim_tt_device.hpp"
+#include "umd/device/utils/error.hpp"
 
 namespace tt::umd {
 
 namespace {
 
-// Builds the device for one simulated chip. The backend (TTSim .so vs RTL directory) is chosen by
-// the simulator path; the host-vs-client role is decided by who owns the socket:
-//   - socket == nullptr: a live host already serves this chip, so attach as a client that reaches
-//     it over the socket via SimulationClient (slim today -- just connect/disconnect; device-op
-//     forwarding lands as the API grows). The client runs no backend, so the only backend-specific
-//     bit is which device class describes it.
-//   - socket != nullptr: we are the host; bring up the in-process backend (the direct hot path) and
-//     hand it the socket to own and expose.
-// The two backends don't share a common factory return type that carries adopt_socket, so dispatch
-// on the concrete device here.
-std::unique_ptr<TTDevice> create_simulation_device(
-    const std::filesystem::path& simulator_directory,
-    ChipId chip_id,
-    int num_host_mem_channels,
-    const std::filesystem::path& socket_path,
-    std::unique_ptr<SimulationServerSocket> socket) {
-    const bool is_ttsim = simulator_directory.extension() == ".so";
+// The role a UMD process takes for a given simulator_path -- decided purely from what the path is,
+// with no socket bind-race:
+//   - a ".so" file                 -> host running the TTSim backend for that library;
+//   - a directory holding per-chip  -> client: a host already serves there, so attach to each
+//     simulation sockets               socket in it (one device per socket), sourcing device
+//                                      identity from the host over the wire;
+//   - any other directory          -> host running the RTL backend from that build directory.
+enum class Role { HOST_TTSIM, HOST_RTL, CLIENT };
 
-    if (socket == nullptr) {
-        auto client = std::make_unique<SimulationClient>(socket_path);
-        if (is_ttsim) {
-            return TTSimTTDevice::create_client(chip_id, std::move(client));
-        }
-        return RtlSimulationTTDevice::create_client(chip_id, std::move(client));
+// The role and, for CLIENT, the sockets found in the directory -- returned together so discover()
+// doesn't scan the directory a second time. The second scan would also be a TOCTOU race: if the host
+// exited between the two scans, the role would still be CLIENT but discover() would build zero
+// devices from a now-empty listing and return silently.
+struct Classification {
+    Role role;
+    std::map<ChipId, std::filesystem::path> sockets;
+};
+
+Classification classify(const std::filesystem::path& simulator_path) {
+    std::error_code ec;
+    if (std::filesystem::is_regular_file(simulator_path, ec) && simulator_path.extension() == ".so") {
+        return {Role::HOST_TTSIM, {}};
     }
+    UMD_ASSERT(
+        std::filesystem::is_directory(simulator_path, ec),
+        error::RuntimeError,
+        fmt::format("Simulator path is neither a .so file nor a directory: {}", simulator_path.string()));
+    // A directory that holds per-chip simulation sockets means a host is already serving there, so
+    // attach as a client; any other directory is an RTL build we host.
+    auto sockets = SimulationServerSocket::sockets_in_directory(simulator_path);
+    if (sockets.empty()) {
+        return {Role::HOST_RTL, {}};
+    }
+    return {Role::CLIENT, std::move(sockets)};
+}
 
-    if (is_ttsim) {
+// Host path: bring up the in-process backend (the direct hot path) and hand it the socket to serve.
+std::unique_ptr<TTDevice> make_host_device(
+    Role role,
+    const std::filesystem::path& simulator_directory,
+    int num_host_mem_channels,
+    std::unique_ptr<SimulationServerSocket> socket) {
+    if (role == Role::HOST_TTSIM) {
         auto device = TTSimTTDevice::create(simulator_directory, num_host_mem_channels);
         device->adopt_socket(std::move(socket));
         return device;
@@ -52,20 +77,66 @@ std::unique_ptr<TTDevice> create_simulation_device(
     return device;
 }
 
+// Client path: build the device class the host reports over the wire. A client runs no local
+// backend, so the class mostly just names the device; the backend kind still comes from the host so
+// the right class is instantiated (and so this stays correct if the classes diverge).
+std::unique_ptr<TTDevice> make_client_device(
+    ChipId chip_id, std::unique_ptr<SimulationClient> client, const SimulationServerDeviceInfo& info) {
+    switch (info.backend_type) {
+        case SimulationBackendType::TTSim:
+            return TTSimTTDevice::create_client(chip_id, std::move(client), info);
+        case SimulationBackendType::Rtl:
+            return RtlSimulationTTDevice::create_client(chip_id, std::move(client), info);
+    }
+    // A value outside the enum means a corrupt/incompatible reply from the host; fail loudly rather
+    // than silently defaulting to a backend class.
+    UMD_THROW(
+        error::RuntimeError,
+        fmt::format("Host reported an unknown simulation backend kind ({})", static_cast<int>(info.backend_type)));
+}
+
 }  // namespace
 
 std::map<ChipId, std::unique_ptr<TTDevice>> SimulationConnector::discover(const SimulationConnectorOptions& options) {
     std::map<ChipId, std::unique_ptr<TTDevice>> devices;
+    const std::filesystem::path& simulator_path = options.simulator_directory;
 
-    // Single-chip for now. Socket-first: try to claim the path; success => we are the host.
+    const Classification classification = classify(simulator_path);
+
+    if (classification.role == Role::CLIENT) {
+        // Multi-chip: one client device per per-chip socket in the directory (enumerated by
+        // classify()). Chip ids come from the socket names, so they match the host's. A failure on
+        // one socket (a dead or wedged host) only skips that chip -- it must not abort attaching to
+        // the healthy ones.
+        for (const auto& [chip_id, socket_file] : classification.sockets) {
+            try {
+                auto client = std::make_unique<SimulationClient>(socket_file);
+                // The connector owns the single GET_DEVICE_INFO fetch: it needs the backend type to
+                // pick the device class, and passes the fetched identity into create_client so it is
+                // not fetched again.
+                const SimulationServerDeviceInfo info = fetch_device_info_from_host(*client);
+                devices.emplace(chip_id, make_client_device(chip_id, std::move(client), info));
+            } catch (const std::exception& e) {
+                log_warning(
+                    LogUMD, "Skipping simulation socket {} (chip {}): {}", socket_file.string(), chip_id, e.what());
+            }
+        }
+        // Every socket was present but none answered -- surface it rather than returning an empty,
+        // successful-looking cluster.
+        UMD_ASSERT(
+            !devices.empty(),
+            error::RuntimeError,
+            fmt::format("No reachable simulation hosts among the sockets in {}", simulator_path.string()));
+        return devices;
+    }
+
+    // Host: single chip for now. Claim the per-chip socket -- create() throws if a live host
+    // already owns it (two hosts cannot serve the same chip) -- and serve it.
     const ChipId chip_id = 0;
-    const std::filesystem::path socket_path = SimulationServerSocket::default_socket_path(chip_id);
-    std::unique_ptr<SimulationServerSocket> socket = SimulationServerSocket::try_create(socket_path);
-
+    auto socket = SimulationServerSocket::create(SimulationServerSocket::default_socket_path(chip_id));
     devices.emplace(
         chip_id,
-        create_simulation_device(
-            options.simulator_directory, chip_id, options.num_host_mem_channels, socket_path, std::move(socket)));
+        make_host_device(classification.role, simulator_path, options.num_host_mem_channels, std::move(socket)));
     return devices;
 }
 
--- a/device/simulation/simulation_server_socket.cpp
+++ b/device/simulation/simulation_server_socket.cpp
@@ -9,8 +9,12 @@
 
 #include <asio.hpp>
 #include <atomic>
+#include <charconv>
+#include <map>
 #include <memory>
 #include <mutex>
+#include <string>
+#include <string_view>
 #include <system_error>
 #include <thread>
 #include <vector>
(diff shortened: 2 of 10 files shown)


device/simulation/simulation_connector.cpp
```diff
@@ -52,20 +63,44 @@ std::unique_ptr<TTDevice> create_simulation_device(
...
+// Client path: build the device class the host reports over the wire. A client runs no local
+// backend, so the class mostly just names the device; the backend kind still comes from the host so
+// the right class is instantiated (and so this stays correct if the classes diverge).
+std::unique_ptr<TTDevice> make_client_device(
+    ChipId chip_id, std::unique_ptr<SimulationClient> client, const SimulationServerDeviceInfo& info) {
+    if (info.backend_kind == SimulationBackendKind::TTSim) {
+        return TTSimTTDevice::create_client(chip_id, std::move(client), info);
+    }
+    return RtlSimulationTTDevice::create_client(chip_id, std::move(client), info);
+}
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed: make_client_device now switches on backend_kind and throws on an unknown value instead of defaulting to RTL.

device/simulation/simulation_server_socket.cpp
```diff
@@ -306,4 +309,45 @@ std::filesystem::path SimulationServerSocket::default_socket_path(ChipId chip_id
...
+        std::error_code sock_ec;
+        if (!entry.is_socket(sock_ec)) {
+            continue;
+        }
+        if (const std::optional<ChipId> chip_id = chip_id_from_socket_path(entry.path())) {
+            sockets.emplace(*chip_id, entry.path());
+        }
+    }
+    return sockets;
+}
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed: the directory_iterator error_code is now checked; a non-directory or unreadable path returns empty (the caller then classifies it as a host build) rather than silently proceeding.

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -75,13 +75,13 @@ std::unique_ptr<TTSimTTDevice> TTSimTTDevice::create_for_chip(
...
     UMD_ASSERT(
         client != nullptr, error::RuntimeError, "Client-mode TTSimTTDevice requires a non-null SimulationClient.");
-    // Source the device identity from the host over the socket -- a client does not read a local
-    // simulator build.
-    const SimulationServerDeviceInfo info = fetch_device_info(*client);
-    SocDescriptor soc_descriptor = build_soc_descriptor(info);
+    // The connector already fetched the device identity over the socket (it needed the backend kind
+    // to pick this class); build the SoC descriptor from it -- a client does not read a local build.
+    SocDescriptor soc_descriptor = build_soc_descriptor(device_info);
     // make_unique can't reach the private client-mode constructor; this static factory can via new.
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed: create_client now asserts device_info.status == 0 before building the device.

device/tt_device/rtl_simulation_tt_device.cpp
```diff
@@ -67,15 +67,14 @@ std::unique_ptr<RtlSimulationTTDevice> RtlSimulationTTDevice::create(
...
         error::RuntimeError,
         "Client-mode RtlSimulationTTDevice requires a non-null SimulationClient.");
-    // Source the device identity from the host over the socket -- a client does not read a local
-    // simulator build.
-    const SimulationServerDeviceInfo info = fetch_device_info(*client);
-    SocDescriptor soc_descriptor = build_soc_descriptor(info);
+    // The connector already fetched the device identity over the socket (it needed the backend kind
+    // to pick this class); build the SoC descriptor from it -- a client does not read a local build.
+    SocDescriptor soc_descriptor = build_soc_descriptor(device_info);
     // make_unique can't reach the private client-mode constructor; this static factory can via new.
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed: create_client now asserts device_info.status == 0 before building the device.

device/simulation/simulation_server_socket.cpp
```diff
@@ -334,4 +337,49 @@ std::filesystem::path SimulationServerSocket::default_socket_path(ChipId chip_id
...
+        name.compare(name.size() - suffix.size(), suffix.size(), suffix) != 0) {
+        return std::nullopt;
+    }
+    const std::string digits = name.substr(prefix.size(), name.size() - prefix.size() - suffix.size());
+    // Require a plain run of digits so e.g. "tt-umd-sim-.sock" or "tt-umd-sim-x.sock" don't parse.
+    if (digits.empty() || digits.find_first_not_of("0123456789") != std::string::npos) {
+        return std::nullopt;
+    }
+    try {
+        return static_cast<ChipId>(std::stoi(digits));
```
nbuncicTT: consider using `std::from_chars` instead of `stoi` and parsing the error code and then returning `nullopt`, because the `from_chars` doesn't throw.

device/simulation/simulation_server_socket.cpp
```diff
@@ -334,4 +337,49 @@ std::filesystem::path SimulationServerSocket::default_socket_path(ChipId chip_id
...
+    // Inverse of default_socket_path()'s "tt-umd-sim-<chip_id>.sock". Kept next to it so the naming
+    // convention lives in exactly one place.
+    static constexpr std::string_view prefix = "tt-umd-sim-";
+    static constexpr std::string_view suffix = ".sock";
+    const std::string name = socket_path.filename().string();
+    if (name.size() <= prefix.size() + suffix.size() || name.compare(0, prefix.size(), prefix) != 0 ||
+        name.compare(name.size() - suffix.size(), suffix.size(), suffix) != 0) {
+        return std::nullopt;
+    }
+    const std::string digits = name.substr(prefix.size(), name.size() - prefix.size() - suffix.size());
```
nbuncicTT: Consider using std::string_view instead of the string. Something like this:
Here is how to eliminate the dynamic allocation by switching to `std::string_view` and `std::from_chars`.
```cpp
if (name.size() >= prefix.size() + suffix.size()) {
    std::string_view digits_view = std::string_view(name).substr(
        prefix.size(), 
        name.size() - prefix.size() - suffix.size()
    );
    
    int number = 0;
    auto [ptr, ec] = std::from_chars(digits_view.data(), digits_view.data() + digits_view.size(), number);
    
    if (ec == std::errc{}) {
        // Use 'number'
    }
}
```

device/simulation/simulation_server_socket.cpp
```diff
@@ -334,4 +337,49 @@ std::filesystem::path SimulationServerSocket::default_socket_path(ChipId chip_id
     return std::filesystem::temp_directory_path() / fmt::format("tt-umd-sim-{}.sock", chip_id);
 }
 
+std::optional<ChipId> SimulationServerSocket::chip_id_from_socket_path(const std::filesystem::path& socket_path) {
+    // Inverse of default_socket_path()'s "tt-umd-sim-<chip_id>.sock". Kept next to it so the naming
+    // convention lives in exactly one place.
+    static constexpr std::string_view prefix = "tt-umd-sim-";
+    static constexpr std::string_view suffix = ".sock";
+    const std::string name = socket_path.filename().string();
+    if (name.size() <= prefix.size() + suffix.size() || name.compare(0, prefix.size(), prefix) != 0 ||
```
nbuncicTT: Use `starts_with`/`ends_with` instead of hand-rolling these strings.

device/simulation/simulation_server_socket.cpp
```diff
@@ -334,4 +337,49 @@ std::filesystem::path SimulationServerSocket::default_socket_path(ChipId chip_id
     return std::filesystem::temp_directory_path() / fmt::format("tt-umd-sim-{}.sock", chip_id);
 }
 
+std::optional<ChipId> SimulationServerSocket::chip_id_from_socket_path(const std::filesystem::path& socket_path) {
+    // Inverse of default_socket_path()'s "tt-umd-sim-<chip_id>.sock". Kept next to it so the naming
+    // convention lives in exactly one place.
+    static constexpr std::string_view prefix = "tt-umd-sim-";
```
nbuncicTT: nit: static is redundant here, but it's fine if you leave it in.

nbuncicTT: also in the var below

device/simulation/simulation_connector.cpp
```diff
@@ -4,45 +4,56 @@
...
+        error::RuntimeError,
+        fmt::format("Simulator path is neither a .so file nor a directory: {}", simulator_path.string()));
+    // A directory that holds per-chip simulation sockets means a host is already serving there, so
+    // attach as a client; any other directory is an RTL build we host.
+    return SimulationServerSocket::sockets_in_directory(simulator_path).empty() ? Role::HostRtl : Role::Client;
+}
 
-    if (is_ttsim) {
+// Host path: bring up the in-process backend (the direct hot path) and hand it the socket to serve.
+std::unique_ptr<TTDevice> make_host_device(
```
nbuncicTT, line 65: Like I said in some of the previous PRs, we are conflating backends (TTSim/RtlSim) with host/client roles. We need to fix this, but as we agreed after all of the PRs are in we will have a refactoring session on this.

device/simulation/simulation_connector.cpp
```diff
@@ -4,45 +4,56 @@
...
+    if (std::filesystem::is_regular_file(simulator_path, ec) && simulator_path.extension() == ".so") {
+        return Role::HostTTSim;
     }
+    UMD_ASSERT(
+        std::filesystem::is_directory(simulator_path, ec),
+        error::RuntimeError,
+        fmt::format("Simulator path is neither a .so file nor a directory: {}", simulator_path.string()));
+    // A directory that holds per-chip simulation sockets means a host is already serving there, so
+    // attach as a client; any other directory is an RTL build we host.
+    return SimulationServerSocket::sockets_in_directory(simulator_path).empty() ? Role::HostRtl : Role::Client;
```
nbuncicTT: `sockets_in_directory()` is called twice on the same path, once here for `.empty()`, and again in `discover()` to build the devices, so the enumerated map is scanned, discarded, and rebuilt.

Beyond the redundant syscalls, the two scans are separate observations of a mutable directory: if the host exits between them, `role` is still `Client`, the loop body never runs, and `discover()` returns an empty map with no error. Silently getting zero devices is a worse outcome than throwing.

Consider having `classify()` return the sockets alongside the role (e.g. a small `{Role role; std::map<ChipId, std::filesystem::path> sockets;}`) so `discover()` iterates the exact set the decision was made on. The filesystem race doesn't disappear, but a socket that dies after the scan then fails loudly in `fetch_device_info_from_host()` instead of silently shrinking the result to nothing — and `Role::Client` becomes guaranteed to carry at least one socket.

device/simulation/simulation_server_socket.hpp
```diff
@@ -73,6 +75,18 @@ class SimulationServerSocket {
...
+    // The per-chip simulation sockets present in a directory: {chip_id -> socket file}, keyed and
+    // ordered by chip id. Only AF_UNIX sockets whose names match the default_socket_path()
+    // convention are included; everything else in the directory is ignored. Empty if the path is
+    // not a directory or holds no such sockets. This is how a client turns a socket directory into
+    // the set of hosts to attach to.
+    static std::map<ChipId, std::filesystem::path> sockets_in_directory(const std::filesystem::path& directory);
+
 private:
     // Non-throwing; only initializes members. Binding happens in bind_and_listen(), driven by
     // the try_create()/create() factories.
```
nbuncicTT, line 92: Now that the role is decided from the path, nothing consumes `try_create()`'s nullptr return anymore (only `create()` at :82 and tests call it), so it should drop out of the public header into a private implementation detail of `create()` with the tests given a proper static `is_live(path)` probe instead of binding-and-destructing to test liveness.

device/simulation/simulation_connector.cpp
```diff
@@ -4,45 +4,56 @@
...
-    std::unique_ptr<SimulationServerSocket> socket) {
-    const bool is_ttsim = simulator_directory.extension() == ".so";
+// The role a UMD process takes for a given simulator_path -- decided purely from what the path is,
+// with no socket bind-race:
+//   - a ".so" file                 -> host running the TTSim backend for that library;
+//   - a directory holding per-chip  -> client: a host already serves there, so attach to each
+//     simulation sockets               socket in it (one device per socket), sourcing device
+//                                      identity from the host over the wire;
+//   - any other directory          -> host running the RTL backend from that build directory.
+enum class Role { HostTTSim, HostRtl, Client };
```
nbuncicTT: nit: Use upper snake case in enums.

device/simulation/simulation_connector.cpp
```diff
@@ -52,20 +63,51 @@ std::unique_ptr<TTDevice> create_simulation_device(
...
+    if (role == Role::Client) {
+        // Multi-chip: one client device per per-chip socket in the directory. Chip ids come from
+        // the socket names, so they match the host's.
+        for (const auto& [chip_id, socket_file] : SimulationServerSocket::sockets_in_directory(simulator_path)) {
+            auto client = std::make_unique<SimulationClient>(socket_file);
+            // The connector owns the single GET_DEVICE_INFO fetch: it needs the backend kind to
+            // pick the device class, and passes the fetched identity into create_client so it is
+            // not fetched again.
+            const SimulationServerDeviceInfo info = fetch_device_info_from_host(*client);
+            devices.emplace(chip_id, make_client_device(chip_id, std::move(client), info));
```
nbuncicTT: `tt-umd-sim-0.sock` and `tt-umd-sim-00.sock` both parse to chip 0 and collide, with emplace silently keeping whichever the directory iterator yielded first. Can you fix this?

nbuncicTT (commented):
Will add more comments

device/simulation/simulation_server_socket.cpp
```diff
@@ -334,4 +337,49 @@ std::filesystem::path SimulationServerSocket::default_socket_path(ChipId chip_id
...
+    // either case there are no per-chip sockets to report (the caller then classifies it as a host
+    // build, not a client socket directory), so return empty rather than silently proceeding.
+    std::error_code ec;
+    std::filesystem::directory_iterator it(directory, ec);
+    if (ec) {
+        return sockets;
+    }
+    for (const auto& entry : it) {
+        std::error_code sock_ec;
+        if (!entry.is_socket(sock_ec)) {
```
nbuncicTT, line 380: `is_socket()` can't tell a live host from a stale file left by a crashed one, so a single leftover socket permanently classifies the directory as a client target and then hard-fails on connect — this silently drops the stale-socket reclamation `try_create()` used to do, and the new baremetal test literally seeds the directory with `leave_stale_socket()` to prove the enumeration accepts them.

device/simulation/simulation_connector.cpp
```diff
@@ -52,20 +63,51 @@ std::unique_ptr<TTDevice> create_simulation_device(
...
+        }
+        return devices;
+    }
+
+    // Host: single chip for now. Claim the per-chip socket -- create() throws if a live host
+    // already owns it (two hosts cannot serve the same chip) -- and serve it.
+    const ChipId chip_id = 0;
+    auto socket = SimulationServerSocket::create(SimulationServerSocket::default_socket_path(chip_id));
```
nbuncicTT, line 136: The socket is bound here but doesn't start accepting until `adopt_socket()` calls `serve()`, which happens after the backend is up. A client connecting in that window gets a successful `connect()` (the listen backlog accepts it) and then blocks forever in `transact()`, since there's no timeout anywhere on the client path. With a slow `.so` load this is a plausible hang rather than a theoretical one — either publish the socket only once serving has started, or add a timeout in `SimulationClient`.

device/simulation/simulation_connector.cpp
```diff
@@ -52,20 +63,51 @@ std::unique_ptr<TTDevice> create_simulation_device(
...
-            options.simulator_directory, chip_id, options.num_host_mem_channels, socket_path, std::move(socket)));
+    if (role == Role::Client) {
+        // Multi-chip: one client device per per-chip socket in the directory. Chip ids come from
+        // the socket names, so they match the host's.
+        for (const auto& [chip_id, socket_file] : SimulationServerSocket::sockets_in_directory(simulator_path)) {
+            auto client = std::make_unique<SimulationClient>(socket_file);
+            // The connector owns the single GET_DEVICE_INFO fetch: it needs the backend kind to
+            // pick the device class, and passes the fetched identity into create_client so it is
+            // not fetched again.
+            const SimulationServerDeviceInfo info = fetch_device_info_from_host(*client);
```
nbuncicTT: A throw from `fetch_device_info_from_host()` on any one socket aborts the whole discovery, so a single dead or wedged host takes down an otherwise healthy multi-chip attach. Worth catching per-socket and either skipping that chip with a warning or failing with a message that names which socket was responsible.

nbuncicTT (approved):

device/simulation/simulation_server_socket.cpp
pjanevskiTT: Switched to `std::from_chars` (non-throwing); it must consume every digit or the name is rejected.

device/simulation/simulation_server_socket.cpp
pjanevskiTT: The parse now runs over a `std::string_view`, so the `std::string` allocation is gone.

device/simulation/simulation_server_socket.cpp
pjanevskiTT: `starts_with`/`ends_with` are C++20 and this repo builds as C++17, so they don't compile here. The allocation they'd have removed is gone anyway -- the prefix/suffix checks now run over a `std::string_view`.

device/simulation/simulation_server_socket.cpp
pjanevskiTT: Dropped the redundant `static` from both.

device/simulation/simulation_connector.cpp
pjanevskiTT: Agreed -- leaving the backend-vs-role separation for the refactoring session after the stack lands, as we discussed.

device/simulation/simulation_connector.cpp
pjanevskiTT: `classify()` now returns the sockets alongside the role, so `discover()` reuses them instead of re-scanning. And when the listing is non-empty but nothing answers, discover() now throws instead of returning an empty, successful-looking cluster -- the TOCTOU silent-zero you flagged.

device/simulation/simulation_connector.cpp
pjanevskiTT: Renamed to UPPER_SNAKE: `HOST_TTSIM` / `HOST_RTL` / `CLIENT`.

device/simulation/simulation_connector.cpp
pjanevskiTT: Fixed -- `chip_id_from_socket_path` now rejects a leading zero (except the lone `0`), so `tt-umd-sim-0` and `tt-umd-sim-00` can't both map to chip 0. Added a regression test.

device/simulation/simulation_connector.cpp
pjanevskiTT: Wrapped each per-socket attach in try/catch: a dead or wedged host is skipped with a warning naming the socket, and discover() only fails if none of them attach.

device/simulation/simulation_server_socket.hpp
pjanevskiTT: Deferring to the refactoring session after the stack lands (tracked alongside the backends-vs-roles cleanup). Left open as a tracker.

device/simulation/simulation_server_socket.cpp
pjanevskiTT: Deferring to the same post-stack refactoring session: the fix is a liveness probe during classification, which is tied to the try_create()/is_live() change above and shifts the stale-socket test expectations, so it's cleaner to do together. Left open as a tracker.

device/simulation/simulation_connector.cpp
pjanevskiTT: Tracking as a separate follow-up PR: the fix is a connect/read timeout in SimulationClient, which is outside this PR's diff. Left open until that lands.

device/simulation/simulation_connector.cpp
pjanevskiTT: Tracked in #3081.

device/simulation/simulation_server_socket.hpp
pjanevskiTT: Tracked in #3081.

device/simulation/simulation_server_socket.cpp
pjanevskiTT: Tracked in #3081.

device/simulation/simulation_connector.cpp
pjanevskiTT: Fixed in #3084: `SimulationClient` now takes a timeout (default `SIMULATION_SOCKET_TIMEOUT` = 30s) enforced with a `poll()` before each send/recv in `transact()`, so a client that connects to a bound-but-not-serving host throws instead of hanging. (poll() rather than SO_RCVTIMEO, which asio's sync read masks.) That PR is on `main`, independent of this stack.

PR #3025, opened 2026-07-14
Simulation: route the Cluster SIMULATION path through SimulationConnector

pjanevskiTT:
### Issue
Part of #2677 (simulation server process). Wires the host/client role decision (built in #3020) up to the `Cluster` level, so opening a simulation `Cluster` becomes a host or a client depending on `simulator_directory`. Stacked on the cluster-descriptor-over-socket PR.

### Description
`Cluster(ChipType::SIMULATION)` now branches on `SimulationConnector::role_for(simulator_directory)`:

- **Host** (`.so` file / RTL directory): keeps the existing multichip in-process device creation, and **additionally exposes each simulation chip's device on its per-chip socket** (`adopt_socket`) so a separate client process can attach and drive it. Multichip host behaviour is unchanged — this only adds socket serving; each device serves device-memory I/O plus `GET_DEVICE_INFO` / `GET_CLUSTER_DESCRIPTOR`.
- **Client** (a directory of `tt-umd-sim-<id>.sock` sockets): reconstructs the `ClusterDescriptor` over the socket (`GET_CLUSTER_DESCRIPTOR`; empty ⇒ a mock built from the served device info), then `SimulationConnector::discover()` attaches one client device per socket. The chip loop wraps each connector-created device in a `SimulationChip` instead of building a backend.

Ownership seam: the connector owns simulation device creation + `SocDescriptor`; `Cluster` owns the `ClusterDescriptor` and `Chip` wrapping.

**Scope:** exercised single-chip. A multichip host (one process publishing N sockets with remote-over-sim-eth routing) is deferred and not built here; the client reconstruction already handles N chips, and a reconciliation guard throws if the served topology names a chip with no socket.

### List of the changes
- `SimulationConnector` — public `Role { Host, Client }` + `role_for()`, so `Cluster` branches without duplicating the path classifier (internal enum renamed to `PathKind`).
- `cluster.{hpp,cpp}` — SIMULATION path: client reconstructs topology over the socket + attaches via the connector; host exposes each chip on its socket via an extracted `Cluster::serve_simulation_devices_over_sockets()` method; `construct_chip_from_cluster` wraps a connector-provided device.
- Tests: CI-gated `HostAndClientClustersShareDeviceMemory` (host Cluster from `.so`, client Cluster from the socket directory, cluster-level write/read across the socket).

### Testing
Build + pre-commit clean; 182 baremetal + all sim unit suites pass. The Cluster-level e2e and the other connector e2e tests run in the simulation CI lane (`TT_UMD_SIMULATOR`).

### API Changes
`SimulationConnector` gains `Role` + `role_for()`. Simulation-internal otherwise.

### Rebased onto #2955 (multichip ttsim)
#2955 is now in `main` and this PR has been rebased onto it. Both edit the SIMULATION block of the chip-construction path; resolved by the ordering inside the block (client wrap first, then #2955's host remote/gateway logic):
```cpp
if (tt_device != nullptr) { /* this PR: client — wrap the connector-created device */ }
if (simulator_directory.extension() == ".so" && cluster_desc->is_chip_remote(chip_id))
    return create_simulation_remote_chip(...);   // #2955: host remote chip
return SimulationChip::create(...);              // host gateway
```
For the host `tt_device` is null, so #2955's remote/gateway logic runs unchanged; for the client `tt_device` is non-null and gets wrapped. `add_chip` is edited only by #2955, so its version applies as-is — no conflict there.

Behavior note (no code change needed for single-chip): post-#2955 the host's remote chips are `RemoteChip`s backed by a remote `TTDevice`, not a `SimulationTTDevice`, so the adopt-socket post-pass here `dynamic_cast`s them away and only the **gateway** serves a socket. Exposing remote chips over sockets so a multichip *client* can fully attach is the deferred multichip-over-sockets work, and #2955's gateway+remotes model is what should define how it's done — best built after #2955 lands, not guessed here.

### DIFF
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -28,6 +28,7 @@
 #include <vector>
 
 #include "hugepage.hpp"
+#include "simulation/simulation_server_socket.hpp"
 #include "tracy.hpp"
 #include "umd/device/chip/local_chip.hpp"
 #include "umd/device/chip/mock_chip.hpp"
@@ -48,6 +49,9 @@
 #include "umd/device/cluster_descriptor.hpp"
 #include "umd/device/pcie/pci_device.hpp"
 #include "umd/device/simulation/simulation_chip.hpp"
+#include "umd/device/simulation/simulation_client.hpp"
+#include "umd/device/simulation/simulation_connector.hpp"
+#include "umd/device/simulation/simulation_device_identity.hpp"
 #include "umd/device/soc_descriptor.hpp"
 #include "umd/device/topology/topology_discovery.hpp"
 #include "umd/device/topology/topology_discovery_options.hpp"
@@ -249,6 +253,12 @@ std::unique_ptr<Chip> Cluster::construct_chip_from_cluster(
     }
     if (chip_type == ChipType::SIMULATION) {
 #ifdef TT_UMD_BUILD_SIMULATION
+        if (tt_device != nullptr) {
+            // A device the connector already created (client mode): wrap it rather than building a
+            // new backend. Its SoC descriptor was sourced over the socket, so read it back.
+            SocDescriptor device_soc = tt_device->get_soc_descriptor();
+            return std::make_unique<SimulationChip>(simulator_directory, device_soc, chip_id, std::move(tt_device));
+        }
         if (simulator_directory.extension() == ".so" && cluster_desc->is_chip_remote(chip_id)) {
             return create_simulation_remote_chip(chip_id, cluster_desc, soc_desc);
         }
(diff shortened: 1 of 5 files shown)


device/cluster.cpp
```diff
@@ -567,6 +633,24 @@ Cluster::Cluster(ClusterOptions options) {
     }
 #endif  // TT_UMD_BUILD_SIMULATION
 
+#ifdef TT_UMD_BUILD_SIMULATION
```
aleksamarkovicTT, line 636: This should be in a separate method.

pjanevskiTT: Extracted into `Cluster::serve_simulation_devices_over_sockets()` (sim-gated private method); the constructor now just calls it.

aleksamarkovicTT (approved):

PR #3027, opened 2026-07-15
Add opt-in per-op timing instrumentation to device MMIO memcpy

### LINKED ISSUE #2915: Add opt-in per-op timing instrumentation to device MMIO memcpy
**Description**

When investigating slow or stalling MMIO traffic over a TLB window, there is currently no way to see how long the individual reads/writes inside `memcpy_to_device` / `memcpy_from_device` actually took. We want an opt-in instrumentation mode that records a `steady_clock` delta for each TLB-touching store/load and prints a per-call summary to stderr, so anyone interested in the data can see exactly how long all the reads/writes took.

This sits alongside the per-op MMIO timeout work in #2629 — same TLB-touching transfer paths in `device/pcie/device_memcpy.cpp`, but purely diagnostic: it observes and prints, it never aborts a transfer.

**Scope**

- Add an opt-in env var (`TT_UMD_MEMCPY_TIMING`) read once at process start; zero overhead when disabled.
- Instrument each TLB-touching op: once per 256-byte block in the bulk AVX2 phase, once per op in the 32 / 16 / 4-byte and byte-wide tail phases. Host-side loads/stores (no TLB access) are not instrumented.
- On function exit, dump to stderr: function name, total `size`, op count, `min` / `max` / `mean` / `total` in ns, and the full per-op `ns:bytes` list.
- Keep recording cheap — a thread-local buffer and a single thread-local check per op when enabled.

**Example output**

```
[memcpy_to_device] size=1024 ops=4 min=820ns max=1310ns mean=1002ns total=4011ns ops_ns:bytes=1310:256,920:256,820:256,961:256
```

**Is your feature request related to a problem?**

Diagnosing slow post-reset reads and posted-write pile-ups currently requires guesswork or external tracing. Surfacing per-op timings directly from the memcpy path makes the data available to anyone debugging MMIO throughput/latency.

**Definition of Done**

- `TT_UMD_MEMCPY_TIMING` toggles the instrumentation; default-off path has no measurable overhead.
- Both `memcpy_to_device` and `memcpy_from_device` emit a per-call timing summary when enabled.
- Per-op deltas and the min/max/mean/total aggregate are printed to stderr.

ajovanovicTT:
### Issue
#2915

### Description
Opt-in, diagnostic-only timing for MMIO transfers. When TT_UMD_MEMCPY_TIMING is set, each TLB-touching store/load in memcpy_to_device and memcpy_from_device is timed and a per-call summary is printed to stderr. It only observes, never alters a transfer. Zero overhead when disabled.

### List of the changes
 - Added env var TT_UMD_MEMCPY_TIMING, read once at startup, to toggle the instrumentation.
 - Instrumented each TLB-touching op in memcpy_to_device and memcpy_from_device, host-side transfers are not timed.
 - On function exit, print one stderr line: function name, size, op count, min, max, mean, total ns, and the per-op ns to bytes list. Recorded into a reused thread-local buffer, no allocation on the default path.

### Testing
 - CI
 - Manual runs on n300 with added env var set to ON, and checking if reported memcpy's match the test behavior and report proper sizes

### API Changes
This PR has no API changes.


ajovanovicTT:
Checked sizes on run of TestDeviceIOFixture.SimpleIOAllTargets/TENSIX:

Screen Shot 2026-07-15 at 14 25 15 PM

device/pcie/device_memcpy.cpp
```diff
@@ -39,6 +45,108 @@ auto make_mmio_timeout_guard(
...
+    }
+
+    MemcpyTimingRecorder(const MemcpyTimingRecorder&) = delete;
+    MemcpyTimingRecorder& operator=(const MemcpyTimingRecorder&) = delete;
+
+    ~MemcpyTimingRecorder() {
+        if (MEMCPY_TIMING_ENABLED) {
+            dump();
+        }
+    }
```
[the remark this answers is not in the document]
ajovanovicTT: Resolved with new commit.

device/pcie/device_memcpy.cpp
```diff
@@ -39,6 +45,108 @@ auto make_mmio_timeout_guard(
...
+
+    void record(std::chrono::steady_clock::time_point start, std::uint32_t bytes) {
+        if (!MEMCPY_TIMING_ENABLED) {
+            return;
+        }
+        const auto delta =
+            std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::steady_clock::now() - start);
+        g_memcpy_op_timings.push_back({delta.count(), bytes});
+    }
```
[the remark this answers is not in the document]
ajovanovicTT: Resolved with new commit.

device/pcie/device_memcpy.cpp
```diff
@@ -39,6 +45,108 @@ auto make_mmio_timeout_guard(
         });
 }
 
+// Opt-in per-op MMIO timing instrumentation, toggled by TT_UMD_MEMCPY_TIMING.
+const bool MEMCPY_TIMING_ENABLED = std::getenv("TT_UMD_MEMCPY_TIMING") != nullptr;
+
```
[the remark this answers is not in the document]
ajovanovicTT: This was intended, since per op mmio timeout is also not set on this path.

device/pcie/device_memcpy.cpp
```diff
@@ -87,12 +213,14 @@ void memcpy_to_device(volatile void* dest, const void* src, std::size_t size, co
...
 
     auto timer = make_mmio_timeout_guard("store", original_size, size, on_timeout);
+    MemcpyTimingRecorder timing("memcpy_to_device", original_size);
 
     // Phase 0: Align device destination to 4 bytes using byte-wide volatile stores.
     while (size > 0 && (reinterpret_cast<std::uintptr_t>(d) % 4) != 0) {
         auto t = std::chrono::steady_clock::now();
         *d++ = *s++;
         size--;
+        timing.record(t, 1);
```
pjanevskiTT: seems a bit weird to have both timing and timer here close in the code? Can we maybe rename timing to "recorder"

pjanevskiTT: ok let's measure and put in a PR description
- whether this slows down memcpy to device when env var is not turned on
- how much does it slow down execution when the env var is turned on

device/pcie/device_memcpy.cpp
```diff
@@ -39,6 +47,124 @@ auto make_mmio_timeout_guard(
...
+        std::int64_t total_ns = 0;
+        std::int64_t min_ns = g_memcpy_op_timings.empty() ? 0 : g_memcpy_op_timings.front().ns;
+        std::int64_t max_ns = 0;
+        for (const auto& op : g_memcpy_op_timings) {
+            total_ns += op.ns;
+            min_ns = std::min(min_ns, op.ns);
+            max_ns = std::max(max_ns, op.ns);
+        }
+        const std::size_t count = g_memcpy_op_timings.size();
+        const std::int64_t mean_ns = count == 0 ? 0 : total_ns / static_cast<std::int64_t>(count);
```
pjanevskiTT: is this mean or average?

device/pcie/device_memcpy.cpp
```diff
@@ -39,6 +47,124 @@ auto make_mmio_timeout_guard(
...
+
+        // Build into the reused thread-local buffer with std::to_chars.
+        std::string& out = g_memcpy_dump_buffer;
+        out.clear();
+        out.reserve(96 + count * 12);
```
pjanevskiTT: random constants in the code...let's document it why are we using these ones

device/pcie/device_memcpy.cpp
```diff
@@ -39,6 +47,124 @@ auto make_mmio_timeout_guard(
...
+                record_warn_emitted_ = true;
+                log_warning(LogUMD, "MemcpyTimingRecorder::record() dropped op timing on allocation failure");
+            }
+        }
+    }
+
+private:
+    // Emits one line to stderr: function name, total size, op count, min/max/mean/total ns, and the full
+    // per-op ns:bytes list in recording order.
+    void dump() const {
```
pjanevskiTT: this dump function seems very messy, let's refactor it a bit

device/pcie/device_memcpy.cpp
```diff
@@ -39,6 +47,124 @@ auto make_mmio_timeout_guard(
...
+        } catch (...) {
+            if (!record_warn_emitted_) {
+                record_warn_emitted_ = true;
+                log_warning(LogUMD, "MemcpyTimingRecorder::record() dropped op timing on allocation failure");
+            }
+        }
+    }
+
+private:
+    // Emits one line to stderr: function name, total size, op count, min/max/mean/total ns, and the full
```
pjanevskiTT: is this visible in stdout as well?

device/pcie/device_memcpy.cpp
pjanevskiTT: it would maybe be useful to somehow mark in the dump if there is some op that took more than mmio timeout config time?

device/pcie/device_memcpy.cpp
```diff
@@ -39,6 +47,124 @@ auto make_mmio_timeout_guard(
...
+
+// Reused across calls on the same thread so dump() reallocates at most once (after warmup) instead of
+// building a fresh string every call. Same non-nesting safety argument as g_memcpy_op_timings.
+thread_local std::string g_memcpy_dump_buffer;
+
+// RAII per-call timing recorder. Clears the thread-local buffer on construction, appends one entry per
+// instrumented op via record(), and dumps the aggregate summary on destruction. Dumping from the
+// destructor means a transfer that aborts via a timeout throw still reports its ops, including the
+// stalling one.
+class MemcpyTimingRecorder {
```
pjanevskiTT: maybe move this class to separate file anyway?

device/pcie/device_memcpy.cpp
pjanevskiTT: Nit: `dump()` runs unconditionally in the destructor, so a zero-op transfer (e.g. a `size == 0` memcpy) still emits a line — `ops=0 min=0 max=0 mean=0 total=0 ops_ns:bytes=` — adding noise to the diagnostic output. Consider an early `if (g_memcpy_op_timings.empty()) { return; }` at the top of `dump()`.

device/pcie/device_memcpy.cpp
```diff
@@ -39,6 +47,124 @@ auto make_mmio_timeout_guard(
...
+                log_warning(LogUMD, "MemcpyTimingRecorder::record() dropped op timing on allocation failure");
+            }
+        }
+    }
+
+private:
+    // Emits one line to stderr: function name, total size, op count, min/max/mean/total ns, and the full
+    // per-op ns:bytes list in recording order.
+    void dump() const {
+        std::int64_t total_ns = 0;
```
pjanevskiTT: Refactor idea: the manual min/max/total reduction — plus the `empty() ? front()` min seed and the `count == 0` mean guard — collapses to standard algorithms once the empty case is handled by an early return (see the nit above):

```cpp
const auto [mn, mx] = std::minmax_element(
    g_memcpy_op_timings.begin(), g_memcpy_op_timings.end(),
    [](const auto& a, const auto& b) { return a.ns < b.ns; });
const std::int64_t total_ns = std::accumulate(
    g_memcpy_op_timings.begin(), g_memcpy_op_timings.end(), std::int64_t{0},
    [](std::int64_t acc, const auto& op) { return acc + op.ns; });
const std::int64_t mean_ns = total_ns / static_cast<std::int64_t>(g_memcpy_op_timings.size());
```

Reads more clearly and drops both special-cases.

device/pcie/device_memcpy.cpp
```diff
@@ -39,6 +47,124 @@ auto make_mmio_timeout_guard(
...
+        for (const auto& op : g_memcpy_op_timings) {
+            total_ns += op.ns;
+            min_ns = std::min(min_ns, op.ns);
+            max_ns = std::max(max_ns, op.ns);
+        }
+        const std::size_t count = g_memcpy_op_timings.size();
+        const std::int64_t mean_ns = count == 0 ? 0 : total_ns / static_cast<std::int64_t>(count);
+
+        // Build into the reused thread-local buffer with std::to_chars.
+        std::string& out = g_memcpy_dump_buffer;
```
pjanevskiTT: Refactor idea: this block — the `append_int` lambda, the `num[24]` scratch, and the ~20 sequential `out += ...` / `append_int(...)` calls — is a lot of manual scaffolding for a line that only prints on the opt-in diagnostic path, where the no-allocation buffer reuse buys little (this path is never hot). `fmt` is already available via tt-logger, so consider:

```cpp
fmt::format_to(std::back_inserter(out),
    "[{}] size={} ops={} min={}ns max={}ns mean={}ns total={}ns ops_ns:bytes=",
    fn_name_, total_size_, count, min_ns, max_ns, mean_ns, total_ns);
for (std::size_t i = 0; i < count; ++i) {
    fmt::format_to(std::back_inserter(out), "{}{}:{}",
        i ? "," : "", g_memcpy_op_timings[i].ns, g_memcpy_op_timings[i].bytes);
}
out += '\n';
```

Collapses the block to a few lines while keeping the reused buffer.

device/pcie/device_memcpy.cpp
ajovanovicTT: Mean or average, not median.

device/pcie/device_memcpy.cpp
ajovanovicTT: Not worth it. The fmt version is byte-identical but ~2-3x slower here (runtime format-string parsing per op). Would go for more efficient solution here rather then go for readability.

device/pcie/device_memcpy.cpp
ajovanovicTT: Done.

device/pcie/device_memcpy.cpp
ajovanovicTT: Done.

device/pcie/device_memcpy.cpp
ajovanovicTT: Keep the recorder in device_memcpy.cpp. It is used only by the two functions in this file and it reads and writes state that lives in this file's anonymous namespace: MEMCPY_TIMING_ENABLED and the thread-local buffers g_memcpy_op_timings and g_memcpy_dump_buffer. Moving the class out would drag that state along and split reasoning that is deliberately co-located. It is small and used only here, so a separate file adds indirection without a benefit.

device/pcie/device_memcpy.cpp
ajovanovicTT: No, in current behavior just to stderr, as it was requested in issue, I could change that if wanted.

device/pcie/device_memcpy.cpp
ajovanovicTT: Documented in comments above them.

device/pcie/device_memcpy.cpp
ajovanovicTT: Added trailing '!' to every op that exceeds budget.

ajovanovicTT: Refactored. Split dump into compute stats plus append summary and append op list helpers, added append int helper.

ajovanovicTT:
Instrumentation overhead on N300 hardware
Measured on real N300 hardware over the full write_to_device / read_from_device path (Static TLB - DRAM).

With TT_UMD_MEMCPY_TIMING off (default / production):

~3 ns per call, and unmeasurable for transfers above 64 KB.
Effectively free - safe to leave compiled in for production.

With TT_UMD_MEMCPY_TIMING on:

Adds ~56 ns per instrumented op, plus a fixed per-call summary dump.
Reads: <1% overhead. Non-posted PCIe reads are already slow (~7.7 µs/op), so 56 ns is negligible.
Bulk posted writes: throughput drops from 2.62 → 1.66 GB/s (~1.6× slower). Writes are fast enough that 56 ns/op is a large fraction of per-op cost.
Small transfers: dominated by the per-call dump (~1 µs).

device/pcie/device_memcpy.cpp
ajovanovicTT: Added comment based on measurements.

device/pcie/device_memcpy.cpp
```diff
@@ -39,6 +48,167 @@ auto make_mmio_timeout_guard(
...
+            return acc + op.ns;
+        });
+    const std::size_t count = ops.size();
+    return {count, min_it->ns, max_it->ns, total_ns / static_cast<std::int64_t>(count), total_ns};
+}
+
+// Reused across calls on the same thread so steady-state recording never allocates. A single shared
+// buffer is safe because the two instrumented entry points (memcpy_to_device / memcpy_from_device) never nest on one
+// thread.
+thread_local std::vector<MemcpyOpTiming> g_memcpy_op_timings;
```
nbuncicTT: Drop the thread_local `std::vector<MemcpyOpTiming> g_memcpy_op_timings;`. In steady state every op takes a similar amount of time, so the full list is just noise. The only interesting ops are the anomalies (ones that exceed the per-op budget but are vetoed as false alarms, so the transfer continues instead of throwing).

Store just those, in a small fixed-size array on the stack, which is bounded, no heap, no TLS. (Or a std::vector if you prefer, but it would now grow far more slowly since it only holds anomalies.)

The output loses the `ops_ns:bytes=...` list and gains a slow_ops field instead, e.g.:

`... min=.. max=.. mean=.. total=.. slow_ops=[op#1729:98211ns, ...]`

device/pcie/device_memcpy.cpp
```diff
@@ -87,12 +257,14 @@ void memcpy_to_device(volatile void* dest, const void* src, std::size_t size, co
     auto* s = static_cast<const std::uint8_t*>(src);
 
     auto timer = make_mmio_timeout_guard("store", original_size, size, on_timeout);
+    MemcpyTimingRecorder recorder("memcpy_to_device", original_size);
```
nbuncicTT: Instead of duplicating the same (t, bytes) args across two calls at every op site, connect the recorder and the timeout guard with a local lambda. The two classes stay fully decoupled (neither references the other), and the two per-op calls fold into a single call site:
```cpp
auto timer = make_mmio_timeout_guard("store", original_size, size, on_timeout);
MemcpyTimingRecorder recorder("memcpy_to_device", original_size);   // own class, RAII, dumps on dtor
auto check = [&](std::chrono::steady_clock::time_point t, std::uint32_t b) {
    recorder.record(t, b);          // record first: a throwing check still counts this op
    timer.record_and_check(t, b);   // may throw on overrun
};
// ...per op, one call:
check(t, 256);
```
Ordering matters: record before check, so an op that trips the timeout throw is still counted.

device/pcie/device_memcpy.cpp
```diff
@@ -39,6 +48,167 @@ auto make_mmio_timeout_guard(
...
+}
+
+// Reused across calls on the same thread so steady-state recording never allocates. A single shared
+// buffer is safe because the two instrumented entry points (memcpy_to_device / memcpy_from_device) never nest on one
+// thread.
+thread_local std::vector<MemcpyOpTiming> g_memcpy_op_timings;
+
+// Reused across calls on the same thread so dump() reallocates at most once (after warmup) instead of
+// building a fresh string every call. Same non-nesting safety argument as g_memcpy_op_timings.
+thread_local std::string g_memcpy_dump_buffer;
```
nbuncicTT: Drop `g_memcpy_dump_buffer` and its thread-local, along with `append_summary` / `append_int`. It's thread-local only to reuse the buffer across calls without threads racing on a shared one, but with the per-op list gone, the summary line is small and bounded, so there's nothing worth reusing.

Format straight into a fixed-size stack `std::array<char, N>` with `fmt::format_to_n` (type-safe, bounded, non-allocating): thread-safe by construction, zero heap, no global state, and it replaces all the manual out += ... / append_int token-building.

```cpp
std::array<char, 256> buf;
auto [end, _] = fmt::format_to_n(buf.data(), buf.size(),
    "[{}] size={} ops={} min={}ns max={}ns mean={}ns total={}ns",
    fn_name_, total_size_, count, min_ns_, max_ns_, mean_ns_, total_ns_);
// emit: fwrite(buf.data(), 1, end - buf.data(), stderr)
```

device/pcie/device_memcpy.cpp
```diff
@@ -39,6 +48,167 @@ auto make_mmio_timeout_guard(
...
+
+    MemcpyTimingRecorder(const MemcpyTimingRecorder&) = delete;
+    MemcpyTimingRecorder& operator=(const MemcpyTimingRecorder&) = delete;
+
+    ~MemcpyTimingRecorder() noexcept {
+        if (!MEMCPY_TIMING_ENABLED) {
+            return;
+        }
+
+        try {
```
nbuncicTT: The `try`/`catch` in both `~MemcpyTimingRecorder()` and `record()` (plus the `record_warn_emitted_` flag) can go. They only guard `std::bad_alloc` from the two heap allocations: `g_memcpy_op_timings.push_back` and the string append in `dump()`. In the reworked design nothing on the path allocates: the aggregate is 4 scalar members, anomalies are a fixed `std::array<SlowOp, 16>`, and the line is formatted into a fixed stack char buffer. No allocation means no throw so there is no need to catch. `record()` becomes `noexcept`, and the dtor is just `if (enabled_ && count_ > 0) dump();`.

This can be:
```cpp
struct SlowOp {
    std::size_t  op;   // which op (1-based index in recording order)
    std::int64_t ns;   // how long it took
};
```

device/pcie/device_memcpy.cpp
```diff
@@ -39,6 +48,167 @@ auto make_mmio_timeout_guard(
...
+// building a fresh string every call. Same non-nesting safety argument as g_memcpy_op_timings.
+thread_local std::string g_memcpy_dump_buffer;
+
+// RAII per-call timing recorder. Clears the thread-local buffer on construction, appends one entry per
+// instrumented op via record(), and dumps the aggregate summary on destruction. Dumping from the
+// destructor means a transfer that aborts via a timeout throw still reports its ops, including the
+// stalling one.
+class MemcpyTimingRecorder {
+public:
+    MemcpyTimingRecorder(const char* fn_name, std::size_t total_size) : fn_name_(fn_name), total_size_(total_size) {
```
nbuncicTT: The blocker you raised is gone if you delete the thread locals. i.e. if the two thread-locals are being deleted, and `MEMCPY_TIMING_ENABLED` doesn't need to move, the recorder takes the enable flag as a constructor argument, so it stays here and gets injected at the call site. The class no longer reads or writes any file-local state, so there's nothing to "drag along" and no reasoning to split. It's now a self-contained ~30-line RAII helper with no ties to the transfer code, so moving it to `device/pcie/memcpy_timing_recorder.hpp` keeps this file focused on the SIMD path. Env var stays here for now (its home is the still-open logging question).

nbuncicTT (commented):
Check with the new changes if the stack allocated approach works better (without the storing of the expected values, and only with the anomalies).

ajovanovicTT:
Instrumentation overhead comparison, MicrobenchmarkTLB.DRAM, TT_UMD_MEMCPY_TIMING on vs off.

At 813f1139, small transfers (1 to 8 bytes) showed 75 to 870 percent overhead in ns per byte; large transfers (1MB+) showed 60 to 65 percent for writes, under 1 percent for reads.

At the latest commit, every case is lower: small transfers now 55 to 315 percent, mid size ops improved the most (e.g. static write 4096B: 142 to 74 percent, 8192B: 90 to 16 percent), and large transfers now show 0 percent overhead across the board.

Conclusion: overhead is better now, no regressions. Refactors since 813f1139 cut the fixed per call instrumentation cost roughly 10 to 50 percent, biggest wins on mid and large transfers.

With instrumentation OFF, there is effectively no overhead. The OFF numbers at 813f1139 and latest match closely (small transfers wobble a few percent, likely host noise; everything at 1024B and up is identical to 2 decimal places). All the cost only appears when TT_UMD_MEMCPY_TIMING=1 is set - when unset, the code path is just a cached bool check, so OFF overhead is negligible/zero.

ajovanovicTT:
Slow op caused by sleep for example:

Screenshot 2026-07-24 at 11 48 31

ajovanovicTT:
Work on optimization to be conitnued with https://github.com/tenstorrent/tt-umd/issues/3073.

nbuncicTT (approved):

PR #3029, opened 2026-07-15
Simulation: SHUTDOWN command for graceful host shutdown over the socket

pjanevskiTT:
### Issue
Part of #2677 (simulation server process). To manage long-running simulation hosts (a follow-up `sim_server` tool with start/list/kill), a host needs to be stoppable **in-band over its socket** rather than by PID/signal: the per-chip socket is deliberately world-writable and cross-user, while a signal is same-uid only — so only a socket message reaches exactly the set of processes that can reach the server. Stacked on #3025.

### Description
Adds a `SHUTDOWN` command to the wire protocol and the host-side hook + client-side error to make graceful shutdown work end to end.

- **Wire:** `SHUTDOWN` added to `SimulationServerCommand`; acked with a plain `SimulationServerResponse` (no new message table).
- **Host:** the shutdown handler is passed into `SimulationTTDevice::adopt_socket(socket, handler)` and captured before serving starts — immutable, so no setter, no member, no lock. `handle_request` invokes it on `SHUTDOWN` and acks. A host that passes none acks as a **no-op**, so a simulation embedded in another program can't be torn down by a stray client. The handler must only *signal* — it runs on a serving thread, so tearing down there would self-join the serving threads and deadlock; a dedicated server signals its main thread to exit, which then drops the `Cluster` for a graceful teardown. It must also be safe to call more than once, since every attached client that sends SHUTDOWN invokes it.
- **Client (graceful failure):** `recv_framed` now maps a clean peer close (EOF / connection reset) to a **distinct, clearly-worded error** ("…the remote end (e.g. a stopped simulation host) has gone away") instead of the generic read-failure message, so a client blocked in / next calling `transact()` when the host is stopped fails with an unmistakable exception and unwinds gracefully. On teardown the host cleanly closes every client connection, so all attached clients get this error on their blocked/next call.

### List of the changes
- `simulation_server_protocol.{fbs,hpp}` — `SHUTDOWN` command.
- `simulation_tt_device.{hpp,cpp}` — `adopt_socket()` takes an optional shutdown handler, captured into the serving lambda before the accept loop starts (no member/mutex); `handle_request` SHUTDOWN branch invokes it and acks (no-op if none).
- `cluster.{hpp,cpp}` — `ClusterOptions::simulation_shutdown_handler`, plumbed through `serve_simulation_devices_over_sockets()` into `adopt_socket()`.
- `simulation_server_transport.cpp` — `recv_framed` distinct peer-close error.
- Tests: `SHUTDOWN` request round-trip (protocol lane); `recv_framed` reports a peer close distinctly (always-built transport lane).

### Testing
Build + pre-commit clean; 180 baremetal + 12 protocol tests pass. End-to-end shutdown (via the follow-up tool) is CI-gated.

### API Changes
`SimulationTTDevice::adopt_socket()` gains an optional `std::function<void()>` shutdown handler; `ClusterOptions::simulation_shutdown_handler` added; `SimulationServerCommand::Shutdown` added. Simulation-internal.

### DIFF
--- a/device/api/umd/device/simulation/simulation_server_protocol.hpp
+++ b/device/api/umd/device/simulation/simulation_server_protocol.hpp
@@ -29,6 +29,9 @@ enum class SimulationServerCommand : int8_t {
     Write = 1,
     GetDeviceInfo = 2,
     GetClusterDescriptor = 3,
+    // Asks the host to shut down gracefully; the host acks with a SimulationServerResponse then tears
+    // down (closing every client connection). A host that didn't opt in acks as a no-op.
+    Shutdown = 4,
 };
 
 // Which simulator the host runs; mirrors wire::SimulationBackendType. Served as part of the device
--- a/device/api/umd/device/tt_device/simulation_tt_device.hpp
+++ b/device/api/umd/device/tt_device/simulation_tt_device.hpp
@@ -7,6 +7,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <filesystem>
+#include <functional>
 #include <memory>
 #include <mutex>
 #include <vector>
@@ -37,8 +38,18 @@ class SimulationTTDevice : public TTDevice {
 public:
     ~SimulationTTDevice() override;
 
-    // Takes ownership of the serving socket that exposes this device (created by discovery).
-    void adopt_socket(std::unique_ptr<SimulationServerSocket> socket);
+    // Takes ownership of the serving socket that exposes this device (created by discovery) and
+    // begins serving. An optional shutdown_handler is invoked when a client sends SHUTDOWN: a
+    // dedicated server (e.g. the sim_server tool) passes one to signal its main thread to exit and
+    // tear this device down; a host that passes none acks SHUTDOWN as a no-op, so a simulation
+    // embedded in another program can't be torn down by a stray client. The handler is fixed here,
+    // before serving starts, and never mutated afterwards -- so it needs no locking.
+    //
+    // The handler runs on a serving thread and MUST be non-blocking -- it must only signal (set a
+    // flag / notify a condition variable / write a self-pipe); tearing down from within it would
+    // join the serving threads from one of them and deadlock. It must also be safe to call more than
+    // once and concurrently: every attached client that sends SHUTDOWN invokes it.
+    void adopt_socket(std::unique_ptr<SimulationServerSocket> socket, std::function<void()> shutdown_handler = {});
 
     // --- TTDevice overrides whose behavior is identical across both simulation backends ---
     void read_from_device(
@@ -160,7 +171,8 @@ private:
     // socket's connection threads; read/write take device_lock, so concurrent host + client
     // access is serialized. The socket layer stays protocol-agnostic; this is where the protocol
     // is (de)serialized.
-    std::vector<uint8_t> handle_request(const std::vector<uint8_t>& request_bytes);
+    std::vector<uint8_t> handle_request(
+        const std::vector<uint8_t>& request_bytes, const std::function<void()>& shutdown_handler);
 
     // The device serves one of two disjoint roles; read_from_device/write_to_device dispatch on
     // this rather than a bare client_ null-check so the intent is named at the call site.
--- a/device/tt_device/simulation_tt_device.cpp
+++ b/device/tt_device/simulation_tt_device.cpp
@@ -46,13 +46,19 @@ void SimulationTTDevice::detach_client() {
     }
 }
 
-void SimulationTTDevice::adopt_socket(std::unique_ptr<SimulationServerSocket> socket) {
+void SimulationTTDevice::adopt_socket(
+    std::unique_ptr<SimulationServerSocket> socket, std::function<void()> shutdown_handler) {
     socket_ = std::move(socket);
-    // Begin serving remote clients now that the backend is up.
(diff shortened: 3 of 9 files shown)


device/tt_device/simulation_tt_device.cpp
```diff
@@ -52,6 +52,8 @@ void SimulationTTDevice::adopt_socket(std::unique_ptr<SimulationServerSocket> so
     socket_->serve([this](const std::vector<uint8_t>& request_bytes) { return handle_request(request_bytes); });
 }
 
+void SimulationTTDevice::set_shutdown_handler(std::function<void()> handler) { shutdown_handler_ = std::move(handler); }
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed: `shutdown_handler_` is now guarded by `shutdown_handler_mutex_`. `set_shutdown_handler()` assigns under the lock, and `handle_request()` copies the handler under the lock then invokes the copy outside it.

device/tt_device/simulation_tt_device.cpp
```diff
@@ -81,6 +83,17 @@ std::vector<uint8_t> SimulationTTDevice::handle_request(const std::vector<uint8_
...
+    // Shutdown: invoke the opt-in handler (a dedicated server sets it to signal its main thread to
+    // exit and tear down; an embedded host leaves it empty, making this a no-op) and ack. The
+    // handler must only signal -- it must not tear down from this serving thread. The ack is sent
+    // before any teardown, and this serving thread hits EOF and is joined during that teardown.
+    if (request.command == SimulationServerCommand::Shutdown) {
+        if (shutdown_handler_) {
+            shutdown_handler_();
+        }
+        return encode(SimulationServerResponse{});  // status 0
+    }
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed: `shutdown_handler_` is now guarded by `shutdown_handler_mutex_`. `set_shutdown_handler()` assigns under the lock, and `handle_request()` copies the handler under the lock then invokes the copy outside it.

device/api/umd/device/simulation/simulation_server_protocol.hpp
```diff
@@ -29,6 +29,9 @@ enum class SimulationServerCommand : int8_t {
     Write = 1,
     GetDeviceInfo = 2,
     GetClusterDescriptor = 3,
+    // Asks the host to shut down gracefully; the host acks with a SimulationServerResponse then tears
+    // down (closing every client connection). A host that didn't opt in acks as a no-op.
+    Shutdown = 4,
```
nbuncicTT, line 34: nit: **Please** figure out a way to retain memory on comments I've written for the simulator epic - I've told multiple times that for enum classes to use upper snake case.

device/api/umd/device/tt_device/simulation_tt_device.hpp
```diff
@@ -40,6 +41,16 @@ class SimulationTTDevice : public TTDevice {
...
 
+    // Opt into remote shutdown: a client's SHUTDOWN request will invoke this callback (see
+    // handle_request). A dedicated server (e.g. the sim_server tool) sets it to signal its main
+    // thread to exit and tear this device down; a host that never sets one acks SHUTDOWN as a no-op,
+    // so a simulation embedded in another program can't be torn down by a stray client.
+    //
+    // The callback runs on a serving thread and MUST be non-blocking -- it must only signal (set a
+    // flag / notify a condition variable / write a self-pipe). Tearing down from within it would
+    // join the serving threads from one of those threads and deadlock.
+    void set_shutdown_handler(std::function<void()> handler);
```
nbuncicTT: Since the handler is only ever set once, before serving starts, consider passing it into `adopt_socket()` instead of exposing a separate setter. That makes it immutable once the socket is published, which removes `shutdown_handler_mutex_`, the per-request `std::function` copy on the serving thread, and both new protected members from the base class — and it closes the window where SHUTDOWN's meaning changes underneath an already-connected client. Worth documenting whether the handler must be idempotent, since two clients sending SHUTDOWN will invoke it twice.

nbuncicTT (approved):

device/api/umd/device/simulation/simulation_server_protocol.hpp
pjanevskiTT: Done in the dedicated casing PR #3082 (stacked on top of the whole stack): both `SimulationServerCommand` and `SimulationBackendType` are now UPPER_SNAKE, matching the .fbs. Tracked and resolved there.

pjanevskiTT: We'll cover all the enum-class casing across the epic in a single PR stacked on top of the whole stack (#3082) -- both `SimulationServerCommand` and `SimulationBackendType`, matching the `.fbs`. Leaving this open until #3082 lands rather than resolving it here, since this PR's own diff still shows the old casing.

device/api/umd/device/tt_device/simulation_tt_device.hpp
pjanevskiTT: Done. The handler is now passed into `adopt_socket()` (the tool supplies it via `ClusterOptions::simulation_shutdown_handler`), captured into the request handler before serving starts. That removes `set_shutdown_handler()`, `shutdown_handler_`, `shutdown_handler_mutex_`, and the per-request copy, and closes the window you flagged (the handler is fixed before the sockets accept). Documented on `adopt_socket()` that it must be non-blocking and safe to call more than once, since each attached client that sends SHUTDOWN invokes it.

PR #3030, opened 2026-07-15
Simulation: sim_server tool to start / list / kill simulation hosts

pjanevskiTT:
### Issue
Part of #2677 (simulation server process). People want to run a long-running simulation host in the background and have other UMD processes attach to it as clients. This adds a small CLI to manage those hosts. Stacked on #3099 (per-server socket directories), which lets it list and target individual servers.

### Description
`tools/sim_server` — a management CLI with three subcommands:

- **`start <simulator.so | rtl-dir>`** — pre-allocates a fresh server directory, then daemonizes (fork + `setsid` + stdio→log) and brings up a `Cluster{ChipType::SIMULATION}` host in it, opting into socket serving via `ClusterOptions::serve_simulation_devices_over_sockets` (off by default after #3112 — publishing the per-chip sockets for clients is this tool's whole purpose), so it serves each chip's socket + device-info + cluster-descriptor, prints the pid and directory, and returns. The daemon opts into over-socket shutdown: it passes a handler via `ClusterOptions::simulation_shutdown_handler` (from #3029) that writes a self-pipe to unblock its main thread, fixed before the sockets start serving; `SIGTERM`/`SIGINT` do the same. The self-pipe's write end is non-blocking, so signalling from the serving thread or a signal handler never blocks. On wake it drops the `Cluster` for a graceful teardown that closes every client connection.
- **`list`** — enumerates the open servers (via `SimulationConnector::list_servers()`) and shows `SERVER · CHIP · STATE · ARCH · SOCKET`, probing liveness + identity. A socket with no live host shows as `unreachable` (stale).
- **`kill <server>`** — connects to that server's socket and sends `SHUTDOWN`, addressing the server by its index (chip id is not unique across servers). In-band over the socket (not PID/signal), matching the socket's world-writable, cross-user access model.

### List of the changes
- `tools/sim_server.cpp` (new) + `tools/CMakeLists.txt` — the tool, gated on `TT_UMD_BUILD_SIMULATION`, following the existing tool scheme (cxxopts, `tools_common`, `tt::LogUMD`).
- `.github/workflows/run-ttsim-tests.yml` — build/exercise the tool in the simulation CI lane.

### Testing
Build + pre-commit clean. `--help` / `list` / `kill <unknown>` / bad-subcommand exercised locally (`list` correctly flags a stale socket as `unreachable`). The `start` daemon + end-to-end start/list/kill cycle needs a real simulator `.so` and runs in the simulation CI lane.

### API Changes
New `sim_server` tool binary; consumes `SimulationConnector::list_servers()` (added in #3099). Otherwise internal.

### DIFF
--- a/.github/workflows/run-ttsim-tests.yml
+++ b/.github/workflows/run-ttsim-tests.yml
@@ -150,8 +150,11 @@ jobs:
       - name: Run sim_server start / list / kill on tt-sim blackhole
         timeout-minutes: 2
         run: |
+          # Resolve the simulator before cd'ing: this job runs in a container, where github.workspace
+          # is the *host* path, so use the working directory the sim paths were prepared in instead.
+          simulator=$(pwd)/sim_bh/libttsim.so
           cd tt-metal/tt_metal/third_party/umd/build/tools/umd
-          ./sim_server.sh start ${{ github.workspace }}/sim_bh/libttsim.so
+          ./sim_server.sh start "$simulator"
           ./sim_server.sh list | tee /tmp/sim_server_list.txt
           grep -q "live" /tmp/sim_server_list.txt
           ./sim_server.sh kill 0
--- a/.github/workflows/run-ttsim-tests.yml
+++ b/.github/workflows/run-ttsim-tests.yml
@@ -159,6 +159,15 @@ jobs:
           grep -q "live" /tmp/sim_server_list.txt
           ./sim_server.sh kill 0
 
+      # The point of the server: a second process attaches to a hosted simulation and device I/O
+      # crosses the socket. SimulationConnector covers exactly that at the Cluster level (a host
+      # Cluster writes, a client Cluster reads it back), but was in no gtest filter, so it never ran.
+      - name: Run UMD SimulationConnector tests on tt-sim blackhole
+        timeout-minutes: 2
+        run: |
+          export TT_UMD_SIMULATOR=$(pwd)/sim_bh/libttsim.so
+          tt-metal/tt_metal/third_party/umd/build/test/umd/api/api_tests --gtest_filter="SimulationConnector.*"
+
       # Following single card tests should be really fast, only a couple of seconds each, so we set a short timeout for
       # them and if they time out, it indicates some issue with the simulator setup rather than an actual test failure.
       - name: Run tt-sim wormhole tests
--- a/.github/workflows/run-ttsim-tests.yml
+++ b/.github/workflows/run-ttsim-tests.yml
@@ -144,6 +144,18 @@ jobs:
           # builds the tools with simulation OFF, which excludes it).
           cmake --build tt-metal/tt_metal/third_party/umd/build --target api_tests umd_microbenchmark sim_server
 
+      # Exercises the tool, not just its build: a single-chip simulator ships no cluster_descriptor.yaml,
+      # which is the case where a host with no chips would serve no sockets and still look like it started.
+      # `list` must report the chip as live, and `kill` must be able to reach it over that socket.
+      - name: Run sim_server start / list / kill on tt-sim blackhole
+        timeout-minutes: 2
+        run: |
+          cd tt-metal/tt_metal/third_party/umd/build/tools/umd
+          ./sim_server.sh start ${{ github.workspace }}/sim_bh/libttsim.so
+          ./sim_server.sh list | tee /tmp/sim_server_list.txt
+          grep -q "live" /tmp/sim_server_list.txt
+          ./sim_server.sh kill 0
+
       # Following single card tests should be really fast, only a couple of seconds each, so we set a short timeout for
       # them and if they time out, it indicates some issue with the simulator setup rather than an actual test failure.
       - name: Run tt-sim wormhole tests
--- a/.github/workflows/run-ttsim-tests.yml
+++ b/.github/workflows/run-ttsim-tests.yml
@@ -128,7 +128,7 @@ jobs:
           cp tt-metal/tt_metal/third_party/umd/tests/cluster_descriptor_examples/blackhole_P300_both_mmio.yaml \
             sim_bh_x2/cluster_descriptor.yaml
 
-      - name: Build UMD api_tests and microbenchmark with simulation support
+      - name: Build UMD api_tests, microbenchmark and sim_server with simulation support
         run: |
           cmake -B tt-metal/tt_metal/third_party/umd/build -G Ninja \
             -S tt-metal/tt_metal/third_party/umd \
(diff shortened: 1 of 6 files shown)


tools/sim_server.cpp
```diff
@@ -0,0 +1,235 @@
...
+    ::setsid();
+    if (::chdir("/") != 0) { /* best effort; not fatal */
+    }
+    redirect_stdio_to_log();
+
+    if (::pipe(g_stop_pipe) != 0) {
+        _exit(1);
+    }
+    std::signal(SIGTERM, on_signal);
+    std::signal(SIGINT, on_signal);
```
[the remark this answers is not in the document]
pjanevskiTT: Done. The self-pipe's write end is set non-blocking (`fcntl(..., O_NONBLOCK)`) right after `pipe()`, so `request_stop()` from the signal handler or a serving thread cannot block: a full pipe just fails with `EAGAIN`, which is harmless since a wakeup byte is already pending. The read end stays blocking (the main thread blocks on it until stopped).

device/simulation/simulation_connector.cpp
```diff
@@ -88,6 +88,12 @@ SimulationConnector::Role SimulationConnector::role_for(const std::filesystem::p
     return classify(simulator_directory) == PathKind::Client ? Role::Client : Role::Host;
 }
 
+std::map<ChipId, std::filesystem::path> SimulationConnector::list_servers() {
+    // Per-chip sockets live directly under the system temp dir (see default_socket_path); scanning
+    // it yields every open server without connecting to any.
+    return SimulationServerSocket::sockets_in_directory(std::filesystem::temp_directory_path());
```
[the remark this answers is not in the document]
pjanevskiTT: The stacked #3051 reworks `list_servers()` to call `SimulationServerSocket::list_server_directories()`, moving the temp-dir + naming knowledge into `SimulationServerSocket` (next to `default_socket_path`) instead of hard-coding it here — the single-source-of-truth this asks for. Resolving as handled in the follow-up.

pjanevskiTT:
Addressed the three low-confidence findings from the latest Copilot review (commit `055bfb8a`):

- **`kill` error message (`sim_server.cpp`)** — was `"requires a <chip_id> argument"`; `kill` takes a **server index**, matching the CLI help and README. Now says `<server>`.
- **`fork()` failure leaked the claimed directory** — `allocate_server_directory()` creates the dir eagerly before the fork, so a failed start orphaned an empty `tt-umd-sim-server-*` dir (cluttering `list`, consuming indices). `cmd_start` now best-effort removes it on the fork-failure path.
- **pipe / Cluster-construction failure leaked it too** (same root cause) — same best-effort cleanup added to the child's `pipe()`-failure and `Cluster` construction-failure paths.

On the success path the running host still owns the directory and its `SimulationServerSocket` removes it on graceful shutdown, so the cleanup only covers the paths that never get that far.

.github/workflows/run-ttsim-tests.yml
```diff
@@ -128,19 +128,36 @@ jobs:
...
+      - name: Build UMD api_tests, microbenchmark and sim_server with simulation support
         run: |
           cmake -B tt-metal/tt_metal/third_party/umd/build -G Ninja \
             -S tt-metal/tt_metal/third_party/umd \
             -DTT_UMD_BUILD_TESTS=ON \
             -DTT_UMD_BUILD_SIMULATION=ON \
             -DTT_UMD_BUILD_PYTHON=OFF \
             -DTT_UMD_BUILD_EXAMPLES=OFF \
-            -DTT_UMD_BUILD_TOOLS=OFF \
+            -DTT_UMD_BUILD_TOOLS=ON \
```
broskoTT, line 139: I don't like this being part of "tools". That directory is meant for utility tools, whereas this will be ran as part of CI or some other production testing pipelines. I think it should be in a separate directory, and probably a totally separate target

tools/sim_server.cpp
```diff
@@ -0,0 +1,197 @@
...
+                chip_id,
+                live ? "live" : "unreachable",
+                live ? arch : "-",
+                socket_path.string());
+        }
+    }
+    return 0;
+}
+
+int cmd_kill(int server_index) {
```
broskoTT, line 129: Maybe it is easier to just send a SIGTERM from sim_server.sh, rather than writing a custom kill command in C++? Or am i missing some hidden feature here

broskoTT (approved):
Thanks to both for contribution on this front

.github/workflows/run-ttsim-tests.yml
```diff
@@ -128,19 +128,36 @@ jobs:
...
+      # which is the case where a host with no chips would serve no sockets and still look like it started.
+      # `list` must report the chip as live, and `kill` must be able to reach it over that socket.
+      - name: Run sim_server start / list / kill on tt-sim blackhole
+        timeout-minutes: 2
+        run: |
+          # Resolve the simulator before cd'ing: this job runs in a container, where github.workspace
+          # is the *host* path, so use the working directory the sim paths were prepared in instead.
+          simulator=$(pwd)/sim_bh/libttsim.so
+          cd tt-metal/tt_metal/third_party/umd/build/tools/umd
+          ./sim_server.sh start "$simulator"
```
broskoTT, line 157: We should definitely also test what we vouch this server is solving (not up to date whether such tests already maybe exist):
We should start a simulation server, either through this script or through regular metal unit test. Then pause it, and then run another process which attaches and checks that is indeed the same server (maybe by inspecting some random device location.

This can be a separate PR, but in that case please create an issue

tools/sim_server.cpp
nbuncicTT: No pid is written anywhere on disk, so `SIGTERM` could only reach servers started by that same shell, whereas the socket is world-writable and works across terminals and users and `SimulationServerCommand::Shutdown` already exists from #3029, so cmd_kill is just connect/send/check. Your instinct does point at a real gap though: a pidfile would let us reap a host that has wedged and stopped answering its socket, which kill can't do today. I'll create a follow-up.

.github/workflows/run-ttsim-tests.yml
nbuncicTT: The test already exists but never ran: `SimulationConnector.HostAndClientClustersShareDeviceMemory` hosts one `Cluster`, attaches a second as a client, and checks a pattern written by the host reads back over the socket, but `SimulationConnector` is in no `gtest_filter`, and it would have failed if it had run. Fixed and added a CI step for it here rather than a separate PR; passes on `sim_bh` and `sim_bh_x2`.

.github/workflows/run-ttsim-tests.yml
nbuncicTT: Enabling `TT_UMD_BUILD_TOOLS` does not build the other tools, a clean configure plus --target sim_server produces only sim_server and libtt-umd, so the flag costs configure time only. That leaves it purely as a placement preference; the target is already independently gated on `TT_UMD_BUILD_SIMULATION`, so it can be moved tools/simulation/, but as a follow up.

PR #3036, opened 2026-07-17
Use kmdlib in read_device_info()

aleksamarkovicTT:
### Issue
#2749 

### Description
Use kmdlib in read_device_info()

### List of the changes
- Add tt_device_get_attrs() to kmdlib.
- Remove ioctl call in read_device_info()
- Change read_device_info() to take path instead of fd.

### Testing
CI

### API Changes
This PR has API changes.


device/api/umd/device/pcie/pci_device.hpp
```diff
@@ -149,11 +149,11 @@ class PCIDevice {
...
 
     /**
-     * Read device information from sysfs.
-     * @param fd Integer corresponding to the character device descriptor of the PCI device.
+     * Read device information for the given device.
+     * @param device_path Path to the character device (e.g. "/dev/tenstorrent/0").
      * @return PciDeviceInfo struct containing the device information.
      */
-    static PciDeviceInfo read_device_info(int fd);
+    static PciDeviceInfo read_device_info(const std::string &device_path);
```
pjanevskiTT, line 156: why change this?

aleksamarkovicTT: Because in kmdlib, tt_device_t is an opaque type. You cannot create one with an existing fd, and I don't think it makes sense even if you could. Design wise, I think that fds are not at the correct abstraction layer for PciDevice anyway, the goal is for everything to go through kmdlib.

pjanevskiTT: ok makes sense, let's just make sure this is not called multiple times in any of the important paths (ctor, IO, etc...)

pjanevskiTT (approved):

PR #3052, opened 2026-07-20
Replace pin/unpin pages ioctl with kmdlib

aleksamarkovicTT:
### Issue
#2749 

### Description
Use kmdlib for pinning/unpinning pages in PciDevice.

### List of the changes
- Add tt_pin_pages() and tt_unpin_pages() to kmdlib.
- Remove ioctl calls in PciDevice related to pinning pages.

### Testing
CI

### API Changes
This PR has API changes.


device/api/umd/device/tt_kmd_lib/tt_kmd_lib.h
```diff
@@ -287,6 +287,16 @@ enum tt_dma_map_flags {
...
+
+    /**
+     * @brief Attests that the pages backing the buffer are physically
+     * contiguous.
+     *
+     * When the system IOMMU is not active, the driver requires this attestation
+     * to expose the buffer as a single contiguous physical (or NOC) address
+     * range, e.g. for hugepages. It has no effect when an IOMMU is active.
+     */
+    TT_DMA_FLAG_CONTIGUOUS = 1 << 2,
```
nbuncicTT, line 299: These flags don't really describe "DMA." Two separate things live here: NONE/NOC/NOC_TOP_DOWN control NOC-to-host aperture placement (device-side NOC access), and a contiguity attestation would be a pinning-layout concern. Neither touches the PCIe DMA engine. The underlying op is "pin pages, optionally expose to NOC," matching `TENSTORRENT_IOCTL_PIN_PAGES`. 

I'd rename the surface to pin-pages terms (`tt_pin_pages_flags`, `TT_PIN_PAGES_`*). Note `tt_dma_map()` itself carries the "dma" name too, so a rename should cover the function or none of it. Happy to split into a follow-up PR / discussion if preferred.

device/pcie/pci_device.cpp
```diff
@@ -601,32 +601,30 @@ PCIDevice::~PCIDevice() {
...
         return 0;
     }
 
     log_debug(
         LogUMD,
         "Pinning pages for Hugepage: virtual address {:#x} and size {:#x} pinned to physical address {:#x}",
-        pin_pages.in.virtual_address,
-        pin_pages.in.size,
-        pin_pages.out.physical_address);
+        vaddr,
```
nbuncicTT: can we be a bit more consistent with naming, and if `physical_address` is named like this, can we also name `vaddr` as `virtual_address`?

nbuncicTT (approved):

device/api/umd/device/tt_kmd_lib/tt_kmd_lib.h
aleksamarkovicTT: I agree, but let's do this in another PR. The names stem from Joel's original implementation

PR #3053, opened 2026-07-20
Simulation: user-facing doc for the shared simulation server

pjanevskiTT:
### Issue
Part of #2677 (simulation server process). The shared host/client simulation model had no user-facing documentation — how the role is chosen, how to run and attach to a server, and how to connect from code. Stacked on #3099 (per-server socket directories), which this doc describes.

### Description
Adds `docs/SIMULATION_SERVER.md`, a concise, flow-first guide for external users of the shared simulation server. It covers:

- **Host and client** — the role is decided by what you point UMD at (a simulator ⇒ host; a server's socket directory ⇒ client), with a topology diagram.
- **More than one server** — each host owns its own directory, so several servers coexist on one machine (even with the same chip id); a second topology diagram shows two side by side.
- **The flow** — `start` / `list` / attach / `kill`, with a sequence diagram of a server's lifetime (start → attach → I/O → shutdown → "server stopped").
- **Connecting from code** — `.cpp` snippets at the `Cluster` level and the `SimulationConnector::discover()` (discovery) level.
- **Where things live** — server directories (`tt-umd-sim-server-<index>`), the per-chip sockets inside them, and the server logs.

Diagrams are Mermaid (rendered inline by GitHub) so they live in the `.md` and stay reviewable in diffs. It deliberately stays on the shared host/client model and does not document the simulator backend internals (`.so` / RTL) or the full API surface.

### List of the changes
- `docs/SIMULATION_SERVER.md` (new) — the user-facing doc, with three Mermaid diagrams.

### Testing
Docs only. Mermaid diagrams verified to render.

### API Changes
None (documentation only).

### DIFF
--- a/docs/SIMULATION_SERVER.md
+++ b/docs/SIMULATION_SERVER.md
@@ -1,9 +1,12 @@
 # Simulation Server
 
-UMD can run against a *simulated* device instead of real hardware. The **simulation server** lets one
-process **host** a simulated device and have other UMD processes **attach** to it and drive it exactly
-as they would a real cluster. That way a single running simulation can be shared by several processes
-at once, instead of each process spinning up its own.
+UMD can run against a *simulated* device instead of real hardware, using
+[ttsim](https://github.com/tenstorrent/ttsim) (a `libttsim.so` for a given architecture) or an RTL
+simulator build. The **simulation server** lets one process **host** a simulated device and have other
+UMD processes **attach** to it and drive it exactly as they would a real cluster.
+
+Without it each process gets its own independent simulation, so two processes never see each other's
+writes.
 
 ## Host and client
 
@@ -15,6 +18,18 @@ You don't pick a role explicitly — UMD decides it from what you point it at:
 
 A host and its clients run on the same machine and communicate over per-chip sockets kept in that
 server's directory (see [Where things live](#where-things-live)).
+
+### Any UMD process can be the host
+
+Hosting is a `Cluster` option, not something special to the tool:
+
+```cpp
+options.serve_simulation_devices_over_sockets = true;   // default: false
+```
+
+It is **off by default**, so an ordinary simulator run — a UMD or tt-metal test pointed at a
+`libttsim.so` — hosts a private in-process simulation and publishes nothing. `sim_server` is just a
+host that sets the flag and stays alive.
 
 ```mermaid
 flowchart TB
@@ -86,11 +101,21 @@ flowchart TB
    1        0      live    blackhole/ttsim  /tmp/tt-umd-sim-server-1/tt-umd-sim-0.sock
    ```
 
-3. **Use it from your program.** Open a UMD cluster in simulation mode pointed at a server's socket
-   directory (the `SOCKET`'s parent above, e.g. `/tmp/tt-umd-sim-server-0`). UMD sees the live
-   sockets there, attaches as a client, and from then on you use the cluster exactly as you would
-   against real hardware — reading and writing device memory and so on. Your program's code is the
-   same whether it runs on hardware or against a shared simulation.
+3. **Use it from your program.** Point UMD at the server *directory* (the `SOCKET`'s parent above,
+   e.g. `/tmp/tt-umd-sim-server-0`) instead of at a simulator. UMD sees the live sockets, attaches as
+   a client, and from then on you drive the cluster exactly as you would real hardware. There is no
+   separate "connect" call — the path you pass is what makes you a client.
+
+   Wherever you would name a `libttsim.so` — `TT_UMD_SIMULATOR`, `TT_METAL_SIMULATOR`, or
+   `ClusterOptions::simulator_directory` — name the server directory instead. So an existing test runs
+   against a server with no code change:
+
+   ```
+   TT_UMD_SIMULATOR=/tmp/tt-umd-sim-server-0 \
+     ./build/test/umd/api/api_tests --gtest_filter="TestDeviceIOFixture.RegReadWrite"
+   ```
+
+   There is no Python entry point today.
 
 4. **Stop the server.**
 
@@ -208,10 +233,11 @@ client — there is no separate "connect" call.
 
 Each simulated chip is exposed as a socket in its server's directory. The host answers device
 operations and describes the cluster topology over that socket; a client forwards its operations to
-the host and applies the replies, so on the client side the device behaves like any other. Stopping a
-server tears it down cleanly and closes its client connections, so a client's in-flight or next
-operation fails with a clear "server stopped" error instead of hanging — letting the client notice
-and exit gracefully.
+the host and applies the replies, so on the client side the device behaves like any other.
+
+Stopping a server does **not** stop its clients — it closes their connections, so a client's next
+operation fails with a clear "server stopped" error instead of hanging, and the client decides what to
+do. Each process is started and stopped on its own.
 
 ## Good to know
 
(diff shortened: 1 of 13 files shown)


docs/SIMULATION_SERVER.md
```diff
@@ -0,0 +1,202 @@
...
+options.chip_type = ChipType::SIMULATION;
+options.simulator_directory = "/tmp/tt-umd-sim-server-0";  // a server's directory (from `list`)
+Cluster cluster(options);
+
+// From here the cluster behaves like any other -- the same code runs against
+// hardware or against a shared simulation.
+for (ChipId chip : cluster.get_target_device_ids()) {
+    uint32_t value = 0;
+    cluster.read_from_device(&value, chip, core, addr, sizeof(value));
+}
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed. The `Cluster` snippet now defines both: `core` is taken from the chip's SoC descriptor (first Tensix core) and `addr` is a literal `0x1000`, and the example includes `umd/device/types/core_coordinates.hpp` for `CoreCoord`/`CoreType`, so it's copy-paste self-contained.

docs/SIMULATION_SERVER.md
```diff
@@ -0,0 +1,224 @@
+# Simulation Server
+
+UMD can run against a *simulated* device instead of real hardware. The **simulation server** lets one
```
broskoTT: link tt sim here

docs/SIMULATION_SERVER.md
```diff
@@ -0,0 +1,224 @@
...
+process **host** a simulated device and have other UMD processes **attach** to it and drive it exactly
+as they would a real cluster. That way a single running simulation can be shared by several processes
```
broskoTT: This lest sentence holds 0 additional information

docs/SIMULATION_SERVER.md
```diff
@@ -0,0 +1,224 @@
...
+   Lists the open servers. Each row is one chip of one server: the server index, the chip id,
+   whether it is reachable, its arch/backend, and the socket it is served on.
+
+   ```
+   SERVER   CHIP   STATE   ARCH             SOCKET
+   0        0      live    blackhole/ttsim  /tmp/tt-umd-sim-server-0/tt-umd-sim-0.sock
+   1        0      live    blackhole/ttsim  /tmp/tt-umd-sim-server-1/tt-umd-sim-0.sock
+   ```
+
+3. **Use it from your program.** Open a UMD cluster in simulation mode pointed at a server's socket
```
broskoTT: It is unclear how to do it. How do i run any unit test on a simulator? Or from python? Should I set an env variable?

docs/SIMULATION_SERVER.md
```diff
@@ -0,0 +1,224 @@
...
+    k1(["tt-umd-sim-0.sock"])
+    h1 --- k1
+  end
+  ca["client"] -->|attach| k0
+  cb["client"] -->|attach| k1
+```
+
+## The flow
+
+1. **Start a server.** Launch a host in the background with the `sim_server.sh` wrapper:
```
broskoTT, line 71: Is this the main entry point? I'd change the instruction to be other way around: UMD backend is the main thing, it starts the simulation server. So you run anything spinning up UMD backend and the server will start. The sim_server.sh is a secondhand citizen, it is just a helper for managing server/process lifetime

docs/SIMULATION_SERVER.md
```diff
@@ -0,0 +1,224 @@
...
+  Tool->>Host: start a simulator, in the background
+  Host->>Dir: create dir + bind tt-umd-sim-0.sock
+  Host-->>Tool: "server directory: ..." once serving
+  Note over Client,Dir: later, a separate process
+  Client->>Dir: open the sockets in the directory
+  Client->>Host: attach, then read / write
+  Host-->>Client: replies (device memory, topology)
+  Tool->>Host: kill the server (SHUTDOWN over socket)
+  Host->>Dir: remove sockets + directory
+  Host--xClient: closed → "server stopped"
```
broskoTT, line 155: hmmm, this implies that the host process can control the lifetime of the client's process, through simulation backend's lifetime? I think I might be confusing some things, can the SimulationServer be started through metal unit test (so just a client from these diagrams)?

docs/SIMULATION_SERVER.md
```diff
@@ -0,0 +1,224 @@
...
+- **Server directories.** One per running server, under the system temporary directory (`/tmp` on
+  Linux), named `tt-umd-sim-server-<index>`. The index counts up from 0; each `start` claims the
+  lowest free one.
+- **Sockets.** One per chip, inside that server's directory, named `tt-umd-sim-<chip_id>.sock`. A
+  server directory *is* what a client points at, and what `sim_server list` scans — there is no
+  central registry, just the directories present on disk. When a server shuts down it removes its
+  sockets and its (now-empty) directory.
+- **Server logs.** Started through `sim_server.sh`, a host is detached from your terminal and its
+  output goes to a per-server log in the temporary directory, named after the server directory:
+  `sim_server-tt-umd-sim-server-<index>.log`. Check it if a server did not come up. Started directly
```
broskoTT, line 229: not sure this is true for the logs

docs/SIMULATION_SERVER.md
```diff
@@ -0,0 +1,224 @@
...
+server tears it down cleanly and closes its client connections, so a client's in-flight or next
+operation fails with a clear "server stopped" error instead of hanging — letting the client notice
+and exit gracefully.
+
+## Good to know
+
+- **Multiple servers coexist.** Each host has its own directory, so you can run several simulators on
+  one machine at once — including two of the same chip id. Point each client at the server directory
+  it wants (from `sim_server list`).
+- **Clients fail cleanly.** If the server goes away, a client's next operation raises a clear error
```
broskoTT, line 247: Yeah this explains some of the questions I raised in some of the previous comments, still don't know how this is true

broskoTT (approved):

docs/SIMULATION_SERVER.md
nbuncicTT: Added — the intro now links [ttsim](https://github.com/tenstorrent/ttsim) and mentions the RTL simulator alternative.

docs/SIMULATION_SERVER.md
nbuncicTT: Replaced with the part that actually adds something: without a server each process gets its own independent simulation, so two processes never see each other's writes.

docs/SIMULATION_SERVER.md
nbuncicTT: Made it concrete: wherever you would name a `libttsim.so` — `TT_UMD_SIMULATOR`, `TT_METAL_SIMULATOR`, or `ClusterOptions::simulator_directory` — name the server directory instead, and an existing test attaches with no code change. Verified against a running server:

```
TT_UMD_SIMULATOR=/tmp/tt-umd-sim-server-0 ./api_tests --gtest_filter="TestDeviceIOFixture.RegReadWrite"   # passes
```

No Python entry point today; that is now stated in the doc.

docs/SIMULATION_SERVER.md
nbuncicTT: Restructured along those lines, with one correction: it is not automatic. Serving is opt-in via `ClusterOptions::serve_simulation_devices_over_sockets`, **default false** since #3112 — so a UMD or tt-metal test pointed at a `libttsim.so` hosts a *private* in-process simulation and publishes nothing. Added an "Any UMD process can be the host" section saying hosting is a `Cluster` flag and `sim_server` is just a host that sets it and stays alive.

docs/SIMULATION_SERVER.md
nbuncicTT: It does not — clarified. Stopping a server closes its clients' connections, so a client's next operation fails with a clear error and the client decides what to do; each process is started and stopped independently. And yes, a metal/UMD test can be the host, but only if it sets `serve_simulation_devices_over_sockets` — by default it is a private in-process simulation, which is the same point as the comment above.

docs/SIMULATION_SERVER.md
nbuncicTT: You were right about the old text. Rewritten and verified: through `sim_server.sh` the log is `/tmp/sim_server-tt-umd-sim-server-<index>.log`, named by server index rather than pid; run directly, `sim_server start` is foreground and logs to your terminal.

docs/SIMULATION_SERVER.md
nbuncicTT: Added the evidence: the sockets are created world-writable (`srw-rw-rw-`), so any user on the machine can attach — which is also why `kill` goes over the socket rather than by signal.

PR #3055, opened 2026-07-20
Use kmdlib for querying mappings

aleksamarkovicTT:
### Issue
#2749 

### Description
Use kmdlib for querying BAR mappings in PciDevice.

### List of the changes
- Add tt_device_query_bar_mappings() to kmdlib.
- Remove ioctl calls in PciDevice related to querying mappings.
### Testing
CI

### API Changes
This PR has API changes.


device/api/umd/device/tt_kmd_lib/tt_kmd_lib.h
```diff
@@ -181,6 +181,61 @@ int tt_device_get_attrs(tt_device_t* dev, tt_device_attrs_t* out_attrs);
...
+ *
+ * Retrieves the base offsets and sizes of the device's memory-mapped BAR
+ * resources. The returned `base` values are suitable as offsets to mmap() on
+ * the device file descriptor.
+ *
+ * @param dev Device handle
+ * @param out_mappings Populated with the device's BAR mappings on success
+ * @return 0 on success, error code on failure
+ */
+int tt_device_query_bar_mappings(tt_device_t* dev, tt_bar_mappings_t* out_mappings);
```
pjanevskiTT, line 237: could we maybe have a signature of this function be something like

> int ...(tt_device_t* dev, tt_bar_mapping_t* mappings, int num_mappings)

in order to make it a bit more general for future use as well. For different architectures we may not have the same amount of BARs. Or if there is some fixed number of BARs that linux supports maybe we should limit ourselves by that?

aleksamarkovicTT: The PCI spec defines 6x32bit BAR or 3x64bit BAR mappings. KMD introduces the UC/WC distinction. I think it's better to have it be an array though.

broskoTT: Let's not over engineer, if we have a case for it today, I'm fine with making it more complex, otherwise, not

broskoTT: Wait, was the suggestion to control how many bars are to be mapped? Or was "num_mappings" supposed to be an output parameter saying how many bars are available for this architecture?

For the former, it could sound useful, but let's not spend too much time on it if we don't have a usecase today. For the latter, the number of bars available could be part of tt_bar_mappings_t directly. Then it would probably make sense to have its contents an array as Aleksa mentioned

broskoTT (approved):

PR #3057, opened 2026-07-21
Re-enable KMD power mode awareness in UMD

nbuncicTT:
### Issue
#2531 

### Description
Restore the power-aware device open path that PR #2512 disabled while investigating the NOC1 issue in #2531:

### List of the changes
- pci_device.cpp: open the device with O_APPEND on Blackhole with KMD >= 2.6.0, and pass O_APPEND as extra_flags to tt_device_open.
- topology_discovery.cpp: call set_power_state(true) when not in low power mode.
- nanobind: drop the now-stale "temporarily disabled" comment.

### Testing
Purpose: run the test suite many times on Blackhole CI to check whether the NOC1 hang from #2531 still reproduces: https://github.com/tenstorrent/tt-umd/actions/runs/29589352072

Manual + CI

### API Changes
/


pjanevskiTT (approved):
approved if we don't see the problem anymore

PR #3060, opened 2026-07-21
Superset benchmark upload

### LINKED ISSUE #2154: Ingestion perf to superset
Aleksa already talked about this with Vladimir Milosevic to setup an instance for UMD so we can analyze perf

ajovanovicTT:
### Issue
#2154

### Description
Wires UMD microbenchmarks to Superset via SFTP ingest (DATA-1567).

### List of the changes
 - New emit_benchmark_export.py folds nanobench results + host spec into a validated Superset export JSON
 - run-benchmarks.yml emits and uploads the export JSON as an artifact
 - New upload-benchmark-superset.yml SFTPs export artifacts to the ingest bucket, no-ops if secret unset
 - build-and-run-all-benchmarks.yml adds the upload-superset job

### Testing
CI.

### API Changes
This PR has no API changes.


ajovanovicTT:
Successfully uploaded benchmark data over sftp, visible in job https://github.com/tenstorrent/tt-umd/actions/runs/29859827443/job/88738438110

ajovanovicTT:
Verified data is uploaded from CI run and visible in superset, example chart : https://superset.tenstorrent.com/explore/?form_data_key=Jc408XQl5sM&slice_id=7809 (only two data points uploaded at this moment).

tests/microbenchmark/tools/emit_benchmark_export.py
```diff
@@ -0,0 +1,312 @@
...
+    except (TypeError, ValueError):
+        return None
+    return f if math.isfinite(f) else None
+
+
+def derive_arch(card):
+    """Silicon arch from the board/card label (n*/sim-wh -> wormhole_b0, p*/sim-bh -> blackhole)."""
+    if not card:
+        return None
+    c = card.lower()
```
aleksamarkovicTT: Why not modify gather_host_specs.py instead of hacking the strings like this?

tests/microbenchmark/tools/emit_benchmark_export.py
```diff
@@ -0,0 +1,312 @@
...
+    commit_timedate = _git(repo_dir, "show", "-s", "--format=%cI", commit_sha or "HEAD")
+
+    pipeline_id = os.environ.get("GITHUB_RUN_ID")
+    try:
+        pipeline_id = int(pipeline_id) if pipeline_id else None
+    except ValueError:
+        pipeline_id = None
+
+    run_ts = (
+        os.environ.get("RUN_TS")
```
aleksamarkovicTT, line 87: Where is RUN_TS set?

tests/microbenchmark/tools/emit_benchmark_export.py
```diff
@@ -0,0 +1,312 @@
...
+            continue
+        try:
+            data = load_nanobench_json(path)
+        except (OSError, json.JSONDecodeError) as e:
+            print(f"WARN: skipping {path.name}: {e}", file=sys.stderr)
+            continue
+        for nb_result in data.get("results", []):
+            record = build_result(nb_result)
+            missing = [k for k in _REQUIRED_RESULT_NUMERICS if record[k] is None]
+            if missing:
```
aleksamarkovicTT: Any reason not to fail instead of warning here? When would this be acceptable? we expect nanobench to always have these fields.

tests/microbenchmark/tools/emit_benchmark_export.py
ajovanovicTT: Removed this hacking and updated logic in gather_host_specs.

tests/microbenchmark/tools/emit_benchmark_export.py
ajovanovicTT: Wired RUN_TS setting at the begging of run-benchmarks workflow.

tests/microbenchmark/tools/emit_benchmark_export.py
ajovanovicTT: Yeah, even in some malformed case run, some bench data would appear for all this fields in that case. Removed this redundant check, left only validation against pydantic model, and made it mandatory, so any bad data in json would fail there.

aleksamarkovicTT (approved):

PR #3063, opened 2026-07-22
Add support for exporting TLB windows as dma-buf for peer-to-peer DMA

### LINKED ISSUE #2987: Integrate TENSTORRENT_IOCTL_EXPORT_TLB_DMABUF for RDMA / peer-to-peer DMA
## Summary

tt-kmd (as of `ttkmd-2.10.0-rc1`) added `TENSTORRENT_IOCTL_EXPORT_TLB_DMABUF`. It wraps an already-allocated, already-configured TLB window in a dma-buf and returns a file descriptor. That fd can be handed to another device's driver, e.g. an RDMA NIC via `ibv_reg_dmabuf_mr()`, so the peer does peer-to-peer PCIe DMA directly into and out of the window. Traffic routes onto the NOC according to the window's `CONFIGURE_TLB` setup, with no Tenstorrent-CPU involvement on the data path.

We would like UMD to expose this so callers can export a NOC-addressable region as a dma-buf without hand-rolling ioctls.

## Kernel API

```c
#define TENSTORRENT_IOCTL_EXPORT_TLB_DMABUF  _IO(TENSTORRENT_IOCTL_MAGIC, 16)

struct tenstorrent_export_tlb_dmabuf {
    __u32 argsz;    // = sizeof(struct)
    __u32 flags;    // reserved, must be 0
    __u32 tlb_id;   // IN:  id from TENSTORRENT_IOCTL_ALLOCATE_TLB
    __s32 fd;       // OUT: the dma-buf fd
    __u64 offset;   // IN:  page-aligned byte offset within the window
    __u64 size;     // IN:  page-aligned; 0 means to end of window
};
```

Usage sequence (see the [reference tool](https://github.com/tenstorrent/tt-kmd/blob/8c32c2b03aa3359ed6acf929f920e472197e8280/tools/rdma/nic_bh_p2p_dma.c)): `ALLOCATE_TLB`, `CONFIGURE_TLB`, `EXPORT_TLB_DMABUF`, hand the fd to the importer, then `FREE_TLB` on teardown.

## Semantics UMD must respect

- Ownership: the export must be issued on the same chardev fd that allocated the window (kmd checks the per-fd bitmap).
- Alignment: `offset` and `size` must be page-aligned and lie within the window.
- The fd is the unit of ownership and is self-sufficient. An export pins the window and the device by refcount, so it survives `FREE_TLB`, `close()` of the owning fd, and even device removal. The window is returned to the pool only once the last export is released. UMD needs no RAII object to keep the export valid.
- `RESET_DEVICE` (destructive flavors) is refused with `-EBUSY` while any export is live.
- The ioctl requires Linux 5.8+.

## Proposed UMD surface

How to integrate this is up to the UMD team. One shape that fits the existing tt-umd TLB stack is a single public entry point that always allocates a dedicated window, so there is no shared-window reconfigure hazard.

```cpp
// Allocate a fresh TLB window aimed at (chip, core, addr), export it as a
// dma-buf for peer-to-peer DMA, and return an fd the caller owns. The caller
// must close() the fd when done; that releases the export and frees the window.
int Cluster::export_dmabuf(ChipId chip, CoreCoord core, uint64_t addr,
                           size_t size, uint64_t ordering = tlb_data::Relaxed);
```

Plumbing (bottom to top): a `tt_kmd_lib` C shim (`tt_tlb_export_dmabuf`), a `TlbHandle::export_dmabuf` virtual implemented by `SiliconTlbHandle` (the one class holding both the fd and the tlb id) and stubbed as unsupported in the sim handles, a thin `TlbWindow::export_dmabuf`, and the `Cluster` entry point above. Internally it configures a fresh window, exports it, and drops the handle immediately; `FREE_TLB` runs at once while the kmd keeps the window pinned for the fd's life.

ajovanovicTT:
### Issue
#2987 

### Description
Adds UMD support for exporting a NOC addressable region as a Linux dma-buf file descriptor for peer to peer PCIe DMA (for example importing into an RDMA NIC via ibv_reg_dmabuf_mr). This wraps the new tt-kmd ioctl TENSTORRENT_IOCTL_EXPORT_TLB_DMABUF, available from tt-kmd 2.10.0-rc1 onward, and additionally requires a Linux 5.8+ kernel.

### List of the changes
 - Added TENSTORRENT_IOCTL_EXPORT_TLB_DMABUF and its struct to the vendored kmd ioctl header
 - Added tt_tlb_export_dmabuf C shim in tt_kmd_lib
 - Added KMD_TLB_DMABUF_EXPORT version constant and PCIDevice::is_tlb_dmabuf_export_supported
 - Added PCIDevice::export_tlb_dmabuf, which runs the driver documented sequence directly (allocate TLB, configure via CONFIGURE_TLB, export, free), independent of the TlbHandle/TlbWindow classes since an exported window is never read or written by the host
 - Added PcieProtocol::export_dmabuf (declared on the new PcieInterface), which validates the request (non-zero and page-aligned address and size), picks the smallest per-arch TLB window size class that covers the requested range, and builds the TLB configuration
 - Added TTDevice::export_dmabuf, which resolves the coordinate, rejects remote devices, and forwards to the PCIe interface
 - Added LocalChip::export_dmabuf and Cluster::export_dmabuf as thin propagators. export_dmabuf is intentionally only on LocalChip, not the Chip base class, since the export wraps a PCIe aperture that a remote chip does not have; Cluster::export_dmabuf goes through get_local_chip, which already rejects a non-local chip id
 - Added a host wide check, has_any_active_rdma_port in device/common/utils.hpp, which scans /sys/class/infiniband for a port in the ACTIVE state. PcieProtocol::export_dmabuf asserts this before allocating any window, since there is nothing on the host that could consume the fd without an active RDMA NIC

### Testing
 - Using TEMP test from rdma_testing/* (paired sources run at the same time on bh-glx6u-18 and bh-glx6u-22) which are on branch ajovanovic/rdma_testing_tools
 - CI

### API Changes
This PR has API changes. New public methods: Cluster::export_dmabuf, TTDevice::export_dmabuf, LocalChip::export_dmabuf, PCIDevice::export_tlb_dmabuf, PCIDevice::is_tlb_dmabuf_export_supported and PCIDevice::read_kernel_version, and a new pure virtual PcieInterface::export_dmabuf implemented by PcieProtocol.


ajovanovicTT:
Example run on configured machines:

Screen Shot 2026-07-28 at 11 54 50 AM
Screen Shot 2026-07-28 at 11 54 26 AM

device/api/umd/device/pcie/pci_device.hpp
```diff
@@ -273,6 +273,12 @@ class PCIDevice {
      */
     static SemVer read_kmd_version();
 
+    /**
+     * Read the running Linux kernel version (from uname(2)'s release field), as major.minor.patch.
+     * The trailing distro suffix (e.g. "-91-generic") is dropped; only major/minor/patch are parsed.
+     */
+    static SemVer read_kernel_version();
```
pjanevskiTT, line 280: I think it makes sense to be added as separate PR, this can be useful in multiple use cases

pjanevskiTT (commented):
will continue with comments

device/pcie/tlb_window.cpp
```diff
@@ -127,6 +127,8 @@ uint64_t TlbWindow::get_base_address() const {
     return handle_ref().get_config().local_offset + offset_from_aligned_addr;
 }
 
+int TlbWindow::export_dmabuf(uint64_t offset, uint64_t size) const { return tlb_handle->export_dmabuf(offset, size); }
```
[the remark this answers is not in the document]
ajovanovicTT: Applied, since tlb window is allocated to aligned address, exported offset must be set to offset_from_aligned_address.

ajovanovicTT:
### Reading over RDMA on the same host (loopback):
Screen Shot 2026-08-04 at 10 53 51 AM

ajovanovicTT:
On x8 lanes connected chip, hitting Ethernet link limit :
image

device/api/umd/device/pcie/pci_device.hpp
nbuncicTT: Agreed, please make separate PRs for adding general things like reading the kernel version, etc.

device/api/umd/device/pcie/pci_device.hpp
```diff
@@ -326,6 +332,32 @@ class PCIDevice {
      */
     static bool is_arch_agnostic_reset_supported();
 
+    /**
+     * Checks if exporting a TLB window as a dma-buf (TENSTORRENT_IOCTL_EXPORT_TLB_DMABUF) is
```
nbuncicTT, line 336: Edit this comment to not mention the IOCTL, as it's an implementation detail of the tt-kmb-lib.

nbuncicTT (commented):
Please separate this PR into small self-contained ones along these lines:
- tt kmd lib changes
- PCIDevice changes
- TTDevice changes
- Cluster/Chip changes

device/api/umd/device/pcie/pci_device.hpp
ajovanovicTT: Edited in newly opened PR : https://github.com/tenstorrent/tt-umd/pull/3142.

device/api/umd/device/pcie/pci_device.hpp
ajovanovicTT: #3145 Opened new PR for this.

PR #3067, opened 2026-07-23
Pin TLB static_vc class/buddy by direction on Blackhole

### LINKED ISSUE #2735: BH IO ordering/membar
**Describe the bug**
We potentially don't guarantee IO ordering on BH, effectivelly our membar doesn't work, and this can lead to many different issues. This might be related to other bugs already reported in UMD.

This draws roots from HW issue where using same static VC for both reads and writes through the same TLB exhibits a HW bug and crashes the host.
We solved that (ugly) by using dynamic VC for all traffic from host.

One potential solution is to tie TLBs to be configured only for reads or writes.
We shouldn't have a problem in single threaded use (to be confirmed).
We already have Strict TLB ordering everywhere, that shouldn't be the problem. 

**Stack trace**
General issue

pjanevskiTT:
### Issue
Closes #2735

### Description
**Goal: bring Blackhole to parity with Wormhole so the memory barrier works on Blackhole too.** The membar — and any ordering-dependent host→device sequence — needs writes to a tile to stay ordered on the NoC. Wormhole gets this for free: its PCIe tile hardwires static VC with reads and writes on separate virtual channels, and the barrier works there. Blackhole runs its cached host TLBs on dynamic VC, where the NoC can reorder two writes to the same tile once they leave the PCIe tile — the suspected root of the Blackhole ordering/membar issues in #2735.

Static VC was left off on Blackhole because turning it on crashed the host: a single static VC carrying both reads and writes hits a hardware bug.

This PR makes Blackhole mimic Wormhole: pin the static VC by direction — writes on buddy 0, reads on buddy 1, multicast writes on class `0b10` — so reads and writes never share a VC. That dodges the crash while restoring the same write ordering Wormhole has, so the barrier behaves on Blackhole as it does on Wormhole. These are exactly the values the Wormhole PCIe tile hardwires.

**Scope:** only the windows that are reconfigured per op and can therefore commit to a direction — the cached data, register and DMA windows. A window that stays configured across both reads and writes asks for `BIDIRECTIONAL` and keeps the architecture default, so the persistent-TLB path remains on dynamic VC on Blackhole and cannot re-trigger the crash. Wormhole hardwires buddy/class in hardware, so only `static_vc` is set there; the TTSim/RTL-sim backends do not model static VC, so those fields are stripped in their TLB handles.

### List of the changes
- Add `static_vc_buddy`/`static_vc_class` to `tlb_data` and the Blackhole/Grendel `tlb_offsets`, and pack them in `apply_offset`/`check`.
- Add a `TlbVcDirection` enum and `ArchitectureTlbs::get_static_vc()`, which returns the static VC settings for an architecture and direction.
- Route the cached data (`make_tlb_config`), register and DMA paths through it by direction; persistent and dma-buf-export windows keep the architecture default.
- Strip the buddy/class fields in the TTSim/RTL-sim TLB handles, which do not model static VC.

### Testing
CI

### API Changes
Additive only: new `static_vc_buddy`/`static_vc_class` fields and a `set_static_vc()` setter on `tlb_data`, a new `TlbVcDirection` enum and `tlb_static_vc` struct in `types/tlb.hpp`, and a `static_vc_is_configurable` field plus `get_static_vc()` on `ArchitectureTlbs`.

### DIFF
--- a/device/api/umd/device/arch/blackhole_implementation.hpp
+++ b/device/api/umd/device/arch/blackhole_implementation.hpp
@@ -38,7 +38,10 @@ inline constexpr auto TLB_2M_OFFSET = tlb_offsets{
     .linked = 72,
     .static_vc = 73,
     // missing .stream_header
-    .static_vc_end = 75};
+    .static_vc_end = 75,
+    .static_vc_buddy = 75,
+    .static_vc_class = 76,
+    .static_vc_class_end = 78};
 
 inline constexpr auto TLB_4G_OFFSET = tlb_offsets{
     .local_offset = 0,
@@ -52,7 +55,10 @@ inline constexpr auto TLB_4G_OFFSET = tlb_offsets{
     .linked = 61,
     .static_vc = 62,
     // missing .stream_header
-    .static_vc_end = 64};
+    .static_vc_end = 64,
+    .static_vc_buddy = 64,
+    .static_vc_class = 65,
+    .static_vc_class_end = 67};
 
 enum class arc_message_type {
     NOP = 0x11,  // Do nothing
@@ -499,7 +505,10 @@ public:
 
     size_t get_cached_tlb_size() const override { return blackhole::STATIC_TLB_SIZE; }
 
-    bool get_static_vc() const override { return false; }  // False due to a known HW issue.
+    // Gates static VC on the persistent-TLB path (configure_tlb), where a single window services both
+    // reads and writes -- a shared static VC there crashes the host. The cached data/register/DMA
+    // paths reconfigure per op and instead pin the static VC by direction via set_static_vc().
+    bool get_static_vc() const override { return false; }
 
     std::optional<uint8_t> get_ubb_tray_id(uint16_t bus_id) const override {
         const uint16_t bus_high = static_cast<uint16_t>(bus_id & 0xF0);
--- a/device/api/umd/device/arch/blackhole_implementation.hpp
+++ b/device/api/umd/device/arch/blackhole_implementation.hpp
@@ -38,7 +38,9 @@ inline constexpr auto TLB_2M_OFFSET = tlb_offsets{
     .linked = 72,
     .static_vc = 73,
     // missing .stream_header
-    .static_vc_end = 75};
+    .static_vc_buddy = 75,
+    .static_vc_class = 76,
+    .static_vc_class_end = 78};
 
 inline constexpr auto TLB_4G_OFFSET = tlb_offsets{
     .local_offset = 0,
@@ -52,7 +54,9 @@ inline constexpr auto TLB_4G_OFFSET = tlb_offsets{
     .linked = 61,
     .static_vc = 62,
     // missing .stream_header
-    .static_vc_end = 64};
+    .static_vc_buddy = 64,
+    .static_vc_class = 65,
+    .static_vc_class_end = 67};
 
 enum class arc_message_type {
     NOP = 0x11,  // Do nothing
--- a/device/api/umd/device/types/tlb.hpp
+++ b/device/api/umd/device/types/tlb.hpp
@@ -9,6 +9,8 @@
 #include <stdexcept>
 #include <utility>
 
+#include "umd/device/types/arch.hpp"
+
 namespace tt::umd {
 
 struct tlb_offsets {
@@ -23,6 +25,21 @@ struct tlb_offsets {
     uint32_t linked;
     uint32_t static_vc;
     uint32_t static_vc_end;
+    // Per-TLB pin of the NoC virtual channel buddy bit and class bits, used when static_vc is set.
+    // Only defined on architectures that expose them as configuration bits (Blackhole/Grendel).
+    // Wormhole's PCIe tile hardwires the equivalent in hardware regardless of config, so these
+    // offsets are left unset (0) there and no buddy/class bits are packed.
+    uint32_t static_vc_buddy;
+    uint32_t static_vc_class;
+    uint32_t static_vc_class_end;
+};
+
+// Direction of a TLB transaction, used to pin the static VC so reads and writes never share a
+// virtual channel (a single static VC carrying both crashes the host on Blackhole).
+enum class TlbVcDirection {
+    UnicastWrite,
+    UnicastRead,
+    MulticastWrite,
 };
 
 struct tlb_data {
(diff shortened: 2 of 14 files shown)


device/api/umd/device/types/tlb.hpp
```diff
@@ -23,6 +25,21 @@ struct tlb_offsets {
     uint32_t linked;
     uint32_t static_vc;
     uint32_t static_vc_end;
+    // Per-TLB pin of the NoC virtual channel buddy bit and class bits, used when static_vc is set.
```
nbuncicTT: omit this comment, as there are no comments for other fields.

device/api/umd/device/types/tlb.hpp
```diff
@@ -46,6 +65,13 @@ struct tlb_data {
     std::pair<std::uint64_t, std::uint64_t> apply_offset(const tlb_offsets &offset) const;
 };
 
+// Populates the static VC fields of `config` for the given architecture and transaction direction.
+// Blackhole/Grendel expose buddy/class as config bits, so they are pinned by direction (write=buddy0,
+// read=buddy1, multicast=class 0b10, unicast=class 0b00) to keep reads and writes on separate static
```
nbuncicTT: The enums precisely exist to omit the implementation detail of what are the values behind (e.g. the classes). Shorten the comments so it describes the API from a higher level without going into impl. details.

device/pcie/tlb_window.cpp
```diff
@@ -25,14 +25,20 @@ TlbWindow::TlbWindow(std::unique_ptr<TlbHandle> handle, const tlb_data config) :
...
+    bool mcast,
+    tt_xy_pair core_start) const {
     tlb_data config{};
     config.local_offset = addr;
     config.x_end = core_end.x;
     config.y_end = core_end.y;
     config.noc_sel = static_cast<uint64_t>(noc_id);
     config.ordering = ordering;
-    config.static_vc = handle_ref().get_arch() != tt::ARCH::BLACKHOLE;
+    set_static_vc(config, handle_ref().get_arch(), direction);
```
nbuncicTT: I'd prefer an API with a return value for the `static_vc` and not an output parameter which can modify the whole struct and not one field.

Something like:
`config.static_vc = get_static_vc(handle_ref().get_arch(), direction);`

I'm aware `get_static_vc` is already taken as a name, so pick some other name like `pick_static_vc` or something like that.

device/types/tlb.cpp
```diff
@@ -8,10 +8,32 @@
 #include <string>
 #include <utility>
 
+#include "umd/device/types/arch.hpp"
 #include "umd/device/utils/error.hpp"
 
 namespace tt::umd {
 
+void set_static_vc(tlb_data &config, tt::ARCH arch, TlbVcDirection direction) {
```
nbuncicTT: Not sure if this makes sense to put into architecture_implementation (if it's modified to return the `static_vc` like I said with the `pick_static_vc`/`get_static_vc` proposal in a previous comment).

nbuncicTT (approved):

device/api/umd/device/arch/blackhole_implementation.hpp
```diff
@@ -38,7 +38,10 @@ inline constexpr auto TLB_2M_OFFSET = tlb_offsets{
     .linked = 72,
     .static_vc = 73,
     // missing .stream_header
-    .static_vc_end = 75};
+    .static_vc_end = 75,
```
nbuncicTT: `static_vc_end` is redundant wherever buddy/class exist, since check() could derive `static_vc`'s width from `static_vc_buddy`'s start the way it does for every other field, and it survives only because Wormhole declares no buddy/class and needs it as a terminator.

Remove .`static_vc_end` because it overlaps with .`static_vc_buddy` in BH/Grendel, leave it in WH.

device/types/tlb.cpp
```diff
@@ -8,10 +8,32 @@
...
+        case tt::ARCH::WORMHOLE_B0:
+            // The Wormhole PCIe tile hardwires buddy/class (buddy=0 writes, buddy=1 reads,
+            // class 0b10 multicast, 0b00 unicast) regardless of config, so only enable static_vc.
+            config.static_vc = 1;
+            break;
+        default:
+            config.static_vc = 0;
+            break;
+    }
+}
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed structurally. The policy is now `ArchitectureTlbs::get_static_vc(TlbVcDirection)`, which returns a fully-populated `tlb_static_vc` for every architecture (Wormhole returns buddy=0/class=0 explicitly rather than leaving them untouched), and `tlb_data::set_static_vc()` assigns all three fields. A reused config can no longer carry stale buddy/class bits into `apply_offset()`.

device/api/umd/device/types/tlb.hpp
pjanevskiTT: Removed.

device/api/umd/device/types/tlb.hpp
pjanevskiTT: Reworded. The enum comment now only says what a window carries and that a single-direction window can be given a channel of its own; the buddy/class encodings live inside `ArchitectureTlbs::get_static_vc()` and nowhere in the header.

device/pcie/tlb_window.cpp
pjanevskiTT: Changed to a return value:

```cpp
config.set_static_vc(get_architecture_tlbs(handle_ref().get_arch()).get_static_vc(direction));
```

One caveat on the shape you suggested: the policy sets three coupled fields (`static_vc`, `static_vc_buddy`, `static_vc_class`), so a single scalar return would not cover it. `get_static_vc()` returns a small `tlb_static_vc` value instead, and `tlb_data::set_static_vc()` applies it as a unit. No out-param.

device/types/tlb.cpp
pjanevskiTT: Done, though the destination moved: `architecture_implementation::get_static_vc()` no longer exists on main, TLB policy now lives in `ArchitectureTlbs`. So this went to `ArchitectureTlbs::get_static_vc(TlbVcDirection)` in `arch/architecture_tlbs.{hpp,cpp}`, right next to `use_static_vc` and gated by a new `static_vc_is_configurable` flag. `types/tlb.cpp` no longer references `tt::ARCH` at all now.

device/api/umd/device/arch/blackhole_implementation.hpp
pjanevskiTT: Done. `.static_vc_end` is gone from the Blackhole and Grendel offset tables and kept on Wormhole. `check()` derives static_vc's terminator from `static_vc_buddy` when it is set, and falls back to `static_vc_end` otherwise.

PR #3074, opened 2026-07-24
Modify check so that it doesn't fail CI for ThermTrip

nbuncicTT:
### Issue
/

### Description
Test modification for CI because off false positivies

### List of the changes
Changed test expectations

### Testing
CI

### API Changes
/

### DIFF
--- a/tests/api/test_firmware_info_provider.cpp
+++ b/tests/api/test_firmware_info_provider.cpp
@@ -622,8 +622,6 @@ TEST_F(TestFirmwareInfoProvider, ThermTripCount) {
         EXPECT_TRUE(trip_count.has_value());
         if (trip_count.has_value()) {
             log_info(tt::LogUMD, "therm_trip_count={}", trip_count.value());
-        } else {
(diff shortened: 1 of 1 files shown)


tests/api/test_firmware_info_provider.cpp
```diff
@@ -610,10 +610,20 @@ TEST_F(TestFirmwareInfoProvider, ThermTripCount) {
...
+            EXPECT_FALSE(trip_count.has_value());
+            continue;
+        }
+
+        EXPECT_TRUE(trip_count.has_value());
         if (trip_count.has_value()) {
-            EXPECT_EQ(trip_count.value(), 0u);
+            log_info(tt::LogUMD, "therm_trip_count={}", trip_count.value());
+        } else {
+            log_warning(tt::LogUMD, "therm_trip_count not available");
```
pjanevskiTT: nit: maybe this log is not needed, now the test will fail if it's not available, that should be good enough to know it's not available

pjanevskiTT (approved):

aleksamarkovicTT:
Issue #3079

PR #3076, opened 2026-07-25
SWEmuleChip: wire up a real SimulationSysmemManager for host-facing addressing

arminaleTT:
### Issue
Part of the tt-emule software-emulator host-facing socket bring-up (companions: https://github.com/tenstorrent/tt-metal/pull/51108). No standalone tt-umd issue.

### Description
Upstream (mergeable) edition — the same change as the staging PR #3075, cherry-picked cleanly onto umd `main`. Wires `SWEmuleChip` up to a real `SimulationSysmemManager` so a device-IO address at/above `pcie_base` resolves to a host-mapped buffer — the substrate for the emulator's host-fed socket path.

**No impact on silicon code paths. This is for emulation support only.**

### List of the changes
- `SWEmuleChip` owns a `SimulationSysmemManager`; `get_sysmem_manager()` returns it (was `nullptr`).
- New `SimulationSysmemManager::get_mapped_host_ptr(device_io_addr, size)` zero-copy lookup, with its
  lifetime contract documented (valid only while the buffer stays mapped; lock guards the lookup).
- Constructor assert `dram_bank_size_ < get_pcie_base_for_arch(arch)`.
- New `api_tests` case `GetMappedHostPtrResolvesOffsetAndMiss`.

### Testing
tt-emule tests and the new `get_mapped_host_ptr` case. The socket runtime is built/verified on the staging edition (#3075).

### API Changes
This PR has API changes: adds public `SimulationSysmemManager::get_mapped_host_ptr`; `SWEmuleChip::get_sysmem_manager()` now returns a live manager rather than `nullptr`.

### DIFF
--- a/device/api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
+++ b/device/api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
@@ -70,7 +70,9 @@ public:
     // kernel's in-place read/write. The pointer is valid only while the mapped buffer stays
     // mapped: the caller must not retain it across an unpin_or_unmap_sysmem()/SysmemBuffer
     // teardown. The registry lock guards the lookup, not the returned pointer's lifetime.
-    void* get_mapped_host_ptr(uint64_t device_io_addr, uint32_t size);
+    // This is a pure address translation — the caller owns the access length (the returned bare
+    // pointer carries none); the bounded copies write/read_mapped_buffer are where a size is checked.
+    void* get_mapped_host_ptr(uint64_t device_io_addr);
 
 protected:
     bool init_sysmem(uint32_t num_host_mem_channels) override;
--- a/device/api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
+++ b/device/api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
@@ -62,9 +62,14 @@ public:
     bool write_mapped_buffer(uint64_t device_io_addr, const void* src, uint32_t size);
     bool read_mapped_buffer(uint64_t device_io_addr, void* dst, uint32_t size);
 
-    // Same lookup as write/read_mapped_buffer, but returns the host pointer directly instead of
-    // performing the copy — for callers (e.g. tt-emule's NOC address resolver) whose contract is
-    // to resolve an address to a pointer once and let the caller memcpy. Returns nullptr on a miss.
+    // Zero-copy address→host-pointer lookup (nullptr on a miss). Unlike
+    // write/read_mapped_buffer, which copy, this returns the mapping directly for in-place
+    // access. emule needs it: its NOC address resolver (__emule_resolve_noc_addr) maps a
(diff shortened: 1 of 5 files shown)


xanderchin (approved):

tests/api/test_simulation_sysmem_manager.cpp
```diff
@@ -181,6 +181,40 @@ TEST_P(ApiSimulationSysmemManagerByArch, WriteReadThroughMappedBuffer) {
...
+
+    // The pointer aliases the backing (zero-copy): a write through it is visible via read_mapped_buffer.
+    std::vector<uint8_t> pattern = {0x11, 0x22, 0x33, 0x44};
+    std::memcpy(base_ptr, pattern.data(), pattern.size());
+    std::vector<uint8_t> readback(pattern.size(), 0);
+    ASSERT_TRUE(sysmem->read_mapped_buffer(device_addr, readback.data(), readback.size()));
+    EXPECT_EQ(pattern, readback);
+
+    // A miss (address in no mapped buffer) returns nullptr.
+    EXPECT_EQ(sysmem->get_mapped_host_ptr(/*device_io_addr=*/1, 1), nullptr);
```
pjanevskiTT: how do we know that there won't be device_io_addr equal to 1?

pjanevskiTT: that is true for current architecture but I wouldn't generalize that assumption

device/api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
```diff
@@ -62,6 +62,16 @@ class SimulationSysmemManager : public SysmemManager {
...
 
+    // Zero-copy address→host-pointer lookup (nullptr on a miss). Unlike
+    // write/read_mapped_buffer, which copy, this returns the mapping directly for in-place
+    // access. emule needs it: its NOC address resolver (__emule_resolve_noc_addr) maps a
+    // host-facing (PCIe) NOC address to a host pointer the emulated kernel dereferences
+    // directly — the way silicon's NOC reaches host memory — and a copy cannot back the
+    // kernel's in-place read/write. The pointer is valid only while the mapped buffer stays
+    // mapped: the caller must not retain it across an unpin_or_unmap_sysmem()/SysmemBuffer
+    // teardown. The registry lock guards the lookup, not the returned pointer's lifetime.
+    void* get_mapped_host_ptr(uint64_t device_io_addr, uint32_t size);
```
pjanevskiTT: I am not sure how much sense does size parameter have here? If the point is just to return the pointer in host memory that matches device_io_addr, why do we even care about the size?

arminaleTT: You're right. Removed

tests/api/test_simulation_sysmem_manager.cpp
arminaleTT: Changed the test to assert missing based on the actual allocated buffer address

pjanevskiTT (approved):

PR #3085, opened 2026-07-27
Only run unified tests for TSan

aleksamarkovicTT:
### Issue
/

### Description
Only run unified tests for TSan and other improvements.

### List of the changes
- Only run unified tests on TSan.
- Skip running TSan build on baremetal.
- Skip ttsim outside of release build type.
- Skip exalens tests, tt-smi tests on TSan.

### Testing
CI

### API Changes
This PR has (no) API changes.

### DIFF
--- a/.github/workflows/build-and-run-all-tests.yml
+++ b/.github/workflows/build-and-run-all-tests.yml
@@ -159,6 +159,8 @@ jobs:
         #   PR/merge queue runtime down.
         # - g13blx01 is the Blackhole Galaxy machine. It is fully disabled on all events for now as
         #   it has been causing CI problems; re-enable once the machine is stable (see #3070).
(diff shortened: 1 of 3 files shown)


.github/workflows/build-and-run-all-tests.yml
```diff
@@ -159,6 +159,8 @@ jobs:
         #   PR/merge queue runtime down.
```
pjanevskiTT, line 159: I guess in the future we might even have some threads inside libdevice as well, but hopefully all multithreaded tests expose any races we have. To be more precise we shouldn't have a bad behavior that is not exposed in multithreaded tests but it exists in single threaded. That being sad, I am ok with this change for now

pjanevskiTT (approved):

Issue #3089, opened 2026-07-28
Add workflow for running tests on exabox


dpopovTT:
**Description**
As we have flaky bh galaxies in Austin we need to have access to new cluster in Markham where all new machines are being put

**Scope**
Add workflow

**Is your feature request related to a problem? Please describe.**
Not being able to run CI on bh galaxies

**Definition of Done**
CI workflow that runs on exabox machiens

PR #3090, opened 2026-07-28
Add exabox workflow

dpopovTT:
### Issue
[(Link to Github issue)](https://github.com/tenstorrent/tt-umd/issues/3089)

### Description
Add exabox testing workflow in order to run tests on cluster in Markham

### List of the changes

1. New workflow that runs on exabox cluster
2. Added installation of openssh server and openmpi in order to trigger tests on multiple galaxies at once
3. Add a user that sshs into machines

### Testing
Running test on sc1 (single galaxy): [job](https://github.com/tenstorrent/tt-umd/actions/runs/30345985442)
### API Changes
This PR has no API changes.

### DIFF
--- a/.github/docker_install_common.sh
+++ b/.github/docker_install_common.sh
@@ -31,10 +31,11 @@ apt-get update && apt-get install -y \
     xxd \
     rpm \
     dpkg-dev \
-    fakeroot
+    fakeroot \
+    jq
 
 # Install Python dependencies
-python3 -m pip install --no-cache-dir pytest pyyaml
+python3 -m pip install --no-cache-dir pytest pyyaml tt-smi
 
 # gcc-11 should be available only for ubuntu 22 and not 20
 if apt-cache show gcc-11 > /dev/null 2>&1; then
--- a/.github/docker_install_common.sh
+++ b/.github/docker_install_common.sh
@@ -86,10 +86,13 @@ grep -q '^StrictModes no' /etc/ssh/sshd_config || echo "StrictModes no" >> /etc/
 
 # OpenMPI with ULFM — must match the exabox runner's launch agent path
 # (/opt/openmpi-v5.0.7-ulfm/bin/prted). Same package metal installs for multihost.
+# SHA256 pinned for supply-chain integrity of the GitHub release artifact.
 OMPI_DEB_URL="https://github.com/tenstorrent/ompi/releases/download/v5.0.7/openmpi-ulfm_5.0.7-1_amd64.deb"
+OMPI_DEB_SHA256="954e872d9105e8bf8c31368ff7a5db8670a3d549e2e7eb1ab6072cffcae7984d"
 OMPI_DEB_FILE="$(basename "$OMPI_DEB_URL")"
 OMPI_TMP="$(mktemp -d)"
 wget -q -O "${OMPI_TMP}/${OMPI_DEB_FILE}" "${OMPI_DEB_URL}"
+echo "${OMPI_DEB_SHA256}  ${OMPI_TMP}/${OMPI_DEB_FILE}" | sha256sum -c -
 apt-get install -y --no-install-recommends "${OMPI_TMP}/${OMPI_DEB_FILE}"
 rm -rf "${OMPI_TMP}"
 test -x /opt/openmpi-v5.0.7-ulfm/bin/prted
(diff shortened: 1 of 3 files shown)


.github/docker_install_common.sh
```diff
@@ -70,3 +71,25 @@ apt install -y clang-format-20 && \
...
+
+# OpenSSH server for TTOP / multihost MPI workers (mpirun SSHes into this image).
+# Keep config tweaks after the apt install so openssh-server does not overwrite them.
+apt-get install -y --no-install-recommends openssh-server sudo
+if ! id -u user >/dev/null 2>&1; then
+    adduser --uid 1001 --shell /bin/bash --disabled-password --gecos "" user
+fi
+usermod -aG sudo user
+echo 'user ALL=(ALL) NOPASSWD:ALL' > /etc/sudoers.d/user
+chmod 0440 /etc/sudoers.d/user
```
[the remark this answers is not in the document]
dpopovTT: Agreed that `NOPASSWD:ALL` is pretty broad for a generic CI image, but this is a TTOP multihost worker rather than a standard root-by-default job container.

TTOP SSHes into the worker as user (`uid 1001`) for MPI/mpirun `--pernode`, and that flow needs passwordless sudo for things like bootstrap and device resets (`tt-smi`). This is also consistent with the shared worker image in `tenstorrent/tt-orchestration` (`images/job-runner/Dockerfile`), which uses `user ALL=(ALL) NOPASSWD:ALL`.

So we can’t really remove sudo here without breaking the Exabox flow. An allowlist would definitely be better long term, but right now the commands we need aren’t a small, stable set. I’d prefer to stay consistent with the existing TTOP worker setup for now, and tighten it later once we have a better-defined bootstrap/reset surface.

.github/docker_install_common.sh
```diff
@@ -31,10 +31,11 @@ apt-get update && apt-get install -y \
     xxd \
     rpm \
     dpkg-dev \
-    fakeroot
+    fakeroot \
+    jq
 
 # Install Python dependencies
-python3 -m pip install --no-cache-dir pytest pyyaml
+python3 -m pip install --no-cache-dir pytest pyyaml tt-smi
```
broskoTT, line 38: installing tt-smi which is a tooling that depends on umd is not how we envision it... See the other comment to avoid this dep

.github/workflows/run-tests-on-exabox.yml
```diff
@@ -0,0 +1,337 @@
...
+      - name: Reset cards and validate cluster
+        working-directory: /tmp
+        timeout-minutes: 45
+        run: |
+          set -euo pipefail
+          MAX_RECOVERY_ATTEMPTS=10
+          SYSTEM_HEALTH=/home/user/tt-umd/build/tools/umd/system_health
+          export LD_LIBRARY_PATH=/home/user/tt-umd/build/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}
+
+          echo "--- Pre-check: tt-smi visibility on all hosts ---"
```
broskoTT, line 251: Checking that tt-smi sees all devices is no different than UMD seeing all devices.
There are a two options you can use:

- install tt-umd python package and then you can call directly some python statements here
- Run topology tool https://github.com/tenstorrent/tt-umd/tree/main/tools#topology-tool

.github/workflows/run-tests-on-exabox.yml
```diff
@@ -0,0 +1,337 @@
...
+              bash -lc "cd /home/user/tt-umd && '${SYSTEM_HEALTH}'"
+          }
+
+          for attempt in $(seq 1 "$MAX_RECOVERY_ATTEMPTS"); do
+            echo "=== Recovery attempt $attempt of $MAX_RECOVERY_ATTEMPTS ==="
+            reset_success=false
+            validation_success=false
+
+            echo "--- Resetting cards with tt-smi ---"
+            if mpirun --pernode tt-smi -glx_reset; then
```
broskoTT, line 271: Again, on newest tt-smi this is just a passthrough to UMD's reset api. you have same two options:

- call reset from python here, given you install tt-umd package. For inspiration you can use https://github.com/tenstorrent/tt-umd/blob/main/nanobind/tests/test_py_warm_reset.py
- Use warm reset cpp tool https://github.com/tenstorrent/tt-umd/blob/main/nanobind/tests/test_py_warm_reset.py

broskoTT (approved):
Approved if you don't add tt-smi to our docker

dpopovTT:
We have new exabox runners so I will close this one and open a new one with new runners

PR #3103, opened 2026-07-29
init arc/eth on ttsim device without cluster descriptor

micahwhiteTT:
### Description
This pull request introduces several improvements and new features for simulation device support and topology discovery, as well as enhancements to device register access and simulation device initialization. The main focus is on enabling topology discovery starting from an already-initialized simulation device, improving simulation device register access, and providing more accurate simulation device behavior. Behavior is unchanged when a cluster descriptor is passed.

### List of the changes

**Simulation device and topology discovery enhancements:**

* Added a new static method `TopologyDiscovery::discover_from_local_device` that allows topology discovery to begin from an already-initialized local device, which is essential for simulation and backends without OS device enumeration. Supporting methods such as `add_initialized_local_device` and `create_topology_discovery` were also introduced. [[1]](diffhunk://#diff-745853b66c9301ff1459255757a5fa84128f7e01f3564823b05924bde2fe04f8R39-R48) [[2]](diffhunk://#diff-745853b66c9301ff1459255757a5fa84128f7e01f3564823b05924bde2fe04f8R62-R75) [[3]](diffhunk://#diff-539c05723aa71ef8ffff4cf0f8eed4a9c900fd5d2bcd1dd632d86d2d6285e4f0R79-R86) [[4]](diffhunk://#diff-539c05723aa71ef8ffff4cf0f8eed4a9c900fd5d2bcd1dd632d86d2d6285e4f0R148-R171) [[5]](diffhunk://#diff-539c05723aa71ef8ffff4cf0f8eed4a9c900fd5d2bcd1dd632d86d2d6285e4f0R188-R218)
* Updated the simulation build flow in `Cluster::Cluster` to use the new `discover_from_local_device` method, initializing the simulation device and passing it to topology discovery.

**Simulation device register and memory access:**

* Implemented `read_from_device_reg` and `write_to_device_reg` in `SimulationTTDevice`, which delegate to the existing device read/write methods, enabling register access for simulation devices. [[1]](diffhunk://#diff-2fbddc65f4395d38e9645ba395d517c3f564b458f6d438cbb71fbab199ee6f96R59-R62) [[2]](diffhunk://#diff-7ae0b9e85d1aaf0b6e3d839480bea987d4c64733128174a416bf7ff247ae66f6R179-R188)
* Updated ARC APB and CSM register access methods in `SimulationTTDevice` to actually perform register reads/writes using the SoC descriptor, instead of throwing errors, improving simulation fidelity.

**Simulation device initialization and hang detection:**

* Added `init_tt_device_for_simulation` to `TTDevice` and ensured simulation devices set up a no-op hang detector, preventing errors from hang detection logic in simulation. [[1]](diffhunk://#diff-58a2231bd0578162c6f39e09858b7c8ea22e41ac24eb770a999a0355ac69c591R416-R418) [[2]](diffhunk://#diff-7ae0b9e85d1aaf0b6e3d839480bea987d4c64733128174a416bf7ff247ae66f6R17-R57)
* Integrated the new simulation device initialization flow into chip construction and remote chip creation logic in the cluster, ensuring correct device setup for both local and remote simulation chips. [[1]](diffhunk://#diff-ca95f78d8d9bce89552686931843a3572b94971a5ee2f9d03ca07d84ec0ede45R221-L229) [[2]](diffhunk://#diff-ca95f78d8d9bce89552686931843a3572b94971a5ee2f9d03ca07d84ec0ede45R255-R275)

**Other improvements and bug fixes:**

* Improved the logic for determining MMIO protocol in `HangDetectorImplementation` to avoid false positives when the protocol is null.
* Fixed a minor issue in cluster descriptor population to only record PCI BDFs when a PCI device is present.
* Provided architecture-specific minimum clock frequency values for simulation devices, improving simulation accuracy.

These changes collectively make simulation device support more robust, enable new simulation workflows, and improve the accuracy and maintainability of the simulation backend.

### Testing
tt-metal tests that failed with ttsim now pass

### DIFF
--- a/device/api/umd/device/cluster.hpp
+++ b/device/api/umd/device/cluster.hpp
@@ -743,13 +743,14 @@ private:
         SocDescriptor& soc_desc,
         int num_host_mem_channels,
         const std::filesystem::path& simulator_directory,
+        bool initialize_simulation_from_firmware,
         std::unique_ptr<TTDevice> tt_device = nullptr);
 #ifdef TT_UMD_BUILD_SIMULATION
-    // Builds the RemoteChip for a non-MMIO chip in a multichip ttsim cluster. A simulated remote chip has no
-    // ARC/topology discovery, so its RemoteCommunication and TTDevice (carrying the cluster-derived SocDescriptor
-    // and ChipInfo) are constructed here against the MMIO gateway.
+    // Builds the RemoteChip for a non-MMIO chip in a multichip ttsim cluster. Its RemoteCommunication and TTDevice
+    // are constructed here against the MMIO gateway; firmware initialization is optional so explicitly supplied
+    // mock descriptors can remain authoritative.
     std::unique_ptr<RemoteChip> create_simulation_remote_chip(
-        ChipId chip_id, ClusterDescriptor* cluster_desc, const SocDescriptor& soc_desc);
+        ChipId chip_id, ClusterDescriptor* cluster_desc, const SocDescriptor& soc_desc, bool initialize_from_firmware);
 
     // Host simulation Cluster only: exposes each simulation chip's device on its per-chip socket so a
     // separate client process (a Cluster pointed at the socket directory) can attach and drive it. A
--- a/device/api/umd/device/simulation/tt_sim_communicator.hpp
+++ b/device/api/umd/device/simulation/tt_sim_communicator.hpp
@@ -34,7 +34,8 @@ public:
         const std::filesystem::path &simulator_directory,
         bool copy_sim_binary = false,
         uint32_t chip_id = 0,
-        uint32_t num_chips = 1);
+        uint32_t num_chips = 1,
+        bool force_shared_bdf_mode = false);
 
     /**
      * Destructor that properly cleans up library handles and file descriptors.
@@ -109,6 +110,9 @@ public:
      * @return 32-bit value read from configuration space
      */
     uint32_t pci_config_read32(uint32_t bus_device_function, uint32_t offset);
+
+    // Enumerate consecutive host-visible PCI functions exposed by the loaded simulator.
+    uint32_t get_num_mmio_devices();
 
     /**
      * Advance the simulator clock.
@@ -211,6 +215,7 @@ private:
 
     uint32_t chip_id_ = 0;
     uint32_t num_chips_ = 1;
+    bool force_shared_bdf_mode_ = false;
     static void *s_shared_handle_;
     static int s_shared_refcount_;
     static bool s_sim_initialized_;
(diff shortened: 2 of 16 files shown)


aleksamarkovicTT (changes requested):
I'll give this a thorough review later, I disagree with the interface changes in TopologyDiscovery.

micahwhiteTT:
> I'll give this a thorough review later, I disagree with the interface changes in TopologyDiscovery.

Any high level comments I could use to refine it in the meantime?

mcraigheadTT:
Hey @aleksamarkovicTT - removing the cluster descriptor and using the simulator's ARC/eth paths instead is a high priority fix for us. Can you please elaborate on your concerns? Thanks!

aleksamarkovicTT:
These are my concerns with the PR:
- The PR is too large to give a thorough review, with almost 500 added lines; features, bug fixes and interface changes in one. My suggestion is to close it and divide it into smaller changes that we can discuss every decision in detail.
- The additions to public interfaces diverge from our committed specification of the UMD base API, which is a high priority for us. Work towards aligning to the base API is planned to be completed soon, after which we plan on restructuring the project to separate silicon and simulation into different targets. The main problem is, right now, TopologyDiscovery is designed only for silicon, and simulation needs to use ClusterDescriptor for initialization. Changing this can only be done with a thorough design discussion beforehand.
- All mixing of simulation-specific code with silicon code should be gated with the `TT_UMD_BUILD_SIMULATION` preprocessor definition (until we separate simulation into a different target). Although not all simulation code is gated like this, we need to work towards it.

Let's work on small changes towards your goals by splitting up the PR and discussing each decision separately.

mcraigheadTT:
I understand the PR size concerns, however I want to flag the following:
>[...] after which we plan on restructuring the project to separate silicon and simulation into different targets. The main problem is, right now, TopologyDiscovery is designed only for silicon, and simulation needs to use ClusterDescriptor for initialization. Changing this can only be done with a thorough design discussion beforehand.

Unless I am misunderstanding, this is the exact opposite of what we require from a simulation perspective. What we need is for the simulation and silicon paths to be fully unified to the maximum extent possible. We need simulation to converge on the exact same topology discovery, harvesting detection, etc. paths as silicon, and we need to get rid of the unacceptable and error-prone user experience of requiring sim users to specify a cluster descriptor for multichip runs. In fact, we also need to get rid of the need for simulation users to specify an SOC descriptor.

Not only is this a critical UX fix for sim users, it also ensures that we have a way to test these silicon code paths in simulation, and it aligns the UMD-based simulation paths with the paths that are used under ttsim-qemu, which greatly reduces the risk of ttsim-qemu-specific bugs (which are much more difficult to debug than ones that can be reproduced under the UMD-based simulation environment).

More generally, we on the simulation team need to be part of these conversations. If there are plans to "[restructure] the project to separate silicon and simulation into different targets", please come talk to me about this first before making any such changes. We can't have another repeat of the mess that happened with the sim server socket, which was also not aligned with our simulation team requirements.

Something is not right if major UMD sim changes are happening and my team and I (who write the simulator) are not part of the conversation.

micahwhiteTT:
All functionality implemented by UMD team. No longer necessary for reference.

PR #3109, opened 2026-07-30
Add timeout on membar

### LINKED ISSUE #3111: Fix `set_membar_flag` that polls forever when a core returns a corrupted barrier value
**Describe the bug**

`LocalChip::set_membar_flag` writes the barrier value to every core of a group and polls for it to be read back, with no timeout. A core that answers reads but returns a value that never matches makes the loop spin forever, and the value actually read is never logged.

Hit on a 6U BH Galaxy: one DRAM channel came up with its upper 16 data bits stuck high, so `0xbb` read back as `0xffff00bb`. Firmware reported the channel trained and unharvested, and reads returned at normal latency (1.15M reads in 2 s), so it was a hot spin, not a stalled access. `initialize_membars` runs serially over all 32 devices, so the stalled rank blocked the MPI collectives and hung all four hosts (128 ASICs) for an hour, with no indication of which device or core was involved.

**Environment**
- UMD commit (branch if applicable): reproduced on `main`; fix on `nbuncic/membar_poll_timeout` (PR #3109), branched from `db90f47e`
- FW version: 19.12.0
- KMD version: 2.10.0
- Board/system type: 6U Galaxy (BH), 4 hosts x 32 ASICs

**Logs**

Last lines from the hung host, from a locally instrumented `set_membar_flag` (TENSIX and ETH passed on the same device microseconds earlier):

```
[membar-probe] device 3 grp@0x0 : polling 8 core(s) for 0xbb
```

What the poll was reading, from a bounded probe of the same sequence:

```
chip 3 pci 3 asic_loc 4 ch 0 sub 0 core d17-12 (TRANSLATED) addr 0x0 : TIMEOUT
  expected 0x000000bb before 0xffffbeef last 0xffff00bb reads 1157157 elapsed 2000.003 ms
```

`before 0xffffbeef` is a `0xdeadbeef` poison write read back, so the low 16 bits round-trip and the high 16 always read as ones. The other 127 ASICs passed the identical probe.

**Stack trace (optional)**

None — gdb was attached to `mpirun` and the `bash` wrappers, not the test binary. Location established from the instrumentation above.

**To reproduce**

Needs a chip whose DRAM channel returns corrupted data; intermittent hardware fault, seen twice in ~50 boots.

1. Boot until a DRAM channel comes up corrupted while firmware still reports it trained (`dram_harvesting_mask 0x0`).
2. Call `start_device`, which runs `initialize_membars`.
3. `set_membar_flag` never returns.

Generally: any core returning something other than the barrier value at `barrier_addr` hangs the process.

**Expected behavior**

Give up after a bounded time and throw, naming the core and the value actually read.

Monotonic 1 s deadline (`MEMBAR_SYNC_TIMEOUT`) on all three barrier groups, throwing with device, core, address, expected value, **value read**, and cores synced. The deadline uses the `noexcept` overload of `check_timeout` so the hot path does no formatting.

```
Memory barrier timed out after 1000 ms on device 3 core d17-12 (TRANSLATED) at 0x0:
expected 0xbb, read 0xffff00bb. 0 of 8 core(s) synced.
```

`0xffff00bb` is the part that matters — it identified a dead 16-bit half of one channel in a minute, where the unbounded loop gave a one-hour hang and no location.

Re-issuing the write inside the poll was rejected: the write lands correctly here, so it would only hide the fault. UMD cannot catch this earlier — firmware training status and BIST both report the channel good, so `wait_dram_cores_training` passes it. That is being raised with the firmware team separately.

nbuncicTT:
### Issue
#3111 

### Description
This pull request introduces a timeout mechanism for memory barrier synchronization in the `LocalChip::set_membar_flag` method, enhancing error handling and preventing indefinite hangs when cores fail to synchronize. The key changes are as follows:

### List of the changes

**Timeouts and Synchronization Improvements:**

* Added `MEMBAR_SYNC_TIMEOUT` (1 second) to `timeouts.hpp` to limit how long the system waits for a core to echo the memory barrier value, preventing hangs if a core fails to respond.
* In `LocalChip::set_membar_flag`, implemented a timeout check using `utils::check_timeout` and the new `MEMBAR_SYNC_TIMEOUT`. If a core doesn't synchronize in time, the code now throws a detailed runtime error, including diagnostic information about the core and the value read.

**Code Maintenance:**

* Included the `utils.hpp` header in `local_chip.cpp` to support the new timeout utility function.

### Testing
Manually on faulty BH Galaxy

### API Changes
/

### DIFF
--- a/device/chip/local_chip.cpp
+++ b/device/chip/local_chip.cpp
@@ -37,6 +37,7 @@
 #include "umd/device/types/xy_pair.hpp"
 #include "umd/device/utils/error.hpp"
 #include "umd/device/utils/timeouts.hpp"
+#include "utils.hpp"
 
 namespace tt::umd {
 
@@ -413,6 +414,7 @@ void LocalChip::set_membar_flag(
         write_to_device(core, barrier_val_vec.data(), barrier_addr, barrier_val_vec.size() * sizeof(uint32_t));
     }
     tt_driver_atomics::sfence();  // Ensure that all writes in the Host WC buffer are flushed
+    const auto start = std::chrono::steady_clock::now();
     while (cores_synced.size() != cores.size()) {
         for (const auto& core : cores) {
             if (cores_synced.find(core) == cores_synced.end()) {
@@ -420,12 +422,23 @@ void LocalChip::set_membar_flag(
                 read_from_device(core, &readback_val, barrier_addr, sizeof(std::uint32_t));
                 if (readback_val == barrier_value) {
                     cores_synced.insert(core);
-                } else {
-                    log_trace(
-                        LogUMD,
-                        "Waiting for core {} to recieve mem bar flag {} in function",
-                        core.str(),
-                        barrier_value);
+                } else if (utils::check_timeout(start, timeout::MEMBAR_SYNC_TIMEOUT)) {
+                    // Report the value actually read. A core that answers but returns corrupted data
+                    // (e.g. a DRAM channel that trained with a dead byte lane) never converges, and the
+                    // readback is what identifies it.
+                    UMD_THROW(
+                        error::RuntimeError,
+                        fmt::format(
+                            "Memory barrier timed out after {} ms on device {} core {} at {:#x}: expected {:#x}, "
+                            "read {:#x}. {} of {} core(s) synced.",
+                            timeout::MEMBAR_SYNC_TIMEOUT.count(),
+                            tt_device_->get_pci_device()->get_device_num(),
+                            core.str(),
+                            barrier_addr,
+                            barrier_value,
+                            readback_val,
+                            cores_synced.size(),
+                            cores.size()));
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


device/chip/local_chip.cpp
```diff
@@ -413,19 +414,31 @@ void LocalChip::set_membar_flag(
...
+                    // Report the value actually read. A core that answers but returns corrupted data
+                    // (e.g. a DRAM channel that trained with a dead byte lane) never converges, and the
+                    // readback is what identifies it.
+                    UMD_THROW(
+                        error::RuntimeError,
+                        fmt::format(
+                            "Memory barrier timed out after {} ms on device {} core {} at {:#x}: expected {:#x}, "
+                            "read {:#x}. {} of {} core(s) synced.",
+                            timeout::MEMBAR_SYNC_TIMEOUT.count(),
+                            tt_device_->get_pci_device()->get_device_num(),
```
aleksamarkovicTT: Use the device ID method.

device/chip/local_chip.cpp
```diff
@@ -413,19 +414,31 @@ void LocalChip::set_membar_flag(
...
+                            core.str(),
```
aleksamarkovicTT: This will emit an exception per core, consider logging all cores that are being waited on upon timeout, if you think it's worth doing so. (I do)

aleksamarkovicTT (approved):

device/chip/local_chip.cpp
nbuncicTT: https://github.com/tenstorrent/tt-umd/pull/3114

Issue #2747, opened 2026-06-01
Refactor cluster's start_device


broskoTT:
Remove deassert_resets_and_set_power_state
Move all functionality to CHip's start_device.

Shouldn't be that hard to do, but there might be some bits and pieces to be careful about.

nbuncicTT:
Multiple PRs open

broskoTT:
currently in progress: removing iatu, deassert_risc_reset in chip

broskoTT:
Investigating further into llrt, this is a more complex task involving investigating tt-metal's flow

PR #3116, opened 2026-08-01
Remove init_pcie_iatus() and enforce minimum KMD version 2.0.0

aleksamarkovicTT:
### Issue
#2747 

### Description
Remove init_pcie_iatus() call in Chip's start_device. New KMD supersedes the need to do this in userspace.
Enforce minimum KMD version 2.0.0. The previous minimum version was 1.34.0.

### Testing
CI

### API Changes
This PR has no API changes.

### DIFF
--- a/device/api/umd/device/utils/kmd_versions.hpp
+++ b/device/api/umd/device/utils/kmd_versions.hpp
@@ -9,8 +9,13 @@
 namespace tt::umd {
 
 /**
+ * The minimum KMD version that can be used to interact with Tenstorrent devices in UMD.
+ */
+inline constexpr SemVer KMD_MINIMUM_VERSION = SemVer(2, 0, 0);
+
+/**
  * KMD version 1.29.0 introduced IOMMU support. UMD requires at least this version to run with IOMMU enabled.
- * With never versions of KMD, UMD will still work when IOMMU is disabled on the system.
+ * With newer versions of KMD, UMD will still work when IOMMU is disabled on the system.
  */
 inline constexpr SemVer KMD_IOMMU = SemVer(1, 29, 0);
 
--- a/device/chip/local_chip.cpp
+++ b/device/chip/local_chip.cpp
@@ -145,10 +145,6 @@ void LocalChip::start_device(uint32_t dram_membar_subchannel) {
     chip_started_lock_.emplace(acquire_mutex(MutexType::CHIP_IN_USE, tt_device_->get_pci_device()->get_device_num()));
 
     sysmem_manager_->pin_or_map_sysmem_to_device();
-    if (!tt_device_->get_pci_device()->is_mapping_buffer_to_noc_supported()) {
-        // If this is supported by the newer KMD, UMD doesn't have to program the iatu.
-        init_pcie_iatus();
-    }
     initialize_membars(dram_membar_subchannel);
 }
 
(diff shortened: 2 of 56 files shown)


joelsmithTT (approved):
Code deletion, I love to see it!

device/api/umd/device/utils/kmd_versions.hpp
```diff
@@ -8,9 +8,14 @@
 
 namespace tt::umd {
 
+/**
+ * The minimum KMD version that can be used to interact with Tenstorrent devices in UMD.
+ */
+inline constexpr SemVer KMD_MINIMUM_VERSION = SemVer(2, 0, 0);
```
broskoTT, line 14: I'd rather have a clear minimal PR which states that the KMD older than 2.0.0 will not be supported going forward. Then after that PR we can go and remove all the deadcode. I'm against coupling the two.

device/chip/local_chip.cpp
```diff
@@ -146,10 +146,6 @@ void LocalChip::start_device(uint32_t dram_membar_subchannel) {
     chip_started_lock_.emplace(acquire_mutex(MutexType::CHIP_IN_USE, tt_device_->get_pci_device()->get_device_num()));
 
     sysmem_manager_->pin_or_map_sysmem_to_device();
-    if (!tt_device_->get_pci_device()->is_mapping_buffer_to_noc_supported()) {
```
broskoTT, line 149: Not sure if this is full story... This code branches out on the initialization path, but on the destruction path you have similar branching somewhere else like "SiliconSysmemManager::unpin_or_unmap_sysmem". You should remove all of it and remove this function from pciDevice if you want to remove all is_mapping_buffer_to_noc_supported branches.

device/chip/local_chip.cpp
```diff
@@ -382,28 +378,6 @@ std::unique_lock<RobustMutex> LocalChip::acquire_mutex(MutexType mutex_type, int
     return lock_manager_.acquire_mutex(mutex_type, pci_device_id);
 }
 
-void LocalChip::init_pcie_iatus() {
```
broskoTT, line 385: This function should definitely go, but as an easier step you can just move it inside silicon sysmem manager for now, and hide it behind that api, that won't change any behavior.
After we stop supporting older KMD we can remove dead code obviously

device/api/umd/device/utils/kmd_versions.hpp
broskoTT: Also tt-smi not supporting KMD < 2.0.0 might be a problem, i.e. we have to think about our clients also.

To be frank, either than cleaning up a smaller part of our codebase I don't see a compelling reason not to still support older KMD versions

broskoTT (changes requested):

device/api/umd/device/utils/kmd_versions.hpp
aleksamarkovicTT: I will scope the PR to first remove support for older KMD. I see a clear benefit in not supporting older components - less code, less bug workarounds etc. If we don't test for it we should not support it, and I don't think there is any benefit and I really doubt there are any clients using KMD older than 2.0.0, and if they are, they are really out of date and should not continue with the practice. This is beta software and should move quickly.

From Joel:
> Something to consider is whether a higher version floor is worthwhile.
On one hand, the ABI is stable across major versions.  New UMD trying to use a new KMD API against old KMD will behave predictably (the ioctl will fail) and old UMD + new KMD should just work (KMD aims for backward compatibility).
On the other hand, if UMD isn't testing against 2.0.0 in CI, then it's hard to know whether it will actually work as expected.  What happens if you try to run latest tt-metal on KMD 2.0.0?  Maybe it works fine.  But it wouldn't surprise me if there are problems.

aleksamarkovicTT: Also, the iATU programming method in TTDevice will go shortly after, Joel said that it's not safe to call without locking.

joelsmithTT:
Please make the KMD version floor whatever it is you guys test in CI or expand your CI to cover things like KMD 2.0.0 if you want the floor there.

I don't see any value in testing against ancient KMD.  If a user has a problem with old KMD vs new UMD, then initial triage will involve upgrading the KMD.  An error message from UMD, "Upgrade the KMD to at least version x.y.z" is unambiguous and actionable.  That's a better user experience than an untested code path failing.

PR #3126, opened 2026-08-05
Fix cumulative remote I/O timeout accounting

micahwhiteTT:
### Description
Reset ethernet timeouts when progress is measured

### Testing
Metal tests pass with ttsim wh_x2


aleksamarkovicTT:
What is the motivation behind this change?

micahwhiteTT:
> What is the motivation behind this change?

TTSim times out on tt-metal tests here - relaxing this timeout allows it to pass. It is justifiable to reset the timeout everytime progress is observed.

aleksamarkovicTT (approved):

PR #3134, opened 2026-08-07
Create DmaInterface per Base API Specification

### LINKED ISSUE #2880: Create DmaInterface based on Base API Specification
**Description**

`DmaInterface` does not exist as a standalone component. DMA functionality is currently embedded in `PcieInterface`. Create `DmaInterface` per the Base API Specification as a dedicated component for bounce buffer and zero-copy DMA transfers, and migrate DMA-related behavior out of `PcieInterface`.

**Scope**

- Create `DmaInterface` per the Base API Specification
- Migrate DMA-related behavior from existing components into it
- Update all call sites

**Timeline**

DMA functionality is bundled into `PcieInterface` rather than being a standalone component, preventing reuse across different transports.

**Definition of Done**

- `DmaInterface` exists per the Base API Specification
- DMA behavior migrated from existing components
- All call sites updated

ajovanovicTT:
### Issue
#2880

### Description
DMA transfers were bundled into PcieInterface, preventing reuse across different transports. Introduce a standalone DmaInterface (bounce-buffer dma_read/dma_write/dma_multicast_write plus zero-copy variants) matching the Base API spec exactly, implemented by PcieProtocol alongside DeviceProtocol and PcieInterface.

Migrate SysmemBuffer's DMA transfers onto the new interface: it previously duplicated TLB/AXI-address resolution and called the raw-address zero-copy primitives directly, which only worked for PCIe. It now resolves through DmaInterface's zero-copy methods.

### List of the changes
 - Create DmaInterface matching the Base API spec exactly (bounce-buffer + zero-copy DMA), implemented by PcieProtocol
 - Remove the 3 core DMA methods from PcieInterface, now covered by DmaInterface
 - Migrate SysmemBuffer's DMA transfers onto DmaInterface's zero-copy methods, dropping its duplicated address resolution
 - Add TTDevice::dma_read_zero_copy/dma_write_zero_copy matching the spec, used by SysmemBuffer

### Testing
CI.

### API Changes
This PR has API changes.

### DIFF
--- a/device/api/umd/device/chip_helpers/sysmem_buffer.hpp
+++ b/device/api/umd/device/chip_helpers/sysmem_buffer.hpp
@@ -7,10 +7,8 @@
 #include <cstddef>
 #include <cstdint>
 #include <functional>
-#include <memory>
 #include <optional>
 
-#include "umd/device/pcie/tlb_window.hpp"
 #include "umd/device/types/xy_pair.hpp"
 
 namespace tt::umd {
@@ -124,8 +122,6 @@ private:
      */
     void validate(const size_t offset) const;
 
-    TlbWindow* get_cached_tlb_window();
-
     PCIDevice* pci_device_;
     TTDevice* tt_device_;
 
@@ -148,8 +144,6 @@ private:
     // Address that is used on the NOC to access the buffer.  NOC target must be
     // the PCIE core that is connected to the host and this address.
     std::optional<uint64_t> noc_addr_;
-
-    std::unique_ptr<TlbWindow> cached_tlb_window = nullptr;
 
     std::function<void()> unmap_callback_;
 };
--- a/device/api/umd/device/tt_device/protocol/dma_interface.hpp
+++ b/device/api/umd/device/tt_device/protocol/dma_interface.hpp
@@ -1,0 +1,110 @@
+/*
+ * SPDX-FileCopyrightText: © 2026 Tenstorrent Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+
+#include "umd/device/types/noc_id.hpp"
+#include "umd/device/types/xy_pair.hpp"
+
+namespace tt::umd {
+
+/**
+ * @brief DMA transfer operations between host and device memory.
+ *
+ * Two transfer modes:
+ * - Bounce buffer: copies through an internal pinned staging buffer. Returns false if DMA is
+ *   unavailable, letting the caller fall back.
+ * - Zero-copy: operates directly on caller-managed pinned memory identified by IOVA, bypassing
+ *   the staging buffer.
+ *
+ * All core coordinates are tt_xy_pair: raw (x, y) with no translation.
+ */
+class DmaInterface {
+public:
+    virtual ~DmaInterface() = default;
+
+    /**
+     * @brief Device-to-Host DMA via internal staging buffer.
+     * @param dst Pointer to the destination host buffer.
+     * @param src_addr Source address on the target core.
+     * @param size Number of bytes to transfer.
+     * @param core Source core coordinates.
+     * @param noc_id NOC to route through.
+     * @return true on success; false if DMA is unavailable.
+     */
+    [[nodiscard]] virtual bool dma_read(void* dst, uint64_t src_addr, size_t size, tt_xy_pair core, NocId noc_id) = 0;
+
+    /**
+     * @brief Host-to-Device DMA via internal staging buffer.
+     * @param src Pointer to the source host memory.
+     * @param dst_addr Destination address on the target core.
+     * @param size Number of bytes to transfer.
+     * @param core Target core coordinates.
+     * @param noc_id NOC to route through.
+     * @return true on success; false if DMA is unavailable.
+     */
+    [[nodiscard]] virtual bool dma_write(
+        const void* src, uint64_t dst_addr, size_t size, tt_xy_pair core, NocId noc_id) = 0;
+
+    /**
+     * @brief Multicast Host-to-Device DMA via internal staging buffer.
+     * @param src Pointer to the source host memory.
+     * @param dst_addr Destination address on the target cores.
+     * @param size Number of bytes to transfer.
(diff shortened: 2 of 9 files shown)


device/api/umd/device/tt_device/protocol/dma_interface.hpp
```diff
@@ -0,0 +1,110 @@
...
+     * @param noc_id NOC to route through.
+     * @return true on success; false if DMA is unavailable.
+     */
+    virtual bool dma_multicast_write_zero_copy(
+        uint64_t src_iova,
+        uint64_t dst_addr,
+        size_t size,
+        tt_xy_pair core_start,
+        tt_xy_pair core_end,
+        NocId noc_id) = 0;
```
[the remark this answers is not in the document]
ajovanovicTT: Left this to match base api exactly.

device/chip_helpers/sysmem_buffer.cpp
```diff
@@ -68,114 +61,27 @@ SysmemBuffer::SysmemBuffer(
...
 
-    if (pci_device_->get_dma_buffer().buffer == nullptr) {
-        UMD_THROW(
-            error::RuntimeError,
-            fmt::format(
-                "DMA buffer is not allocated on PCI device {}, PCIe DMA operations not supported.",
-                pci_device_->get_device_num()));
-    }
-
     validate(offset);
```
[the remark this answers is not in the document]
ajovanovicTT: Opened separate PR for this.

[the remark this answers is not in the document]
ajovanovicTT: https://github.com/tenstorrent/tt-umd/pull/3135

device/api/umd/device/tt_device/protocol/dma_interface.hpp
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * @brief DMA transfer operations between host and device memory.
+ *
+ * Two transfer modes:
+ * - Bounce buffer: copies through an internal pinned staging buffer. Returns false if DMA is
+ *   unavailable, letting the caller fall back.
+ * - Zero-copy: operates directly on caller-managed pinned memory identified by IOVA, bypassing
+ *   the staging buffer.
+ *
+ * All core coordinates are tt_xy_pair: raw (x, y) with no translation.
```
nbuncicTT: nit: I think this is a redundant comment, as we assume that xy_pair is not being translated across the whole codebase. I'd remove this part.

device/tt_device/tt_device.cpp
```diff
@@ -804,6 +811,36 @@ void TTDevice::dma_multicast_write(
...
+    if (is_remote_tt_device) {
+        UMD_THROW(error::RuntimeError, "DMA zero-copy write not supported for remote device.");
+    }
+    auto pcie_dma_lock =
+        lock_manager.acquire_mutex(MutexType::PCIE_DMA, communication_device_id_, communication_device_type_);
+
+    bool dma_success =
+        get_dma_interface()->dma_write_zero_copy(src_iova, dst_addr, size, resolve_coordinate(core, noc_id), noc_id);
+    if (!dma_success) {
+        UMD_THROW(error::RuntimeError, "DMA zero-copy write requires a functional DMA interface.");
```
nbuncicTT: Why do you assume the DMA is not functional here? 

I'd add two checks instead of one here:
- DMA presence (which is maybe already present in `get_dma_interface()`) - omit if the check is present of course
- DMA operation failure/pass

device/tt_device/tt_device.cpp
```diff
@@ -368,6 +370,13 @@ PcieInterface *TTDevice::get_pcie_interface() {
     return pcie_capabilities_;
 }
 
+DmaInterface *TTDevice::get_dma_interface() {
```
nbuncicTT, line 373: As said in the downward comment, because we have this check, no need to add in also in `dma_read_zero_copy` and other APIs as it would unnecessary clutter the code.

device/api/umd/device/chip_helpers/sysmem_buffer.hpp
```diff
@@ -149,8 +145,6 @@ class SysmemBuffer {
     // the PCIE core that is connected to the host and this address.
     std::optional<uint64_t> noc_addr_;
 
-    std::unique_ptr<TlbWindow> cached_tlb_window = nullptr;
```
nbuncicTT, line 152: Create a task with the label feature which we will use as a basis for an investigation on the following topics:
- Evaluate TLB caching/allocation usage for internal bounce buffer vs system memory buffer (i.e. we have zero copy and internal bounce buffer approaches, we need to evaluate if we should use different TLB sizes for these cases)
- Evaluate TLB size usage for these two API pairs 
- Evaluate fallbacks if there is no available TLB of the wanted size

This is UMD internal heuristic investigation, it's not a client-facing issue, so we as a team should align on this, and then provide some perf assumptions for arch-specific behavior (i.e. we could use a 4GB TLB if the size passed is bigger than some mark, etc.).

Just make an issue with this in mind, and we'll figure out the best path later on.

nbuncicTT (approved):

device/tt_device/tt_device.cpp
ajovanovicTT: Presence is already checked separately in get_dma_interface(), which throws if the DMA capability doesn't exist. The bool here is a second, different check: false only when no DMA buffer was allocated for zero copy, not op failed (there is no real failure signal today, hangs throw separately). Making the bool mean pass or fail of the transfer itself would be a bigger change, out of scope here. Reworded the message to match what it actually checks.

device/tt_device/tt_device.cpp
ajovanovicTT: Check in dma_read_zero_copy checks only if dma buffer is allocated, so there is actually no redundant check regarding this there. Just updated throw message in dma_read_zero_copy /write to tell precisely why there was a throw.

device/api/umd/device/chip_helpers/sysmem_buffer.hpp
ajovanovicTT: Opened issue : https://github.com/tenstorrent/tt-umd/issues/3148

PR #3139, opened 2026-08-07
Align TTDevice DMA method names/signatures with Base API Spec

### LINKED ISSUE #3300: Align TTDevice DMA method names and parameter order with Base API Spec
**Description**
TTDevice DMA method names and parameter order were inconsistent with the Base API Spec (tt_device_doxy.hpp).

**Scope**
- Rename dma_write_to_device → dma_write, dma_read_from_device → dma_read, dma_multicast_write → dma_write_to_core_range
- Reorder parameters to (buf, addr, size, core[, core_end], noc) per spec
- Update doc comments to match spec wording
- Update all call sites (SimulationTTDevice, LocalChip, nanobind internals, multiprocess test)
- Keep Python-exposed nanobind names unchanged (tt-exalens dependency); add TODOs

**Timeline**
No hard deadline; followup to #2880.

DMA method names and signatures diverged from the Base API Spec.

**Definition of Done**
Aligned with BaseAPI spec.

ajovanovicTT:
### Issue
#3300

### Description
Aligns TTDevice's DMA method names and parameter order with tt_device_doxy.hpp (Base API Spec).

### List of the changes
 - Rename dma_write_to_device to dma_write, dma_read_from_device to dma_read, dma_multicast_write to dma_write_to_core_range; reorder params to (buf, addr, size, core[, core_end], noc)
 - Update doc comments to match spec wording
 - Update all call sites (SimulationTTDevice, LocalChip, nanobind internals, multiprocess test)
 - Leave nanobind Python-exposed names unchanged (tt-exalens dependency); add TODOs

### Testing
CI.

### API Changes
This PR has API changes.


selormwalker (commented):
Approved changes. The implementation adheres to repo guidelines.

pjanevskiTT (approved):

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -178,53 +178,84 @@ class TTDevice {
...
+     * Copies from the user-provided buffer into an internal pinned staging buffer
+     * before issuing the hardware DMA to the device.
+     *
+     * @param src Pointer to the user-provided buffer containing the data to send.
+     * @param dst_addr Destination address on the target device core.
+     * @param size Number of bytes to transfer.
+     * @param core Target core coordinate on the device.
+     * @param noc_id Physical network to route the transaction over. Defaults to NocId::DEFAULT_NOC.
+     */
+    virtual void dma_write(
```
pjanevskiTT, line 196: This is a hard rename on an installed public header (`device/api/...`): `dma_write_to_device`, `dma_read_from_device` and `dma_multicast_write` are deleted outright with no `[[deprecated]]` forwarder. Anything downstream calling these on a `TTDevice` (tt-metal, tt-exalens' C++ side, tt-mlir) breaks at uplift with no migration window.

The convention we've been following for public UMD API renames is two stacked PRs: phase 1 gives the new names the real bodies and leaves the old names as one-line `[[deprecated]]` forwarders, phase 2 deletes the forwarders once downstream has migrated. The forwarders here are trivial:

```cpp
[[deprecated("Use dma_write() instead.")]]
void dma_write_to_device(
    const void* src, size_t size, CoreCoord core, uint64_t addr, NocId noc_id = NocId::DEFAULT_NOC) {
    dma_write(src, addr, size, core, noc_id);
}
```

nanobind/py_api_tt_device.cpp
```diff
@@ -652,7 +658,7 @@ void bind_tt_device(nb::module_ &m) {
                 tt_xy_pair core = {core_x, core_y};
                 {
                     nb::gil_scoped_release release;
-                    self.dma_read_from_device(data_ptr, data_size, core, addr, get_selected_noc_id());
+                    self.dma_read(data_ptr, addr, data_size, core, get_selected_noc_id());
```
pjanevskiTT, line 667: `noc_id` is validated at line 652 and then discarded — the transfer goes out on `get_selected_noc_id()`, which is thread-local and settable from Python via `set_thread_noc_id` (`nanobind/py_api_basic_types.cpp:30`).

So `umd.set_thread_noc_id(NocId.NOC1)` followed by `dev.dma_read_from_device(0, x, y, addr, buf)` passes the `noc_id must be 0` guard and then reads over NOC1. The same coordinates resolve to a different core on NOC0 vs NOC1, so this silently returns data from the wrong core while claiming to enforce NOC0.

Either pass the validated `noc_id` through, or drop the parameter and the guard rather than half-honouring it. (Pre-existing, and the sibling `noc_read`/`noc_write`/`noc_read32` bindings have the same shape — but this line is being touched here.)

device/chip/local_chip.cpp
```diff
@@ -292,19 +292,19 @@ void LocalChip::read_from_device(CoreCoord core, void* dest, uint64_t l1_src, si
...
+    tt_device_->dma_read(dst, addr, size, core, get_selected_noc_id());
 }
 
 int LocalChip::export_dmabuf(CoreCoord core, uint64_t addr, size_t size, uint64_t ordering) {
     return tt_device_->export_dmabuf(core, addr, size, ordering, get_selected_noc_id());
 }
 
 void LocalChip::dma_multicast_write(void* src, size_t size, CoreCoord core_start, CoreCoord core_end, uint64_t addr) {
-    tt_device_->dma_multicast_write(src, size, core_start, core_end, addr, get_selected_noc_id());
+    tt_device_->dma_write_to_core_range(src, addr, size, core_start, core_end, get_selected_noc_id());
```
pjanevskiTT, line 305: The rename stops at the `TTDevice` layer, so the same operation now has two names one frame apart: `Cluster::dma_multicast_write` -> `Chip::dma_multicast_write` -> `TTDevice::dma_write_to_core_range`. The Base API spec names the Chip-level entry point `dma_write_to_core_range()` too, so the Chip/Cluster layer is left mid-migration and its eventual rename becomes a second, unrelated public-API break.

The argument-order split is the sharper half of this. `Chip`/`Cluster`/`SysmemBuffer` keep `(src, size, core, addr)`; `TTDevice` is now `(src, addr, size, core)`, and each of these forwarders transposes two integers across a one-line boundary (lines 295 and 299 too). Writing `dma_write(src, size, addr, core, ...)` compiles with zero diagnostics — `size_t` <-> `uint64_t`, no `-Wconversion` and no `-Werror` in the build — and DMAs `addr` bytes to device address `size`.

Either rename Chip/Cluster in the same stack so the order is uniform, or note in the PR body which follow-up tracks it.

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -178,53 +178,84 @@ class TTDevice {
     virtual void write_to_device_reg(
         const void *mem_ptr, CoreCoord core, uint64_t addr, size_t size, NocId noc_id = NocId::DEFAULT_NOC);
 
-    virtual void dma_write_to_device(
-        const void *src, size_t size, CoreCoord core, uint64_t addr, NocId noc_id = NocId::DEFAULT_NOC);
+    /**
+     * @brief Executes a Host-to-Device (H2D) DMA transfer using an internal bounce buffer.
```
pjanevskiTT, line 185: The new doc states the DMA semantics unconditionally, but `TTDevice::dma_write` (`tt_device.cpp:760-768`) checks `dma_success` and, when DMA is unavailable, unlocks the PCIE_DMA mutex and falls back to `write_to_device()` — a TLB/MMIO path with different ordering, throughput and locking. `dma_read` does the same at 779-787.

A caller reading only this block has no way to know the transfer degraded. The fallback *is* documented, but only inside a different method ("Unlike dma_read, there is no non-DMA fallback", line 235). Worth stating it on `dma_write`/`dma_read`/`dma_write_to_core_range` themselves, since the doc blocks are being rewritten anyway.

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -178,53 +178,84 @@ class TTDevice {
...
-     * but uses DMA for better performance. Multicast writes data to a grid of cores. Cores must be specified in the
-     * translated coordinate system so that the write lands on the intended cores.
+     * @brief Executes a multicast Host-to-Device DMA transfer using an internal bounce buffer.
      *
-     * @param src pointer to memory from which the data is sent
-     * @param size number of bytes
-     * @param core_start starting core coordinates (x,y) of the multicast write
-     * @param core_end ending core coordinates (x,y) of the multicast write
-     * @param addr address on the device where data will be written
+     * Broadcasts data to a rectangular grid of cores via the internal staging buffer. Cores must be
```
pjanevskiTT, line 216: "Cores must be specified in the translated coordinate system" isn't accurate: `dma_write_to_core_range` calls `resolve_coordinate(core_start, noc_id)`, which translates from *any* non-LITERAL `CoordSystem` (`tt_device.cpp:850-858`), so NOC0/VIRTUAL/LOGICAL coords all work fine.

The real footgun is the opposite one, and it's documented on none of the three methods: `CoreCoord`'s default `coord_system` is `CoordSystem::LITERAL` (`types/core_coordinates.hpp:147`), which `resolve_coordinate` passes through untranslated — so `dma_write(src, addr, size, CoreCoord{1, 1})` targets raw NOC (1,1), not logical Tensix (1,1), with no error. Since these doc blocks are being rewritten, that's the sentence worth having.

tests/unified/multiprocess.cpp
```diff
@@ -36,13 +36,13 @@ using namespace tt::umd;
...
  * current architecture. On Wormhole B0, PCIe DMA reads are required/preferred, so
- * dma_read_from_device is used. On other architectures (including Blackhole), the standard
+ * dma_read is used. On other architectures (including Blackhole), the standard
  * read_from_device path is used instead.
  */
 void read_data_based_on_architecture(
     TTDevice& tt_device, CoreCoord core, void* mem_ptr, uint64_t address, size_t size) {
     if (tt_device.get_arch() == tt::ARCH::WORMHOLE_B0) {
-        tt_device.dma_read_from_device(mem_ptr, size, core, address);
+        tt_device.dma_read(mem_ptr, address, size, core);
```
pjanevskiTT, line 45: `dma_read` silently degrades to `read_from_device` when DMA isn't available (`tt_device.cpp:779-787`), so this helper — and `Multiprocess.DMAWriteReadRaceCondition`, which exists specifically to stress the concurrent DMA path — can run 4 threads x 500 iterations of pure MMIO, pass every `ASSERT_EQ`, and report success without exercising the PCIE_DMA lock path at all.

Worth asserting DMA is actually available up front, or using `dma_read_zero_copy`/`dma_write_zero_copy`, which throw instead of degrading.

device/tt_device/tt_device.cpp
```diff
@@ -749,7 +749,7 @@ int TTDevice::export_dmabuf(CoreCoord core, uint64_t addr, size_t size, uint64_t
     return get_pcie_interface()->export_dmabuf(resolve_coordinate(core, noc_id), addr, size, ordering, noc_id);
 }
 
-void TTDevice::dma_write_to_device(const void *src, size_t size, CoreCoord core, uint64_t addr, NocId noc_id) {
+void TTDevice::dma_write(const void *src, uint64_t dst_addr, size_t size, CoreCoord core, NocId noc_id) {
     ZoneScopedC(tracy::Color::MediumPurple);
     if (is_remote_tt_device) {
         UMD_THROW(error::RuntimeError, "DMA write to device not supported for remote device.");
```
pjanevskiTT: The message still names the deleted method: this throws from `dma_write` but says "DMA write to device". Same at line 774 ("DMA read from device" from `dma_read`) and line 794 ("DMA multicast write" from `dma_write_to_core_range`).

A user who hits one of these greps for `dma_write_to_device` / `dma_multicast_write` at the TTDevice layer and finds nothing, because those symbols no longer exist.

device/tt_device/simulation_tt_device.cpp
```diff
@@ -313,8 +313,8 @@ void SimulationTTDevice::noc_multicast_write(const void* src, size_t size, uint6
     noc_multicast_write(src, size, start, end, addr, noc_id);
 }
 
-void SimulationTTDevice::dma_multicast_write(
-    void* src, size_t size, CoreCoord core_start, CoreCoord core_end, uint64_t addr, NocId noc_id) {
+void SimulationTTDevice::dma_write_to_core_range(
+    const void* src, uint64_t dst_addr, size_t size, CoreCoord core_start, CoreCoord core_end, NocId noc_id) {
     UMD_THROW(error::RuntimeError, "DMA multicast write is not supported for simulation devices.");
```
pjanevskiTT: Same staleness as `tt_device.cpp`: `dma_write_to_core_range` throws "DMA multicast write is not supported...", a name that no longer exists anywhere in the TTDevice API after this PR.

device/chip/local_chip.cpp
```diff
@@ -292,19 +292,19 @@ void LocalChip::read_from_device(CoreCoord core, void* dest, uint64_t l1_src, si
...
 void LocalChip::dma_read_from_device(void* dst, size_t size, CoreCoord core, uint64_t addr) {
-    tt_device_->dma_read_from_device(dst, size, core, addr, get_selected_noc_id());
+    tt_device_->dma_read(dst, addr, size, core, get_selected_noc_id());
 }
 
 int LocalChip::export_dmabuf(CoreCoord core, uint64_t addr, size_t size, uint64_t ordering) {
     return tt_device_->export_dmabuf(core, addr, size, ordering, get_selected_noc_id());
 }
 
 void LocalChip::dma_multicast_write(void* src, size_t size, CoreCoord core_start, CoreCoord core_end, uint64_t addr) {
```
pjanevskiTT, line 304: Minor const-correctness: `TTDevice::dma_write_to_core_range` is tightened to `const void* src` in this PR, but `Chip::dma_multicast_write` (`chip.hpp:68`), `LocalChip` (`local_chip.hpp:64`), `RemoteChip`, `MockChip`, `SWEmuleChip`, `SimulationChip` and `Cluster::dma_multicast_write` (`cluster.hpp:473`) all keep `void* src`.

A caller holding a `const std::vector<uint8_t>&` payload still has to `const_cast` for a pure-read operation, even though every layer below is now const-correct — and `Chip::dma_write_to_device` directly above already takes `const void*`. Widening these is source-compatible for all existing callers.

nanobind/py_api_tt_device.cpp
```diff
@@ -473,14 +473,16 @@ void bind_tt_device(nb::module_ &m) {
             nb::arg("soft_reset_raw_value"),
             release_gil(),
             "Set the raw soft reset register value for a core in translated coordinates. ")
+        // TODO: rename to dma_read to match TTDevice::dma_read once tt-exalens's umd_device.py (which
```
pjanevskiTT: The same two-line comment is pasted at 476, 494 and 646 — worth collapsing to one above the group.

On the substance: "updated in lockstep" isn't achievable. tt-exalens pins/vendors a UMD version and uplifts on its own schedule, so there's no build where both sides flip atomically — whoever picks this up will hard-rename the binding and break exalens the same way the C++ side does. The workable plan is the same two-phase shape: register the new Python name as an additional `.def` alias, let exalens migrate, then drop the old alias. Worth putting that in the TODO so the next person doesn't have to rediscover it.

device/api/umd/device/tt_device/tt_device.hpp
ajovanovicTT: Deprecated forwarders aren't needed here — none of the named consumers actually call these TTDevice methods directly today: tt-metal only goes through Cluster::dma_write_to_device/dma_read_from_device (untouched, separate layer), tt-exalens only touches the nanobind Python names (kept stable in this PR), and tt-mlir doesn't call this API at all (only vendors it transitively).

nanobind/py_api_tt_device.cpp
ajovanovicTT: This is pre existing behavior out of scope for this renaming PR, it has to stay stateful noc for now.

tests/unified/multiprocess.cpp
ajovanovicTT: Out of scope for this PR, might update this in separate PR.

device/chip/local_chip.cpp
ajovanovicTT: Again just respecting BaseApi spec, layers above update are not part of this PR and this change breaks nothing.

device/chip/local_chip.cpp
ajovanovicTT: This is current part of aligning to BaseAPI spec, not touching layers above. Leaving as is, with later updates to layers above.

device/api/umd/device/tt_device/tt_device.hpp
ajovanovicTT: Wouldn't state implementation details here in doc, respected Base api spec.

device/api/umd/device/tt_device/tt_device.hpp
ajovanovicTT: Will leave this doc comment as is, woudn't spill implementation details of core coordinate translation here

PR #3144, opened 2026-08-10
Expose export_dmabuf on LocalChip and Cluster (4/4)

ajovanovicTT:
Description
Part 4 of 4 splitting #3063 into a stack (kmdlib -> pcidevice -> ttdevice -> cluster). Stacked on PR 3/4, completes the feature. No behavior change versus #3063, and the cumulative diff of this PR against main is byte for byte identical to #3063.

List of the changes
Added LocalChip::export_dmabuf and Cluster::export_dmabuf as thin propagators
export_dmabuf is intentionally only on LocalChip, not the Chip base class, since the export wraps a PCIe aperture that a remote chip does not have. Cluster::export_dmabuf goes through get_local_chip, which already rejects a non-local chip id
Added TestClusterExportDmabuf in tests/unified/test_tlb.cpp

Testing
Using TEMP test from rdma_testing/* on branch ajovanovic/rdma_testing_tools, paired sources run at the same time on bh-glx6u-18 and bh-glx6u-22
CI

API Changes
This PR has API changes. New public methods LocalChip::export_dmabuf and Cluster::export_dmabuf

### DIFF
--- a/device/api/umd/device/cluster.hpp
+++ b/device/api/umd/device/cluster.hpp
@@ -516,6 +516,30 @@ public:
      */
     TlbWindow* get_static_tlb_window(const ChipId chip, const CoreCoord core);
 
+    /**
+     * Allocate a dedicated TLB window aimed at (chip, core, addr), export it as a dma-buf for
+     * peer-to-peer PCIe DMA, and return an fd the caller owns.
+     * - The caller must close() the returned fd when done; this releases the kmd-side pin and,
+     *   once the last export on the window is released, returns the window to the allocation pool.
+     * - A dedicated window is always allocated (never a shared/static one), so there is no
+     *   reconfigure hazard with other TLB users.
+     * - `size` is the exported length, not a TLB window size: the underlying window is rounded up
+     *   internally to the smallest per-arch size class that can cover [addr, addr + size), so
+     *   callers need not know the arch's valid window sizes. The dma-buf is exactly `size` bytes.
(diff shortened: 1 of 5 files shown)


device/api/umd/device/cluster.hpp
```diff
@@ -516,6 +516,30 @@ class Cluster {
      */
     TlbWindow* get_static_tlb_window(const ChipId chip, const CoreCoord core);
 
+    /**
+     * Allocate a dedicated TLB window aimed at (chip, core, addr), export it as a dma-buf for
```
nbuncicTT: Delete any mention of TLB windows. This is an abstraction which shouldn't know about low-level details. We will rename the enum for ordering in some other PR. Don't spill impl. details like "dedicated TLB window is exclusive etc."

tests/unified/test_tlb.cpp
```diff
@@ -93,6 +98,45 @@ TEST_F(TestTlb, TestTlbWindowAllocateNew) {
     }
 }
 
+TEST_F(TestTlb, TestClusterExportDmabuf) {
```
nbuncicTT, line 101: Does it make sense to add a loop-back test here as it requires one host? You can also do this in another PR.

nbuncicTT (approved):

tests/unified/test_tlb.cpp
ajovanovicTT: Will open a separate PR with new test under test_tlb.cpp

PR #3155, opened 2026-08-12
Reject device descriptors that index past core/NOC tables in CoordinateManager

### LINKED ISSUE #3154: Out-of-bounds read in CoordinateManager from device-descriptor grid/core/table sizes
Describe the bug
================
Grid sizes and core coordinates parsed from a device-descriptor YAML are used as vector indices without validating that the indices are within the sizes of the vectors they index. A descriptor that declares a tensix grid larger than the number of cores it provides, or a core whose NOC0 coordinate exceeds the NOC0->NOC1 translation table, performs an out-of-bounds heap read in CoordinateManager and aborts the process.

Affected path: https://github.com/tenstorrent/tt-umd/blob/main/device/coordinates/coordinate_manager.cpp

Failure mechanism
-----------------
1) translate_tensix_coords() (coordinate_manager.cpp:222 on main):

    const tt_xy_pair& tensix_core = tensix_cores[y * grid_size_x + x];

grid_size_x/y come from the descriptor fields grid.x_size / grid.y_size; tensix_cores comes from functional_workers. Nothing enforces tensix_cores.size() == grid_size_x * grid_size_y, so a large grid indexes past the end of the vector.

2) add_noc1_to_noc0_mapping() (coordinate_manager.cpp:852 on main):

    noc0_x_to_noc1_x[tensix_core.x], noc0_y_to_noc1_y[tensix_core.y]

The core coordinates and the NOC0->NOC1 tables all come from the descriptor. A core whose coordinate exceeds the table size indexes past the end.

Reproduction
------------
Compile the real coordinate_manager.cpp and its architecture coordinate-manager sources under AddressSanitizer and drive create_coordinate_manager() with a descriptor grid larger than its core list. The process aborts:

    AddressSanitizer: heap-buffer-overflow on address ...
    READ of size 16 ... 
    CoordinateManager::translate_tensix_coords() coordinate_manager.cpp:222
    CoordinateManager::add_core_translation() coordinate_manager.cpp:91

The trigger is a single malicious descriptor; no hardware is required.

Expected behavior
-----------------
A descriptor whose declared grid does not match its provided core list, or whose core coordinates are outside the NOC translation tables, should be rejected with a clear runtime error instead of causing an out-of-bounds read.

Environment
-----------
- UMD commit: f63270792de36d0806b4d170d720c8314b001fd0 (main)
- Reproduction is host-side and hardware-free (AddressSanitizer on the real translation unit)

Proposed fix
------------
Added bounds validation in CoordinateManager before indexing:
- Reject a tensix grid larger than the provided core list, and bounds-check the flat loop index in translate_tensix_coords().
- Reject cores whose NOC0 coordinates exceed the size of the NOC0->NOC1 translation tables in add_noc1_to_noc0_mapping().

Also pinging @nbuncicTT per the coordination with the tt-metal maintainers.

peter941221:
### Issue
Fixes https://github.com/tenstorrent/tt-umd/issues/3154

### Description
Grid sizes, core coordinates, and NOC0->NOC1 translation tables all come from a device descriptor YAML. There was no invariant tying them together, so a descriptor could declare a tensix grid larger than its core list or a core whose NOC0 coordinate exceeds the translation table, causing an out-of-bounds heap read in CoordinateManager.

Add bounds validation before indexing:
- translate_tensix_coords(): reject a tensix grid larger than the number of provided cores, and bounds-check the flat loop index.
- add_noc1_to_noc0_mapping(): reject cores whose NOC0 coordinates exceed the size of the NOC0->NOC1 translation tables.

A crafted descriptor now produces a clean RuntimeError at descriptor parsing instead of a heap-buffer-overflow abort.

### List of the changes
- device/coordinates/coordinate_manager.cpp
  - guard in translate_tensix_coords() against an oversized tensix grid and against an out-of-range flat index
  - bounds check on tensix_core.x / tensix_core.y in add_noc1_to_noc0_mapping()
- tests/baremetal/test_core_coord_translation_wh.cpp
  - two focused gtests asserting the malformed descriptors are rejected with UmdException<RuntimeError>

### Testing
- Local (hardware-free): compiled coordinate_manager.cpp and the architecture coordinate-manager sources with clang and drove create_coordinate_manager() with the malformed inputs. Both new guards throw UmdException<RuntimeError>:
  - "tensix core list size N is smaller than declared grid WxH"
  - "NOC1 mapping index (X,Y) is out of range"
- Before the fix, the same driver does not reject the input and performs the out-of-bounds heap read.
- The two gtests cover the same paths and run in the baremetal CI lane.

### API Changes
This PR has no API changes.

### DIFF
--- a/device/coordinates/coordinate_manager.cpp
+++ b/device/coordinates/coordinate_manager.cpp
@@ -215,25 +215,11 @@ void CoordinateManager::translate_tensix_coords() {
     size_t grid_size_x = tensix_grid_size.x;
     size_t grid_size_y = tensix_grid_size.y;
 
-    if (grid_size_x != 0 && grid_size_y != 0 && tensix_cores.size() / grid_size_x < grid_size_y) {
-        UMD_THROW(
-            error::RuntimeError,
-            fmt::format(
-                "CoordinateManager: tensix core list size {} is smaller than declared grid {}x{}",
-                tensix_cores.size(),
-                grid_size_x,
-                grid_size_y));
-    }
-
     size_t logical_y = 0;
     for (size_t y = 0; y < grid_size_y; y++) {
         if (!(harvesting_masks.tensix_harvesting_mask & (1 << y))) {
             for (size_t x = 0; x < grid_size_x; x++) {
-                const size_t idx = y * grid_size_x + x;
-                if (idx >= tensix_cores.size()) {
-                    UMD_THROW(error::RuntimeError, "CoordinateManager: tensix core index out of range");
-                }
-                const tt_xy_pair& tensix_core = tensix_cores[idx];
+                const tt_xy_pair& tensix_core = tensix_cores.at(y * grid_size_x + x);
 
                 CoreCoord logical_coord = CoreCoord(x, logical_y, CoreType::TENSIX, CoordSystem::LOGICAL);
                 add_core_translation(logical_coord, tensix_core);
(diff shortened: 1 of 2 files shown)


device/coordinates/coordinate_manager.cpp
```diff
@@ -215,11 +215,25 @@ void CoordinateManager::translate_tensix_coords() {
...
     size_t logical_y = 0;
     for (size_t y = 0; y < grid_size_y; y++) {
         if (!(harvesting_masks.tensix_harvesting_mask & (1 << y))) {
             for (size_t x = 0; x < grid_size_x; x++) {
-                const tt_xy_pair& tensix_core = tensix_cores[y * grid_size_x + x];
+                const size_t idx = y * grid_size_x + x;
+                if (idx >= tensix_cores.size()) {
+                    UMD_THROW(error::RuntimeError, "CoordinateManager: tensix core index out of range");
+                }
+                const tt_xy_pair& tensix_core = tensix_cores[idx];
```
aleksamarkovicTT: We'd prefer just using .at() instead of throwing exceptions.

peter941221: Updated to use .at() for the tensix core list and both NOC mapping tables. The regression tests now assert std::out_of_range. Rebuilt baremetal_tests and ran CoordinateManager.*: 57/57 passed.

aleksamarkovicTT (approved):

aleksamarkovicTT:
Thanks for the contribution.

PR #3156, opened 2026-08-12
Add read-only host page pinning support

### LINKED ISSUE #2680: Add read_only flag to SysmemManager::map_sysmem_buffer
We want to be able to pin read-only files onto the device. I've filed https://github.com/tenstorrent/tt-kmd/issues/240 and Joel is adding a flag to the KMD to allow that, so I'd like to be able to specify that flag through the UMD API.

jbaumanTT:
### Issue
Closes #2680

### Description
Adds support for pinning host pages so that the device may read them but not write them.

KMD 2.9.0 introduced a read-only pinning mode: the IOMMU mapping permits device reads and faults device writes, which additionally allows device-readable mappings of read-only and shared file-backed memory. Without it, a host buffer that is only ever a DMA *source* still has to be pinned read/write, so memory the process cannot write — an `mmap(PROT_READ)` of a weights file, for example — cannot be pinned at all and has to be copied first.

This exposes that mode through `DeviceBufferAccess` and threads it down to the pin ioctl.

### List of the changes
- New `DeviceBufferAccess` enum (`ReadWrite` / `ReadOnly`) in `device/api/umd/device/types/host_memory.hpp`.
- `PCIDevice::map_for_dma()` and `PCIDevice::map_buffer_to_noc()` take a `DeviceBufferAccess` (defaulting to `ReadWrite`) and set `TT_DMA_FLAG_READ_ONLY` when read-only is requested.
- New `PCIDevice::is_read_only_page_pinning_supported()`, true only when an IOMMU is active and the KMD is at least 2.9.0. Requesting a read-only mapping without support throws rather than silently widening.
- New `KMD_READ_ONLY_PAGE_PINNING = 2.9.0` in `kmd_versions.hpp`.
- New `TT_DMA_FLAG_READ_ONLY` in the tt-kmd-lib flag enum, translated to `TENSTORRENT_PIN_PAGES_READ_ONLY` in `tt_dma_map()`.
- `SysmemBuffer` records its access mode and exposes `get_device_access()`. `dma_read_from_device()` now throws on a read-only mapping, since that direction writes host memory.
- `SysmemManager::map_sysmem_buffer()` (silicon and simulation) plumbs the access mode through.

### Testing
New `ApiSysmemManager.ReadOnlySharedFileMapping` in `tests/api/test_sysmem_manager.cpp`, gated on the capability of the device the mapping is actually made through. It seeds a file with a position-dependent pattern, maps it `PROT_READ`, pins it read-only, has the device DMA it into Tensix L1, and reads that back to compare — so it fails if the KMD accepts the flag but maps the pages wrongly, which an accessor-only test could not detect. It also asserts `dma_read_from_device` throws on a read-only mapping, since that direction writes host memory. Readback goes over MMIO rather than DMA because D2H DMA is unsupported on Blackhole, which would otherwise make the test skip there.

`api_tests --gtest_filter='*SysmemManager*'` on `yyzo-bh-07`: 24 passed, 2 skipped (the pre-existing DMA siblings, which skip on Blackhole).

Verified on `yyzo-bh-07` (Blackhole p150, KMD 2.9.0, IOMMU `DMA-FQ`) through the consumer side in tt-metal — the tt-metal branch below builds against this commit and its four pinned-memory suites pass, including a 36 MiB H2D upload from a `PROT_READ` file mapping, which is the case that cannot work without this change. Note that the CI runner fleet is still on KMD 2.8.0, so the read-only paths skip there; on that path the tt-metal side widens to read/write and keeps working.

Paired with tenstorrent/tt-metal#52893, which consumes this and is related to tenstorrent/tt-metal#46132. This PR needs to land first — that one bumps its umd submodule to this branch.

### API Changes
This PR has API changes.

- `PCIDevice::map_for_dma()` and `PCIDevice::map_buffer_to_noc()` gain a trailing `DeviceBufferAccess access = DeviceBufferAccess::ReadWrite` parameter. Existing callers are source compatible.
- `SysmemManager::map_sysmem_buffer()` gains the same trailing defaulted parameter.
- New public `DeviceBufferAccess` type, `SysmemBuffer::get_device_access()`, and `PCIDevice::is_read_only_page_pinning_supported()`.

### DIFF
--- a/device/api/umd/device/pcie/pci_device.hpp
+++ b/device/api/umd/device/pcie/pci_device.hpp
@@ -230,10 +230,11 @@ public:
      * Map a buffer so it is accessible by the device NOC.
      * @param buffer must be page-aligned
      * @param size must be a multiple of the page size
+     * @param device_access READ_ONLY requires is_read_only_page_pinning_supported()
      * @return uint64_t NOC address, uint64_t PA or IOVA
      */
     std::pair<uint64_t, uint64_t> map_buffer_to_noc(
-        void *buffer, size_t size, DeviceBufferAccess access = DeviceBufferAccess::ReadWrite);
+        void *buffer, size_t size, DeviceBufferAccess device_access = DeviceBufferAccess::READ_WRITE);
 
     /**
      * Map a hugepage so it is accessible by the device NOC.
@@ -251,9 +252,10 @@ public:
      *
      * @param buffer must be page-aligned
      * @param size must be a multiple of the page size
+     * @param device_access READ_ONLY requires is_read_only_page_pinning_supported()
      * @return uint64_t PA (no IOMMU) or IOVA (with IOMMU) for use by the device
      */
-    uint64_t map_for_dma(void *buffer, size_t size, DeviceBufferAccess access = DeviceBufferAccess::ReadWrite);
+    uint64_t map_for_dma(void *buffer, size_t size, DeviceBufferAccess device_access = DeviceBufferAccess::READ_WRITE);
 
     /**
      * @return whether this device and KMD support device-read-only host mappings
--- a/device/api/umd/device/pcie/pci_device.hpp
+++ b/device/api/umd/device/pcie/pci_device.hpp
@@ -22,6 +22,7 @@
 #include "umd/device/pcie/silicon_tlb_handle.hpp"
 #include "umd/device/pcie/tlb_handle.hpp"
 #include "umd/device/types/arch.hpp"
+#include "umd/device/types/host_memory.hpp"
 #include "umd/device/types/tlb.hpp"
 #include "umd/device/utils/semver.hpp"
 
@@ -231,7 +232,8 @@ public:
      * @param size must be a multiple of the page size
      * @return uint64_t NOC address, uint64_t PA or IOVA
      */
-    std::pair<uint64_t, uint64_t> map_buffer_to_noc(void *buffer, size_t size);
+    std::pair<uint64_t, uint64_t> map_buffer_to_noc(
+        void *buffer, size_t size, DeviceBufferAccess access = DeviceBufferAccess::ReadWrite);
 
     /**
      * Map a hugepage so it is accessible by the device NOC.
@@ -251,7 +253,12 @@ public:
      * @param size must be a multiple of the page size
      * @return uint64_t PA (no IOMMU) or IOVA (with IOMMU) for use by the device
      */
-    uint64_t map_for_dma(void *buffer, size_t size);
+    uint64_t map_for_dma(void *buffer, size_t size, DeviceBufferAccess access = DeviceBufferAccess::ReadWrite);
+
+    /**
+     * @return whether this device and KMD support device-read-only host mappings
+     */
+    bool is_read_only_page_pinning_supported() const;
 
     /**
      * Access the device's DMA buffer.  This buffer is not guaranteed to exist.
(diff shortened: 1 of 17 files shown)


device/api/umd/device/types/host_memory.hpp
```diff
@@ -0,0 +1,20 @@
...
+namespace tt::umd {
+
+/**
+ * @brief Device permissions for a pinned host-memory mapping.
+ *
+ * ReadOnly prevents the device from writing the mapping. It does not restrict
+ * host access to the underlying virtual memory.
+ */
+enum class DeviceBufferAccess {
+    ReadWrite,
```
nbuncicTT: nit: Can you make the enum class values in upper snake case, e.g.:
```
enum class DeviceBufferAccess {
    READ_WRITE,
    READ_ONLY,
};
```

device/chip_helpers/silicon_sysmem_manager.cpp
```diff
@@ -464,9 +464,9 @@ std::unique_ptr<SysmemBuffer> SiliconSysmemManager::allocate_sysmem_buffer(
 }
 
 std::unique_ptr<SysmemBuffer> SiliconSysmemManager::map_sysmem_buffer(
-    void *buffer, size_t sysmem_buffer_size, const bool map_to_noc) {
+    void *buffer, size_t sysmem_buffer_size, const bool map_to_noc, DeviceBufferAccess access) {
```
nbuncicTT: nit: I'd like the parameter to hold more meaning, can it be renamed to `device_access`?

device/api/umd/device/pcie/pci_device.hpp
```diff
@@ -251,7 +253,12 @@ class PCIDevice {
      * @param size must be a multiple of the page size
      * @return uint64_t PA (no IOMMU) or IOVA (with IOMMU) for use by the device
      */
-    uint64_t map_for_dma(void *buffer, size_t size);
+    uint64_t map_for_dma(void *buffer, size_t size, DeviceBufferAccess access = DeviceBufferAccess::ReadWrite);
+
+    /**
+     * @return whether this device and KMD support device-read-only host mappings
+     */
+    bool is_read_only_page_pinning_supported() const;
```
nbuncicTT, line 263: We intend to omit access to low level classes such as `PCIDevice` in the near future and push everything needed for these kinds of changes through high level interfaces.

That being said, does it make sense to query the `SysmemManager` itself so see if the mapping is supported (it would be a pass-through for this function)?

If you don't want to do it now, that's fine, you can leave the PR as it is, but just as a heads-up - we'll do it in the upcoming weeks and that may require a two-step change if you depend on it in tt-metal.

nbuncicTT (approved):

device/api/umd/device/types/host_memory.hpp
jbaumanTT: Done.

device/chip_helpers/silicon_sysmem_manager.cpp
jbaumanTT: Done.

device/api/umd/device/pcie/pci_device.hpp
jbaumanTT: Ok, added the query to SysmemManager. Let me know if that matches what you were thinking.

nbuncicTT: It's great, thanks!

Issue #2876, opened 2026-06-22
Migrate FirmwareInfoProvider interface to Base API Specification


nbuncicTT:
**Description**

`FirmwareInfoProvider` already exists and closely matches the Base API Specification. It needs to be made virtual (currently `final`) to fit into `TTDeviceModel`, some return types need to become `std::optional` per the spec, and a few missing getters need to be added. Static helpers and telemetry engine internals stay in the concrete implementation.

**Scope**

- Make `FirmwareInfoProvider` virtual per the Base API Specification
- Adjust return types to `std::optional` where the spec requires it
- Add missing getters per the spec
- Move static helpers and internals into the concrete implementation

**Timeline**

No hard deadline. Part of the UMD Base API migration epic.

**Is your feature request related to a problem? Please describe.**

`FirmwareInfoProvider` is currently a final class, preventing it from being used as a component interface in `TTDeviceModel`.

**Definition of Done**

- `FirmwareInfoProvider` interface matches the Base API Specification
- Concrete implementation retains internal helpers
- All consumers updated
- Existing tests pass or are updated

broskoTT:
PRs progressing

PR #3158, opened 2026-08-12
Make all FirmwareInfoProvider methods virtual, deprecate get_eth_fw_version()

aleksamarkovicTT:
### Issue
#2876 

### Description
Make all FirmwareInfoProvider methods virtual, deprecate get_eth_fw_version().

### List of the changes
- Make all FirmwareInfoProvider methods virtual.
- Deprecate get_eth_fw_version().
- Nanobind can use deprecate get_eth_fw_version_semver() as it is not bound by the base API.

### Testing
CI

### API Changes
This PR has API changes.


device/api/umd/device/firmware/firmware_info_provider.hpp
```diff
@@ -39,181 +39,180 @@ class FirmwareInfoProvider {
...
     virtual ~FirmwareInfoProvider() = default;
 
-    FirmwareBundleVersion get_firmware_version() const;
+    virtual FirmwareBundleVersion get_firmware_version() const;
 
-    uint64_t get_board_id() const;
+    virtual uint64_t get_board_id() const;
 
-    std::optional<uint32_t> get_eth_fw_version() const;
+    // TODO: Will be removed in UMD Base API 2.0.0
```
nbuncicTT: Instead of just making a TODO, please add an issue with the context why should this be removed, then link the issue in the comment - that way we know if it's solved or not.

nbuncicTT (approved):

PR #3159, opened 2026-08-13
[base API] Optional return values in FirmwareInfoProvider

aleksamarkovicTT:
### Issue
#2876 

### Description
Add missing optional return values to methods in FirmwareInfoProvider to match spec.

### List of the changes
- Add stub impl. for `get_min_clock_freq()`.
- Add missing optional return value for `get_board_id()`
- Add missing optional return value for `get_asic_temperature()`
- Add missing optional return value for `get_asic_location()`
- Add missing optional return value for `get_heartbeat()`
- Add missing optional return value for `get_max_clock_freq()`
- Adapt tests.

### Testing
CI

### API Changes
This PR has API changes.


tools/telemetry.cpp
```diff
@@ -38,7 +38,7 @@ std::string run_default_telemetry(tt::ChipId chip_id, FirmwareInfoProvider* firm
         return fmt::format("Could not get information for chip ID {}.", chip_id);
     }
 
-    double asic_temperature = firmware_info_provider->get_asic_temperature();
+    double asic_temperature = firmware_info_provider->get_asic_temperature().value_or(0.0);
```
nbuncicTT, line 41: I feel the tool is misleading to restoring to the 0. I feel that it should report N/A instead of 0, but this is a follow up issue - no action required for this PR. Create the follow up issue and let's refactor the tool when we have time.

nbuncicTT (approved):

PR #3160, opened 2026-08-13
Dma reads mixed bug repro test

### LINKED ISSUE #3448: Add automated repro test for DMA reads mixing between TLB windows on device re-init
**Description**
DMA reads could occasionally get mixed between TLB windows after a device re-init: a DMA read targeting one NOC core would return data belonging to a different NOC core's in-flight DMA read, mis-routing the transfer to the wrong TLB window/requester. See #3095 for the original report, repro steps, and logs. The underlying bug has since been fixed; this issue tracks adding an automated regression test for it.

**Scope**
- Add an automated test reproducing the mixed-core DMA read scenario from #3095 (repeated device re-init, concurrent DMA reads across multiple cores).
- Verify the test passes against the current fix with zero foreign/mis-routed reads across many iterations/cores.

**Timeline**
No hard deadline.

Yes - #3095 was a silent data mis-routing bug (no crash) that's now fixed, but there's no regression coverage for it in the test suite.

**Definition of Done**
Automated repro test for the #3095 scenario is merged into the UMD test suite and passes reliably.

ajovanovicTT:
### Issue
#3448

### Description
Test to reproduce bug from above linked issue.

### List of the changes
 - Added new test

### Testing
Local runs, bug reproduced on aus-wh-08 and bgd-lab-16 (both n150)

### API Changes
This PR has no API changes.


tests/unified/multiprocess.cpp
```diff
@@ -491,3 +495,200 @@ TEST(Multiprocess, DISABLED_DMAWriteReadRaceConditionProcessIsolation) {
...
+        {
+            std::unique_ptr<TTDevice> probe_device = TTDevice::create(pci_device_id);
+            probe_device->init_tt_device();
+            std::vector<CoreCoord> cores =
+                probe_device->get_soc_descriptor().get_cores(CoreType::TENSIX, CoordSystem::TRANSLATED);
+            core = cores.at(worker_id % cores.size());
+        }
+        result->core = core;
+
+        pthread_barrier_wait(start_barrier);
```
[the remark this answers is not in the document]
ajovanovicTT: Its intended for workers to synchronize right before they start to do writes/reads.

tests/unified/multiprocess.cpp
```diff
@@ -491,3 +495,200 @@ TEST(Multiprocess, DISABLED_DMAWriteReadRaceConditionProcessIsolation) {
 
     std::cout << "DMA race condition test (real fork) completed" << std::endl;
 }
+
+namespace dma_reads_mixed_repro {
```
nbuncicTT, line 499: Why are you using a separate namespace?

tests/unified/multiprocess.cpp
```diff
@@ -491,3 +495,200 @@ TEST(Multiprocess, DISABLED_DMAWriteReadRaceConditionProcessIsolation) {
...
+
+// Tag layout mirrors the tt-metal/ttexalens repro this test is modeled on: worker(4b) |
```
nbuncicTT, line 501: Remove these noise adding comments. The test is the documentation here.

tests/unified/multiprocess.cpp
```diff
@@ -491,3 +495,200 @@ TEST(Multiprocess, DISABLED_DMAWriteReadRaceConditionProcessIsolation) {
...
+// noc_x(6b) | noc_y(6b) | iteration(16b). Every word of the payload carries the same tag, so a
+// short readback still identifies who actually produced the data.
+constexpr int NUM_WORKERS = 16;
```
nbuncicTT, line 504: No need to scatter around structs, constants, functions, etc.

Try to use tools we have already for these situations, e.g. use TEST_F and create a test class that encapsulates this.

tests/unified/multiprocess.cpp
```diff
@@ -491,3 +495,200 @@ TEST(Multiprocess, DISABLED_DMAWriteReadRaceConditionProcessIsolation) {
...
+};
+
+// Core of one worker: pin to a NOC core, then repeatedly tear down and re-create the TTDevice
+// (the mechanism under test, standing in for ttexalens' per-iteration init_ttexalens() in the
+// original repro), write a tagged payload with a plain MMIO write ("noc_write" in the repro),
+// and read it back via dma_read_from_device. A "foreign" result means the DMA read landed on
+// data tagged for a different worker/core -- the completed transfer was matched to the wrong
+// requester.
+void run_worker(int worker_id, int pci_device_id, WorkerResult* result, pthread_barrier_t* start_barrier) {
+    try {
```
nbuncicTT, line 552: Why are you doing this in a try catch block?

nbuncicTT: I mean, I'm completely fine with having a thrown exception which stops the execution of the test - this means the device is not working well and we'll see it better in the CI.

tests/unified/multiprocess.cpp
```diff
@@ -491,3 +495,200 @@ TEST(Multiprocess, DISABLED_DMAWriteReadRaceConditionProcessIsolation) {
...
+// while several processes drive concurrent DMA traffic against distinct cores on the same chip.
+// Modeled directly on the tt-metal/ttexalens repro (same worker|noc_x|noc_y|iteration tag, same
+// stale-vs-foreign classification) but reinitializes via UMD's own TTDevice::create() /
+// init_tt_device() instead of ttexalens, and drives I/O the same way UMD itself would: a plain
+// write_to_device ("noc_write") followed by a dma_read_from_device readback.
+//
+// Disabled by default, like DISABLED_DMAWriteReadRaceConditionProcessIsolation above (real
+// fork() alongside gtest has known flakiness, see issue #2579) -- run explicitly against real
+// Wormhole hardware with --gtest_also_run_disabled_tests to check for the bug.
+TEST(Multiprocess, DISABLED_DmaReadMixedCoreRepro) {
```
nbuncicTT, line 615: Why is this disabled? I understand that we have flakiness in some situations, but what's here holding us back?

tests/unified/multiprocess.cpp
```diff
@@ -491,3 +495,200 @@ TEST(Multiprocess, DISABLED_DMAWriteReadRaceConditionProcessIsolation) {
...
+        }
+    } catch (const std::exception& e) {
+        result->errored = true;
+        std::strncpy(result->error_message, e.what(), sizeof(result->error_message) - 1);
+    }
+}
+
+}  // namespace dma_reads_mixed_repro
+
+// Reproduces a reported bug: a PCIe DMA read targeting one NOC core occasionally returns the
```
nbuncicTT, line 604: Don't explain the code in a paragraph and then do it. The code should be understandable enough - maybe a one sentence explanation can be made here.

tests/unified/multiprocess.cpp
```diff
@@ -491,3 +495,200 @@ TEST(Multiprocess, DISABLED_DMAWriteReadRaceConditionProcessIsolation) {
...
+// init_tt_device() instead of ttexalens, and drives I/O the same way UMD itself would: a plain
+// write_to_device ("noc_write") followed by a dma_read_from_device readback.
+//
+// Disabled by default, like DISABLED_DMAWriteReadRaceConditionProcessIsolation above (real
+// fork() alongside gtest has known flakiness, see issue #2579) -- run explicitly against real
+// Wormhole hardware with --gtest_also_run_disabled_tests to check for the bug.
+TEST(Multiprocess, DISABLED_DmaReadMixedCoreRepro) {
+    using namespace dma_reads_mixed_repro;
+
+    std::vector<int> pci_device_ids = PCIDevice::enumerate_devices();
```
nbuncicTT, line 618: The three lines here are a `SetUp`, where you stop the execution of the test if this doesn't work, it's boiler plate code and doesn't add any value to the test.

tests/unified/multiprocess.cpp
```diff
@@ -491,3 +495,200 @@ TEST(Multiprocess, DISABLED_DMAWriteReadRaceConditionProcessIsolation) {
...
+    ASSERT_FALSE(pci_device_ids.empty());
+    const int pci_device_id = pci_device_ids.at(0);
+
+    void* barrier_mem =
```
nbuncicTT, line 622: If this barrier is needed for the test setup, push it into a private part of the test class, and a getter and initialize it in SetUp, again, it doesn't add value to the test.

tests/unified/multiprocess.cpp
```diff
@@ -491,3 +495,200 @@ TEST(Multiprocess, DISABLED_DMAWriteReadRaceConditionProcessIsolation) {
...
+    pthread_barrierattr_init(&barrier_attr);
+    pthread_barrierattr_setpshared(&barrier_attr, PTHREAD_PROCESS_SHARED);
+    ASSERT_EQ(pthread_barrier_init(start_barrier, &barrier_attr, NUM_WORKERS), 0);
+
+    void* results_mem =
+        mmap(nullptr, sizeof(WorkerResult) * NUM_WORKERS, PROT_READ | PROT_WRITE, MAP_SHARED | MAP_ANONYMOUS, -1, 0);
+    ASSERT_NE(results_mem, MAP_FAILED);
+    WorkerResult* results = static_cast<WorkerResult*>(results_mem);
+
+    std::cout << "Testing DMA read mixed-core repro on PCI device " << pci_device_id << " with " << NUM_WORKERS
```
nbuncicTT, line 636: Remove this `std::cout`, we don't need to know this.

tests/unified/multiprocess.cpp
```diff
@@ -491,3 +495,200 @@ TEST(Multiprocess, DISABLED_DMAWriteReadRaceConditionProcessIsolation) {
...
+
+        total_stale += r.stale;
+        total_foreign += r.foreign;
+        total_iterations += r.completed_iterations;
+    }
+
+    std::cout << "\ntotal: " << total_stale << " stale, " << total_foreign << " foreign / " << total_iterations
+              << " inits" << std::endl;
+
+    pthread_barrier_destroy(start_barrier);
```
nbuncicTT, line 688: From this line and below it's a `TearDown` sequence, please use this.

tests/unified/multiprocess.cpp
nbuncicTT: This is also called in a fork, so the exit status will signalize that something went wrong.

tests/unified/multiprocess.cpp
```diff
@@ -491,3 +495,200 @@ TEST(Multiprocess, DISABLED_DMAWriteReadRaceConditionProcessIsolation) {
...
+// original repro), write a tagged payload with a plain MMIO write ("noc_write" in the repro),
+// and read it back via dma_read_from_device. A "foreign" result means the DMA read landed on
+// data tagged for a different worker/core -- the completed transfer was matched to the wrong
+// requester.
+void run_worker(int worker_id, int pci_device_id, WorkerResult* result, pthread_barrier_t* start_barrier) {
+    try {
+        CoreCoord core;
+        {
+            std::unique_ptr<TTDevice> probe_device = TTDevice::create(pci_device_id);
+            probe_device->init_tt_device();
```
nbuncicTT, line 556: this `init_tt_device` is for getting the tensix cores? And the other ones in the for loop are there to trigger the behavior?

I'm asking this because I think that you could reuse the `TTDevice` there. I don't see benefit in destroying the object - additionally if there is no need for `init_tt_device` for the repro, why is it in for loop?

tests/unified/multiprocess.cpp
```diff
@@ -491,3 +495,200 @@ TEST(Multiprocess, DISABLED_DMAWriteReadRaceConditionProcessIsolation) {
...
+            std::unique_ptr<TTDevice> tt_device = TTDevice::create(pci_device_id);
+            tt_device->init_tt_device();
+
+            std::fill(
+                payload.begin(),
+                payload.end(),
+                encode_tag(static_cast<uint32_t>(worker_id), core.x, core.y, static_cast<uint32_t>(iteration)));
+            std::fill(readback.begin(), readback.end(), 0);
+
+            tt_device->write_to_device(payload.data(), core, SCRATCH_ADDR, NUM_BYTES);
```
nbuncicTT, line 578: In the code, I think one line comments could explain why this is the repro, i.e. here we expect to write and read the same data, but previously it was being corrupted by the fact that the TLB reconfiguration has been done later/midflight etc. (you know the details).

From this point of view, the test isn't all that clear.

nbuncicTT (commented):
Please re-do the test, make it more clear and concise what is the goal of the test and refactor it so the intention is clear through the code with a minimal repro and boilerplate code moved in the test fixture.

tests/unified/multiprocess.cpp
ajovanovicTT: Unnecessary, since there are no naming conflicts, dropped this.

tests/unified/multiprocess.cpp
ajovanovicTT: Removed unnecessary comments.

tests/unified/multiprocess.cpp
ajovanovicTT: Unnecessary again, leftover from mirroring existing test in tt-metal on which the bug was initialy reproduced, so removed.

tests/unified/multiprocess.cpp
ajovanovicTT: Probe init is just for the soc descriptor to get the core list. The init in the loop was only needed for coordinate translation, so I take the x,y from the probe and pass
it as LITERAL now. Keeping the create and destroy, that part is the actual repro.

tests/unified/multiprocess.cpp
ajovanovicTT: Clarified with one additional comment on this place.

tests/unified/multiprocess.cpp
ajovanovicTT: Temporarily enabled it to see how it behaves in the CI run.

tests/unified/multiprocess.cpp
ajovanovicTT: Applied.

nbuncicTT (approved):

tests/unified/multiprocess.cpp
ajovanovicTT: CI passes, leaving it enabled.

PR #3165, opened 2026-08-14
Fix DMA reads returning another core's data via unflushed TLB config

### LINKED ISSUE #3095: DMA reads get mixed between TLB windows after device re-init
**Describe the bug**

When UMD is re-initialized repeatedly (device/link init in a loop), a DMA read targeting one
NOC core returns the data of a *different* DMA read issued against a *different* NOC core.

Concretely: process A does `noc_write` + `dma_read_from_device` on core `(1,1)`, process B does the
same on core `(1,2)`. Occasionally the read on `(1,1)` comes back holding the payload that was
written to `(1,2)` — i.e. the completed DMA transfer is matched to the wrong TLB window / wrong
requester. The payload is intact and self-consistent, it just belongs to another core, so this is a
mis-routed transfer and not data corruption.

The repro tags every write with `worker_id | noc_x | noc_y | iteration`, so the data read back
identifies exactly which core/worker it actually came from.

It only reproduces when the device is re-initialized inside the loop; with a single init at the
start the reads are clean. `gc.collect()` is required in the loop, otherwise the TLBs are exhausted
before enough iterations run — so TLB allocation/lifetime looks related.

**Environment**
- UMD commit (branch if applicable): `6fd12eee0`, `main`
- FW version: not relevant — reproduces regardless
- KMD version: not relevant — reproduces regardless
- Board/system type: Wormhole

**Logs**

```text
<paste the output of: python umd.py --cores 16 --iterations 300>
```

Failing lines look like this (script output format — replace with a real run):

```text
worker  3 core=  1,1 inits=300 stale=0 foreign=2:  FOREIGN
        iter 117: DMA read returned worker 5's data from core (1,2) iter 117  (expected worker 3 core 1,1)
```

**Stack trace (optional)**

n/a — no exception, the read succeeds and returns the wrong buffer.

**To reproduce**

1. Build the UMD Python bindings and install them into the ttexalens environment.
2. Get the repro script: https://github.com/tenstorrent/tt-metal/compare/main...iklikovac/exalens-repro
   → `tt_metal/tt-llk/tests/umd.py` (branch `iklikovac/exalens-repro`, commit `eb6c595`).
   The relevant part is per iteration: re-init the link, `noc_write(x, y, 0x10000, 256B)`,
   then `dma_read_from_device(x, y, 0x10000, 256B)`, compare.
3. Run with several processes hitting different cores concurrently:
   ```
   python umd.py --cores 16 --iterations 300
   ```
4. Look for `foreign` > 0 in the summary. Each sample line prints which core the data actually
   came from.

**Expected behavior**

`dma_read_from_device(x, y, addr, n)` returns the contents of `(x, y)` at `addr` — always, and
regardless of how many times the device has been re-initialized or how many other processes are
doing DMA reads on other cores at the same time. A DMA read must never return another window's
data.

ajovanovicTT:
### Issue
#3095

### Description
PCIe DMA transfers could be routed through the DMA TLB window's previous configuration and
return another core's data. MMIO writes are posted, so the window config write is not
guaranteed to have landed when the doorbell starts the DMA engine, which is a separate
requester reached through a different register block and so is not ordered behind it.

### List of the changes
 - PCIDevice::configure_tlb takes an optional verify flag that polls the config register until
it reads back the written low 64 bits, throwing on timeout.
 - TlbHandle::set_verify_config marks a window as needing that verification, passed through by
SiliconTlbHandle on every configure. Off by default, so MMIO paths pay nothing.
 - PcieProtocol enables it on the PCIe DMA window only, set before the window is constructed so
the first configure is covered too.

### Testing
CI, local overhead measurement via existing benchmarks

### API Changes
This PR has API changes.

### DIFF
--- a/device/api/umd/device/pcie/tlb_handle.hpp
+++ b/device/api/umd/device/pcie/tlb_handle.hpp
@@ -52,6 +52,18 @@ public:
      */
     int get_tlb_id() const { return tlb_id_; }
 
+    /**
+     * Require every subsequent configure() to confirm the new configuration has reached the device
+     * before returning, at the cost of a PCIe round trip per reconfigure.
+     *
+     * Off by default, which is correct whenever the host itself is the next user of the window: the
+     * host's access to the window travels behind the config write, into the same block, so it
+     * cannot be translated by the old configuration. Turn it on for a window whose next user is
+     * on-chip -- notably the PCIe DMA engine, which is a separate requester started by a doorbell to
+     * a different register block, and so is not ordered behind the config write at all.
+     */
+    void set_verify_config(const bool verify) { verify_config_ = verify; }
+
     virtual tt::ARCH get_arch() const = 0;
 
 protected:
@@ -65,6 +77,7 @@ protected:
     size_t tlb_size_ = 0;
     tlb_data tlb_config_{};
     TlbMapping tlb_mapping_ = TlbMapping::UC;
+    bool verify_config_ = false;
 
 private:
     /**
--- a/device/api/umd/device/pcie/tlb_handle.hpp
+++ b/device/api/umd/device/pcie/tlb_handle.hpp
@@ -52,18 +52,6 @@ public:
      */
     int get_tlb_id() const { return tlb_id_; }
 
-    /**
-     * Require every subsequent configure() to confirm the new configuration has reached the device
-     * before returning, at the cost of a PCIe round trip per reconfigure.
-     *
-     * Off by default, which is correct whenever the host itself is the next user of the window: the
-     * host's access to the window travels behind the config write, into the same block, so it
-     * cannot be translated by the old configuration. Turn it on for a window whose next user is
-     * on-chip -- notably the PCIe DMA engine, which is a separate requester started by a doorbell to
-     * a different register block, and so is not ordered behind the config write at all.
-     */
-    void set_verify_config(const bool verify) { verify_config_ = verify; }
-
     virtual tt::ARCH get_arch() const = 0;
 
 protected:
@@ -72,12 +60,31 @@ protected:
      */
     TlbHandle() = default;
 
+    /**
+     * Protected constructor letting derived classes fix verify_config_ for the lifetime of the handle.
+     *
+     * @param verify_config Require every subsequent configure() to confirm the new configuration has
+     *                       reached the device before returning, at the cost of a PCIe round trip per
+     *                       reconfigure. False is correct whenever the host itself is the next user of
+     *                       the window: the host's access to the window travels behind the config
+     *                       write, into the same block, so it cannot be translated by the old
+     *                       configuration. Pass true for a window whose next user is on-chip --
+     *                       notably the PCIe DMA engine, which is a separate requester started by a
+     *                       doorbell to a different register block, and so is not ordered behind the
+     *                       config write at all.
+     */
+    explicit TlbHandle(const bool verify_config) : verify_config_(verify_config) {}
+
+    /**
+     * Whether configure() must confirm the new configuration reached the device before returning.
+     */
+    bool get_verify_config() const { return verify_config_; }
+
     int tlb_id_ = 0;
     uint8_t* tlb_base_ = nullptr;
     size_t tlb_size_ = 0;
     tlb_data tlb_config_{};
     TlbMapping tlb_mapping_ = TlbMapping::UC;
-    bool verify_config_ = false;
 
 private:
     /**
@@ -85,6 +92,8 @@ private:
      * Implemented by derived classes.
      */
     virtual void free_tlb() noexcept = 0;
+
+    bool verify_config_ = false;
 };
 
 }  // namespace tt::umd
--- a/device/pcie/pci_device.cpp
+++ b/device/pcie/pci_device.cpp
@@ -37,6 +37,7 @@
 #include "umd/device/types/arch.hpp"
 #include "umd/device/utils/error.hpp"
 #include "umd/device/utils/kmd_versions.hpp"
+#include "umd/device/utils/timeouts.hpp"
 #include "utils.hpp"
 
 namespace tt::umd {
@@ -810,7 +811,7 @@ std::unique_ptr<TlbHandle> PCIDevice::allocate_tlb(const size_t tlb_size, const 
     }
 }
 
-void PCIDevice::configure_tlb(const uint32_t tlb_index, const tlb_data &tlb_config) {
+void PCIDevice::configure_tlb(const uint32_t tlb_index, const tlb_data &tlb_config, const bool verify) {
     // Get the TLB configuration for this index.
     auto tlb_configuration = arch_impl_->get_tlb_configuration(tlb_index);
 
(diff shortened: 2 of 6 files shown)


device/pcie/pci_device.cpp
```diff
@@ -832,12 +833,37 @@ void PCIDevice::configure_tlb(const uint32_t tlb_index, const tlb_data &tlb_conf
...
+    if (verify) {
+        static constexpr auto TLB_CONFIG_BUSY_POLL_WINDOW =
+            std::chrono::duration_cast<std::chrono::microseconds>(timeout::MMIO_OP_TIMEOUT);
+        static constexpr auto TLB_CONFIG_POLL_INTERVAL = std::chrono::microseconds(1);
+
+        const bool config_landed = utils::poll_until(
+            [&]() { return tlb_reg_ptr[0] == config_words[0] && tlb_reg_ptr[1] == config_words[1]; },
+            timeout::MMIO_OP_TIMEOUT,
+            TLB_CONFIG_BUSY_POLL_WINDOW,
+            TLB_CONFIG_POLL_INTERVAL);
```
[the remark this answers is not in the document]
ajovanovicTT: Intentionally added check for just lower ones.

device/pcie/pci_device.cpp
```diff
@@ -832,12 +833,37 @@ void PCIDevice::configure_tlb(const uint32_t tlb_index, const tlb_data &tlb_conf
 
     // Write the TLB register values
     // Wormhole uses 64-bit registers (8 bytes), Blackhole uses 96-bit registers (12 bytes).
-    tlb_reg_ptr[0] = static_cast<uint32_t>(lower_64);
-    tlb_reg_ptr[1] = static_cast<uint32_t>(lower_64 >> 32);
-
-    if (arch == tt::ARCH::BLACKHOLE) {
-        // Blackhole needs the upper 32 bits as well (96-bit total).
-        tlb_reg_ptr[2] = static_cast<uint32_t>(upper_64);
+    const uint32_t config_words[] = {
```
nbuncicTT: You're using C-style arrays, which is discouraged, please use std::array instead.

device/api/umd/device/pcie/tlb_handle.hpp
```diff
@@ -52,6 +52,18 @@ class TlbHandle {
...
+     * Require every subsequent configure() to confirm the new configuration has reached the device
+     * before returning, at the cost of a PCIe round trip per reconfigure.
+     *
+     * Off by default, which is correct whenever the host itself is the next user of the window: the
+     * host's access to the window travels behind the config write, into the same block, so it
+     * cannot be translated by the old configuration. Turn it on for a window whose next user is
+     * on-chip -- notably the PCIe DMA engine, which is a separate requester started by a doorbell to
+     * a different register block, and so is not ordered behind the config write at all.
+     */
+    void set_verify_config(const bool verify) { verify_config_ = verify; }
```
nbuncicTT: Before exposing any APIs, and adding any state to the class, please first measure the impact of reconfiguration (before and after adding the check) and how big of an impact it would have in general (i.e. how much perf do we lose with the overhead of reading the value back, because I think it would maybe be insignificant - but we need the measurement for that). 

If it's a significant overhead, then I'd like to omit adding a general API for a TLB window which will always retain this state (I don't see a scenario where you would create a TLB then use this API to set and unset the verification).. What I'd do instead is add another constructor which has one additional parameter - then you can set the verification parameter through it and use the other constructor for delegated creation - that way the DMA TLB will always retain it's state and there is no need to add an API.

device/pcie/pci_device.cpp
```diff
@@ -832,12 +833,37 @@ void PCIDevice::configure_tlb(const uint32_t tlb_index, const tlb_data &tlb_conf
...
+        static_cast<uint32_t>(lower_64), static_cast<uint32_t>(lower_64 >> 32), static_cast<uint32_t>(upper_64)};
+    const size_t num_config_words = arch == tt::ARCH::BLACKHOLE ? 3 : 2;
```
nbuncicTT: don't rely on operator precedence, use parenthesis

device/pcie/pci_device.cpp
```diff
@@ -832,12 +833,37 @@ void PCIDevice::configure_tlb(const uint32_t tlb_index, const tlb_data &tlb_conf
...
+
+    // MMIO writes are posted, so read the low 64 bits back to prove the mapping is live before
+    // anything uses it. Costs a PCIe round trip, hence opt-in.
+    if (verify) {
+        static constexpr auto TLB_CONFIG_BUSY_POLL_WINDOW =
+            std::chrono::duration_cast<std::chrono::microseconds>(timeout::MMIO_OP_TIMEOUT);
+        static constexpr auto TLB_CONFIG_POLL_INTERVAL = std::chrono::microseconds(1);
+
+        const bool config_landed = utils::poll_until(
+            [&]() { return tlb_reg_ptr[0] == config_words[0] && tlb_reg_ptr[1] == config_words[1]; },
```
nbuncicTT: Again, don't rely on operator precedence, state the intention of what should be evaluated first with parenthesis

nbuncicTT (commented):
Let's evaluate first if it's a big overhead with measurements, and then take appropriate action.

device/api/umd/device/pcie/tlb_handle.hpp
ajovanovicTT: Added delegating constructor in tlb_handle.hpp, and made changes accordingly, as we discussed offline.

ajovanovicTT:
### Measured the readback overhead on n150 WH.

Per reconfiguration, isolated on an idle 1 MiB window: 171 ns to 1262 ns, so 1.09 us.
The readback was satisfied by the first read in every configure, the poll never spins.

Stress test, 4 GiB written to a DRAM core over the DMA path. The buffered path
reconfigures the window once per DMA buffer chunk, and the buffer is 512 KiB, so
4 GiB is 8192 reconfigurations per direction. Window size does not matter here,
the chunk is the buffer, not the window. 8 interleaved reps per variant:

- Write: 0.859 s without readback to 0.885 s with the readback, 2.9 percent.
- Read: not resolvable, standard deviation 0.28 s across reps in every variant.
  Same 8192 reconfigurations at 1.09 us bound it under 1.6 percent.
- Small transfers pay a flat 2.15 us per call, so the relative cost is what varies:
  88 percent on a 4 B write, 17 percent at 64 KiB, 3 percent at 512 KiB.

MMIO 4 byte ops are unchanged, 400 ns versus 399 ns baseline. With the earlier
unconditional readback bug solution in configure_tlb they cost 1606 ns, 4x. Same DMA numbers,
so the opt-in is what buys the MMIO path.

Zero-copy DMA could not be measured on machine I was testing this on (no iommu). The buffered path measured
here reconfigures more often anyway, 8192 versus 4096 per 4 GiB.

device/tt_device/protocol/pcie_protocol.cpp
```diff
@@ -397,8 +397,11 @@ bool PcieProtocol::dma_transfer_zero_copy(
 
 TlbWindow* PcieProtocol::get_cached_dma_tlb_window(tlb_data config) {
     if (cached_dma_tlb_window_ == nullptr) {
-        cached_dma_tlb_window_ = std::make_unique<SiliconTlbWindow>(
-            pci_device_->allocate_tlb(get_dma_tlb_size(pci_device_->get_arch()), TlbMapping::WC), config);
+        // The DMA engine, not the host, reads through this window, so it is not ordered behind our
+        // config write - pass verify_config so every configure() confirms it landed first.
+        auto handle = pci_device_->allocate_tlb(
+            get_dma_tlb_size(pci_device_->get_arch()), TlbMapping::WC, /*verify_config=*/true);
```
nbuncicTT, line 403: nit: in `/*verify_config=*/true` no need for the `/*verify_config=*/` part, most modern IDEs do this automatically (clangd also does it for you).

Also, a bool is a bad representation for these kinds of situations, it's better to use an enum class, but as we are refactoring this for the base API spec, we will change this once more so no need to fix it here (no action needed).

device/pcie/pci_device.cpp
```diff
@@ -832,12 +835,37 @@ void PCIDevice::configure_tlb(const uint32_t tlb_index, const tlb_data &tlb_conf
...
-    tlb_reg_ptr[1] = static_cast<uint32_t>(lower_64 >> 32);
-
-    if (arch == tt::ARCH::BLACKHOLE) {
-        // Blackhole needs the upper 32 bits as well (96-bit total).
-        tlb_reg_ptr[2] = static_cast<uint32_t>(upper_64);
+    const std::array<uint32_t, 3> config_words = {
+        static_cast<uint32_t>(lower_64), static_cast<uint32_t>(lower_64 >> 32), static_cast<uint32_t>(upper_64)};
+    const size_t num_config_words = (arch == tt::ARCH::BLACKHOLE) ? 3 : 2;
+    for (size_t i = 0; i < num_config_words; i++) {
+        tlb_reg_ptr[i] = config_words[i];
```
nbuncicTT, line 842: We should be using .at() here instead of [], but this also would have a perf impact. I know you measured this and it work, but please make a draft follow-up PR, which will add this .at() so you can measure it and apply if acceptable - no action for this PR needed.

nbuncicTT (approved):
Well done

device/tt_device/protocol/pcie_protocol.cpp
ajovanovicTT: Ok, will addess this comment with Base API work.

device/pcie/pci_device.cpp
ajovanovicTT: Will do in a followup, since overhead from that change is not validated yet (I have missed to measure that at this point).

PR #3170, opened 2026-08-14
Remove init_pcie_iatus()

aleksamarkovicTT:
### Issue
#2747 

### Description
Remove init_pcie_iatus() call in Chip's start_device. New KMD supersedes the need to do this in userspace.

### Testing
CI

### API Changes
This PR has no API changes.


broskoTT (approved):
The change will work, but my original comment still holds: To logically properly match  you should remove all the branches where is_mapping_buffer_to_noc_supported is called (which will now always be true).

I'm fine with this change going in as is, but I think you can just remove KMD_MAP_TO_NOC and anything that follows from it as dead code, just delete the code until it compiles.

PR #3138, opened 2026-08-07
Delete Grayskull board types (E75, E150, E300)

vjymisal0:
### Issue
Closes #3137

### Description
Removes the Grayskull-specific board types E75, E150, and E300 from the `BoardType` enum, since Grayskull board support has been dropped from the codebase.

### List of the changes
- `device/api/umd/device/types/cluster_descriptor_types.hpp`: removed `E75`/`E150`/`E300` from the `BoardType` enum, removed their `board_type_name_map` and `board_type_canonical_name_map` entries, and re-numbered the remaining `static_assert` checks to stay contiguous from 0.
- `nanobind/py_api_basic_types.cpp`: removed the corresponding `E75`/`E150`/`E300` bindings from the Python `BoardType` enum.

A full-repo case-insensitive search for `E75`, `E150`, `E300`, and `Grayskull` was done. The only other `Grayskull` hits are in `scripts/iommu_detect.sh`, which mentions "Grayskull/Wormhole" generically in IOMMU-detection log messages about chip architecture, not the E75/E150/E300 board types — left untouched per the issue's scope, flagging here in case a maintainer wants a separate cleanup pass for that.

### Testing
Verified with a full-repo grep that no other source, test, or binding file references `E75`, `E150`, or `E300` after this change. Could not run the CMake/Ninja build in this environment (no `cmake`/`ninja` available, and this is Linux-only driver code); please let me know if CI surfaces anything and I'll follow up.

### API Changes
This PR has API changes: `BoardType::E75`, `BoardType::E150`, and `BoardType::E300` are removed, and the remaining enum values are renumbered (values are not stable ABI, per the existing `static_assert`s).

PR #3177, opened 2026-08-15
Delete Grayskull board types (E75, E150, E300)

### LINKED ISSUE #3137: Delete Grayskull board types
Delete all references to Grayskull board types: E75, E150, E300.

lukiod:
Closes #3137

Deletes the retired Grayskull board types `E75`, `E150`, and `E300` from `BoardType` and every place that referenced them:

- `device/api/umd/device/types/cluster_descriptor_types.hpp`: removed the three enum entries, the corresponding `static_assert`s, and their entries in `board_type_name_map` and `board_type_canonical_name_map`. The `static_assert`s for the remaining values are renumbered to match the enum's new implicit values (`N150` starts at 0).
- `nanobind/py_api_basic_types.cpp`: removed the matching `.value("E75"/"E150"/"E300", ...)` Python bindings.

No other file in the tree references `E75`, `E150`, or `E300` (grepped the full source tree, excluding `third_party/`), so this is a self-contained removal with no follow-up call sites.

Verified `device/api/umd/device/types/cluster_descriptor_types.hpp` compiles standalone with `g++ -std=c++20 -fsyntax-only` after the change (no other translation unit in the tree depends on the removed values, per the grep above).

### DIFF
--- a/device/api/umd/device/types/cluster_descriptor_types.hpp
+++ b/device/api/umd/device/types/cluster_descriptor_types.hpp
@@ -19,9 +19,6 @@
 namespace tt {
 
 enum BoardType : uint32_t {
-    E75,
-    E150,
-    E300,
     N150,
     N300,
     P100,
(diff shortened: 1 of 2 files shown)


aleksamarkovicTT:
Closing in favor of #3138

Issue #2745, opened 2026-06-01
Switch to KMD based locks


broskoTT:
**Description**
User space locking assumes correct management of /dev/shm memory between the guest and host systems (not VMs, but docker instances). Using KMD's LOCK_CTL would avoid that issue.

**Scope**
- Support and expose KMD locks
- Switch the current RobustMutex usages to use the new API

**Timeline**

**Is your feature request related to a problem? Please describe.**
Some additional context available in https://tenstorrent.slack.com/archives/C07L16Z59AP/p1779988273699179 and https://tenstorrent.slack.com/archives/C07L16Z59AP/p1779970786376479

**Definition of Done**
No need to share /dev/shm to have tt-telemetry working on the host and tt-metal working in docker.

broskoTT:
Also https://tenstorrent.slack.com/archives/C07L16Z59AP/p1777327724945219

broskoTT:
KMD locks:

don't require additional share like /dev/shm, just /dev/tenstorrent. Seems important for provisioning
are slower - will anayze by how much, but shouldn't be to impactful
all locks must be tied to a local device - shouldn't be a problem, will test this in following days

broskoTT:
Will be working on this very soon

broskoTT:
I wanted to check perf for locks.
Joel's comment was that I should use the KMDs convention for what locks are used for

kluongTT:
@broskoTT is out until after July I believe - @pjanevskiTT - any chance know much about this, or know if anyone else has interest in picking this up?

ppetrovicTT:
We've added some metrics relying on `CHIP_IN_USE` lock: https://github.com/tenstorrent/tt-ansible/commit/5fb6e33398b8418674dd0a2b43a5e77e979a770a
Please keep it in mind and let me know once these locks move to KMD, so that we can update our scripts.
Ideally, `tt-telemetry` should be able to read if the device is locked by a user.

pjanevskiTT:
@kluongTT sorry for the late reply. No one is planning on doing this at the moment. Are there any benefits to this for you? If so, we could prioritize this

kluongTT:
@pjanevskiTT - see this thread for a possible usecase - usage tracking: https://tenstorrent.slack.com/archives/C048YNQCXU0/p1784644340476759

locks are in `/dev/shm` and containers have their own `/dev/shm` namespace usually. we want to keep this isolation

broskoTT:
[...]

PR #3180, opened 2026-08-17
Remove usage of global ARC_MSG lock

broskoTT:
### Issue
On a path to https://github.com/tenstorrent/tt-umd/issues/2745

### Description
A follow up of https://github.com/tenstorrent/tt-umd/pull/3179
NOTE: Not to be merged until dependant PR hits all clients to ensure correctness.

### List of the changes
- Don't use global ARC_MSG lock in UMD

### Testing
Existing test, and specifically multiprocess tests should guarantee this still works as intended.

### API Changes
This PR has no API changes.


broskoTT:
Superseded by #3398, which does the same removal against the current code.

This PR's diff targets `device/arc/wormhole_arc_messenger.cpp`, which #3322 deleted when it routed all ARC commands through `DeviceFirmware::send_device_command`. That refactor also dropped the per device lock #3179 had added, so it had to be restored first: #3397 puts it back, and #3398 removes the global one on top of it.

Same gate as here - #3398 waits until #3397 has reached all clients.

PR #3182, opened 2026-08-17
KMDMutex init

broskoTT:
### Issue
#2745 

### Description
Perf analysis can be found here: https://tenstorrent.atlassian.net/wiki/spaces/UMD/pages/2564948067/Locks+perf+analysis
We are diverging away from requiring /dev/shm as it doesn't scale well on all environments, so we want to stop using RobustMutex on the main path.

This PR introduces KMDMutex class, and adds relevant tests which output some perf of locks

### List of the changes
- Introduce KMDMutex
- Add missing entry points to tt_kmd_lib
- Add test for different lock types. Note this includes FLock based mutex which I tested but don't want to commit the implementation to our codebase.

### Testing
This PR includes perf tests. For functional test, CI test in this tip of the stack: https://github.com/tenstorrent/tt-umd/pull/3233

### API Changes
This PR has API changes, introduces KMDMutex

### DIFF
--- a/device/api/umd/device/utils/kmd_mutex.hpp
+++ b/device/api/umd/device/utils/kmd_mutex.hpp
@@ -1,0 +1,119 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <sys/types.h>  // pid_t
+
+#include <chrono>
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <utility>
+
+namespace tt::umd {
+
+// Conventional KMD resource-lock indices. These mirror the TENSTORRENT_LOCK_INDEX_* defines in KMD's
+// ioctl.h: indices 0..15 are reserved by convention for ERISC cores. They are copied here (the KMD
+// header lives in a separate repo and cannot be included/linked from UMD); keep this in sync with
+// KMD. Callers coordinating their own resources should use indices outside this reserved range.
+enum class KmdLockIndex : uint8_t {
+    ETH00 = 0,
+    ETH01 = 1,
+    ETH02 = 2,
+    ETH03 = 3,
+    ETH04 = 4,
+    ETH05 = 5,
+    ETH06 = 6,
+    ETH07 = 7,
+    ETH08 = 8,
+    ETH09 = 9,
+    ETH10 = 10,
+    ETH11 = 11,
+    ETH12 = 12,
+    ETH13 = 13,
+    ETH14 = 14,
+    ETH15 = 15,
+};
+
+// KmdMutex is a cross-process lock backed by the Tenstorrent kernel-mode driver (KMD) resource locks
+// (TENSTORRENT_IOCTL_LOCK_CTL). See device/utils/README.md for how the available locking backends
+// compare.
+//
+// Properties:
+//   - The lock is tied to the device itself, not to a host filesystem. Any process that can open
+//     /dev/tenstorrent/<N> contends over the same lock, even across different containers or mount
+//     namespaces, with no /dev/shm (or other filesystem) sharing required. This makes it the right
+//     primitive for serializing whole workloads against one device.
+//   - Every operation is an ioctl (a syscall); there is no userspace fast path. A contended lock()
+//     polls the non-blocking acquire with a short backoff (the ioctl set has no efficient blocking
+//     wait), so waiting for a held lock costs more than a single syscall.
+//   - The lock is held by the open file descriptor, and KMD releases all of an fd's locks when the
+//     fd is closed. Since the kernel closes every fd of a process when it dies, a crashed holder
+//     cannot leak the lock - it is reclaimed automatically.
+//   - Scope is limited to a single local device: there is no global lock that spans devices.
+//
+// Each KMD device exposes TENSTORRENT_RESOURCE_LOCK_COUNT (64) independent lock indices (see
+// KmdLockIndex for the reserved ones).
+//
+// This class owns its own dedicated chardev fd, separate from any fd used to run a workload. That
+// is deliberate: a device reset invalidates every fd that was open across it (further ioctls return
+// ENODEV), but the lock itself survives the reset and stays held until its fd is closed. Keeping the
+// lock on a dedicated fd lets a caller hold the lock, reset the device, then open fresh fds to run a
+// workload. lock() transparently reopens its fd and retries if it observes that a reset happened
+// while it was waiting.
+//
+// It meets the C++ Lockable requirement (lock()/try_lock()/unlock()), so it can be used with RAII
+// helpers like std::lock_guard and std::unique_lock.
+class KmdMutex {
+public:
+    // @param pci_device_num  N in /dev/tenstorrent/N (a UMD logical device id / PCIDevice number).
+    // @param lock_index      KMD resource lock index in [0, 64). Avoid 0..15 (reserved, see
+    //                        KmdLockIndex).
+    KmdMutex(int pci_device_num, uint8_t lock_index);
+    ~KmdMutex() noexcept;
+
+    // Opens the dedicated chardev fd used to hold the lock, after verifying the installed KMD supports
+    // resource locks (throws otherwise). Must be called before lock()/try_lock(). Kept separate from
+    // the constructor so that failures during setup are still cleaned up by the destructor.
+    void initialize();
+
+    // Move-only so it can live in STL containers. Copying would alias the owned fd.
+    KmdMutex(KmdMutex&& other) noexcept;
+    KmdMutex& operator=(KmdMutex&& other) noexcept;
+    KmdMutex(const KmdMutex&) = delete;
+    KmdMutex& operator=(const KmdMutex&) = delete;
+
+    // Blocks until the lock is acquired. If a reset is detected while waiting (the fd becomes
+    // unusable), the fd is transparently reopened and the wait is retried.
+    void lock();
+
+    // Attempts to acquire without blocking. Returns true if acquired, false if held by another fd.
+    bool try_lock();
+
+    // Tries to acquire the lock (immediately if timeout is zero, otherwise polling until timeout).
+    // Returns std::nullopt if the lock was acquired (the caller now holds it). On contention returns a
+    // {pid, tid} pair, but KMD does not expose the owner, so it is always {0, 0} here - the pair's
+    // presence signals "held by someone else", nothing more.
+    std::optional<std::pair<pid_t, pid_t>> probe_lock(std::chrono::seconds timeout = std::chrono::seconds(0));
+
+    // Releases the lock. It is also released automatically when this object (and thus its fd) is
+    // destroyed, or when the owning process exits.
+    void unlock();
+
+    // Best-effort query of whether the lock is currently held by any fd (including this one). This
+    // is inherently racy and intended for debugging/diagnostics only.
+    bool is_locked_by_anyone();
+
+private:
+    void open_device_fd();
+    void close_fd() noexcept;
+
+    int pci_device_num_;
+    uint8_t lock_index_;
+    int fd_ = -1;
+    std::string device_path_;
+};
+
+}  // namespace tt::umd
--- a/device/api/umd/device/utils/kmd_mutex.hpp
+++ b/device/api/umd/device/utils/kmd_mutex.hpp
@@ -12,6 +12,10 @@
 #include <string>
 #include <utility>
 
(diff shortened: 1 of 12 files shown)


device/api/umd/device/utils/kmd_mutex.hpp
```diff
@@ -0,0 +1,124 @@
...
+struct tt_device_t;
+
+namespace tt::umd {
+
+// Conventional KMD resource-lock indices. These mirror the TENSTORRENT_LOCK_INDEX_* defines in KMD's
+// ioctl.h: indices 0..15 are reserved by convention for ERISC cores. They are copied here (the KMD
+// header lives in a separate repo and cannot be included/linked from UMD); keep this in sync with
+// KMD. Callers coordinating their own resources should use indices outside this reserved range.
+enum class KmdLockIndex : uint8_t {
+    ETH00 = 0,
```
pjanevskiTT: maybe we should move this somewhere where we put ioctl.h so we keep these files in sync?

device/api/umd/device/utils/kmd_mutex.hpp
```diff
@@ -0,0 +1,124 @@
...
+// include path, which UMD links privately.
+struct tt_device_t;
+
+namespace tt::umd {
+
+// Conventional KMD resource-lock indices. These mirror the TENSTORRENT_LOCK_INDEX_* defines in KMD's
+// ioctl.h: indices 0..15 are reserved by convention for ERISC cores. They are copied here (the KMD
+// header lives in a separate repo and cannot be included/linked from UMD); keep this in sync with
+// KMD. Callers coordinating their own resources should use indices outside this reserved range.
+enum class KmdLockIndex : uint8_t {
```
pjanevskiTT: do we need these constants here, if we are not using it? Maybe just keeping a constant on number of reserved locks would be fine?

tests/microbenchmark/host_benchmark/test_locks.cpp
```diff
@@ -0,0 +1,280 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Microbenchmarks comparing the available UMD locking backends:
+//   - RobustMutex     : pthread robust mutex in a /dev/shm file (userspace fast path).
+//   - KmdMutex        : KMD resource lock via ioctl (syscall per op, tied to a silicon device).
+//
+// FlockMutex (flock(2) over a /dev/shm file) was a third backend; its implementation has been
```
pjanevskiTT: I would probably remove this. I would understand if we had an implementation of FlockMutex. But if we don't have it I would remove it from the benchmark as well. If we ever add any other type of mutex it is easy to add it to make it similar to robust and kmd. This way if we change anything about robust and kmd mutex, we need to make sure comments for flock mutex are in sync

device/utils/kmd_mutex.cpp
```diff
@@ -0,0 +1,177 @@
...
+#include "tt-kmd-lib/tt_kmd_lib.h"
+#include "umd/device/utils/error.hpp"
+
+namespace tt::umd {
+
+KmdMutex::KmdMutex(int pci_device_num, uint8_t lock_index) :
+    pci_device_num_(pci_device_num),
+    lock_index_(lock_index),
+    device_path_(fmt::format("/dev/tenstorrent/{}", pci_device_num)) {
+    UMD_ASSERT(
```
pjanevskiTT, line 27: do we want to maybe log warning if someone locks on the ETH indices? Or we are going to use it for our own ETH cores?

pjanevskiTT (approved):

device/api/umd/device/utils/kmd_mutex.hpp
```diff
@@ -0,0 +1,124 @@
...
+// the reserved ones).
+//
+// This class owns its own dedicated device handle (chardev fd), separate from any fd used to run a
+// workload. That is deliberate: a device reset invalidates every fd that was open across it (further
+// ioctls return ENODEV), but the lock itself survives the reset and stays held until its fd is
+// closed. Keeping the lock on a dedicated handle lets a caller hold the lock, reset the device, then
+// open fresh fds to run a workload. lock() transparently reopens its handle and retries if it
+// observes that a reset happened while it was waiting.
+//
+// It meets the C++ Lockable requirement (lock()/try_lock()/unlock()), so it can be used with RAII
```
pjanevskiTT, line 54: The lock is owned by the *file descriptor*, so it's worth stating explicitly whether one `KmdMutex` shared between threads of a process actually excludes them. `tt_lock_acquire` on a handle that already holds the index presumably reports acquired, in which case two `std::lock_guard`s on the same `KmdMutex` in two threads would both proceed.

This matters for the stated goal of moving off `RobustMutex`: `LockManager` keeps one mutex object per (type, device) and hands `std::unique_lock<RobustMutex>` to every thread in the process, relying on the pthread mutex for thread-level exclusion. A drop-in `KmdMutex` would lose that unless it holds a process-local `std::mutex` alongside the KMD lock. The benchmark sidesteps this by giving each thread its own instance/fd, so nothing here exercises it.

Either document "one instance per thread, no intra-process exclusion", or add the process-local mutex.

device/utils/kmd_mutex.cpp
```diff
@@ -0,0 +1,177 @@
...
+}
+
+void KmdMutex::lock() {
+    UMD_ASSERT(device_ != nullptr, error::RuntimeError, "KmdMutex::lock() called before initialize()");
+
+    // KMD exposes only non-blocking acquire, so a blocking lock is implemented by polling it with a
+    // short backoff. This also deliberately avoids KMD's blocking-acquire path, which has had a
+    // deadlock against the reset ioctl. try_lock() already handles a reset (ENODEV) by reopening the
+    // handle.
+    while (!try_lock()) {
```
pjanevskiTT, line 134: `lock()` polls forever with no diagnostic. `RobustMutex::lock()` does a 1-second timed attempt first and logs `Waiting for lock '{}' which is currently held by thread TID: {}, PID: {}` before blocking, so a stuck lock is visible in the log; here a caller that never gets the lock just hangs silently. A periodic warning (even without an owner pid, since KMD doesn't expose one) would keep the two backends comparable for debugging.

Separately, the class comment and README both describe a "short backoff", but this is a fixed 1 ms sleep — either implement a backoff or reword.

device/api/umd/device/utils/kmd_versions.hpp
```diff
@@ -53,4 +53,11 @@ inline constexpr SemVer KMD_TLB_DMABUF_EXPORT = SemVer(2, 10, 0, 1);
  * introduced in that release.
  */
 inline constexpr SemVer MIN_KERNEL_TLB_DMABUF_EXPORT = SemVer(5, 8, 0);
+
+/**
+ * KMD version 1.26.0 introduced the TENSTORRENT_IOCTL_LOCK_CTL IOCTL, a generic per-device resource
+ * locking primitive (acquire/release/test of one of TENSTORRENT_RESOURCE_LOCK_COUNT named locks).
+ * KmdMutex requires at least this version.
+ */
+inline constexpr SemVer KMD_RESOURCE_LOCKS = SemVer(1, 26, 0);
```
pjanevskiTT: `KMD_RESOURCE_LOCKS` isn't referenced anywhere in the PR — `KmdMutex` never checks it. On KMD older than 1.26.0 the `LOCK_CTL` ioctl fails and the user gets `tt_lock_acquire() failed for lock 32 on /dev/tenstorrent/0 errno: 25` rather than "KMD too old". Worth checking it in `initialize()` and throwing with this constant in the message; otherwise the constant is unused.

broskoTT: Yes this was leftover...

device/utils/kmd_mutex.cpp
broskoTT: Those are just suggested indexes for such usage. They are also meant to be used by us as well as by other clients. Nothing is preventing us from using them.

device/api/umd/device/utils/kmd_mutex.hpp
broskoTT: Fair question, but the lock is not arbitrated on the file descriptor. KMD does `test_and_set_bit` on
a **device wide** bitmap, and only records the fd afterwards (chardev.c, `LOCK_CTL_ACQUIRE`):

```c
if (!test_and_set_bit(in.index, priv->device->resource_lock)) {
    set_bit(in.index, priv->resource_lock);
    out.value = 1;  // Acquired
} else {
    out.value = 0;  // Failed to acquire
}
```

So an acquire fails while the lock is held, including from the handle already holding it. Two threads
on one `KmdMutex` do exclude each other, and no process local `std::mutex` is needed.

There is an asymmetry, on the other operation: `RELEASE` clears the **per fd** bit, so a thread can
release a lock a different thread took - KMD tracks handles, not threads. `std::lock_guard` keeps that
from happening.

Both are in the class comment now, and you were right that nothing exercised this - three tests added
in `test_kmd_mutex.cpp`.

device/utils/kmd_mutex.cpp
```diff
@@ -0,0 +1,177 @@
...
+    // If the handle died due to a reset, the lock is still held in KMD until the handle is closed;
+    // the destructor's close will release it, so this is not fatal.
+    if (result == -ENODEV) {
+        log_warning(
+            tt::LogUMD,
+            "tt_lock_release() for lock {} on {} hit ENODEV (device reset); lock will be released on close",
+            lock_index_,
+            device_path_);
+        return;
+    }
```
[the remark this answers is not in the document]
broskoTT: Agreed, fixed - the handle is closed and reopened there, which is what hands the lock back.

It mattered more than it looks: locks live for the lifetime of the process, so "released when the
handle closes" meant "released at process exit", blocking everyone else until then.

device/utils/kmd_mutex.cpp
broskoTT: Both fixed. `lock()` warns once after a second, so a lock nobody releases shows up in the log instead
of looking like a hang. No owner to name - KMD does not expose one.

A real blocking wait would be better, but tt-kmd-lib only wires up the non-blocking acquire, and
KMD's `ACQUIRE_BLOCKING` is the path we deliberately avoid over the reset deadlock. So it keeps
polling, quietly. The wording was just wrong - it is a fixed 1 ms sleep, not a backoff - reworded in
the header and README.

device/api/umd/device/utils/kmd_versions.hpp
broskoTT: Removed. Wiring it up would not have helped anyway: `KMD_MINIMUM_VERSION` is 2.0.0 and this was
1.26.0, so UMD already refuses every KMD that lacks the ioctl.

device/utils/kmd_mutex.cpp
```diff
@@ -0,0 +1,177 @@
...
+    // O_APPEND tells KMD (>= 2.6) not to request high power on open; a lock-only handle should not
+    // change the device's power state. The flag is harmless on older KMD versions.
+    int result = tt_device_open(device_path_.c_str(), &device_, O_APPEND);
+    UMD_ASSERT(
+        result == 0,
+        error::RuntimeError,
+        fmt::format("tt_device_open() failed for KMD lock device {} errno: {}", device_path_, -result));
+}
+
+void KmdMutex::initialize() { open_device(); }
```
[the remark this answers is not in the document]
broskoTT: Guarded - `initialize()` returns if the handle is already open. Not reachable through `LockManager`,
but it costs nothing.

tests/microbenchmark/host_benchmark/test_locks.cpp
broskoTT: Removed, all of it. Git history has it if we ever want it back.

device/utils/kmd_mutex.cpp
broskoTT: No change, per your answer on the thread.

device/api/umd/device/utils/kmd_mutex.hpp
broskoTT: Removed the enum - no users, and `tt_kmd_lib.h` already documents the index range next to
`TT_RESOURCE_LOCK_COUNT`, so there is no second copy to keep in sync. Which indices UMD takes, and
why, is documented where they are chosen (#3232).

device/api/umd/device/utils/kmd_mutex.hpp
broskoTT: Covered by the reply on the thread just above: the enum is removed, and `tt_kmd_lib.h` is the one place that tracks KMD, so there is nothing left to keep in sync.

Issue #2873, opened 2026-06-22
Migrate ArchitectureImplementation interface to Base API Specification


nbuncicTT:
**Description**

`architecture_implementation` exists but needs renaming to `ArchitectureImplementation` and trimming to match the Base API Specification. Methods that don't belong here should be migrated to their appropriate components. The internal implementation may retain additional APIs, but those should only be accessible within the concrete `TTDeviceModel`, not exposed externally.

**Scope**

- Rename `architecture_implementation` to `ArchitectureImplementation`
- Trim public interface to match the Base API Specification
- Keep internal-only APIs accessible only within concrete `TTDeviceModel` implementations
- Migrate excess methods to appropriate components
- Update all call sites and implementations

**Timeline**

No hard deadline. Part of the UMD Base API migration epic.

**Is your feature request related to a problem? Please describe.**

The current interface is bloated with methods that belong in other components, making the interface boundary unclear.

**Definition of Done**

- `ArchitectureImplementation` public interface matches the Base API Specification
- Internal-only APIs not exposed outside of `TTDeviceModel`
- Excess methods migrated to appropriate components
- Existing tests pass or are updated

PR #3196, opened 2026-08-19
Move ARC register offsets to ArchitectureRegisters

broskoTT:
### Issue

### Description
Stacked on #3218.

The ARC register offsets and message ids are not part of the Base API specification. The
offsets join the other register addresses in `ArchitectureRegisters`, so there is still one
struct per architecture filled in one place. Message ids need nothing new, the per architecture
`arc_message_type` enums already exist and are already used around these call sites.

`probe_arc` was the only thing in the base `TTDevice` which needed these, so it becomes a hook
the concrete devices implement.

### List of the changes
- Add the APB BAR0 offset, CSM mailbox BAR0 offset, the two reset unit scratch offsets and the
  APB NOC base address to `ArchitectureRegisters`.
- Move `probe_arc` from `TTDevice` into the concrete devices, which hold their register map.
- Remove the ten ARC offset and message getters from `ArchitectureImplementation` and update all
  call sites. The concrete devices already read some of these from their own constants, so those
  lines go through the struct too.
- The tt_device example only needed some BAR readable address, so it no longer derives one.

### Testing
Code builds. Existing CI covers ARC init and warm reset paths on both architectures.

### API Changes
This PR has API changes, ten methods removed from `ArchitectureImplementation` and ARC entries
added to `ArchitectureRegisters`.

### Full stack diff

### DIFF
--- a/device/api/umd/device/arch/architecture_registers.hpp
+++ b/device/api/umd/device/arch/architecture_registers.hpp
@@ -21,6 +21,13 @@ using NocRegAddrResolver = uint64_t (*)(CoreType core_type, uint32_t noc, uint32
 // different addresses. Architecture specific code reads its own constants directly; this exists for
 // callers which are not tied to one architecture.
 struct ArchitectureRegisters {
+    uint32_t arc_apb_bar0_offset;
+    uint32_t arc_csm_bar0_mailbox_offset;
+    // ARC reset unit scratch registers, relative to the APB window.
+    uint32_t arc_reset_scratch_offset;
+    uint32_t arc_reset_scratch_2_offset;
+    uint64_t arc_apb_noc_base_address;
+
     // BAR0 offset of the NOC0 node id register. Reachable before the NOC is up, which is what makes
     // it usable as the bus hang check.
     uint32_t noc_node_id_bar_offset;
--- a/device/api/umd/device/tt_device/tt_device.hpp
+++ b/device/api/umd/device/tt_device/tt_device.hpp
@@ -576,9 +576,11 @@ protected:
 
     virtual void set_arc_coordinate() {}
 
-private:
-    void probe_arc();
+    // TODO: temporary. The register to probe is architecture specific, so only the concrete devices
+    // can implement it. Goes away once DeviceFirmware::init_firmware owns ARC startup.
+    virtual void probe_arc() {}
 
+private:
     void log_aiclk_timeout_warning(uint32_t target_aiclk, std::chrono::milliseconds timeout_ms);
 
     void assign_soc_arch_descriptor(const std::shared_ptr<SocArchDescriptor> &soc_arch_descriptor);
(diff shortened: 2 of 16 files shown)


device/api/umd/device/arch/architecture_registers.hpp
```diff
@@ -21,6 +21,15 @@ using NocRegAddrResolver = uint64_t (*)(CoreType core_type, uint32_t noc, uint32
 // different addresses. Architecture specific code reads its own constants directly; this exists for
 // callers which are not tied to one architecture.
 struct ArchitectureRegisters {
+    // ARC register windows. BAR0 offsets are usable before the device is initialized, the NOC base
```
nbuncicTT: nit: Omit the comment as this is an implementation detail, and doesn't hold true for a specific board implementation of blackhole.

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -576,9 +576,9 @@ class TTDevice {
 
     virtual void set_arc_coordinate() {}
 
-private:
-    void probe_arc();
+    virtual void probe_arc() {}
```
nbuncicTT, line 581: You've changed the probe_arc from public to private and made it virtual - which is what we are trying to remove explicitly. Why is this necessary? If it's temporary, please put a comments over it.

The point is that `init_firmware()` from `DeviceFirmware` should meddle with the `probe_arc`.

nbuncicTT (approved):

tests/api/test_warm_reset.cpp
```diff
@@ -419,8 +432,8 @@ TEST_P(ClusterWarmResetScratchMethodTest, ClusterWarmResetScratch) {
     auto tt_device = cluster->get_chip(chip_id)->get_tt_device();
 
     tt_device->bar_write32(
-        tt_device->get_architecture_implementation()->get_arc_axi_apb_peripheral_offset() +
-            tt_device->get_architecture_implementation()->get_arc_reset_scratch_2_offset(),
+        get_arc_address_map(tt_device->get_arch()).apb_bar0_offset +
+            get_arc_address_map(tt_device->get_arch()).reset_scratch_2_offset,
         write_test_data);
```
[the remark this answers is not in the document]
broskoTT: Hoisted the lookup: all four sites now go through a single `arc_scratch_2_bar_address(tt_device)` helper, so one map lookup per site instead of one per field.

Skipped the architecture gate. `get_architecture_registers()` has an explicit `QUASAR` case, so it only throws for `ARCH::Invalid`, which a warm reset test on silicon cannot hit. (The name in the comment is also stale after the rebase, it is `get_architecture_registers()` now, not `get_arc_address_map()`.)

device/api/umd/device/arch/architecture_registers.hpp
broskoTT: Removed. Kept only the note that the scratch offsets are relative to the APB window, since that is needed to use the fields correctly rather than a claim about behavior.

device/api/umd/device/tt_device/tt_device.hpp
broskoTT: Added a TODO above it naming `DeviceFirmware::init_firmware` as the eventual owner, so the reason it exists and when it goes away are both written down.

Small correction: it went from private to protected, not public to private. It had to become overridable because the register to probe is architecture specific, so the base class can no longer implement it once the offset leaves `ArchitectureImplementation`.

PR #3197, opened 2026-08-19
[Bug fix] Route TTSim TLB window reads and writes through arch-aware addressing so Quasar targets cores by coordinate

sagarwalTT:
### Description
On Quasar, the simulated TLB handle has no base address — it does not program TLB registers the way Wormhole and Blackhole do — so computing a physical address from it and sending that to the simulator does not reliably identify the intended destination core.

This change adds a per-architecture decision inside the memory-access path that chooses whether to address the simulator by physical address (Wormhole, Blackhole) or by explicit core coordinate plus device address (Quasar). All read and write entry points on the TLB window now route through the two block-transfer methods that make this decision, rather than each computing a physical address independently, so the addressing rules live in one place going forward.

### List of the changes
- The per-architecture choice between physical-address addressing and core-coordinate addressing is now made inside the existing `write_block` and `read_block` methods.
- The remaining read and write entry points (16-bit, 32-bit, and register transfers) now delegate to `write_block`/`read_block` instead of each computing a physical address directly.
- For Quasar, addressing now uses the core's row and column coordinates (`x_end`, `y_end`) together with a device address derived from the TLB configuration, sent via `tile_write_bytes`/`tile_read_bytes`.
- Behavior for Wormhole and Blackhole is unchanged; both continue on the existing physical-address path (`pci_mem_write_bytes`/`pci_mem_read_bytes`).
- The Quasar path is selected by an architecture check rather than a capability check, because the simulator does not currently model Quasar TLBs at all; this should become a capability check if that ever changes.

### Testing
Quasar fast dispatch tests in `tt-metal` that utilize the full Quasar 8x4 grid on craq-sim were manually run with this change.

### API Changes
None

### DIFF
--- a/device/api/umd/device/pcie/tt_sim_tlb_window.hpp
+++ b/device/api/umd/device/pcie/tt_sim_tlb_window.hpp
@@ -49,6 +49,14 @@ private:
      */
     uint64_t get_physical_address(uint64_t offset) const;
 
+    /**
+     * Single entry points for all reads and writes, so the choice between
+     * addressing the simulator by physical address and addressing it by core
+     * coordinate is made in exactly one place.
+     */
+    void translate_and_write(uint64_t offset, const void* data, size_t size);
+    void translate_and_read(uint64_t offset, void* data, size_t size);
+
     TTSimCommunicator* sim_communicator_;
 };
 
--- a/device/api/umd/device/pcie/tt_sim_tlb_window.hpp
+++ b/device/api/umd/device/pcie/tt_sim_tlb_window.hpp
@@ -49,14 +49,6 @@ private:
      */
     uint64_t get_physical_address(uint64_t offset) const;
 
-    /**
-     * Single entry points for all reads and writes, so the choice between
-     * addressing the simulator by physical address and addressing it by core
-     * coordinate is made in exactly one place.
-     */
-    void translate_and_write(uint64_t offset, const void* data, size_t size);
-    void translate_and_read(uint64_t offset, void* data, size_t size);
-
     TTSimCommunicator* sim_communicator_;
 };
 
--- a/device/pcie/tt_sim_tlb_window.cpp
+++ b/device/pcie/tt_sim_tlb_window.cpp
@@ -57,9 +57,7 @@ void TTSimTlbWindow::translate_and_read(uint64_t offset, void* data, size_t size
         static_cast<uint32_t>(size));
 }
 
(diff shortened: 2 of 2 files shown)


device/api/umd/device/pcie/tt_sim_tlb_window.hpp
```diff
@@ -49,6 +49,14 @@ class TTSimTlbWindow : public TlbWindow {
      */
     uint64_t get_physical_address(uint64_t offset) const;
 
+    /**
+     * Single entry points for all reads and writes, so the choice between
+     * addressing the simulator by physical address and addressing it by core
+     * coordinate is made in exactly one place.
+     */
+    void translate_and_write(uint64_t offset, const void* data, size_t size);
```
nbuncicTT: Rather than adding `translate_and_write`/`translate_and_read`, put the arch branch directly in `write_block`/`read_block` and make the other six entry points one-line forwarders to them. They already have the (offset, data, size) signature the helpers were introduced to provide, and all eight are byte-identical in this class today, so this gets the same single decision point with no new declarations in `device/api/`, dropping the API Changes note from the PR. The `write_register`/`write_block` distinction is real on silicon but unrepresentable over `tile_*_bytes`, so collapsing it here loses nothing.

device/pcie/tt_sim_tlb_window.cpp
```diff
@@ -20,49 +21,69 @@ TTSimTlbWindow::TTSimTlbWindow(
...
 
-void TTSimTlbWindow::write16(uint64_t offset, uint16_t value) {
-    validate(offset, sizeof(uint16_t));
-    sim_communicator_->pci_mem_write_bytes(get_physical_address(offset), &value, sizeof(uint16_t));
+void TTSimTlbWindow::translate_and_write(uint64_t offset, const void* data, size_t size) {
+    validate(offset, size);
+    // Quasar's TLB handle programs no TLB registers and is left without a base address, so a
+    // physical address derived from it carries no identity for the core this window was created
+    // for. Name the core explicitly. Wormhole and Blackhole keep the physical-address path.
+    if (tlb_handle->get_arch() != tt::ARCH::QUASAR) {
```
nbuncicTT, line 52: Please note in the comment that Quasar TLBs aren't modeled by the simulator, so if that ever changes, this needs to become a capability check rather than an arch check.

nbuncicTT (approved):

PR #3204, opened 2026-08-20
[Base API] Make BlackholeArcMessageQueue depend on protocol interfaces

nbuncicTT:
## Issue

`BlackholeArcMessageQueue` took a `TTDevice*` and read the thread-local NOC selection directly, which made it untestable and kept hidden global state on the ARC messaging path.

## Description

The queue now takes what it actually uses: a `DeviceProtocol` for the data path and a `BlackholeArcApb` for the firmware interrupt. `BlackholeArcMessenger`, its only caller, supplies them.

Four reads of the thread-local NOC (`get_selected_noc_id()` / `is_selected_noc1()`) move out of the queue and become a parameter. The messenger reads the thread-local at its own entry points, so the remaining dependency on global state is explicit and confined to the old path.

Intended to be behavior-neutral: `read_from_device` -> `read_data` is equivalent since `resolve_coordinate` is a pass-through for `LITERAL` coordinates, and the JTAG branch swaps `IODeviceType::JTAG` for a non-null `JtagInterface` under the invariant that a `TTDevice` is built for exactly one protocol.

## List of the changes

- Queue constructor and factory take `DeviceProtocol*`, `JtagInterface*`, `BlackholeArcApb*`, and the NOC translation state instead of a `TTDevice*`.
- `send_message` and the internals take the NOC to route over.
- `BlackholeArcMessenger` builds a `BlackholeArcApb` and passes the interfaces down.

## Testing

Covered by the tests two PRs up: a full `send_message` round trip against a fake ARC memory (request entry layout, both ring pointers advancing, firmware interrupt, argument limit, error status) plus descriptor decode over PCIe and JTAG. Full baremetal suite passes.

## API Changes

- `BlackholeArcMessageQueue`'s constructor, `send_message`, and `get_blackhole_arc_message_queue` signatures change. Both are public headers, but `BlackholeArcMessenger` is the only in-tree caller.

One deliberate semantic change: the ARC core coordinate used to be resolved once at queue creation while reads used the thread-local NOC per call, so a NOC change after creation would send a NOC0-resolved coordinate over NOC1. It is now resolved per call from the NOC parameter, which removes that mismatch.

Stacked on #3203.

### DIFF
--- a/device/arc/blackhole_arc_messenger.cpp
+++ b/device/arc/blackhole_arc_messenger.cpp
@@ -9,7 +9,10 @@
 #include <chrono>
 #include <tt-logger/tt-logger.hpp>
 
+#include "noc_access.hpp"
 #include "umd/device/pcie/pci_device.hpp"
+#include "umd/device/tt_device/protocol/jtag_interface.hpp"
+#include "umd/device/tt_device/protocol/pcie_interface.hpp"
 #include "umd/device/tt_device/tt_device.hpp"
 #include "umd/device/types/blackhole_arc.hpp"
 #include "umd/device/utils/lock_manager.hpp"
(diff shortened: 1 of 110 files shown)


device/arc/blackhole_arc_messenger.cpp
```diff
@@ -9,16 +9,33 @@
...
+        tt_device->get_device_protocol(), pcie_interface, jtag_interface, tt_device->get_architecture_implementation());
+
     blackhole_arc_msg_queue = BlackholeArcMessageQueue::get_blackhole_arc_message_queue(
-        tt_device, BlackholeArcMessageQueueIndex::APPLICATION);
+        tt_device->get_device_protocol(),
+        jtag_interface,
+        arc_apb.get(),
+        tt_device->get_noc_translation_enabled(),
+        BlackholeArcMessageQueueIndex::APPLICATION,
+        get_selected_noc_id());
```
broskoTT, line 37: So this is now pushed up the stack, nice.

device/arc/blackhole_arc_messenger.cpp
```diff
@@ -9,16 +9,33 @@
...
 #include "umd/device/pcie/pci_device.hpp"
+#include "umd/device/tt_device/protocol/jtag_interface.hpp"
+#include "umd/device/tt_device/protocol/pcie_interface.hpp"
 #include "umd/device/tt_device/tt_device.hpp"
 #include "umd/device/types/blackhole_arc.hpp"
 #include "umd/device/utils/lock_manager.hpp"
 
 namespace tt::umd {
 
 BlackholeArcMessenger::BlackholeArcMessenger(TTDevice* tt_device) : ArcMessenger(tt_device) {
```
broskoTT, line 22: I assume in one of the following PRs this component will receive the protocols, arch implementation, etc?

broskoTT (approved):

device/arc/blackhole_arc_messenger.cpp
nbuncicTT: Yes

PR #3206, opened 2026-08-20
[Base API] Add tests for BlackholeArcApb and BlackholeArcMessageQueue

nbuncicTT:
## Issue

Neither component was covered. The ARC message queue in particular was untestable before it stopped taking a `TTDevice`.

## Description

Both components now take protocol interfaces, so they can be driven against gmock fakes with no card and no simulation build. The tests live in the baremetal lane alongside the other cardless tests.

## List of the changes

- `test_blackhole_arc_apb.cpp`: all three routes an access can take (JTAG, over the NOC when the ARC tile is not reachable over AXI, and the PCIe BAR) for both read and write; that the ARC core and NOC are taken per call; and that an out-of-range offset is rejected.
- `test_blackhole_arc_message_queue.cpp`: a full `send_message` round trip against a fake ARC memory, checking the request entry layout, both ring pointers advancing, the firmware interrupt hitting `ARC_FW_INT_ADDR`, the seven-argument limit, and the error status path. Plus descriptor decode in the factory over both PCIe and JTAG.

## Testing

15 new tests pass; the full baremetal suite is 187 tests and passes.

## API Changes

None. Tests only.

Stacked on #3205.

### DIFF
--- a/tests/baremetal/test_blackhole_arc_apb.cpp
+++ b/tests/baremetal/test_blackhole_arc_apb.cpp
@@ -17,8 +17,10 @@ using ::testing::Return;
 
 namespace {
 
-// The PCIe tile's x-coordinate decides whether the ARC tile is reachable over AXI: 11 means it is,
-// anything else means the access has to go over the NOC.
+// Blackhole has two PCIe tiles and only one of them is wired to ARC over AXI, so which tile the
+// board brought up decides whether an ARC access can bypass the NOC. The tile identifies itself by
+// its x-coordinate: blackhole::PCIE_CORES_TYPE1_NOC0 at x=11 has the direct AXI path, while
+// blackhole::PCIE_CORES_TYPE2_NOC0 at x=2 does not and has to go over the NOC.
 constexpr uint32_t PCIE_X_ARC_OVER_AXI = 11;
 constexpr uint32_t PCIE_X_ARC_OVER_NOC = 2;
 
--- a/tests/baremetal/test_blackhole_arc_apb.cpp
+++ b/tests/baremetal/test_blackhole_arc_apb.cpp
@@ -28,14 +28,14 @@ const tt_xy_pair ARC_CORE = {8, 0};
 class BlackholeArcApbTest : public ::testing::Test {
 protected:
     void expect_pcie_x_coordinate(uint32_t x) {
-        EXPECT_CALL(pcie_, bar_read32(arch_.get_read_checking_offset())).WillRepeatedly(Return(x));
+        EXPECT_CALL(pcie_, bar_read32(blackhole::NIU_CFG_NOC0_BAR_PCIE_ADDR + blackhole::NOC_NODE_ID_OFFSET))
+            .WillRepeatedly(Return(x));
     }
 
-    uint64_t noc_address() const { return arch_.get_arc_apb_noc_base_address() + APB_OFFSET; }
+    uint64_t noc_address() const { return blackhole::ARC_NOC_XBAR_ADDRESS_START + APB_OFFSET; }
 
     uint32_t bar_address() const { return blackhole::ARC_APB_BAR0_XBAR_OFFSET_START + APB_OFFSET; }
 
-    blackhole_implementation arch_;
     NiceMock<MockDeviceProtocol> protocol_;
     NiceMock<MockPcieInterface> pcie_;
     NiceMock<MockJtagInterface> jtag_;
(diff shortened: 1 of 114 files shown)


tests/baremetal/test_blackhole_arc_apb.cpp
```diff
@@ -0,0 +1,139 @@
...
+using namespace tt::umd;
+using namespace tt::umd::test_utils;
+using ::testing::_;
+using ::testing::NiceMock;
+using ::testing::Return;
+
+namespace {
+
+// The PCIe tile's x-coordinate decides whether the ARC tile is reachable over AXI: 11 means it is,
+// anything else means the access has to go over the NOC.
```
broskoTT: Reading this comment confused me more about this value that if the comment was left out, the var name told me more about what it is.
The comment should not explain the implementation, should explain why.
I'd mention "on the BH chip which has PCI tile at 11,  ARC is wired to PCI tile directly outside of NOC" or something like that

tests/baremetal/test_blackhole_arc_apb.cpp
```diff
@@ -0,0 +1,139 @@
...
+// device protocol, one word at a time regardless of the requested size.
+TEST_F(BlackholeArcApbTest, ReadOverJtagGoesThroughDeviceProtocol) {
+    BlackholeArcApb apb(&protocol_, /*pcie_interface=*/nullptr, &jtag_, &arch_);
+
+    EXPECT_CALL(
+        protocol_,
+        read_ctrl(_, ARC_CORE, blackhole::ARC_NOC_XBAR_ADDRESS_START + APB_OFFSET, sizeof(uint32_t), NocId::NOC1));
+
+    uint32_t value = 0;
+    apb.read(&value, APB_OFFSET, sizeof(uint64_t), ARC_CORE, NocId::NOC1);
```
broskoTT: Having both read and write seems redundant to test the logic you want, only one should be enough.

tests/baremetal/test_blackhole_arc_apb.cpp
```diff
@@ -0,0 +1,139 @@
...
+};
+
+// A device opened over JTAG has no PcieInterface, so the access is routed over the NOC through the
+// device protocol, one word at a time regardless of the requested size.
+TEST_F(BlackholeArcApbTest, ReadOverJtagGoesThroughDeviceProtocol) {
+    BlackholeArcApb apb(&protocol_, /*pcie_interface=*/nullptr, &jtag_, &arch_);
+
+    EXPECT_CALL(
+        protocol_,
+        read_ctrl(_, ARC_CORE, blackhole::ARC_NOC_XBAR_ADDRESS_START + APB_OFFSET, sizeof(uint32_t), NocId::NOC1));
```
broskoTT: This is locking in the implementation too much. What you actually want to test is that this went to the range that you'd expect for accessing APB. Can you write something like this:
```
using ::testing::_;
using ::testing::Between;

// Define your expected address boundaries
const uint64_t range_min = blackhole::ARC_NOC_XBAR_ADDRESS_START;
const uint64_t range_max = blackhole::ARC_NOC_XBAR_ADDRESS_START + APB_MAX_OFFSET;

EXPECT_CALL(
    protocol_,
    read_ctrl(
        _,                        // Ignore 1st arg
        _,                        // Ignore 2nd arg (or keep ARC_CORE if required)
        Between(range_min, range_max), // Check 3rd arg address is in range
        _,                        // Ignore 4th arg
        _                         // Ignore 5th arg
    )
);
```

Same for other tests

broskoTT: I see now ReadUsesBarWhenArcAvailableOverAxi below, I think that's enough actually

tests/baremetal/test_blackhole_arc_message_queue.cpp
```diff
@@ -0,0 +1,251 @@
...
+        for (size_t i = 0; i < return_values.size(); i++) {
+            memory_.at(word_address(response_offset + 1 + i)) = return_values[i];
+        }
+    }
+
+    blackhole_implementation arch_;
+    NiceMock<MockDeviceProtocol> protocol_;
+    NiceMock<MockPcieInterface> pcie_;
+    NiceMock<MockJtagInterface> jtag_;
+    FakeArcMemory memory_;
```
broskoTT: Seems like a redundant class, you can just have a map here directly

broskoTT (approved):

tests/baremetal/test_blackhole_arc_apb.cpp
nbuncicTT: read() and write() each have their own copy of the JTAG/NOC/BAR branch, so covering only one would miss a wrong branch or base in the other.

tests/baremetal/test_blackhole_arc_apb.cpp
nbuncicTT: Agreed, that comment was restating the variable names. Rewrote it as the hardware reason.

tests/baremetal/test_blackhole_arc_message_queue.cpp
nbuncicTT: It holds the byte→word address splitting both ON_CALLs need; happy to flatten it if you prefer.

PR #3207, opened 2026-08-20
[Base API] Add DeviceFirmware component and its Blackhole implementation

nbuncicTT:
## Issue

The firmware operations a device performs (startup, ARC command dispatch, chip identity, training, power and clock control) live directly on `TTDevice` and its per-arch subclasses, mixed in with device I/O.

## Description

Introduces `DeviceFirmware`, the per-architecture interface for those operations, and `BlackholeDeviceFirmware` implementing it. Every method takes the NOC to route over, so the component holds no NOC state.

`BlackholeDeviceFirmware` is built from interfaces rather than a `TTDevice`: the device protocol, the optional PCIe and JTAG interfaces, the architecture implementation, `FirmwareInfoProvider` and `FirmwareTelemetryReader`. It drives ARC registers through `BlackholeArcApb` and ARC messages through `BlackholeArcMessageQueue`, serialized by the same cross-process `ARC_MSG` mutex `ArcMessenger` uses. Behavior is ported from `BlackholeTTDevice` and `TTDevice`.

## List of the changes

- `DeviceFirmware` interface, plus `DeviceCommandResult`, `ClockState` and `EthTrainingStatus`.
- `BlackholeDeviceFirmware`: firmware startup and probe, ARC command dispatch under the ARC mutex, chip info (identity from `FirmwareInfoProvider`, all five harvesting masks from telemetry), ethernet and DRAM training with retrains, and clock and power state.
- Both ARC core coordinates resolved once in the constructor, so `get_firmware_noc_coord` is a lookup rather than a NIU read per access.
- Wormhole and simulation implementations are placeholders.

## Testing

Library links with no unresolved symbols and the full baremetal suite passes (187 tests). Nothing constructs these classes yet, so there is no runtime path to exercise on hardware.

## API Changes

- New public headers under `umd/device/tt_device/firmware/`.
- Three behaviors differ from the methods they replace, because the signatures do: `wait_eth_core_training` returns whether training settled rather than an elapsed time that was always zero; `wait_dram_channel_training` returns false where the old code warned and returned early on a missing training status; `set_clock_state` takes `ClockState`, which maps onto the two AICLK messages the old `PowerState` overload sent.

Not integrated into `TTDevice` yet. Known follow-ups: constructor null-validation, `FirmwareStartupError`'s device identity, the Wormhole and simulation implementations, and then integration, after which `TTDevice::set_arc_coordinate`, `get_arc_core` and eventually `ArcMessenger` can go.

Stacked on #3204.

### DIFF
--- a/device/api/umd/device/tt_device/firmware/device_firmware.hpp
+++ b/device/api/umd/device/tt_device/firmware/device_firmware.hpp
@@ -1,0 +1,146 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <chrono>
+#include <cstdint>
+#include <vector>
+
+#include "umd/device/types/cluster_descriptor_types.hpp"
+#include "umd/device/types/eth_training_status.hpp"
+#include "umd/device/types/noc_id.hpp"
+#include "umd/device/types/power_state.hpp"
+#include "umd/device/types/xy_pair.hpp"
+
+namespace tt::umd {
+
+/**
+ * @brief Result of a firmware command execution.
+ *
+ * Bundles the exit code and any return values from the firmware into a single
+ * return type, eliminating the need for out-parameters.
+ */
+struct DeviceCommandResult {
+    uint32_t exit_code;
+    std::vector<uint32_t> return_values;
+};
+
+/**
+ * @brief Requested clock state for a device.
+ */
+enum class ClockState {
+    BUSY,  ///< Maximum AICLK frequency.
+    IDLE,  ///< Minimum AICLK frequency.
+};
+
+/**
+ * @brief Abstract firmware interface for a single device.
+ *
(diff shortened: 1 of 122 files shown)


device/api/umd/device/tt_device/firmware/device_firmware.hpp
```diff
@@ -0,0 +1,146 @@
...
+     * @param noc_id NOC to route through.
+     * @return true if translation is enabled.
+     */
+    virtual bool get_noc_translation_enabled([[maybe_unused]] NocId noc_id = NocId::DEFAULT_NOC) = 0;
+
+    /**
+     * @brief Returns the NOC coordinate of the management firmware core.
+     * @param noc_id NOC to route through.
+     */
+    virtual tt_xy_pair get_firmware_noc_coord([[maybe_unused]] NocId noc_id = NocId::DEFAULT_NOC) const = 0;
```
aleksamarkovicTT: Can you add a method that returns a map of firmware NOC coordinates for all valid NocIds?

aleksamarkovicTT (approved):

PR #3208, opened 2026-08-21
SWEmuleChip: deterministic L1 slot map + shared-segment identity

xanderchin:
Makes a `SWEmuleChip`'s worker-L1 layout and identity stable enough for two PROCESSES to share it.

This is the base of the emule multi-rank work: it is what lets two emulated MPI ranks read and write
each other's chip L1. tt-metal #53673 depends on it — that PR compiles without this, but every
cross-rank delivery drops, because a peer looks for a named segment nothing created.

Non-emule builds are unaffected: construction stays behind `TT_UMD_BUILD_EMULE`, and shared backing
additionally requires `TT_EMULE_CHIP_SHM`.

## Why

Two processes independently compute a core → slot index and then reach into each other's L1 through
it. If they ever disagree, one rank writes into the wrong core's memory and the symptom is corrupted
tensor data somewhere else entirely, with no error at the point of the mistake. Both halves of that
agreement — the layout and the name of the thing holding it — have to be derived, not incidental.

## What

**The slot map is built once, in the constructor, from the SoC descriptor's Tensix list.** Assigning
on first touch could not survive sharing: touch order differs between a process that dispatches
kernels (a grid walk) and one that only host-writes buffers (write order), so the same core would land
on a different slot in each.

**It is keyed under every naming that can reach `get_core`.** `write_to_device` passes a `CoreCoord`'s
raw x/y and drops its `coord_system`, while `get_cores` defaults to `NOC0` — so a `TRANSLATED`-only map
would let a host write miss the pool and take a private mapping for a core a peer is writing through
the shared slot, leaving one physical core with two backings. Aliases are merged only when
collision-free: one naming's coordinate for core A can equal another's for core B, and letting the
later insert win would point a key at the wrong core's L1 — the exact corruption this map exists to
prevent. `TRANSLATED` stays authoritative, so an arch whose namings genuinely conflict falls back to
today's behaviour (private mapping plus the existing warning) rather than to silent corruption.

**A chip's segment identity is its globally stable unique id, never its `ChipId`.**
`TT_VISIBLE_DEVICES` renumbers each process's visible chips to `0..N-1`, so two processes holding
different physical chips would both call theirs chip 0 and collide on one segment. The harvesting masks
join the key because they change the core list and therefore the layout — packed into disjoint 20-bit
fields rather than XOR-folded, because folding overlapping shifts is not injective and two chips with
genuinely different core lists could otherwise collide, attach to each other's segment, and resolve
through the wrong slot map. A mask too wide for its field throws instead of truncating into an alias.

**Identity is `std::optional`,** so "no stable id" is distinguishable from an id that is legitimately
`0` — the legacy descriptor computes `chip << 32`, which is exactly `0` for chip 0. With sharing
requested and no id available, construction fails rather than quietly handing back process-private L1
that no peer can attach; the chip reports the same case itself, because otherwise the only visible
symptom is dropped cross-rank writes somewhere unrelated.

**A slot-map miss is fatal under sharing.** A Tensix core reaching `get_core` without a slot means a
naming was unavailable or was dropped as ambiguous; mapping it privately would leave a peer addressing
that core through the shared pool and seeing nothing. Non-Tensix coordinates (`write_to_device` routes
ETH/PCIE/DISPATCH/ARC through the same call, since it only special-cases DRAM) legitimately have no
slot and are still mapped privately, so the two cases are told apart by a Tensix classification set
built across every naming independently of slot assignment.

**No tt-emule header leaves the emule translation unit.** `device/cluster.cpp` is compiled in every UMD
build while tt-emule is optional, so it passes the identity through and `SWEmuleChip` — compiled only
under `TT_UMD_BUILD_EMULE` — decides what a missing identity means.

**New members are appended, never interleaved.** Keeping the one-argument constructor preserves that
mangled symbol, but moving an existing member's offset would break a pre-existing accessor that an
already-built emule binary had inlined: the old `l1_size()` offset would land on `dram_bank_size_`.

**A core's L1 must fit its slot.** Quasar's 4 MiB `worker_l1_size` exceeds the 2 MB slot stride and
would have overrun into the next core's slot, so that now throws at construction.

**The one-argument constructor is kept as a real overload,** not a defaulted parameter: a default
argument is applied at the call site, so collapsing the two would drop the old mangled symbol and break
any emule binary already linked against it.

## Behaviour change

Construction now **hard-fails** when a descriptor's `worker_l1_size` exceeds the `L1Pool` slot
stride (e.g. `tests/soc_descs/quasar_simulation_1x1.yaml`, 4 MiB vs 2 MB). Base silently overran
into the next core's slot. Throwing is right, but it is a change and is called out here.

## Verification

Built and run with `-DTT_UMD_BUILD_EMULE=ON -DTT_EMULE_PATH=<local tt-emule checkout>`, since the
pinned revision predates the tt-emule API this uses (see merge order below).

| Check | Result |
|---|---|
| umd build, `TT_UMD_BUILD_EMULE=ON` | clean |
| `baremetal_tests --gtest_filter=ApiEmuleClusterTest.*` | 7 passed (2 pre-existing + 5 new) |
| tt-metal `build_emule` / `build_emule_dist` | both link clean |
| Single-rank blaze gate, 35 files | 35 passed, 0 failed, 0 skipped |
| 2-rank chained d2d wave (MPI) | 2 passed |
| 4-rank x 4x2 galaxy chained wave | passed on all 4 ranks |
| Shared-pool miss warnings / absorbed fabric sends | 0 / 0 |

The 4-rank run is the load-bearing one: a page crosses 3 d2d hops over 4 meshes, with two interior
ranks holding no host sockets, and the tail asserts the decoded page is bitwise identical to the
head. Zero pool-miss warnings means owner and peer derived the same layout from the shared builder.

The new `NonTranslatedCoreNamingStillResolvesToAPoolSlot` test was confirmed to have teeth: with
only the entry-point translation reverted it fails on every core, at `NOC0 (1,1)` /
`TRANSLATED (18,18)`.

## Merge order

`chip_store.hpp` and the 3-arg `L1Pool` this uses are authored in the tt-emule PR, so the pin here
can only move once that lands:

tt-emule -> **pin bump here** -> this PR -> tt-metal #53673 -> tt-blaze

Note for a CI lane: tt-metal's umd pin `33a7e52e` has diverged from umd main (not an ancestor) and
tt-metal does not build against main at `tt_cluster.cpp:778`, independent of this PR.

## Review

Second round (@pjanevskiTT, 14 comments) — all addressed; three were substantive:

- **Coordinates were never normalized on entry.** Every other `Chip` translates; this one looked a
  raw `{x,y}` up in whichever naming the caller picked, so a NOC0-named worker got a private mmap
  invisible to peers. Fixed at both entry points, with a test that catches it.
- **A synthesized unique id could name a shared segment.** Where the descriptor carried no
  `chip_unique_ids` block it invents them (`chip << 32`, or the bare logical id for mock), so two
  ranks each holding a different physical chip 0 computed one key. `ClusterDescriptor` now reports
  provenance and `Cluster` withholds an invented id. Our own multi-rank runs could not have caught
  this — every descriptor we use carries real ids.
- **The slot map had two implementations.** tt-metal rebuilt it independently for peer chips, so one
  cross-process invariant lived in two places. `build_worker_slot_map()` is now the single source.

Deferred with an issue: #3272 (key the segment on `tensix_harvesting_mask` alone — changes the
segment name, so three repos must move together).

## Earlier review rounds

An adversarial review of this commit raised four issues, all addressed here: the coordinate-naming
mismatch, `chip_uid == 0` being conflated with "no identity", an owner/peer disagreement on acceptable
segment size (the owner accepted `>=` where peers require `==`, so a stale larger segment let the owner
run while every peer declined — fixed on the tt-emule side), and the constructor ABI. It confirmed
clean that `TT_VISIBLE_DEVICES` remapping preserves unique-id values while remapping only keys, that
the mask packing is injective, and that Tensix enumeration is deterministic on both the generic and
Blackhole paths.


device/api/umd/device/chip/sw_emule_chip.hpp
```diff
@@ -32,7 +33,16 @@ class SimulationSysmemManager;
 /// All non-memory operations (barriers, resets, power management) are no-ops.
 class SWEmuleChip : public Chip {
 public:
+    // chip_uid is the chip's GLOBALLY stable unique id, used to name the shared L1 segment under
+    // TT_EMULE_CHIP_SHM. It must not be the ChipId: TT_VISIBLE_DEVICES makes the cluster descriptor
+    // renumber each process's visible chips to 0..N-1, so two processes holding different physical
```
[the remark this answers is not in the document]
xanderchin: Fixed — the sentence was truncated when an adjacent line was removed. It now completes: "…so two processes holding different physical chips would both call theirs chip 0 and collide on one segment."

device/api/umd/device/chip/sw_emule_chip.hpp
```diff
@@ -116,12 +138,20 @@ class SWEmuleChip : public Chip {
...
     // Host-facing (PCIe) address resolution — an existing, already-upstream UMD class (also used
     // by TTSimTTDevice), not emule-specific. No host-memory channels to pre-reserve (SWEmuleChip
     // has no hugepage concept), so 0.
     std::unique_ptr<SimulationSysmemManager> sysmem_manager_;
+
+    // Appended, never interleaved: an existing member's offset must not move, or an accessor that a
+    // previously built emule binary inlined (l1_size / dram_bank_size) would read the wrong field.
+    std::unordered_map<tt_xy_pair, size_t> slot_of_;
+    uint64_t shm_harvest_mask_ = 0;
+    std::unordered_map<tt_xy_pair, size_t> tensix_coords_;
```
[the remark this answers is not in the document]
xanderchin: You are right, and the comment was wrong. I had only preserved the one-argument *symbol*; appending members still changes `sizeof(SWEmuleChip)`, so a client that allocated the old size and then runs the new constructor is writing past its allocation — worse than the shifted-offset problem I thought I had fixed.

Rather than add indirection to an opt-in emule class, the comment now states it plainly: this is an ABI break, the one-argument constructor is for **source** compatibility only, and anything linking the old layout must be rebuilt. emule is gated behind `TT_UMD_BUILD_EMULE` and carries no ABI promise.

device/chip/sw_emule_chip.cpp
```diff
@@ -42,14 +44,113 @@ SWEmuleChip::SWEmuleChip(const SocDescriptor& soc_descriptor) :
...
+        bool consistent = true;
+        for (size_t i = 0; i < named.size() && consistent; ++i) {
+            auto it = slot_of_.find(tt_xy_pair(named[i].x, named[i].y));
+            consistent = (it == slot_of_.end()) || (it->second == i);
+        }
+        if (!consistent) {
+            continue;  // ambiguous: this naming would repoint an already-mapped coordinate
+        }
+        for (size_t i = 0; i < named.size(); ++i) {
+            slot_of_.emplace(tt_xy_pair(named[i].x, named[i].y), i);
```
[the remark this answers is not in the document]
xanderchin: Agreed, and this one changed the design: the aliasing is gone.

Skipping an inconsistent naming was never sufficient, exactly as you describe — the canonical entry stays in `slot_of_` and `get_core` sees only `{x,y}`. The map is now keyed on one naming and never aliased. That makes a wrong answer impossible; the cost is that an access expressed in another naming misses the pool, which is now reported, with the shared-pool case named specifically because a private mapping is then invisible to peers.

The real fix for the underlying gap is to stop discarding the `CoreCoord` in `write_to_device`/`read_from_device` and key on system+type. That is a UMD API change and does not belong in this commit.

device/chip/sw_emule_chip.cpp
```diff
@@ -42,14 +44,113 @@ SWEmuleChip::SWEmuleChip(const SocDescriptor& soc_descriptor) :
...
+            continue;
+        }
+    }
+    // Alias the other namings onto the same slots, because write_to_device passes a CoreCoord's raw
+    // x/y and drops its coord_system while get_cores defaults to NOC0 -- a TRANSLATED-only map lets a
+    // host write miss the pool and take a private mapping for a core a peer reaches through the
+    // shared slot. Merge a naming ONLY if it is collision-free against what is already mapped: one
+    // naming's coordinate for core A can equal another's for core B, and letting the later write win
+    // would point a key at the wrong core's L1, which is the exact corruption this map exists to stop.
+    for (auto system : {CoordSystem::NOC0, CoordSystem::NOC1, CoordSystem::LOGICAL}) {
```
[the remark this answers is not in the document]
xanderchin: Same conclusion — `LOGICAL` is no longer added. Logical TENSIX `(0,0)` colliding with PCIE/ARC `(0,0)` while `write_to_device` drops `core_type` is unanswerable from a raw `{x,y}` key, so the map keys on one naming only.

Worth noting the two runs that exercise this pass with no core falling back to a private mapping, so the emulator does address every pooled worker in the canonical naming — the aliasing was solving a problem that does not arise here, at the cost of one that would.

device/chip/sw_emule_chip.cpp
```diff
@@ -42,14 +44,113 @@ SWEmuleChip::SWEmuleChip(const SocDescriptor& soc_descriptor) :
...
+            slot_of_.emplace(tt_xy_pair(named[i].x, named[i].y), i);
+        }
+    }
+    size_t pool_size = (!tensix.empty() ? tensix.size() : 128);  // 128 = WH/BH fallback if SOC reports 0
+    // A slot must hold a whole core's L1. Quasar's worker_l1_size (4 MiB) exceeds L1Pool's 2 MB
+    // slot stride, which would silently overrun into the next core's slot.
+    if (static_cast<size_t>(l1_size_) > tt_emule::L1Pool::SLOT_SIZE) {
+        throw std::runtime_error(
+            "SWEmuleChip: worker_l1_size " + std::to_string(l1_size_) + " exceeds the L1Pool slot stride " +
+            std::to_string(tt_emule::L1Pool::SLOT_SIZE) + " — adjacent cores' L1 would overlap");
```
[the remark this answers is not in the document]
xanderchin: Fixed — all five sites now use `UMD_THROW(error::RuntimeError, ...)` with `utils/error_detail.hpp` included.

pjanevskiTT (approved):

device/api/umd/device/chip/sw_emule_chip.hpp
```diff
@@ -32,7 +33,17 @@ class SimulationSysmemManager;
 /// All non-memory operations (barriers, resets, power management) are no-ops.
 class SWEmuleChip : public Chip {
 public:
+    // chip_uid is the chip's GLOBALLY stable unique id, used to name the shared L1 segment under
+    // TT_EMULE_CHIP_SHM. It must not be the ChipId: TT_VISIBLE_DEVICES makes the cluster descriptor
```
pjanevskiTT, line 45: we made chip IDs in cluster descriptor stable. Why are we not using those IDs here?

device/chip/sw_emule_chip.cpp
```diff
@@ -42,14 +45,77 @@ SWEmuleChip::SWEmuleChip(const SocDescriptor& soc_descriptor) :
...
+        if (tensix_mask > kMax || dram_mask > kMax || eth_mask > kMax) {
+            UMD_THROW(
+                error::RuntimeError,
+                "SWEmuleChip: harvesting mask exceeds the " + std::to_string(kField) +
+                    "-bit field width of the shared-segment key; widen the key before sharing this chip");
+        }
+        shm_harvest_mask_ = tensix_mask | (dram_mask << kField) | (eth_mask << (2 * kField));
+    }
+    if (chip_uid.has_value() && tt_emule::chip_store_shared()) {
+        worker_pool_ = std::make_unique<tt_emule::L1Pool>(pool_size, *chip_uid, shm_harvest_mask_);
```
pjanevskiTT, line 105: This does not compile against the tt-emule revision this PR pins. `third_party/CMakeLists.txt:245` still reads `GIT_TAG acf8473de76a3138086acfd3e48aa2748a025126`, unchanged from base, and at that commit `include/tt_emule/l1_pool.hpp` declares exactly one constructor — `explicit L1Pool(size_t num_slots)` — and no `is_shared()`. `chip_store_shared` does not appear anywhere in that tree.

Three of the four new tt-emule uses are missing at the pin:
- `L1Pool(pool_size, *chip_uid, shm_harvest_mask_)` — this line
- `tt_emule::chip_store_shared()` — lines 94 and 100
- `worker_pool_->is_shared()` — line 163

`L1Pool::SLOT_SIZE` is the one that does exist (`l1_pool.hpp:17`).

This will not turn CI red. `grep -ril emule .github/` returns nothing across all 22 workflows, `TT_UMD_BUILD_EMULE` defaults `OFF` (`CMakeLists.txt:62`), so `chip/sw_emule_chip.cpp` is excluded from the target (`device/CMakeLists.txt:69`) and this whole block is preprocessed away — the PR goes green while not compiling.

Please bump the pin in this PR to a revision that provides all three. A configure+build-only lane with `-DTT_UMD_BUILD_EMULE=ON` would have caught it; worth adding given the Verification section is currently unfalsifiable from the repo.

device/chip/sw_emule_chip.cpp
```diff
@@ -42,14 +45,77 @@ SWEmuleChip::SWEmuleChip(const SocDescriptor& soc_descriptor) :
...
-    // costs mesh capacity; get_core() falls back to an individual mmap if exceeded.
-    size_t num_tensix = soc.get_cores(tt::CoreType::TENSIX).size();
-    size_t pool_size = (num_tensix > 0 ? num_tensix : 128);  // 128 = WH/BH fallback if SOC reports 0
-    worker_pool_ = std::make_unique<tt_emule::L1Pool>(pool_size);
+    // only WORKER cores use the pool), so num_tensix is an exact bound — no padding.
+    //
+    // TRANSLATED, because that is the naming get_core() is keyed on. The descriptor's core order is
+    // a deterministic row-major walk, so this map is identical in every process given the same
+    // descriptor + harvesting mask — which is what makes a slot index a portable identity.
+    const auto tensix = soc.get_cores(tt::CoreType::TENSIX, CoordSystem::TRANSLATED);
```
pjanevskiTT: The comment above says TRANSLATED "is the naming `get_core()` is keyed on", but nothing on the way in translates. `write_to_device` and `read_from_device` (lines 189 and 197) call `get_core(tt_xy_pair(core.x, core.y))` with the caller's raw coordinate in whatever `CoordSystem` they supplied. Every other `Chip` normalizes first — `LocalChip` does `get_soc_descriptor().translate_chip_coord_to_translated(core, get_selected_noc_id())` at `local_chip.cpp:248/275/323/346` — and `Cluster::write_to_device` forwards the `CoreCoord` untranslated, so translation is the `Chip`'s job.

On the one path in this repo that exercises the class: `tests/baremetal/test_sw_emule_chip.cpp:43,59` use `get_cores(CoreType::TENSIX)`, which defaults to `CoordSystem::NOC0` (`soc_descriptor.hpp:76`). WH NOC0 Tensix are x∈{1-4,6-9}, y∈{1-5,7-11}; TRANSLATED Tensix start at (18,18). Disjoint sets — so every lookup misses `slot_of_`, every worker takes the `else` at line 158 and gets a private mmap plus a warning, and the pool is allocated and never indexed once.

That is also a regression against base: `next_slot_` handed pool slots to whatever raw (x,y) arrived, so a NOC0 caller did get the MAP_32BIT 2 MB-aligned slot the class doc promises (`sw_emule_chip.hpp:30-31`). I take the point from your earlier reply that your emule runs show no fallbacks, so tt-metal is passing TRANSLATED — but the contract shouldn't depend on which naming the caller happens to pick, and the in-repo test currently proves nothing about slot determinism.

Translating at the top of `write_to_device`/`read_from_device` as `LocalChip` does would make the lines 50-52 comment true and make the lines 54-60 aliasing argument unnecessary.

device/cluster.cpp
```diff
@@ -245,7 +245,20 @@ std::unique_ptr<Chip> Cluster::construct_chip_from_cluster(
...
 #ifdef TT_UMD_BUILD_EMULE
-        return std::make_unique<SWEmuleChip>(soc_desc);
+        // Pass the chip's globally stable unique id (NOT chip_id, which cluster_desc renumbers to
+        // 0..N-1 per visible-device subset) so a shared L1 segment can be keyed on it.
+        std::optional<uint64_t> chip_uid;
+        if (cluster_desc != nullptr) {
+            const auto& uids = cluster_desc->get_chip_unique_ids();
+            auto it = uids.find(chip_id);
+            if (it != uids.end()) {
+                chip_uid = it->second;  // 0 is a legitimate id: the legacy descriptor computes chip << 32
```
pjanevskiTT: `sw_emule_chip.hpp:36-39` says the id "must not be the ChipId" because two processes would otherwise both call theirs chip 0 and collide on one segment. `get_chip_unique_ids()` does not deliver that on either path reachable here:

- **Legacy/mock YAML:** `cluster_descriptor.cpp:815-819` synthesizes `chip_unique_ids.insert({chip, static_cast<uint64_t>(chip) << 32})` when the YAML has no `chip_unique_ids` block. `tests/cluster_descriptor_examples/wormhole_N150.yaml` has no such block, so its chip gets uid `0`.
- **Mock cluster:** `cluster_descriptor.cpp:577` inserts `{logical_id, logical_id}`, then `fill_mock_hardcoded_data` at `:594` tries `insert({logical_id, kSimUniqueIdBase + logical_id})` — a no-op on an existing key, so the uid stays the logical id. Chip 0 → `0`.

Two ranks each emulating a *different* physical chip, each with its own single-chip descriptor, then compute the same `(uid=0, mask)` key and attach to each other's worker L1. That is the collision the header comment claims to design out, reached by a different route.

Either gate the shared path on an id that is genuinely board-derived (`get_board_id_for_chip` + `get_asic_location`, or a flag for "the descriptor really carried `chip_unique_ids`" so `nullopt` is passed otherwise), or have the segment carry an identity header that is validated on attach.

Separately, `if (cluster_desc != nullptr)` on line 251 is dead: the only caller (line 566) already dereferenced `cluster_desc` at lines 541 and 551. Worth dropping, since as written it reads as covering the case above.

device/chip/sw_emule_chip.cpp
```diff
@@ -81,23 +147,33 @@ tt_emule::Core* SWEmuleChip::get_core(tt_xy_pair core_xy) {
...
-    if (next_slot_ < worker_pool_->num_slots()) {
-        // Worker cores: allocate from L1Pool (aligned slots, bitmask offset extraction).
-        size_t slot = next_slot_++;
-        uint8_t* slot_mem = worker_pool_->slot_ptr(slot);
-        core = std::make_unique<tt_emule::Core>(coord, slot_mem, static_cast<size_t>(l1_size_));
-        core_to_slot_[core_xy] = slot;
+    auto slot = slot_of_.find(core_xy);
+    if (slot != slot_of_.end()) {
+        core = std::make_unique<tt_emule::Core>(
+            coord, worker_pool_->slot_ptr(slot->second), static_cast<size_t>(l1_size_));
```
pjanevskiTT, line 175: The removed code guarded every pool access with `if (next_slot_ < worker_pool_->num_slots())`. That guard is gone, and the index is now only nominally bounded by `pool_size == tensix.size()` from line 64.

That holds for a locally-created pool, but not necessarily for the shared branch. If `L1Pool(pool_size, uid, mask)` *attaches* to a segment a peer already created, `num_slots()` reflects the existing segment rather than the local request. A peer that computed a different `pool_size` under the same key — a different `sdesc_path`, or the `tensix.empty()` → 128 fallback against a real 72 — gives `slot->second >= num_slots()`. At the pinned revision `slot_ptr` returns `nullptr` in that case (`l1_pool.hpp:52-53`), so the `Core` gets a null L1 base and the `memcpy` in `write_to_device` segfaults on first touch instead of reporting anything.

I can't see the new 3-arg ctor, so if it already validates the attached segment size against `pool_size`, say so and this is moot. Otherwise: assert `slot->second < worker_pool_->num_slots()` before `slot_ptr`, and add `pool_size` to the segment key so a mismatched layout can't silently attach — the key is currently only `(chip_uid, shm_harvest_mask_)`.

device/chip/sw_emule_chip.cpp
```diff
@@ -42,14 +45,77 @@ SWEmuleChip::SWEmuleChip(const SocDescriptor& soc_descriptor) :
...
+    // write_to_device, which passes a CoreCoord's raw x/y and drops both its coord_system and its
+    // core_type -- so a raw {x,y} key cannot say which space it came from. Aliasing would then be
+    // unsound in two ways on Blackhole: logical Tensix (1,2) and translated (1,2) can be different
+    // workers, and logical TENSIX (0,0) collides with logical PCIE/ARC (0,0), which would resolve a
+    // non-Tensix access into a pooled slot. Keying on one naming keeps a wrong answer impossible; the
+    // cost is that an access expressed in another naming misses the pool and is reported below.
+    for (size_t i = 0; i < tensix.size(); ++i) {
+        slot_of_[tt_xy_pair(tensix[i].x, tensix[i].y)] = i;
+    }
+    size_t pool_size = (!tensix.empty() ? tensix.size() : 128);  // 128 = WH/BH fallback if SOC reports 0
```
pjanevskiTT: In base this fallback was live: `next_slot_++` handed out slots on first touch, so 128 slots were usable even with an empty core list. Now if `tensix` is empty then `slot_of_` is empty too, line 157 is unreachable, and no core can ever index the pool — yet `L1Pool` still mmaps and `memset`s 128 × 2 MB = 256 MB in the scarce low-4 GB MAP_32BIT window, committing it as RSS for nothing. The comment this PR deleted from base made exactly that argument ("the pool sits in the scarce low-4 GB window … so over-sizing costs mesh capacity").

Under sharing it also publishes a 128-slot segment under the same key a 72-slot peer would compute — see the note on line 157.

`size_t pool_size = tensix.size();` covers it: `L1Pool(0)` early-returns without mmapping at the pinned revision, and `num_pool_slots()` already tolerates a null `worker_pool_`.

device/chip/sw_emule_chip.cpp
```diff
@@ -17,13 +17,16 @@
 #include "tt_emule/l1_pool.hpp"
 #include "umd/device/chip_helpers/simulation_sysmem_manager.hpp"
 #include "umd/device/soc_descriptor.hpp"
+#include "umd/device/utils/error_detail.hpp"
```
pjanevskiTT: `error_detail.hpp` is a private header that `#error`s if included directly:

```
#ifndef UMD_ERROR_HPP_INTERNAL_INCLUDE
#error "error_detail.hpp is a private header. Include umd/device/utils/error.hpp instead."
#endif
```

This compiles only because line 19 sorts before line 20 and transitively pulls `error.hpp` (via `coordinate_manager.hpp` → `cluster_descriptor_types.hpp`), making line 20 a `#pragma once` no-op. If that chain changes, or clang-format reorders the block, the file breaks with a confusing `#error` in a lane nobody runs.

Also `error::RuntimeError` is declared in `error.hpp:34`, not in the detail header — so as written this include doesn't actually provide the symbol the new throws use. This is the only direct includer of `error_detail.hpp` in the repo; every other `UMD_THROW` site includes `umd/device/utils/error.hpp`. Should be that.

While here: `<stdexcept>` on line 11 is now unused — both `std::runtime_error` throws became `UMD_THROW`, and there's no other use in the file.

device/chip/sw_emule_chip.cpp
```diff
@@ -42,14 +45,77 @@ SWEmuleChip::SWEmuleChip(const SocDescriptor& soc_descriptor) :
...
+    // workers, and logical TENSIX (0,0) collides with logical PCIE/ARC (0,0), which would resolve a
+    // non-Tensix access into a pooled slot. Keying on one naming keeps a wrong answer impossible; the
+    // cost is that an access expressed in another naming misses the pool and is reported below.
+    for (size_t i = 0; i < tensix.size(); ++i) {
+        slot_of_[tt_xy_pair(tensix[i].x, tensix[i].y)] = i;
+    }
+    size_t pool_size = (!tensix.empty() ? tensix.size() : 128);  // 128 = WH/BH fallback if SOC reports 0
+    // A slot must hold a whole core's L1. Quasar's worker_l1_size (4 MiB) exceeds L1Pool's 2 MB
+    // slot stride, which would silently overrun into the next core's slot.
+    if (static_cast<size_t>(l1_size_) > tt_emule::L1Pool::SLOT_SIZE) {
```
pjanevskiTT, line 67: This PR adds four new behaviours and touches no test file, though `tests/baremetal/test_sw_emule_chip.cpp` exists. All four are testable from a `SocDescriptor` alone — no shared memory, no second process:

1. **Slot determinism**, the premise of the PR: two `SWEmuleChip` built from the same descriptor agree on `slot_of()` for every Tensix core, and the slot set is exactly `{0 … num_pool_slots()-1}` with no gaps or duplicates. Asserting `slot_of()` for a NOC0-named core would also have surfaced the issue noted on line 53.
2. `num_pool_slots() == soc.get_cores(TENSIX, TRANSLATED).size()`, which pins the line 64 fallback.
3. **This throw:** `tests/soc_descs/quasar_simulation_1x1.yaml` already has `worker_l1_size: 4194304` > 2 MiB, so it's an `EXPECT_THROW(..., error::RuntimeError)` with no new fixture data. Worth calling out in the PR body too — a SWEMULE chip with such a descriptor now hard-fails at construction where base silently overran into the next slot. Throwing is right; it just isn't documented as a behaviour change.
4. **The 20-bit packing** at line 92: that two descriptors differing only in `dram_harvesting_mask` produce different keys, plus `EXPECT_THROW` for a mask `> 0xFFFFF`. Injectivity of that packing is the load-bearing anti-collision argument in the comment above and nothing checks it.

Also: `wormhole_N150.yaml` has no `chip_unique_ids` while `wormhole_N150_unique_ids.yaml` does, so the existing test already walks both `chip_uid` paths — without asserting anything about either.

device/chip/sw_emule_chip.cpp
```diff
@@ -42,14 +45,77 @@ SWEmuleChip::SWEmuleChip(const SocDescriptor& soc_descriptor) :
...
+    // core_type -- so a raw {x,y} key cannot say which space it came from. Aliasing would then be
+    // unsound in two ways on Blackhole: logical Tensix (1,2) and translated (1,2) can be different
+    // workers, and logical TENSIX (0,0) collides with logical PCIE/ARC (0,0), which would resolve a
+    // non-Tensix access into a pooled slot. Keying on one naming keeps a wrong answer impossible; the
+    // cost is that an access expressed in another naming misses the pool and is reported below.
+    for (size_t i = 0; i < tensix.size(); ++i) {
+        slot_of_[tt_xy_pair(tensix[i].x, tensix[i].y)] = i;
+    }
+    size_t pool_size = (!tensix.empty() ? tensix.size() : 128);  // 128 = WH/BH fallback if SOC reports 0
+    // A slot must hold a whole core's L1. Quasar's worker_l1_size (4 MiB) exceeds L1Pool's 2 MB
```
pjanevskiTT: The rationale doesn't hold for the arch it names. `tt::ARCH::QUASAR` uses `grendel::TENSIX_L1_SIZE` = 1572864 (1.5 MiB, `grendel_implementation.hpp:268`), and `tests/soc_descs/quasar_32_arch.yaml` is 1499136 — neither trips this guard. The only 4 MiB `worker_l1_size` in the repo is the `tests/soc_descs/quasar_simulation_*.yaml` set. As written, a reader concludes SWEmuleChip cannot run Quasar at all, and the next maintainer may adjust the threshold to match a condition that doesn't exist.

The guard itself is worth keeping — suggest rewording to "a SOC descriptor whose `worker_l1_size` exceeds the slot stride (e.g. `tests/soc_descs/quasar_simulation_1x1.yaml`, 4 MiB)".

device/chip/sw_emule_chip.cpp
```diff
@@ -42,14 +45,77 @@ SWEmuleChip::SWEmuleChip(const SocDescriptor& soc_descriptor) :
...
+    // slot stride, which would silently overrun into the next core's slot.
+    if (static_cast<size_t>(l1_size_) > tt_emule::L1Pool::SLOT_SIZE) {
+        UMD_THROW(
+            error::RuntimeError,
+            "SWEmuleChip: worker_l1_size " + std::to_string(l1_size_) + " exceeds the L1Pool slot stride " +
+                std::to_string(tt_emule::L1Pool::SLOT_SIZE) + " — adjacent cores' L1 would overlap");
+    }
+    // Shared backing needs a stable identity; without a uid we can only be process-private.
+    // The harvesting mask joins the key because it changes the core list, hence the slot layout.
```
pjanevskiTT: `HarvestingMasks` has five fields; this key takes tensix/dram/eth and silently drops `pcie_harvesting_mask` and `l2cpu_harvesting_mask`. And the stated rationale — "it changes the core list, hence the slot layout" — holds only for `tensix_harvesting_mask`: `get_tensix_cores()` and both `fill_tensix_noc0_translated_mapping()` implementations read no other mask.

So the key over-discriminates on dram/eth (two chips with identical worker layouts refuse to share) while omitting two masks with no stated reason. Keying on `tensix_harvesting_mask` alone would match the reasoning and make the field-width question trivially safe.

On reachability of the throw below: the WH tensix mask spans 10 rows, BH 14 columns, Quasar 4 rows, and the dram/eth masks are narrower still — `> kMax` cannot fire for any descriptor in the repo. Fine as documentation, just not a guard that will run.

Minor, on the constants: `device/` is overwhelmingly SCREAMING_CASE for `constexpr`, and these names mislead — `kField` is a width in bits, `kMax` is a mask. `HARVEST_FIELD_BITS` / `HARVEST_FIELD_MASK` would read better.

device/chip/sw_emule_chip.cpp
```diff
@@ -81,23 +147,33 @@ tt_emule::Core* SWEmuleChip::get_core(tt_xy_pair core_xy) {
...
-            LogUMD,
-            "L1Pool exhausted (all {} slots used); core ({},{}) falling back to individual mmap",
-            worker_pool_->num_slots(),
-            core_xy.x,
-            core_xy.y);
+        // No slot: either a genuinely non-pooled core (write_to_device routes ETH/PCIE/DISPATCH/ARC
+        // here too, since it only special-cases DRAM) or a worker addressed in a naming this map is
+        // not keyed on. A raw (x,y) cannot tell those apart, so this stays a private mapping rather
+        // than a hard error -- but under a shared pool it is invisible to peers, so say which.
+        if (worker_pool_ && worker_pool_->is_shared()) {
```
pjanevskiTT, line 180: In base this warning fired only on genuine pool exhaustion, which was rare. It now fires for every core with no slot, and the comment above concedes that includes expected traffic — ETH/PCIE/DISPATCH/ARC all route through here since only DRAM is special-cased. Combined with the line 53 issue that's one warning per Tensix core per chip (72 lines on a WH N150 in the existing test), scaling with mesh size, which trains people to ignore the log.

Suggest `log_debug` for the non-shared branch and warn-once (or a summary at close) for the shared branch, where it's genuinely actionable. Neither message names the chip or arch, so on a multi-chip cluster the lines aren't attributable.

device/api/umd/device/chip/sw_emule_chip.hpp
```diff
@@ -94,16 +105,28 @@ class SWEmuleChip : public Chip {
...
 
-    // L1Pool for worker cores — single contiguous MAP_32BIT mmap with
-    // 2 MB aligned slots for bitmask offset extraction.
+    // L1Pool for worker cores — one contiguous mmap carved into fixed-size slots.
     std::unique_ptr<tt_emule::L1Pool> worker_pool_;
-    size_t next_slot_ = 0;
 
-    // Slot index tracking: physical core → pool slot.
-    std::unordered_map<tt_xy_pair, size_t> core_to_slot_;
+    // Tensix core → pool slot, built ONCE from the SoC descriptor rather than on first touch.
```
pjanevskiTT, line 127: Moving `slot_of_` and `shm_harvest_mask_` to lines 155-156 left their documentation behind:

- this block ("Tensix core → pool slot, built ONCE…") is followed by a blank line and then `cores_` at line 132, so it reads as documenting `cores_`;
- line 142 ("Folded once in the ctor so the segment key and the peer-side lookup cannot drift apart.") dangles above `SocDescriptor soc_descriptor_;`;
- the ABI note at 151-154 says "The one-argument constructor **below**" — it's at line 43, above — and a class-wide ABI statement wedged mid-member-list reads as documentation for `slot_of_`/`shm_harvest_mask_`.

Moving the two members up next to their own comments fixes the first two and gives the ABI note a reason to live in the class doc block at 28-33. That block, separately, still describes the pool as "MAP_32BIT mmap with 2 MB aligned slots" while line 123 was deliberately generalised to "one contiguous mmap carved into fixed-size slots" — worth making them agree.

On the `sizeof` claim at line 152: removed `size_t` (8) + `unordered_map` (56) = 64 bytes, added `unordered_map` (56) + `uint64_t` (8) = 64 bytes, so on libstdc++ x86-64 `sizeof(SWEmuleChip)` most likely does not change. Member *offsets* do, which is enough for the note's actual point — worth trimming it to that.

device/api/umd/device/chip/sw_emule_chip.hpp
```diff
@@ -32,7 +33,17 @@ class SimulationSysmemManager;
 /// All non-memory operations (barriers, resets, power management) are no-ops.
 class SWEmuleChip : public Chip {
 public:
+    // chip_uid is the chip's GLOBALLY stable unique id, used to name the shared L1 segment under
+    // TT_EMULE_CHIP_SHM. It must not be the ChipId: TT_VISIBLE_DEVICES makes the cluster descriptor
+    // renumber each process's visible chips to 0..N-1, so two processes holding different physical
+    // chips would both call theirs chip 0 and collide on one segment.
+    //
+    // Two overloads rather than a defaulted parameter, so the one-argument signature keeps compiling
```
pjanevskiTT: "a default argument is applied at the call site and would remove it" — a defaulted parameter doesn't break source compatibility; every one-argument call still compiles. The only thing a separate overload preserves is the mangled symbol, i.e. ABI. And lines 151-154 concede the class is already an ABI break because member offsets changed, so a client that can't be rebuilt is broken by the layout change regardless. The extra overload buys nothing, and the two comments contradict each other.

`SWEmuleChip(const SocDescriptor&, std::optional<uint64_t> chip_uid = std::nullopt);` and drop this paragraph. If you'd rather keep the overloads, the new one on line 46 should be `explicit` for consistency with line 43.

device/api/umd/device/chip/sw_emule_chip.hpp
```diff
@@ -94,16 +105,28 @@ class SWEmuleChip : public Chip {
     // coord of a channel resolves here, so a noc=1 read sees a noc=0 / host write.
     tt_emule::Core* get_dram_channel_backing(uint32_t channel);
 
+    // Pool slot for a worker core, or SIZE_MAX if it has none. The map is derived from the SoC
+    // descriptor alone, so a chip owned by a PEER PROCESS with the same arch and harvesting has an
+    // identical layout — which is what lets a rank resolve a NOC address into a chip it does not own,
+    // from the peer's shared segment. See tt-emule docs/fabric-ccl-emulation.md.
+    size_t slot_of(tt_xy_pair core_xy) const;
```
pjanevskiTT: These three getters have no caller in the repo at head — `git grep` finds only the declarations and their definitions. The header is installed unconditionally (`device/CMakeLists.txt:200`), so this is public surface added for an out-of-repo consumer.

Two specifics:
- `slot_of` takes a bare `tt_xy_pair`, which re-exports into the public API the exact coord-system ambiguity that lines 54-60 of the .cpp identify as unsound, and it signals absence with `SIZE_MAX` where this repo (and this PR) otherwise uses `std::optional`. `std::optional<size_t> get_pool_slot(CoreCoord) const` would match both the convention and the neighbouring `get_core` / `get_dram_channel_backing` naming.
- `shm_harvest_mask()` exports the internal 3×20-bit packing as a public `uint64_t`, so any external reader is coupled to a layout the .cpp comment says may need widening. That's the weakest of the three.

Also, the `See tt-emule docs/fabric-ccl-emulation.md` pointer on line 111 sends readers of an always-installed public header to a doc in an optional, separately-hosted repo — and that file does not exist at the pinned revision (`docs/` there contains only `historical_reports`).

Making the tests from the `sw_emule_chip.cpp:67` note the in-repo callers would settle the YAGNI question for the first two.

device/chip/sw_emule_chip.cpp
xanderchin: This circular dep unfortunately. Need this to land the emule changes.

device/chip/sw_emule_chip.cpp
xanderchin: Fixed. `write_to_device`/`read_from_device` now call `translate_chip_coord_to_translated`, as `LocalChip` does; those two are the only `get_core` sites (`*_reg`/`dma_*` delegate, both multicasts throw). LITERAL still passes through unchanged, so callers that already resolved the coord are unaffected.

Added `NonTranslatedCoreNamingStillResolvesToAPoolSlot`, which writes via NOC0 and reads via TRANSLATED. I confirmed it catches the regression: reverting only the translation fails it on every core, at exactly the coordinates you predicted — `NOC0 (1,1)` / `TRANSLATED (18,18)` "resolved to different backings".

device/cluster.cpp
xanderchin: Fixed, and you're right that `get_chip_unique_ids()` doesn't deliver it — verified both paths you named.

Rather than derive a board id, `ClusterDescriptor` now reports provenance: `has_authentic_chip_unique_ids()` is true only for a real `chip_unique_ids` block or real ASIC discovery (`topology_discovery.cpp:431`), and `Cluster` withholds a synthesized id so `SWEmuleChip`'s existing refusal fires. Provenance propagates through the `TT_VISIBLE_DEVICES` filter and the chip-id remap — filtering doesn't make an invented id authentic.

Worth noting our multi-rank runs couldn't have caught this: every descriptor we use carries real `chip_unique_ids`, so the synthesized path is unreachable from them. Dead `cluster_desc != nullptr` dropped.

device/chip/sw_emule_chip.cpp
xanderchin: Correct, and the CI blindness is the real defect — a green PR over an uncompiled file is exactly how we shipped a broken JIT path before.

The stale pin is a merge-order dependency: `chip_store.hpp` and the 3-arg `L1Pool` are authored in the tt-emule PR under review, so the pin can only move once that lands. Order is tt-emule → pin bump here → this PR → tt-metal → tt-blaze.

I verified locally with `-DTT_UMD_BUILD_EMULE=ON -DTT_EMULE_PATH=<local checkout>`: builds clean, tests pass. One caveat for the lane you suggest — tt-metal's pin `33a7e52e` has diverged from umd main (not an ancestor) and tt-metal fails against main at `tt_cluster.cpp:778`, so it can't simply be "build tt-metal against this PR".

device/chip/sw_emule_chip.cpp
xanderchin: Fixed — `pool_size = slot_of_.size()`, fallback gone. You're right it was dead and costly: nothing could index those slots, yet it committed 256 MB of the MAP_32BIT window. Restored the reasoning I'd deleted from base.

device/chip/sw_emule_chip.cpp
xanderchin: Fixed — now includes `umd/device/utils/error.hpp`. Agreed it only compiled by include-order accident, and `error::RuntimeError` lives in the public header anyway. `<stdexcept>` dropped.

device/chip/sw_emule_chip.cpp
xanderchin: Fixed. You're right to the line: `grendel_implementation.hpp:268` is 1572864 and `quasar_32_arch.yaml` is 1499136 — only `quasar_simulation_*.yaml` is 4 MiB. Reworded to name the descriptor rather than the arch, and added `RefusesDescriptorWhoseL1ExceedsTheSlotStride` using that existing fixture.

device/chip/sw_emule_chip.cpp
xanderchin: Partly fixed. Confirmed only `tensix_harvesting_mask` feeds the Tensix core list, so the key over-discriminates and omits two masks as you say.

Narrowing it changes the segment NAME, which tt-emule's peer resolver and a tt-metal parser both depend on, so it has to move across three repos together. Filing an issue rather than doing it half-way here — happy to be overruled if you'd rather it block.

device/chip/sw_emule_chip.cpp
xanderchin: Fixed — `log_debug` for the non-shared branch, warn-once naming the arch for the shared one. Largely moot after the translation fix, but the branch stays reachable for ETH/PCIE/ARC.

device/api/umd/device/chip/sw_emule_chip.hpp
xanderchin: Fixed — members moved back next to their own comments, class doc reconciled with the member comment.

The ABI paragraph is deleted rather than trimmed: your `sizeof` arithmetic is right, and the only consumer builds umd from source as a submodule, so the note was both wrong and moot.

device/api/umd/device/chip/sw_emule_chip.hpp
xanderchin: You're right — a defaulted parameter doesn't break source compatibility, and my comment asserted the opposite while contradicting the ABI note below it. Collapsed to one `explicit` ctor with `= std::nullopt`.

device/api/umd/device/chip/sw_emule_chip.hpp
xanderchin: Fixed, and it was worse than "no in-repo caller": nothing called these anywhere, in this repo or the three sibling PRs — tt-metal was rebuilding the map itself from the peer `SocDescriptor`.

That's two implementations of one cross-process invariant, so `build_worker_slot_map(const SocDescriptor&)` is now the single source and tt-metal calls it. Free function, not the member you suggested, because a peer-owned chip has no local `SWEmuleChip`. `slot_of` returns `std::optional`; kept `tt_xy_pair` since peer callers start from a NOC address with no `CoreCoord`. Doc pointer dropped from the installed header.

device/chip/sw_emule_chip.cpp
xanderchin: Not reachable, but taking the assert. `chip_store_map_named_ex` ftruncates as creator and as attacher requires `st_size == bytes` exactly, throwing otherwise — so a 128-slot segment can't silently serve a 72-slot request and `slot_ptr` can't return null here. That also makes adding `pool_size` to the key unnecessary. Added the bounds check anyway; it's free.

device/chip/sw_emule_chip.cpp
xanderchin: Agreed, and done — five tests: slot determinism across two independent builds, pool sizing, the cross-naming write/read, the L1-stride throw on `quasar_simulation_1x1.yaml`, and both uid provenance paths.

Also documenting the behaviour change in the PR body, as you asked: construction now hard-fails on `worker_l1_size > SLOT_SIZE` where base silently overran.

device/api/umd/device/chip/sw_emule_chip.hpp
xanderchin: Which accessor do you mean? `get_chip_unique_ids()` is what this used, and it's the one that isn't stable on the legacy/mock paths (see my reply on `cluster.cpp:255`) — chip 0 gets uid 0 in both. If there's a newer board-derived id I should key on instead, I'd rather use it than the provenance flag I added.

device/chip/sw_emule_chip.cpp
xanderchin: Took this one seriously and implemented it — narrowed the key to `tensix_harvesting_mask`, updated tt-metal's decoder to match, rebuilt. **K=2 then failed deterministically**: `kernel on core (12,8) failed`, `delivery_sum=0`, and `TT_EMULE_SHM_TRACE` showed *zero* peer lookups. A/B confirmed it (stash the change, rebuild → `2 passed`).

Root cause, from a standalone probe against libtt-umd:

| eth mask | constructing a Blackhole SocDescriptor |
|---|---|
| 288 (what the packed name carries) | OK, 140 tensix cores |
| 0 (what tensix-only yields) | **throws** — "Exactly 2 or 14 ETH cores should be harvested on full Blackhole" (`blackhole_coordinate_manager.cpp:69-71`) |

Your analysis of *which masks feed the core list* is correct — only `tensix_harvesting_mask` does. But a peer rebuilds this chip's `SocDescriptor` from the segment name and must get a **constructible** one before it can ask for cores at all, and Blackhole validates the ETH mask at construction. On our descriptor `tensix` and `dram` are both 0 and `eth` is 288, so the tensix mask carries no information — narrowing the name deletes the only field that distinguishes it, and every peer rebuild throws.

So the key can't narrow. What I did take from the comment:

- **You were right that my rationale was wrong.** It said the masks are there because they change the core list. They aren't — they're there so the peer rebuild constructs at all. Comment rewritten to say that.
- **Naming**: `HARVEST_FIELD_BITS` / `HARVEST_FIELD_MASK`.
- **pcie/l2cpu**: no longer silent. Five 20-bit fields don't fit a 64-bit name component, and I verified dropping a non-zero pcie mask (ours is 2) yields an *identical* 140-core translated Tensix list — so it's safe today. Documented, with a note to revisit if that stops holding.
- The width guard stays and is now load-bearing: it protects the packing rather than being decorative.

Re-verified after the rewrite: K=2 `2 passed`, K=4 `1 passed` on all 4 ranks, zero kernel failures.

Closing tenstorrent/tt-umd#3272 as won't-do, since its premise doesn't survive the experiment.

PR #3212, opened 2026-08-21
[Base API] Add WormholeArcWindow component

nbuncicTT:
## Issue

Wormhole ARC APB register access lived inside `WormholeTTDevice`, so anything else needing an ARC register had to go through a whole `TTDevice`.

## Description

Extracts that access into its own component, built from protocol interfaces instead of a `TTDevice`: the device protocol plus exactly one of the PCIe, JTAG and remote interfaces, which is what picks the route.

The ARC tile exposes several windows that differ only in where they sit and whether they hold registers or memory, so the component takes those as a `Config` and is named `WormholeArcWindow`; APB is `WormholeArcWindow::arc_apb()`. The CSM window follows in #3213 as a second config rather than a second class.

Routing is unchanged from `WormholeTTDevice::read_from_arc_apb`/`write_to_arc_apb`: over the NOC when the device is reached remotely, over JTAG when it is reached that way, and otherwise through the PCIe BAR. One deliberate difference: the JTAG branch used to be pinned to `wormhole::ARC_CORES_NOC0[0]` and `NocId::DEFAULT_NOC`, and now uses the coordinate and NOC the caller passed, so the component holds no NOC state. That is documented on the class and on the parameters.

Two defects inherited from the original are fixed while it is being made public:
- the range check validated only the first byte, and against the window size with `>`, so a word access at `ARC_APB_ADDRESS_RANGE` ran past the end;
- `size` was honored only on the remote path, so a caller passing anything other than a word to the JTAG or BAR path got an overrun or a short access instead of an error.

## List of the changes

- New `WormholeArcWindow` with `read`/`write` and an `arc_apb()` factory.
- Whole-transfer bounds check; non-word accesses rejected on the word-sized routes.
- Component tests covering all three routes, the transport invariant and both checks.
- New `tests/test_utils/protocol_mocks.hpp` (same mocks as #3206, plus a `MockRemoteInterface`).
- Added to the library sources and the installed `api` FILE_SET.

## Testing

`WormholeArcApbWindowTest`, 10 gmock tests, no card needed.

## API Changes

- New public header `umd/device/tt_device/firmware/wormhole_arc_window.hpp`.

Stacked on #3207.


device/tt_device/firmware/wormhole_arc_apb.cpp
```diff
@@ -0,0 +1,80 @@
...
+    if (arc_addr_offset > wormhole::ARC_APB_ADDRESS_RANGE) {
+        UMD_THROW(error::RuntimeError, "Address is out of ARC APB address range.");
+    }
+
+    if (remote_interface_ != nullptr) {
+        device_protocol_->read_ctrl(
+            mem_ptr, arc_core, wormhole::ARC_APB_NOC_BASE_ADDRESS + arc_addr_offset, size, noc_id);
+        return;
+    }
+    if (jtag_interface_ != nullptr) {
```
pjanevskiTT: The JTAG route is no longer pinned to NOC0. `WormholeTTDevice::read_from_arc_apb` hardcodes `wormhole::ARC_CORES_NOC0[0]` and `NocId::DEFAULT_NOC` on this branch regardless of the thread's selected NOC; here it takes the caller's `arc_core` and `noc_id`. Only the remote branch used a NOC-dependent core (`get_arc_core()`).

Arguably the better contract, but it means "routing is unchanged" holds only if the call site added in a later PR passes NOC0 coordinates for JTAG devices. Worth either documenting that on `read`/`write`, or keeping this branch explicitly on NOC0.

device/api/umd/device/tt_device/firmware/wormhole_arc_apb.hpp
```diff
@@ -0,0 +1,74 @@
...
+        DeviceProtocol* device_protocol,
+        PcieInterface* pcie_interface,
+        JtagInterface* jtag_interface,
+        RemoteInterface* remote_interface);
+
+    /**
+     * @brief Reads from the ARC APB window.
+     * @param mem_ptr Destination buffer.
+     * @param arc_addr_offset Offset into the ARC APB window.
+     * @param size Bytes to read; only honored on the remote path, the other paths read one word.
```
pjanevskiTT: `size` being honored only on the remote path is now part of a public component's contract, and the two other paths write a full `uint32_t` into `mem_ptr` regardless — a caller passing `size = 2` with a 2-byte buffer gets an overrun rather than an error. Consider asserting `size == sizeof(uint32_t)` on the JTAG/BAR paths instead of silently widening.

Related: the range check is `arc_addr_offset > ARC_APB_ADDRESS_RANGE` and ignores `size`, so `arc_addr_offset == ARC_APB_ADDRESS_RANGE` passes and a word read runs 3 bytes past the window. Both are inherited from `WormholeTTDevice`, but they were private there.

device/api/umd/device/tt_device/firmware/wormhole_arc_apb.hpp
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * @brief Access to the Wormhole ARC APB register window.
+ *
+ * A single access is routed one of three ways, in order: over the NOC when the device is reached
+ * remotely, over JTAG when it is reached that way, and otherwise through the PCIe BAR. Callers pass
+ * the ARC core coordinate and the NOC to route over, so this holds no NOC state of its own.
+ *
+ * The interfaces are non-owning and must outlive this object.
+ */
+class WormholeArcApb {
```
pjanevskiTT: maybe add few tests for the component?

pjanevskiTT: do we have this for Blackhole as well? I guess it can look the same just with other addresses and ranges?

pjanevskiTT: In the future we can also include enums for all APB registers so we change to API maybe to something like read_reg(register enum), but we are far away from it

pjanevskiTT (approved):

device/tt_device/firmware/wormhole_arc_apb.cpp
```diff
@@ -0,0 +1,80 @@
...
+        transports == 1,
+        error::RuntimeError,
+        "WormholeArcApb requires exactly one of a PcieInterface, a JtagInterface or a RemoteInterface, since which "
+        "one is present is how it picks the route for an access.");
+}
+
+void WormholeArcApb::read(void* mem_ptr, uint64_t arc_addr_offset, size_t size, tt_xy_pair arc_core, NocId noc_id) {
+    if (arc_addr_offset > wormhole::ARC_APB_ADDRESS_RANGE) {
+        UMD_THROW(error::RuntimeError, "Address is out of ARC APB address range.");
+    }
```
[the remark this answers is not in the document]
nbuncicTT: Fixed. The bound now covers the whole transfer rather than the first byte, and rejects zero-length accesses:

```cpp
UMD_ASSERT(size != 0, ...);
UMD_ASSERT(size <= config_.size_bytes && arc_addr_offset <= config_.size_bytes - size, ...);
```

Written as a subtraction so it cannot wrap. `ARC_APB_ADDRESS_RANGE` is the window *size*, not its last valid offset, so the old `>` comparison also let a word access at exactly `ARC_APB_ADDRESS_RANGE` through — that is covered now too.

Both windows share this check since the APB/CSM classes were merged into `WormholeArcWindow`. Test: `AccessMustFitEntirelyInsideTheWindow`.

device/tt_device/firmware/wormhole_arc_apb.cpp
nbuncicTT: Good catch, and you are right that the PR description was the wrong part. The behaviour is deliberate — we want the component to hold no NOC state — so the caller-passed `arc_core`/`noc_id` stay, but I removed the blanket "routing is unchanged" claim and documented the difference on the class and on both `arc_core` parameters:

> Taking both per call is a deliberate difference from `WormholeTTDevice::read_from_arc_apb`, which pinned its JTAG branch to `wormhole::ARC_CORES_NOC0[0]` and `NocId::DEFAULT_NOC` whatever NOC the thread had selected [...] a caller that wants the old JTAG behaviour has to pass the NOC0 coordinate itself.

Test `JtagReadUsesTheCallersCoreAndNoc` pins that the JTAG route uses what it was given.

device/api/umd/device/tt_device/firmware/wormhole_arc_apb.hpp
nbuncicTT: Both fixed. The JTAG and BAR paths now reject any size they cannot honour instead of silently widening:

```cpp
UMD_ASSERT(remote_interface_ != nullptr || size == sizeof(uint32_t), ...);
```

so `size = 2` throws rather than overrunning a 2-byte buffer, and an oversized request throws rather than short-reading. Every in-tree caller passes exactly 4 bytes, so nothing changes in practice — it only closes the public contract.

The range check in your second paragraph is fixed as described in the Copilot thread. Tests: `NonWordSizeIsRejectedOnTheWordSizedRoutes`, `AccessMustFitEntirelyInsideTheWindow`.

device/api/umd/device/tt_device/firmware/wormhole_arc_apb.hpp
nbuncicTT: Added — `tests/baremetal/test_wormhole_arc_window.cpp`, 10 tests for the APB window (6 more for CSM in #3213). Covers all three routes, the exactly-one-transport invariant, and both access checks. Runs in the no-card baremetal lane.

Also adds `tests/test_utils/protocol_mocks.hpp`, matching the mocks the Blackhole component tests use, plus a `MockRemoteInterface` for the remote route.

nbuncicTT: Yes — `BlackholeArcApb`, added in #3203 and already merged.

They are deliberately not the same class though: Blackhole routes on JTAG → is-ARC-reachable-over-AXI → BAR, while Wormhole routes on remote → JTAG → BAR, so the branch *order* differs rather than just the constants. Your CSM comment on #3213 does apply, however — the two Wormhole windows do have identical branch order, so those two are now one class.

nbuncicTT: Agreed, and it would be a clear improvement over passing raw offsets. Out of scope here — this PR is a straight extraction plus the two checks above — but noting it as a follow-up on the Base API work.

PR #3213, opened 2026-08-21
[Base API] Add the ARC CSM window

nbuncicTT:
## Issue

Wormhole ARC CSM access lived inside `WormholeTTDevice`. It has consumers beyond the device: `FirmwareInfoProvider` reaches through a `TTDevice` to read the runtime telemetry address and size from CSM.

## Description

CSM is the same three routes as the APB window with three constants changed, so it is added as a second `WormholeArcWindow` config rather than a second class. `WormholeArcWindow::arc_csm()`.

The one real difference is preserved: CSM holds memory rather than registers, so a remote access takes the data path where APB takes the register path. That is now a `Content` field instead of a duplicated branch. The JTAG route stays on the register path for both windows, as `WormholeTTDevice::read_from_arc_csm` did — the old class comment claimed memory-vs-registers as the distinguishing property and then contradicted it on that branch, so `Content` documents that it selects the remote route only.

Blackhole has no counterpart: its CSM accessors only throw, so there is nothing to extract there.

## List of the changes

- `WormholeArcWindow::arc_csm()` and the ARC CSM window config.
- Component tests for the CSM window: the data-path remote route, the register-path JTAG route, the CSM BAR offset and the CSM bound.

## Testing

`WormholeArcCsmWindowTest`, 6 gmock tests, no card needed.

## API Changes

- New `WormholeArcWindow::arc_csm()` factory. No new header.

Stacked on #3212.


device/api/umd/device/tt_device/firmware/wormhole_arc_csm.hpp
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * @brief Access to the Wormhole ARC CSM memory window.
+ *
+ * Same three routes as WormholeArcApb (remote over the NOC, JTAG, then the PCIe BAR), but the remote
+ * path is a data access rather than a register access, and the offsets are the CSM ones. Callers pass
+ * the ARC core coordinate and the NOC to route over, so this holds no NOC state of its own.
+ *
+ * The interfaces are non-owning and must outlive this object.
+ */
+class WormholeArcCsm {
```
pjanevskiTT: This is `WormholeArcApb` with three constants changed and `read_ctrl` -> `read_data` on one branch: same constructor, same exactly-one-transport invariant, same branch order, same doc comments. The PR description argues against a shared base for the Wormhole/Blackhole pair because the branch orders differ and a base would need about four virtual hooks — that reasoning doesn't apply here, where the branch order is identical.

A single `WormholeArcWindow{noc_base, bar0_base, address_range, remote_access_kind}` constructed twice would drop ~140 duplicated lines and keep the two windows from drifting.

device/tt_device/firmware/wormhole_arc_csm.cpp
```diff
@@ -0,0 +1,79 @@
...
+    if (remote_interface_ != nullptr) {
+        // CSM is memory rather than registers, so the remote path is a data access. Mirrors
+        // WormholeTTDevice::read_from_arc_csm, which uses read_from_device here and
+        // read_from_device_reg on the APB path.
+        device_protocol_->read_data(
+            mem_ptr, arc_core, wormhole::ARC_CSM_NOC_BASE_ADDRESS + arc_addr_offset, size, noc_id);
+        return;
+    }
+    if (jtag_interface_ != nullptr) {
+        device_protocol_->read_ctrl(
```
pjanevskiTT: The remote branch uses `read_data` because "CSM is memory rather than registers", but the JTAG branch on the same memory window uses `read_ctrl`. That matches `WormholeTTDevice::read_from_arc_csm` so it isn't a regression, but the class comment states memory-vs-registers as this component's distinguishing property and this branch contradicts it. Either use `read_data` here too, or note why JTAG stays on the control path.

device/api/umd/device/tt_device/firmware/wormhole_arc_csm.hpp
pjanevskiTT: can we add some tests for the component?

pjanevskiTT: same as for APB, do we have this for BH as well?

pjanevskiTT (approved):

device/tt_device/firmware/wormhole_arc_csm.cpp
```diff
@@ -0,0 +1,79 @@
...
+        transports == 1,
+        error::RuntimeError,
+        "WormholeArcCsm requires exactly one of a PcieInterface, a JtagInterface or a RemoteInterface, since which "
+        "one is present is how it picks the route for an access.");
+}
+
+void WormholeArcCsm::read(void* mem_ptr, uint64_t arc_addr_offset, size_t size, tt_xy_pair arc_core, NocId noc_id) {
+    if (arc_addr_offset > wormhole::ARC_CSM_ADDRESS_RANGE) {
+        UMD_THROW(error::RuntimeError, "Address is out of ARC CSM address range.");
+    }
```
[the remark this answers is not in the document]
nbuncicTT: Fixed, and now shared rather than duplicated — the CSM and APB classes were merged into one `WormholeArcWindow`, so there is a single bound check:

```cpp
UMD_ASSERT(size != 0, ...);
UMD_ASSERT(size <= config_.size_bytes && arc_addr_offset <= config_.size_bytes - size, ...);
```

Each window is bounded by its own `size_bytes`, so CSM is checked against `ARC_CSM_ADDRESS_RANGE`. The word-sized JTAG/BAR paths additionally reject any size other than `sizeof(uint32_t)`. Test: `AccessIsBoundedByTheCsmWindow`.

device/api/umd/device/tt_device/firmware/wormhole_arc_csm.hpp
nbuncicTT: Agreed, and done as you described. `WormholeArcApb` and `WormholeArcCsm` are gone; there is one `WormholeArcWindow` holding the routing once, with the differences as data:

```cpp
struct Config { const char* name; uint64_t noc_base_address;
                uint32_t bar0_offset_start; uint32_t size_bytes; Content content; };

static WormholeArcWindow arc_apb(protocol, pcie, jtag, remote);   // #3212
static WormholeArcWindow arc_csm(protocol, pcie, jtag, remote);   // #3213
```

`Content` is an enum rather than a bool so the register-vs-memory distinction is named. Factories rather than public config constants so the `wormhole::` address constants stay out of the public header.

You were right that my Wormhole-vs-Blackhole argument did not transfer: those two differ in branch *order*, these two do not. #3212 now introduces the generic window and this PR is down to a factory plus a config — 3 files, +125/-1.

device/tt_device/firmware/wormhole_arc_csm.cpp
nbuncicTT: You were right that the comment contradicted the code. The behaviour is unchanged from `WormholeTTDevice`, so I documented the asymmetry rather than altering the routing — it is now a property of the `Content` enum:

> This only selects the remote route's access kind. The JTAG route issues a register access for every window, memory ones included, which is what `WormholeTTDevice::read_from_arc_csm` did: it is reading a single word out of the ARC's address space over a debug transport, where the register path is the one that is wired up, so the distinction does not arise.

Test `JtagReadStaysOnTheRegisterPath` pins it so it cannot drift silently.

device/api/umd/device/tt_device/firmware/wormhole_arc_csm.hpp
nbuncicTT: Added — 6 tests in `tests/baremetal/test_wormhole_arc_window.cpp` alongside the APB ones. They cover what actually differs for CSM: the data-path remote route, the register-path JTAG route, the CSM BAR offset, and the CSM window bound.

nbuncicTT: No, and nothing to extract — `BlackholeTTDevice::read_from_arc_csm` and `write_to_arc_csm` only throw "CSM read/write not supported for Blackhole". Noted in the class doc so the asymmetry is not mistaken for an oversight.

PR #3223, opened 2026-08-22
Move TLB layout to ArchitectureTlbs

broskoTT:
### Issue

### Description
Stacked on #3222.

The TLB window layout is not part of the Base API specification, so it moves out of
`ArchitectureImplementation` into `ArchitectureTlbs`, filled per architecture in one file, same
shape as `ArchitectureRegisters`. Resolving a window index to its configuration is per
architecture logic, so that one entry is a lookup and its implementation stays private to that
file.

`PCIDevice` and `SimulationTlbAllocator` are left holding an `ArchitectureImplementation` they no
longer read anything from. Dropping it is a small follow up PR.

### List of the changes
- Add `ArchitectureTlbs` with the base and count per window size, the available sizes, the cached
  window size, the static VC flag, the config register address and size, and the window
  configuration lookup.
- Remove the ten TLB getters from `ArchitectureImplementation` and its implementations, and move
  the window configuration resolution out of the arch implementations.
- Update the TLB call sites in PCIDevice, the PCIe protocol, TLB manager, local chip, TTDevice,
  the simulation TLB allocator and the tlb_virus tool.

### Testing
Code builds. TLB allocation and device IO tests cover these paths in CI.

### API Changes
This PR has API changes, ten methods removed from `ArchitectureImplementation` and a new
`ArchitectureTlbs`.

### Full stack diff

### DIFF
--- a/device/api/umd/device/arch/architecture_tlbs.hpp
+++ b/device/api/umd/device/arch/architecture_tlbs.hpp
@@ -1,0 +1,47 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+#include <utility>
+#include <vector>
+
+#include "umd/device/types/arch.hpp"
+#include "umd/device/types/tlb.hpp"
+
+namespace tt::umd {
+
+// Resolves a TLB window index to its configuration.
+using TlbConfigurationResolver = tlb_configuration (*)(uint32_t tlb_index);
+
+// The TLB window layout of one architecture. Architecture specific code reads its own constants
+// directly; this exists for callers which are not tied to one architecture.
+struct ArchitectureTlbs {
+    // Base index and window count per window size. A zero count means the architecture has no
+    // windows of that size.
+    std::pair<uint32_t, uint32_t> base_and_count_1m;
+    std::pair<uint32_t, uint32_t> base_and_count_2m;
+    std::pair<uint32_t, uint32_t> base_and_count_16m;
+    std::pair<uint32_t, uint32_t> base_and_count_4g;
+
+    // Window sizes the architecture provides, smallest first.
+    std::vector<size_t> sizes;
+
+    // Preferred window size, the group with the largest count available.
+    size_t cached_window_size;
+
+    // Whether static_vc should be used for window configuration.
+    bool use_static_vc;
+
+    uint32_t static_cfg_addr;
+    uint64_t cfg_reg_size_bytes;
+
+    TlbConfigurationResolver get_configuration;
+};
+
+const ArchitectureTlbs& get_architecture_tlbs(const tt::ARCH arch);
+
+}  // namespace tt::umd
--- a/device/CMakeLists.txt
+++ b/device/CMakeLists.txt
@@ -63,6 +63,7 @@ target_sources(
     PRIVATE
         arch/architecture_implementation.cpp
         arch/architecture_registers.cpp
+        arch/architecture_tlbs.cpp
         chip/chip.cpp
         chip/local_chip.cpp
         chip/mock_chip.cpp
@@ -185,6 +186,7 @@ if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.23)
             FILES
                 api/umd/device/arch/architecture_implementation.hpp
                 api/umd/device/arch/architecture_registers.hpp
+                api/umd/device/arch/architecture_tlbs.hpp
                 api/umd/device/arc/arc_messenger.hpp
                 api/umd/device/arc/arc_telemetry_reader.hpp
                 api/umd/device/arc/firmware_telemetry_reader.hpp
--- a/device/api/umd/device/arch/architecture_implementation.hpp
+++ b/device/api/umd/device/arch/architecture_implementation.hpp
@@ -16,7 +16,6 @@
 #include "umd/device/types/cluster_types.hpp"
 #include "umd/device/types/core_coordinates.hpp"
 #include "umd/device/types/risc_type.hpp"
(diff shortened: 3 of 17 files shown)


device/api/umd/device/arch/architecture_tlbs.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include "umd/device/types/tlb.hpp"
+
+namespace tt::umd {
+
+// Resolves a TLB window index to its configuration.
+using TlbConfigurationResolver = tlb_configuration (*)(uint32_t tlb_index);
+
+// The TLB window layout of one architecture. Architecture specific code reads its own constants
+// directly; this exists for callers which are not tied to one architecture.
+struct ArchitectureTlbs {
```
nbuncicTT, line 22: I agree with this refactor in spirit, the TLB layout should not live in `ArchitectureImplementation`. Let's land it and open a follow-up issue for the design concerns below:

- **The window size is encoded in the field names** (`base_and_count_1m/2m/16m/4g`), so a new size means a new field plus a new arm in every consumer, and it forces the `{0, 0}` placeholders and the "a zero count means the architecture has no windows of that size" sentinel.
- **The `base` half of all four pairs is never read** — the only consumer, `get_tlb_count_for_size` in `tools/tlb_virus.cpp:29`, takes `.second` four times and never touches `.first`.
- **`base` also duplicates data the struct already exposes**, since `DYNAMIC_TLB_2M_BASE = TLB_BASE_2M` (`wormhole_implementation.hpp:223`) is the same value `get_configuration` returns as `tlb_configuration.base`.
- **`base` and `count` are not the same kind of thing** — `base` is a byte offset within a BAR while the index base is a separate constant (`TLB_BASE_INDEX_2M = 156` vs `TLB_BASE_2M = TLB_COUNT_1M * (1 << 20)`, `wormhole_implementation.hpp:207-211`) — so an index-keyed lookup is carrying the wrong one.
- **`sizes` is a parallel array over the same window classes**, so the sizes and their base/count pairs have to be kept in agreement by hand in three separate table literals.
- **The three `tlb_configuration_for` functions are one generic operation** — find the class whose index range contains the index and offset within it — which as per-arch data plus a single resolver would remove roughly 100 of the 165 lines in `architecture_tlbs.cpp` along with the `TlbConfigurationResolver` alias and the function pointer.
- **`cfg_addr` and `cached_window_size` are both derivable from fields already in the struct** (`static_cfg_addr + base_index * cfg_reg_size_bytes`, and the class with the largest count), so they are redundant constants that can silently drift from the values beside them.
- **`SimulationTlbAllocator` duplicates numbers this struct now owns** — `count = 156` and `count = 202` at `simulation_tlb_allocator.cpp:136-146` are `wormhole::TLB_COUNT_1M` and `blackhole::TLB_COUNT_2M`, `tlb_reg_size_bytes_` repeats `cfg_reg_size_bytes`, and `tlb_sizes[0]`/`[1]` indexes `sizes` positionally.
- **Out-of-range TLB indices resolve silently instead of throwing**, because Blackhole and Grendel fall through to a 2M configuration for anything outside the 4G range and Wormhole's descending cascade has no upper bound, both yielding a nonsense `index_offset`.
- **The BAR is not represented anywhere**, so Blackhole's 2M windows (BAR0) and 4G windows (BAR4) both report a base of 0 with nothing distinguishing them.

nbuncicTT (approved):

device/api/umd/device/arch/architecture_tlbs.hpp
nbuncicTT: The issues which are resolved by this PR (https://github.com/tenstorrent/tt-umd/pull/3226) you may ignore, sorry for not reading through earlier.

broskoTT: Sorry for wasting your time, unfortunately this is the downside of smaller PRs...

broskoTT: Went through all ten. Six are already handled by #3226, which is in the stack above this one:

- size encoded in field names, `{0, 0}` placeholders and the zero-count sentinel: gone, replaced by a `std::vector<TlbWindowGroup>`
- `base` never read: dropped rather than moved
- `base` duplicating `tlb_configuration.base`: gone with it
- `base` being a BAR byte offset rather than the index base: the group now carries `base_index`, sourced from `TLB_BASE_INDEX_*`
- `sizes` as a parallel array: folded into the group
- BAR not represented: the group carries whether its windows are mapped through BAR4

Four are real and still open, and I checked each against the constants:

- the three `tlb_configuration_for` functions being one generic operation
- `cfg_addr` and `cached_window_size` being derivable. Confirmed both: every arch defines `DYNAMIC_TLB_*_CFG_ADDR = STATIC_TLB_CFG_ADDR + (TLB_BASE_INDEX_* * TLB_CFG_REG_SIZE_BYTES)`, and `cached_window_size` is the largest count group on all three (WH 1MB with 156 windows, BH and Grendel 2MB with 202)
- out of range indices resolving silently instead of throwing
- `tlb_reg_size_bytes_` in `SimulationTlbAllocator` still repeating `cfg_reg_size_bytes` (the duplicated counts and positional size indexing from that point are fixed by #3226)

Those four are one coherent change, since a single resolver driven by per group data fixes the first, third and most of the second. Rather than deferring to an issue I am putting them in a PR on top of the stack and will link it here.

broskoTT: Landed the four remaining points in #3240, on top of the stack: one generic resolver replacing the three per architecture functions and the function pointer, throwing for an index no group owns, `cfg_addr` derived, and the simulation allocator taking its config register size from the architecture.

One deliberate exception: `cached_window_size` stays a stored constant rather than being derived from the largest group. It matches the most numerous window size on all three architectures today, but it is a choice about which pool UMD caches from, so it should stay settable independently.

Issue #2864, opened 2026-06-21
Create TTDeviceModel skeleton with all component interfaces


nbuncicTT:
**Description**

Add the `TTDeviceModel` base class with all component interface declarations as defined in the Base API Specification, no concrete implementations. This establishes the composition root that `TTDevice` will receive at construction. In this phase, `TTDeviceModel` is not wired into `TTDevice` - it exists as standalone code for architecture review.

**Scope**

- Implement `TTDeviceModel` base class with pure virtual getters for all components per the Base API Specification: DeviceProtocol, DeviceFirmware, ArchitectureImplementation, SocArchDescriptor, IoWindow, FirmwareInfoProvider, HangDetector, FirmwareTelemetryReader, PcieInterface, DmaInterface, JtagInterface, RemoteInterface
- Implement stub/empty interfaces for each component where they don't yet exist
- No concrete implementations in this issue
- Not wired into TTDevice

**Timeline**

No hard deadline. Part of the UMD Base API migration epic. This is the first TTDeviceModel issue, all others depend on it.

**Is your feature request related to a problem? Please describe.**

There is no central composition root for device components. Device-specific behavior is scattered across TTDevice subclasses rather than assembled through a model.

**Definition of Done**

- `TTDeviceModel` base class exists with all component interface getters
- All component interfaces are defined (even if empty)
- Compiles cleanly with no concrete implementations

broskoTT:
PRs going in, will add more interfaces as they come in

PR #3225, opened 2026-08-22
[base API] TTDeviceModel skeleton wired into TTDevice

broskoTT:
### Issue
#2864

### Description
Adds the `TTDeviceModel` composition root and hands it to `TTDevice` at
construction. Models are specialized by architecture, and by backend for
simulation -- transport is deliberately not a subclass axis, since the
transport-specific components (DeviceProtocol and its interface facets, system
memory, I/O windows) will be built from the transport object and injected. One
model per architecture therefore serves PCIe, JTAG and remote alike, avoiding an
architecture x transport cross product.

Deviates from the issue on one point: the model is wired into `TTDevice` right
away instead of standing alone. A model nothing constructs cannot be reviewed
for the part that matters -- ownership and construction order -- and wiring it
now puts every creation path under the existing test suite. A model still
carries only device identity; components move in one PR at a time.

### List of the changes
- Add `TTDeviceModel` base with identity accessors and nullptr-returning
optional component getters.
- Add `WormholeTTDeviceModel`, `BlackholeTTDeviceModel` and
`SimulationTTDeviceModel` under `device/tt_device_model/`.
- `TTDevice` now requires a model at construction; drop `arch`,
`communication_device_type_` and `communication_device_id_` and read identity
through the model instead.
- Construct the model in `TTDevice::create` and in the simulation device
constructors.

### Testing
CI

### API Changes
This PR has API changes.

### Full stack diff


device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -523,24 +524,27 @@ class TTDevice {
...
 
 protected:
-    IODeviceType communication_device_type_ = IODeviceType::UNDEFINED;
-    int communication_device_id_ = -1;
     std::unique_ptr<ArchitectureImplementation> architecture_impl_;
-    tt::ARCH arch = tt::ARCH::Invalid;
     LockManager lock_manager;
 
-    TTDevice() = default;
+    // Every TTDevice is built around a model, which supplies its identity and -- as components are
```
nbuncicTT: No need for this comment, it's self explanatory that TTDevice is built around a model.

device/api/umd/device/tt_device_model/tt_device_model.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+    // Device identity.
+    //
+    // TODO: temporary - not part of the Base API. TTDevice derived its identity from the transport
+    // object it used to own; the model answers for it during the migration. These go away once the
+    // components that carry this information (DeviceProtocol, ArchitectureImplementation) have moved
+    // here and can be asked directly.
+    virtual tt::ARCH get_arch() const = 0;
+
+    virtual IODeviceType get_communication_device_type() const = 0;
```
nbuncicTT: This is not a part of the spec, hence state that's it also TODO as with get_arch.

nbuncicTT (approved):

device/api/umd/device/tt_device/tt_device.hpp
broskoTT: Agreed, the first part just restated the signature. Trimmed to the one thing that is not self-evident -- which of the overloads is the transport-less one. That line disappears anyway once the remaining constructors collapse into a single `TTDevice(std::unique_ptr<TTDeviceModel>)` later in the stack.

device/api/umd/device/tt_device_model/tt_device_model.hpp
broskoTT: Done, and went a bit further after re-reading the spec.

- `get_arch()`: marked `TODO: temporary`, naming `get_architecture_impl()->get_architecture()` as what answers it once ArchitectureImplementation moves into the model.
- `get_communication_device_id()`: marked `TODO: temporary`, naming `get_device_protocol()->get_mmio_id()`.
- `get_communication_device_type()`: removed from the model entirely. The spec declares it on TTDevice (tt_device_doxy.hpp:621), not on TTDeviceModel, so TTDevice now derives it from the transport interface it holds -- PCIe/JTAG directly, remote reporting the local device's transport, UNDEFINED otherwise. Same values the constructors used to store, with no stored state.

Knock-on: the silicon models no longer take an `IODeviceType` at all (they only stored it to answer that getter), both remote factory sites stopped reaching into the local device for it, and `tt_device_model.hpp` no longer includes `communication_protocol.hpp`.

device/api/umd/device/tt_device_model/tt_device_model.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+    // Identifies the device within its transport: the PCI device number for PCIe, the JLink id for
+    // JTAG. A remote device reports the identity of the local device it is reached through.
+    int get_communication_device_id() const;
+
+    virtual HangDetector *get_hang_detector() { return nullptr; }
+
+    virtual FirmwareTelemetryReader *get_firmware_telemetry_reader() { return nullptr; }
+
+    virtual PcieInterface *get_pcie_interface() { return nullptr; }
```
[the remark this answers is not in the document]
broskoTT: Declining this one, for three reasons.

1. It asks for these as *nullable* getters, but per the Base API spec DeviceProtocol, DeviceFirmware, ArchitectureImplementation, SocArchDescriptor, IoWindow, FirmwareInfoProvider and SystemMemoryAllocator are **required** components -- pure virtual, not nullptr-defaulting. Only HangDetector, FirmwareTelemetryReader and the four transport interfaces are optional, and those are the ones declared here.

2. Three of them do not exist as types yet in this repo -- DeviceFirmware, IoWindow and SystemMemoryAllocator. Declaring getters for them would mean inventing stub interfaces, which is explicitly not what this PR is doing: components are declared here as they are actually decoupled from TTDevice, so the header never advertises something nothing implements.

3. It is already addressed by the rest of the stack rather than being missing: DeviceProtocol lands in #3236, SocArchDescriptor in #3237, ArchitectureImplementation in #3254, HangDetector in #3264. The class comment carries a TODO noting components arrive incrementally.

PR #3226, opened 2026-08-22
Describe TLB windows as size classes instead of one field per size

broskoTT:
### Issue

### Description
Stacked on #3224.

`ArchitectureTlbs` had a base and count field per window size plus a separate list of sizes, so
architectures filled `{0, 0}` for sizes they do not have and callers matched size literals to find
a count. One list of size classes replaces all of it. The class also carries the base window
index and the BAR4 mapping, both of which the architecture already defines.

`SimulationTlbAllocator` kept a near copy of the same structure, so its pools are built from the
classes now instead of a fixed array indexed by a size class enum.

### List of the changes
- Replace the four base and count fields and the size list with a list of `TlbSizeClass`.
- Drop the unused base offset, which was a BAR offset rather than the window index it claimed.
- `SimulationTlbAllocator` keeps one pool per class, so its size class enum, its duplicated window
  counts and its index layout pass are gone.

### Testing
Code builds. Simulation TLB allocator unit tests pass, including the address from index cases.

### API Changes
This PR has API changes, `ArchitectureTlbs` describes size classes instead of a field per size.

### Full stack diff

### DIFF
--- a/device/api/umd/device/arch/architecture_tlbs.hpp
+++ b/device/api/umd/device/arch/architecture_tlbs.hpp
@@ -6,7 +6,6 @@
 
 #include <cstddef>
 #include <cstdint>
-#include <utility>
 #include <vector>
 
 #include "umd/device/types/arch.hpp"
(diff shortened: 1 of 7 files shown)


device/api/umd/device/arch/architecture_tlbs.hpp
```diff
@@ -6,29 +6,30 @@
...
-#include <utility>
 #include <vector>
 
 #include "umd/device/types/arch.hpp"
 #include "umd/device/types/tlb.hpp"
 
 namespace tt::umd {
 
+// A set of TLB windows which all share one size, contiguous in the window index space.
+struct TlbWindowGroup {
```
nbuncicTT: nit: `TlbSizeClass` makes more sense for me than `TlbWindowGroup`

device/api/umd/device/chip_helpers/simulation_tlb_allocator.hpp
```diff
@@ -67,31 +67,27 @@ class SimulationTlbAllocator {
...
 
 private:
-    // A pool of TLB indices that all share a single size.
-    struct TlbSizeClass {
-        size_t size = 0;         // Bytes per TLB; 0 means unused on the current arch.
-        size_t count = 0;        // Number of TLBs in the pool.
-        size_t start_index = 0;  // First global TLB index belonging to this pool.
+    // A window group and which of its windows are currently allocated.
+    struct TlbPool {
+        TlbWindowGroup group;
```
nbuncicTT: nit: `TlbSizeClass layout;` would make more sense for me if you'd adjust the group to size change from the first nit comment.

nbuncicTT (approved):

device/api/umd/device/arch/architecture_tlbs.hpp
broskoTT: Renamed to `TlbSizeClass`, and the field it lives in to `size_classes` so the two read together. Loop variables and the comments that said window group follow.

One thing to flag: the struct now carries the size, window count, base window index and which BAR the windows are in, so it is a hardware description rather than only a size bucket. Size class is allocator vocabulary, but it matches the wording already used in the simulation allocator, so happy to go with it.

device/api/umd/device/chip_helpers/simulation_tlb_allocator.hpp
broskoTT: Done, the pool now holds `TlbSizeClass layout` and reads as `pool.layout.size` and so on.

PR #3227, opened 2026-08-22
Remove Tensix location and UBB tray getters from ArchitectureImplementation

broskoTT:
### Issue

### Description
Stacked on #3226.

Neither of these is in the Base API specification, and neither needs a new home.

`get_t6_x_locations` and `get_t6_y_locations` had one caller, the legacy ETH broadcast path, which
was already constructing a Wormhole implementation explicitly to reach them. It reads the Wormhole
constants directly now.

`get_ubb_tray_id` mapped a PCI bus id to a UBB tray. `ClusterDescriptor::get_tray_id` already
knows the board type before asking, so it picks the tray table from the board rather than
constructing an architecture implementation to dispatch on the architecture. The tables stay per
architecture, the lookup over them is written once.

### List of the changes
- Remove `get_t6_x_locations`, `get_t6_y_locations` and `get_ubb_tray_id` from the interface and
  its implementations.
- The ETH broadcast helpers read the Wormhole Tensix locations directly.
- `ClusterDescriptor::get_tray_id` selects the tray table by board type.

### Testing
Code builds. Cluster descriptor and Wormhole broadcast paths are covered in CI.

### API Changes
This PR has API changes, three methods removed from `ArchitectureImplementation`.

### Full stack diff

### DIFF
--- a/device/tt_device/ethernet_broadcast.cpp
+++ b/device/tt_device/ethernet_broadcast.cpp
@@ -16,7 +16,6 @@
 #include <unordered_map>
 #include <vector>
 
-#include "umd/device/arch/architecture_implementation.hpp"
 #include "umd/device/arch/wormhole_implementation.hpp"
 #include "umd/device/tt_device/remote_communication.hpp"
 #include "umd/device/types/arch.hpp"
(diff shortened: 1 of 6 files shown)


device/tt_device/ethernet_broadcast.cpp
```diff
@@ -328,8 +323,7 @@ void EthernetBroadcast::broadcast_write_to_cluster(
         }
     } else {
         UMD_ASSERT(
-            use_translated_coords or
-                valid_tensix_broadcast_grid(rows_to_exclude_virtual, cols_to_exclude_virtual, arch_impl.get()),
+            use_translated_coords or valid_tensix_broadcast_grid(rows_to_exclude_virtual, cols_to_exclude_virtual),
```
nbuncicTT: nit: Instead of using the modern `or` be consistent with the rest of the code base and use `||`

nbuncicTT (approved):

device/tt_device/ethernet_broadcast.cpp
broskoTT: Changed to `||`.

Left the other one alone for now: line 289 in the same function still has `cols_to_exclude_virtual.find(0) == cols_to_exclude_virtual.end() or ...`, which this PR does not otherwise touch. Happy to sweep it in the same change if you would rather the whole function be consistent.

PR #3228, opened 2026-08-22
Align ArchitectureImplementation naming with the Base API specification

broskoTT:
### Issue

### Description
Stacked on #3227.

Last of the series. With the misplaced methods gone, this renames and retypes what is left to
match the Base API specification, so the interface reads as the specification describes it.

The idle AICLK frequency the specification asks for already existed as `TTDevice::get_min_clock_freq`
with per architecture overrides reading `AICLK_IDLE_VAL`. It moves onto the interface and the
concrete devices delegate, so the value has one definition. The busy frequency does not stay, it had a single
consumer in `test_firmware_info_provider` which now resolves `AICLK_BUSY_VAL` for its own
architecture, so the interface carries only what the specification lists.

### List of the changes
- Rename the two power state messages to `get_firmware_message_go_busy` and `_go_idle`, and drop
  the `arc_` prefix from the reset unit refclk offsets.
- Widen the refclk offsets and the Tensix soft reset address to `uint64_t`.
- Add `get_min_clock_freq` and remove `get_aiclk_busy_val`.
- Document what `get_harvesting_noc_locations` returns.

### Testing
Code builds. ARC message, clock and soft reset paths are covered in CI.

### API Changes
This PR has API changes, renamed and retyped methods on `ArchitectureImplementation`.

### Full stack diff

### DIFF
--- a/device/api/umd/device/arch/architecture_implementation.hpp
+++ b/device/api/umd/device/arch/architecture_implementation.hpp
@@ -8,7 +8,6 @@
 #include <cstdint>
 #include <iterator>
 #include <memory>
-#include <optional>
 #include <utility>
 #include <vector>
 
@@ -43,16 +42,10 @@ public:
     virtual uint32_t get_soft_reset_staggered_start() const = 0;
     // Replace with std::span once we enable C++20.
     virtual const std::vector<uint32_t>& get_harvesting_noc_locations() const = 0;
-    virtual const std::vector<uint32_t>& get_t6_x_locations() const = 0;
-    virtual const std::vector<uint32_t>& get_t6_y_locations() const = 0;
 
     virtual DeviceL1AddressParams get_l1_address_params() const = 0;
 
     static std::unique_ptr<ArchitectureImplementation> create(tt::ARCH architecture);
-
-    // Map a PCI bus id to a UBB tray id (1..4). Returns std::nullopt for archs without UBB
-    // boards or when the bus id does not correspond to a known tray.
-    virtual std::optional<uint8_t> get_ubb_tray_id(uint16_t bus_id) const { return std::nullopt; }
 };
 
 }  // namespace tt::umd
--- a/device/api/umd/device/arch/architecture_implementation.hpp
+++ b/device/api/umd/device/arch/architecture_implementation.hpp
@@ -29,17 +29,20 @@ public:
     virtual ~ArchitectureImplementation() = default;
 
     virtual tt::ARCH get_architecture() const = 0;
-    virtual uint32_t get_arc_message_arc_go_busy() const = 0;
-    virtual uint32_t get_arc_message_arc_go_long_idle() const = 0;
-    virtual uint32_t get_arc_reset_unit_refclk_low_offset() const = 0;
-    virtual uint32_t get_arc_reset_unit_refclk_high_offset() const = 0;
+    virtual uint32_t get_firmware_message_go_busy() const = 0;
+    virtual uint32_t get_firmware_message_go_idle() const = 0;
+    virtual uint64_t get_reset_unit_refclk_low_offset() const = 0;
+    virtual uint64_t get_reset_unit_refclk_high_offset() const = 0;
     virtual uint32_t get_dram_banks_number() const = 0;
-    virtual uint32_t get_aiclk_busy_val() const = 0;
+    // AICLK frequency the chip parks at when idle, in MHz.
+    virtual uint32_t get_min_clock_freq() const = 0;
     virtual uint32_t get_num_eth_channels() const = 0;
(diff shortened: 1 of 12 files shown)


device/api/umd/device/arch/architecture_implementation.hpp
```diff
@@ -29,17 +29,21 @@ class ArchitectureImplementation {
...
-    virtual uint32_t get_arc_reset_unit_refclk_high_offset() const = 0;
+    virtual uint32_t get_firmware_message_go_busy() const = 0;
+    virtual uint32_t get_firmware_message_go_idle() const = 0;
+    virtual uint64_t get_reset_unit_refclk_low_offset() const = 0;
+    virtual uint64_t get_reset_unit_refclk_high_offset() const = 0;
     virtual uint32_t get_dram_banks_number() const = 0;
-    virtual uint32_t get_aiclk_busy_val() const = 0;
+    // AICLK frequencies the chip parks at when idle and claims when busy, in MHz.
+    virtual uint32_t get_min_clock_freq() const = 0;
+    virtual uint32_t get_max_clock_freq() const = 0;
```
nbuncicTT: We have `get_max_clock_freq` in `FirmwareInfoProvider` which gives us the max clock freq from the device itself and it not a constant, maybe we should rename these two to: `get_min_clock_freq_limit` and `get_max_clock_freq_limit` use the `[[deprecated]]` attribute for `get_min_clock_freq` so that we have API parity.

nbuncicTT (approved):

device/tt_device/tt_device.cpp
```diff
@@ -606,11 +606,11 @@ uint64_t TTDevice::get_refclk_counter() {
...
     uint32_t low_addr = 0;
-    read_from_arc_apb(&high1_addr, architecture_impl_->get_arc_reset_unit_refclk_high_offset(), sizeof(high1_addr));
-    read_from_arc_apb(&low_addr, architecture_impl_->get_arc_reset_unit_refclk_low_offset(), sizeof(low_addr));
-    read_from_arc_apb(&high1_addr, architecture_impl_->get_arc_reset_unit_refclk_high_offset(), sizeof(high1_addr));
+    read_from_arc_apb(&high1_addr, architecture_impl_->get_reset_unit_refclk_high_offset(), sizeof(high1_addr));
+    read_from_arc_apb(&low_addr, architecture_impl_->get_reset_unit_refclk_low_offset(), sizeof(low_addr));
+    read_from_arc_apb(&high1_addr, architecture_impl_->get_reset_unit_refclk_high_offset(), sizeof(high1_addr));
     if (high2_addr > high1_addr) {
-        read_from_arc_apb(&low_addr, architecture_impl_->get_arc_reset_unit_refclk_low_offset(), sizeof(low_addr));
+        read_from_arc_apb(&low_addr, architecture_impl_->get_reset_unit_refclk_low_offset(), sizeof(low_addr));
```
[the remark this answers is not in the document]
broskoTT: This is pre-existing behavior, not something this PR changes. The three reads and the rollover check were already there, this PR only renamed the offset getter on those lines. Ignoring it here.

device/api/umd/device/arch/architecture_implementation.hpp
broskoTT: Dropped `get_max_clock_freq` from `ArchitectureImplementation` instead of renaming it. It had a single consumer, `test_firmware_info_provider`, which supplied the expected value for the telemetry reading, so it now resolves `AICLK_BUSY_VAL` for its own architecture and the interface carries only what the specification lists.

Worth noting the collision was wider than the two names: `TTDevice::get_max_clock_freq()` already exists and forwards to `FirmwareInfoProvider`, so `tt_device.cpp:284` reads `target_aiclk = get_max_clock_freq()` meaning telemetry. Removing the constant one takes that ambiguity away.

`get_min_clock_freq` still overlaps three ways though, the architecture constant, `TTDevice::get_min_clock_freq()` forwarding to it, and `FirmwareInfoProvider::get_min_clock_freq()` from telemetry. I kept the plain name because that is what the Base API specification lists, but say if you would rather have `get_min_clock_freq_limit` and I will rename it.

On the `[[deprecated]]` part, which symbol did you mean? If it is `TTDevice::get_min_clock_freq()`, it has four in tree callers (`local_chip`, `remote_chip` twice, `tt_device.cpp:287`) plus a nanobind binding and a test, so with `-Werror` the attribute breaks the build unless all of those move over first. Happy to do it, just want to confirm the scope before turning a rename PR into a `TTDevice` API change.

broskoTT: So the get_max_clock_freq was removed, ignore the rest

PR #3229, opened 2026-08-22
Probe locks in lock_virus through LockManager

broskoTT:
### Issue
#2745

### Description
lock_virus built the names of the locks it expected from LockManager's internal name map, and
probed them by opening the backing RobustMutex itself. That tied the tool to how locks happen to
be implemented, right before that implementation changes.

It now walks every mutex type LockManager exposes, for every PCIe and JTAG device found, and
probes each one through LockManager. Whatever ends up backing a lock, the tool reports the lock
UMD would actually take.

Output before, on a host with no card. What it lists is whatever is left in /dev/shm, here mostly
locks left behind by unit tests, and it has no idea whether UMD would ever take them:

```
=== UMD locks found in /dev/shm (5) ===
  [TT_UMD_LOCK.ARC_MSG                                    ]  FREE
  [TT_UMD_LOCK.CREATE_ETH_MAP                             ]  FREE
  [TT_UMD_LOCK.UMD_TEST_MUTEX_INTERFACE                   ]  FREE
  [TT_UMD_LOCK.UMD_TEST_ROBUST_MUTEX_EXCEPTION            ]  FREE
  [TT_UMD_LOCK.UMD_TEST_ROBUST_MUTEX_HANG                 ]  FREE

=== PCIe devices found (0) ===
  (none)

=== Expected locks missing from /dev/shm (0) ===
  (none)
```

Output after, on the same host. Each lock UMD can take is listed under the device it belongs to,
one block per device found:

```
=== System wide locks ===
    [ARC_MSG         ]  FREE
    [CREATE_ETH_MAP  ]  FREE

=== PCIe devices found (0) ===

=== JTAG devices found (0) ===
```

### List of the changes
- Add LockManager::probe_mutex, which reports a lock's owner without leaving it held
- Add to_string(MutexType)
- Rewrite lock_virus to enumerate mutex types and devices instead of scanning /dev/shm
- --hold-lock takes a mutex type, with --device / --device-type, instead of a lock name

### Testing
Ran lock_virus on a host with no card, and against a second instance holding ARC_MSG via
--hold-lock, which it reports as LOCKED with the holder's PID/TID.

### API Changes
This PR has API changes, adds LockManager::probe_mutex and to_string(MutexType).

### Full stack diff

### DIFF
--- a/device/utils/lock_manager.cpp
+++ b/device/utils/lock_manager.cpp
@@ -13,6 +13,8 @@
 
 namespace tt::umd {
 
+std::string to_string(MutexType mutex_type) { return LockManager::MUTEX_TYPE_TO_STRING.at(mutex_type); }
+
 void LockManager::initialize_mutex(MutexType mutex_type) {
     initialize_mutex_internal(MUTEX_TYPE_TO_STRING.at(mutex_type));
 }
@@ -58,6 +60,17 @@ std::unique_lock<RobustMutex> LockManager::acquire_mutex(
     return acquire_mutex_internal(mutex_name);
 }
 
+std::optional<std::pair<pid_t, pid_t>> LockManager::probe_mutex(MutexType mutex_type) {
+    return probe_mutex_internal(MUTEX_TYPE_TO_STRING.at(mutex_type));
+}
+
+std::optional<std::pair<pid_t, pid_t>> LockManager::probe_mutex(
+    MutexType mutex_type, int device_id, IODeviceType device_type) {
+    std::string mutex_name = MUTEX_TYPE_TO_STRING.at(mutex_type) + "_" + std::to_string(device_id) + "_" +
+                             DeviceTypeToString.at(device_type);
+    return probe_mutex_internal(mutex_name);
+}
+
 void LockManager::initialize_mutex_internal(const std::string& mutex_name) {
     if (mutexes.find(mutex_name) != mutexes.end()) {
         log_warning(LogUMD, "Mutex already initialized: {}", mutex_name);
@@ -82,6 +95,19 @@ std::unique_lock<RobustMutex> LockManager::acquire_mutex_internal(const std::str
         UMD_THROW(error::RuntimeError, "Mutex not initialized: " + mutex_name);
     }
     return std::unique_lock(mutexes.at(mutex_name));
+}
+
+std::optional<std::pair<pid_t, pid_t>> LockManager::probe_mutex_internal(const std::string& mutex_name) {
+    if (mutexes.find(mutex_name) == mutexes.end()) {
+        UMD_THROW(error::RuntimeError, "Mutex not initialized: " + mutex_name);
+    }
+    RobustMutex& mutex = mutexes.at(mutex_name);
+    std::optional<std::pair<pid_t, pid_t>> owner = mutex.probe_lock(std::chrono::seconds(0));
+    if (!owner.has_value()) {
+        // The mutex was free, which means probing it took it. Release it so that probing stays a query.
+        mutex.unlock();
+    }
+    return owner;
 }
 
 }  // namespace tt::umd
--- a/tools/lock_virus.cpp
+++ b/tools/lock_virus.cpp
@@ -2,87 +2,96 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-// lock_virus: inspect all UMD shared-memory locks present in /dev/shm, report
-// their state (free / held, owner PID/TID), and cross-check against the locks
-// expected for every enumerated PCIe device.
+// lock_virus: report the state (free / held, owner PID/TID) of every lock UMD can take - each mutex
+// type, for every PCIe and JTAG device found. Locks are probed through LockManager, so what gets
+// reported is the lock UMD would actually take, whichever mutex happens to back it.
 
-#include <dirent.h>
-
-#include <algorithm>
-#include <cerrno>
-#include <cstdlib>
-#include <cstring>
 #include <cxxopts.hpp>
 #include <iostream>
-#include <set>
+#include <memory>
+#include <optional>
 #include <string>
 #include <thread>
 #include <tt-logger/tt-logger.hpp>
+#include <utility>
 #include <vector>
 
+#include "umd/device/jtag/jtag_device.hpp"
 #include "umd/device/pcie/pci_device.hpp"
 #include "umd/device/utils/lock_manager.hpp"
-#include "umd/device/utils/robust_mutex.hpp"
 
 using namespace tt::umd;
 
-// ── lock-name helpers ─────────────────────────────────────────────────────────
-
-static std::vector<std::string> chip_specific_mutex_names(int device_id, const std::string& device_type = "PCIe") {
-    std::vector<std::string> names;
-    names.reserve(LockManager::CHIP_SPECIFIC_MUTEX_TYPES.size());
-    for (MutexType type : LockManager::CHIP_SPECIFIC_MUTEX_TYPES) {
-        std::string name = LockManager::MUTEX_TYPE_TO_STRING.at(type);
-        name.append("_").append(std::to_string(device_id)).append("_").append(device_type);
-        names.push_back(std::move(name));
-    }
-    return names;
-}
-
-// ── /dev/shm scan ─────────────────────────────────────────────────────────────
-
-static std::vector<std::string> scan_umd_locks() {
-    static constexpr std::string_view prefix = RobustMutex::SHM_FILE_PREFIX;
-
-    std::vector<std::string> found;
-    DIR* dir = opendir("/dev/shm");
-    if (!dir) {
-        log_warning(tt::LogUMD, "Could not open /dev/shm: {}", std::to_string(errno));
-        return found;
-    }
-
-    struct dirent* entry;
-    while ((entry = readdir(dir)) != nullptr) {
-        std::string name(entry->d_name);
-        if (name.rfind(std::string(prefix), 0) == 0) {
-            found.push_back(name);
-        }
-    }
-    closedir(dir);
-    std::sort(found.begin(), found.end());
-    return found;
-}
-
 // ── reporting ─────────────────────────────────────────────────────────────────
 
-static void report_lock(const std::string& shm_name) {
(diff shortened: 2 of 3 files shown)


device/utils/lock_manager.cpp
```diff
@@ -84,4 +97,17 @@ std::unique_lock<RobustMutex> LockManager::acquire_mutex_internal(const std::str
     return std::unique_lock(mutexes.at(mutex_name));
 }
 
+std::optional<std::pair<pid_t, pid_t>> LockManager::probe_mutex_internal(const std::string& mutex_name) {
+    if (mutexes.find(mutex_name) == mutexes.end()) {
+        UMD_THROW(error::RuntimeError, "Mutex not initialized: " + mutex_name);
+    }
+    RobustMutex& mutex = mutexes.at(mutex_name);
+    std::optional<std::pair<pid_t, pid_t>> owner = mutex.probe_lock(std::chrono::seconds(0));
```
pjanevskiTT, line 105: `probe_lock()` returns `nullopt` in two different situations: the mutex really was free, and `EOWNERDEAD`, where a previous owner crashed holding it and `probe_lock` recovers the mutex via `pthread_mutex_consistent()` before returning (`robust_mutex.cpp`, the `EOWNERDEAD` branch of `probe_lock`).

So `probe_mutex` reports FREE in both cases. Two consequences: `lock_virus` prints `FREE` for a lock that was orphaned by a crashed process right up until we probed it, which is exactly the state someone running this tool is hunting for; and a query API silently repairs cross-process lock state as a side effect.

The header comment ("Returns std::nullopt if the mutex was free") should say this, and it would be worth having `lock_virus` distinguish "free" from "recovered from a dead owner".

tools/lock_virus.cpp
```diff
@@ -2,87 +2,96 @@
...
+}
+
+// ── testing mode ──────────────────────────────────────────────────────────────
 
-    struct dirent* entry;
-    while ((entry = readdir(dir)) != nullptr) {
-        std::string name(entry->d_name);
-        if (name.rfind(std::string(prefix), 0) == 0) {
-            found.push_back(name);
+static std::optional<MutexType> find_mutex_type(const std::string& mutex_name) {
```
pjanevskiTT: `find_mutex_type` searches both `SYSTEM_WIDE_MUTEX_TYPES` and `CHIP_SPECIFIC_MUTEX_TYPES`, and between them they cover all seven enumerators, so every name resolves regardless of whether `--device` was given. `hold_lock` then picks system wide vs chip specific purely on the presence of `--device`.

That makes two silent no-ops possible. `--hold-lock NON_MMIO` without `--device` initializes and holds a system wide `NON_MMIO` lock that UMD never takes, and the reporting side only walks `SYSTEM_WIDE_MUTEX_TYPES` for that section, so it never prints it. `--hold-lock CREATE_ETH_MAP --device 0` holds `CREATE_ETH_MAP_0_PCIe`, also never reported. In both cases this instance prints "Holding lock" while a parallel `lock_virus` run shows everything FREE, which defeats the point of the testing mode.

Worth validating membership: with `--device`, require the type to be in `CHIP_SPECIFIC_MUTEX_TYPES`; without it, in `SYSTEM_WIDE_MUTEX_TYPES`.

tools/lock_virus.cpp
```diff
@@ -2,87 +2,96 @@
...
+    } else {
+        log_info(tt::LogUMD, "    [{:<16}]  FREE", mutex_name);
     }
-    return names;
 }
 
-// ── /dev/shm scan ─────────────────────────────────────────────────────────────
+static void report_device_locks(LockManager& lock_manager, int device_id, IODeviceType device_type) {
+    for (MutexType mutex_type : LockManager::CHIP_SPECIFIC_MUTEX_TYPES) {
+        lock_manager.initialize_mutex(mutex_type, device_id, device_type);
```
pjanevskiTT, line 53: `initialize_mutex` creates the backing `/dev/shm/TT_UMD_LOCK.*` file when it does not exist, so a single `lock_virus` run now materializes all six chip specific types for every PCIe and JTAG device found, plus the two system wide ones, and leaves them behind.

Permissions are fine (`open_shm_file` clears the umask), so nothing breaks. But this turns a read-only inspection tool into one that leaves state, and it means "which lock files exist" stops being a usable signal after the first run, which is what the removed *Expected locks missing from /dev/shm* section reported on. If that trade is intended, it is worth one line in the `--help` text so nobody is surprised by a freshly populated `/dev/shm`.

tools/lock_virus.cpp
```diff
@@ -110,85 +123,52 @@ int main(int argc, char* argv[]) {
...
-    if (result.count("hold-lock")) {
-        const std::string mutex_name = result["hold-lock"].as<std::string>();
-        RobustMutex m(mutex_name);
-        m.initialize();
-        if (auto owner = m.probe_lock()) {
-            log_error(
-                tt::LogUMD, "Lock '{}' is already held by PID={} TID={}", mutex_name, owner->first, owner->second);
+    try {
+        const std::string device_type_name = result["device-type"].as<std::string>();
+        if (device_type_name != "pcie" && device_type_name != "jtag") {
```
pjanevskiTT: maybe store these literals in constant somewhere. Also, specify in help whether we support just lowercase or any letters

tools/lock_virus.cpp
```diff
@@ -110,85 +123,52 @@ int main(int argc, char* argv[]) {
...
-        log_info(tt::LogUMD, "=== UMD locks found in /dev/shm ({}) ===", found_locks.size());
-        if (found_locks.empty()) {
-            log_info(tt::LogUMD, "  (none)");
-        } else {
-            for (const auto& name : found_locks) {
-                report_lock(name);
+        const IODeviceType device_type = device_type_name == "jtag" ? IODeviceType::JTAG : IODeviceType::PCIe;
+
+        // ── Testing mode: hold a single lock and spin ─────────────────────
+        if (result.count("hold-lock")) {
```
pjanevskiTT, line 168: same for this, let's maybe store all the literals as statics here in the tool since they repeat in lots of places

pjanevskiTT (approved):

device/utils/lock_manager.cpp
broskoTT: Right on both counts, and the doc was the misleading part - `RobustMutex::probe_lock` already spells
the recovery out, `probe_mutex` did not. It now says that not held covers a lock nobody had taken and
one whose owner died holding it, and that probing recovers rather than reports the latter.

Keeping the recovery: a robust mutex whose owner died is meant to be made consistent by the next
acquirer, and probing is an acquire. Reporting it as a distinct state would need `probe_lock` to
return more than an optional, which is a bigger change than this tool is worth.

tools/lock_virus.cpp
broskoTT: Good catch, and the "prints Holding lock while a parallel run shows everything FREE" framing is
exactly what made it worth fixing rather than documenting.

`find_mutex_type` now takes the list to search, and `--device` picks which: chip specific with it,
system wide without. Both cases you name fail loudly now, and the error lists what is valid:

```
'NON_MMIO' is not a system wide mutex type. Expected one of: ARC_MSG, CREATE_ETH_MAP
'CREATE_ETH_MAP' is not a chip specific mutex type. Expected one of: ARC_MSG, REMOTE_ARC_MSG, ...
```

The lists come from `LockManager`, so a new mutex type cannot fall out of sync with the tool.

tools/lock_virus.cpp
```diff
@@ -110,85 +123,52 @@ int main(int argc, char* argv[]) {
...
-            log_info(tt::LogUMD, "  (none)");
-        } else {
-            for (int id : device_ids) {
-                log_info(tt::LogUMD, "  device {}", id);
+            std::optional<int> device_id;
+            if (result.count("device")) {
+                device_id = result["device"].as<int>();
             }
+            hold_lock(*mutex_type, device_id, device_type);
         }
```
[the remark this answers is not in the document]
broskoTT: Same as the thread above, fixed there - the type is now resolved only in the form `--device` asks
for, and the error names the valid types.

tools/lock_virus.cpp
broskoTT: Intended, and you are right that it was nowhere the user would see it. It is in the `--help` text now,
and in the tool's comment along with why it probes every lock instead of only the ones that exist: the
job is finding out who holds a lock when something is stuck, including when nobody knows what the
holder is, so a lock that was never taken is still worth reporting as free.

tools/lock_virus.cpp
broskoTT: Done - `DEVICE_TYPE_PCIE` and `DEVICE_TYPE_JTAG`, used by the check and the message, and `--help` now
says lowercase. The option's `default_value` still takes the literal, since cxxopts wants a
`std::string` there.

tools/lock_virus.cpp
broskoTT: Covered by the same change.

tools/lock_virus.cpp
```diff
@@ -2,87 +2,96 @@
...
-
-#include <algorithm>
-#include <cerrno>
-#include <cstdlib>
-#include <cstring>
 #include <cxxopts.hpp>
 #include <iostream>
-#include <set>
+#include <memory>
+#include <optional>
```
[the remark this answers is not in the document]
broskoTT: Correct, it was working only through `<thread>`. Added.

tools/lock_virus.cpp
```diff
@@ -2,87 +2,96 @@
...
-        return found;
+// Returns the JTAG device indices, or an empty list if JTAG is not usable on this host. The J-Link
+// library it needs is loaded at runtime and is absent on most systems.
+static std::vector<int> enumerate_jtag_devices() {
+    try {
+        std::unique_ptr<JtagDevice> jtag_device = JtagDevice::create();
+        std::vector<int> device_ids;
+        device_ids.reserve(jtag_device->get_device_cnt());
+        for (uint32_t index = 0; index < jtag_device->get_device_cnt(); index++) {
+            device_ids.push_back(static_cast<int>(index));
```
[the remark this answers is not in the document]
broskoTT: Correct, and it can reach the J-Link library. Read once into a local now.

PR #3230, opened 2026-08-22
Introduce MutexInterface

broskoTT:
### Issue
#2745

### Description
RobustMutex and KmdMutex already expose the same operations, but nothing states that, so whoever
holds a lock has to know which of the two backs it. MutexInterface states it, and both implement it.

LockManager keeps and hands out its mutexes through the interface, so which backend guards a lock
stops being visible to its callers. It still only ever creates RobustMutex; choosing between the
two comes later in the stack.

### List of the changes
- Add MutexInterface with initialize / lock / unlock / probe_lock
- RobustMutex and KmdMutex implement it
- LockManager holds std::unique_ptr<MutexInterface> and hands out std::unique_lock<MutexInterface>

### Testing
No behavior change, the one backend in use is the same one as before. umd_misc_tests, plus a
lock_virus run, which goes through LockManager and so through the interface.

### API Changes
This PR has API changes, adds MutexInterface and acquire_mutex now returns
std::unique_lock<MutexInterface>.

### Full stack diff

### DIFF
--- a/device/api/umd/device/utils/kmd_mutex.hpp
+++ b/device/api/umd/device/utils/kmd_mutex.hpp
@@ -86,7 +86,7 @@ public:
     // Returns std::nullopt if the lock was acquired (the caller now holds it). On contention returns a
     // {pid, tid} pair, but KMD does not expose the owner, so it is always {0, 0} here - the pair's
     // presence signals "held by someone else", nothing more.
-    std::optional<std::pair<pid_t, pid_t>> probe_lock(std::chrono::seconds timeout = std::chrono::seconds(0)) override;
+    std::optional<std::pair<pid_t, pid_t>> probe_lock(std::chrono::seconds timeout) override;
 
     // Releases the lock. It is also released automatically when this object (and thus its handle) is
     // destroyed, or when the owning process exits.
--- a/device/api/umd/device/utils/kmd_mutex.hpp
+++ b/device/api/umd/device/utils/kmd_mutex.hpp
@@ -12,6 +12,8 @@
 #include <string>
 #include <utility>
 
+#include "umd/device/utils/mutex_interface.hpp"
+
 // tt-kmd-lib device handle. Forward declared so this public header does not require tt-kmd-lib's
 // include path, which UMD links privately.
 struct tt_device_t;
@@ -53,7 +55,7 @@ namespace tt::umd {
 //
 // It meets the C++ Lockable requirement (lock()/try_lock()/unlock()), so it can be used with RAII
 // helpers like std::lock_guard and std::unique_lock.
-class KmdMutex {
+class KmdMutex : public MutexInterface {
 public:
     // @param pci_device_num  N in /dev/tenstorrent/N (a UMD logical device id / PCIDevice number).
     // @param lock_index      KMD resource lock index in [0, TT_RESOURCE_LOCK_COUNT). See
@@ -64,7 +66,7 @@ public:
     // Opens the dedicated device handle used to hold the lock. Must be called before
     // lock()/try_lock(). Kept separate from the constructor so that failures during setup are still
     // cleaned up by the destructor. Calling it again once the handle is open does nothing.
-    void initialize();
+    void initialize() override;
 
     // Move-only so it can live in STL containers. Copying would alias the owned handle.
     KmdMutex(KmdMutex&& other) noexcept;
@@ -74,7 +76,7 @@ public:
 
     // Blocks until the lock is acquired. If a reset is detected while waiting (the handle becomes
     // unusable), the handle is transparently reopened and the wait is retried.
-    void lock();
+    void lock() override;
 
     // Attempts to acquire without blocking. Returns true if acquired, false if held by another
     // handle.
@@ -84,11 +86,11 @@ public:
     // Returns std::nullopt if the lock was acquired (the caller now holds it). On contention returns a
     // {pid, tid} pair, but KMD does not expose the owner, so it is always {0, 0} here - the pair's
     // presence signals "held by someone else", nothing more.
-    std::optional<std::pair<pid_t, pid_t>> probe_lock(std::chrono::seconds timeout = std::chrono::seconds(0));
+    std::optional<std::pair<pid_t, pid_t>> probe_lock(std::chrono::seconds timeout = std::chrono::seconds(0)) override;
 
     // Releases the lock. It is also released automatically when this object (and thus its handle) is
     // destroyed, or when the owning process exits.
-    void unlock();
+    void unlock() override;
 
     // Best-effort query of whether the lock is currently held by any handle (including this one).
     // This is inherently racy and intended for debugging/diagnostics only.
--- a/device/api/umd/device/utils/lock_manager.hpp
+++ b/device/api/umd/device/utils/lock_manager.hpp
@@ -6,6 +6,7 @@
 
 #include <sys/types.h>
 
+#include <memory>
 #include <mutex>
 #include <optional>
 #include <string>
@@ -14,7 +15,7 @@
 #include <vector>
 
 #include "umd/device/types/communication_protocol.hpp"
-#include "umd/device/utils/robust_mutex.hpp"
+#include "umd/device/utils/mutex_interface.hpp"
 
 namespace tt::umd {
 
@@ -38,8 +39,8 @@ enum class MutexType {
 // Name of a mutex type, for logging and diagnostics.
 std::string to_string(MutexType mutex_type);
 
-// Note that the returned std::unique_lock<RobustMutex> should never outlive the LockManager which holds underlying
-// RobustMutexes. Also note that clear_mutex doesn't need to be explicitly called, since the mutexes will all get
+// Note that the returned std::unique_lock<MutexInterface> should never outlive the LockManager which holds
+// underlying mutexes. Also note that clear_mutex doesn't need to be explicitly called, since the mutexes will all get
 // cleared automatically when the LockManager goes out of scope. We could implement these lock such that initialization
 // is not needed, and they are initialized every time they're locked, but since that communicates with the OS filesystem
 // it might be slower do to it each time. This way, locking/unlocking should be faster.
@@ -75,19 +76,19 @@ public:
     // This set of functions is used to manage mutexes which are system wide and not chip specific.
     void initialize_mutex(MutexType mutex_type);
     void clear_mutex(MutexType mutex_type);
-    std::unique_lock<RobustMutex> acquire_mutex(MutexType mutex_type);
+    std::unique_lock<MutexInterface> acquire_mutex(MutexType mutex_type);
 
     // This set of functions is used to manage mutexes which are chip specific.
     void initialize_mutex(MutexType mutex_type, int device_id, IODeviceType device_type = IODeviceType::PCIe);
     void clear_mutex(MutexType mutex_type, int device_id, IODeviceType device_type = IODeviceType::PCIe);
-    std::unique_lock<RobustMutex> acquire_mutex(
+    std::unique_lock<MutexInterface> acquire_mutex(
         MutexType mutex_type, int device_id, IODeviceType device_type = IODeviceType::PCIe);
 
     // This set of functions is used to manage mutexes which are chip specific. This variant accepts custom mutex name.
     void initialize_mutex(
         const std::string& mutex_prefix, int device_id, IODeviceType device_type = IODeviceType::PCIe);
     void clear_mutex(const std::string& mutex_prefix, int device_id, IODeviceType device_type = IODeviceType::PCIe);
-    std::unique_lock<RobustMutex> acquire_mutex(
+    std::unique_lock<MutexInterface> acquire_mutex(
         const std::string& mutex_prefix, int device_id, IODeviceType device_type = IODeviceType::PCIe);
 
     // Reports whether a mutex is currently held, without holding it afterwards. Returns the owning {pid, tid} if it is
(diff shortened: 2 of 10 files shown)


device/api/umd/device/utils/kmd_mutex.hpp
pjanevskiTT, line 1: can we rebase this properly?

device/utils/kmd_mutex.cpp
```diff
@@ -0,0 +1,177 @@
...
+    // O_APPEND tells KMD (>= 2.6) not to request high power on open; a lock-only handle should not
+    // change the device's power state. The flag is harmless on older KMD versions.
+    int result = tt_device_open(device_path_.c_str(), &device_, O_APPEND);
+    UMD_ASSERT(
+        result == 0,
+        error::RuntimeError,
+        fmt::format("tt_device_open() failed for KMD lock device {} errno: {}", device_path_, -result));
+}
+
+void KmdMutex::initialize() { open_device(); }
```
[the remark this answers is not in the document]
broskoTT: Wrongly rebased

pjanevskiTT (approved):

device/api/umd/device/utils/mutex_interface.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+#include <sys/types.h>  // pid_t
+
+#include <chrono>
+#include <optional>
+#include <utility>
+
+namespace tt::umd {
+
+// Common interface of UMD's cross-process locking backends, so that a lock can be guarded by whichever
```
pjanevskiTT, line 15: The interface says a lock can be guarded by "whichever backend suits it without its users having to know which one", but for the four operations declared here the two implementations differ observably, and nothing states which of that behavior is contractual:

- **re-`lock()` by the owning thread:** `RobustMutex` surfaces it as a throw - glibc's robust mutex returns `EDEADLK`, which falls into `probe_lock`'s `else` branch (`robust_mutex.cpp:458-465`), reached from `lock()`'s initial `probe_lock(1s)` at `robust_mutex.cpp:470`. `KmdMutex::lock()` polls `try_lock()` in an unbounded loop (`kmd_mutex.cpp:134`), and per `kmd_mutex.hpp:39-42` an acquire fails if the lock is held *including by the handle asking for it* - so the same code hangs instead.
- **`unlock()` by a non-owner / double unlock:** `RobustMutex` throws on the non-zero `pthread_mutex_unlock` return (`robust_mutex.cpp:418-423`); `KmdMutex` logs at debug and returns (`kmd_mutex.cpp:194-198`), and `kmd_mutex.hpp:43-45` notes releasing is not owner-checked at all.
- **TSAN:** `RobustMutex` is annotated, `KmdMutex` is not.

Not a bug in this PR - `initialize_mutex_internal` only ever constructs a `RobustMutex`, and `KmdMutex` is not constructed anywhere under `device/`. But this comment is now the only place a caller can learn what it may rely on, since the point of the interface is that callers stop knowing their backend. The re-entrancy row is the sharp edge: code that gets a diagnosable exception today would silently hang the first time a lock is flipped to the KMD backend, which is what the rest of the stack sets up.

A few lines here would pin it down - that the mutex is non-recursive, and that whether misuse (re-lock, non-owner unlock, double unlock) is *diagnosed* is unspecified per backend.

Same theme, one line up the file: `lock_manager.hpp:94-96` documents `probe_mutex` as returning "the owning `{pid, tid}`" with no mention of the `{0, 0}` case that lines 31-32 here reserve.

device/api/umd/device/utils/robust_mutex.hpp
```diff
@@ -63,7 +65,7 @@ class RobustMutex {
     // Note: the returned PID/TID pair is a snapshot that may already be stale, since there is a race condition inherent
     // in trying to inspect a lock without acquiring it. So consider the information best-effort and for debugging
     // purposes only.
-    std::optional<std::pair<pid_t, pid_t>> probe_lock(std::chrono::seconds timeout = std::chrono::seconds(0));
+    std::optional<std::pair<pid_t, pid_t>> probe_lock(std::chrono::seconds timeout = std::chrono::seconds(0)) override;
```
pjanevskiTT: Default argument on a virtual. `MutexInterface::probe_lock` (`mutex_interface.hpp:33`) declares no default, while both overrides do - here and `kmd_mutex.hpp:89`. Default arguments bind statically, so the same call is well-formed or not depending on the static type: `probe_lock()` compiles through a `RobustMutex&` and is a hard error through a `MutexInterface&`. That asymmetry is newly reachable now that `LockManager` hands out the base type.

Cheapest fix is dropping the defaults from both overrides so the timeout is always explicit - `tests/microbenchmark/host_benchmark/test_locks.cpp:185` and `:195` are the only two call sites relying on it.

device/api/umd/device/utils/mutex_interface.hpp
```diff
@@ -0,0 +1,36 @@
...
+// backend suits it without its users having to know which one. See device/utils/README.md for how the
+// backends compare.
+// Implementations meet the C++ BasicLockable requirement, so they work with std::lock_guard and
+// std::unique_lock.
+class MutexInterface {
+public:
+    virtual ~MutexInterface() = default;
```
pjanevskiTT, line 31: `MutexInterface` is an abstract base with a public implicitly-declared copy constructor and copy assignment, so `a = b;` on two `MutexInterface&` compiles as a silent slicing no-op. Both derived classes already `= delete` copy, but that doesn't cover assignment made through a base reference.

Suggest deleting them alongside the destructor (or making them protected):

```cpp
MutexInterface(const MutexInterface&) = delete;
MutexInterface& operator=(const MutexInterface&) = delete;
```

Safe with respect to the existing moves: `RobustMutex` and `KmdMutex` both hand-write their move constructor/assignment and neither touches the stateless base subobject.

device/api/umd/device/utils/lock_manager.hpp
```diff
@@ -75,19 +76,19 @@ class LockManager {
...
-    std::unique_lock<RobustMutex> acquire_mutex(
+    std::unique_lock<MutexInterface> acquire_mutex(
         MutexType mutex_type, int device_id, IODeviceType device_type = IODeviceType::PCIe);
 
     // This set of functions is used to manage mutexes which are chip specific. This variant accepts custom mutex name.
     void initialize_mutex(
         const std::string& mutex_prefix, int device_id, IODeviceType device_type = IODeviceType::PCIe);
     void clear_mutex(const std::string& mutex_prefix, int device_id, IODeviceType device_type = IODeviceType::PCIe);
-    std::unique_lock<RobustMutex> acquire_mutex(
+    std::unique_lock<MutexInterface> acquire_mutex(
```
pjanevskiTT, line 91: *(Note: this belongs on `tests/misc/test_robust_mutex.cpp`, which isn't in this PR's diff - placing it on the change that invalidates it.)*

Three comments in `tests/misc/test_robust_mutex.cpp` become false with this return-type change:

- `:35` - "because LockManager hands out a `std::unique_lock<RobustMutex>` whose destructor runs during stack unwinding"
- `:58-60` - "An exception thrown while a `std::unique_lock<RobustMutex>` is in scope ... This is exactly the type LockManager returns"
- `:67` - "Same RAII wrapper `LockManager::acquire_mutex()` returns to callers like `read_non_mmio()`"

The test's coverage is unaffected - it uses the concrete type deliberately - but those comments are load-bearing: they're the stated argument for why the test reproduces production behavior. It's `std::unique_lock<MutexInterface>` now.

device/api/umd/device/utils/kmd_mutex.hpp
broskoTT: Done. #3182 has since merged, so this branch now sits on main and the diff is the interface and its
two implementations, nothing else.

device/api/umd/device/utils/mutex_interface.hpp
broskoTT: This is the one worth having caught, and the framing is right: the interface exists so callers stop
knowing their backend, which makes this comment the only place they can learn what to rely on. The
re-entrancy row in particular - an exception today, a silent hang once a lock moves to KMD - is not
something anyone would find out from the signatures.

The interface now states that the mutex is not recursive, and that misuse - re-lock, non-owner unlock,
double unlock - is not required to be diagnosed, so callers cannot depend on any particular outcome.
Written as what an implementation owes its callers rather than as what today's two happen to do, so it
stays true of the next one. `probe_mutex` picks up the `{0, 0}` case too.

device/api/umd/device/utils/robust_mutex.hpp
broskoTT: Correct, and it is exactly the kind of thing that only bites once callers hold the base type. Dropped
from both overrides, so the timeout is always explicit. The two call sites you found in
`test_locks.cpp` were indeed the only ones.

device/api/umd/device/utils/mutex_interface.hpp
broskoTT: Right, and the point that the derived `= delete`s do not cover assignment through a base reference is
the part that is easy to miss. Both deleted on the interface. A default constructor had to be declared
alongside, since declaring any constructor suppresses the implicit one and both backends default
construct the base subobject in their hand written moves.

device/api/umd/device/utils/lock_manager.hpp
broskoTT: Fixed all three. Also said explicitly in the one at the lock site that the test uses the concrete type
because it is about RobustMutex rather than about which backend is underneath - which is the thing the
comments were implying and no longer quite true.

Placing it on the change that invalidates it was the right call, thanks.

device/utils/README.md
```diff
@@ -0,0 +1,65 @@
+# UMD locking backends
+
+UMD provides two cross-process lock implementations. They expose the same small interface
+(`initialize()`, `lock()`, `try_lock()`, `unlock()`, `probe_lock()`) and meet the C++ `Lockable`
+requirement, so they work with `std::lock_guard` / `std::unique_lock` and can be swapped for one
+another. They differ in *how* processes find each other and in the performance/robustness trade-off.
```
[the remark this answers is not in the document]
broskoTT: Correct - neither `RobustMutex` nor `MutexInterface` has `try_lock`, so it is `BasicLockable`.
Reworded, and the README now points at `MutexInterface` for the shared set and notes that `KmdMutex`
adds `try_lock` on top.

tools/lock_virus.cpp
```diff
@@ -2,87 +2,96 @@
...
-    for (MutexType type : LockManager::CHIP_SPECIFIC_MUTEX_TYPES) {
-        std::string name = LockManager::MUTEX_TYPE_TO_STRING.at(type);
-        name.append("_").append(std::to_string(device_id)).append("_").append(device_type);
-        names.push_back(std::move(name));
+static void report_state(const std::string& mutex_name, const std::optional<std::pair<pid_t, pid_t>>& owner) {
+    if (owner.has_value()) {
+        log_info(tt::LogUMD, "    [{:<16}]  LOCKED  PID={} TID={}", mutex_name, owner->first, owner->second);
+    } else {
+        log_info(tt::LogUMD, "    [{:<16}]  FREE", mutex_name);
     }
```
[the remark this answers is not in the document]
broskoTT: Real, but not reachable yet: `LockManager` only ever constructs a `RobustMutex` until the PR that
routes chip specific PCIe locks to KMD. Fixing it there, where the sentinel first becomes reachable
and where the tool's output actually changes.

device/utils/kmd_mutex.cpp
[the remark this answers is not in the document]
broskoTT: All three were fixed in #3182, which has since merged - the branch was showing them because it was
mis-rebased at the time. `initialize()` returns when the handle is open, `unlock()` closes and reopens
on `-ENODEV` through a non-throwing path, and `tests/misc/test_kmd_mutex.cpp` covers acquisition,
contention between handles and release.

device/utils/kmd_mutex.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+void KmdMutex::unlock() {
+    UMD_ASSERT(device_ != nullptr, error::RuntimeError, "KmdMutex::unlock() called before initialize()");
+
+    int was_held = 0;
+    int result = tt_lock_release(device_, lock_index_, &was_held);
+
+    // If the handle died due to a reset, the lock is still held in KMD until the handle is closed;
+    // the destructor's close will release it, so this is not fatal.
+    if (result == -ENODEV) {
```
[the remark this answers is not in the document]
broskoTT: All three were fixed in #3182, which has since merged - the branch was showing them because it was
mis-rebased at the time. `initialize()` returns when the handle is open, `unlock()` closes and reopens
on `-ENODEV` through a non-throwing path, and `tests/misc/test_kmd_mutex.cpp` covers acquisition,
contention between handles and release.

device/api/umd/device/utils/kmd_mutex.hpp
```diff
@@ -0,0 +1,126 @@
...
+// This class owns its own dedicated device handle (chardev fd), separate from any fd used to run a
+// workload. That is deliberate: a device reset invalidates every fd that was open across it (further
+// ioctls return ENODEV), but the lock itself survives the reset and stays held until its fd is
+// closed. Keeping the lock on a dedicated handle lets a caller hold the lock, reset the device, then
+// open fresh fds to run a workload. lock() transparently reopens its handle and retries if it
+// observes that a reset happened while it was waiting.
+//
+// It meets the C++ Lockable requirement (lock()/try_lock()/unlock()), so it can be used with RAII
+// helpers like std::lock_guard and std::unique_lock.
+class KmdMutex : public MutexInterface {
```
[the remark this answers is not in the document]
broskoTT: All three were fixed in #3182, which has since merged - the branch was showing them because it was
mis-rebased at the time. `initialize()` returns when the handle is open, `unlock()` closes and reopens
on `-ENODEV` through a non-throwing path, and `tests/misc/test_kmd_mutex.cpp` covers acquisition,
contention between handles and release.

tools/lock_virus.cpp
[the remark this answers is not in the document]
broskoTT: Done in #3232, where the sentinel first becomes reachable: a lock whose backend cannot name the holder now reports "LOCKED by an unknown holder" instead of PID=0 TID=0.

PR #3231, opened 2026-08-22
Address locks only by mutex type

broskoTT:
### Issue
#2745

### Description
Minor cleanup of LockManager, mostly removing unused members from public API.

### List of the changes
- Move MUTEX_TYPE_TO_STRING to lock_manager.cpp
- Drop the initialize / clear / acquire_mutex variants taking a mutex name
- Drop the defaulted device_type on the chip specific entry points, and take the device the lock
  belongs to from TTDevice at every site. Several callers were relying on the default and so named
  their lock after a PCIe device even on JTAG or simulation, and BlackholeArcMessenger dereferenced
  a null PCIDevice on a JTAG device

### Testing
umd_misc_tests, plus a lock_virus run.

Lock names change for JTAG and simulation: NON_MMIO_<id>_PCIe becomes NON_MMIO_<id>_JTAG, or
NON_MMIO_-1_Undefined for simulation. Initialization and both acquire sites move together, so nothing
breaks in process. Two UMD versions would stop serializing against each other on such a device, which
is the concern that holds back the last PR of this stack, but JTAG and simulation are bring-up paths
where a second process on a different UMD version does not turn up. PCIe lock names are unchanged.

### API Changes
This PR has API changes, removes the name taking lock entry points from LockManager and makes
device_type a required argument on the chip specific ones.

### Full stack diff

### DIFF
--- a/device/api/umd/device/utils/lock_manager.hpp
+++ b/device/api/umd/device/utils/lock_manager.hpp
@@ -87,7 +87,8 @@ private:
     std::optional<std::pair<pid_t, pid_t>> probe_mutex_internal(const std::string& mutex_name);
 
     // Maps from mutex name to an initialized mutex.
-    // Mutex names are made from the mutex type name, combined with device number for chip specific ones.
+    // Mutex names are made from the mutex type name, combined with the device number and device type for chip
+    // specific ones.
     // Note that once LockManager is out of scope, all the mutexes will be cleared up automatically.
     std::unordered_map<std::string, std::unique_ptr<MutexInterface>> mutexes;
 };
--- a/device/api/umd/device/utils/lock_manager.hpp
+++ b/device/api/umd/device/utils/lock_manager.hpp
@@ -46,17 +46,6 @@ std::string to_string(MutexType mutex_type);
 // it might be slower do to it each time. This way, locking/unlocking should be faster.
 class LockManager {
 public:
-    // Maps MutexType enum values to their string names used in shared-memory lock names.
-    inline static const std::unordered_map<MutexType, std::string> MUTEX_TYPE_TO_STRING = {
-        {MutexType::ARC_MSG, "ARC_MSG"},
-        {MutexType::REMOTE_ARC_MSG, "REMOTE_ARC_MSG"},
-        {MutexType::NON_MMIO, "NON_MMIO"},
-        {MutexType::MEM_BARRIER, "MEM_BARRIER"},
-        {MutexType::CREATE_ETH_MAP, "CREATE_ETH_MAP"},
-        {MutexType::CHIP_IN_USE, "CHIP_IN_USE"},
-        {MutexType::PCIE_DMA, "PCIE_DMA"},
-    };
-
     // Mutex types that are initialized per chip (combined with device_id + device_type).
     inline static const std::vector<MutexType> CHIP_SPECIFIC_MUTEX_TYPES = {
         MutexType::ARC_MSG,
@@ -79,17 +68,9 @@ public:
     std::unique_lock<MutexInterface> acquire_mutex(MutexType mutex_type);
 
     // This set of functions is used to manage mutexes which are chip specific.
-    void initialize_mutex(MutexType mutex_type, int device_id, IODeviceType device_type = IODeviceType::PCIe);
-    void clear_mutex(MutexType mutex_type, int device_id, IODeviceType device_type = IODeviceType::PCIe);
-    std::unique_lock<MutexInterface> acquire_mutex(
-        MutexType mutex_type, int device_id, IODeviceType device_type = IODeviceType::PCIe);
-
-    // This set of functions is used to manage mutexes which are chip specific. This variant accepts custom mutex name.
-    void initialize_mutex(
-        const std::string& mutex_prefix, int device_id, IODeviceType device_type = IODeviceType::PCIe);
-    void clear_mutex(const std::string& mutex_prefix, int device_id, IODeviceType device_type = IODeviceType::PCIe);
-    std::unique_lock<MutexInterface> acquire_mutex(
-        const std::string& mutex_prefix, int device_id, IODeviceType device_type = IODeviceType::PCIe);
+    void initialize_mutex(MutexType mutex_type, int device_id, IODeviceType device_type);
+    void clear_mutex(MutexType mutex_type, int device_id, IODeviceType device_type);
+    std::unique_lock<MutexInterface> acquire_mutex(MutexType mutex_type, int device_id, IODeviceType device_type);
 
     // Reports whether a mutex is currently held, without holding it afterwards. Returns the owning {pid, tid} if it is
     // held, and std::nullopt if it is not - which covers both a mutex nobody had taken and one whose owner died holding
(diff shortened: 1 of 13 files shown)


device/chip/local_chip.cpp
```diff
@@ -97,14 +97,14 @@ void LocalChip::initialize_default_chip_mutexes() {
     // Initialize non-MMIO mutexes for WH devices regardless of number of chips, since these may be used for
     // ethernet broadcast
     if (tt_device_->get_arch() == tt::ARCH::WORMHOLE_B0) {
-        lock_manager_.initialize_mutex(MutexType::REMOTE_ARC_MSG, pci_device_id);
+        lock_manager_.initialize_mutex(MutexType::REMOTE_ARC_MSG, pci_device_id, IODeviceType::PCIe);
```
pjanevskiTT: maybe the proper thing to do here would be to get device type from TTDevice?

pjanevskiTT: apply this to all places where we have hardcoded PCIdevice in this call in the PR

pjanevskiTT (approved):

device/arc/blackhole_arc_messenger.cpp
```diff
@@ -26,7 +26,8 @@ uint32_t BlackholeArcMessenger::send_message(
     std::vector<uint32_t>& return_values,
     const std::vector<uint32_t>& args,
     const std::chrono::milliseconds timeout_ms) {
-    auto lock = lock_manager.acquire_mutex(MutexType::ARC_MSG, tt_device->get_pci_device()->get_device_num());
+    auto lock = lock_manager.acquire_mutex(
+        MutexType::ARC_MSG, tt_device->get_pci_device()->get_device_num(), IODeviceType::PCIe);
```
pjanevskiTT: This is the one caller left assuming PCIe, and it is the one where the assumption does not hold.

The mutex acquired here is initialized in `ArcMessenger`'s constructor as `ARC_MSG_<get_communication_device_id()>_<get_communication_device_type()>` (`arc_messenger.cpp:37`), and `WormholeArcMessenger` acquires it with those same two accessors (`wormhole_arc_messenger.cpp:75`). Keying it off `get_pci_device()->get_device_num()` plus a literal `IODeviceType::PCIe` only agrees with that on a PCIe backed device.

The JTAG path is reachable: `TTDevice::init_tt_device` creates an `ArcMessenger` unconditionally (`tt_device.cpp:115`), and `BlackholeTTDeviceModel(std::unique_ptr<JtagDevice>, uint8_t jlink_id)` leaves `pci_device_` null while setting `communication_device_id_ = jlink_id`. So on a JTAG Blackhole `get_pci_device()->get_device_num()` dereferences null, and had it survived, `ARC_MSG_<n>_PCIe` would not match the `ARC_MSG_<jlink_id>_JTAG` that was initialized, so `acquire_mutex_internal` would throw "Mutex not initialized".

This is pre-existing, since the old defaulted call resolved to PCIe the same way, but the PR is touching this line specifically to remove that assumption. Using `tt_device->get_communication_device_id()` and `tt_device->get_communication_device_type()` here matches the Wormhole path and fixes both.

device/tt_device/remote_communication.cpp
```diff
@@ -20,7 +20,10 @@ namespace tt::umd {
 
 RemoteCommunication::RemoteCommunication(TTDevice* local_tt_device, SysmemManager* sysmem_manager) :
     local_tt_device_(local_tt_device), sysmem_manager_(sysmem_manager) {
-    lock_manager_.initialize_mutex(MutexType::NON_MMIO, local_tt_device->get_communication_device_id());
+    lock_manager_.initialize_mutex(
```
pjanevskiTT, line 23: This changes the lock name for JTAG and simulated devices: `NON_MMIO_<id>_PCIe` becomes `NON_MMIO_<id>_JTAG`, or `NON_MMIO_-1_Undefined` for simulation, since `SimulationTTDeviceModel::get_communication_device_id()` returns -1 and the type resolves to `UNDEFINED`. Initialization and both acquire sites move together, so nothing breaks in process.

The question is cross version: two processes on different UMD versions stop serializing against each other on the same device. That is exactly what the deprecated global `ARC_MSG` lock at `wormhole_arc_messenger.cpp:84` exists for, and the reason #3233 is held until #3232 is deployed everywhere. If JTAG and simulation never see concurrent UMD versions in practice then this is a non issue, but it would be worth saying so in the description rather than leaving it implicit.

Minor, related: the Testing section says "No behavior change", which the third bullet of the description contradicts. This is an intended behavior change, not a no-op.

device/arc/blackhole_arc_messenger.cpp
broskoTT: Checked the whole chain and it holds up. `get_pci_device()` returns nullptr rather than throwing when
the capability is absent, deliberately (`tt_device.cpp:201`), `create_arc_messenger` runs
unconditionally from `init_tt_device`, and Blackhole has live JTAG paths - so this is a null deref on
a JTAG Blackhole, and the name would not have matched even if it had survived.

Fixed as you suggest, with `get_communication_device_id()` and `get_communication_device_type()`,
which is what the constructor initialized the lock under and what the Wormhole path already uses.

device/chip/local_chip.cpp
broskoTT: Agreed, done everywhere: the three `initialize_mutex` calls, the `CHIP_IN_USE` acquire in
`start_device`, the `MEM_BARRIER` acquire in `insert_host_to_device_barrier`, and the ARC one above.
No lock site dereferences `get_pci_device()` any more.

For PCIe this is the same value - `communication_device_id_` is initialized from
`pci_device->get_device_num()` - so nothing changes there.

One thing this surfaces rather than fixes: `initialize_default_chip_mutexes()` only runs when
`tlb_manager_` is set, which is PCIe only, so a JTAG LocalChip never initializes `MEM_BARRIER` and
`insert_host_to_device_barrier` now throws "Mutex not initialized" where it used to deref null. Both
are broken, this just makes the failure legible. Initializing chip mutexes for JTAG too is a bigger
change than this PR, so I left it - happy to open an issue.

broskoTT: Agreed, done everywhere: the three `initialize_mutex` calls, the `CHIP_IN_USE` acquire in
`start_device`, the `MEM_BARRIER` acquire in `insert_host_to_device_barrier`, and the ARC one above.
No lock site dereferences `get_pci_device()` any more.

For PCIe this is the same value - `communication_device_id_` is initialized from
`pci_device->get_device_num()` - so nothing changes there.

One thing this surfaces rather than fixes: `initialize_default_chip_mutexes()` only runs when
`tlb_manager_` is set, which is PCIe only, so a JTAG LocalChip never initializes `MEM_BARRIER` and
`insert_host_to_device_barrier` now throws "Mutex not initialized" where it used to deref null. Both
are broken, this just makes the failure legible. Initializing chip mutexes for JTAG too is a bigger
change than this PR, so I left it - happy to open an issue.

device/tt_device/remote_communication.cpp
broskoTT: You are right that the Testing section contradicts the description, fixed - this is an intended
behavior change, not a no-op.

On the cross version question: the concern that gates #3233 is two UMD versions driving the same
silicon device at once. JTAG and simulation are bring-up and development paths, where that does not
happen - a JTAG debug session or a simulator is not something a second process joins on a different
UMD version. Said so in the description rather than leaving it implicit.

device/api/umd/device/utils/lock_manager.hpp
```diff
@@ -105,7 +85,7 @@ class LockManager {
     std::optional<std::pair<pid_t, pid_t>> probe_mutex_internal(const std::string& mutex_name);
 
     // Maps from mutex name to an initialized mutex.
-    // Mutex names are made from mutex type name or directly mutex name combined with device number.
+    // Mutex names are made from the mutex type name, combined with device number for chip specific ones.
```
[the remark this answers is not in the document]
broskoTT: Correct, chip specific names carry the device type as well as the number. Updated.

PR #3232, opened 2026-08-22
Back chip specific PCIe locks with KMD resource locks

broskoTT:
### Issue
#2745

### Description
Perf analysis: https://tenstorrent.atlassian.net/wiki/spaces/UMD/pages/2564948067/Locks+perf+analysis

Shared memory locks require every participant to see the same /dev/shm, which does not hold across
containers. Chip specific locks on a PCIe device now also take a KMD resource lock, which lives in
the driver and needs nothing shared beyond the device itself.

They take both for now. A process running an older UMD knows only about the shared memory lock, so
dropping it would leave old and new processes not serializing at all. The shared memory half goes
away once every client is on KMD locks. Locks with no KMD equivalent, meaning system wide ones and
locks on a JTAG device, are unchanged.

### List of the changes
- Split LockManager's internals into a RobustMutex backed and a KMD backed set
- Add the mutex type to KMD lock index mapping, above the range KMD reserves for ERISC cores
- Add CompositeMutex, holding both backends in a fixed order, for the duration of the transition
- Add a test asserting a chip specific PCIe lock is held in both backends

### Testing
umd_misc_tests. The new test needs a card and is skipped without one.

### API Changes
This PR has no API changes.

### Full stack diff

### DIFF
--- a/device/utils/lock_manager.cpp
+++ b/device/utils/lock_manager.cpp
@@ -11,6 +11,7 @@
 #include <unordered_map>
 
 #include "umd/device/utils/error.hpp"
+#include "umd/device/utils/kmd_mutex.hpp"
 #include "umd/device/utils/robust_mutex.hpp"
 
 namespace tt::umd {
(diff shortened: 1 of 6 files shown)


pjanevskiTT (approved):

pjanevskiTT (commented):
won't block on the comment for deadlock for exception, it's easy to solve

device/utils/lock_manager.cpp
```diff
@@ -33,6 +34,72 @@ std::string get_mutex_name(MutexType mutex_type, int device_id, IODeviceType dev
...
+public:
+    CompositeMutex(std::unique_ptr<MutexInterface> first, std::unique_ptr<MutexInterface> second) :
+        first_(std::move(first)), second_(std::move(second)) {}
+
+    void initialize() override {
+        first_->initialize();
+        second_->initialize();
+    }
+
+    void lock() override {
```
pjanevskiTT, line 78: `lock()` is not exception safe, and the state it leaves behind does not recover on its own.

`second_` is a `KmdMutex`, and `KmdMutex::lock()` can throw. It polls `try_lock()`, which throws on any `tt_lock_acquire` result other than `0` or `-ENODEV`, and on the `-ENODEV` recovery path it calls `open_device()`, which throws if `tt_device_open` fails - the "device was reset while we were waiting and is not back yet" case `KmdMutex`'s own header discusses.

When that throws, `first_` is still locked, and nothing releases it. `acquire_kmd_mutex` builds the lock via `std::unique_lock(mutex)`, whose constructor calls `lock()`; if that throws the `unique_lock` is never constructed, so no destructor runs and it never took ownership. The exception propagates to the UMD caller, which may catch and retry, so the process stays alive - and `RobustMutex`'s `EOWNERDEAD` recovery only fires when the holder dies. The shared memory half is then held for the lifetime of the process, and every other process on the host blocks on that lock indefinitely.

`unlock()` has the same shape: `second_->unlock()` can throw (`RobustMutex::unlock` on `pthread_mutex_unlock` failure, `KmdMutex::unlock` on any non-`ENODEV` failure), leaving `first_` held. There it is worse, because `unlock()` is reached from `~unique_lock` and an escaping exception in a destructor is `std::terminate`. The terminate part already exists with a bare `RobustMutex`; what is new here is the second throw site and the stranded `first_`.

Wrapping `second_->lock()` in a try/catch that unlocks `first_` and rethrows covers it, and the same for `unlock()`.

The ordering itself is right, for what it is worth: `lock()`, `probe_lock()` and `initialize()` all go first then second, `unlock()` reverses it, and `probe_lock` correctly gives `first_` back when `second_` turns out to be taken.

broskoTT: Right, and the reasoning about why it does not recover is the part that makes it matter: the caller
never receives the `unique_lock` when the constructor's `lock()` throws, so no destructor runs, and
`EOWNERDEAD` only fires when the holder dies rather than catches. A caller that retries leaves the
shared memory half held for the lifetime of the process, with every other process on the host behind
it.

Fixed in `lock()` and `unlock()`, and in `probe_lock()` too - `second_->probe_lock()` can throw for
the same reasons, and it had the same hole even though the contention path was already handled.

On `unlock()`: it still rethrows, so the terminate through `~unique_lock` is unchanged, since that is
pre-existing with a bare RobustMutex and not something to bury here. What changes is that the first
one is no longer stranded on the way out.

Thanks for checking the ordering as well - that half was deliberate, and it is useful to have it
confirmed by someone else rather than just asserted in the comment.

device/utils/lock_manager.cpp
```diff
@@ -32,58 +33,179 @@ std::string get_mutex_name(MutexType mutex_type, int device_id, IODeviceType dev
...
+        first_->lock();
+        second_->lock();
+    }
```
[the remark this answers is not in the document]
broskoTT: Correct on both, fixed - `lock()`, `unlock()` and `probe_lock()` all give the first one back if taking
the second does not work out. See the longer thread on the same code for why a stranded lock here does
not recover on its own.

PR #3236, opened 2026-08-23
[base API] Move DeviceProtocol into TTDeviceModel

broskoTT:
### Issue
#2864

### Description
The model now builds its own `DeviceProtocol` from the transport object it is
handed, so the transport no longer reaches `TTDevice` at all. Each constructor
knows the concrete protocol type it built, so the interfaces that protocol
provides (PCIe, DMA, JTAG, remote) are recorded by plain assignment -- no
runtime type checks anywhere.

Transport selects a model constructor rather than a model subclass, so one
architecture model still serves PCIe, JTAG and remote alike. With the protocol
out of `TTDevice`, its three silicon constructors collapse into one and six
members go away. The model returns nullptr for an interface a transport does not
provide, while `TTDevice`'s getters keep throwing, so its public behaviour is
unchanged.

### List of the changes
- `TTDeviceModel` gains one constructor per transport, each building its protocol
and recording the interfaces it provides; the architecture models forward to them.
- `TTDevice` drops `device_protocol_` and the five cached capability pointers, and
serves the protocol and interface getters from the model.
- Collapse the three silicon `TTDevice` constructors, and the `WormholeTTDevice` /
`BlackholeTTDevice` constructors, into one each.
- `TTDevice::create` and the remote factories build the model from the transport.

### Testing
CI

### API Changes
This PR has API changes.

### Full stack diff

### DIFF
--- a/device/api/umd/device/tt_device_model/blackhole_tt_device_model.hpp
+++ b/device/api/umd/device/tt_device_model/blackhole_tt_device_model.hpp
@@ -4,21 +4,49 @@
 
 #pragma once
 
+#include <memory>
+
 #include "umd/device/tt_device_model/tt_device_model.hpp"
 
 namespace tt::umd {
 
-// Model for a Blackhole device, over any transport that reaches one.
+class JtagDevice;
+class RemoteCommunication;
+
+// Model for a Blackhole device, over any transport that reaches one. Each constructor builds the
+// protocol for its transport and, knowing the concrete protocol type, records the interfaces that
+// protocol provides -- the rest stay null.
 class BlackholeTTDeviceModel : public TTDeviceModel {
 public:
-    explicit BlackholeTTDeviceModel(int communication_device_id);
+    BlackholeTTDeviceModel(std::unique_ptr<PCIDevice> pci_device, bool use_safe_api);
+    BlackholeTTDeviceModel(std::unique_ptr<JtagDevice> jtag_device, uint8_t jlink_id);
 
     tt::ARCH get_arch() const override;
 
     int get_communication_device_id() const override;
 
+    DeviceProtocol *get_device_protocol() override;
+
+    PcieInterface *get_pcie_interface() override;
+
+    DmaInterface *get_dma_interface() override;
+
+    JtagInterface *get_jtag_interface() override;
+
+    RemoteInterface *get_remote_interface() override;
+
+    PCIDevice *get_pci_device() override;
+
 private:
     int communication_device_id_;
+
+    std::unique_ptr<DeviceProtocol> protocol_;
+    PcieInterface *pcie_interface_ = nullptr;
+    DmaInterface *dma_interface_ = nullptr;
+    JtagInterface *jtag_interface_ = nullptr;
+    RemoteInterface *remote_interface_ = nullptr;
+    // Owned by the PCIe protocol; retained only to serve get_pci_device().
+    PCIDevice *pci_device_ = nullptr;
 };
 
 }  // namespace tt::umd
--- a/device/api/umd/device/tt_device_model/tt_device_model.hpp
+++ b/device/api/umd/device/tt_device_model/tt_device_model.hpp
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include "umd/device/tt_device/protocol/device_protocol.hpp"
 #include "umd/device/types/arch.hpp"
 
 namespace tt::umd {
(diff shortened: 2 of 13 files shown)


device/api/umd/device/tt_device_model/tt_device_model.hpp
```diff
@@ -48,25 +55,44 @@ class TTDeviceModel {
     // JTAG. A remote device reports the identity of the local device it is reached through.
     int get_communication_device_id() const;
 
-    virtual HangDetector *get_hang_detector() { return nullptr; }
+    // Null for a backend reached without a host transport, such as simulation.
+    DeviceProtocol *get_device_protocol() const;
```
nbuncicTT: `TTDeviceModel` is an interface, not an owning class. It is up to the derived class to own whichever things it needs, otherwise, there's no point in making this class. I understand that this will make some duplication in the derived classes of the silicon-based ones, but we can make a `Transport` class which encapsulates all of the transports and we can re-use this via composition.

Same goes for all other private attributes, they mustn't be in here.

device/api/umd/device/tt_device_model/blackhole_tt_device_model.hpp
```diff
@@ -0,0 +1,55 @@
...
+// protocol for its transport and, knowing the concrete protocol type, records the interfaces that
+// protocol provides -- the rest stay null.
+class BlackholeTTDeviceModel : public TTDeviceModel {
+public:
+    BlackholeTTDeviceModel(std::unique_ptr<PCIDevice> pci_device, bool use_safe_api);
+    BlackholeTTDeviceModel(std::unique_ptr<JtagDevice> jtag_device, uint8_t jlink_id);
+
+    tt::ARCH get_arch() const override;
+
+    IODeviceType get_communication_device_type() const override;
```
nbuncicTT: `get_communication_device_type` isn't part of the spec, hence remove it or if it's needed from the model temporary, state so in a comment.

nbuncicTT (approved):

device/api/umd/device/tt_device_model/tt_device_model.hpp
broskoTT: Already the case after the #3225 rework — you were on the earlier push. The base has no members or non-public sections now; each model owns its own protocol and interface pointers.

Agreed on `Transport` as a follow-up. The duplication is real — the unity build already forced arch-specific names on the two `resolve_*_soc_arch_descriptor` helpers because their copies collided in one TU.

device/api/umd/device/tt_device_model/blackhole_tt_device_model.hpp
broskoTT: Already removed, via the route you asked for on #3225: gone from all three models, and `TTDevice` derives it from the transport interface it holds. The silicon models no longer take an `IODeviceType` at all.

device/api/umd/device/tt_device_model/tt_device_model.hpp
```diff
@@ -51,6 +53,9 @@ class TTDeviceModel {
     // JTAG. A remote device reports the identity of the local device it is reached through.
     virtual int get_communication_device_id() const = 0;
 
+    // Required components.
+    virtual DeviceProtocol *get_device_protocol() = 0;
+
```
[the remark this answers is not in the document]
broskoTT: `SimulationTTDeviceModel::get_device_protocol()` already documents why it is null: a simulation backend has no transport protocol, because the simulation `TTDevice` overrides the read/write paths instead of routing them through one. Nothing dereferences it blindly — those overrides never reach the protocol.

device/tt_device/tt_device.cpp
```diff
@@ -173,15 +139,11 @@ void TTDevice::init_tt_device(const std::chrono::milliseconds timeout_ms) {
         switch (arch) {
             case ARCH::WORMHOLE_B0:
                 return std::unique_ptr<WormholeTTDevice>(new WormholeTTDevice(
-                    std::make_unique<WormholeTTDeviceModel>(IODeviceType::JTAG, device_number),
-                    std::move(jtag_device),
-                    device_number,
+                    std::make_unique<WormholeTTDeviceModel>(std::move(jtag_device), device_number),
                     soc_arch_descriptor));
             case ARCH::BLACKHOLE:
```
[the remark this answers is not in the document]
broskoTT: Pre-existing, not introduced here. On `main`, `WormholeTTDevice`/`BlackholeTTDevice` already take `uint8_t jlink_id` and `create()` already passes `device_number` (int) straight in — tt_device.cpp:178/181. This PR only moves where the narrowing happens. Worth its own issue rather than folding a JTAG-path fix into the model migration.

PR #3237, opened 2026-08-23
[base API] Move SocArchDescriptor into TTDeviceModel

broskoTT:
### Issue
#2864

### Description
The model resolves its own SoC architecture descriptor: it validates a supplied
one against the architecture it reports, or builds that architecture's default
when none is given. `TTDevice` reads it from the model when it constructs its
`SocDescriptor`, and no longer takes or stores one.

This leaves `TTDevice` with a single constructor -- the descriptor was the last
thing separating the silicon path from the transport-less one -- and the
architecture devices with a single constructor each, taking nothing but a model.

### List of the changes
- `TTDeviceModel` resolves and owns the SoC architecture descriptor; its
architecture check moves here from `TTDevice::assign_soc_arch_descriptor()`.
- Drop `TTDevice::assign_soc_arch_descriptor()` and `soc_arch_descriptor_`;
`init_tt_device()` builds the `SocDescriptor` from the model's descriptor.
- Collapse `TTDevice` to one constructor, and `WormholeTTDevice` /
`BlackholeTTDevice` to one each taking only a model.
- `TTDevice::create` passes the caller's descriptor to the model.

### Testing
CI

### API Changes
This PR has API changes.

### Full stack diff

### DIFF
--- a/device/tt_device_model/blackhole_tt_device_model.cpp
+++ b/device/tt_device_model/blackhole_tt_device_model.cpp
@@ -4,48 +4,23 @@
 
 #include "umd/device/tt_device_model/blackhole_tt_device_model.hpp"
 
-#include <fmt/format.h>
-
 #include <utility>
 
+#include "tt_device_model/soc_arch_descriptor_resolver.hpp"
 #include "umd/device/jtag/jtag_device.hpp"
 #include "umd/device/pcie/pci_device.hpp"
 #include "umd/device/pcie/silicon_tlb_window.hpp"
-#include "umd/device/soc_arch_descriptor.hpp"
 #include "umd/device/tt_device/protocol/jtag_protocol.hpp"
 #include "umd/device/tt_device/protocol/pcie_protocol.hpp"
-#include "umd/device/utils/error.hpp"
 
 namespace tt::umd {
-
-namespace {
-
-// A caller may supply the descriptor; otherwise it comes from the architecture's constants.
-// Named per architecture on purpose: a unity build folds the model sources into one translation
-// unit, where a shared name in this anonymous namespace would collide with the other model's copy.
-std::shared_ptr<SocArchDescriptor> resolve_blackhole_soc_arch_descriptor(
-    const std::shared_ptr<SocArchDescriptor> &soc_arch_descriptor) {
-    if (soc_arch_descriptor == nullptr) {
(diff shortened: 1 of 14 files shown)


device/tt_device_model/blackhole_tt_device_model.cpp
```diff
@@ -4,19 +4,47 @@
...
 #include "umd/device/tt_device/protocol/pcie_protocol.hpp"
+#include "umd/device/utils/error.hpp"
 
 namespace tt::umd {
 
-BlackholeTTDeviceModel::BlackholeTTDeviceModel(std::unique_ptr<PCIDevice> pci_device, bool use_safe_api) :
+namespace {
+
+// A caller may supply the descriptor; otherwise it comes from the architecture's constants.
+std::shared_ptr<SocArchDescriptor> resolve_soc_arch_descriptor(
```
nbuncicTT: Why is this in an anonymous namespace? 

Also if you have the need for this kind of helper on all concrete silicon models, create a template which is invoked on the ones that have an ARCH check and check the template parameter against the passed parameter (i.e. the <T> vs the arch passed to the helper).

nbuncicTT (approved):

device/tt_device_model/blackhole_tt_device_model.cpp
broskoTT: Anonymous namespace was just the file-local idiom, but you picked at a real weak spot: it is exactly what broke the tt-metal build. Unity builds fold both model sources into one TU, so the two anonymous namespaces merged and the identically-named helpers collided. I had patched that by giving them arch-specific names.

Took your template suggestion instead, which fixes the cause rather than the symptom — one definition in a private header, architecture as a non-type template parameter, supplied descriptor checked against it:

```cpp
template <tt::ARCH arch>
std::shared_ptr<SocArchDescriptor> resolve_soc_arch_descriptor(
    const std::shared_ptr<SocArchDescriptor> &soc_arch_descriptor);
```

Called as `resolve_soc_arch_descriptor<tt::ARCH::WORMHOLE_B0>(...)`. Net -50/+34: the duplicated `UMD_ASSERT` is gone, the arch-specific names and the comment explaining the collision are gone, the anonymous namespace is unnecessary, and four includes drop out of each model source. A template in a header is one entity across TUs, so the collision class no longer exists.

PR #3239, opened 2026-08-24
[Base API] Remove TTDevice dependency from FirmwareInfoProvider

### LINKED ISSUE #2876: Migrate FirmwareInfoProvider interface to Base API Specification

aleksamarkovicTT:
### Issue
Finally closes #2876 

### Description
Remove TTDevice dependency from FirmwareInfoProvider

### List of the changes
- Remove TTDevice dependency from FirmwareInfoProvider
- Update constructor and static builder method calls.
- Update nanobind

### Testing
CI

### API Changes
This PR has API changes.

### DIFF
--- a/device/api/umd/device/firmware/firmware_info_provider_implementation.hpp
+++ b/device/api/umd/device/firmware/firmware_info_provider_implementation.hpp
@@ -12,6 +12,7 @@
 
 #include "umd/device/firmware/firmware_info_provider.hpp"
 #include "umd/device/firmware/firmware_telemetry_mapping.hpp"
+#include "umd/device/tt_device/tt_device.hpp"
 #include "umd/device/types/cluster_descriptor_types.hpp"
 #include "umd/device/types/core_coordinates.hpp"
 #include "umd/device/types/gddr_telemetry.hpp"
@@ -23,7 +24,8 @@ enum class ARCH;
 }  // namespace tt
 
 namespace tt::umd {
-class TTDevice;
+class DeviceProtocol;
+class FirmwareTelemetryReader;
 
 /*
  * FirmwareInfoProvider is a data-driven class that abstracts away the details of specific firmware
@@ -34,9 +36,20 @@ class TTDevice;
  */
 class FirmwareInfoProviderImplementation : public FirmwareInfoProvider {
 public:
-    static std::unique_ptr<FirmwareInfoProvider> create_firmware_info_provider(TTDevice* tt_device);
+    static std::unique_ptr<FirmwareInfoProvider> create_firmware_info_provider(
+        tt::ARCH arch,
+        DeviceProtocol* device_protocol,
+        xy_pair arc_core_noc0,
+        xy_pair arc_core_noc1,
+        FirmwareTelemetryReader* telemetry);
 
-    FirmwareInfoProviderImplementation(TTDevice* tt_device);
+    FirmwareInfoProviderImplementation(
+        FirmwareBundleVersion firmware_version,
+        tt::ARCH arch,
+        DeviceProtocol* device_protocol,
+        xy_pair arc_core_noc0,
+        xy_pair arc_core_noc1,
+        FirmwareTelemetryReader* telemetry);
 
     virtual FirmwareBundleVersion get_firmware_version([[maybe_unused]] NocId noc_id = NocId::DEFAULT_NOC) const;
 
@@ -146,7 +159,11 @@ private:
      */
     std::vector<std::pair<CoreCoord, bool>> parse_eth_status_bitmask(uint16_t bitmask) const;
 
-    TTDevice* tt_device = nullptr;
+    tt::ARCH arch_ = ARCH::Invalid;
+    DeviceProtocol* device_protocol_ = nullptr;
+    FirmwareTelemetryReader* telemetry_ = nullptr;
+    xy_pair arc_core_noc0_;
+    xy_pair arc_core_noc1_;
 
     FirmwareBundleVersion firmware_version = FirmwareBundleVersion(0, 0, 0);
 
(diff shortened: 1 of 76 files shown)


nbuncicTT (approved):

device/firmware/firmware_info_provider_implementation.cpp
```diff
@@ -811,27 +830,29 @@ FirmwareInfoProviderImplementation::get_eth_retrain_status_per_core(NocId noc_id
...
 std::optional<uint32_t> FirmwareInfoProviderImplementation::get_runtime_telemetry_buffer_address(NocId noc_id) const {
     uint32_t address = 0;
-    switch (tt_device->get_arch()) {
+    switch (arch_) {
         case ARCH::WORMHOLE_B0:
             if (firmware_version < FirmwareBundleVersion(19, 13, 0)) {
                 return std::nullopt;
             }
-            tt_device->read_from_device(
+            device_protocol_->read_ctrl(
```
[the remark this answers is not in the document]
aleksamarkovicTT: I prefer read_ctrl here.

device/api/umd/device/firmware/firmware_info_provider_implementation.hpp
```diff
@@ -12,6 +12,7 @@
 
 #include "umd/device/firmware/firmware_info_provider.hpp"
 #include "umd/device/firmware/firmware_telemetry_mapping.hpp"
+#include "umd/device/tt_device/tt_device.hpp"
```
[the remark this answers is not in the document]
aleksamarkovicTT: this will be removed.

device/firmware/firmware_info_provider_implementation.cpp
```diff
@@ -32,37 +34,64 @@
...
+        UMD_THROW(error::RuntimeError, "FirmwareTelemetryReader cannot be nullptr.");
     }
-    firmware_version = FirmwareBundleVersion(0, 0, 0);
-    if (tt_device->get_arch() == ARCH::WORMHOLE_B0 && dynamic_cast<SmBusArcTelemetryReader*>(telemetry) != nullptr) {
+    if (device_protocol == nullptr) {
+        UMD_THROW(error::RuntimeError, "DeviceProtocol cannot be nullptr.");
+    }
+
+    FirmwareBundleVersion firmware_version = FirmwareBundleVersion(0, 0, 0);
+    if (dynamic_cast<SmBusArcTelemetryReader*>(telemetry) != nullptr) {
```
pjanevskiTT, line 74: The `tt_device->get_arch() == ARCH::WORMHOLE_B0 &&` guard was dropped even though `arch` is now a parameter. Today it's equivalent — `create_arc_telemetry_reader` only builds an `SmBusArcTelemetryReader` in the `WORMHOLE_B0` case — but this factory is public and takes `arch` and the reader independently, so a Blackhole caller passing an SmBus reader would now read `wormhole::LegacyTelemetryTag::FW_BUNDLE_VERSION`. Cheap to keep the conjunct.

device/firmware/firmware_info_provider_implementation.cpp
```diff
@@ -811,27 +830,29 @@ FirmwareInfoProviderImplementation::get_eth_retrain_status_per_core(NocId noc_id
...
+    switch (arch_) {
         case ARCH::WORMHOLE_B0:
             if (firmware_version < FirmwareBundleVersion(19, 13, 0)) {
                 return std::nullopt;
             }
-            tt_device->read_from_device(
+            device_protocol_->read_ctrl(
                 &address,
-                tt_device->get_arc_core(),
+                noc_id == NocId::NOC1 ? arc_core_noc1_ : arc_core_noc0_,
```
pjanevskiTT, line 840: This changes which ARC core is targeted for threads that use `NocIdSwitcher`. Before, the core came from `TTDevice::get_arc_core()` (`is_selected_noc1() ? arc_core_noc1 : arc_core_noc0`, i.e. thread state) while the routing NOC came from `read_from_device`'s default argument (`DEFAULT_NOC` = NOC0) — so a NOC1-selected thread sent NOC1 ARC coordinates over NOC0. Now both come from `noc_id`.

The new form is self-consistent and fixes that mismatch, but it isn't a no-op: callers relying on the thread-selected NOC now have to pass `noc_id` explicitly. Worth calling out, since the stated scope is only removing the `TTDevice` dependency.

device/api/umd/device/firmware/firmware_info_provider_implementation.hpp
```diff
@@ -34,9 +35,20 @@ class TTDevice;
...
-    FirmwareInfoProviderImplementation(TTDevice* tt_device);
+    static std::unique_ptr<FirmwareInfoProvider> create_firmware_info_provider(
+        tt::ARCH arch,
+        DeviceProtocol* device_protocol,
+        xy_pair arc_core_noc0,
+        xy_pair arc_core_noc1,
+        FirmwareTelemetryReader* telemetry);
+
+    FirmwareInfoProviderImplementation(
+        FirmwareBundleVersion firmware_version,
```
pjanevskiTT, line 46: The constructor no longer derives `firmware_version` from telemetry — the factory does, and the constructor trusts what it's handed and builds `firmware_feature_map` from it. Since the constructor is public API, a direct construction (default `FirmwareBundleVersion(0, 0, 0)`) now silently yields the oldest feature map, which wasn't possible before. Making it protected with the factory as the sanctioned path, or documenting that the version must come from telemetry, would keep that from being an easy mistake.

Issue #3243, opened 2026-08-24
Reduce 6u CI runtime by bounding per-chip test sweeps


broskoTT:
The `6u` label resolves to a single runner, and every job targeting it serialises. A merge queue entry holds it for ~21 minutes back to back, and PR jobs queue behind that - in one recent PR run the test-all job waited ~18 minutes after its build artifact was ready. This has grown in priority with the PR count from the hackathon.

Most of the cost is tests whose inner loop is O(chips) or O(chips x cores). The same tests that take 86s on an N300 take 367s on 6u, since 6u has 32 chips and all of them are MMIO capable.

Tracking test optimization work to bring that down.

PR #3244, opened 2026-08-24
Bound the per-chip sweeps that make CI tests scale with cluster size

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/3243

### Description
Various optimizations for 6u tests which are currently the bottleneck of our CI.

### List of the changes
- `SysmemReadWrite` exercises at most 2 MMIO chips. Two chips with distinct sentinel patterns
  already detect a misrouted host DMA; each further chip repeats that at the cost of a 1 GiB
  fill and a full offset sweep per channel.
- `ClusterAssertDeassertRiscsTest` gives each of its 15 RISC-subset parameters one chip
  instead of all of them. `RiscSetUnderTest` becomes a struct carrying the subset's index,
  which selects the chip, so on a cluster with at least 15 chips every subset lands on a
  different one. `TestRiscProgram.DeassertResetBrisc` still sweeps every chip and core.
- The three `ClusterWH` broadcast tests read back 6 cores per chip - two from the front, two
  from the middle and two from the back of the broadcast target grid - rather than every
  tensix core for every one of 15 broadcast sizes. The row/column exclusion filter is hoisted
  out of the inner loops into one shared helper, replacing four copies of it.
- `test_utils::is_iommu_available` reads sysfs via the new `PCIDevice::detect_iommu` instead
  of building a whole Cluster. It was constructing a 32-chip Cluster, twice, to read one flag.
- Demote two per-item `log_info` calls to `log_debug`: the per-chip-per-core-per-NOC line in
  `TestNocTranslatedCoordinates` (~15k lines of a passing 6u run) and the per-allocation line
  in `tlb_virus` (~5.8k lines; the per-size summary it already prints is what matters). A 6u
  job log is 47k lines against N300's 7k.

### Testing
Builds clean, pre-commit passes, baremetal tests pass. Measured on 6u, comparing this PR's run
against the main run on this PR's base commit 6e34a1a7d, so the two are directly comparable.
All 381 tests present in both, same pass/skip status.

**6u Release job: 7m24s -> 5m42s.** gtest total across the three binaries 384.8s -> 283.6s.
Job log 47575 -> 26859 lines.

| | main | PR | delta |
|---|---|---|---|
| `api_tests` step | 238.5s | 175.8s | -62.7s |
| `wormhole_b0/unit_tests` step | 80.8s | 40.0s | -40.8s |
| `unified_tests` step | 65.9s | 68.2s | +2.3s |
| `TestDeviceIOFixture.SysmemReadWrite` | 65.1s | 29.7s | -35.4s |
| `ClusterAssertDeassertRiscsTest` (15 params) | 43.2s | 3.6s | -39.6s |
| `ClusterWH.BroadcastWrite` | 23.9s | 5.2s | -18.8s |
| `ClusterWH.VirtualCoordinateBroadcast` | 14.3s | 3.4s | -10.9s |
| `ClusterWH.VirtualCoordinateBroadcastPerChip` | 14.3s | 3.4s | -10.9s |
| `WarmResetTest.StaleFileDescriptorClusterRecovery` | 63.2s | 75.2s | +11.9s |
| `Multiprocess.MultipleThreadsMultipleClustersOpenClose` | 40.8s | 43.1s | +2.3s |

The last two are not touched by this PR. Both overflow the host's ~32 GiB hugepage pool and fall
back to regular pages, so they vary with what else is mapped; total regular-page volume is
identical at 210 GiB in both runs. They are also why the `unified_tests` step is slightly up.

`SysmemReadWrite` still spends 21.5s of its 29.7s in Cluster construction, unchanged between the
runs - the chip cap only removes the offset sweep. See the note below on `ClusterOptions`.

### API Changes
This PR has API changes: `PCIDevice::detect_iommu(const PciDeviceInfo&)` is now public. It was
already implemented as a file-static in `pci_device.cpp`; nothing is removed or renamed.

Co-authored-by: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/device/api/umd/device/pcie/pci_device.hpp
+++ b/device/api/umd/device/pcie/pci_device.hpp
@@ -157,6 +157,12 @@ public:
     static PciDeviceInfo read_device_info(const std::string &device_path);
 
     /**
+     * Whether an IOMMU protects the host from the given device. Reads sysfs only, so callers that
+     * just need the IOMMU state do not have to open the device.
+     */
+    static bool detect_iommu(const PciDeviceInfo &device_info);
+
+    /**
      * PCI device constructor.
      *
      * Opens the character device file descriptor, reads device information from
--- a/tests/api/test_risc_program.cpp
+++ b/tests/api/test_risc_program.cpp
@@ -9,6 +9,7 @@
 #include <array>
 #include <cstdint>
 #include <cstring>
+#include <iterator>
 #include <memory>
 #include <optional>
 #include <set>
@@ -183,7 +184,7 @@ TEST_P(ClusterAssertDeassertRiscsTest, TriscNcriscAssertDeassertTest) {
         }
     };
 
-    const auto& configurations_of_risc_cores = GetParam();
+    const auto& configurations_of_risc_cores = GetParam().riscs;
 
     constexpr uint64_t brisc_code_address = 0x20;
 
@@ -193,7 +194,15 @@ TEST_P(ClusterAssertDeassertRiscsTest, TriscNcriscAssertDeassertTest) {
     auto tensix_l1_size = cluster->get_soc_descriptor(0).worker_l1_size;
     std::vector<uint32_t> zero_data(tensix_l1_size / sizeof(uint32_t), 0);
 
-    auto chip_ids = cluster->get_target_device_ids();
+    // Every parameter runs the same sequence over the same core grid, differing only in which RISCs it
+    // brings up, so running each of them on every chip is redundant work that grows with cluster size.
+    // Give each parameter one chip instead: on a cluster with at least as many chips as parameters each
+    // one lands on a different chip. All-chip coverage of assert/deassert stays with
+    // TestRiscProgram.DeassertResetBrisc, which still sweeps every chip and core.
+    const auto cluster_chip_ids = cluster->get_target_device_ids();
+    const std::vector<ChipId> chip_ids = {
+        *std::next(cluster_chip_ids.begin(), GetParam().index % cluster_chip_ids.size())};
+
     for (auto& chip_id : chip_ids) {
         auto brisc_configuration_program = get_brisc_configuration_program_for_chip(cluster.get(), chip_id);
 
--- a/tests/wormhole/test_cluster_wh.cpp
+++ b/tests/wormhole/test_cluster_wh.cpp
@@ -4,6 +4,7 @@
 
 #include <gtest/gtest.h>
 
+#include <algorithm>
 #include <cstddef>
 #include <cstdint>
 #include <initializer_list>
@@ -16,6 +17,7 @@
 #include <string>
 #include <thread>
 #include <unordered_set>
+#include <utility>
 #include <vector>
 
 #include "tests/test_utils/device_test_utils.hpp"
@@ -58,12 +60,14 @@ static const std::vector<uint32_t> T6_X_TRANSLATED_LOCATIONS = {18, 19, 20, 21, 
 static const std::vector<uint32_t> T6_Y_TRANSLATED_LOCATIONS = {18, 19, 20, 21, 22, 23, 24, 25, 26, 27};
 
 // Broadcast payload sizes, a single word up to 64 KiB, covering the chunking in
-// broadcast_write_to_cluster. Only the first broadcast test sweeps all of them; the rest take the
-// first, middle and last size, since repeating the full sweep multiplies out with the number of chips
-// without covering anything the first test does not.
+// broadcast_write_to_cluster. Only the first broadcast test sweeps all of them; repeating the full
+// sweep in the others multiplies out with the number of chips without covering anything new. The
+// reduced set keeps both extremes plus 256 words, which is the exact boundary of the
+// `size_in_bytes > 256 * DATA_WORD_SIZE` check deciding whether a remote write stages through host
+// DRAM, so a `>` vs `>=` slip there still fails.
 static const std::vector<uint32_t> BROADCAST_SIZES = {
     1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384};
-static const std::vector<uint32_t> REDUCED_BROADCAST_SIZES = {1, 128, 16384};
+static const std::vector<uint32_t> REDUCED_BROADCAST_SIZES = {1, 128, 256, 16384};
 
 static void set_barrier_params(Cluster& cluster) {
     // Populate address map and NOC parameters that the driver needs for memory barriers and remote transactions.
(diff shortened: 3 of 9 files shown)


device/api/umd/device/pcie/pci_device.hpp
```diff
@@ -156,6 +156,12 @@ class PCIDevice {
      */
     static PciDeviceInfo read_device_info(const std::string &device_path);
 
+    /**
+     * Whether an IOMMU protects the host from the given device. Reads sysfs only, so callers that
+     * just need the IOMMU state do not have to open the device.
+     */
+    static bool detect_iommu(const PciDeviceInfo &device_info);
```
pjanevskiTT, line 163: maybe this can be in it's own PR, but not that important since it's a very small change

pjanevskiTT (approved):

tests/wormhole/test_cluster_wh.cpp
```diff
@@ -56,12 +57,52 @@ static const std::vector<tt_xy_pair> ETH_CORES_TRANSLATION_ON = {
...
 
+// Two items from the front, two from the middle and two from the back, or all of them if there are
+// fewer than that. Used to sample a chip's core grid instead of sweeping all of it.
+template <typename T>
+static std::vector<T> sample_front_middle_back(const std::vector<T>& items) {
+    if (items.size() <= 6) {
+        return items;
+    }
+    const size_t middle = items.size() / 2 - 1;
+    return {items[0], items[1], items[middle], items[middle + 1], items[items.size() - 2], items.back()};
```
pjanevskiTT: The six cores here are three *adjacent pairs* out of a row-major list (`get_cores(CoreType::TENSIX)` iterates y outer, x inner - `coordinate_manager.cpp:506`), so they collapse onto roughly three grid positions clustered at the head, middle and tail. The count isn't the issue; the issue is which columns fall out of it.

`BroadcastWrite` (`cols_to_exclude = {0, 5}`, NOC0): tensix x are `{1,2,3,4,6,7,8,9}`, so 8x10 = 80 targets and the sample lands on x `{1,2,8,9}`. Columns 4 and 6 - the two flanking excluded column 5 - are never read back. If the mask were off by one and excluded 6 instead of 5, column 6 would never receive the broadcast and this test still passes.

`VirtualCoordinateBroadcast` / `...PerChip` (`cols_to_exclude = {16,17,20}`, translated): targets are y `{18,19,21,23,24,25}` x x `{18,19,21,22,23,24,25}` = 42, sample lands on x `{18,19,24,25}` and y `{18,21,23,25}`. x = 21 - the core immediately after excluded column 20 - is never read, nor are 22 and 23; rows 19 and 24 are skipped too.

Since the broadcast is driven by per-row/per-column exclusion masks mapped through `TRANSLATED_TO_VIRTUAL_X/Y` (`ethernet_broadcast.cpp:243-267`), a one-off in that mapping drops or adds a single row/column strip - which is exactly what these tests exist to catch, and it lands on cores the sample skips. The coverage also shifts with harvesting, since the grid shape changes.

Suggestion, at essentially the same cost: walk a staircase through the target rectangle rather than taking adjacent pairs, e.g. `for i in 0..max(rows.size(), cols.size()): (rows[i % rows.size()], cols[i % cols.size()])`. That is 7 cores for `VirtualCoordinateBroadcast` and 8-10 for `BroadcastWrite`, and it touches every target row *and* every target column, so any single dropped or added strip fails.

tests/wormhole/test_cluster_wh.cpp
```diff
@@ -56,12 +57,52 @@ static const std::vector<tt_xy_pair> ETH_CORES_TRANSLATION_ON = {
 static const std::vector<uint32_t> T6_X_TRANSLATED_LOCATIONS = {18, 19, 20, 21, 22, 23, 24, 25};
 static const std::vector<uint32_t> T6_Y_TRANSLATED_LOCATIONS = {18, 19, 20, 21, 22, 23, 24, 25, 26, 27};
 
+// Broadcast payload sizes, a single word up to 64 KiB, covering the chunking in
+// broadcast_write_to_cluster. Only the first broadcast test sweeps all of them; the rest take the
+// first, middle and last size, since repeating the full sweep multiplies out with the number of chips
+// without covering anything the first test does not.
+static const std::vector<uint32_t> BROADCAST_SIZES = {
+    1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384};
+static const std::vector<uint32_t> REDUCED_BROADCAST_SIZES = {1, 128, 16384};
```
pjanevskiTT: The one real behavioral switch across this sweep is `use_host_dram = size_in_bytes > 256 * DATA_WORD_SIZE` (`remote_communication_legacy_firmware.cpp:159`), and `{1, 128, 16384}` does still exercise both sides of it (128 below, 16384 above), so this reduction is largely fine. What it drops is `256` - the exact boundary value, and the only size that tested it. A `>` vs `>=` slip there now goes undetected. Adding `256` back costs 1 KiB per iteration.

tests/wormhole/test_cluster_wh.cpp
broskoTT: Good catch, you are right and I took your suggestion in 618086546.

Confirmed the ordering: `get_tensix_cores()` is y outer / x inner (`coordinate_manager.cpp:503-505`), so my three "front/middle/back" pairs were three *adjacent* pairs and collapsed onto ~3 rows and 4 columns. I reproduced your numbers exactly - `BroadcastWrite` sampled x `{1,2,8,9}` of 8 columns, `VirtualCoordinateBroadcast` x `{18,19,24,25}` and y `{18,21,23,25}`. So the columns flanking an excluded column were never read, and an off-by-one in the mask would have passed.

`broadcast_readback_cores` now walks the staircase you described: collect the distinct target rows and columns, then pair the i-th row with the i-th column cycling the shorter axis. Verified coverage:

| test | full grid | sample | all rows | all cols |
|---|---|---|---|---|
| `BroadcastWrite` | 80 | 10 | yes | yes |
| `VirtualCoordinateBroadcast` / `PerChip` | 42 | 7 | yes | yes |
| `BroadcastWrite`, 2 rows harvested | 64 | 8 | yes | yes |

I put the reasoning in the comment on the helper so the simpler version does not come back. There is a `find()` guard for a non-rectangular target set - on Wormhole harvesting drops whole rows so the remainder stays rectangular, but it keeps the sample honest if that ever changes.

tests/wormhole/test_cluster_wh.cpp
broskoTT: Agreed, 256 is back in `REDUCED_BROADCAST_SIZES` as of 618086546.

One correction to the mechanism though, which I hit while checking this. There are two `use_host_dram` sites, and the broadcast one short-circuits the size term:

```cpp
// remote_communication_legacy_firmware.cpp:406
bool use_host_dram = (broadcast || (size_in_bytes > 256 * DATA_WORD_SIZE)) && system_mem_available;
```

With `broadcast == true` that is true at every size, so on the broadcast path itself the boundary is never the deciding factor. What actually reaches the `> 256 * DATA_WORD_SIZE` check is `write_to_non_mmio` with `broadcast == false` (line 168) - the per-core readbacks and zeroing writes inside the loop, on remote chips.

So your conclusion holds, just via the point-to-point path rather than the broadcast itself: `{1, 128, 16384}` gives 512 and 65536 bytes but nothing at exactly 1024, which is the only value that separates `>` from `>=`. Added it with that spelled out in the comment.

device/api/umd/device/pcie/pci_device.hpp
broskoTT: Fair, and it would have been cleaner as its own PR. Leaving it here since it is three lines and the PR is already reviewed - splitting it now is more churn than it saves. Will keep it separate next time.

tests/api/test_risc_program.cpp
```diff
@@ -193,7 +194,15 @@ TEST_P(ClusterAssertDeassertRiscsTest, TriscNcriscAssertDeassertTest) {
...
-    auto chip_ids = cluster->get_target_device_ids();
+    // Every parameter runs the same sequence over the same core grid, differing only in which RISCs it
+    // brings up, so running each of them on every chip is redundant work that grows with cluster size.
+    // Give each parameter one chip instead: on a cluster with at least as many chips as parameters each
+    // one lands on a different chip. All-chip coverage of assert/deassert stays with
+    // TestRiscProgram.DeassertResetBrisc, which still sweeps every chip and core.
+    const auto cluster_chip_ids = cluster->get_target_device_ids();
+    const std::vector<ChipId> chip_ids = {
+        *std::next(cluster_chip_ids.begin(), GetParam().index % cluster_chip_ids.size())};
+
```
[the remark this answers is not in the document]
broskoTT: I do not think this is reachable. The test already dereferences chip 0 twice before this line - `cluster->get_tt_device(0)->get_arch()` in the Blackhole skip, and `cluster->get_soc_descriptor(0).worker_l1_size` for the zero buffer - so an empty cluster throws there first and never reaches the modulus.

The claim that the test previously no-oped is also not right: the old code would have had to get past those same two calls before reaching the chip loop.

Leaving it as is rather than adding an assertion that cannot fire.

Issue #3249, opened 2026-08-25
Migrate SystemMemory interface to Base API Specification


pjanevskiTT:
**Description**

The current codebase has `SysmemBuffer` and `SysmemManager`, which together cover what the Base API Specification defines as `SystemMemoryBuffer` and `SystemMemoryAllocator`. They differ from the spec in naming, in how buffer memory is owned, and in how a NOC address is bound. Align them with the spec. The legacy channel/hugepage API stays on the concrete implementations for now and is retired separately.

**Scope**

- Rename `SysmemBuffer` to `SystemMemoryBuffer` and its accessors to the spec names
- Add `write_to_sysmem()` / `read_from_sysmem()` on the buffer
- Add `get_communication_id()` on the buffer and the allocator
- Give the buffer ownership of its memory through the spec's `Deleter`
- Create the `SystemMemoryAllocator` interface per the spec and have `SysmemManager` implement it
- Add deferred `bind_noc_address()` with the spec's `NocBinder`, absorbing `LocalChip::init_pcie_iatus()`
- Make the buffer constructors private with the allocator as friend

**Is your feature request related to a problem? Please describe.**

The system memory classes do not match the Base API Specification, and the buffer does not own the memory it hands out — `SiliconSysmemManager::allocate_sysmem_buffer()` mmaps a mapping that is never munmapped.

**Definition of Done**

- `SystemMemoryBuffer` and `SystemMemoryAllocator` match the Base API Specification
- Buffer memory is owned via the spec's `Deleter`, closing the mmap leak
- NOC binding goes through `bind_noc_address()`
- Concrete implementations retain the legacy channel/hugepage API
- All consumers updated
- Existing tests pass or are updated

PR #3252, opened 2026-08-25
Add host copy methods to SysmemBuffer

pjanevskiTT:
### Issue
#3249

### Description
Adds the buffer-side host copy methods the Base API Specification puts on `SysmemBuffer`. Both are pure host-side `memcpy`s against the buffer VA — the device is not involved. They are bounded by the user-requested size, so on an unaligned buffer they address the caller's VA rather than the larger page-aligned mapping pinned underneath.

Note these are deliberately not gated on `DeviceBufferAccess`: `READ_ONLY` constrains what the device may do to the mapping and explicitly does not restrict host access, so writing into a device-read-only buffer from the host is the intended pattern.

Stacked on #3250.

### List of the changes
- Added `SysmemBuffer::write_to_sysmem()` and `read_from_sysmem()`, both routed through the existing `validate(offset, size)` bounds check.
- Added simulation-side tests for the round trip at zero and non-zero offsets, the last addressable byte, and the out-of-bounds cases in both directions.
- Added a silicon-side test that a copy on a deliberately unaligned mapping lands at the caller's VA and is bounded by the requested size.

### Testing
CI

### API Changes
This PR has no API changes.

### DIFF
--- a/device/api/umd/device/chip_helpers/sysmem_buffer.hpp
+++ b/device/api/umd/device/chip_helpers/sysmem_buffer.hpp
@@ -68,6 +68,27 @@ public:
     ~SysmemBuffer();
 
     /**
+     * Copies data from a caller-provided source buffer into this system memory buffer.
+     * Pure host-side operation, the device is not involved. This is allowed regardless of
+     * DeviceBufferAccess, which only constrains what the device may do to the mapping.
+     *
+     * @param src Pointer to the source host memory.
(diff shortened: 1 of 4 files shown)


pjanevskiTT:
Superseded by #3265 - please review there instead.

The commit itself is unchanged - it was re-pushed onto the new base as #3265.

PR #3253, opened 2026-08-25
Add get_communication_id to SysmemBuffer and SysmemManager

pjanevskiTT:
### Issue
#3249

### Description
Adds `get_communication_id()` to `SysmemBuffer` and `SysmemManager`, per the Base API Specification. A manager pins for exactly one device and stamps that device's id into every buffer it produces, so a caller holding a buffer can confirm it belongs to the device it is about to be used with — its IOVA is only valid there. The value matches `TTDevice::get_communication_device_id()`.

`SiliconSysmemManager` takes the id from its `TTDevice`; `SimulationSysmemManager` uses its chip id. The id is carried on the manager as a protected member alongside `pcie_base_`, following what that class already does.

Stacked on #3252.

### List of the changes
- Added `SysmemBuffer::get_communication_id()`, populated from the `TTDevice` in the device-backed constructor and passed explicitly in the pre-mapped one.
- Added `SysmemManager::get_communication_id()`, set by both concrete managers.
- Added tests that a manager's buffers carry its id, that managers for different chips stamp different ids, and that the id matches the `TTDevice` on silicon.

### API Changes
This PR has API changes.

- `SysmemBuffer`'s pre-mapped constructor takes a new required `communication_id` parameter after `device_io_addr`. The device-backed constructor is unchanged.

### Testing
CI

### DIFF
--- a/device/api/umd/device/chip_helpers/sysmem_buffer.hpp
+++ b/device/api/umd/device/chip_helpers/sysmem_buffer.hpp
@@ -58,10 +58,17 @@ public:
         size_t buffer_size,
         bool map_to_noc = false,
         DeviceBufferAccess device_access = DeviceBufferAccess::READ_WRITE);
+    /**
+     * Constructor for a buffer that was already made visible to the device by the caller.
+     *
+     * @param communication_id Identifier of the device this buffer's IOVA is valid for. Supplied by the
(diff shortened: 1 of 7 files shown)


pjanevskiTT:
Superseded by #3266 - please review there instead.

The commit itself is unchanged - it was re-pushed onto the new base as #3266.

PR #3255, opened 2026-08-25
Give SysmemBuffer ownership of its memory via a Deleter

pjanevskiTT:
### Issue
#3249

### Description
Gives `SysmemBuffer` ownership of its memory through the Base API Specification's `Deleter`, replacing the argument-less `unmap_callback_`. The deleter is composed at construction and receives the page-aligned start of the mapping, so it releases the address the pages were actually pinned at rather than the caller's VA.

This closes a leak. `SiliconSysmemManager::allocate_sysmem_buffer()` mmaps the backing memory and hands the pointer to the buffer, but the buffer only ever unpinned it — the mapping was never munmapped, so every allocated buffer leaked its full size for the life of the process. The allocate path now supplies a deleter that frees it; the map path supplies none, since that memory belongs to the caller.

Stacked on #3253.

### List of the changes
- Added `SysmemBuffer::Deleter` and replaced `unmap_callback_` with a `std::unique_ptr<void, Deleter>` owning the buffer's page-aligned start.
- Composed the deleter in the device-backed constructor as unpin, then release the backing memory if the buffer owns it; `allocate_sysmem_buffer()` now passes a releasing deleter and `map_sysmem_buffer()` does not.
- Reshaped the simulation registry-erase callback to the new deleter signature.
- Added tests that the deleter runs exactly once with the page-aligned start, that a buffer without one destructs cleanly, and that repeated allocate/destroy cycles on silicon do not grow RSS.

### API Changes
This PR has API changes.

- Added the public `SysmemBuffer::Deleter` alias.
- `SysmemBuffer`'s device-backed constructor takes a new optional trailing `release_backing_memory` parameter.
- `SysmemBuffer`'s pre-mapped constructor takes a `Deleter` where it previously took a `std::function<void()>` unmap callback.
- `SysmemBuffer` is no longer copyable, since it now owns its memory. It was already only ever handled through `std::unique_ptr`.

### Testing
CI

### DIFF
--- a/device/api/umd/device/chip_helpers/sysmem_buffer.hpp
+++ b/device/api/umd/device/chip_helpers/sysmem_buffer.hpp
@@ -7,6 +7,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <functional>
+#include <memory>
 #include <optional>
 
 #include "umd/device/types/host_memory.hpp"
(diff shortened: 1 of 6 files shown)


pjanevskiTT:
Superseded by #3267 - please review there instead.

The commit itself is unchanged - it was re-pushed onto the new base as #3267.

PR #3256, opened 2026-08-25
Add the SystemMemoryAllocator interface

pjanevskiTT:
### Issue
#3249

### Description
Adds the `SystemMemoryAllocator` interface from the Base API Specification and makes `SysmemManager` implement it. The interface is pure virtual with no `.cpp`, matching how the other base-layer interfaces in the repo are laid out.

The three methods forward to the existing `allocate_sysmem_buffer()` / `map_sysmem_buffer()` implementations, so nothing changes behaviourally — this only puts the spec's names and shape in front of what is already there. The legacy channel and hugepage surface stays on `SysmemManager`; retiring it is separate work.

One deliberate difference from the spec: `map_user_buffer()` carries the repo's `DeviceBufferAccess` parameter, defaulted to `READ_WRITE`. The spec has no such parameter, but dropping it would lose read-only page pinning (#2680), so the interface is a superset here.

Stacked on #3255.

### List of the changes
- Added `umd/device/chip_helpers/system_memory_allocator.hpp` declaring `allocate_buffer()`, `map_user_buffer()` and `get_communication_id()`.
- Made `SysmemManager` derive from `SystemMemoryAllocator` and implement the three by forwarding to the existing methods.
- Registered the new header in the installed header set.
- Dropped a duplicate `sysmem_buffer.hpp` include in `sysmem_manager.hpp` while adding the new one.
- Added a test driving a manager purely through a `SystemMemoryAllocator*`.

### API Changes
This PR has API changes.

- Added the `SystemMemoryAllocator` interface. `SysmemManager` now derives from it and gains `allocate_buffer()` and `map_user_buffer()` alongside the existing `allocate_sysmem_buffer()` and `map_sysmem_buffer()`.

### Testing
CI

### DIFF
--- a/device/CMakeLists.txt
+++ b/device/CMakeLists.txt
@@ -206,6 +206,7 @@ if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.23)
                 api/umd/device/chip_helpers/sysmem_manager.hpp
                 api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
                 api/umd/device/chip_helpers/silicon_sysmem_manager.hpp
+                api/umd/device/chip_helpers/system_memory_allocator.hpp
                 api/umd/device/chip_helpers/sysmem_buffer.hpp
                 api/umd/device/chip_helpers/tlb_manager.hpp
                 api/umd/device/chip_helpers/simulation_tlb_allocator.hpp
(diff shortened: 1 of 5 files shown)


pjanevskiTT:
Superseded by #3268 - please review there instead.

The commit itself is unchanged - it was re-pushed onto the new base as #3268.

PR #3257, opened 2026-08-25
Add bind_noc_address to SysmemBuffer

pjanevskiTT:
### Issue
#3249

### Description
Adds `SysmemBuffer::bind_noc_address()` from the Base API Specification. A NOC address lets every tile on the device reach the buffer, rather than only the PCIe tile.

The spec models binding as something that can be deferred, with a binder callable that programs the translation on demand. That does not apply here: the driver assigns the NOC address when the pages are pinned, so binding is decided by the `bind_to_noc` argument at allocate or map time. UMD used to program the iATU itself on the pre-2.0 driver, which could have bound after the fact, but that path is gone — #3170 removed `init_pcie_iatus()`, #3171 removed `is_mapping_buffer_to_noc_supported()`, and `KMD_MINIMUM_VERSION` is 2.0.0.

So this is a no-op on a bound buffer and throws on an unbound one. The throw is the point: a buffer that was not pinned with NOC access can never be given a NOC address, and failing at the mistaken assumption is much easier to diagnose than the device access that would otherwise fail later.

Stacked on #3256.

### List of the changes
- Added `SysmemBuffer::bind_noc_address()`.
- Added tests that binding an already-bound buffer is a no-op and is idempotent, and that binding an unbound buffer throws.

### API Changes
This PR has API changes.

- Added `SysmemBuffer::bind_noc_address()`.

### Testing
CI

### DIFF
--- a/device/api/umd/device/chip_helpers/sysmem_buffer.hpp
+++ b/device/api/umd/device/chip_helpers/sysmem_buffer.hpp
@@ -134,6 +134,17 @@ public:
 
     std::optional<uint64_t> get_noc_addr() const { return noc_addr_; }
 
+    /**
+     * Confirms this buffer is bound to a NOC address, so that every tile on the device can reach it rather
+     * than only the PCIe tile.
+     *
+     * The driver assigns the NOC address when the pages are pinned, so binding is decided at construction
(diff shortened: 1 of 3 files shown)


pjanevskiTT:
Superseded by #3269 - please review there instead.

The commit itself is unchanged - it was re-pushed onto the new base as #3269.

PR #3260, opened 2026-08-25
Hold locks in one registry per process

broskoTT:
### Issue
#2745

### Description
LocalChip, TTDevice, ArcMessenger and RemoteCommunication each owned a LockManager, so the same
lock was set up once per owner, and again for every cluster a process opened. The locks are process
wide, so that bought nothing.

An initialized mutex may hold a file descriptor for as long as it lives, which makes the
duplication scale descriptors with the number of owners instead of the number of locks. Harmless
while RobustMutex is the only backend, since it closes its descriptor after mapping, but not once a
lock keeps its handle open.

LockManager now keeps one registry per process, with static entry points like WarmReset and
PCIDevice.

### List of the changes
- Make LockManager's entry points static, backed by one registry per process
- Drop the LockManager members from LocalChip, TTDevice, ArcMessenger and RemoteCommunication
- Guard the registry, since clusters are built from several threads
- Initializing an already initialized lock is no longer warned about, it is the expected outcome
  when more than one object works with the same device

### Testing
umd_misc_tests, plus a lock_virus run against a second process holding ARC_MSG, which it reports
as held.

### API Changes
This PR has API changes, LockManager can no longer be instantiated and its entry points are static.

### Full stack diff


pjanevskiTT (approved):

device/utils/lock_manager.cpp
```diff
@@ -66,28 +96,26 @@ std::optional<std::pair<pid_t, pid_t>> LockManager::probe_mutex(
 }
 
 void LockManager::initialize_mutex_internal(const std::string& mutex_name) {
-    if (mutexes.find(mutex_name) != mutexes.end()) {
-        log_warning(LogUMD, "Mutex already initialized: {}", mutex_name);
+    MutexRegistry& registry = get_registry();
+    std::lock_guard<std::mutex> guard(registry.guard);
```
pjanevskiTT: The registry guard is held across `mutex->initialize()`, and that call is neither cheap nor bounded. `RobustMutex::initialize()` goes through `open_shm_file()` and then `CriticalSectionScopeGuard`, whose constructor does `flock(fd, LOCK_EX)` with no `LOCK_NB` - a blocking inter-process lock - plus `pthread_mutex_lock` on the static `multithread_mutex_`, `ftruncate` and `mmap`.

So one thread setting up a lock holds the process wide guard while waiting on another process. Everything else stalls behind it, because `acquire_mutex` and `probe_mutex` need the same guard just for their lookup, and those are hot paths: `dma_write_to_device`, `read_non_mmio`, `insert_host_to_device_barrier`, every ARC message. A thread acquiring an already initialized, unrelated lock ends up waiting on an flock for a lock it has nothing to do with. Since the PR's rationale is that clusters are built from several threads, that is exactly the workload which hits this.

Not a deadlock, to be clear: `RobustMutex::initialize` never calls back into `LockManager`, so there is no lock order inversion. It is unbounded latency.

Constructing and initializing outside the guard, then taking the guard only to insert and discarding the loser if another thread won the race, avoids it. `RobustMutex::initialize()` is already serialized internally by `multithread_mutex_`, so two threads initializing the same name concurrently is harmless.

device/utils/lock_manager.cpp
```diff
@@ -32,6 +33,35 @@ std::string get_mutex_name(MutexType mutex_type, int device_id, IODeviceType dev
...
+// Deliberately never destroyed: a mutex outliving the registry is far less trouble than one being torn down while
+// another thread still holds it, or after the logger it reports through is gone. The OS reclaims the mappings and
+// file descriptors at exit anyway.
+MutexRegistry& get_registry() {
+    // NOLINTNEXTLINE(cppcoreguidelines-owning-memory)
+    static MutexRegistry* registry = new MutexRegistry();
+    return *registry;
+}
+
+// Looking a mutex up hands out a reference into the registry, which stays valid because entries are only ever added.
```
pjanevskiTT, line 52: This invariant is what makes the whole design safe, and it is worth stating in the public header rather than only here.

For the record it does hold: `mutexes` is a node based `unordered_map`, so insertion invalidates iterators but not references to elements, and the reference handed out is to the `MutexInterface` the `unique_ptr` owns, which never moves regardless. With `clear_mutex` gone and the registry never destroyed, nothing erases, so the `std::unique_lock<MutexInterface>` callers hold cannot dangle.

The concern is discoverability. The header previously warned that "the returned `std::unique_lock<MutexInterface>` should never outlive the LockManager which holds underlying mutexes"; that warning is gone, replaced by prose about file descriptors. So the property a caller now depends on - a handed out lock stays valid for the process lifetime, precisely because entries are never removed - is documented in an anonymous namespace in the .cpp. Anyone later reintroducing a `clear_mutex`, or making the registry destructible, turns every outstanding `unique_lock` into a dangling pointer with nothing in the header to warn them off. One line in the class comment would cover it.

device/utils/lock_manager.cpp
broskoTT: Confirmed and fixed. `CriticalSectionScopeGuard` does `flock(fd_, LOCK_EX)` with no `LOCK_NB`
(`robust_mutex.cpp:63`), so holding the registry guard across `initialize()` meant one thread could
park every other thread's lookup behind an inter-process lock, and lookups sit on the IO paths. The
threaded cluster construction this PR is motivated by is exactly what would hit it.

Done the way you suggest: check under the guard, construct and initialize outside it, then take the
guard again and `try_emplace`. A thread that loses the race drops its own mutex, which is harmless -
the backend serializes first-time setup internally, and the loser's destructor only unmaps its own
mapping.

The "not a deadlock, unbounded latency" distinction was the useful part of this, thanks - it made
clear the fix was about where the guard is held rather than about lock ordering.

device/utils/lock_manager.cpp
broskoTT: Agreed, and you are right that dropping the old "should never outlive the LockManager" warning left
nothing in its place. The class comment now says that locks are only ever added and never removed,
that this is what makes a returned `std::unique_lock<MutexInterface>` safe to hold, and that removing
locks again would leave every outstanding one dangling - so the next person to reach for a
`clear_mutex` has to read it first.

PR #3263, opened 2026-08-25
Make SysmemBuffer constructible only by an allocator

pjanevskiTT:
### Issue
#3249

### Description
Makes `SysmemBuffer`'s constructors private with `SystemMemoryAllocator` as their only friend, per the Base API Specification. Buffers can now only come from an allocator, so every buffer necessarily carries the device id and the cleanup behaviour of whatever pinned it — the two things a hand-built buffer could previously get wrong.

Friendship is not inherited and `std::make_unique` is not a friend either, so `SystemMemoryAllocator` gained a protected `create_buffer()` that forwards to the constructors. Derived allocators build through it. That keeps the friend list to one entry instead of one per concrete manager.

Stacked on #3257.

### List of the changes
- Made both `SysmemBuffer` constructors private and declared `SystemMemoryAllocator` a friend.
- Added the protected `SystemMemoryAllocator::create_buffer()` forwarding helper and routed both concrete managers through it.
- Routed the IOMMU microbenchmark's buffer construction through `map_user_buffer()`, which is the path it was approximating anyway.
- Gave the tests that need a hand-built buffer a minimal test-local allocator rather than weakening the encapsulation.

### API Changes
This PR has API changes.

- `SysmemBuffer`'s constructors are now private. Buffers must be obtained from a `SystemMemoryAllocator` via `allocate_buffer()` or `map_user_buffer()`.

### Testing
CI

### DIFF
--- a/device/api/umd/device/chip_helpers/sysmem_buffer.hpp
+++ b/device/api/umd/device/chip_helpers/sysmem_buffer.hpp
@@ -40,54 +40,6 @@ public:
      */
     using Deleter = std::function<void(void*)>;
 
-    /**
-     * Constructor for SysmemBuffer. Start of the buffer must be aligned
-     * to page size. In case of unaligned buffer start address, the buffer will be aligned to the page size and the
-     * buffer size will be adjusted accordingly. However, the adjusted buffer size won't be visible to the user. It will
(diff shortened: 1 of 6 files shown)


pjanevskiTT:
Superseded by #3270 - please review there instead.

The commit itself is unchanged - it was re-pushed onto the new base as #3270.

PR #3267, opened 2026-08-25
Give SysmemBuffer ownership of its memory via a Deleter

pjanevskiTT:
### Issue
#3249

### Description
Gives `SysmemBuffer` ownership of its memory through the Base API Specification's `Deleter`, replacing the argument-less `unmap_callback_`. The deleter is composed at construction and receives the page-aligned start of the mapping, so it releases the address the pages were actually pinned at rather than the caller's VA.

This closes a leak. `SiliconSysmemManager::allocate_sysmem_buffer()` mmaps the backing memory and hands the pointer to the buffer, but the buffer only ever unpinned it — the mapping was never munmapped, so every allocated buffer leaked its full size for the life of the process. The allocate path now supplies a deleter that frees it; the map path supplies none, since that memory belongs to the caller.

Stacked on #3253.

### List of the changes
- Added `SysmemBuffer::Deleter` and replaced `unmap_callback_` with a `std::unique_ptr<void, Deleter>` owning the buffer's page-aligned start.
- Composed the deleter in the device-backed constructor as unpin, then release the backing memory if the buffer owns it; `allocate_sysmem_buffer()` now passes a releasing deleter and `map_sysmem_buffer()` does not.
- Reshaped the simulation registry-erase callback to the new deleter signature.
- Added tests that the deleter runs exactly once with the page-aligned start, that a buffer without one destructs cleanly, and that repeated allocate/destroy cycles on silicon do not grow RSS.

### API Changes
This PR has API changes.

- Added the public `SysmemBuffer::Deleter` alias.
- `SysmemBuffer`'s device-backed constructor takes a new optional trailing `release_backing_memory` parameter.
- `SysmemBuffer`'s pre-mapped constructor takes a `Deleter` where it previously took a `std::function<void()>` unmap callback.
- `SysmemBuffer` is no longer copyable, since it now owns its memory. It was already only ever handled through `std::unique_ptr`.

### Testing
CI

---
Replaces #3255, which GitHub auto-closed as merged when the stack was reordered (its head branch became an ancestor of its base). Same commit, new base.

### DIFF
--- a/device/api/umd/device/chip_helpers/sysmem_buffer.hpp
+++ b/device/api/umd/device/chip_helpers/sysmem_buffer.hpp
@@ -7,6 +7,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <functional>
+#include <memory>
 #include <optional>
 
 #include "umd/device/types/host_memory.hpp"
@@ -31,6 +32,22 @@ class TTDevice;
 class SysmemBuffer {
 public:
     /**
+     * Cleanup callable invoked on destruction, receiving the page-aligned start of the buffer.
+     *
+     * It encodes the difference between the two ownership models: for a buffer the allocator itself
+     * allocated, it unpins the pages and frees the backing memory; for a buffer mapped from a caller's
+     * pointer, it only unpins, leaving the memory to its owner.
+     *
+     * NOTE: the Base API Specification keeps this alias, and the constructors below, private, with the
+     * allocator a friend of this class. There the allocator is the sole author of the deleter and states
+     * the ownership model outright: it composes unpin-and-free for memory UMD owns, and unpin-only for
+     * memory the client owns. Here both are public and the model is instead implied by whether a
+     * release callable was passed, which is not what we want to keep. Moving to the spec's shape is a
+     * follow-up, together with making the constructors private.
+     */
+    using Deleter = std::function<void(void*)>;
+
+    /**
      * Constructor for SysmemBuffer. Start of the buffer must be aligned
      * to page size. In case of unaligned buffer start address, the buffer will be aligned to the page size and the
      * buffer size will be adjusted accordingly. However, the adjusted buffer size won't be visible to the user. It will
@@ -51,18 +68,26 @@ public:
      * @param buffer_va Pointer to the virtual address of the buffer in the process address space.
      * @param buffer_size Size of the buffer requested by the user.
      * @param map_to_noc If true, the buffer will be mapped to be accessible over NOC from device.
+     * @param release_backing_memory Optional callable that frees the pages behind buffer_va, invoked after
+     * the buffer has been unpinned from the device. Pass it when the allocator owns the memory it is handing
+     * over; leave it empty when the caller owns the memory and only wants it unpinned on destruction.
+     * Signalling ownership through the presence of this parameter is the interim shape described in the
+     * note on Deleter, not the one the specification settles on.
      */
     SysmemBuffer(
         TTDevice* tt_device,
         void* buffer_va,
         size_t buffer_size,
         bool map_to_noc = false,
-        DeviceBufferAccess device_access = DeviceBufferAccess::READ_WRITE);
+        DeviceBufferAccess device_access = DeviceBufferAccess::READ_WRITE,
+        Deleter release_backing_memory = {});
     /**
      * Constructor for a buffer that was already made visible to the device by the caller.
      *
      * @param communication_id Identifier of the device this buffer's IOVA is valid for. Supplied by the
      * allocator, which pins for exactly one device. Matches TTDevice::get_communication_device_id().
+     * @param deleter Cleanup callable invoked on destruction with the page-aligned start of the buffer.
+     * Defaults to releasing nothing, since this constructor takes a mapping the caller already owns.
      */
     SysmemBuffer(
         void* buffer_va,
@@ -70,7 +95,7 @@ public:
         uint64_t device_io_addr,
         int communication_id,
         std::optional<uint64_t> noc_addr = std::nullopt,
-        std::function<void()> unmap_callback = {},
+        Deleter deleter = {},
         DeviceBufferAccess device_access = DeviceBufferAccess::READ_WRITE);
     ~SysmemBuffer();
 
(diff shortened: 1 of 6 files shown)


device/api/umd/device/chip_helpers/sysmem_buffer.hpp
```diff
@@ -30,6 +31,15 @@ class TTDevice;
...
 class SysmemBuffer {
 public:
+    /**
+     * Cleanup callable invoked on destruction, receiving the page-aligned start of the buffer.
+     *
+     * It encodes the difference between the two ownership models: for a buffer the allocator itself
+     * allocated, it unpins the pages and frees the backing memory; for a buffer mapped from a caller's
+     * pointer, it only unpins, leaving the memory to its owner.
+     */
+    using Deleter = std::function<void(void*)>;
```
nbuncicTT, line 48: In the spec this is private, I get that it isn't now because it's an incremental change, but please note this in a comment.

device/chip_helpers/sysmem_buffer.cpp
```diff
@@ -99,19 +134,8 @@ void SysmemBuffer::dma_read_from_device(const size_t offset, size_t size, const
...
-        return;
-    }
-    try {
-        pci_device_->unmap_for_dma(buffer_va_, mapped_buffer_size_);
-    } catch (...) {
-        log_warning(
-            LogUMD, "Failed to unmap sysmem buffer (size: {:#x}, IOVA: {:#x}).", mapped_buffer_size_, device_io_addr_);
-    }
+    // Runs the deleter composed at construction: unpin, and free the backing memory if this buffer owns it.
+    system_memory_ptr_.reset();
```
nbuncicTT: We don't need `.reset()` explicitly, the `unique_ptr` invokes the deleter for us.

device/api/umd/device/chip_helpers/sysmem_buffer.hpp
```diff
@@ -51,26 +61,32 @@ class SysmemBuffer {
...
+     * over; leave it empty when the caller owns the memory and only wants it unpinned on destruction.
      */
     SysmemBuffer(
         TTDevice* tt_device,
         void* buffer_va,
         size_t buffer_size,
         bool map_to_noc = false,
-        DeviceBufferAccess device_access = DeviceBufferAccess::READ_WRITE);
+        DeviceBufferAccess device_access = DeviceBufferAccess::READ_WRITE,
+        Deleter release_backing_memory = {});
```
nbuncicTT, line 83: There's another reason which is why this is private in the spec: It's because we don't want this deallocation/release of backing memory to be the only reason to have a deleter. We want the deleter to be fully defined by the allocator (that's why it's a friend class of the buffer and the buffer has private ctors). 

What this means is:
- If the client wants UMD to own the buffer, the allocator (not the client) sets the deleter which releases the memory and does the unpinning
- If the client want to own the buffer, but wants it mapped, the allocator sets the deleter which only does the unpinning - this means it's up to the client to do the memory management

And we are achieving this implicitly by having or not having the deleter depending on the case, and we don't want this.

Again if it's a temporary solution, I'm fine with it, just add a comment or have it in mind when going forward (no action for now).

nbuncicTT (approved):

tests/api/test_sysmem_manager.cpp
```diff
@@ -227,6 +228,62 @@ TEST(ApiSysmemManager, SysmemBufferFunctions) {
...
+    const size_t rss_before = read_rss_kib();
+    if (rss_before == 0) {
+        GTEST_SKIP() << "Could not read VmRSS from /proc/self/status.";
+    }
+
+    for (int i = 0; i < iterations; i++) {
+        std::unique_ptr<SysmemBuffer> buffer = sysmem_manager->allocate_sysmem_buffer(buf_size);
+        ASSERT_NE(buffer, nullptr);
+        // Touch the buffer so the pages are definitely resident.
+        std::memset(buffer->get_buffer_va(), i, 4096);
```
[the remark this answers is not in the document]
pjanevskiTT: Correct — with only the first page touched, a leaked mapping would show up as roughly one page per iteration and sit well under the threshold, so the assertion proved nothing. The loop now writes one byte per page across the whole 64 MiB buffer, so a retained mapping registers at full size.

device/chip_helpers/silicon_sysmem_manager.cpp
```diff
@@ -447,7 +447,25 @@ std::unique_ptr<SysmemBuffer> SiliconSysmemManager::allocate_sysmem_buffer(
             error::RuntimeError,
             fmt::format("Failed to allocate sysmem buffer of size {:#x} bytes with mmap.", sysmem_buffer_size));
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed. `allocate_sysmem_buffer()` now wraps the construction in try/catch and releases the mapping before rethrowing, so a failure to pin no longer leaks it.

Went with try/catch rather than a `unique_ptr` guard because clang-tidy's `bugprone-unused-return-value` rejects the discarded `release()` and the build runs with `-Werror`.

device/chip_helpers/sysmem_buffer.cpp
```diff
@@ -41,6 +59,23 @@ SysmemBuffer::SysmemBuffer(
...
+         iova = device_io_addr_,
+         release = std::move(release_backing_memory)](void* aligned_va) {
+            try {
+                pci_device->unmap_for_dma(aligned_va, mapped_size);
+            } catch (...) {
+                log_warning(LogUMD, "Failed to unmap sysmem buffer (size: {:#x}, IOVA: {:#x}).", mapped_size, iova);
+            }
+            if (release) {
+                release(aligned_va);
+            }
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed at both sites, and this also covers the destructor finding that was suppressed in the review body. `or_no_op` is now `make_non_throwing`, which handles the empty-function case and catches anything the wrapped deleter throws, logging instead of letting it escape; both constructors route their deleter through it. Added `ThrowingBufferDeleterDoesNotEscapeDestruction` to hold that.

device/api/umd/device/chip_helpers/sysmem_buffer.hpp
pjanevskiTT: Noted in a `NOTE:` block on the alias — it records that the spec keeps this and the constructors private with the allocator as a friend, and that moving to that shape is follow-up work.

device/chip_helpers/sysmem_buffer.cpp
pjanevskiTT: Removed. The destructor now leaves it to `system_memory_ptr_`'s own destruction, with the comment saying as much.

device/api/umd/device/chip_helpers/sysmem_buffer.hpp
pjanevskiTT: Folded into the same `NOTE:` on `Deleter`, since it is the same underlying point. It now spells out that in the spec the allocator is the sole author of the deleter and states the ownership model outright — unpin-and-free for memory UMD owns, unpin-only for memory the client owns — rather than having it implied by whether a release callable was passed, which is what this step does. The `release_backing_memory` param doc points at that note.

No behavioural change here, as you said.

PR #3268, opened 2026-08-25
Add the SystemMemoryAllocator interface

pjanevskiTT:
### Issue
#3249

### Description
Adds the `SystemMemoryAllocator` interface from the Base API Specification and makes `SysmemManager` implement it. The interface is pure virtual with no `.cpp`, matching how the other base-layer interfaces in the repo are laid out.

The three methods forward to the existing `allocate_sysmem_buffer()` / `map_sysmem_buffer()` implementations, so nothing changes behaviourally — this only puts the spec's names and shape in front of what is already there. The legacy channel and hugepage surface stays on `SysmemManager`; retiring it is separate work.

`map_user_buffer()` carries device access: it takes a trailing `DeviceBufferAccess device_access = DeviceBufferAccess::READ_WRITE`, so read-only page pinning (#2680) is reachable through the interface rather than only through `map_sysmem_buffer()` on the concrete manager. The default keeps every call site source-compatible.

Stacked on #3255.

### List of the changes
- Added `umd/device/chip_helpers/system_memory_allocator.hpp` declaring `allocate_buffer()`, `map_user_buffer()` and `get_communication_id()`.
- Made `SysmemManager` derive from `SystemMemoryAllocator` and implement the three by forwarding to the existing methods.
- Registered the new header in the installed header set.
- Dropped a duplicate `sysmem_buffer.hpp` include in `sysmem_manager.hpp` while adding the new one.
- Added a test driving a manager purely through a `SystemMemoryAllocator*`, including a read-only mapping made through the interface.

### API Changes
This PR has API changes.

- Added the `SystemMemoryAllocator` interface. `SysmemManager` now derives from it and gains `allocate_buffer()` and `map_user_buffer()` alongside the existing `allocate_sysmem_buffer()` and `map_sysmem_buffer()`.

### Testing
CI

### DIFF
--- a/device/chip_helpers/sysmem_manager.cpp
+++ b/device/chip_helpers/sysmem_manager.cpp
@@ -77,6 +77,15 @@ void SysmemManager::read_from_sysmem(uint16_t channel, void *dest, uint64_t sysm
     memcpy(dest, user_scratchspace, size);
 }
 
+std::unique_ptr<SysmemBuffer> SysmemManager::allocate_buffer(size_t size, bool bind_to_noc) {
+    return allocate_sysmem_buffer(size, bind_to_noc);
+}
+
+std::unique_ptr<SysmemBuffer> SysmemManager::map_user_buffer(
+    void *user_ptr, size_t size, bool bind_to_noc, DeviceBufferAccess device_access) {
+    return map_sysmem_buffer(user_ptr, size, bind_to_noc, device_access);
+}
+
 uint64_t SysmemManager::get_pcie_base_for_arch(tt::ARCH arch) {
     switch (arch) {
         case tt::ARCH::WORMHOLE_B0:
--- a/device/CMakeLists.txt
+++ b/device/CMakeLists.txt
@@ -217,6 +217,7 @@ if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.23)
                 api/umd/device/chip_helpers/sysmem_manager.hpp
                 api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
                 api/umd/device/chip_helpers/silicon_sysmem_manager.hpp
+                api/umd/device/chip_helpers/system_memory_allocator.hpp
                 api/umd/device/chip_helpers/sysmem_buffer.hpp
                 api/umd/device/chip_helpers/tlb_manager.hpp
                 api/umd/device/chip_helpers/simulation_tlb_allocator.hpp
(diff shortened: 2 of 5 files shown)


device/chip_helpers/sysmem_manager.cpp
```diff
@@ -77,6 +77,16 @@ void SysmemManager::read_from_sysmem(uint16_t channel, void *dest, uint64_t sysm
...
 }
 
+std::unique_ptr<SysmemBuffer> SysmemManager::allocate_buffer(size_t size, bool bind_to_noc) {
+    return allocate_sysmem_buffer(size, bind_to_noc);
+}
+
+std::unique_ptr<SysmemBuffer> SysmemManager::map_user_buffer(void *user_ptr, size_t size, bool bind_to_noc) {
+    // The spec's interface has no notion of device access. Callers needing a device-read-only mapping use
+    // map_sysmem_buffer() on the concrete manager.
+    return map_sysmem_buffer(user_ptr, size, bind_to_noc, DeviceBufferAccess::READ_WRITE);
```
nbuncicTT: We must have the `DeviceBufferAccess` in `SystemMemoryAllocator` interface, but we mustn't make a breaking change, hence create another API which overloads the `map_user_buffer` with another parameter which has a default value  `DeviceBufferAccess::READ_WRITE`. e.g.
```cpp
virtual std::unique_ptr<SysmemBuffer> map_user_buffer(void* user_ptr, size_t size, bool bind_to_noc = false, DeviceBufferAccess device_access = DeviceBufferAccess::READ_WRITE) = 0;
```

nbuncicTT (approved):

device/chip_helpers/sysmem_manager.cpp
nbuncicTT: And of course implement it in sysmem_manager..

pjanevskiTT: Done, both parts. `SystemMemoryAllocator::map_user_buffer()` now takes a trailing `DeviceBufferAccess device_access = DeviceBufferAccess::READ_WRITE`, and `SysmemManager` implements it by forwarding straight through to `map_sysmem_buffer()`. The default keeps every existing call site source-compatible.

Extended the interface test to map a buffer `READ_ONLY` through a `SystemMemoryAllocator*` and assert `get_device_access()` comes back `READ_ONLY`. Also corrected the PR description, which claimed device access was not expressible through the interface.

PR #3269, opened 2026-08-25
Add bind_noc_address to SysmemBuffer

pjanevskiTT:
### Issue
#3249

### Description
Adds `bind_noc_address()` and the `NocBinder` it uses to `SysmemBuffer`, continuing the alignment of the sysmem classes with the Base API Specification.

No allocator supplies a binder today: the KMD assigns a NOC address only when pages are pinned, and the pre-2.0 iATU path that could have bound after the fact is gone (#3170, #3171). The binder is injected rather than built in, so an allocator that can bind another way is free to supply one.

### List of the changes
- Added `SysmemBuffer::bind_noc_address()`: runs the injected binder once and caches the result, a no-op once bound, and throws when the buffer is unbound and has no binder.
- Added `SysmemBuffer::NocBinder` and an optional trailing constructor parameter for it.
- Added tests for the bound, unbound, and injected-binder cases.

### API Changes
This PR has API changes.

- Added `SysmemBuffer::bind_noc_address()` and `SysmemBuffer::NocBinder`.
- `SysmemBuffer`'s pre-mapped constructor takes a new optional trailing `noc_binder` parameter.

### Testing
CI

### Full stack diff
https://github.com/tenstorrent/tt-umd/compare/main...pjanevski/sysmem-remove-deprecated-names

### DIFF
--- a/tests/api/test_simulation_sysmem_manager.cpp
+++ b/tests/api/test_simulation_sysmem_manager.cpp
@@ -2,6 +2,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+#include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
 #include <algorithm>
@@ -23,6 +24,10 @@
 #include "umd/device/types/host_memory.hpp"
 
 using namespace tt::umd;
+
+using ::testing::InSequence;
+using ::testing::MockFunction;
+using ::testing::Return;
 
 const uint32_t HUGEPAGE_REGION_SIZE = 1ULL << 30;  // 1GB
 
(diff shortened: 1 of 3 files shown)


tests/api/test_simulation_sysmem_manager.cpp
```diff
@@ -134,6 +134,93 @@ TEST(ApiSimulationSysmemManager, BufferDeleterRunsOnceWithAlignedStart) {
     EXPECT_EQ(deleter_saw, aligned_start);
 }
 
+// The driver assigns the NOC address at pin time, so bind_noc_address() only confirms what already
+// happened. On a bound buffer it is a no-op.
+TEST(ApiSimulationSysmemManager, BindNocAddressIsNoOpWhenAlreadyBound) {
```
nbuncicTT, line 146: The PR content is great in general, but I think the test is reinventing the wheel with manually counting the number of invocations by putting a value in a lambda. So from my point of view, I think it would be better to use the gmock framework for counting the number of invocations of a function. If you think it's a big hassle to do this now - state in the comment with a TODO as a future task (use gmock framework on this test suite).

nbuncicTT (approved):

tests/api/test_simulation_sysmem_manager.cpp
pjanevskiTT: Converted to gmock rather than leaving a TODO — gmock is already linked into `test_common`, so it was cheap.

`MockFunction<void(void*)>` for the deleter and `MockFunction<uint64_t()>` for the binder, with the counts expressed as expectations instead of hand-rolled counters: `EXPECT_CALL(deleter, Call(aligned_start))` covers both the cardinality and the argument, `WillOnce(Return(...))` makes a second binder call fail on its own, and the already-bound case is `Times(0)`. The 'not run before destruction' part of the deleter test is a sequenced checkpoint mock, since plain `Times(1)` would not catch it running at construction.

Fixed in 85310b96.

Issue #2875, opened 2026-06-22
Migrate IoWindow interface to Base API Specification


nbuncicTT:
**Description**

The current codebase has two separate classes (`TlbHandle` and `TlbWindow`) that together cover what the Base API Specification defines as `IoWindow`. Merge them into a single `IoWindow` interface per the spec. Methods not in the spec remain in concrete implementations as internal behavior.

**Scope**

- Create `IoWindow` interface per the Base API Specification
- Merge `TlbHandle` and `TlbWindow` into `IoWindow`
- Move methods not in the spec into concrete implementations
- Update all consumers

**Timeline**

No hard deadline. Part of the UMD Base API migration epic.

**Is your feature request related to a problem? Please describe.**

TLB windowing is split across two classes with implementation details exposed in the public interface.

**Definition of Done**

- `IoWindow` interface exists per the Base API Specification
- `TlbHandle` and `TlbWindow` are merged into `IoWindow`
- Implementation-specific methods are internal to concrete implementations
- All consumers updated
- Existing tests pass or are updated

broskoTT:
pending PR reviews, work in progress, working on ttdevice modifications

PR #3271, opened 2026-08-25
[Base API] Implement IoWindow interface in TlbWindow

ajovanovicTT:
### Issue
Part of #2875

### Description
TlbWindow now implements the Base API IoWindow interface. Additive only: no consumer changes, no behavior change, nothing removed or renamed.

### List of the changes
 - Added IoOrdering enum in io_window_config.hpp, values matching the tlb_data ordering encoding
 - Added configure with an explicit ordering argument, plus get_io_ordering, to the IoWindow interface
 - TlbWindow derives from IoWindow and implements write_aligned, read_aligned, both configure overloads, get_target_config, get_io_ordering, get_memory_caching_type
 - Added a using declaration in SiliconTlbWindow so its configure override does not hide the new overloads

### Testing
 - Build and new test run locally
 - CI

### API Changes
This PR has API changes.

### DIFF
--- a/device/api/umd/device/io_window/io_window.hpp
+++ b/device/api/umd/device/io_window/io_window.hpp
@@ -96,7 +96,7 @@ public:
      * Updates the underlying hardware mapping so that subsequent read/write operations
      * target the new device address.
      *
-     * Applies IoOrdering::Strict. Callers that need another ordering mode must use the
+     * By default ordering is IoOrdering::Strict. Callers that need another ordering mode must use the
      * two-argument overload below.
      *
      * @param config Device-side target describing the core, address, and optional NOC.
@@ -104,7 +104,7 @@ public:
     virtual void configure(const TargetIoWindowConfig& config) = 0;
 
     /**
-     * @brief Reconfigures the window, selecting the transaction ordering mode explicitly.
+     * @brief Configures the window with the transaction ordering mode explicitly set.
      *
      * @param config Device-side target describing the core, address, and optional NOC.
      * @param ordering Transaction ordering mode to apply to the new mapping.
--- a/device/api/umd/device/io_window/io_window.hpp
+++ b/device/api/umd/device/io_window/io_window.hpp
@@ -96,15 +96,32 @@ public:
      * Updates the underlying hardware mapping so that subsequent read/write operations
      * target the new device address.
      *
+     * Applies IoOrdering::Strict. Callers that need another ordering mode must use the
+     * two-argument overload below.
+     *
      * @param config Device-side target describing the core, address, and optional NOC.
      */
     virtual void configure(const TargetIoWindowConfig& config) = 0;
+
+    /**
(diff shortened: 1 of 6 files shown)


device/api/umd/device/io_window/io_window.hpp
```diff
@@ -96,16 +96,33 @@ class IoWindow {
      * Updates the underlying hardware mapping so that subsequent read/write operations
      * target the new device address.
      *
+     * Applies IoOrdering::Strict. Callers that need another ordering mode must use the
```
nbuncicTT: nit: state by default the ordering is strict.

device/api/umd/device/io_window/io_window.hpp
```diff
@@ -96,16 +96,33 @@ class IoWindow {
...
+     * two-argument overload below.
+     *
      * @param config Device-side target describing the core, address, and optional NOC.
      */
     virtual void configure(const TargetIoWindowConfig& config) = 0;
 
+    /**
+     * @brief Reconfigures the window, selecting the transaction ordering mode explicitly.
```
nbuncicTT: nit: Configures the window, not reconfigures. Also this can sound nicer: 
```cpp
* @brief Configures the window with the transaction ordering mode explicitly set.
```

device/pcie/tlb_window.cpp
```diff
@@ -121,6 +134,52 @@ void TlbWindow::configure(const tlb_data& new_config) {
...
+    UMD_ASSERT(
+        config.flags == WindowFlags::None,
+        error::RuntimeError,
+        "WindowFlags are not supported by TLB-backed IoWindows.");
+
+    const bool mcast = config.core_end.has_value();
+    configure(make_tlb_config(
+        config.addr,
+        mcast ? config.core_end.value() : config.core_start,
+        config.noc.value_or(NocId::NOC0),
```
nbuncicTT: Ok, so to just clear things up a bit, there *can* be memory mapped peripherals which we own which won't be reachable over the NOC but exposed in BAR space in some IP configurations, hence a user doesn't have to define the NOC - i.e. it can be left `std::nullopt`.

TLBs most certainly don't have this option, they have to have a NOC and silently regressing to NOC0 is not what we want, hence, use the `UMD_ASSERT` to throw if somebody passes to `TlbWindow` `std::nullopt` on this concrete class.

nbuncicTT (approved):

device/pcie/tlb_window.cpp
ajovanovicTT: Done.

device/api/umd/device/io_window/io_window.hpp
ajovanovicTT: Done.

device/api/umd/device/io_window/io_window.hpp
ajovanovicTT: Done.

PR #3273, opened 2026-08-26
[Base API] Move firmware startup from TTDevice to DeviceFirmware

nbuncicTT:
## Issue

`DeviceFirmware` and its per-arch implementations exist, but nothing constructs them — `TTDevice` still owns firmware startup through `wait_arc_core_start`.

## Description

Builds a `DeviceFirmware` and moves one thing onto it: waiting for the firmware to come up. `wait_arc_core_start(timeout)` becomes `get_device_firmware()->init_firmware(timeout, noc_id)` at both call sites, and the per-arch implementations are deleted.

Each concrete device builds its own firmware from its own constructor, with the call explicitly qualified. That matters: `create_device_firmware()` is virtual, and calling it from `TTDevice`'s constructor resolves to the base — which handed simulation devices a silicon firmware built on a model with no protocol, and segfaulted ttsim.

`init_firmware()` fills the telemetry and info-provider slots `TTDevice` owns, passed as `std::unique_ptr<T>&` so ownership never moves. This replaces the raw-pointer form on `main`: those components cannot exist before the firmware is up, and the firmware is what brings it up. Temporary — once `wait_firmware_ready()` is its own API they can be built first and passed as plain pointers.

**Scope is startup only.** `TTDevice` still answers `get_chip_info`, `get_noc_translation_enabled`, `set_power_state`, `set_clock_state` and ARC messaging itself; those move in #3289 onward. `init_firmware()` is guarded so a second call is a no-op, because warm reset only wants the wait.

## List of the changes

- `build_device_firmware()` / `create_device_firmware()`; construction from each most-derived constructor.
- Simulation backends get their own `DeviceFirmware` rather than a silicon one.
- `wait_arc_core_start` removed; the wait split into `wait_firmware_ready()` in all five implementations.
- Firmware constructors validate pointers before dereferencing them.
- `FirmwareStartupError` constructors defined.

## Testing

Simulation on and off; `baremetal_tests` 254/254; no undefined symbols. ttsim covers the simulation construction path.

## API Changes

`TTDevice::wait_arc_core_start` removed — use `get_device_firmware()->init_firmware(...)`. `get_device_firmware()` added.


device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -591,6 +586,31 @@ class TTDevice {
...
+    /**
+     * @brief Takes ownership of the firmware this device runs on.
+     *
+     * Called from the most derived constructor, with the result of an explicitly qualified
+     * create_device_firmware(). Both halves matter: create_device_firmware() is virtual, so calling
+     * it from TTDevice's own constructor would resolve to TTDevice's implementation rather than the
+     * device's, and qualifying it at the call site says which one runs without the reader having to
+     * know that.
+     */
+    void build_device_firmware(std::unique_ptr<DeviceFirmware> firmware);
```
broskoTT, line 598: I assume this goes away once we move it to tt device model

device/tt_device/tt_device.cpp
```diff
@@ -75,6 +77,41 @@ TTDevice::TTDevice(std::unique_ptr<TTDeviceModel> model) : model_(std::move(mode
     }
 }
 
+void TTDevice::build_device_firmware(std::unique_ptr<DeviceFirmware> firmware) {
+    device_firmware_ = std::move(firmware);
+}
+
+std::unique_ptr<DeviceFirmware> TTDevice::create_device_firmware() {
+    // The model owns the transport now, so the firmware is built from its interfaces rather than
```
broskoTT, line 85: also hope this is temporary

broskoTT (approved):

nbuncicTT:
Closing in favor of #3317 / #3318 / #3319 / #3320 — the same moves rebuilt function-by-function, each PR deleting the TTDevice code it replaces.

PR #3275, opened 2026-08-26
[Base API] Extract *_reconfigure to free functions over IoWindow&

ajovanovicTT:
### Issue

### Description
Extracts the *_reconfigure family (read_block, write_block, read_register, write_register, noc_multicast_write) out of TlbWindow into free functions expressed purely against the IoWindow spec surface. PcieProtocol's non-safe path, LocalChip, the TTDevice hang probe, and SimulationTTDevice only needed this family, not the rest of TlbWindow — they now depend on IoWindow instead.

### List of the changes
 - New device/pcie/io_window_reconfigure.{hpp,cpp}: the 5 free functions + the chunking loop, built on configure(TargetIoWindowConfig, IoOrdering) + read_block/write_block/read_aligned/write_aligned/get_size.
 - TlbWindow keeps thin forwarding members (unchanged behavior) so SiliconTlbWindow's safe_* variants keep compiling; resolving safe_* is separate follow-up work.
 - Updated the 4 non-safe call sites in PcieProtocol, LocalChip, TTDevice, SimulationTTDevice to call the free functions directly.
 - Per transfer traffic direction is now a two bit field in WindowFlags (bits 2 and 3, zero meaning bidirectional), so it reaches the window through the spec surface and TlbVcDirection is no longer needed. Without this the helpers would configure every mapping as bidirectional and drop the static VC pinning added in #3067.

### Testing
No behavior change - the helpers set the same direction the TlbWindow members did, so the static VC settings are identical to main on every path. Existing TLB/window tests cover this path.

### API Changes
This PR has API changes.
 - WindowFlags gains Bidirectional, UnicastWrite, UnicastRead, MulticastWrite and DirectionMask.
 - TlbVcDirection is removed; ArchitectureTlbs::get_static_vc takes WindowFlags instead.

### DIFF
--- a/device/pcie/io_window_reconfigure.cpp
+++ b/device/pcie/io_window_reconfigure.cpp
@@ -1,0 +1,110 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "pcie/io_window_reconfigure.hpp"
+
+#include <algorithm>
+#include <optional>
+
+namespace tt::umd {
+
+namespace {
+
+// Each chunk is configured immediately before it is transferred, so the mapping only ever carries one
+// direction of traffic and says so. The direction follows from the transfer rather than from the caller.
+template <typename buffer_pointer, typename io_operation>
+void transfer_and_reconfigure(
+    IoWindow& window,
+    TargetIoWindowConfig config,
+    IoOrdering ordering,
+    buffer_pointer buffer,
+    size_t size,
+    io_operation op) {
+    while (size > 0) {
+        window.configure(config, ordering);
+        size_t transfer_size = std::min(size, window.get_size());
+        op(buffer, transfer_size);
+        size -= transfer_size;
+        config.addr += transfer_size;
+        buffer += transfer_size;
+    }
+}
+
+}  // namespace
+
+void read_block_reconfigure(
+    IoWindow& window, void* mem_ptr, tt_xy_pair core, uint64_t addr, size_t size, NocId noc_id, IoOrdering ordering) {
+    transfer_and_reconfigure(
+        window,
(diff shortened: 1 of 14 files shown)


device/pcie/io_window_reconfigure.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+#include <algorithm>
+#include <optional>
+
+namespace tt::umd {
+
+namespace {
+
+template <typename buffer_pointer, typename io_operation>
+void transfer_and_reconfigure(
```
nbuncicTT, line 17: Why is this in an anonymous namespace?

nbuncicTT (approved):

device/pcie/io_window_reconfigure.cpp
ajovanovicTT: Internal helper only used by the free functions in this file, not declared in the header so anonymous namespace gives it internal linkage so it can't collide with or leak to other translation units.

device/pcie/tlb_window.cpp
```diff
@@ -54,52 +53,28 @@ tlb_data TlbWindow::make_tlb_config(
...
+// keep working unchanged. Callers with no safe/unsafe distinction to make should call the free
+// functions directly instead of going through a TlbWindow.
 void TlbWindow::read_block_reconfigure(
     void* mem_ptr, tt_xy_pair core, uint64_t addr, size_t size, NocId noc_id, uint64_t ordering) {
-    transfer_and_reconfigure(
-        make_tlb_config(addr, core, noc_id, ordering),
-        static_cast<uint8_t*>(mem_ptr),
-        size,
-        [this](uint8_t* buf, size_t sz) { read_block(0, buf, sz); });
+    tt::umd::read_block_reconfigure(*this, mem_ptr, core, addr, size, noc_id, static_cast<IoOrdering>(ordering));
```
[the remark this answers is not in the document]
ajovanovicTT: Applied suggestion.

PR #3280, opened 2026-08-26
[Base API] Resolve safe_* as window-level policy

ajovanovicTT:
### Issue

### Description
use_safe_api_ is PcieProtocol-lifetime state, not a per-call choice, and the hang check is already installed per-window. This makes safety the same kind of window-level policy, collapsing the mirrored safe_*/*_reconfigure method family and the if constexpr(safe) dispatch into direct calls through the free *_reconfigure functions.

### List of the changes
 - TlbWindow gains an IoSafety constructor parameter (defaults to Disabled, fixed for the window's lifetime); its safe_* and *_reconfigure member forwarders are removed.
 - SiliconTlbWindow routes each memory-access op through execute_safe() only when IoSafety::Enabled; sim windows drop their trivial safe_write16/safe_read16 overrides.
 - PcieProtocol constructs the cached window with IoSafety derived from use_safe_api_ and calls the free reconfigure functions unconditionally, removing the write_data_impl<bool safe>/read_data_impl<bool safe> templates.

### Testing
 - CI
 - Local runs of adjusted warm reset tests from tests/api/test_warm_reset.cpp 

### API Changes
- removing the safe_*/*_reconfigure member methods from TlbWindow
- new IoSafety enum, taken as a defaulted TlbWindow/SiliconTlbWindow constructor parameter

### DIFF
--- a/device/api/umd/device/pcie/tlb_window.hpp
+++ b/device/api/umd/device/pcie/tlb_window.hpp
@@ -149,7 +149,7 @@ protected:
         tt_xy_pair core_end,
         NocId noc_id,
         uint64_t ordering,
-        TlbVcDirection direction,
+        WindowFlags flags,
         bool mcast = false,
         tt_xy_pair core_start = {}) const;
 
--- a/device/api/umd/device/pcie/tlb_window.hpp
+++ b/device/api/umd/device/pcie/tlb_window.hpp
@@ -20,13 +20,28 @@
 namespace tt::umd {
 
 /**
+ * Safe-I/O policy of a window, fixed for its lifetime.
+ *
+ * With Enabled, the window's memory-access methods recover from a SIGBUS raised by a failing device
+ * access and throw SigbusError instead of letting the fault take the process down. Only
+ * SiliconTlbWindow can offer this, since simulation windows never touch mapped device memory; where it
+ * cannot be offered the policy is ignored.
+ *
+ * The underlying type is bool so that a caller still holding a flag can cast it directly.
+ */
+enum class IoSafety : bool {
+    Disabled = false,
+    Enabled = true,
+};
+
+/**
  * Base class for TlbWindow implementations that contains all shared logic.
  * The memory access methods are pure virtual to allow different implementations
  * for silicon (direct memory access) vs simulation (communicator-based access).
  */
 class TlbWindow : public IoWindow {
 public:
-    TlbWindow(std::unique_ptr<TlbHandle> handle, const tlb_data config = {});
+    TlbWindow(std::unique_ptr<TlbHandle> handle, const tlb_data config = {}, IoSafety io_safety = IoSafety::Disabled);
 
     virtual ~TlbWindow() = default;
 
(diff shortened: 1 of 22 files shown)


device/api/umd/device/pcie/tlb_window.hpp
```diff
@@ -49,91 +49,17 @@ class TlbWindow : public IoWindow {
...
     // Installs a per-op MMIO timeout hang check used by the timed memcpy path. No-op by default; only
     // SiliconTlbWindow consults it (simulation windows do not run the timed path). See SiliconTlbWindow.
     virtual void set_io_timeout_hang_check(const std::function<bool(NocId)>& hang_check) {}
 
+    // Installs safe-I/O as window-level policy: once enabled, the memory-access methods above recover
+    // from SIGBUS by throwing SigbusError instead of crashing the process. No-op by default; only
+    // SiliconTlbWindow can actually offer this (simulation windows never touch mapped device memory).
+    // Callers needing chunked transfers use the free functions in io_window_reconfigure.hpp, which honor
+    // whatever policy is installed here since they operate through this window's ops.
+    virtual void set_safe_io(bool enable) {}
```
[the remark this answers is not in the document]
ajovanovicTT: Changed PR desc

device/pcie/silicon_tlb_window.cpp
```diff
@@ -99,27 +99,103 @@ void SiliconTlbWindow::update_io_timeout_callback() {
     io_timeout_callback_ = [hang_check, noc]() -> bool { return !hang_check(noc); };
 }
 
+void SiliconTlbWindow::set_safe_io(bool enable) { safe_io_ = enable; }
+
+template <typename Func, typename... Args>
```
nbuncicTT: Why are we making a diff by reshuffling the code in the PR? Please make only the diff visible and I'll do a review on the PR then.

ajovanovicTT: Restructured changes to show minimal diff, behavior remained unchanged.

device/pcie/silicon_tlb_window.cpp
```diff
@@ -287,64 +289,66 @@ decltype(auto) SiliconTlbWindow::execute_safe(Func &&func, Args &&...args) {
...
+    }
+    return read16_impl(offset);
 }
 
-void SiliconTlbWindow::safe_read_block(uint64_t offset, void *data, size_t size) {
-    execute_safe(&SiliconTlbWindow::read_block, offset, data, size);
+void SiliconTlbWindow::write32(uint64_t offset, uint32_t value) {
+    if (safe_io_) {
+        execute_safe(&SiliconTlbWindow::write32_impl, offset, value);
+    } else {
```
nbuncicTT: nit: Follow the pattern from other impl. where you early return, you can do so in functions with the void return value by stating `return;` in the if branch.

Do so in other APIs you're touching.

device/api/umd/device/pcie/tlb_window.hpp
nbuncicTT: Instead of making this a setter, why isn't this a constructor parameter? We won't change this after the window inception and also we don't have a getter.

Proposal: Make a new, overloaded ctor for TlbWindow which accepts this as a parameter, by default it's false. 

Also I dislike booleans as parameters as general and even here I prefer an enum class - this can be done here elegantly and we cast the boolean from TTDevice for now.

And lastly, this removes the setter.

nbuncicTT (approved):

Issue #3283, opened 2026-08-27
Add RDMA dma-buf tests: single-host loopback (unit test) + cross-host benchmark


ajovanovicTT:
**Description**
UMD's Cluster::export_dmabuf() has no test coverage that actually exercises real RDMA traffic against the exported dma-buf. We need two levels: a fast single-host correctness check runnable in CI, and a cross-host bandwidth benchmark for manual/lab use. Right now neither exists, so a regression in the export path (kmd version gating, TLB window export, dma-buf lifetime) would only surface on real multi-host hardware, if at all.

**Scope**
 - Single-host RDMA loopback test as a gtest in tests/unified/test_tlb.cpp: seed DRAM via write_to_device(), export via Cluster::export_dmabuf(), RDMA-READ it back over a loopback QP pair on one NIC port, verify against the pattern. Gate on KMD version and an active RDMA port; skip cleanly when libibverbs-dev isn't available at build time.
 - Cross-host RDMA dma-buf initiator/target benchmark under examples/ (needs two hosts with RDMA NICs; not CI-runnable, so it belongs with the other examples rather than the unit test suite).
 - README.md covering both: prerequisites (kmd/kernel version, RDMA NIC, libibverbs-dev), build flags, and run instructions for each.

**Timeline**
No hard deadline.

**Is your feature request related to a problem? Please describe.**
export_dmabuf() is currently validated only by fd-validity checks (TestClusterExportDmabuf), not by an actual RDMA transfer.

**Definition of Done**
 - Loopback test merged into tests/unified/test_tlb.cpp, passing on real Blackhole + RDMA hardware, skipping cleanly elsewhere.
 - Cross-host initiator/target tool merged under examples/.
 - README documents prerequisites and run steps for both and is linked from wherever examples/tests are indexed.

broskoTT:
pending PR reviews, used tests to validate RDMA, missing machine configuration

ajovanovicTT:
Done with last PR merged https://github.com/tenstorrent/tt-umd/pull/3285

PR #3176, opened 2026-08-14
Add single-host RDMA dma-buf loopback test

ajovanovicTT:
### Issue
#3283

### Description
Adds a single-host RDMA loopback test that validates Cluster::export_dmabuf() end to end by RDMA-reading real device DRAM through the exported dma-buf and verifying it against a pattern seeded through a separate TLB window.

### List of the changes
 - Added tests/rdma/: rdma_loopback.hpp and rdma_loopback.cpp hold an RdmaLoopback class that brings up a loopback RC QP pair on the first ACTIVE RDMA port and reads an exported dma-buf; test_dmabuf_loopback.cpp holds the test itself. Includes a glossary of the RDMA terms used and per-method documentation.
 - Added tests/rdma/CMakeLists.txt with an rdma_tests target that requires libibverbs. It is not part of the default build and is deliberately left out of the umd_tests aggregate that CI runs, so the test is not disabled at the gtest level.
 - tests/unified/ is untouched; the RDMA code lives entirely under tests/rdma/.

### Testing
Built rdma_tests and unified_tests, both warning free. The test itself needs an ACTIVE RoCEv2 NIC and a kmd with dma-buf export support, so it was not executed.

### API Changes
This PR has no API changes.

PR #3284, opened 2026-08-27
Add cross-host RDMA dma-buf P2P example

ajovanovicTT:
### Issue
#3283

### Description
Adds a two-host RDMA example that validates Cluster::export_dmabuf() over real RDMA hardware and measures sustained bandwidth. The target exports a TLB window over a DRAM core as a dma-buf and registers it as an RDMA MR; the initiator RDMA-writes or reads into it from a peer host with no host-CPU copy on the target data path.

This is the cross-host counterpart to the single-host loopback gtest in #3176. It needs two hosts with RDMA NICs, so it is not CI-runnable and lives under examples/ rather than the test suite. Currently intended for Blackhole Galaxy systems and only exercised there.

### List of the changes
 - Added examples/rdma_dmabuf_p2p/: dmabuf_target.cpp (exports the dma-buf via UMD, verifies via an independent read path), dmabuf_initiator.cpp (peer side, links no UMD), rdma_common.hpp (TCP handshake, RDMA device and GID selection, RC QP bring-up and verification pattern shared by both binaries).
 - Added examples/rdma_dmabuf_p2p/CMakeLists.txt and wired it into examples/CMakeLists.txt.
 - Review fixes to the bring-up: hop_limit 64 rather than 1 so packets survive a routed fabric, path_mtu negotiated from the port active_mtu rather than hardcoded, and device and port selection that requires IBV_PORT_ACTIVE, plus a --dev flag.
 - Review fixes to verification: a position dependent pattern compared over the whole buffer rather than a byte periodic one over the first 4 KiB, dram_membar before the UMD readback, and a pass or fail byte from the initiator that the target propagates into its exit status.
 - Robustness: bounds checked argument values, --size rejected above the uint32 work request length and the port max_msg_sz, a 2 MiB default that matches a Blackhole TLB size class, a padding free handshake struct with a pinned size and a magic guard that also carries the MR length, MSG_NOSIGNAL and EINTR handling, and an --ordering flag for the export.

### Testing
Built both binaries with TT_UMD_BUILD_EXAMPLES=ON and TT_UMD_BUILD_RDMA=ON, warning free, and exercised the argument parsing and validation paths locally. End to end run on paired RDMA hosts was done earlier on the original branch and not re-run against this rebase.

### API Changes
This PR has no API changes.

### DIFF
--- a/examples/rdma_dmabuf_p2p/CMakeLists.txt
+++ b/examples/rdma_dmabuf_p2p/CMakeLists.txt
@@ -1,0 +1,31 @@
+# This example drives real RDMA verbs, so it needs libibverbs. It is skipped rather than failing
+# the configure step when libibverbs-dev is not installed, so enabling TT_UMD_BUILD_EXAMPLES stays
+# safe on hosts without an RDMA stack.
+find_library(IBVERBS_LIBRARY NAMES ibverbs)
+if(NOT IBVERBS_LIBRARY)
+    message(STATUS "rdma_dmabuf_p2p example: libibverbs not found, skipping")
+    return()
+endif()
+
+add_executable(dmabuf_target dmabuf_target.cpp)
+
+target_link_libraries(
+    dmabuf_target
+    PRIVATE
+        umd::tt-umd
+        tt-logger::tt-logger
+        ${IBVERBS_LIBRARY}
+)
+
+# The initiator only speaks RDMA against a plain host buffer, so it does not link UMD.
+add_executable(dmabuf_initiator dmabuf_initiator.cpp)
+
+target_link_libraries(dmabuf_initiator PRIVATE ${IBVERBS_LIBRARY})
+
+set_target_properties(
+    dmabuf_target
+    dmabuf_initiator
+    PROPERTIES
+        RUNTIME_OUTPUT_DIRECTORY
+            ${CMAKE_BINARY_DIR}/examples/rdma_dmabuf_p2p/
+)
--- a/examples/rdma_dmabuf_p2p/CMakeLists.txt
+++ b/examples/rdma_dmabuf_p2p/CMakeLists.txt
@@ -1,9 +1,40 @@
 # This example drives real RDMA verbs, so it needs libibverbs. It is skipped rather than failing
 # the configure step when libibverbs-dev is not installed, so enabling TT_UMD_BUILD_EXAMPLES stays
 # safe on hosts without an RDMA stack.
+#
+# Both the library and the headers are located: a host with libibverbs1 but not libibverbs-dev passes
+# find_library and would then hard-fail the build on "#include <infiniband/verbs.h>: No such file or
+# directory" — exactly the case the skip is meant to cover.
 find_library(IBVERBS_LIBRARY NAMES ibverbs)
-if(NOT IBVERBS_LIBRARY)
-    message(STATUS "rdma_dmabuf_p2p example: libibverbs not found, skipping")
+find_path(IBVERBS_INCLUDE_DIR NAMES infiniband/verbs.h)
+if(NOT (IBVERBS_LIBRARY AND IBVERBS_INCLUDE_DIR))
+    message(STATUS "rdma_dmabuf_p2p example: libibverbs library or headers not found, skipping")
+    return()
+endif()
+
+# The headers also have to be new enough: ibv_reg_dmabuf_mr() needs rdma-core >= v33 and
+# ibv_query_gid_ex() >= v32, and both are called unconditionally below. Without this check an older
+# rdma-core fails with a bare "was not declared in this scope" that says nothing about the cause.
+include(CheckCXXSymbolExists)
+set(CMAKE_REQUIRED_INCLUDES ${IBVERBS_INCLUDE_DIR})
+set(CMAKE_REQUIRED_LIBRARIES ${IBVERBS_LIBRARY})
+check_cxx_symbol_exists(
+    ibv_reg_dmabuf_mr
+    "infiniband/verbs.h"
+    HAVE_IBV_REG_DMABUF_MR
+)
+check_cxx_symbol_exists(
+    ibv_query_gid_ex
+    "infiniband/verbs.h"
+    HAVE_IBV_QUERY_GID_EX
+)
+unset(CMAKE_REQUIRED_INCLUDES)
+unset(CMAKE_REQUIRED_LIBRARIES)
+if(NOT (HAVE_IBV_REG_DMABUF_MR AND HAVE_IBV_QUERY_GID_EX))
+    message(
+        STATUS
+        "rdma_dmabuf_p2p example: libibverbs is too old (need rdma-core >= v33 for ibv_reg_dmabuf_mr and >= v32 for ibv_query_gid_ex), skipping"
+    )
     return()
 endif()
 
@@ -16,11 +47,13 @@ target_link_libraries(
         tt-logger::tt-logger
         ${IBVERBS_LIBRARY}
 )
+target_include_directories(dmabuf_target PRIVATE ${IBVERBS_INCLUDE_DIR})
 
 # The initiator only speaks RDMA against a plain host buffer, so it does not link UMD.
 add_executable(dmabuf_initiator dmabuf_initiator.cpp)
 
 target_link_libraries(dmabuf_initiator PRIVATE ${IBVERBS_LIBRARY})
+target_include_directories(dmabuf_initiator PRIVATE ${IBVERBS_INCLUDE_DIR})
 
 set_target_properties(
     dmabuf_target
--- a/examples/rdma_dmabuf_p2p/dmabuf_initiator.cpp
+++ b/examples/rdma_dmabuf_p2p/dmabuf_initiator.cpp
@@ -1,0 +1,326 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// RDMA initiator for the two-host dma-buf bandwidth test. Registers a normal host-memory MR filled
+// with a known test pattern, brings up an RC QP against the target, and issues repeated RDMA WRITEs
+// into the target's dma-buf-backed MR (see dmabuf_target.cpp) to measure sustained write bandwidth.
+// No UMD dependency here at all.
+#include <infiniband/verbs.h>
+
+#include <algorithm>
+#include <chrono>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <iostream>
+#include <string>
+#include <vector>
+
+#include "rdma_common.hpp"
+
+namespace {
+
+struct Args {
+    std::string host;
+    uint16_t port = 9999;
+    // 64 MiB: must match dmabuf_target's --size (Blackhole DRAM_BANK_SIZE is 4 GiB, so there's
+    // plenty of headroom vs the old Tensix-L1-backed 1.5 MiB ceiling).
+    uint64_t size = 64ull << 20;
+    uint64_t iters = 100;  // number of RDMA ops of --size bytes to issue, timed as one batch
+    int ib_port = 1;
+    int gid_index = -1;        // -1 = auto-detect a RoCEv2 GID (see resolve_gid_index in rdma_common.hpp)
+    std::string op = "write";  // "write" = host -> device NOC; "read" = device NOC -> host
+};
+
+Args parse_args(int argc, char** argv) {
+    Args a;
+    for (int i = 1; i < argc; ++i) {
+        std::string arg = argv[i];
+        auto next = [&]() { return std::string(argv[++i]); };
+        if (arg == "--host") {
+            a.host = next();
+        } else if (arg == "--port") {
+            a.port = static_cast<uint16_t>(std::stoi(next()));
+        } else if (arg == "--size") {
+            a.size = std::stoull(next());
+        } else if (arg == "--iters") {
+            a.iters = std::stoull(next());
+        } else if (arg == "--ib-port") {
+            a.ib_port = std::stoi(next());
+        } else if (arg == "--gid-index") {
+            a.gid_index = std::stoi(next());
+        } else if (arg == "--op") {
+            a.op = next();
+        }
+    }
+    return a;
+}
+
+// Number of in-flight, unsignaled RDMA_WRITEs between each polled completion. RC QPs complete
+// WQEs in order, so polling one CQE every SIGNAL_INTERVAL writes still confirms all of them landed.
+// Kept at 1 (fully synchronous, one WRITE in flight at a time): at the default 64 MiB --size,
+// ibv_poll_cq overhead is negligible next to per-message transfer time, so there's nothing to gain
+// from batching completions — and batching here is actively dangerous, since going over 32 previously
+// left ~2 GiB of unpaced, unacknowledged WRITEs outstanding and blew through the QP's retry budget
+// ("transport retry counter exceeded") on a real fabric. Only raise this if you've confirmed the
+// fabric tolerates more bytes in flight for your chosen --size.
+constexpr uint64_t SIGNAL_INTERVAL = 1;
+
+}  // namespace
+
+int run(int argc, char** argv) {
+    Args args = parse_args(argc, argv);
+    if (args.host.empty()) {
+        std::cerr << "Usage: dmabuf_initiator --host <target-host> [--port N] [--size N] [--iters N]"
+                  << " [--op write|read] [--gid-index N]" << std::endl;
+        return 1;
+    }
+    if (args.op != "write" && args.op != "read") {
+        std::cerr << "--op must be 'write' or 'read', got '" << args.op << "'" << std::endl;
+        return 1;
+    }
+    const bool is_read = (args.op == "read");
+
+    std::vector<uint8_t> local_buf(args.size);
+    if (is_read) {
+        // RDMA_READ pulls device memory into this buffer, so prefill with a sentinel that is NOT the
+        // expected pattern — otherwise a read that silently moved nothing would still "verify".
+        // The target seeds the real pattern into device memory via UMD before the handshake.
+        std::fill(local_buf.begin(), local_buf.end(), 0xAA);
+    } else {
+        // Test pattern: byte i = i & 0xFF, so the target can verify without needing to share the
+        // buffer contents out of band.
+        for (size_t i = 0; i < args.size; ++i) {
+            local_buf[i] = static_cast<uint8_t>(i & 0xFF);
+        }
+    }
+
+    int num_devices = 0;
+    ibv_device** dev_list = ibv_get_device_list(&num_devices);
+    if (!dev_list || num_devices == 0) {
+        std::cerr << "No RDMA devices found" << std::endl;
+        return 1;
+    }
+    ibv_context* ctx = ibv_open_device(dev_list[0]);
+    ibv_free_device_list(dev_list);
+    if (!ctx) {
+        std::cerr << "ibv_open_device failed" << std::endl;
+        return 1;
+    }
+
+    ibv_port_attr port_attr{};
+    if (ibv_query_port(ctx, args.ib_port, &port_attr)) {
+        std::cerr << "ibv_query_port failed" << std::endl;
+        return 1;
+    }
+    bool is_roce = (port_attr.link_layer == IBV_LINK_LAYER_ETHERNET);
+    if (is_roce) {
+        args.gid_index = resolve_gid_index(ctx, args.ib_port, args.gid_index);
+    }
+
+    ibv_pd* pd = ibv_alloc_pd(ctx);
+    ibv_mr* mr = ibv_reg_mr(pd, local_buf.data(), local_buf.size(), IBV_ACCESS_LOCAL_WRITE);
+    if (!mr) {
+        perror("ibv_reg_mr failed");
+        return 1;
+    }
+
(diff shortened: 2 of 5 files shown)


examples/rdma_dmabuf_p2p/dmabuf_initiator.cpp
```diff
@@ -0,0 +1,326 @@
...
+        attr.dest_qp_num = remote.qpn;
+        attr.rq_psn = remote.psn;
+        attr.max_dest_rd_atomic = 1;
+        attr.min_rnr_timer = 12;
+        attr.ah_attr.port_num = args.ib_port;
+        if (is_roce) {
+            attr.ah_attr.is_global = 1;
+            std::memcpy(attr.ah_attr.grh.dgid.raw, remote.gid, 16);
+            attr.ah_attr.grh.sgid_index = args.gid_index;
+            attr.ah_attr.grh.hop_limit = 1;
```
pjanevskiTT: `hop_limit = 1` means the packet dies at the first router. This is a *cross-host* example and `resolve_gid_index()` deliberately prefers the routable IPv4-mapped RoCEv2 GID, so the moment the two hosts aren't on the same L2 segment every packet is dropped and the initiator fails with "transport retry counter exceeded".

That's exactly the symptom `rdma_common.hpp:45-50` spends 12 lines attributing to a RoCEv1 GID, so this will be actively misdiagnosed.

`connect_qp()` on the base branch (`tests/unified/test_tlb.cpp:315`) uses `hop_limit = 64`. Same fix needed at `dmabuf_target.cpp:214`.

examples/rdma_dmabuf_p2p/dmabuf_initiator.cpp
```diff
@@ -0,0 +1,326 @@
...
+    int conn = tcp_connect(args.host, args.port);
+    QpExchangeInfo remote{};
+    recv_all(conn, &remote, sizeof(remote));
+    send_all(conn, &local, sizeof(local));
+    std::cout << "Target QPN=" << remote.qpn << " rkey=" << remote.rkey << std::endl;
+
+    {
+        ibv_qp_attr attr{};
+        attr.qp_state = IBV_QPS_RTR;
+        attr.path_mtu = IBV_MTU_1024;
```
pjanevskiTT: `path_mtu` is hardcoded rather than negotiated from `port_attr`, which is already queried at line 112 and then used only for `link_layer`.

- On a jumbo-frame RoCE fabric (`active_mtu == IBV_MTU_4096`) this pins the QP to a 1 KiB MTU, inflating per-packet header overhead ~4x — i.e. the bandwidth number this example exists to produce comes out silently low.
- On a port whose `active_mtu` is below 1024, the RTR transition fails or oversized packets are dropped.

`setup_loopback_qp_pair` on the base branch does `mtu = std::min(pa.active_mtu, IBV_MTU_4096)` (`tests/unified/test_tlb.cpp:393`). Same at `dmabuf_target.cpp:204`.

examples/rdma_dmabuf_p2p/dmabuf_initiator.cpp
```diff
@@ -0,0 +1,326 @@
...
+        }
+    }
+
+    int num_devices = 0;
+    ibv_device** dev_list = ibv_get_device_list(&num_devices);
+    if (!dev_list || num_devices == 0) {
+        std::cerr << "No RDMA devices found" << std::endl;
+        return 1;
+    }
+    ibv_context* ctx = ibv_open_device(dev_list[0]);
```
pjanevskiTT: `dev_list[0]` plus a hardcoded default `--ib-port 1`, with no check that the port is `IBV_PORT_ACTIVE` and no `--dev` flag.

On a multi-NIC host where `mlx5_0` is DOWN and the wired NIC is index 1, `Cluster::export_dmabuf()` succeeds (`has_any_active_rdma_port()` finds the *second* device) and then this side either throws "No RoCEv2 GID found on this port" or brings the QP up on a dead port and times out.

`open_first_active_rdma_device()` on the base branch (`tests/unified/test_tlb.cpp:196`) already does the correct scan: iterate devices, iterate `1..phys_port_cnt`, require `IBV_PORT_ACTIVE`. Same at `dmabuf_target.cpp:102`.

examples/rdma_dmabuf_p2p/dmabuf_initiator.cpp
```diff
@@ -0,0 +1,326 @@
...
+    int ib_port = 1;
+    int gid_index = -1;        // -1 = auto-detect a RoCEv2 GID (see resolve_gid_index in rdma_common.hpp)
+    std::string op = "write";  // "write" = host -> device NOC; "read" = device NOC -> host
+};
+
+Args parse_args(int argc, char** argv) {
+    Args a;
+    for (int i = 1; i < argc; ++i) {
+        std::string arg = argv[i];
+        auto next = [&]() { return std::string(argv[++i]); };
```
pjanevskiTT: `next()` does `argv[++i]` with no bounds check against `argc`. `dmabuf_initiator --host` — a trailing flag, easy typo — constructs `std::string` from `argv[argc]`, which is the guaranteed NULL terminator: UB, in practice a segfault inside `strlen` before the usage message at line 75 or `main`'s try/catch can turn it into a clean error. `i` is also left past `argc`.

Same at `dmabuf_target.cpp:47`, which has no usage message at all.

examples/rdma_dmabuf_p2p/dmabuf_initiator.cpp
```diff
@@ -0,0 +1,326 @@
...
+    // device memory is never modified.
+    //
+    // Direction note: WRITE pushes host -> device NOC and is "posted" on the target's PCIe link
+    // (fire-and-forget, pipelines well). READ pulls device NOC -> host, which makes the target NIC
+    // issue *non-posted* PCIe reads against the card's BAR — each needs a completion to come back, so
+    // throughput is bounded by read-completion concurrency and latency rather than raw link
+    // bandwidth. Expect READ to be substantially slower than WRITE on the same link.
+    ibv_sge sge{};
+    sge.addr = reinterpret_cast<uint64_t>(local_buf.data());
+    sge.length = static_cast<uint32_t>(local_buf.size());
```
pjanevskiTT, line 195: `sge.length` is `uint32_t` and `--size` is `uint64_t`, narrowed here with no range check.

`--size 4294967296` (4 GiB — a natural choice on Blackhole, whose TLB size classes are exactly {2 MiB, 4 GiB}, and which `PcieProtocol::export_dmabuf` accepts at `addr=0`) truncates `sge.length` to 0. Every WQE then moves nothing, every CQE is `IBV_WC_SUCCESS`, the target's first-4 KiB check passes on seeded data, and line 282 prints a bandwidth computed from `args.size * args.iters` — a fabricated number reported as PASS.

Below 4 GiB there's still `port_attr.max_msg_sz` (typically 1 GiB on mlx5), which is never checked. Worth rejecting `size > max_msg_sz` up front.

examples/rdma_dmabuf_p2p/dmabuf_initiator.cpp
```diff
@@ -0,0 +1,326 @@
...
+#include "rdma_common.hpp"
+
+namespace {
+
+struct Args {
+    std::string host;
+    uint16_t port = 9999;
+    // 64 MiB: must match dmabuf_target's --size (Blackhole DRAM_BANK_SIZE is 4 GiB, so there's
+    // plenty of headroom vs the old Tensix-L1-backed 1.5 MiB ceiling).
+    uint64_t size = 64ull << 20;
```
pjanevskiTT: The 64 MiB default only reasons about Blackhole. On Wormhole the largest TLB size class is 16 MiB (`device/arch/architecture_tlbs.cpp:132`), so `export_dmabuf` throws "No TLB window size can cover 67108864 bytes at address 0x0" and the run dies as `Fatal: ...` with nothing pointing at `--size`.

Either pick a default valid on both (the base-branch loopback test uses 2 MiB precisely because it's a valid size class on WH and BH), or have the target check the arch and produce an actionable error. Same default at `dmabuf_target.cpp:37`.

examples/rdma_dmabuf_p2p/dmabuf_target.cpp
```diff
@@ -0,0 +1,293 @@
...
+        // The data moved device -> host, so the initiator holds it and verifies on its side; there is
+        // nothing to check here. Device memory should be unchanged from what we seeded.
+        std::cout << "Initiator signaled read complete; it verifies the pattern on its side." << std::endl;
+    } else {
+        std::cout << "Initiator signaled write complete; verifying via UMD readback..." << std::endl;
+
+        std::vector<uint8_t> readback(args.size);
+        cluster->read_from_device(readback.data(), args.chip, core, args.addr, args.size);
+
+        for (size_t i = 0; i < args.size && i < 4096; ++i) {
```
pjanevskiTT: Two problems here, and the comment overstates what this catches:

1. The pattern `byte i = i & 0xFF` is 256-byte periodic, so it cannot detect a wrong-address bug at all — any offset that's a multiple of 256 (a mis-computed `window_offset`, an rkey/iova shift by a page) verifies clean.
2. Only the first 4 KiB is compared, so a WRITE where only the first page landed also passes.

The base branch's loopback gtest deliberately uses a golden-ratio multiplier per 64-bit word (`test_tlb.cpp:496`) "so a partial or misaddressed transfer cannot accidentally verify" — this regresses that. A position-dependent pattern plus a full compare fixes both, and also removes the waste at lines 256-257, where 64 MiB is pulled back over MMIO so that 4 KiB of it can be looked at.

Same at `dmabuf_initiator.cpp:291` for `--op read`.

examples/rdma_dmabuf_p2p/dmabuf_target.cpp
```diff
@@ -0,0 +1,293 @@
...
+    bool all_match = true;
+    if (is_read) {
+        // The data moved device -> host, so the initiator holds it and verifies on its side; there is
+        // nothing to check here. Device memory should be unchanged from what we seeded.
+        std::cout << "Initiator signaled read complete; it verifies the pattern on its side." << std::endl;
+    } else {
+        std::cout << "Initiator signaled write complete; verifying via UMD readback..." << std::endl;
+
+        std::vector<uint8_t> readback(args.size);
+        cluster->read_from_device(readback.data(), args.chip, core, args.addr, args.size);
```
pjanevskiTT, line 219: This readback races the RDMA writes. The initiator's CQE only proves the target NIC ACKed at the RoCE layer — it says nothing about the posted PCIe -> TLB -> NOC -> DRAM writes having landed, and `Cluster::export_dmabuf`'s window uses the default `tlb_data::Relaxed` ordering, which imposes nothing against traffic on the *different* window this `read_from_device()` goes through.

Expected symptom: an intermittent, unreproducible "Mismatch at byte N" that reads as a dma-buf bug. `Cluster::dram_membar(chip, cores)` exists for this and isn't called. It'd also be worth plumbing `export_dmabuf`'s `ordering` parameter through as a flag so Posted/Strict can be tried.

examples/rdma_dmabuf_p2p/dmabuf_target.cpp
```diff
@@ -0,0 +1,293 @@
...
+        if (ibv_modify_qp(qp, &attr, flags)) {
+            std::cerr << "Failed to move QP to RTS" << std::endl;
+            return 1;
+        }
+    }
+
+    // Wait for the initiator to signal its RDMA batch has completed (a real hardware ACK, not just
+    // "posted").
+    char done = 0;
+    recv_all(conn, &done, 1);
```
pjanevskiTT: `done` is written here and never read — and the initiator sends `1` unconditionally, including after it has printed FAIL (`dmabuf_initiator.cpp:300-306`).

So in `--op read` with corrupted data: the initiator prints "Mismatch at byte N" / FAIL and returns 1, while the target prints "Initiator signaled read complete" and returns **0**. Any CI harness keying off the target's exit status — the natural side to monitor, since it owns the device — records a clean pass on a data-corruption run.

Send the pass/fail in that byte and propagate it into the target's return value.

examples/rdma_dmabuf_p2p/rdma_common.hpp
```diff
@@ -0,0 +1,191 @@
...
+#include <string>
+
+// Out-of-band info exchanged between the two hosts to bring up the RC QP.
+// Sent as a flat, fixed-size struct over a plain TCP socket (no serialization library needed).
+struct QpExchangeInfo {
+    uint32_t qpn = 0;
+    uint32_t psn = 0;
+    uint16_t lid = 0;          // valid only if the port's link layer is InfiniBand
+    uint8_t gid[16] = {0};     // valid only if the port's link layer is Ethernet (RoCE)
+    uint32_t rkey = 0;         // target's dma-buf MR rkey; unused/zero from the initiator
```
pjanevskiTT: The struct carries `rkey` and `remote_addr` but not the MR length, and the two binaries are launched independently with their own `--size`.

- Initiator larger than target: every WRITE overruns the MR and dies with `IBV_WC_REM_ACCESS_ERR` at iter 0, indistinguishable from a permissions/dma-buf failure.
- Initiator smaller: it succeeds, the target's 4 KiB check passes, and a bandwidth figure is printed for a window that was never fully written.

Lines 30-31 justify shipping `remote_addr` "so the initiator doesn't have to hardcode it" — length has the identical argument. Add it and have the initiator reject a mismatch with a clear message.

examples/rdma_dmabuf_p2p/rdma_common.hpp
```diff
@@ -0,0 +1,191 @@
...
+    ibv_query_gid(ctx, ib_port, idx, &gid);
+    std::cout << "Auto-selected RoCEv2 GID index " << idx << " (" << gid_to_string(gid) << ")" << std::endl;
+    return idx;
+}
+
+inline void send_all(int fd, const void* buf, size_t len) {
+    const char* p = static_cast<const char*>(buf);
+    size_t sent = 0;
+    while (sent < len) {
+        ssize_t n = send(fd, p + sent, len - sent, 0);
```
pjanevskiTT: `send()` without `MSG_NOSIGNAL`, and neither binary ignores SIGPIPE. When the peer dies mid-run (Ctrl-C, or a failed WQE), the next `send()` on the closed socket kills the process with signal 13 — no exception, no "connection error", and no cleanup of the exported dma-buf / TLB window.

`dmabuf_target.cpp:285-286` promises the opposite: "send_all() throw[s] on a dropped connection (e.g. the initiator dying mid-run); catch here so a peer failure prints a clean error". Either `MSG_NOSIGNAL` here or `signal(SIGPIPE, SIG_IGN)` at startup.

Minor, same function and `recv_all` below: `n < 0` with `EINTR` is treated as fatal rather than retried.

examples/rdma_dmabuf_p2p/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+# This example drives real RDMA verbs, so it needs libibverbs. It is skipped rather than failing
+# the configure step when libibverbs-dev is not installed, so enabling TT_UMD_BUILD_EXAMPLES stays
+# safe on hosts without an RDMA stack.
+find_library(IBVERBS_LIBRARY NAMES ibverbs)
```
pjanevskiTT, line 8: The guard only checks for the library, so it fails open exactly where the comment above says it won't: a host with `libibverbs1` but not `libibverbs-dev` (or headers outside the default include path) passes `find_library`, doesn't `return()`, and then hard-fails the build on `#include <infiniband/verbs.h>: No such file or directory`.

The base branch already gets this right in `tests/unified/CMakeLists.txt:11-16` — `find_library` **and** `find_path(IBVERBS_INCLUDE_DIR NAMES infiniband/verbs.h)`, plus `target_include_directories`. Worth mirroring here, including the include dirs on both targets.

Both binaries also call `ibv_reg_dmabuf_mr` (rdma-core >= v33) and `ibv_query_gid_ex` (>= v32) unconditionally, with nothing checking the headers are new enough.

examples/rdma_dmabuf_p2p/rdma_common.hpp
```diff
@@ -0,0 +1,191 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Shared helpers for the two-host dma-buf RDMA example: the out-of-band TCP handshake used to
```
pjanevskiTT: Higher-level: this header plus the two `.cpp`s re-implement the RDMA bring-up that this PR's own base branch just added to `tests/unified/test_tlb.cpp` — `open_first_active_rdma_device()`, `pick_rocev2_gid()`, `make_qp()`, `connect_qp()`, `setup_loopback_qp_pair()`.

The copy has already diverged in three places, and each divergence is one of the bugs I flagged above: `hop_limit` 64 -> 1, `mtu = min(active_mtu, 4096)` -> hardcoded `IBV_MTU_1024`, first-ACTIVE-port scan -> `dev_list[0]`. Every future fix to the bring-up now has to be made twice.

Suggest lifting the shared bring-up into one header that both the loopback gtest and this example include, and keeping `rdma_common.hpp` to just the TCP handshake and `QpExchangeInfo`.

examples/rdma_dmabuf_p2p/dmabuf_initiator.cpp
ajovanovicTT: Fixed. hop_limit is now 64 in connect_rc_qp, shared by both binaries.

examples/rdma_dmabuf_p2p/dmabuf_initiator.cpp
ajovanovicTT: Fixed. path_mtu is now min(port_attr.active_mtu, IBV_MTU_4096) in connect_rc_qp.

examples/rdma_dmabuf_p2p/dmabuf_initiator.cpp
ajovanovicTT: Fixed. open_active_rdma_port scans all devices and ports for IBV_PORT_ACTIVE, and a --dev flag was added. An explicit device or port that is not ACTIVE is rejected.

examples/rdma_dmabuf_p2p/dmabuf_initiator.cpp
ajovanovicTT: Fixed. arg_value throws when a flag is trailing, in both binaries. dmabuf_target also got a usage message and rejects unknown args. Verified: dmabuf_initiator --host now prints an error and exits 1 instead of segfaulting.

examples/rdma_dmabuf_p2p/dmabuf_initiator.cpp
ajovanovicTT: Fixed. --size is rejected above UINT32_MAX and above port_attr.max_msg_sz, before the buffer is allocated.

examples/rdma_dmabuf_p2p/dmabuf_initiator.cpp
ajovanovicTT: Fixed. Default is now 2 MiB on both sides, the only size class valid on Wormhole and Blackhole. Documented in the README and in the Args comment.

examples/rdma_dmabuf_p2p/dmabuf_target.cpp
ajovanovicTT: Fixed. Both sides now use a position dependent pattern with the same golden ratio multiplier as the loopback test, and compare the whole buffer instead of the first 4 KiB.

examples/rdma_dmabuf_p2p/dmabuf_target.cpp
ajovanovicTT: Fixed. dram_membar is called before the readback. Also added an --ordering flag so the export can run with Posted or Strict.

examples/rdma_dmabuf_p2p/dmabuf_target.cpp
ajovanovicTT: Fixed. The initiator now sends a pass or fail byte and the target propagates it into its exit status.

examples/rdma_dmabuf_p2p/rdma_common.hpp
ajovanovicTT: Fixed. mr_length is in the handshake and the initiator refuses a size mismatch with a clear message.

examples/rdma_dmabuf_p2p/rdma_common.hpp
ajovanovicTT: Fixed. send uses MSG_NOSIGNAL and EINTR is retried in both send_all and recv_all.

examples/rdma_dmabuf_p2p/CMakeLists.txt
ajovanovicTT: Fixed. find_path for the headers plus target_include_directories on both targets, and a check_cxx_symbol_exists guard for ibv_reg_dmabuf_mr and ibv_query_gid_ex.

examples/rdma_dmabuf_p2p/rdma_common.hpp
ajovanovicTT: Partly done. The bring up now exists once in rdma_common.hpp (open_active_rdma_port, create_rc_qp, connect_rc_qp) and both binaries use it, so the two sides cannot diverge again. All three divergences you listed are fixed. Sharing it with the loopback gtest is not done: after the base branch restructuring that test lives in tests/rdma, and a helper shared with examples/ would need a home that neither directory owns. Happy to do it if you have a preferred location.

pjanevskiTT (approved):

PR #3285, opened 2026-08-27
Document the RDMA dma-buf tests

ajovanovicTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/3283

### Description
Documents how to run the two RDMA dma-buf tests added in the preceding PRs in this stack: the single-host loopback gtest and the two-host peer-to-peer example. Also makes both of them opt-in at configure time, which came out of the review discussion on this PR.

### List of the changes
 - Added examples/rdma_dmabuf_p2p/README.md: scope, prerequisites split into distro packages, NIC driver and fabric configuration, build steps, run instructions for both sides, a table of the available flags, and notes on DRAM bank limits, TLB window size classes, RoCEv2 GID selection and pinned memory limits. Cross-references the loopback gtest for the single-host case.
 - Listed the new example in examples/README.md.
 - Added the TT_UMD_BUILD_RDMA option, OFF by default, gating both tests/rdma and examples/rdma_dmabuf_p2p. With it ON, missing or too-old libibverbs is a configure error naming what to install rather than a silent skip, so the host environment no longer decides what gets built.
 - Added cmake/ibverbs.cmake, which locates libibverbs, checks that rdma-core is new enough for ibv_reg_dmabuf_mr and ibv_query_gid_ex, and defines an imported ibverbs::ibverbs target. Both consumers previously carried the same lookup separately.
 - Stated that this is currently intended for Blackhole Galaxy and only exercised there, and dropped the Wormhole size class guidance from the example.

### Testing
Configured with the option off (no RDMA targets and no libibverbs lookup), with it on (all three targets build warning free), and with it on and libibverbs hidden (a single configure error pointing at the fix).

### API Changes
This PR has no API changes.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -8,7 +8,7 @@ This directory contains examples demonstrating how to use various software compo
 Demonstrates TTDevice usage, showcasing basic device operations and the difference between functionality available before and after calling `init_tt_device()`.
 
 ### `rdma_dmabuf_p2p/`
-Two-host benchmark for `Cluster::export_dmabuf()`: a peer NIC RDMA-writes (or reads) directly against a TLB window over a DRAM core on the other host's card. Requires RDMA hardware on both hosts and `libibverbs`; skipped at configure time if `libibverbs` is missing.
+Two-host benchmark for `Cluster::export_dmabuf()`: a peer NIC RDMA-writes (or reads) directly against a TLB window over a DRAM core on the other host's card. Requires RDMA hardware on both hosts and libibverbs (library + headers, e.g. `libibverbs-dev`); skipped at configure time if the ibverbs library is missing.
 
 ## Building Examples
 
--- a/examples/README.md
+++ b/examples/README.md
@@ -7,6 +7,9 @@ This directory contains examples demonstrating how to use various software compo
 ### `tt_device_example/`
 Demonstrates TTDevice usage, showcasing basic device operations and the difference between functionality available before and after calling `init_tt_device()`.
 
+### `rdma_dmabuf_p2p/`
+Two-host benchmark for `Cluster::export_dmabuf()`: a peer NIC RDMA-writes (or reads) directly against a TLB window over a DRAM core on the other host's card. Requires RDMA hardware on both hosts and `libibverbs`; skipped at configure time if `libibverbs` is missing.
+
 ## Building Examples
 
 Examples are not built by default. To build them:
--- a/examples/README.md
+++ b/examples/README.md
@@ -8,7 +8,7 @@ This directory contains examples demonstrating how to use various software compo
 Demonstrates TTDevice usage, showcasing basic device operations and the difference between functionality available before and after calling `init_tt_device()`.
 
 ### `rdma_dmabuf_p2p/`
-Two-host benchmark for `Cluster::export_dmabuf()`: a peer NIC RDMA-writes (or reads) directly against a TLB window over a DRAM core on the other host's card. Requires RDMA hardware on both hosts and libibverbs (library + headers, e.g. `libibverbs-dev`); skipped at configure time if the ibverbs library is missing.
+Two-host benchmark for `Cluster::export_dmabuf()`: a peer NIC RDMA-writes (or reads) directly against a TLB window over a DRAM core on the other host's card. Blackhole only at the moment (intended for Blackhole Galaxy systems). Requires RDMA hardware on both hosts and libibverbs (library + headers, e.g. `libibverbs-dev`). Built only with `-DTT_UMD_BUILD_RDMA=ON`, which is OFF by default; with it ON, missing or too-old libibverbs is a configure error rather than a silent skip.
 
 ## Building Examples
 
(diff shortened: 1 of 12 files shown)


examples/rdma_dmabuf_p2p/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Build (both hosts)
+
+```bash
+cmake -B build -DCMAKE_BUILD_TYPE=Release -DTT_UMD_BUILD_EXAMPLES=ON
+cmake --build build -j"$(nproc)"
+```
+
+Binaries land in `build/examples/rdma_dmabuf_p2p/`. If `libibverbs` is not installed the example is
+skipped at configure time rather than failing the build.
```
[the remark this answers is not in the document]
ajovanovicTT: This check will be added to both loopback and p2p test.

examples/rdma_dmabuf_p2p/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+```bash
+# tt-kmd version — need >= 2.10.0-rc1 for TENSTORRENT_IOCTL_EXPORT_TLB_DMABUF
+cat /sys/module/tenstorrent/version
+
+# kernel — need >= 5.8
+uname -r
+
+# RDMA NIC present and active
+ibv_devices
```
broskoTT: any prerequisites on how to install these dependancies? Or is that part of installing the NIC?

examples/README.md
```diff
@@ -7,6 +7,9 @@ This directory contains examples demonstrating how to use various software compo
 ### `tt_device_example/`
 Demonstrates TTDevice usage, showcasing basic device operations and the difference between functionality available before and after calling `init_tt_device()`.
 
+### `rdma_dmabuf_p2p/`
+Two-host benchmark for `Cluster::export_dmabuf()`: a peer NIC RDMA-writes (or reads) directly against a TLB window over a DRAM core on the other host's card. Requires RDMA hardware on both hosts and libibverbs (library + headers, e.g. `libibverbs-dev`); skipped at configure time if the ibverbs library is missing.
```
broskoTT: `skipped at configure time if the ibverbs library is missing`
As I mentioned some other way on the related PR, not explicitly like this, but I think it is better to have more control here.
Rather than having environment dictate what we build and what we don't, let's introduce a variable something like "TT_UMD_BUILD_RDMA_TEST/EXAMPLES", then if it is set to ON, having ibverbs is required and cmake errors out if it can't find it. If it is off you skip the build of related components of course.

examples/rdma_dmabuf_p2p/README.md
```diff
@@ -0,0 +1,110 @@
...
+*posted* (fire-and-forget, they pipeline), while reads are *non-posted* — the NIC must issue read
+requests against the card's BAR and wait for each completion, so throughput is governed by
+read-completion concurrency and latency rather than raw link bandwidth.
+
+## Notes
+
+- `addr + size` must stay within a single DRAM bank. On Blackhole the usable bank space ends short
+  of the nominal `DRAM_BANK_SIZE` — keep clear of the tail when picking large `--size`/`--addr`.
+- The underlying TLB window size is chosen by `tt_tlb_alloc()`, which accepts only specific size
+  classes (1/2/16 MiB on Wormhole, 2 MiB or 4 GiB on Blackhole). `export_dmabuf()` rounds `--size`
```
broskoTT: Does all of this even work on Wormhole? If not, remove it here, but also add it somewhere, to the beginning of this guide that all of this is BH specific at the moment

broskoTT (approved):

examples/rdma_dmabuf_p2p/README.md
ajovanovicTT: Added. Prerequisites is now split into four steps: the distro packages, the NIC driver and firmware which come with the NIC rather than with those packages, the RoCE port IP configuration the GID is derived from, and the tt-kmd version. Your guess was right, most of the RDMA stack is part of bringing the NIC up.

examples/README.md
ajovanovicTT: Done. Added TT_UMD_BUILD_RDMA, OFF by default, gating both tests/rdma and examples/rdma_dmabuf_p2p. With it ON, missing or too-old libibverbs is a configure error naming what to install, so the environment no longer decides what gets built. The lookup and the rdma-core version checks now live once in cmake/ibverbs.cmake as an imported ibverbs::ibverbs target, since both consumers were carrying the same 35 lines. rdma_tests also no longer needs EXCLUDE_FROM_ALL, the option is the gate.

examples/rdma_dmabuf_p2p/README.md
ajovanovicTT: Done. The Wormhole size class line is gone and the guide now states up front that this is intended for Blackhole Galaxy and only exercised there. The 2 MiB default is explained in Blackhole terms, anything larger consumes one of the few 4 GiB windows. Also added the same scope note to the loopback test. Nothing in the code rejects Wormhole, but it is unverified so the docs no longer suggest it.

PR #3289, opened 2026-08-27
[Base API] Delegate set_clock_state to DeviceFirmware

nbuncicTT:
## Issue

Both architectures implement `DeviceFirmware::set_clock_state`, but nothing calls it — `TTDevice` still answers with its own per-arch copy.

## Description

`TTDevice::set_clock_state` stays on the facade and forwards to the firmware; it just stops being virtual, since the per-arch behaviour has moved. `Chip::set_clock_state` still calls `tt_device->set_clock_state(...)`, now with a `ClockState`.

The replacement landed earlier in the stack, so the diff here only shows the deletion. This is the removed Wormhole body against the firmware body replacing it ([`wormhole_device_firmware.cpp#L452-L475`](https://github.com/tenstorrent/tt-umd/blob/8d55c9d9e0de36bdb37c19ff301a3aebabe515d2/device/tt_device/firmware/wormhole_device_firmware.cpp#L452-L475)), class prefixes and the enum rename normalised away:

```diff
--- TTDevice
+++ DeviceFirmware
@@ -1,10 +1,24 @@
-void set_clock_state(ClockState state, NocId /*noc_id*/) {
-    ZoneScoped;
-    uint32_t msg = get_power_state_arc_msg(state);
-    int exit_code = get_arc_messenger()->send_message(msg, {0, 0});
+void set_clock_state(ClockState state, NocId noc_id) {
+    uint32_t msg_code = wormhole::ARC_MSG_COMMON_PREFIX;
+    uint32_t target_aiclk = 0;
+    switch (state) {
+        case ClockState::BUSY:
+            msg_code |= architecture_impl_->get_firmware_message_go_busy();
+            target_aiclk = firmware_info_provider_->get_max_clock_freq().value_or(0);
+            break;
+        case ClockState::IDLE:
+            msg_code |= architecture_impl_->get_firmware_message_go_idle();
+            target_aiclk = wormhole::AICLK_IDLE_VAL;
+            break;
+        default:
+            UMD_THROW(error::RuntimeError, "Unrecognized clock state.");
+    }
+
+    DeviceCommandResult result = send_device_command(msg_code, {0, 0}, timeout::ARC_MESSAGE_TIMEOUT, noc_id);
     UMD_ASSERT(
-        exit_code == 0,
+        result.exit_code == 0,
         error::RuntimeError,
-        fmt::format("Failed to set clock state to {} with exit code: {}", (int)state, exit_code));
-    wait_for_aiclk_value(state);
+        fmt::format("Failed to set clock state to {} with exit code: {}", (int)state, result.exit_code));
+
+    wait_for_aiclk_value(target_aiclk, noc_id);
 }
```

Two intended differences: the ARC message goes through `send_device_command()`, and the BUSY target comes from `FirmwareInfoProvider` rather than `TTDevice` — which is why `wait_for_aiclk_value()` now takes the target instead of the state. Blackhole is the same shape.

## List of the changes

- `TTDevice::set_clock_state` forwards to the firmware and is no longer virtual; it takes `ClockState`.
- Deleted: both arch overrides, `get_power_state_arc_msg`, `wait_for_aiclk_value`, `log_aiclk_timeout_warning`, `AICLK_TOLERANCE_PERCENT`.

## Testing

## API Changes

`TTDevice::set_clock_state` now takes `ClockState` instead of `TTDevice::PowerState`, and is no longer virtual. `TTDevice::wait_for_aiclk_value` removed.


device/tt_device/tt_device.cpp
```diff
@@ -248,80 +245,8 @@ void TTDevice::set_power_state(TTDevice::PowerState state, NocId /*noc_id*/) {
...
-}
-
-void TTDevice::log_aiclk_timeout_warning(uint32_t target_aiclk, std::chrono::milliseconds timeout_ms) {
-    const uint32_t aiclk = get_clock();
-
-    auto *telemetry = get_firmware_telemetry_reader();
-    std::string arb_max_info;
-    if (telemetry != nullptr && telemetry->is_entry_available(TelemetryTag::AICLK_ARB_MAX)) {
-        const uint32_t arb_max = telemetry->read_entry(TelemetryTag::AICLK_ARB_MAX);
-        arb_max_info = fmt::format(
```
broskoTT, line 301: I don't see this code in DeviceFirmware, but it was added due to https://tenstorrent.atlassian.net/browse/DCAMP-186

nbuncicTT:
Closing in favor of #3323.

PR #3290, opened 2026-08-27
[Base API] Delegate set_power_state to DeviceFirmware

nbuncicTT:
## Issue

`TTDevice::set_power_state` reached past its own abstractions to `get_pci_device()`, while `PcieInterface::set_power_state(PowerState)` had no callers.

## Description

It forwards to the firmware instead, which makes that API live:

```diff
-    if (is_remote() || model_->get_pcie_interface() == nullptr) {
-        return;
-    }
-    get_pci_device()->set_power_state(state == TTDevice::PowerState::BUSY);
+    get_device_firmware()->set_power_state(
+        state == TTDevice::PowerState::BUSY ? tt::umd::PowerState::HIGH : tt::umd::PowerState::LOW, noc_id);
```

Unlike the rest of the stack this deletes nothing — `TTDevice::set_power_state` survives as a converting forwarder, so the two enums can be collapsed separately (~90 sites, 85 in tests).

The early return moves into the firmware, which returns on a null `PcieInterface`. `TTDevice` also returned early on `is_remote()`; equivalent, since a remote model exposes no `PcieInterface`.

## List of the changes

- `TTDevice::set_power_state` delegates and converts the enum.

## Testing

## API Changes

None — the signature is unchanged.


broskoTT (approved):

nbuncicTT:
Closing in favor of #3324.

PR #3291, opened 2026-08-27
[Base API] Delegate get_noc_translation_enabled to DeviceFirmware

nbuncicTT:
## Issue

`TTDevice::get_noc_translation_enabled` was pure virtual with three overrides, each duplicating a register read the firmware already does.

## Description

The pure virtual becomes a forwarder; the Blackhole, Wormhole and simulation overrides are deleted. Removed Wormhole body against its replacement ([`wormhole_device_firmware.cpp#L313-L320`](https://github.com/tenstorrent/tt-umd/blob/8d55c9d9e0de36bdb37c19ff301a3aebabe515d2/device/tt_device/firmware/wormhole_device_firmware.cpp#L313-L320)):

```diff
--- TTDevice
+++ DeviceFirmware
@@ -1,7 +1,8 @@
-bool get_noc_translation_enabled() {
-    uint32_t niu_cfg = 0x0;
+bool get_noc_translation_enabled(NocId noc_id) {
     constexpr uint32_t ARC_APB_NIU_0_OFFSET = 0x50000;
     constexpr uint32_t NIU_CFG_0_OFFSET = 0x100;
-    read_from_arc_apb(&niu_cfg, ARC_APB_NIU_0_OFFSET + NIU_CFG_0_OFFSET, sizeof niu_cfg);
+
+    uint32_t niu_cfg = 0x0;
+    read_from_arc_apb(&niu_cfg, ARC_APB_NIU_0_OFFSET + NIU_CFG_0_OFFSET, sizeof(niu_cfg), noc_id);
     return (niu_cfg & (1 << 14)) != 0;
 }
```

Same register, same bit. The only change is that the firmware holds the interfaces directly and takes the NOC per call.

This was scoped as blocked by constructor ordering. It isn't: the firmware is built at the top of `BlackholeTTDevice`'s constructor and computes its own answer, so it never needs `TTDevice` to have answered first.

## List of the changes

- `TTDevice::get_noc_translation_enabled()` delegates instead of being pure virtual.
- Three overrides deleted.

## Testing

## API Changes

`TTDevice::get_noc_translation_enabled()` is no longer virtual. Behaviour unchanged.

### DIFF
--- a/device/api/umd/device/tt_device/blackhole_tt_device.hpp
+++ b/device/api/umd/device/tt_device/blackhole_tt_device.hpp
@@ -33,8 +33,6 @@ public:
 
     uint32_t get_min_clock_freq() override;
 
-    bool get_noc_translation_enabled() override;
-
     void read_from_arc_apb(void *mem_ptr, uint64_t arc_addr_offset, size_t size) override;
 
     void write_to_arc_apb(const void *mem_ptr, uint64_t arc_addr_offset, size_t size) override;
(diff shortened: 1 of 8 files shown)


aleksamarkovicTT (approved):

nbuncicTT:
Closing in favor of #3325.

PR #3295, opened 2026-08-27
[Base API] Delegate get_chip_info to DeviceFirmware

nbuncicTT:
## Issue

`get_chip_info` was a base filling the common fields plus two arch overrides adding harvesting — duplicating firmware implementations that already do both.

## Description

The base becomes a forwarder and all three overrides are deleted.

One of them could not be dropped as-is. `TTSimDeviceFirmware::get_chip_info` returned a bare `ChipInfo{}`, while `TTSimTTDevice` set `eth_harvesting_mask = 0x120` for Blackhole — without which SocDescriptor construction throws:

```diff
--- TTDevice
+++ DeviceFirmware
@@ -1,10 +1,10 @@
-ChipInfo get_chip_info() {
-    // No firmware_info_provider on the simulator; mirror the defaults used inside
-    // create(). BH SocDescriptor construction rejects an empty eth_harvesting_mask
-    // ("Exactly 2 or 14 ETH cores should be harvested on full Blackhole"), so apply the same 0x120
-    // default here. Keep in sync with create() above.
+ChipInfo get_chip_info(NocId noc_id) {
+    // There is no FirmwareInfoProvider on the simulator, so the defaults mirror the ones
+    // create() uses. Blackhole SocDescriptor construction rejects an empty
+    // eth_harvesting_mask ("Exactly 2 or 14 ETH cores should be harvested on full Blackhole"), so the
+    // same 0x120 default is applied here. Keep in sync with create().
     ChipInfo chip_info{};
-    if (get_arch() == tt::ARCH::BLACKHOLE) {
+    if (arch_ == tt::ARCH::BLACKHOLE) {
         chip_info.harvesting_masks.eth_harvesting_mask = 0x120;
     }
     return chip_info;
```

So the default moves with the function, and `TTSimDeviceFirmware` now takes the architecture. Worth generalising: a firmware method is not automatically equivalent to the one it replaces — diff the bodies rather than trusting the name.

## List of the changes

- `TTDevice::get_chip_info()` delegates and is no longer virtual.
- Wormhole, Blackhole and TTSim overrides deleted.
- `TTSimDeviceFirmware` takes `tt::ARCH` and carries the Blackhole default.

## Testing

## API Changes

`TTDevice::get_chip_info()` is no longer virtual. `TTSimDeviceFirmware` gains a constructor argument.

### DIFF
--- a/.github/workflows/build-and-run-all-tests.yml
+++ b/.github/workflows/build-and-run-all-tests.yml
@@ -154,6 +154,7 @@ jobs:
         # - P300-viommu is a shared pool of 3 P300 machines.
         # - bh-loudbox is a single machine in CIv2. It also has an additional problem where
         #   it actually has a ~20min startup time for viommu variant, so this one can be really slow.
+        #   It recently became congested, so it is blacklisted from merge queue for now, but will still run on main.
         # - The wh-glx-6u is also one machine, but not yet fully proven, so don't want to block merge queue with it.
         # - 6u is a single machine and the main merge queue bottleneck, so in the merge queue it
         #   runs on one build type only. Release is dropped because the PR gate already covers it,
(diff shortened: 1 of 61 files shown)


nbuncicTT:
Closing in favor of #3326.

PR #3298, opened 2026-08-27
[Base API] Delegate get_firmware_noc_coord to DeviceFirmware

nbuncicTT:
## Issue

`TTDevice` cached the ARC core coordinate in `arc_core_noc0`/`arc_core_noc1`, filled by a `set_arc_coordinate()` hook each architecture overrode — duplicating what the firmware already computes.

## Description

The cache and the hook go; `get_arc_core()` forwards to `get_firmware_noc_coord()`, which is the same selection over the firmware's own members.

The coordinates themselves are computed the same way, just in the firmware's constructor: Blackhole calls `blackhole::get_arc_core(...)` exactly as `set_arc_coordinate()` did, Wormhole uses the same fixed constants.

Scoped as hard because of ~20 callers, but most were other classes' own `get_arc_core()` — `ArcTelemetryReader` and `BlackholeArcMessageQueue` each compute their own.

## List of the changes

- Both `get_arc_core()` overloads delegate.
- Deleted: `arc_core_noc0`/`arc_core_noc1`, virtual `set_arc_coordinate()` and both overrides.
- `get_device_firmware()` is now `const`.

## Testing

## API Changes

`TTDevice::set_arc_coordinate()` removed. `get_device_firmware()` is now `const`.

### DIFF
--- a/device/api/umd/device/tt_device/blackhole_tt_device.hpp
+++ b/device/api/umd/device/tt_device/blackhole_tt_device.hpp
@@ -58,7 +58,6 @@ protected:
     // Number of retrain attempts is chosen based on syseng team testing.
     uint32_t get_max_dram_retrain_attempts() const override { return 3; }
 
-    void set_arc_coordinate() override;
 
     void probe_arc() override;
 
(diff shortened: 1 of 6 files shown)


nbuncicTT:
Closing in favor of #3327.

PR #3299, opened 2026-08-27
[Base API] Route production ARC commands through DeviceFirmware

nbuncicTT:
## Issue

Every ARC command goes through `ArcMessenger` via `TTDevice`, while `DeviceFirmware::send_device_command` — already implemented by both architectures — had no callers outside the firmware.

## Description

Sixteen production call sites move to the firmware: `warm_reset`, `Chip::arc_msg`, `retrain_dram_core`, `get_clock`, `configure_iatu_region`, and ten in the two SPI devices. Nothing moved bodily, so there is no body-diff to show — each site changes shape like this:

```diff
-    uint32_t exit_code = get_arc_messenger()->send_message(msg_code, return_values, args, timeout);
+    DeviceCommandResult result = get_device_firmware()->send_device_command(msg_code, args, timeout);
```

The SPI flows keep their old shape via a file-local adapter, since `send_message` filled the caller's vector in place. The timeout is passed explicitly everywhere rather than defaulted — defaults on a virtual dispatch statically.

**`ArcMessenger` is not deleted here.** Removing it also means rewriting two card-only test files that construct it directly, which deserves a hardware run as its own change. After this it has zero production callers.

## List of the changes

- 16 call sites moved to `send_device_command`.
- `send_arc_command()` adapter local to each SPI device.

## Testing

Simulation on and off; `baremetal_tests` 254/254. The ARC paths need a card — `test_arc_messages_bh/wh` and the SPI tests cover them.

## API Changes

None. `TTDevice::get_arc_messenger()` remains but is unused in production.

### DIFF
--- a/.github/workflows/build-and-run-all-tests.yml
+++ b/.github/workflows/build-and-run-all-tests.yml
@@ -154,6 +154,7 @@ jobs:
         # - P300-viommu is a shared pool of 3 P300 machines.
         # - bh-loudbox is a single machine in CIv2. It also has an additional problem where
         #   it actually has a ~20min startup time for viommu variant, so this one can be really slow.
+        #   It recently became congested, so it is blacklisted from merge queue for now, but will still run on main.
         # - The wh-glx-6u is also one machine, but not yet fully proven, so don't want to block merge queue with it.
         # - 6u is a single machine and the main merge queue bottleneck, so in the merge queue it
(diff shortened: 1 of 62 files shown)


nbuncicTT:
Closing in favor of #3322 — the command path now lands together with its production routing.

PR #3305, opened 2026-08-28
[Base API] Add IoWindow factory (TLBManager removal, Phase 1)

ajovanovicTT:
Issue

Description
Phase 1 of removing TLBManager from UMD public API, revised in scope. The team decided static TLBs are not needed at all once tt-metal moves off them, since regular write_to_device is equally fast. This PR adds the one factory tt-metal needs to create its own RAII windows, and drops the lookup added earlier in this PR.

List of the changes
 - Added TTDevice::create_io_window(TargetIoWindowConfig, HostIoWindowConfig), a spec shaped factory returning unique_ptr to IoWindow. Rounds a requested size up to the smallest architecture size class that covers it, and defaults the NOC to the one selected for the calling thread when the target does not name one. Multicast targets take the same NOC translation guard noc_multicast_write has.
 - Added Cluster::create_io_window(ChipId, CoreCoord core_start, addr, HostIoWindowConfig, optional core_end, WindowFlags, optional NocId), the chip and core level counterpart tt-metal will call directly. A multicast grid, transaction flags or a NOC can be named at creation instead of reconfiguring the window afterwards. Parameters are ordered so the common unicast call passes only chip, core, addr and size. Returns nullptr for mock and emulated chips, which have no device to map.
 - Added make_io_window_target, a free function over SocDescriptor that builds a TargetIoWindowConfig from coordinates in any coordinate system. Cluster uses it to translate cores against the chip mapping, and any caller holding a SocDescriptor can reach the TTDevice factory the same way. TargetIoWindowConfig itself is unchanged and still names its cores in the translated system.
 - get_static_tlb_window, get_tlb_configuration, get_tlb_manager, and TLBManager itself are untouched and will be removed in a later phase once tt-metal migrates onto create_io_window.
 - Clarified the IoWindow::get_size doc comment: it reports bytes accessible from the current window base, which can be less than the backing mapping size after reconfiguring to a less aligned address.

Testing
CI. Added TestTlb.CreateIoWindow covering both factories: size round up, NOC defaulting, target round trip, and the too large request throw. Added TestTlb.CreateMulticastIoWindow, which writes once through a grid window and reads both corners back.

API Changes
New public API: TTDevice::create_io_window, Cluster::create_io_window, and make_io_window_target in the new header umd/device/io_window/io_window_target.hpp. Chip::get_static_io_window and Cluster::get_static_io_window, added earlier in this PR, are removed before merge. No other existing API removed or changed.

### DIFF
--- a/device/api/umd/device/cluster.hpp
+++ b/device/api/umd/device/cluster.hpp
@@ -281,21 +281,21 @@ public:
         uint64_t ordering = tlb_data::Relaxed);
 
     /**
-     * Maps a core into host address space, anchored at an address on that core. Reads and writes
-     * through the returned window address it as offsets from that anchor. The window is created
-     * large enough to cover the requested size, rounded up to a size the architecture provides.
+     * Maps a target into host address space, anchored at the address the target names. Reads and
+     * writes through the returned window address it as offsets from that anchor. The window is
+     * created large enough to cover the requested size, rounded up to a size the architecture
+     * provides.
      *
      * The caller owns the window: the mapping is released when it is destroyed, so it must be held
      * for as long as it is used, and it must not outlive this Cluster. Returns nullptr for chips
      * with no device behind them (mock and emulated).
      *
      * @param chip Device to target.
-     * @param core Core to map.
-     * @param addr Address on the core the window is anchored at.
+     * @param target Device-side target describing the core(s), address, optional NOC and flags.
      * @param host Host-side window properties (caching strategy and requested size).
      */
     std::unique_ptr<IoWindow> create_io_window(
-        const ChipId chip, const CoreCoord core, uint64_t addr, HostIoWindowConfig host = {});
+        const ChipId chip, TargetIoWindowConfig target, HostIoWindowConfig host = {});
 
     /**
      * Pass in ethernet cores with active links for a specific MMIO chip. When called, this function will force UMD to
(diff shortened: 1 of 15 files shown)


device/api/umd/device/cluster.hpp
```diff
@@ -279,6 +280,23 @@ class Cluster {
...
+     * for as long as it is used, and it must not outlive this Cluster. Returns nullptr for chips
+     * with no device behind them (mock and emulated).
+     *
+     * @param chip Device to target.
+     * @param core Core to map.
+     * @param addr Address on the core the window is anchored at.
+     * @param host Host-side window properties (caching strategy and requested size).
+     */
+    std::unique_ptr<IoWindow> create_io_window(
+        const ChipId chip, const CoreCoord core, uint64_t addr, HostIoWindowConfig host = {});
```
nbuncicTT: Why don't we have `TargetIoWindowConfig` in the API itself? We have less flexibility when we create the IoWindow and it'll surely be followed by a `configure` API invocation on the window...

nbuncicTT (approved):

device/api/umd/device/cluster.hpp
ajovanovicTT: Synced on this. Just changed the order of agreed parameters so that default values have to be passed at least possible explicitly currently.

PR #3306, opened 2026-08-28
[base API] Answer arch and communication id from TTDeviceModel components

broskoTT:
### Issue
Related to #2864

### Description
TTDeviceModel carried get_arch() and get_communication_device_id() as temporary getters,
each with a TODO saying it would be answered by a real component once that component moved
into the model. Both components are now there, so the getters go away and TTDevice asks the
components directly.

Note that get_communication_device_id() changes when it resolves, not what it returns. The
models captured the id at construction; RemoteProtocol::get_mmio_id() reaches through to the
local device on every call instead. That is the same reach-through that caused the teardown
segfault on #3236: the local device now has to outlive every remote device routed through it.
Taken deliberately, to keep identity in one place rather than copied into each model.

### List of the changes
- Drop get_arch() and get_communication_device_id() from TTDeviceModel and its three models,
  along with the now-unused communication_device_id_ and arch_ members.
- TTDevice::get_arch() delegates to get_architecture_impl()->get_architecture().
- TTDevice::get_communication_device_id() delegates to get_device_protocol()->get_mmio_id(),
  guarding the null protocol a simulation model reports.

### Testing
Existing CI. No behavioural change for PCIe and JTAG: PcieProtocol::get_mmio_id() returns the
same PCI device number the models captured, and JtagProtocol::get_mmio_id() the same JLink id.

### API Changes
This PR has API changes: two getters are removed from the TTDeviceModel interface. They were
never part of the Base API spec.

### DIFF
--- a/device/tt_device/tt_device.cpp
+++ b/device/tt_device/tt_device.cpp
@@ -83,9 +83,10 @@ void TTDevice::init_tt_device(const std::chrono::milliseconds timeout_ms) {
         is_pcie_hung();
     }
     // The hang detector is an optional component, so a model that provides none simply skips the check.
-    HangDetector *hang_detector = model_->get_hang_detector();
+    // Going through is_noc_hung() rather than the detector directly picks up its warning for a
+    // protocol that cannot run the check at all.
     const NocId hang_check_noc = is_selected_noc1() ? NocId::NOC1 : NocId::NOC0;
-    if (hang_detector != nullptr && hang_detector->is_noc_hung(hang_check_noc).value_or(false)) {
+    if (model_->get_hang_detector() != nullptr && is_noc_hung(hang_check_noc, HangAction::RETURN)) {
         UMD_THROW(error::NocHangError, *this, hang_check_noc);
     }
     // Waits for the firmware and builds the components that read what it publishes; the model's
--- a/device/tt_device/tt_device.cpp
+++ b/device/tt_device/tt_device.cpp
@@ -22,6 +22,7 @@
 #include "tracy.hpp"
 #include "umd/device/arc/arc_telemetry_reader.hpp"
 #include "umd/device/arc/firmware_telemetry_reader.hpp"
+#include "umd/device/arch/architecture_implementation.hpp"
 #include "umd/device/arch/architecture_tlbs.hpp"
 #include "umd/device/driver_atomics.hpp"
 #include "umd/device/jtag/jtag_device.hpp"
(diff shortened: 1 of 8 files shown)


device/api/umd/device/tt_device/protocol/pcie_protocol.hpp
aleksamarkovicTT, line 117: Why rename from set? I think set makes more sense.

broskoTT: It literally says "wired in" in the comment. Maybe "configure" is a better name, but set is a bit misleading given the actual hang detector class is not passed in this function

device/tt_device/tt_device.cpp
```diff
@@ -73,17 +75,19 @@ TTDevice::TTDevice(std::unique_ptr<TTDeviceModel> model) : model_(std::move(mode
...
         is_pcie_hung();
     }
-    bool noc_hang_check_result =
-        hang_detector_->is_noc_hung(is_selected_noc1() ? NocId::NOC1 : NocId::NOC0).value_or(false);
-    if (noc_hang_check_result) {
-        UMD_THROW(error::NocHangError, *this, is_selected_noc1() ? NocId::NOC1 : NocId::NOC0);
+    // The hang detector is an optional component, so a model that provides none simply skips the check.
+    HangDetector *hang_detector = model_->get_hang_detector();
+    const NocId hang_check_noc = is_selected_noc1() ? NocId::NOC1 : NocId::NOC0;
+    if (hang_detector != nullptr && hang_detector->is_noc_hung(hang_check_noc).value_or(false)) {
```
nbuncicTT: I think you should evaluate if the hang detection is supported, it's not a requirement as it's return value is optional. If you think it's required, then leave it like this (I'm not saying for our particular case, we will solve this in our concrete class in the future where internally we'll make it required).

nbuncicTT (approved):

device/api/umd/device/tt_device/protocol/pcie_protocol.hpp
aleksamarkovicTT: enable_hang_detector?

broskoTT: discussed offline, we can't find better name than wire currently

device/tt_device/tt_device.cpp
broskoTT: warning is already embedded inside is_noc_hung.
Won't throw if it returns nullopt

Issue #3307, opened 2026-08-28
Galaxy x32 ttsim binaries take the legacy per-chip dlopen path


broskoTT:
### Description

The Galaxy ttsim binaries (`libttsim_wh_x32.so`, `libttsim_bh_x32.so`) fall through to
the legacy per-chip memfd + dlopen path in `TTSimCommunicator::initialize()`, so a
32-chip cluster ends up with **32 independent dlopens of a 32-chip simulator** instead
of one shared instance.

Neither of the two shared-image paths matches:

- The multichip ABI probe checks for `libttsim_create_device_by_id` /
  `libttsim_select_device_by_id`. The x32 builds export neither -- their dynamic symbol
  table is byte-identical to the single-chip build:

  ```
  $ diff <(nm -D --defined-only libttsim_wh.so | awk '{print $3}' | sort) \
         <(nm -D --defined-only libttsim_wh_x32.so | awk '{print $3}' | sort)
  # (no output)
  ```

- `self_describing_multi_mmio` requires a `cluster_descriptor.yaml` beside the `.so`.
  tt-metal's `setup-ttsim` action does not place one -- the topology is supplied through
  `TT_METAL_MOCK_CLUSTER_DESC_PATH` -- so this is false as well.

Confirmed in tt-metal's `runtime_sim_galaxy_cpp_unit_tests (wormhole_b0)` job: the
container runs with `LOGURU_LEVEL=INFO`, 32 `Creating Simulation device` lines are
logged, and neither `TTSim multichip mode enabled` nor `TTSim BDF multichip mode`
appears.

### Impact

It works today -- those jobs are green -- but each dlopen carries the full Galaxy
zero-fill BSS, and the multichip modelling in the x32 image is unused:

```
$ readelf -S -W libttsim_wh_x32.so   # .lbss  3.97 GB
$ readelf -S -W libttsim_bh_x32.so   # .lbss  7.25 GB
```

Nothing faults those pages, so RSS stays low, but 32 copies of a 32-chip simulator is
not what either side intends. Chips are also isolated from each other, so the eth links
compiled into the x32 image never carry traffic.

### Notes

Found while switching UMD's ttsim CI galaxy legs onto the x32 binaries, which is what
tt-metal already uses (tenstorrent/tt-umd PR linked below). Worth deciding with the
ttsim owners whether the fix belongs in UMD (recognise an x32 image without a
[...]

PR #3308, opened 2026-08-28
[CI] Run the ttsim galaxy tests against the x32 simulators

broskoTT:
### Issue
#3307 (related -- surfaced by this change, not fixed by it)

### Description
The two galaxy legs ran the *single-chip* `libttsim_wh.so` / `libttsim_bh.so`
under a 32-chip 6U cluster descriptor, so UMD replicated one single-chip
simulator per chip instead of driving a real Galaxy. ttsim publishes
`libttsim_wh_x32.so` / `libttsim_bh_x32.so` for exactly this topology (32 chips,
all MMIO, inter-chip eth links baked into the .so), and that is what tt-metal's
own `runtime_sim_galaxy_cpp_unit_tests` legs use.

The point of these legs is to run tt-metal's galaxy tests the way tt-metal runs
them, against the newest UMD. Pointing them at the x32 binaries and adding the
one env var tt-metal's `setup-ttsim` sets that we were missing makes them match
one-for-one, so a failure here means a real regression rather than a
configuration we drive differently from everyone else.

This also clears the current CI failure: on the single-chip build
`MeshDeviceFixture.TensixInlineWriteDynamicNoc` aborts with
`UndefinedBehavior: rv32_tti: tile_type=E`, while the same test on the same
tt-metal commit passes against x32 in tt-metal's CI.

### List of the changes
- Download `libttsim_wh_x32.so` and `libttsim_bh_x32.so` and stage them in
  `sim_wh_x32` / `sim_bh_x32` with the matching SoC descriptor. No
  `cluster_descriptor.yaml` beside them -- the topology comes from
  `TT_METAL_MOCK_CLUSTER_DESC_PATH`, as in tt-metal.
- Point both galaxy steps at those directories and set
  `TT_METAL_DISABLE_SFPLOADMACRO=1`.

### Testing
CI should show if this diff fixes the issue.

Switching to x32 surfaced #3307: neither shared-image path in
`TTSimCommunicator::initialize()` matches an x32 binary, so the Galaxy legs take the
legacy per-chip dlopen path. That is pre-existing and equally true of tt-metal's own
galaxy legs, so it is filed separately rather than fixed here.

### API Changes
This PR has no API changes.

### DIFF
--- a/.github/workflows/run-ttsim-tests.yml
+++ b/.github/workflows/run-ttsim-tests.yml
@@ -91,6 +91,22 @@ jobs:
           fileName: "libttsim_wh_x2.so"
           out-file-path: /tmp/ttsim-wh-x2
 
+      - name: Download ttsim binary (wh_x32)
+        uses: robinraju/release-downloader@28fc21f50d76778e7023361aa1f863e717d3d56f  # v1.13
+        with:
+          repository: tenstorrent/ttsim
+          latest: true
+          fileName: "libttsim_wh_x32.so"
+          out-file-path: /tmp/ttsim-wh-x32
+
+      - name: Download ttsim binary (bh_x32)
+        uses: robinraju/release-downloader@28fc21f50d76778e7023361aa1f863e717d3d56f  # v1.13
+        with:
+          repository: tenstorrent/ttsim
+          latest: true
+          fileName: "libttsim_bh_x32.so"
+          out-file-path: /tmp/ttsim-bh-x32
+
       - name: Build tt-metal
         run: |
           cd tt-metal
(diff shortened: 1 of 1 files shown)


aleksamarkovicTT (approved):

.github/workflows/run-ttsim-tests.yml
```diff
@@ -91,6 +91,22 @@ jobs:
           fileName: "libttsim_wh_x2.so"
           out-file-path: /tmp/ttsim-wh-x2
 
+      - name: Download ttsim binary (wh_x32)
```
aleksamarkovicTT, line 94: Are sim_wh and sim_bh still used ? if not, remove

broskoTT: they are. In the tests that hit a single chip

Issue #2872, opened 2026-06-22
Create DeviceFirmware interface based on Base API Specification


nbuncicTT:
`DeviceFirmware` does not exist as a standalone component. Create it per the Base API Specification and migrate relevant behavior from existing APIs and components into it. Existing components that are fully absorbed into `DeviceFirmware` should be deleted.

**Scope**

- Create `DeviceFirmware` interface per the Base API Specification
- Identify and migrate firmware-related behavior from existing components
- Remove any components fully absorbed into `DeviceFirmware`
- Implement for supported device backends

**Timeline**

No hard deadline. Part of the UMD Base API migration epic.

**Is your feature request related to a problem? Please describe.**

Firmware management behavior is spread across existing components rather than encapsulated in a dedicated interface.

**Definition of Done**

- `DeviceFirmware` interface exists per the Base API Specification
- Relevant behavior migrated from existing components
- Absorbed components are deleted
- Existing tests pass or are updated

PR #3318, opened 2026-08-31
[Base API] Add the DeviceFirmware seam

nbuncicTT:
## Issue

#2872
The `DeviceFirmware` implementations were removed in #3317 to be re-landed one function at a time. Something has to exist for the first function to land into.

## Description

The seam, and nothing else: an interface with no methods, concrete classes that hold and validate their dependencies, and ownership in the right place.

Per the spec, the firmware is a component of `TTDeviceModel` — each concrete model creates and owns its own implementation, exactly like the protocol and the architecture implementation. The model knows its architecture and backend statically, so no dispatch picks the implementation, which rules out the virtual-call-in-constructor trap by construction (the bug that segfaulted ttsim in the previous attempt).

Every subsequent PR adds one method here together with the `TTDevice` code it deletes, so each is reviewable as a move.

## List of the changes

- `DeviceFirmware` (empty interface), `BlackholeDeviceFirmware`, `WormholeDeviceFirmware`, `SimulationDeviceFirmware` (shared by all sim backends until one needs to differ).
- `TTDeviceModel::get_device_firmware()`; each concrete model constructs and owns its firmware.
- Constructors validate every dependency before use.

## Testing

Simulation on and off; `baremetal_tests` 238/238; no undefined symbols. Behaviour cannot change — nothing calls any of it yet.

## API Changes

- `TTDeviceModel::get_device_firmware()` added (pure virtual).

Stacked on #3317.

### DIFF
--- a/device/tt_device/firmware/blackhole_device_firmware.cpp
+++ b/device/tt_device/firmware/blackhole_device_firmware.cpp
@@ -1,0 +1,35 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "umd/device/tt_device/firmware/blackhole_device_firmware.hpp"
+
+#include "umd/device/tt_device/protocol/device_protocol.hpp"
+#include "umd/device/tt_device/protocol/jtag_interface.hpp"
+#include "umd/device/tt_device/protocol/pcie_interface.hpp"
+#include "umd/device/utils/error.hpp"
+
+namespace tt::umd {
+
+BlackholeDeviceFirmware::BlackholeDeviceFirmware(
+    DeviceProtocol* device_protocol,
+    PcieInterface* pcie_interface,
+    JtagInterface* jtag_interface,
+    ArchitectureImplementation* architecture_impl) :
+    device_protocol_(device_protocol),
+    pcie_interface_(pcie_interface),
+    jtag_interface_(jtag_interface),
+    architecture_impl_(architecture_impl) {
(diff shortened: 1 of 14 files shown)


device/tt_device/firmware/blackhole_device_firmware.cpp
```diff
@@ -0,0 +1,35 @@
...
+    PcieInterface* pcie_interface,
+    JtagInterface* jtag_interface,
+    ArchitectureImplementation* architecture_impl) :
+    device_protocol_(device_protocol),
+    pcie_interface_(pcie_interface),
+    jtag_interface_(jtag_interface),
+    architecture_impl_(architecture_impl) {
+    UMD_ASSERT(device_protocol_ != nullptr, error::RuntimeError, "BlackholeDeviceFirmware requires a DeviceProtocol.");
+    UMD_ASSERT(
+        architecture_impl_ != nullptr,
```
broskoTT, line 25: Not to change in this PR, but a general code style comment for us to think about in the future.

If we don't ever want this being nullptr, maybe passing a pointer is not the right solution, i.e. we can pass a reference? Not sure what are all the implications of passing a reference to an object held by unique_ptr though, it might be worse to do that... Just food for thought

broskoTT (approved):

device/tt_device/firmware/blackhole_device_firmware.cpp
nbuncicTT: Essentially, that's a rabbit hole - going that path creates tight coupling to the owner of this object which has to keep the object being referred to valid during the whole lifetime of the object which holds the reference - this is why reference semantics fail in comparison to value semantics.

I'm more in favor of checking a nullptr for now, but I want to definitely revisit this in the future for a better design.

PR #3321, opened 2026-08-31
Add an opt-in unity build option

nbuncicTT:
## Issue

Compile time dominates our CI. `device` has 156 headers against 105 sources, 144 of them public API, and every source re-parses a large slice of that surface. On the merge queue the `build-device` matrix runs 26 to 33 minutes per entry.

## Description

Adds the option, nothing more. `TT_UMD_ENABLE_UNITY_BUILD` defaults to OFF, so **this PR changes no build anywhere** — it only makes batching available to turn on.

Unity builds concatenate several sources into one translation unit, so header parsing is paid once per batch instead of once per source. `CMAKE_UNITY_BUILD` is set *after* `add_subdirectory(third_party)`, so dependencies keep building one source at a time either way.

Split out of the original PR at review request: the test changes that batching requires, and the CI switch that turns it on, are in the follow-up. This PR is deliberately just the mechanism so it can be reviewed on its own.

Measured effect of turning it on, for context (cold `Build All` on real runners, full matrix): 1.35x to 3.74x, median ~2.4x, worst entry 21m27s -> 8m50s. Numbers and the batch-size sweep are in the follow-up.

## List of the changes

- `TT_UMD_ENABLE_UNITY_BUILD` (default OFF) and `TT_UMD_UNITY_BUILD_BATCH_SIZE` (default 8), set after `third_party`.
- Both added to the configuration summary.

## Testing

Configured and built both ways on this branch, clang-20, Release, `TT_UMD_BUILD_ALL=ON`: default (OFF) is byte-for-byte the same build as before, 463 edges green; explicit `-DTT_UMD_ENABLE_UNITY_BUILD=ON` configures 44 unity blobs and builds green.

## API Changes

This PR has no API changes, and with the default off it does not change how the shipped library is compiled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -114,6 +114,22 @@ endif()
 
 add_subdirectory(third_party)
 
+# Unity builds batch several sources into one translation unit, so the cost of parsing
+# our headers is paid once per batch instead of once per source. Set AFTER third_party
+# so dependencies keep building one source at a time.
+#
+# Opt-in and off by default, so this commit changes nothing on its own. Turning it on also
+# needs the tests to be batching-safe, which is a follow-up; and it changes how the library
+# is compiled (static-initialisation order within a batch, inlining across what used to be
+# separate translation units), which is not something to hand a release build or a
+# developer's incremental loop without asking.
+option(TT_UMD_ENABLE_UNITY_BUILD "Batch UMD sources into unity translation units" OFF)
+set(TT_UMD_UNITY_BUILD_BATCH_SIZE "8" CACHE STRING "Number of sources per unity translation unit")
+if(TT_UMD_ENABLE_UNITY_BUILD)
+    set(CMAKE_UNITY_BUILD ON)
+    set(CMAKE_UNITY_BUILD_BATCH_SIZE "${TT_UMD_UNITY_BUILD_BATCH_SIZE}")
+endif()
+
 # Create a target that aggregates all generated files needed for code analysis.
 # Subdirectories can add dependencies to this target as needed.
 add_custom_target(umd_all_generated_files)
@@ -211,6 +227,8 @@ set(_umd_summary_vars
     TT_UMD_ENABLE_CLANG_TIDY
     TT_UMD_ENABLE_LOGGING
     TT_UMD_ENABLE_TRACY
+    TT_UMD_ENABLE_UNITY_BUILD
+    TT_UMD_UNITY_BUILD_BATCH_SIZE
 )
 message(STATUS "")
 message(STATUS "==================== UMD configuration summary ====================")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -124,7 +124,10 @@ add_subdirectory(third_party)
 # separate translation units), which is not something to hand a release build or a
 # developer's incremental loop without asking.
 option(TT_UMD_ENABLE_UNITY_BUILD "Batch UMD sources into unity translation units" OFF)
-set(TT_UMD_UNITY_BUILD_BATCH_SIZE "8" CACHE STRING "Number of sources per unity translation unit")
+# 16 from a sweep of 1..24 at -j4, the parallelism a CI runner actually has. The curve is
+# almost spent by 8 (2.56x of an available 3.01x) and 16 captures 97% of it; 24 buys 3% more
+# while leaving libtt-umd only 5 units, so the tail idles cores on a 4-core runner.
+set(TT_UMD_UNITY_BUILD_BATCH_SIZE "16" CACHE STRING "Number of sources per unity translation unit")
 if(TT_UMD_ENABLE_UNITY_BUILD)
     set(CMAKE_UNITY_BUILD ON)
     set(CMAKE_UNITY_BUILD_BATCH_SIZE "${TT_UMD_UNITY_BUILD_BATCH_SIZE}")


.github/workflows/build-device.yml
```diff
@@ -16,6 +16,13 @@ on:
           Enable sccache compiler caching.
         type: boolean
         default: true
+      use-unity:
```
aleksamarkovicTT: use-unity-build or enable-unity-build

aleksamarkovicTT (approved):

.github/workflows/build-device.yml
```diff
@@ -34,6 +41,12 @@ env:
   CREATE_MAP_BINARIES_DIR: ./device/bin/silicon
   INSTALL_OUTPUT_DIR: ./install
   WHEEL_OUTPUT_DIR: ./build/wheel
+  # Batch sources into unity translation units, for the merge queue only.
```
broskoTT: ugh... different build configs for merge queue and others... I'd rather not

tests/baremetal/test_soc_descriptor.cpp
```diff
@@ -643,7 +642,7 @@ TEST(SocDescriptor, AllSocDescriptors) {
 
         auto arch = SocDescriptor::get_arch_from_soc_descriptor_path(soc_desc_yaml);
         HarvestingMasks harvesting_masks = {
-            .eth_harvesting_mask = (arch == tt::ARCH::BLACKHOLE) ? example_eth_harvesting_mask : 0};
+            .eth_harvesting_mask = (arch == tt::ARCH::BLACKHOLE) ? example_eth_harvest_mask : 0};
```
broskoTT: Yeah obviously please drop all other refactors from this PR

CMakeLists.txt
```diff
@@ -114,6 +114,23 @@ endif()
...
+# so dependencies keep building one source at a time.
+#
+# Opt-in, and off by default. It is currently only turned on for the cold matrix builds in
+# the merge queue, where it is worth ~2-3x. Everything else is left alone deliberately:
+# builds that get real value from the compiler cache do not need it, and it changes how
+# the library is compiled (static-initialisation order within a batch, inlining across
+# what used to be separate translation units), which is not something to hand a release
+# build or a developer's incremental loop without asking.
+option(TT_UMD_ENABLE_UNITY_BUILD "Batch UMD sources into unity translation units" OFF)
+set(TT_UMD_UNITY_BUILD_BATCH_SIZE "8" CACHE STRING "Number of sources per unity translation unit")
```
broskoTT: Maybe try driving this 1-20 and post the results, both for local build and CI build

tests/baremetal/test_soc_descriptor.cpp
nbuncicTT: The refactors are necessary for the unity builds.

.github/workflows/build-device.yml
nbuncicTT: Renamed to `enable-unity-build`, matching `TT_UMD_ENABLE_UNITY_BUILD`. It moved to the follow-up (#3390) along with the rest of the CI wiring.

CMakeLists.txt
```diff
@@ -114,6 +114,23 @@ endif()
...
+if(TT_UMD_ENABLE_UNITY_BUILD)
+    set(CMAKE_UNITY_BUILD ON)
+    set(CMAKE_UNITY_BUILD_BATCH_SIZE ${TT_UMD_UNITY_BUILD_BATCH_SIZE})
+endif()
```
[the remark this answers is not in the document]
nbuncicTT: Done, quoted. It cannot bite with a numeric value, but it costs nothing to be correct.

tests/baremetal/test_soc_descriptor.cpp
nbuncicTT: Split. This PR is now `CMakeLists.txt` only - the option, default OFF, +18 lines, changes no build anywhere.

The test changes and the CI switch are in #3390, stacked on this one. To be explicit about why they cannot simply be dropped: batching is what *creates* the collisions, so the two have to land in this order - the option is inert on its own, and CI cannot turn batching on until the tests are batching-safe. A PR that enabled it without the test changes would fail its own merge-queue run.

tests/baremetal/test_blackhole_arc_apb.cpp
```diff
@@ -72,7 +72,7 @@ TEST_P(BlackholeArcApbTest, ReadOverJtagGoesThroughDeviceProtocol) {
     EXPECT_CALL(protocol_, read_ctrl(_, arc_core(), noc_address(), sizeof(uint32_t), noc()));
 
     uint32_t value = 0;
-    apb.read(&value, APB_OFFSET, sizeof(uint64_t), arc_core(), noc());
+    apb.read(&value, BLACKHOLE_APB_OFFSET, sizeof(uint64_t), arc_core(), noc());
 }
```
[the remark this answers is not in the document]
nbuncicTT: I do not think this one is right - it would quietly gut the test.

The comment directly above says the access goes "one word at a time regardless of the requested size", and the test exists to prove exactly that:

```cpp
EXPECT_CALL(protocol_, read_ctrl(_, arc_core(), noc_address(), sizeof(uint32_t), noc()));
uint32_t value = 0;
apb.read(&value, BLACKHOLE_APB_OFFSET, sizeof(uint64_t), arc_core(), noc());
```

It requests 8 bytes and asserts the protocol is called with 4. Changing the request to `sizeof(value)` makes it "request 4, expect 4", which asserts nothing. Not a buffer risk either: the implementation only writes one word, which is the behaviour being asserted.

Leaving as is. This PR only renamed a constant in that file; the `sizeof` predates it.

.github/workflows/build-device.yml
nbuncicTT: Agreed, and this is not hypothetical - it is exactly how the gcc `-Werror=subobject-linkage` failures got through. They were invisible on the PR and only appeared once the queue compiled differently, which is the worst place to find them.

Changed in #3390: batching is now on for every automatic event in `build-device`, and in `build-tests` and `run-ttsim-tests` too, so PRs, the merge queue and main all compile the same way and a batching-only failure surfaces on the PR that introduced it.

It costs nothing on the cached lanes. Measured with a warm cache, whole tree, clang-tidy on: unity 36.8s against per-file 100.9s, both at 100% cache hits. clang-tidy runs alongside the compile under `cmake -E __run_co_compile` and is never served from cache, so batching lowers a floor no cache can reach. One build after merge repopulates the cache under the new keys.

`release.yml` is left alone - it ships artifacts, and the option defaults OFF.

nbuncicTT:
Batch size sweep, as requested. Two curves, because the answer differs between a dev box and a runner.

**At `-j4`, the parallelism a GitHub runner actually has.** clang-tidy on, `TT_UMD_BUILD_ALL=ON`, Release, clang-20, whole tree:

| batch | wall | speedup | marginal | tree CPU | libtt-umd units | peak RSS |
| --- | --- | --- | --- | --- | --- | --- |
| off | 599.2s | 1.00x | - | 2213s | 104 | 468M |
| 4 | 309.5s | 1.94x | +93.6% | 1010s | 26 | 618M |
| 8 | 234.5s | 2.56x | +32.0% | 748s | 13 | 680M |
| 12 | 216.1s | 2.77x | +8.5% | 681s | 9 | 667M |
| 16 | 206.0s | 2.91x | +4.9% | 637s | 7 | 732M |
| 24 | 198.7s | 3.01x | +3.3% | 594s | 5 | 734M |

**At `-j16` on a 64-core box**, for contrast, showing every value 1..24:

| batch | 1 | 2 | 3 | 4 | 6 | 8 | 10 | 12 | 16 | 20 | 24 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| wall | 158.7 | 133.5 | 96.3 | 81.7 | 74.2 | **70.9** | **70.2** | 71.7 | 77.1 | 76.4 | 72.1 |
| tree CPU | 2243 | 1795 | 1156 | 1000 | 844 | 762 | 720 | 691 | 645 | 620 | 603 |
| units | 104 | 52 | 35 | 26 | 18 | 13 | 11 | 9 | 7 | 6 | 5 |

Note the local wall time gets *worse* past batch 10 while CPU time keeps falling. That is a parallelism artifact of the wide box: at batch 16 `libtt-umd` is only 7 units, fewer than the 16 cores, so cores idle. It does not happen at `-j4`, which is why the first table is the one to pick from.

**Picked 16.** The curve is almost entirely spent by 8 (2.56x of an available 3.01x) and 16 captures 97% of the total. Going to 24 buys 3% more wall while leaving `libtt-umd` at 5 units, so on 4 cores the tail idles three of them; 7 units packs as 4+3. Memory is not a constraint either way - 732M peak times 4 concurrent is ~2.9GB against 16GB runners.

12 would be the conservative alternative at 2.77x, if we want to limit how much batching hurts cache granularity and incremental rebuilds.

Two caveats on the data, since they would otherwise mislead:

- Batch 12 in the first `-j4` pass read 279.3s / 797.7s tree CPU, i.e. *worse* than batch 8, which is impossible for a monotonic quantity. Re-running gave 216.1s / 681.3s; 8, 16 and 24 all reproduced within 1-4%, so only that point was bad. The table uses the re-run.
- `off` is absent from the `-j16` table because I corrupted that build by touching the working tree while it was in flight. Batch 1 covers it: 2243s against `off`'s 2213s at `-j4`, 1.4% apart, so batch=1 is a sound proxy and the 1->N ratios are real.

CMakeLists.txt
nbuncicTT: Swept 1..24, results posted as a PR comment. Short version: at `-j4` (runner parallelism) the curve is 1.94x at batch 4, 2.56x at 8, 2.91x at 16, 3.01x at 24 - almost spent by 8. Default changed 8 -> 16, which captures 97% of the available win; 24 buys 3% more but leaves libtt-umd only 5 units, so the tail idles cores on 4 of them.

Worth knowing: on a wide box the local wall time gets *worse* past batch 10 even though CPU time keeps falling, because there end up being fewer units than cores. That is why the decision comes from the -j4 numbers rather than the local ones.

CMakeLists.txt
```diff
@@ -114,6 +114,25 @@ endif()
 
 add_subdirectory(third_party)
 
+# Unity builds batch several sources into one translation unit, so the cost of parsing
+# our headers is paid once per batch instead of once per source. Set AFTER third_party
+# so dependencies keep building one source at a time.
+#
+# Opt-in and off by default, so this commit changes nothing on its own. Turning it on also
```
aleksamarkovicTT, line 121: Remove ephemeral development related comments and benchmark results.

PR #3322, opened 2026-08-31
[Base API] Route all ARC commands through DeviceFirmware::send_device_command and delete ArcMessenger

nbuncicTT:
## Issue

#2872
Third move onto the `DeviceFirmware` seam: the firmware command path — and the `ArcMessenger` it replaces dies in this same diff.

## Description

`DeviceFirmware::send_device_command()` arrives wired into production, and **every** `ArcMessenger` caller moves onto it here — including the call sites inside `TTDevice` overrides that later PRs delete (`set_clock_state` on both arches, Wormhole's `get_chip_info` harvesting query), the card tests, and the Python `arc_msg` bindings. With zero callers left, the whole `ArcMessenger` family is deleted in this diff: no window where two copies of the send protocol coexist.

This lands before the remaining delegations (`set_clock_state`, `get_chip_info`, …) because their firmware implementations are built on this call — the same dependency order the original stack used.

Implementation notes, preserved from the reviewed originals:

- **Blackhole** sends through `BlackholeArcMessageQueue`, built in `init_firmware()` — the queue descriptor is published by the ARC firmware while it boots, so that is the earliest it can be read. Same `ARC_MSG` mutex key the messenger used (its device-number argument defaulted the type to PCIe).
- **Wormhole** sends through the APB window scratch registers with the messenger's exact semantics: same arg packing, ethernet flush before the trigger for remote devices, the **same lock set** (per-device mutex — `REMOTE_ARC_MSG` keyed on the gateway for remotes — plus the deprecated system-wide one kept for cross-version exclusion), and the **post-message PCIe-hang tripwire** (same register `WormholeHangDetector` reads; throws `PcieHangError` with identity fields via #3310).
- Commands before `init_firmware()` throw `UninitializedDeviceError`. The Blackhole card tests used to work without init because the messenger built its own queue — a bypass around that guard; they now bring the firmware up explicitly.
- Every converted caller passes the **thread-selected NOC**, which the messenger read itself.
- **Simulation** reports success with no return values, so shared flows that issue commands unconditionally keep working.
- **`WormholeArcWindow::write()` is implemented here** — the command path is its first writer — and its tests arrive with it, the same behavior-migration shape as `read()` arriving with the boot wait (#3346).
- The SPI adapter lives in one shared internal header: tt-metal compiles UMD as a unity build, where per-file anonymous-namespace copies collide (caught by run-ttsim-tests).

## List of the changes

- `DeviceCommandResult` + `DeviceFirmware::send_device_command()`; implemented in WH, BH, simulation.
- `BlackholeDeviceFirmware` owns `arc_msg_queue_`, built in `init_firmware()`.
- All `ArcMessenger` callers converted: `Chip::arc_msg`, both SPI devices, warm reset's legacy flow, WH `get_clock`/`configure_iatu_region`/`get_chip_info`(harvesting)/`set_clock_state`, BH `retrain_dram_core`/`set_clock_state`, `set_tdp_limit`/`restore_default_tdp_limit`, both ARC-message card test suites, the three Python `arc_msg` bindings.
- Deleted: `arc_messenger.{hpp,cpp}`, `wormhole_arc_messenger.{hpp,cpp}`, `blackhole_arc_messenger.{hpp,cpp}`, `TTDevice::get_arc_messenger()`, `arc_messenger_`.

## Testing

`baremetal_tests` 254/254, simulation on and off, Python bindings build, no undefined symbols, unity-build check clean. The converted suites are card tests — validated by card CI.

## API Changes

- Added: `DeviceFirmware::send_device_command`, `DeviceCommandResult`.
- Removed: `TTDevice::get_arc_messenger()`, `ArcMessenger`/`WormholeArcMessenger`/`BlackholeArcMessenger` (no users in tt-metal/tt-exalens; tests and Python bindings converted in this PR).

Stacked on #3320. (Absorbed the previously separate #3329 per review — delete-as-we-delegate, single diff.)


device/tt_device/blackhole_tt_device.cpp
```diff
@@ -188,13 +188,25 @@ uint32_t BlackholeTTDevice::get_min_clock_freq() { return get_architecture_imple
 
 void BlackholeTTDevice::set_clock_state(TTDevice::PowerState state, NocId /*noc_id*/) {
     ZoneScoped;
-    int exit_code = 0;
+    uint32_t exit_code = 0;
     switch (state) {
         case TTDevice::PowerState::BUSY:
```
broskoTT, line 191: nit:
Could be rewritten with a single send_device_command call, where the first parameter is set based on this switch

device/arc/spi_arc_command.hpp
```diff
@@ -0,0 +1,35 @@
...
+#include "umd/device/tt_device/firmware/device_firmware.hpp"
+#include "umd/device/types/noc_id.hpp"
+#include "umd/device/utils/timeouts.hpp"
+
+namespace tt::umd {
+
+// The SPI flows were written against ArcMessenger::send_message, which fills the caller's
+// return-value vector in place and returns the exit code. Kept as an adapter shared by the SPI
+// devices so the conversion to DeviceFirmware does not also rewrite the flows themselves. A named
+// header rather than a per-file anonymous namespace because unity builds (tt-metal's) merge those
```
broskoTT, line 19: Let's not mention tt-metal in comments.
Also, let's not mention historical facts about code "so the conversion to DeviceFirwmare", rather state why is it currently the way it is and can't be modified further

device/arc/spi_arc_command.hpp
```diff
@@ -0,0 +1,35 @@
...
+// into one translation unit, where two copies collide.
+inline uint32_t send_spi_arc_command(
```
broskoTT, line 21: This can probably be avoided...
But as we synced offline, no need to modify in this PR.
Maybe let claude make a following PR to fix this, can also fix the other comment...

device/chip/chip.cpp
```diff
@@ -125,8 +125,13 @@ int Chip::arc_msg(
...
     }
 
-    uint32_t exit_code =
-        get_tt_device()->get_arc_messenger()->send_message(msg_code, arc_msg_return_values, args, timeout_ms);
+    // The ArcMessenger path this replaces read the thread-selected NOC itself; keep that.
+    DeviceCommandResult result =
+        get_tt_device()->get_device_firmware()->send_device_command(msg_code, args, timeout_ms, get_selected_noc_id());
+    const uint32_t exit_code = result.exit_code;
+    for (size_t i = 0; i < arc_msg_return_values.size() && i < result.return_values.size(); i++) {
+        arc_msg_return_values[i] = result.return_values[i];
```
broskoTT, line 133: This was done just because of the next few lines, can be easily rewritten to be nicer, no need for intermediary arc_msg_return_values

broskoTT (approved):

device/tt_device/blackhole_tt_device.cpp
nbuncicTT: This exact switch is deleted in the next PR of the stack (#3323 moves set_clock_state into the firmware, whose implementation already builds the message code from the switch and sends once). The nit dies with the body.

PR #3325, opened 2026-08-31
[Base API] Delegate get_noc_translation_enabled to DeviceFirmware

nbuncicTT:
## Issue

#2872
Delegate `get_noc_translation_enabled` to `DeviceFirmware`.

## Description

The Wormhole read (NIU config over the ARC APB window) and the Blackhole read (BAR/JTAG, never over a NOC) move into the firmware implementations. Blackhole's is the promotion of the private constructor helper from #3319 to the interface method — the temporary duplication flagged there ends here. Simulation returns `false` through the firmware now, same value as its deleted `TTDevice` override.

`TTDevice::get_noc_translation_enabled()` becomes a non-virtual forwarder; the pure virtual and all three overrides die.

## List of the changes

- `DeviceFirmware::get_noc_translation_enabled(NocId)`; implemented in WH (APB window read), BH (BAR/JTAG read, promoted from the private helper), simulation (`false`).
- `TTDevice::get_noc_translation_enabled` non-virtual forwarder; WH/BH/simulation overrides deleted.

## Testing

`baremetal_tests` 254/254, simulation on and off, no undefined symbols.

## API Changes

None — `TTDevice::get_noc_translation_enabled()` keeps its signature (now non-virtual). `DeviceFirmware::get_noc_translation_enabled` added.

Stacked on #3324.

## Adversarial review round

Follow-up: the facade forwards the thread-selected NOC (the deleted overrides' device reads honored it), and the constructor's now-stale `BlackholeTTDevice::` qualifier is dropped.

### DIFF
--- a/device/tt_device/firmware/blackhole_device_firmware.cpp
+++ b/device/tt_device/firmware/blackhole_device_firmware.cpp
@@ -284,7 +284,7 @@ void BlackholeDeviceFirmware::wait_firmware_ready(std::chrono::milliseconds time
     }
 }
 
-bool BlackholeDeviceFirmware::get_noc_translation_enabled() const {
+bool BlackholeDeviceFirmware::get_noc_translation_enabled(NocId /*noc_id*/) {
     uint32_t niu_cfg;
     if (get_io_device_type() == IODeviceType::JTAG) {
         niu_cfg = jtag_interface_->mmio_read32(blackhole::NIU_CFG_NOC0_ARC_ADDR);
--- a/device/api/umd/device/tt_device/blackhole_tt_device.hpp
+++ b/device/api/umd/device/tt_device/blackhole_tt_device.hpp
@@ -33,8 +33,6 @@ public:
 
     uint32_t get_min_clock_freq() override;
 
-    bool get_noc_translation_enabled() override;
-
     void read_from_arc_apb(void *mem_ptr, uint64_t arc_addr_offset, size_t size) override;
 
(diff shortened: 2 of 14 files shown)


device/tt_device/firmware/blackhole_device_firmware.cpp
```diff
@@ -284,7 +284,7 @@ void BlackholeDeviceFirmware::wait_firmware_ready(std::chrono::milliseconds time
     }
 }
 
-bool BlackholeDeviceFirmware::get_noc_translation_enabled() const {
+bool BlackholeDeviceFirmware::get_noc_translation_enabled(NocId /*noc_id*/) {
```
broskoTT, line 287: This is fine, but I have to notice that this part was implemented as part of https://github.com/tenstorrent/tt-umd/pull/3319
I synced offline with you that it might be good to separate out that change out of that PR (among other separations)

broskoTT (approved):

device/tt_device/firmware/blackhole_device_firmware.cpp
nbuncicTT: Done as part of the six-way split of the old #3319: the private NOC-translation read now arrives in #3347 together with the Blackhole boot wait that consumes it (the coordinates depend on it), and this PR only promotes it to the interface and deletes the TTDevice copy. So the separation you asked for is exactly the current stack structure.

PR #3327, opened 2026-08-31
[Base API] Delegate get_firmware_noc_coord to DeviceFirmware

nbuncicTT:
## Issue

#2872
Delegate `get_firmware_noc_coord` to `DeviceFirmware`.

## Description

The firmware implementations have resolved the ARC coordinate per NOC in their constructors since #3319 (Wormhole's is fixed; Blackhole's depends on the NOC-translation state, fixed for the device's lifetime). This promotes that private helper to the interface — the only `const` method on it — and deletes the `TTDevice` copy of the same knowledge: the `arc_core_noc0`/`arc_core_noc1` members, `set_arc_coordinate()` and both overrides. `TTDevice::get_arc_core()` forwards.

Simulation returns the origin, matching what the defaulted members used to hold.

## List of the changes

- `DeviceFirmware::get_firmware_noc_coord(NocId) const`; WH/BH promote their private helper, simulation returns `{0, 0}`.
- `TTDevice`: `arc_core_noc0`/`arc_core_noc1`, `set_arc_coordinate()` + WH/BH overrides deleted; `get_arc_core()` forwards to the firmware.

## Testing

`baremetal_tests` 254/254, simulation on and off, no undefined symbols.

## API Changes

None — `TTDevice::get_arc_core()` keeps its signature and values. `DeviceFirmware::get_firmware_noc_coord` added.

Stacked on #3326.

### DIFF
--- a/device/api/umd/device/tt_device/firmware/simulation_device_firmware.hpp
+++ b/device/api/umd/device/tt_device/firmware/simulation_device_firmware.hpp
@@ -63,6 +63,12 @@ public:
         return chip_info;
     }
 
+    // A simulated device has no management firmware core; callers that need a coordinate get the
+    // origin, matching what TTDevice::get_arc_core() returned for simulation before this existed.
+    tt_xy_pair get_firmware_noc_coord([[maybe_unused]] NocId noc_id = NocId::DEFAULT_NOC) const override {
+        return tt_xy_pair{};
+    }
+
 private:
     tt::ARCH arch_ = tt::ARCH::Invalid;
 };
--- a/device/api/umd/device/tt_device/blackhole_tt_device.hpp
+++ b/device/api/umd/device/tt_device/blackhole_tt_device.hpp
@@ -52,8 +52,6 @@ protected:
     // Number of retrain attempts is chosen based on syseng team testing.
     uint32_t get_max_dram_retrain_attempts() const override { return 3; }
 
-    void set_arc_coordinate() override;
-
 private:
     const ArchitectureRegisters registers_ = get_architecture_registers(tt::ARCH::BLACKHOLE);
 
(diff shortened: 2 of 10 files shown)


device/api/umd/device/tt_device/firmware/simulation_device_firmware.hpp
```diff
@@ -63,6 +63,12 @@ class SimulationDeviceFirmware : public DeviceFirmware {
         return chip_info;
     }
 
+    // A simulated device has no management firmware core; callers that need a coordinate get the
+    // origin, matching what TTDevice::get_arc_core() returned for simulation before this existed.
+    tt_xy_pair get_firmware_noc_coord([[maybe_unused]] NocId noc_id = NocId::DEFAULT_NOC) const override {
+        return tt_xy_pair{};
```
broskoTT, line 69: Can you double check what was the situation for Simulation up to now?
Obviously if the tests pass, everything is fine.
Just hesitant if something might break if this functions returns (likely?) 0,0 here

broskoTT (approved):

device/api/umd/device/tt_device/firmware/simulation_device_firmware.hpp
nbuncicTT: Checked when building this: it is byte-identical behavior. Up to now TTDevice::get_arc_core() for simulation returned the default-initialized arc_core_noc0/noc1 members — {0,0} — because set_arc_coordinate() was the base-class no-op for sims. The sim firmware returning tt_xy_pair{} reproduces exactly that, and nothing in the simulation flow consumes the value (no ARC messenger/telemetry path runs there). The code comment on the override records this.

PR #3329, opened 2026-09-01
[Base API] Delete ArcMessenger; DeviceFirmware::send_device_command is the one command path

nbuncicTT:
## Issue

With every production caller on `DeviceFirmware::send_device_command` (#3322), `ArcMessenger` is a second, TTDevice-coupled sender with no callers left outside tests — and the adversarial review showed the two copies had already drifted (lock set, PCIe-hang tripwire).

## Description

The `ArcMessenger` family is deleted; `send_device_command` is the one command path. `BlackholeArcMessenger` duplicated what the firmware already owns (it built a second APB window and message queue for the same device); `WormholeArcMessenger`'s protocol lives in `WormholeDeviceFirmware::send_device_command` with the exact lock set and hang tripwire restored in #3322's review round.

Consumers converted:

- The **card tests** (`test_arc_messages_wh/bh`) drive the firmware directly. The Blackhole tests used to work without `init_tt_device()` because the messenger built its own queue — a bypass around the pre-init guard; they now bring the firmware up explicitly with `init_firmware()`.
- The **Python `arc_msg` bindings** (three overloads) send through the firmware and read `DeviceCommandResult`.
- `TTDevice::get_arc_messenger()` and the `arc_messenger_` member are gone; `init_tt_device()` no longer builds one.

Note: `TTDevice::read_from_arc_apb`/`write_to_arc_apb` do **not** die here — the Wormhole SPI engine still drives the SPI controller through them directly. They retire when SPI moves onto window components (follow-up).

## List of the changes

- Deleted: `arc_messenger.{hpp,cpp}`, `wormhole_arc_messenger.{hpp,cpp}`, `blackhole_arc_messenger.{hpp,cpp}`, `TTDevice::get_arc_messenger`, `arc_messenger_`.
- Converted: both ARC-message card test suites, the three Python `arc_msg` bindings.
- Stale includes dropped across nine files.

## Testing

`baremetal_tests` 254/254, simulation on and off, Python bindings build, no undefined symbols. The rewritten suites are card tests — validated by card CI.

## API Changes

- Removed: `TTDevice::get_arc_messenger()`, `ArcMessenger`/`WormholeArcMessenger`/`BlackholeArcMessenger` (no users in tt-metal/tt-exalens; tests and Python bindings converted in this PR).

Stacked on #3327.


nbuncicTT:
Folded into #3322 per review: the messenger now dies in the same diff its replacement lands in — #3322 also converts the override-internal call sites (set_clock_state ×2, WH get_chip_info harvesting) that had kept it alive mid-stack.

PR #3332, opened 2026-09-01
[Base API] Delegate the training APIs to DeviceFirmware

nbuncicTT:
## Issue

#2872
Delegate the training APIs to `DeviceFirmware`: `wait_eth_core_training`, `get_eth_core_training_status`, `wait_dram_channel_training`.

## Description

The per-arch bodies move into the firmware implementations, and everything whose only reason to exist was serving them dies in the same diff:

- **Ethernet**: Wormhole's retrain/status/link-error decode (with the UBB timeout-is-only-a-warning rule, read through the info provider) and Blackhole's port-status read move verbatim. `TTDevice::wait_eth_core_training` and `read_eth_core_training_status` become non-virtual forwarders that resolve the `CoreCoord` and pass the thread-selected NOC; they keep their signatures (`wait` still returns the elapsed time its callers subtract from a timeout budget — now measured around the firmware call rather than inside the loop). All four per-arch overrides of each are deleted.
- **DRAM**: the base `TTDevice::wait_dram_channel_training` body moves into the firmwares — Wormhole treats a reported failure as terminal (it cannot retrain), Blackhole retries up to its 3 attempts through a private `retrain_dram_core` built on `send_device_command`. The `TTDevice` retrain machinery dies: `retrain_dram_core` (pure virtual + WH throw-stub + BH + simulation overrides) and `get_max_dram_retrain_attempts`.
- **Simulation**: one arch-aware `SimulationDeviceFirmware` serves the RTL semantics (no wait, `SUCCESS`); TTSim's "not supported" throws were unreachable — the simulation chip flow never waits on training.

Spec note: `get_dram_channel_training_status` is the one spec method deliberately not added — nothing in `TTDevice` maps to it and it would land caller-less; it arrives with its first consumer.

## List of the changes

- `DeviceFirmware`: `wait_eth_core_training`, `get_eth_core_training_status`, `wait_dram_channel_training`; implemented in WH, BH (with private `retrain_dram_core`), simulation.
- `TTDevice`: both eth methods non-virtual forwarders; `wait_dram_channel_training` forwards; `retrain_dram_core`/`get_max_dram_retrain_attempts` family deleted; all per-arch training overrides deleted (WH, BH, TTSim, RTL-sim).

## Testing

`baremetal_tests` 254/254, simulation on and off, no undefined symbols. The moved paths are card-exercised via topology discovery and `Chip` start-up in card CI.

## API Changes

- `TTDevice::wait_eth_core_training`/`read_eth_core_training_status` keep signatures, now non-virtual.
- Removed: `TTDevice::retrain_dram_core`, `TTDevice::get_max_dram_retrain_attempts` (protected; no external users).

Stacked on #3329.


device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -362,8 +362,8 @@ class TTDevice {
      * @param timeout_ms Timeout in ms.
      * @return Time taken in ms.
      */
-    virtual std::chrono::milliseconds wait_eth_core_training(
```
broskoTT, line 365: Doesn't topology discovery use the return value of this to fine control the amount of time passed for all eth training on all cores?

broskoTT (approved):

device/api/umd/device/tt_device/tt_device.hpp
nbuncicTT: Not sure, but the change is about the signature being virtual, hence this is delegated now to device firmware and we measure how much device firmware takes time for this API

broskoTT: We synced offline, found what confused me, all is good.

device/tt_device/firmware/blackhole_device_firmware.cpp
```diff
@@ -355,6 +357,99 @@ tt_xy_pair BlackholeDeviceFirmware::get_firmware_noc_coord(NocId noc_id) const {
...
+    }
+
+    // Number of retrain attempts is chosen based on syseng team testing.
+    constexpr uint32_t MAX_DRAM_RETRAIN_ATTEMPTS = 3;
+    uint32_t num_retrain_dram_core = MAX_DRAM_RETRAIN_ATTEMPTS;
+    auto start = std::chrono::steady_clock::now();
+    while (true) {
+        std::vector<DramTrainingStatus> dram_training_status =
+            firmware_info_provider_->get_dram_training_status(dram_banks_number);
+
```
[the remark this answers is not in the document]
nbuncicTT: Valid — a pre-init call would null-deref where the old path threw UninitializedDeviceError through the facade getter. Guard (same shape as set_clock_state's) queued for the follow-up PR right after the stack lands, kept out of the stack to preserve CI.

Issue #3331, opened 2026-09-01
Run the Python bindings tests in the RTL simulation CI legs


pjanevskiTT:
**Description**

The RTL simulation legs (the UMD-on-VCS jobs in tt-umd-simulators, see #2641 / #2642) build and run the UMD C++ tests, but not the Python module, so no Python test runs against an RTL simulator. `test_py_simulation_connector.py`'s host/client test skips there for the same reason it skips everywhere else.

`SimulationConnector` picks the device class from the backend the host reports, so the RTL path through the bindings is genuinely distinct from the TTSim one (#3329) and worth its own coverage.

**Scope**

- Build the Python module in the RTL legs and run `nanobind/tests` with `TT_UMD_SIMULATOR` pointed at the RTL simulator build directory.
- Confirm what those legs configure today before adding the step; they live in tt-umd-simulators, not in this repo's workflows.

**Timeline**

**Is your feature request related to a problem? Please describe.**

The Python bindings for simulation (#3315) are untested against the RTL backend, which is the half of the simulation stack the TTSim lane does not cover.

**Definition of Done**

- The simulator-gated tests in `nanobind/tests` execute against an RTL simulator in CI rather than skipping.

Issue #3330, opened 2026-09-01
Run the Python bindings tests in the TTSim CI lane


pjanevskiTT:
**Description**

`nanobind/tests` runs only in the pytest lane, which has no simulator, and `run-ttsim-tests.yml` builds with `-DTT_UMD_BUILD_PYTHON=OFF`, so that lane has no `tt_umd` module at all. A Python test that needs a running simulator therefore runs nowhere: `test_py_simulation_connector.py`'s host/client test skips in every lane.

**Scope**

- Build the Python module in the TTSim lane and run `nanobind/tests` there with `TT_UMD_SIMULATOR` pointed at the lane's `libttsim.so`.
- The lane already stages several simulators (`sim_wh`, `sim_bh`, `sim_qsr`, `sim_wh_x2`, `sim_bh_x2`); decide which of them the Python tests run against.

**Timeline**

**Is your feature request related to a problem? Please describe.**

The Python bindings for simulation (#3315) can regress silently: nothing in CI exercises them against a running simulator.

**Definition of Done**

- The simulator-gated tests in `nanobind/tests` execute against TTSim in CI rather than skipping.

Issue #3315, opened 2026-08-31
Simulation: expose SimulationConnector and its options to Python


pjanevskiTT:
**Description**

`SimulationConnector` is the entry point for opening simulated devices: it decides host vs client from the path it is given, discovers the devices, allocates a server directory, and lists the servers currently open on the machine. None of it is bound to Python — `nanobind/` only exposes the per-device factories (`create_simulation_tt_device`, `TTSimTTDevice`, `RtlSimulationTTDevice`) under `TT_UMD_BUILD_SIMULATION`, and `docs/SIMULATION_SERVER.md` states outright that "There is no Python entry point today."

Expose the client-side surface so a Python process can find a running simulation server and attach to it.

**Scope**

- Bind `SimulationConnectorOptions` (`simulator_directory`, `serve_over_sockets`, `server_directory`, `cluster_descriptor`, `num_host_mem_channels`) and `SimulationServerInfo` (`index`, `directory`, `sockets`).
- Bind `SimulationConnector`: `Role`, `role_for()`, `discover()`, `allocate_server_directory()`, `list_servers()`. Settle how `discover()`'s `std::map<ChipId, std::unique_ptr<TTDevice>>` transfers ownership to Python against the existing `TTDevice` bindings.
- Guard the new bindings with `TT_UMD_BUILD_SIMULATION`, as the existing simulation bindings are, and keep the generated `tt_umd.pyi` / `__init__.py` correct.
- Add `nanobind/tests/test_py_simulation_connector.py` covering list / attach / device I/O against a server started in the test. The simulator-gated part of it only runs once the simulation lanes build the Python module: #3330 (TTSim) and #3331 (RTL).
- Update `docs/SIMULATION_SERVER.md`: replace "There is no Python entry point today" with the Python flow.

**Timeline**

**Is your feature request related to a problem? Please describe.**

tt-exalens drives UMD from Python. With no bindings for the connector it cannot attach to a running simulation host, so the simulation server is unusable from it.

**Definition of Done**

- From Python: list the open simulation servers, attach to one, and read/write a simulated device — no C++ involved.
- Bindings are covered by a test that runs in CI.
- Stubs and `docs/SIMULATION_SERVER.md` are updated.

Issue #3314, opened 2026-08-31
Simulation server: tt-exalens integration requirements


pjanevskiTT:
**Description**

The simulation server (#2677) is in place: a host process runs a simulator and serves each simulated chip on a per-chip unix socket, other processes attach as clients by pointing UMD at the server directory, and `sim_server` starts / lists / kills hosts. See [docs/SIMULATION_SERVER.md](https://github.com/tenstorrent/tt-umd/blob/main/docs/SIMULATION_SERVER.md).

tt-exalens is one of the intended consumers, and integrating it end to end needs more than what is there today. This issue tracks the requirements coming from the Exalens team so the work can be sliced into sub-issues and tracked in one place.

**Scope**

- [x] Expose the client-side surface to Python: `SimulationConnector` and `SimulationConnectorOptions`, so a Python client can discover a running server and open devices against it without dropping into C++. #3315

- [x] Report the role (host or client) and the link to the simulator from an opened connector, in both roles. #3316

- [x] `sim_server list`: show which simulator each open server is running, alongside the arch and backend it already shows. #3328

More requirements to be added as they come in from the Exalens team; each becomes a sub-issue here.

**Timeline**

**Is your feature request related to a problem? Please describe.**

Today the simulation server is only reachable from C++ (or by pointing an existing UMD/tt-metal binary at a server directory via `TT_UMD_SIMULATOR`). tt-exalens drives UMD from Python, so it cannot attach to a running simulation host at all.

**Definition of Done**

- Every sub-issue below is closed.
- tt-exalens can attach to a running UMD simulation server and drive it the same way it drives silicon.

pjanevskiTT:
closed by #3333 @adjordjevic-TT @tt-vjovanovic

PR #3333, opened 2026-09-01
Simulation: expose the simulation API to Python

pjanevskiTT:
### Issue

#3315

### Description

`SimulationConnector` is the entry point for opening simulated devices, but only from C++ — `nanobind/` bound the per-device factories and nothing else, and `docs/SIMULATION_SERVER.md` said outright that there was no Python entry point. tt-exalens drives UMD from Python (#3314), so it could not attach to a running simulation host at all.

This binds the connector and its options. Role, discovery, directory allocation and server listing behave as they do in C++, including deciding host-vs-client from the path.

Most of the new tests need no simulator: `role_for` classifies on a real `AF_UNIX` socket named `tt-umd-sim-0.sock`, so a Python test can build one in a tmpdir. The host/client test does need one and skips without `TT_UMD_SIMULATOR` — it will only run once #3330 / #3331 build the Python module in the simulation lanes. It was verified locally against ttsim blackhole v1.10.3: a Python host serves a socket, a Python client attaches to its directory, and a `noc_write32` from the client is read back by the host.

### List of the changes

- New `nanobind/py_api_simulation.cpp`: `SimulationConnectorOptions`, `SimulationServerInfo`, and `SimulationConnector` with `Role.HOST`/`Role.CLIENT`, `role_for`, `discover`, `allocate_server_directory`, `list_servers`. Guarded by `TT_UMD_BUILD_SIMULATION`, like the existing simulation bindings.
- `nanobind/tests/test_py_simulation_connector.py`.
- `docs/SIMULATION_SERVER.md`: a "From Python" section under *Connecting from code*.

### Testing

CI

### API Changes

New Python API only; no C++ change.

### Full stack diff

### DIFF
--- a/nanobind/py_api_simulation.cpp
+++ b/nanobind/py_api_simulation.cpp
@@ -1,0 +1,153 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include <fmt/format.h>
+#include <nanobind/nanobind.h>
+#include <nanobind/stl/filesystem.h>
+#include <nanobind/stl/map.h>
+#include <nanobind/stl/optional.h>
+#include <nanobind/stl/pair.h>
+#include <nanobind/stl/shared_ptr.h>
+#include <nanobind/stl/string.h>
+#include <nanobind/stl/unique_ptr.h>
+#include <nanobind/stl/vector.h>
+
+#include "umd/device/cluster.hpp"
+#include "umd/device/cluster_descriptor.hpp"
+#include "umd/device/simulation/simulation_connector.hpp"
+#include "umd/device/tt_device/tt_device.hpp"
+
+namespace nb = nanobind;
+// Releases Python's Global Interpreter Lock (GIL) for the duration of the C++ call,
+// allowing other Python threads to run in parallel while this binding executes. Pass
+// release_gil() as a call guard to nb::class_::def() on methods that don't touch the
+// Python interpreter (e.g. blocking device I/O), so callers can drive UMD concurrently
+// from multiple Python threads.
+using release_gil = nb::call_guard<nb::gil_scoped_release>;
+
+using namespace tt::umd;
+
+#ifdef TT_UMD_BUILD_SIMULATION
+
+void bind_simulation(nb::module_ &m) {
+    nb::class_<SimulationConnectorOptions>(m, "SimulationConnectorOptions")
+        .def(nb::init<>())
+        .def_rw(
+            "simulator_directory",
+            &SimulationConnectorOptions::simulator_directory,
+            "The simulator build to host (a TTSim .so or an RTL simulator directory), or the directory a running "
+            "server keeps its sockets in to attach as a client.")
+        .def_rw(
+            "serve_over_sockets",
+            &SimulationConnectorOptions::serve_over_sockets,
+            "Host path only: publish each simulated chip over a per-chip socket so other processes can attach as "
+            "clients. Off by default, so discover() yields private in-process devices.")
+        .def_rw(
+            "server_directory",
+            &SimulationConnectorOptions::server_directory,
+            "Host path only: the directory to serve the per-chip sockets in. Empty means allocate a fresh one.")
+        .def_rw(
+            "cluster_descriptor",
+            &SimulationConnectorOptions::cluster_descriptor,
+            "Connectivity/topology used to configure the simulator on the host path. Optional; a client takes the "
+            "topology from the host instead.")
+        .def_rw("num_host_mem_channels", &SimulationConnectorOptions::num_host_mem_channels);
+
+    nb::class_<SimulationServerInfo>(m, "SimulationServerInfo")
+        .def_ro("index", &SimulationServerInfo::index, "The server's index, as used by the sim_server tool.")
+        .def_ro("directory", &SimulationServerInfo::directory, "The directory this server serves its sockets in.")
+        .def_ro(
+            "sockets", &SimulationServerInfo::sockets, "Per-chip sockets present in the directory: chip id -> path.")
+        .def("__repr__", [](const SimulationServerInfo &self) {
+            return fmt::format(
+                "SimulationServerInfo(index={}, directory='{}', chips={})",
+                self.index,
+                self.directory.string(),
+                self.sockets.size());
+        });
+
+    nb::enum_<SimulationBackendType>(m, "SimulationBackendType")
+        .value("TTSIM", SimulationBackendType::TTSIM, "The functional simulator (a libttsim .so).")
+        .value("RTL", SimulationBackendType::RTL, "An RTL simulator build.");
+
+    nb::class_<SimulationConnector> simulation_connector(
+        m,
+        "SimulationConnector",
+        "Entry point for opening simulated devices, mirroring silicon TopologyDiscovery. The role is decided purely "
+        "from what simulator_directory points at.");
+
+    nb::enum_<SimulationConnector::Role>(simulation_connector, "Role")
+        .value("HOST", SimulationConnector::Role::Host, "This process runs the simulation.")
+        .value("CLIENT", SimulationConnector::Role::Client, "This process attaches to a simulation another host runs.");
+
+    nb::class_<SimulationConnector::Connection>(
+        simulation_connector,
+        "Connection",
+        "What a discover() call opened: the role this process took and the simulator behind it.")
+        .def_ro(
+            "role",
+            &SimulationConnector::Connection::role,
+            "Whether this process hosts the simulation or attached to one.")
+        .def_ro(
+            "simulator",
+            &SimulationConnector::Connection::simulator,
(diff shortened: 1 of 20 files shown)


nanobind/tests/test_py_simulation_connector.py
```diff
@@ -0,0 +1,133 @@
...
+
+        options.simulator_directory = "/tmp/libttsim.so"
+        options.serve_over_sockets = True
+        options.server_directory = "/tmp/tt-umd-sim-server-42"
+        options.num_host_mem_channels = 1
+
+        self.assertEqual(options.simulator_directory, Path("/tmp/libttsim.so"))
+        self.assertTrue(options.serve_over_sockets)
+        self.assertEqual(options.server_directory, Path("/tmp/tt-umd-sim-server-42"))
+        self.assertEqual(options.num_host_mem_channels, 1)
```
tt-vjovanovic: This test seems pointless

nanobind/tests/test_py_simulation_connector.py
```diff
@@ -0,0 +1,133 @@
...
+    def test_role_for_directory_with_a_socket_is_client(self):
+        with tempfile.TemporaryDirectory() as directory:
+            # A real AF_UNIX socket under the per-chip name is all the classification looks at, so
+            # this exercises the client path without a simulator build.
+            with socket.socket(socket.AF_UNIX, socket.SOCK_STREAM) as sock:
+                sock.bind(os.path.join(directory, SOCKET_NAME))
+                self.assertEqual(
+                    tt_umd.SimulationConnector.role_for(directory),
+                    tt_umd.SimulationConnector.Role.CLIENT,
+                )
```
tt-vjovanovic: This test should not exist. It should use api to allocated server directory (potentially create server inside) and verify that it returns Role.CLIENT. This way it is "hacked" and might not work in the future. It is using internal knowledge to simulate testing.

nanobind/tests/test_py_simulation_connector.py
```diff
@@ -0,0 +1,133 @@
...
+        try:
+            second = tt_umd.SimulationConnector.allocate_server_directory()
+            try:
+                self.assertTrue(first.is_dir())
+                self.assertTrue(second.is_dir())
+                self.assertNotEqual(first, second)
+            finally:
+                shutil.rmtree(second, ignore_errors=True)
+        finally:
+            shutil.rmtree(first, ignore_errors=True)
```
tt-vjovanovic, line 40: In real life scenarios, is it expected for UMD users to do cleanup manually? Can't it be done on reference count going to 0?
Also, what about if program was killed? Who will do cleanup?

nanobind/tests/test_py_simulation_connector.py
```diff
@@ -0,0 +1,133 @@
...
+        directory = tt_umd.SimulationConnector.allocate_server_directory()
+        try:
+            servers = tt_umd.SimulationConnector.list_servers()
+            listed = [server for server in servers if server.directory == directory]
+            self.assertEqual(len(listed), 1, f"{directory} missing from {servers}")
+            # Nothing serves in it yet, so it has no per-chip sockets.
+            self.assertEqual(listed[0].sockets, {})
+            self.assertGreaterEqual(listed[0].index, 0)
+        finally:
+            shutil.rmtree(directory, ignore_errors=True)
```
tt-vjovanovic, line 52: Note that this test might fail if something else is being executed on the machine (for example you running some long running metal test while executing UMD tests).

tt-vjovanovic (approved):

nanobind/tests/test_py_simulation_connector.py
```diff
@@ -0,0 +1,133 @@
...
+            host_options.server_directory = server_directory
+
+            self.assertEqual(
+                tt_umd.SimulationConnector.role_for(host_options.simulator_directory),
+                tt_umd.SimulationConnector.Role.HOST,
+            )
+
+            host_devices = tt_umd.SimulationConnector.discover(host_options)
+            self.assertEqual(len(host_devices), 1)
+            self.assertTrue((server_directory / SOCKET_NAME).exists())
```
tt-vjovanovic, line 77: Can you please also verify that type of returned `TTDevice` is `RtlSimulationTTDevice`?

nanobind/py_api_simulation.cpp
```diff
@@ -0,0 +1,109 @@
...
+            nb::arg("simulator_directory"),
+            release_gil(),
+            "Host or client, decided from the path: a directory holding live simulation sockets means a host already "
+            "serves there, so attach to it; anything else is a simulator build to host.")
+        .def_static(
+            "discover",
+            &SimulationConnector::discover,
+            nb::arg("options"),
+            release_gil(),
+            "Opens the simulated devices for these options and returns them as {chip_id: TTDevice}.")
```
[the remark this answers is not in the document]
pjanevskiTT: Checked this rather than changed it: the map binding works. `TopologyDiscovery::discover` is bound the same way (`py_api_topology_discovery.cpp`, returning `std::pair<unique_ptr<ClusterDescriptor>, map<ChipId, unique_ptr<TTDevice>>>`), and nanobind's map caster moves out of a moved-from map rather than iterating by lvalue reference. Verified end to end against ttsim: the returned devices are usable from Python, including `noc_write32`/`noc_read32` across the socket, and ownership transfers correctly.

`discover` is bound through a lambda now in any case, since it returns `(connection, devices)`.

nanobind/tests/test_py_simulation_connector.py
pjanevskiTT: Agreed — it exercises nanobind's `def_rw`, not UMD. Deleted in 3e0569652a9c543e9d438be3adec49792d8e4441.

nanobind/tests/test_py_simulation_connector.py
pjanevskiTT: Deleted in 3e0569652a9c543e9d438be3adec49792d8e4441 rather than rewritten: `test_host_serves_a_socket_a_client_can_attach_to` already asserts `role_for(server_directory) == Role.CLIENT` against a directory a real host is serving, so the hand-built socket was a worse duplicate of a test that exists.

Your point proved itself in the meantime — #3340 makes classification require a *live* listener, and the hand-bound socket (bound, never listening) is exactly what a crashed host leaves behind, so it started classifying as stale and the test failed. Internal knowledge, and the internals moved.

nanobind/tests/test_py_simulation_connector.py
pjanevskiTT: Not manual in the normal case, but the API does have a gap here.

A directory a host actually serves cleans itself up: the socket's destructor removes the socket and the now-empty directory, which `SimulationConnector.CreatesHostDeviceAndExposesSocket` asserts. If the process is killed, nothing runs, and the leftovers are cleared by `sim_server.sh prune`, which removes the directories of servers that no longer answer.

The gap is a directory that is allocated and then never served, which is what these tests do — nothing owns it, so nothing removes it. `allocate_server_directory()` hands back a bare path with no owner, so refcounting has nothing to hang off. Making it return an RAII handle (or having the tests go through a real host) would close that; it is a change to the C++ API rather than the bindings, so I would rather do it as its own issue than widen this PR

nanobind/tests/test_py_simulation_connector.py
pjanevskiTT: It should be insulated from that: the test filters `list_servers()` down to the exact directory it just allocated and asserts on that one row, rather than counting rows or assuming it is alone. A concurrent `sim_server` (or a metal run hosting its own simulation) adds rows the filter drops.

The allocation itself is race-safe too — `allocate_server_directory()` claims an index with an atomic `mkdir`, so two racing callers get different directories. Left as is; say the word if you would still rather it went.

nanobind/tests/test_py_simulation_connector.py
pjanevskiTT: Added in 3e0569652a9c543e9d438be3adec49792d8e4441, for both the host and the client device.

One adjustment: under `TT_UMD_SIMULATOR` pointing at a `libttsim.so` the concrete type is `TTSimTTDevice`, not `RtlSimulationTTDevice` — so the assertion picks the expected class from the backend the connection reports, and holds for an RTL host too. Verified the concrete type does surface through the bindings rather than the `TTDevice` base.

nanobind/tests/test_py_simulation_connector.py
pjanevskiTT: Filed as #3342.

Correcting something in my reply above: I said nothing removes an allocated-but-never-served directory. That was wrong — `list` reports it as `empty` and `sim_server.sh prune` does remove it, the same as a crashed host's. It is not a permanent leak.

The real gap is narrower, and #3342 states it that way: the directory has no owner, so releasing it is a manual out-of-band sweep, and `prune` cannot distinguish an abandoned directory from a host that is still coming up — both read as `empty`, which is why `sim_server.sh` warns against pruning while starting one. An RAII handle from `allocate_server_directory()` would close that, which is a C++ API change rather than a bindings one.

broskoTT (approved):

Issue #3316, opened 2026-08-31
Simulation: report the role and the simulator link of an opened SimulationConnector


pjanevskiTT:
**Description**

`SimulationConnector::discover()` returns only the devices. Once it returns, a caller cannot ask what it got: whether this process is hosting the simulation or attached as a client to someone else's host, and what simulator sits behind the devices. `role_for()` answers the first question only before opening, from the path, and `Cluster` takes the simulation fields as `ClusterOptions` inputs and exposes no getters for them.

A consumer needs to know, after opening, what it is connected to.

**Scope**

Report from an opened connector:

- The role: host or client.
- As host: the simulator behind it (the `libttsim.so` or the RTL build directory) and the backend kind; whether it is serving, and if so the server directory and per-chip socket paths it is serving on — including a directory UMD allocated itself, which the caller has no way to learn today.
- As client: the server directory and per-chip socket paths it attached to, and the simulator the host is running — its `libttsim.so` path or RTL build directory, plus the arch and backend kind.

The same information must be reachable through `Cluster` and from Python (#3315).

**Timeline**

**Is your feature request related to a problem? Please describe.**

tt-exalens needs to tell the user what it is connected to — "hosting `libttsim_blackhole.so`" versus "attached to the server in `/tmp/tt-umd-sim-server-0`". The connector reports nothing back, and a host that let UMD allocate the server directory cannot even name the directory it is serving in.

**Definition of Done**

- After opening, a caller can ask whether it is host or client and get the simulator link for that role, from C++ and from Python.
- A host serving in a UMD-allocated directory learns that directory without pre-allocating it.
- Covered by tests for both roles.

PR #3336, opened 2026-09-01
Simulation: report the role and simulator link of an opened connection

pjanevskiTT:
### Issue
#3316. Stacked on #3333, which binds the `discover()` signature this PR changes.

### Description
`SimulationConnector::discover()` returned only the devices, so a caller could not ask what it had opened: whether this process runs the simulation or attached to a host that does, and which simulator sits behind it. `role_for()` answers only the first, only from the path, and only before opening — re-running it afterwards races a host that has since exited. Worse on the host side: with `server_directory` empty the connector allocated one internally and merely logged it, so a caller that did not pre-allocate could never name the directory it was serving in.

`discover()` now returns that connection alongside the devices, and `Cluster` reports the same through `get_simulation_connection()`. A host names the simulator it opened; a client names the directory it attached to plus the simulator the host reports over the wire, which needed a new `simulator_path` field in the device-info message — appended last, so an older host still decodes and the client reads it as empty. Role and backend stay separate fields rather than one enum, since they are orthogonal axes.

### List of the changes
- `SimulationConnector::Connection` (role, simulator, backend, arch, server directory, sockets) and `discover()` returning it with the devices; call sites updated.
- `Cluster::get_simulation_connection()`, recording the host side including a UMD-allocated server directory.
- `simulator_path` added to `SimulationServerDeviceInfo` and served by the host; `SimulationTTDevice::backend_type()` made public.
- Python bindings for the connection and the new `discover()` shape; docs and tests.

### Testing
CI

### API Changes
`discover()` returns `SimulationConnector::Result` instead of the device map. New `Cluster::get_simulation_connection()`. `SimulationTTDevice::backend_type()` is now public. `describe_device()` takes the simulator path. In Python, `SimulationConnector.discover()` returns `(connection, devices)`.

### Full stack diff

### DIFF
--- a/device/api/umd/device/cluster.hpp
+++ b/device/api/umd/device/cluster.hpp
@@ -26,6 +26,9 @@
 #include "umd/device/chip/remote_chip.hpp"
 #include "umd/device/cluster_descriptor.hpp"
 #include "umd/device/soc_descriptor.hpp"
+#ifdef TT_UMD_BUILD_SIMULATION
+#include "umd/device/simulation/simulation_connector.hpp"
+#endif  // TT_UMD_BUILD_SIMULATION
 #include "umd/device/topology/topology_discovery.hpp"
 #include "umd/device/topology/topology_discovery_options.hpp"
 #include "umd/device/tt_device/remote_communication.hpp"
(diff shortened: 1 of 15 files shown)


pjanevskiTT:
Not actually merged — GitHub auto-closed this while the stack was being reordered, because for a moment this PR's head was an ancestor of its own base branch. Superseded by #3341, which carries the same commit based on `main`. Nothing here landed on `main`.

Issue #3081, opened 2026-07-27
Simulation: refactor host/client role & backend classification (post socket stack)


pjanevskiTT:
### Context
The simulation-server socket stack (part of #2677) deferred a few design cleanups so the feature PRs could land first. This issue tracks the agreed post-stack refactoring session. Deferred review threads live on #3020 (decide host/client role from `simulator_path`).

### Items to address

1. **Separate backend from role.** `SimulationConnector` conflates the *backend* (TTSim vs RTL) with the *host/client role*: the internal `PathKind { HOST_TTSIM, HOST_RTL, CLIENT }` mixes both axes. They're orthogonal — a client can attach to either backend — and should be modeled as two separate concepts (role: host/client; backend: TTSim/RTL).

2. **Demote `SimulationServerSocket::try_create()` + add a static `is_live(path)`.** Now that the role is decided from the path, `try_create()`'s `nullptr` return is no longer consumed in production (only `create()` and tests use it). Make it a private implementation detail of `create()`, and give tests a proper `static bool is_live(const std::filesystem::path&)` liveness probe instead of the current bind-and-destruct.

3. **Liveness-aware classification (stale sockets).** `sockets_in_directory()` relies on `is_socket()`, which can't distinguish a live host from a socket a crashed host left behind. One stale socket permanently classifies a directory as a client target and then hard-fails on connect — losing the stale-socket reclamation the old bind-based `try_create()` path did. Classification should probe liveness (count only sockets with a live listener), which depends on item 2's `is_live()`. The baremetal test that seeds stale sockets should be updated to the new semantics.

### Out of scope (tracked separately)
- Client-side connect/read timeout in `SimulationClient` (the bound-but-not-serving hang window, from #3020 review) — a separate small follow-up PR, not this refactor.

PR #3341, opened 2026-09-01
Simulation: report the role and simulator link of an opened connection

pjanevskiTT:
### Issue
#3316. Replaces #3336, which GitHub auto-closed as merged while the stack was reordered — same commit, now based on `main`, with the Python half moved out to #3333, which stacks on this.

### Description
`SimulationConnector::discover()` returned only the devices, so a caller could not ask what it had opened: whether this process runs the simulation or attached to a host that does, and which simulator sits behind it. `role_for()` answers only the first, only from the path, and only before opening — re-running it afterwards races a host that has since exited. Worse on the host side: with `server_directory` empty the connector allocated one internally and merely logged it, so a caller that did not pre-allocate could never name the directory it was serving in.

`discover()` now returns that connection alongside the devices, and `Cluster` reports the same through `get_simulation_connection()`. A host names the simulator it opened; a client names the directory it attached to plus the simulator the host reports over the wire, which needed a new `simulator_path` field in the device-info message — appended last, so an older host still decodes and the client reads it as empty. Role and backend stay separate fields rather than one enum, since they are orthogonal axes; the internal `PathKind` still conflates them, which is #3081 item 1.

### List of the changes
- `SimulationConnector::Connection` (role, simulator, backend, arch, server directory, sockets) and `discover()` returning it with the devices; call sites updated.
- `Cluster::get_simulation_connection()`, recording the host side including a UMD-allocated server directory.
- `simulator_path` added to `SimulationServerDeviceInfo` and served by the host; `SimulationTTDevice::backend_type()` made public.
- Docs and tests.

### Testing
CI

### API Changes
`discover()` returns `SimulationConnector::Result` instead of the device map. New `Cluster::get_simulation_connection()`. `SimulationTTDevice::backend_type()` is now public. `describe_device()` takes the simulator path.

### Full stack diff

### DIFF
--- a/device/api/umd/device/simulation/simulation_connector.hpp
+++ b/device/api/umd/device/simulation/simulation_connector.hpp
@@ -9,6 +9,8 @@
 #include <memory>
 #include <vector>
 
+#include "umd/device/simulation/simulation_server_protocol.hpp"
+#include "umd/device/types/arch.hpp"
 #include "umd/device/types/cluster_descriptor_types.hpp"
 
 namespace tt::umd {
@@ -59,7 +61,36 @@ public:
     enum class Role { Host, Client };
     static Role role_for(const std::filesystem::path& simulator_directory);
 
-    static std::map<ChipId, std::unique_ptr<TTDevice>> discover(const SimulationConnectorOptions& options);
+    // What a discover() call actually opened: which side of the connection this process ended up
+    // on, and what simulator sits behind it. Reported alongside the devices because none of it is
+    // recoverable afterwards -- re-classifying the path answers only the role, and races a host
+    // that has since exited, while a server directory discover() allocated itself is otherwise
+    // never told to the caller at all.
+    struct Connection {
+        Role role = Role::Host;
+        // The simulator behind the devices: a libttsim .so path or an RTL build directory. As a
+        // host, the build this process opened; as a client, the build the host reports it runs --
+        // empty when talking to a host that predates serving that field.
+        std::filesystem::path simulator;
+        // Which kind of simulator that is, and the architecture it simulates.
+        SimulationBackendType backend = SimulationBackendType::TTSIM;
+        tt::ARCH arch = tt::ARCH::Invalid;
+        // As a host, the directory this process serves its sockets in: empty when serving is off,
+        // and the freshly allocated directory when the caller left server_directory empty. As a
+        // client, the directory it attached to.
+        std::filesystem::path server_directory;
+        // The per-chip sockets in that directory ({chip_id -> socket path}). Empty for a host that
+        // is not serving.
+        std::map<ChipId, std::filesystem::path> sockets;
+    };
+
+    // The devices discover() opened, and the connection it opened them over.
+    struct Result {
+        Connection connection;
+        std::map<ChipId, std::unique_ptr<TTDevice>> devices;
+    };
+
+    static Result discover(const SimulationConnectorOptions& options);
 
     // Claims a fresh directory for one simulation server (the lowest free index) and returns it, so
     // a caller can report the location before it starts serving there (pass it back as
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -465,7 +465,9 @@ Cluster::Cluster(ClusterOptions options) {
                 connector_options.simulator_directory = options.simulator_directory;
                 connector_options.num_host_mem_channels =
                     static_cast<int>(options.num_host_mem_ch_per_mmio_device.value_or(0));
-                tt_devices = SimulationConnector::discover(connector_options);
+                auto discovered = SimulationConnector::discover(connector_options);
+                simulation_connection_ = std::move(discovered.connection);
+                tt_devices = std::move(discovered.devices);
 
                 // Every chip the topology names must have a socket-backed device (single-chip today;
                 // a multichip host will publish one socket per chip). Fail clearly rather than later
@@ -654,9 +656,16 @@ Cluster::Cluster(ClusterOptions options) {
 #endif  // TT_UMD_BUILD_SIMULATION
 
 #ifdef TT_UMD_BUILD_SIMULATION
-    if (options.chip_type == ChipType::SIMULATION && options.serve_simulation_devices_over_sockets) {
-        serve_simulation_devices_over_sockets(
-            options.simulator_directory, options.simulator_server_directory, options.simulation_shutdown_handler);
+    if (options.chip_type == ChipType::SIMULATION) {
+        // The client path recorded its connection during discovery, so reaching here without one
+        // means this process runs the simulation itself.
+        if (!simulation_connection_.has_value()) {
+            simulation_connection_ = describe_simulation_host(options.simulator_directory);
+        }
+        if (options.serve_simulation_devices_over_sockets) {
+            serve_simulation_devices_over_sockets(
+                options.simulator_directory, options.simulator_server_directory, options.simulation_shutdown_handler);
+        }
     }
 #endif  // TT_UMD_BUILD_SIMULATION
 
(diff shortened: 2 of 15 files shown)


device/simulation/simulation_connector.cpp
```diff
@@ -142,6 +143,36 @@ std::map<ChipId, std::unique_ptr<TTDevice>> SimulationConnector::discover(const
...
+                        simulator_path.string(),
+                        info.simulator_path,
+                        arch_to_str(static_cast<tt::ARCH>(info.arch)),
+                        info.backend_type == SimulationBackendType::TTSIM ? "ttsim" : "rtl",
+                        result.connection.simulator.string(),
+                        arch_to_str(result.connection.arch),
+                        result.connection.backend == SimulationBackendType::TTSIM ? "ttsim" : "rtl");
+                }
+                result.connection.sockets.emplace(chip_id, socket_file);
                 devices.emplace(chip_id, make_client_device(chip_id, std::move(client), info));
```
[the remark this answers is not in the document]
pjanevskiTT: Confirmed and fixed in 17f57032cb1d4490d971d78ee3217651bceb99f0. `make_client_device()` throws on a backend kind this build doesn't know, so the socket could be recorded for a chip the catch then skipped. The `emplace` now happens after the device is in the map.

The three derived fields were already safe: they are guarded on `result.devices.empty()`, so a socket that fails to build a device leaves `devices` empty and the next successful one overwrites them. That guard now also reads as "first successful" rather than coincidentally being one.

tests/api/test_simulation_server_protocol.cpp
```diff
@@ -120,6 +121,20 @@ TEST(SimulationServerProtocol, DeviceInfoSerializationRoundTrip) {
     EXPECT_EQ(decoded.eth_harvesting_mask, info.eth_harvesting_mask);
     EXPECT_EQ(decoded.l2cpu_harvesting_mask, info.l2cpu_harvesting_mask);
     EXPECT_EQ(decoded.pcie_harvesting_mask, info.pcie_harvesting_mask);
+    EXPECT_EQ(decoded.simulator_path, info.simulator_path);
+}
+
+// simulator_path was appended to the schema after the first hosts shipped, so a host that never
+// sets it must still decode -- as empty, not as a null-string read.
+TEST(SimulationServerProtocol, DeviceInfoWithoutSimulatorPathRoundTrip) {
```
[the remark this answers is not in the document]
pjanevskiTT: Right that it was only covering the empty-string case. Fixed in 17f57032cb1d4490d971d78ee3217651bceb99f0 from the other end: `encode()` now omits `simulator_path` entirely when it is empty rather than writing an empty string, so a message with no simulator path is byte-identical on the wire to one from a host built before the field existed. The test therefore does exercise the absent-field decode, and is renamed `DeviceInfoWithAnAbsentSimulatorPathRoundTrip` to say so.

device/api/umd/device/simulation/simulation_connector.hpp
```diff
@@ -59,7 +61,36 @@ class SimulationConnector {
...
+    struct Connection {
+        Role role = Role::Host;
+        // The simulator behind the devices: a libttsim .so path or an RTL build directory. As a
+        // host, the build this process opened; as a client, the build the host reports it runs --
+        // empty when talking to a host that predates serving that field.
+        std::filesystem::path simulator;
+        // Which kind of simulator that is, and the architecture it simulates.
+        SimulationBackendType backend = SimulationBackendType::TTSIM;
+        tt::ARCH arch = tt::ARCH::Invalid;
+        // As a host, the directory this process serves its sockets in: empty when serving is off,
```
ajovanovicTT, line 78: nit: It would be more readable if comment explanations of struct fields were above struct def.

device/cluster.cpp
```diff
@@ -666,6 +675,28 @@ Cluster::Cluster(ClusterOptions options) {
...
+    SimulationConnector::Connection connection;
+    connection.role = SimulationConnector::Role::Host;
+    connection.simulator = simulator_directory;
+    // Take the backend and arch from a device rather than re-deriving them from the path, so this
+    // reports what was actually built. server_directory and sockets stay empty until (and unless)
+    // serve_simulation_devices_over_sockets() fills them.
+    for (const auto& [chip_id, chip] : chips_) {
+        if (auto* sim_device = dynamic_cast<SimulationTTDevice*>(chip->get_tt_device())) {
+            connection.backend = sim_device->backend_type();
+            connection.arch = sim_device->get_soc_descriptor().arch;
```
ajovanovicTT, line 696: The loop iterates chips_ looking for the first SimulationTTDevice and breaks. If nothing matches (shouldn't happen, but no guard), the returned Connection silently has backend = TTSIM (the default) and arch = Invalid. Worth a log_warning or assert after the loop so a future regression that breaks the host construction order doesn't misreport silently.

device/cluster.cpp
```diff
@@ -686,11 +717,21 @@ void Cluster::serve_simulation_devices_over_sockets(
...
+    }
     for (const auto& [chip_id, chip] : chips_) {
         if (auto* sim_device = dynamic_cast<SimulationTTDevice*>(chip->get_tt_device())) {
-            sim_device->adopt_socket(
-                SimulationServerSocket::create(SimulationServerSocket::default_socket_path(server_directory, chip_id)),
-                shutdown_handler);
+            const std::filesystem::path socket_path =
+                SimulationServerSocket::default_socket_path(server_directory, chip_id);
+            sim_device->adopt_socket(SimulationServerSocket::create(socket_path), shutdown_handler);
+            if (simulation_connection_.has_value()) {
```
ajovanovicTT: The if (simulation_connection_.has_value()) guards in serve_simulation_devices_over_sockets() treat a missing connection as a no-op rather than a bug. Since the connection is always populated before this is called, .value() (which throws on empty optional) would make the invariant explicit and surface any future regression that violates it.

tests/api/test_simulation_connector.cpp
```diff
@@ -289,3 +307,105 @@ TEST(SimulationConnector, HostAndClientClustersShareDeviceMemory) {
...
+    EXPECT_EQ(connection.arch, result.devices.at(0)->get_soc_descriptor().arch);
+    EXPECT_EQ(connection.server_directory, server_directory);
+    EXPECT_EQ(connection.sockets.size(), result.devices.size());
+    EXPECT_EQ(connection.sockets.at(0), SimulationServerSocket::default_socket_path(server_directory, 0));
+}
+
+// The point of reporting the connection on the host side: with server_directory left empty the
+// connector allocates one internally, and the caller has no other way to learn which. Requires
+// TT_UMD_SIMULATOR.
+TEST(SimulationConnector, ReportsTheServerDirectoryItAllocated) {
```
ajovanovicTT, line 343: nit: ReportsTheServerDirectoryItAllocated allocates a real directory but doesn't clean it up. If an ASSERT_* fires early the directory persists and could interfere with later runs. A scope guard calling remove_all would fix it.

ajovanovicTT (approved):

device/api/umd/device/simulation/simulation_connector.hpp
pjanevskiTT: Keeping them inline. `SimulationConnectorOptions`, 20 lines up in this header, documents each field inline, and `ClusterOptions` in `cluster.hpp` does the same with per-field doxygen blocks -- that is the convention here for structs whose fields carry caveats. `SimulationServerInfo` describes its fields above the struct because `index`/`directory`/`sockets` need no explanation; `Connection`'s do: `simulator` is empty when the host predates serving that field, an empty `server_directory` is how "not serving" reads, and `sockets` is empty for that same host. Lifted above the struct, those qualifications stop being findable from the field they qualify.

tests/api/test_simulation_connector.cpp
pjanevskiTT: Cleanup already happens, and it is the socket's job: `~SimulationServerSocket` (simulation_server_socket.cpp:330-344) removes the socket file and then the now-empty server directory, guarded to the `<temp>/tt-umd-sim-server-<index>` naming. `result` is a local and an `ASSERT_*` is just a `return`, so the devices are destroyed on an early exit too and the directory goes with them. `CreatesHostDeviceAndExposesSocket` at the top of this file asserts exactly that: `EXPECT_FALSE(std::filesystem::is_directory(server_directory))` once the device's scope ends.

A `remove_all` guard would also be actively wrong if declared after `result`: destruction runs in reverse, so it would delete a live socket out from under the still-running server. And `remove_all` deletes a non-empty directory, which is the case the destructor deliberately declines to touch -- a multi-chip host's directory has to survive until its last socket is gone.

There is a leak window, but not this one: if `discover()` throws between `allocate_server_directory()` and the first socket bind, the empty directory persists, consumes an index, and enumerates in `list_servers()`. That is in `discover()` and pre-existing -- equally true of the four other tests here that allocate a directory.

device/cluster.cpp
pjanevskiTT: Fixed in https://github.com/tenstorrent/tt-umd/commit/2a1438e9b3642e9adbd070c3d58b46024ed7146a. `describe_simulation_host()` now warns when it finds no simulation device, instead of reporting the defaults as if they had been read off one.

Warning rather than asserting is deliberate: a simulation `Cluster` with zero chips is legal today -- an empty `target_devices` with no `cluster_descriptor.yaml` beside the simulator falls back to a mock descriptor built *from* `target_devices`, so it constructs with no chips at all (`HostAndClientClustersShareDeviceMemory` names chip 0 explicitly for exactly that reason). An assert would turn a constructible cluster into a throw. A future regression that describes the host before the chips exist surfaces through the same warning.

device/cluster.cpp
pjanevskiTT: Fixed in https://github.com/tenstorrent/tt-umd/commit/2a1438e9b3642e9adbd070c3d58b46024ed7146a. The invariant does hold: the only call site is the constructor, immediately after the block that populates `simulation_connection_`. Both guards are gone and the two writes are unconditional, behind one `UMD_ASSERT` after the role check.

`UMD_ASSERT` rather than `.value()` so a violation names the invariant instead of raising a bare `std::bad_optional_access` from two separate sites.

PR #3345, opened 2026-09-01
[Base API] Give the firmware implementations their private plumbing

nbuncicTT:
## Description
Private plumbing the boot waits will consume: device identity for error payloads, the per-NOC ARC coordinates (Blackhole resolves them in the constructor from a private NOC-translation read — the public `TTDevice::get_noc_translation_enabled` survives until its own delegation PR; Wormhole's are fixed constants), the ARC window members, and thin accessors over them (`get_io_device_type`, `get_firmware_noc_coord`, `read_from_arc_apb`, `read_from_arc_csm`). All private; briefly caller-less — the waits that consume them are the next two PRs.

## Testing
`baremetal_tests` 254/254, sim on/off.

## API Changes
None.

Stacked on the window PR.

### DIFF
--- a/device/CMakeLists.txt
+++ b/device/CMakeLists.txt
@@ -103,6 +103,7 @@ target_sources(
         tt_device/hang_detection/blackhole_hang_detector.cpp
         tt_device/firmware/blackhole_arc_apb.cpp
         tt_device/firmware/blackhole_device_firmware.cpp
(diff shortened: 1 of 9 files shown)


nbuncicTT:
Dissolved per review: introduction and reduction must share a diff, so the plumbing now arrives inside the two behavior migrations that consume it — the Wormhole half in #3346, the Blackhole half in #3347.

PR #3349, opened 2026-09-02
Add optional cluster_id to the cluster descriptor

p1-0tr:
Fabric needs one string that identifies the group of accelerators a
descriptor describes, so that FSD-built and live/mock topologies key their
chips the same way. A cluster descriptor had no such field, so tt-metal fell
back to the mock YAML's filename, which is not a hostname.

Adds an optional top-level `cluster_id` to the cluster descriptor: a unique
string identifying the group of Tenstorrent accelerators connected to a
common host / controller / root complex. Its value is currently that group's
bare metal hostname, because that is what the factory system descriptor and
the fabric topology solver join on, but the field identifies the accelerator
group rather than a machine -- hence `cluster_id` and not `hostname`, so the
value scheme can change later without another schema migration.

Discovery stamps it from `TopologyDiscoveryOptions::cluster_id` when the caller
supplies one, else from gethostname(). The option is what makes the field
usable in containers and VMs, where gethostname() returns the container id or
guest name -- stable enough to look convincing while pointing at nothing
physical. UMD deliberately does not read an environment variable for this:
container and VM plumbing belongs to whoever launches the process, and one
field on the options struct is enough for them to pass the value down.
`ClusterOptions` already embeds `TopologyDiscoveryOptions`, so Cluster callers
get it too.

A supplied id that is not legal throws rather than falling back to
gethostname(): the caller passed it on purpose, and quietly substituting a
container hostname would produce a wrong-but-plausible topology, which is the
failure the option exists to prevent. An OS hostname that cannot be used as a
cluster id (over 128 characters, exotic characters) only warns and leaves the
field unset, since that is not something the caller asked for and throwing
there would break discovery on hosts that work today.

Legal ids are 1 to 128 characters from [A-Za-z0-9._-]: a hostname-shaped
charset while cluster ids are hostname-valued, and a length limit that is the
fixed 128-byte buffer consumers pack the id into. Both paths that can set the
field -- YAML parsing and discovery -- validate, so every cluster id on a
descriptor is a legal one. The value is not expected to change over a
descriptor's lifetime, so there is no setter.

Note that `cluster_id` here is unrelated to the existing `EthCoord::cluster_id`,
which is a per-chip integer that groups ethernet-connected chips within one
descriptor.

The field is optional forever: old YAMLs load with it empty, old UMD ignores
the key, serialize() omits it when unset so existing descriptors serialize
unchanged, and it is not in the schema's `required` list. It is copied by
create_constrained_cluster_descriptor and apply_chip_id_remapping (fewer
chips, same accelerator group) and left unset by create_mock_cluster.

Documented in docs/CLUSTER_ID.md.

Offline gtests in tests/baremetal/test_cluster_descriptor_cluster_id.cpp:
parse with and without the key, every existing offline descriptor still
parses, legal and illegal values through YAML, serialize omit/round-trip,
constrained-descriptor copy, a supplied id winning over the hostname, an
illegal supplied id throwing, the hostname default, and a mock cluster having
no cluster id.

Built with the default preset (clang-20, clang-tidy on, -Werror). All 13
gtests pass, and the rest of baremetal_tests is unaffected.

Adds `ClusterDescriptor::get_cluster_id()` and
`TopologyDiscoveryOptions::cluster_id` (both bound in nanobind), and an
optional `cluster_id` key in the cluster descriptor YAML schema.

### DIFF
--- a/device/common/utils.hpp
+++ b/device/common/utils.hpp
@@ -7,6 +7,7 @@
 #include <fmt/ranges.h>
 #include <unistd.h>
 
+#include <array>
 #include <chrono>
 #include <cstddef>
 #include <cstdint>
@@ -15,6 +16,7 @@
 #include <fstream>
 #include <optional>
 #include <string>
+#include <string_view>
 #include <thread>
 #include <tt-logger/tt-logger.hpp>
 #include <type_traits>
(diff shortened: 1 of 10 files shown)


device/common/utils.hpp
```diff
@@ -136,6 +139,105 @@ inline std::unordered_set<int> get_visible_devices(const std::unordered_set<int>
...
+// to a non-empty value, otherwise the OS hostname. Returns nullopt when neither can be used, which
+// leaves the field unset on the cluster descriptor and lets consumers fall back to what they did
+// before.
+//
+// An invalid TT_CLUSTER_ID throws rather than falling back to gethostname(): the variable is set on
+// purpose, and quietly substituting a container hostname would produce a wrong-but-plausible
+// topology, which is exactly what TT_CLUSTER_ID exists to prevent. An unusable OS hostname only
+// warns, because that is not something the operator asked for.
+inline std::optional<std::string> local_cluster_id() {
+    const std::optional<std::string> cluster_id_from_env = get_env_var_value(TT_CLUSTER_ID_ENV.data());
```
Riddy21: @p1-0tr Do you think we should pass this into cluster desrciptor as an option from tt-metal? And migraede the TT_CLUSTER_ID_ENV code to rtoptions.cpp?

I think it might belong better there because any environemnt options as user interface should be passed through the official way

p1-0tr: What about users who don't set up a metal runtime? (e.g. FM)

aleksamarkovicTT: I'm very much against environment variables being used here. Please remove it from the PR.

p1-0tr: Done :)

nanobind/tests/test_py_topology_discovery.py
```diff
@@ -68,3 +69,83 @@ def test_no_remote_discovery_true(self):
...
+        - 0
+asic_locations:
+  0: 0
+"""
+
+
+class TestClusterDescriptorClusterId(unittest.TestCase):
+    """Offline: no device needed, the descriptors are built from YAML."""
+
+    def test_cluster_id_from_yaml(self):
```
aleksamarkovicTT: I think these Python tests are unnecessary, feel free to remove them.

aleksamarkovicTT (approved):

Issue #3353, opened 2026-09-02
Bring up TTSim through the silicon bringup path instead of a supplied cluster descriptor


pjanevskiTT:
TTSim clusters are built from a hand-written `cluster_descriptor.yaml` placed beside `libttsim.so`, not from the device. Everything downstream — harvesting masks, board and ASIC IDs, `chips_with_mmio`, the Ethernet connection map — is read out of that file. Run the real `TopologyDiscovery` path over the simulator's ARC and Ethernet firmware instead, so simulation and silicon share one bringup path.

**Scope**

- Give `SimulationTTDeviceModel` the components the shared bringup path already expects: a `DeviceProtocol`, a `SocArchDescriptor` resolved through the same helper the silicon models use, and a real communication device id
- Enumerate TTSim's PCI endpoints by walking the full bus/device space, and support several MMIO chips on one shared simulator image
- Let `TopologyDiscovery` accept devices created by a backend with no OS enumeration, selected by a simulation field on `TopologyDiscoveryOptions`. `IODeviceType` stays `PCIe` — TTSim simulates PCIe
- Route the no-descriptor TTSim case through discovery in `Cluster`; a supplied descriptor stays authoritative
- Remove the staged `cluster_descriptor.yaml` from the TTSim CI lanes, and assert chip count and harvesting masks in its place
- Report every remaining simulation/silicon divergence to the simulator team rather than special-casing it in UMD

**Is your feature request related to a problem? Please describe.**

Simulation users must author and maintain a descriptor that can disagree with the simulator, and UMD's silicon discovery path is never exercised under simulation, so a regression in it stays invisible until it reaches hardware. The checked-in descriptors already disagree with the device: `wormhole_N300.yaml` declares `harvest_mask: 0` for both chips while the simulator's ARC reports `0x41` and `0x05`.

**Not in scope**

- The SOC descriptor. Still required — the architecture pre-probe needs an arch before any device object exists
- Quasar and Trizor. No ARC, no Ethernet, no host MMIO path in TTSim; they stay on the mock descriptor
- craq-sim. No ARC tile at all, so discovery cannot run there; it keeps its descriptor-driven flow unchanged

**Definition of Done**

- `bh`, `bh_x2`, `wh` and `wh_x2` come up correctly with no `cluster_descriptor.yaml` present
- Both `cp … cluster_descriptor.yaml` steps are gone from `.github/workflows/run-ttsim-tests.yml`
[...]

PR #3358, opened 2026-09-02
Avoid cross-process non-MMIO locking for isolated simulation

micahwhiteTT:
### Description
`TTDevice::get_communication_device_id()` returns `-1` for a device with no device protocol. That id keys UMD's cross-process locks, so every simulator on a machine named the same shared-memory lock (`TT_UMD_LOCK.NON_MMIO_-1_PCIe`) and serialized its ETH transactions against every other, unrelated run -- two independent ttsim tt-metal tests would block each other for no reason.

Report the PID instead: unique among live processes, stable for the model's lifetime, so each simulator's lock name is private to its own process.

### List of the changes
- `TTDevice::get_communication_device_id()` defaults to `getpid()`.

### Testing
CI. The bug is cross-process so no unit test covers it; verified by auditing the call sites above.

### API Changes
This PR has no API changes.

### DIFF
--- a/device/tt_device/tt_device.cpp
+++ b/device/tt_device/tt_device.cpp
@@ -501,7 +501,7 @@ bool TTDevice::is_remote() { return model_->get_remote_interface() != nullptr; }
 // device to be addressed within.
 int TTDevice::get_communication_device_id() const {
     DeviceProtocol *device_protocol = model_->get_device_protocol();
-    return device_protocol != nullptr ? device_protocol->get_mmio_id() : -1;
+    return device_protocol != nullptr ? device_protocol->get_mmio_id() : -getpid();
 }
 
 // Derived from the transport the device actually has, rather than stored: exactly one of these


micahwhiteTT:
@broskoTT I saw you took a look at this - I updated it to match the base API changes

pjanevskiTT:
Heads-up on an overlap: #3375, in the #3353 stack, changes what a simulated device reports as its MMIO id, and that covers the ttsim case this PR is about.

`TTSimProtocol::get_mmio_id()` now answers `(getpid() << 5) | chip_id` for a host-mode TTSim device, resolved once when the transport attaches. The endpoint's PCI device number stays in the low 5 bits -- the width of the BDF device field it comes from -- and the pid occupies the rest, so each chip of a multi-endpoint image keeps its own lock and two processes never agree on a name. That separates exactly the locks described here: `NON_MMIO` on a gateway reaching a chip over ethernet, and the per-device ARC lock the architecture firmware claims. It also stops a simulated chip aliasing silicon's PCI device of the same number, which the bare endpoint number did.

The reason for flagging it rather than letting both land quietly: after #3375 the branch this PR edits is unreachable for simulation. `TTDeviceModel::get_device_protocol()` is pure virtual, there are three models, and all three answer non-null -- `WormholeTTDeviceModel` and `BlackholeTTDeviceModel` return `protocol_.get()`, and `SimulationTTDeviceModel` returns `tt_sim_protocol_.get()` as of #3375. The `: -1` fallback stays as a defensive answer for a future protocol-less model, but nothing in tree reaches it, so `-getpid()` there would no longer change any lock name.

Two things that may still be worth something, your call:

- Until #3375 merges, this PR is the only thing that fixes the bug on current main, and it costs nothing once #3375 lands. If that window matters, merging it soon is reasonable.
- The unattached half is genuinely still open: RTL simulation and socket client mode never attach a transport, so they keep reporting -1 and keep sharing `NON_MMIO_-1_UNDEFINED` across processes. The fix for that is not in the `TTDevice` fallback though -- it belongs in `TTSimProtocol`, resolving the id from the chip id the device already has at construction instead of only in `attach()`.

If the PR stays open, the description is the part worth revisiting: it promises that two independent ttsim tt-metal tests stop blocking each other, and after #3375 the patch will not be what does that.

micahwhiteTT:
#3375 landed, fixing this issue

PR #3361, opened 2026-09-03
Simulation: report the chip id as the communication device id

pjanevskiTT:
### Issue

#3353

### Description

`SimulationTTDeviceModel::get_communication_device_id()` returned `-1` for every simulated device. That id is not purely informational: `RemoteCommunication` initializes and acquires its `NON_MMIO` mutex keyed on it, so every simulated gateway in a process shared a single lock name, and `TTDeviceError` names the device by it, so any error raised from a simulation device reported chip `-1`.

Pass the simulated chip id through to the model instead. All four construction sites already have it in scope. This is also a prerequisite for #3353 Phase 5, where topology discovery writes this value into `chips_with_mmio` — a simulated cluster would otherwise map every chip to `-1`.

### List of the changes

- `SimulationTTDeviceModel`: take the communication device id at construction and return it
- `TTSimTTDevice`, `RtlSimulationTTDevice`: pass `chip_id` when building the model, in both the host and client constructors

### Testing

CI

### API Changes

- `SimulationTTDeviceModel`'s constructor now takes the communication device id alongside the arch. Made required rather than defaulted so every construction site has to state it.

### DIFF
--- a/device/api/umd/device/tt_device_model/simulation_tt_device_model.hpp
+++ b/device/api/umd/device/tt_device_model/simulation_tt_device_model.hpp
@@ -17,7 +17,7 @@ namespace tt::umd {
 // actually differ between them.
 class SimulationTTDeviceModel : public TTDeviceModel {
 public:
-    explicit SimulationTTDeviceModel(tt::ARCH arch);
+    SimulationTTDeviceModel(tt::ARCH arch, int communication_device_id);
 
     ~SimulationTTDeviceModel() override;
 
@@ -35,6 +35,7 @@ public:
 
 private:
     tt::ARCH arch_;
+    int communication_device_id_;
     std::unique_ptr<ArchitectureImplementation> architecture_impl_;
 };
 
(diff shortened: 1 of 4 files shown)


pjanevskiTT:
Closing as superseded by #3375 rather than rebasing: the method this PR fixes no longer exists.

#3306 removed `get_communication_device_id()` from `TTDeviceModel` and derives the id from the protocol instead:

```cpp
// device/tt_device/tt_device.cpp:500-505
// A simulation model reaches its device in-process, so it has no protocol and no communication
// device to be addressed within.
int TTDevice::get_communication_device_id() const {
    DeviceProtocol *device_protocol = model_->get_device_protocol();
    return device_protocol != nullptr ? device_protocol->get_mmio_id() : -1;
}
```

So the `-1` this PR set out to fix is now a consequence of simulation having no protocol, and #3375 gives it one whose `get_mmio_id()` returns the chip id. The fix arrives through the same mechanism silicon uses, which is better than the constructor argument proposed here.

The consumers named in this PR's description are unchanged and are addressed by #3375: `RemoteCommunication` keying its `NON_MMIO` mutex on the id, and `TTDeviceError` reporting the device. Worth noting that the comment quoted above also goes stale once #3375 lands, since a simulation model will then have a protocol.

PR #3362, opened 2026-09-03
Simulation: resolve a SocArchDescriptor on the simulation model

pjanevskiTT:
### Issue

#3353

### Description

Both architecture-descriptor accessors on `SimulationTTDeviceModel` return `nullptr`, so `init_tt_device`'s closing `construct_soc_descriptor()` takes the null branch, rebuilds the descriptor from the architecture's defaults, and silently discards the one the simulator's `soc_descriptor.yaml` describes. This was latent while `init_tt_device()` could not complete on a simulated device; it completes as of #3376, which is why this sits on top of that stack rather than standing alone.

Resolve and hold one instead: a descriptor supplied by the caller if given, otherwise rebuilt from the simulator's own YAML, otherwise the architecture's constants. Threading `SocDescriptor` through const-correctly would avoid the supplied-descriptor parameter entirely, but it spans ~46 sites across all three models, `TopologyDiscovery` and the arch devices, so it belongs in its own change.

### List of the changes

- `SimulationTTDeviceModel`: resolves and holds one architecture descriptor; its constructor takes the `SocDescriptor` callers already have, since that carries both the architecture and the YAML path. The simulator's YAML is preferred only when it still exists -- a socket client rebuilds its descriptor through a temp file that is deleted immediately, so on that path the recorded path is always stale
- `resolve_soc_arch_descriptor`: runtime-architecture overload, with the existing template delegating to it, so simulation resolves through the same helper the silicon models use

### Testing

CI

### API Changes

`SimulationTTDeviceModel`'s constructor takes a `SocDescriptor` and an optional architecture descriptor; `resolve_soc_arch_descriptor` gains a runtime-architecture overload. Nothing removed or renamed.


device/tt_device_model/simulation_tt_device_model.cpp
```diff
@@ -15,6 +20,19 @@ namespace tt::umd {
 
 namespace {
 
+// A simulated device's SoC descriptor comes from the simulator's own YAML, which need not agree with
+// the architecture's constants, so prefer that YAML when none is supplied -- but only while the file
+// exists. A socket client rebuilds its descriptor through a temp YAML that is deleted at once, so on
+// that path the recorded path is stale and the constants are the right fallback.
+std::shared_ptr<SocArchDescriptor> resolve(
+    const SocDescriptor &soc_descriptor, const std::shared_ptr<SocArchDescriptor> &soc_arch_descriptor) {
+    if (soc_arch_descriptor == nullptr && !soc_descriptor.device_descriptor_file_path.empty() &&
```
broskoTT: Maybe add a log_debug at least, so somebody running sim can see which choice was made?

broskoTT (approved):

device/tt_device_model/simulation_tt_device_model.cpp
pjanevskiTT: Added in bc27b045. Split the three cases apart so each logs its own choice — the caller's descriptor, the simulator's own YAML, or the architecture constants. The constants branch also reports the path it found nothing at, which is the case that is confusing without it.

PR #3374, opened 2026-09-03
Add UBB_BLACKHOLE_BIN6 board type for Bin6 Galaxy DRAM harvesting

### LINKED ISSUE #3178: No way to declare an intentionally DRAM-harvested BH galaxy, so valid systems report as inconsistent
## Summary

There is no way to declare that a board is intentionally DRAM-harvested, so a validly-configured system reports itself as inconsistent. On a Blackhole UBB with one GDDR channel soft-harvested per chip, every chip warns:

```
Chip 0 has inconsistent DRAM harvesting information between harvest mask and number of
harvested. Board ubb_blackhole expects 0 units, but harvesting mask indicates 1 units.
                                                       (cluster_descriptor.cpp:1316)
```

Nothing is actually wrong with this system. The harvesting mask is derived from `ENABLED_GDDR` telemetry, so the mask and the hardware agree exactly; the mismatch is only against a static per-board constant that cannot express the configuration.

## The gap

`expected_dram_harvested_units_map`, `device/api/umd/device/types/cluster_descriptor_types.hpp:247`:

```cpp
static const std::unordered_map<BoardType, uint32_t> expected_dram_harvested_units_map = {
    {BoardType::N150, 0},
    {BoardType::N300, 0},
    {BoardType::P100, 1},
    {BoardType::P150, 0},
    {BoardType::P300, 0},
    {BoardType::UBB, 0},
    {BoardType::UBB_BLACKHOLE, 0},
};
```

Each entry is a single constant per board type, which models **harvesting at manufacture**: `P100` declares 1 because that part ships DRAM-harvested, and `UBB_BLACKHOLE` declares 0 because a stock one is not harvested. Both entries are correct for their shipping SKUs.

What has no representation is **harvesting applied at runtime**. Setting `dram_table.soft_harvest_dram_mask` is a supported capability — the firmware exposes it, and UMD reads the result faithfully out of telemetry — but a board that has been harvested this way has no way to be described as consistent. It is compared against the unharvested constant and reported as an outlier.

The consequence is that the warning cannot distinguish "this chip is misconfigured" from "this chip is deliberately harvested and UMD has no vocabulary for that". Both produce identical output, which devalues the check: on a harvested system it fires for every chip on every `ClusterDescriptor` construction, so a genuinely bad chip would be indistinguishable from the surrounding noise.

For scale, on a 32-chip harvested system this warning is 35–57% of total log output across `distributed_unit_tests`, `unit_tests_api`, and `unit_tests_dispatch`. That volume is not the complaint — if a chip really were in a bad harvesting state, warning per-chip and per-construction is reasonable, and knowing exactly which chip is the point. The volume matters only because every one of these lines is spurious.

## Suggested direction

Give a harvested configuration a way to be declared, so the check can compare against what the system is actually meant to be. Whether that is a harvested board variant, a permitted range, an explicit "variable" sentinel, or something that consults the runtime harvesting state rather than a static table is a design question for the owners — I am flagging the gap rather than proposing the shape of the answer.

Loosening or deleting the check would be the wrong fix: it would lose the ability to detect a genuinely inconsistent chip, which is the thing worth keeping.

## Note on the return value

`verify_harvesting_information()` returns a `bool` that propagates through `verify_cluster_descriptor_info()`, and every call site ignores it:

```
device/cluster_descriptor.cpp:142   desc->verify_cluster_descriptor_info();
device/cluster_descriptor.cpp:505   remapped->verify_cluster_descriptor_info();
device/cluster_descriptor.cpp:585   desc->verify_cluster_descriptor_info();
device/topology/topology_discovery.cpp:523  cluster_desc->verify_cluster_descriptor_info(...);
```

Mentioned only because it means the check is currently advisory — the log line is its entire effect. Not a request to change that.

## Environment

tt-umd at `9bbe7bc9` (as vendored in tt-metal `main`), Blackhole UBB, 32 chips, one GDDR channel soft-harvested per chip via `dram_table.soft_harvest_dram_mask`. Firmware 19.13.2, KMD 2.9.0. Verified against the same system with all 8 banks enabled, where the warning does not appear.

ajovanovicTT:
### Issue
#3178 

### Description
UBB_BLACKHOLE always expects 0 DRAM harvested units, so a Bin6 Galaxy with 1 GDDR channel harvested reports as inconsistent on every chip during ClusterDescriptor construction, even though the configuration is valid. Bin6 is a distinct SKU with its own board ID, so it gets its own BoardType rather than a change to the existing expectation.

### List of the changes
 - Added BoardType::UBB_BLACKHOLE_BIN6
 - Mapped UPI 0x202 to the new board type in board_upi_map
 - Added name map entries, chip count, and expected tensix/dram/eth harvested unit entries for the new board type
 - Updated get_tray_id, is_running_on_6u, and the nanobind BoardType enum to treat UBB_BLACKHOLE_BIN6 like UBB_BLACKHOLE
  - According test updates

### Testing
Ci only, needs to be tested on bin6 glx.

### API Changes
New BoardType enum value

### DIFF
--- a/device/api/umd/device/types/cluster_descriptor_types.hpp
+++ b/device/api/umd/device/types/cluster_descriptor_types.hpp
@@ -36,6 +36,8 @@ enum BoardType : uint32_t {
     UBB_BLACKHOLE,
     QUASAR_BOARD,
     UNKNOWN,
+    // Appended after UNKNOWN so the values above keep their numbering/ABI.
+    UBB_BLACKHOLE_BIN6,
 };
 
 static_assert(N150 == 3, "N150 must be 3");
@@ -49,6 +51,7 @@ static_assert(UBB_WORMHOLE == 9, "WH_UBB must equal UBB");
 static_assert(UBB_BLACKHOLE == 10, "BH_UBB must be 10");
 static_assert(QUASAR_BOARD == 11, "QUASAR must be 11");
 static_assert(UNKNOWN == 12, "UNKNOWN must be 12");
+static_assert(UBB_BLACKHOLE_BIN6 == 13, "BH_UBB_BIN6 must be 13");
 
 // Small performant hash combiner taken from boost library.
 // Not using boost::hash_combine due to dependency complications.
@@ -101,6 +104,7 @@ inline const std::unordered_map<std::string_view, BoardType> board_type_name_map
     {"p300", BoardType::P300},
     {"ubb", BoardType::UBB},
     {"ubb_blackhole", BoardType::UBB_BLACKHOLE},
+    {"ubb_blackhole_bin6", BoardType::UBB_BLACKHOLE_BIN6},
     {"ubb_wormhole", BoardType::UBB_WORMHOLE},
     {"quasar", BoardType::QUASAR_BOARD},
     {"unknown", BoardType::UNKNOWN},
@@ -120,6 +124,7 @@ inline const std::unordered_map<BoardType, std::string_view> board_type_canonica
     {BoardType::P300, "p300"},
     {BoardType::UBB, "ubb"},
     {BoardType::UBB_BLACKHOLE, "ubb_blackhole"},
+    {BoardType::UBB_BLACKHOLE_BIN6, "ubb_blackhole_bin6"},
     {BoardType::UBB_WORMHOLE, "ubb_wormhole"},
     {BoardType::QUASAR_BOARD, "quasar"},
     {BoardType::UNKNOWN, "unknown"},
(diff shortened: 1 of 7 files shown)


device/api/umd/device/types/cluster_descriptor_types.hpp
```diff
@@ -36,6 +36,8 @@ enum BoardType : uint32_t {
     UBB_BLACKHOLE,
     QUASAR_BOARD,
     UNKNOWN,
```
broskoTT, line 39: UNKNOWN should be last

broskoTT: Also I don't think this is serialized anywhere, at least up to know, so it is should be safe to put UBB_BLACKHOLE_BIN6 right after UBB_BLACKHOLE

broskoTT (approved):

device/api/umd/device/types/cluster_descriptor_types.hpp
ajovanovicTT: Checked - tt-metal does serialize BoardType as a raw int in physical_system_descriptor_serialization.cpp. No existing fixtures use the values this reorder touches and everything else compares by name, so risk seems low. Moved UBB_BLACKHOLE_BIN6 right after UBB_BLACKHOLE.

PR #3376, opened 2026-09-03
Simulation: read TTSim devices with the architecture's own firmware

pjanevskiTT:
### Issue

#3353

### Description

A host-mode TTSim Wormhole or Blackhole image publishes real firmware state: its ARC is up before host code runs, its telemetry table is populated, and both are reachable over the protocol #3375 adds. So the architecture's own `DeviceFirmware` can read a simulated device, and simulation no longer has to answer firmware questions itself -- which it currently gets wrong twice, reporting `noc_translation_enabled = false` where the simulator's NIU_CFG says `1`, and a hardcoded `eth_harvesting_mask = 0x120` where its telemetry implies `0x3000`.

Quasar, RTL simulation and client mode keep the firmware that reports nothing, which for them is accurate: Quasar models no ARC, ethernet or host BAR path, RTL models no ARC, and a client-mode device has no local simulator to read.

### List of the changes

- `SimulationTTDeviceModel`: `use_arch_device_firmware()` installs `WormholeDeviceFirmware`/`BlackholeDeviceFirmware` and serves their telemetry reader and info provider onward. Both are opted out of the KMD half of the ARC message lock (#3401), which a simulator has no `/dev/tenstorrent` node for
- `TTSimProtocol::attach`: installs the architecture firmware once the transport works, since an architecture firmware reads the device in its constructor and cannot exist before then

### Testing

CI

### API Changes

`SimulationTTDeviceModel` gains `use_arch_device_firmware()` and now returns non-null from `get_firmware_telemetry_reader()`/`get_firmware_info_provider()` for host-mode TTSim WH/BH. Nothing removed or renamed.


device/api/umd/device/tt_device/protocol/tt_sim_protocol.hpp
```diff
@@ -36,7 +37,10 @@ class SimulationTTDevice;
  */
 class TTSimProtocol : public DeviceProtocol, public PcieInterface {
 public:
-    TTSimProtocol() = default;
+    // The model that owns this protocol, so that attaching a working transport can hand it the
+    // architecture's firmware -- which cannot exist any earlier, since it reads the device as it is
+    // constructed.
+    explicit TTSimProtocol(SimulationTTDeviceModel* model);
```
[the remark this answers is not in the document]
pjanevskiTT: `TTSimProtocol` is introduced by #3375, one PR down this stack — it does not exist on `main`, so there are no existing consumers to break, and the only construction site is `SimulationTTDeviceModel`'s member initialiser. Nothing released is removed, so the API section is accurate as written.

device/api/umd/device/tt_device_model/simulation_tt_device_model.hpp
```diff
@@ -31,19 +32,40 @@ class SimulationTTDeviceModel : public TTDeviceModel {
     // whenever one is present, and refuses to proceed without a detector to ask.
     HangDetector *get_hang_detector() override;
 
+    // Replace the firmware that reports nothing with the architecture's own, which reads what the
+    // simulator publishes. Called by a device that has attached a usable protocol, and only then: an
+    // architecture firmware reads the device in its constructor -- Blackhole resolves its ARC
+    // coordinates from NOC translation state over BAR -- so it cannot exist before the transport
+    // works. A no-op for architectures a simulator publishes no firmware state for.
+    void use_arch_device_firmware();
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed in a818ea5f. The method is private now with `TTSimProtocol` as a friend: `attach()` is the only caller, and the only place that can know the precondition holds, so the invariant is structural rather than a comment.

device/tt_device_model/simulation_tt_device_model.cpp
```diff
@@ -46,8 +49,52 @@ DeviceProtocol *SimulationTTDeviceModel::get_device_protocol() { return tt_sim_p
...
+                /*remote_interface=*/nullptr,
+                architecture_impl_.get(),
+                /*kmd_lock_available=*/false);
+            auto *raw = firmware.get();
+            telemetry_reader_lookup_ = [raw]() { return raw->get_firmware_telemetry_reader(); };
+            info_provider_lookup_ = [raw]() { return raw->get_firmware_info_provider(); };
+            device_firmware_ = std::move(firmware);
+            break;
+        }
+        case tt::ARCH::BLACKHOLE: {
```
ajovanovicTT: optional: The WORMHOLE_B0 and BLACKHOLE branches duplicate the same pattern. Worth factoring into a small template helper (e.g. install<WormholeDeviceFirmware>(args...)) to avoid the two branches drifting apart if one is updated and the other isn't.

device/api/umd/device/tt_device_model/simulation_tt_device_model.hpp
ajovanovicTT: Forward declaration class SimulationDeviceFirmware; looks unused in this header: device_firmware_ is typed as std::unique_ptr<DeviceFirmware> (the base class), and the only place SimulationDeviceFirmware is actually named is in simulation_tt_device_model.cpp, which already includes its full header directly. Safe to remove.

ajovanovicTT (approved):

device/tt_device_model/simulation_tt_device_model.cpp
pjanevskiTT: Done in d88139f2. The shared tail -- keep the firmware, serve the telemetry reader and info provider it owns -- moved into a local `install` generic lambda. It takes the firmware already constructed, so the arguments that actually differ between the two (Blackhole has no `remote_interface`) stay at the call site.

device/api/umd/device/tt_device_model/simulation_tt_device_model.hpp
pjanevskiTT: Removed in d88139f2. Confirmed unused: `SimulationDeviceFirmware` is named only in `simulation_tt_device_model.cpp`, which includes its header directly.

PR #3380, opened 2026-09-04
[Base API] Retire the raw ARC APB accessors from TTDevice

nbuncicTT:
### Issue
#2872

### Description
`read_from_arc_apb` / `write_to_arc_apb` were the last pure virtuals on `TTDevice`, keeping the per-arch subclasses alive to carry APB routing the firmware component family already owns (`WormholeArcWindow` / `BlackholeArcApb`). Raw register-window access is not part of the spec on `TTDevice`, `TTDeviceModel`, or `DeviceFirmware` — and no client uses it (verified across tt-metal, nanobind, tools and tests). This PR deletes the accessors and migrates the two internal consumers onto concrete components in the same diff. CSM needs nothing: no public CSM API exists; the window is already a private detail of `WormholeDeviceFirmware`.

### List of the changes
- `TTDevice::read_from_arc_apb` / `write_to_arc_apb` deleted, with the Wormhole, Blackhole and simulation overrides and Blackhole's AXI-availability helpers.
- The SPI devices are architecture-committed by construction, so each holds its concrete firmware (`WormholeDeviceFirmware` / `BlackholeDeviceFirmware`, resolved and checked once in the constructor) and uses its public ARC APB wrappers directly, routing on the thread-selected NOC exactly as before.
- `BlackholeDeviceFirmware` gains the `write` wrapper (one line onto `BlackholeArcApb::write`).
- `get_refclk_counter` (spec TTDevice API, kept) moves its per-arch body verbatim onto `DeviceFirmware` as `get_refclk_counter(NocId)` — the counter lives in the ARC reset unit, reached through the firmware component's windows. The facade forwards; simulation firmware throws, as the deleted path did. Proposed spec amendment: add `get_refclk_counter` to DeviceFirmware's interface.

Behavior notes (pre-existing component semantics, or error-text only): the Wormhole window validates harder than the deleted accessors (zero-length, word-size on BAR/JTAG routes, end-of-window overrun) — semantics that landed with `WormholeArcWindow`, covered by its baremetal tests; JTAG APB routes honor the selected NOC instead of being pinned to NOC0 — the already-approved component contract; simulation `get_refclk_counter` error text becomes "The refclk counter is not available for a simulated device." (same `RuntimeError`).

### Testing
Builds with simulation ON and OFF; `baremetal_tests` 254/254; CI.

### API Changes
This PR has API changes: `TTDevice::read_from_arc_apb` / `write_to_arc_apb` are removed (no known users); `DeviceFirmware` gains `get_refclk_counter(NocId)`.


broskoTT (approved):
I looked at this thoroughly. 
I hope adding get_refclk to Dev Firmware is intentional @nbuncicTT 
I have some comments and modifications for SPITTDevice, but it makes more sense in a following PR which I'll make

PR #3385, opened 2026-09-04
sim: add CRAQ mapped-memory and multi-rank support

nkapreTT:
CRAQ-backed mapped buffers and cross-rank Ethernet need explicit address bounds, reliable rendezvous, and descriptor lifetime management. This change configures a dedicated mapped-memory iATU region and connects same-host remote endpoints through a checked, capability-probed FD transport.

The review revision fixes overflow and unaligned-buffer address translation, preflights allocation capacity, makes IPC setup transactional with one deadline and a readiness handshake, rejects invalid identities/paths, and detaches simulator references before closing descriptors. Session directories are explicitly supplied by the launcher and shared only by that job. Nonblocking retry is preserved. The launch and ownership contract is documented in `docs/craq_multi_rank.md`.

It also restores the canonical GTest package name, removes unused factory additions, and confines reset no-ops to SWEmule. Qualification exposed a pre-existing mismatch with the declared tt-emule dependency; private pools now compile against that pin, while unsupported shared-pool requests fail explicitly without changing the dependency.

## Dependency and scope

- Cross-rank FD transport requires tenstorrent/craq-sim#399, now merged into CRAQ main as `ecc58159`. This PR remains draft pending external RTL CI qualification.
- Current checked FD transport supports WH/BH; Quasar is explicitly unsupported. Local simulation retains compatibility with older simulator libraries.
- The inspected `tt-metal:nkapre/craq-sim` consumer and Blaze integration use ClusterOptions constructors, not the removed factory additions. Their existing dependency pins have not been changed by this revision. Older Metal revisions call `Cluster::get_static_tlb_window`, which current upstream UMD removed; those consumers need a compatible Metal revision or adaptation before updating their pin.
- The existing simulation-server allocator is not used for FIFO-only sessions because its stale-directory pruning can remove active FIFO jobs.

## Validation of review revision

- After merging current main, 43 mapped-memory tests passed using current-source objects without an older UMD library. These include upstream buffer-lifetime tests and our boundary/alignment regressions. Four new cases also failed against the original pre-fix implementation.
- Eight no-card IPC tests passed, including delayed incoming writers, session isolation, invalid/stale paths, deadline handling, MAC collision checks, cleanup, and SIGPIPE-safe handshake failure.
- Current-source communicator ownership regression passed against patched CRAQ BH: duplicate attachment rollback, detach-before-close, and reattachment while another communicator keeps the simulator alive.
- GTest local-only/disconnected configuration passed; the old package name reproduces the configuration failure.
- Current cluster source compiles with simulation enabled and disabled; communicator and SWEmule source compile with warnings as errors, including the exact declared tt-emule headers.
- Repository pre-commit hooks and patch checks passed.

Current main has been merged and conflicts resolved while preserving upstream buffer-owned allocation lifetime and endpoint enumeration. Paired CRAQ/UMD CI passed for UMD `516e1223` in [run 35375441707](https://github.com/tenstorrent/craq-sim/actions/runs/35375441707). These targeted results are not a claim of completed SWEmule runtime, live cross-chip mapped DMA, or Metal/Blaze firmware qualification. Previous-head green checks do not qualify this revision.

## RTL build follow-up

The external RTL job failed during its Debug build, before RTL tests ran: clang-tidy rejected a combined signal-set declaration, mismatched declaration/definition parameter names, and a discarded unique_ptr release result. Commit `ebee6f4e` fixes all three without disabling checks or changing descriptor lifetime behavior.

Clang-tidy 20.1.8 with the repository configuration reproduces the three errors on the old code and passes cluster.cpp, tt_sim_communicator.cpp, and test_sim_multichip.cpp after the fix, including Debug/trace settings matching the supplied CI log. All eight rebuilt IPC tests and pre-commit hooks pass. New-head CI and the external RTL result are pending.

### DIFF
--- a/device/api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
+++ b/device/api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
@@ -31,10 +31,18 @@ public:
     // hugepage physical_address (= iATU target). The simulator's host-side DMA router (dma_route) then
     // routes by which chip's [host_base, host_base + PER_CHIP_HOST_STRIDE) window contains the address.
     static constexpr uint64_t PER_CHIP_HOST_STRIDE = 1ULL << 36;  // 64 GiB; >> per-chip sysmem, non-overlapping
+    // The modeled outbound iATU uses 32-bit base/limit registers. Mapped-buffer IOVAs share this
+    // device-visible window with the fixed-stride host-memory channels. WH's modeled PCIe DMA
+    // aperture ends at offset 0xFFFE0000, so use that conservative limit for every architecture.
+    static constexpr uint64_t DEVICE_IO_WINDOW_SIZE = 0xFFFE0000ULL;
 
     uint64_t get_host_base() const { return host_base_; }
 
     uint64_t get_host_region_size() const { return PER_CHIP_HOST_STRIDE; }
+
+    uint64_t get_mapped_arena_offset() const { return system_memory_size_; }
+
+    uint64_t get_mapped_arena_size() const { return DEVICE_IO_WINDOW_SIZE - system_memory_size_; }
 
     bool pin_or_map_sysmem_to_device() override;
 
--- a/device/api/umd/device/cluster.hpp
+++ b/device/api/umd/device/cluster.hpp
@@ -173,6 +173,21 @@ public:
     // Explicit to prevent implicit conversion from ClusterOptions at call sites.
     explicit Cluster(ClusterOptions options);
 
+    // Build ClusterOptions inside libtt-umd to avoid passing STL-bearing options across a DSO boundary.
+    static std::unique_ptr<Cluster> create_silicon_cluster(
+        std::optional<uint32_t> num_host_mem_ch_per_mmio_device = std::nullopt);
+    static std::unique_ptr<Cluster> create_single_chip_simulation_cluster(
+        uint32_t num_host_mem_ch_per_mmio_device, ChipId chip_id, const char* simulator_path);
+    static std::unique_ptr<Cluster> create_simulation_cluster_with_descriptor(
+        uint32_t num_host_mem_ch_per_mmio_device,
+        const char* simulator_path,
+        const char* sdesc_path,
+        ClusterDescriptor* cluster_descriptor);
+    static std::unique_ptr<Cluster> create_mock_cluster(
+        const char* sdesc_path, ClusterDescriptor* cluster_descriptor = nullptr);
+    static std::unique_ptr<Cluster> create_swemule_cluster(
+        const char* sdesc_path, ClusterDescriptor* cluster_descriptor = nullptr);
+
     /**
      * Closing the device. Should undo everything that was done in the constructor. Break created links, free memory,
      * leave the device in a state where it can be re-initialized.
--- a/device/chip/chip.cpp
+++ b/device/chip/chip.cpp
@@ -88,11 +88,19 @@ RiscType Chip::get_risc_reset_state(CoreCoord core) {
 }
 
 void Chip::assert_risc_reset(CoreCoord core, const RiscType selected_riscs) {
-    get_tt_device()->assert_risc_reset(core, selected_riscs);
+    TTDevice* tt_device = get_tt_device();
+    if (tt_device == nullptr) {
+        return;
+    }
+    tt_device->assert_risc_reset(core, selected_riscs);
 }
 
 void Chip::deassert_risc_reset(CoreCoord core, const RiscType selected_riscs, bool staggered_start) {
-    get_tt_device()->deassert_risc_reset(core, selected_riscs, staggered_start);
+    TTDevice* tt_device = get_tt_device();
+    if (tt_device == nullptr) {
+        return;
+    }
+    tt_device->deassert_risc_reset(core, selected_riscs, staggered_start);
 }
 
 void Chip::assert_risc_reset(const RiscType selected_riscs) {
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -632,11 +632,23 @@ Cluster::Cluster(ClusterOptions options) {
 
         std::vector<PendingFdLink> pending_fd_links;
         const char* eth_ipc_env = std::getenv("TT_SIM_ETH_IPC_DIR");
-        const std::string eth_ipc_dir = eth_ipc_env ? std::string(eth_ipc_env) : std::string("/tmp/ttsim_eth_ipc");
-        ::mkdir(eth_ipc_dir.c_str(), 0777);
+        const std::string eth_ipc_dir =
+            eth_ipc_env != nullptr && eth_ipc_env[0] != '\0' ? eth_ipc_env : "/tmp/ttsim_eth_ipc";
+        const int mkdir_result = ::mkdir(eth_ipc_dir.c_str(), 0777);
+        UMD_ASSERT(
+            mkdir_result == 0 || errno == EEXIST,
+            error::RuntimeError,
+            fmt::format("Failed to create TTSim Ethernet IPC directory {}: {}", eth_ipc_dir, std::strerror(errno)));
         auto eth_fifo_path = [&](uint64_t source, int source_channel, uint64_t destination, int destination_channel) {
             return eth_ipc_dir + "/eth_" + std::to_string(source) + "_" + std::to_string(source_channel) + "__" +
                    std::to_string(destination) + "_" + std::to_string(destination_channel) + ".fifo";
+        };
+        auto ensure_fifo = [](const std::string& path) {
+            const int mkfifo_result = ::mkfifo(path.c_str(), 0666);
+            UMD_ASSERT(
+                mkfifo_result == 0 || errno == EEXIST,
+                error::RuntimeError,
+                fmt::format("Failed to create TTSim Ethernet FIFO {}: {}", path, std::strerror(errno)));
         };
 
         // For every connected eth pair (chip_a:chan_a <-> chip_b:chan_b),
@@ -692,9 +704,9 @@ Cluster::Cluster(ClusterOptions options) {
                 local_comm->register_eth_endpoint(uint32_t(local_channel), eth_sim_mac(local_chip, int(local_channel)));
                 const std::string write_path = eth_fifo_path(local_uid, int(local_channel), remote_uid, remote_channel);
                 const std::string read_path = eth_fifo_path(remote_uid, remote_channel, local_uid, int(local_channel));
-                ::mkfifo(write_path.c_str(), 0666);
-                ::mkfifo(read_path.c_str(), 0666);
-                const int read_fd = ::open(read_path.c_str(), O_RDONLY | O_NONBLOCK);
+                ensure_fifo(write_path);
+                ensure_fifo(read_path);
+                const int read_fd = ::open(read_path.c_str(), O_RDONLY | O_NONBLOCK | O_CLOEXEC);
                 if (read_fd < 0) {
                     log_warning(
                         tt::LogEmulationDriver,
@@ -709,7 +721,7 @@ Cluster::Cluster(ClusterOptions options) {
         for (auto& link : pending_fd_links) {
             int write_fd = -1;
             for (int attempt = 0; attempt < 120000 && write_fd < 0; ++attempt) {
-                write_fd = ::open(link.write_path.c_str(), O_WRONLY | O_NONBLOCK);
+                write_fd = ::open(link.write_path.c_str(), O_WRONLY | O_NONBLOCK | O_CLOEXEC);
                 if (write_fd < 0) {
                     if (errno != ENXIO) {
                         break;
--- a/device/cluster.cpp
+++ b/device/cluster.cpp
@@ -4,13 +4,19 @@
 
 #include "api/umd/device/cluster.hpp"
 
+#include <fcntl.h>
 #include <fmt/format.h>
 #include <fmt/ranges.h>
+#include <sys/stat.h>
+#include <unistd.h>
 
 #include <algorithm>
+#include <cerrno>
 #include <chrono>
 #include <cstddef>
 #include <cstdint>
+#include <cstdlib>
+#include <cstring>
 #include <filesystem>
 #include <initializer_list>
 #include <map>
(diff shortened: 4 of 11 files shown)


third_party/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ include(${PROJECT_SOURCE_DIR}/cmake/ttexalens_private.cmake)
 ####################################################################################################################
 if(TT_UMD_BUILD_TESTS)
     CPMAddPackage(
-        NAME GTest
+        NAME googletest
```
pjanevskiTT: This reverts 118311c2 ("Fix local googletest", #2188), which renamed the CPM package from `googletest` to `GTest` precisely so `CPM_LOCAL_PACKAGES_ONLY` builds resolve it: CPM resolves by calling `find_package(<NAME> <VERSION>)`, and `GTest` is the canonical config/module name while `googletest` is not. Reverting it brings back exactly the failure that PR documents -- `CPM: googletest not found via find_package(googletest 1.13.0)`, then the `FETCHCONTENT_FULLY_DISCONNECTED` error, then `Cannot specify compile options for target "gtest"`. It also looks unrelated to CRAQ; can it be dropped from this PR?

device/cluster.cpp
```diff
@@ -616,6 +622,35 @@ Cluster::Cluster(ClusterOptions options) {
...
+            tt::umd::TTSimCommunicator* comm;
+            uint32_t channel;
+            std::string write_path;
+            int read_fd;
+        };
+
+        std::vector<PendingFdLink> pending_fd_links;
+        const char* eth_ipc_env = std::getenv("TT_SIM_ETH_IPC_DIR");
+        const std::string eth_ipc_dir =
+            eth_ipc_env != nullptr && eth_ipc_env[0] != '\0' ? eth_ipc_env : "/tmp/ttsim_eth_ipc";
```
pjanevskiTT: A single fixed path shared by every simulation run on the machine is going to bite: two unrelated multi-rank ttsim runs on the same CI box compute identical FIFO names and connect to each other. `SimulationServerSocket::allocate_server_directory()` already solves this for the socket server (#3099 gave each host its own directory for exactly this reason) -- reusing that gets the isolation for free.

Two related points on the same lines: the directory is created 0777 (and accepted with whatever owner/mode it already has) with FIFOs at 0666, and `ensure_fifo` treats EEXIST as success without checking the entry is actually a FIFO -- so a stale regular file, or one a local user pre-creates, is opened `O_WRONLY` successfully and every eth frame is written into it while the run looks wired. An `lstat`/`S_ISFIFO` + ownership check and 0700/0600 would close that. And nothing ever unlinks the FIFOs, so later runs reuse whatever is left behind.

device/cluster.cpp
```diff
@@ -616,6 +622,35 @@ Cluster::Cluster(ClusterOptions options) {
...
+        const int mkdir_result = ::mkdir(eth_ipc_dir.c_str(), 0777);
```
pjanevskiTT: This runs for every `.so` simulation cluster, before `remote_conns` is consulted, and its failure is fatal -- so an ordinary single-rank ttsim run now throws out of the `Cluster` constructor if `/tmp/ttsim_eth_ipc` can't be created: read-only `/tmp` in a container, EACCES because another user got there first, or ENOENT when `TT_SIM_ETH_IPC_DIR` names a nested path (raw `mkdir` can't create parents; `std::filesystem::create_directories(..., ec)` can, and is what the rest of the tree uses). Could the whole mkdir/FIFO block move inside a `if (!remote_conns.empty())` guard?

device/cluster.cpp
```diff
@@ -650,6 +685,61 @@ Cluster::Cluster(ClusterOptions options) {
                     mac_b);
             }
         }
+
+        const auto& chip_uids = cluster_desc->get_chip_unique_ids();
```
pjanevskiTT: `get_chip_unique_ids()` can't be used for this. `has_authentic_chip_unique_ids()` exists to say whether the ids came from the source or were synthesized, and its doc comment states the synthesized ones "must not be used as a cross-process identity" -- which is precisely what a FIFO path is. The SWEMULE path in this same file (~line 250) already gates on that predicate with a comment describing this exact collision. With a ttsim `cluster_descriptor.yaml` that has no `chip_unique_ids` block, both ranks synthesize the same ids and therefore the same FIFO names, so a rank silently wires to the wrong peer (or itself) instead of failing loudly. Please gate this loop on `has_authentic_chip_unique_ids()` and throw otherwise.

device/cluster.cpp
```diff
@@ -650,6 +685,61 @@ Cluster::Cluster(ClusterOptions options) {
...
+            }
+            auto* local_comm = get_comm(local_chip);
+            if (local_comm == nullptr) {
+                continue;
+            }
+            const uint64_t local_uid = uid_it->second;
+            for (const auto& [local_channel, remote] : channel_map) {
+                const uint64_t remote_uid = std::get<0>(remote);
+                const int remote_channel = std::get<1>(remote);
+                local_comm->register_eth_endpoint(uint32_t(local_channel), eth_sim_mac(local_chip, int(local_channel)));
```
pjanevskiTT: Two gaps versus the in-process loop just above, which registers both endpoints *and* calls `register_peer` in both directions so "firmware sees correctly wired neighbours at boot time":

1. `eth_sim_mac(local_chip, ...)` encodes the *local* chip id, but `create_constrained_cluster_descriptor` renumbers visible chips to 0-based consecutive indices whenever a rank sees a subset of the cluster (cluster_descriptor.cpp:491-508). So in a 2-rank job both ranks' local chips start at 0, and the two ends of a cross-rank link are handed the **same** MAC. The uids used for the FIFO names survive that remap (only keys are remapped, values are preserved) but the MACs don't -- deriving the cross-rank MAC from the chip uid rather than the logical id would fix it.

2. There is no `register_peer`/DEST_MAC equivalent for these links, which makes sense (it needs the peer's `dev_handle_`, and that doesn't exist across processes) -- but then what supplies the peer MAC? Does `libttsim_configure_eth_link_fd` pre-write it, or does the eth model learn it from the first frame? If neither, cross-rank neighbours won't look wired at boot the way local pairs do.

device/cluster.cpp
```diff
@@ -616,6 +622,35 @@ Cluster::Cluster(ClusterOptions options) {
                 first_chip_comm->switch_reset();
             }
         }
+
+        struct PendingFdLink {
```
pjanevskiTT: Stepping back on the design: UMD already has a cross-process simulation mechanism -- `serve_simulation_devices_over_sockets` / `SimulationConnector` / `SimulationServerSocket`, with the `sim_server` tool, per-server directory allocation, `prune` for stale directories, and `docs/SIMULATION_SERVER.md`. This adds a second, parallel IPC channel with its own env var, its own `/tmp` layout, no cleanup and no documentation.

The use case is genuinely different (wiring two independent simulators' eth tiles vs. sharing one simulation), but the plumbing -- directory allocation, lifetime, collision avoidance, teardown -- is the same problem that code already solves, and the comments below flag several places where this copy gets it wrong. Could this be built on that machinery instead? Either way, `docs/SIMULATION_SERVER.md` or a sibling doc should cover `TT_SIM_ETH_IPC_DIR` and how a multi-rank run is meant to be launched.

device/cluster.cpp
```diff
@@ -650,6 +685,61 @@ Cluster::Cluster(ClusterOptions options) {
...
+                        read_path,
+                        std::strerror(errno));
+                    continue;
+                }
+                pending_fd_links.push_back({local_comm, uint32_t(local_channel), write_path, read_fd});
+            }
+        }
+        for (auto& link : pending_fd_links) {
+            int write_fd = -1;
+            for (int attempt = 0; attempt < 120000 && write_fd < 0; ++attempt) {
```
pjanevskiTT: 120000 attempts x `usleep(1000)` is a hard-coded 120 s stall, and it sits inside the per-link loop, so k links whose peers never appear block `Cluster::Cluster` for k x 120 s with nothing logged while it waits -- then it warns and hands back a cluster with dead links. Start one rank of a galaxy without its peer (crashed rank, wrong `TT_SIM_ETH_IPC_DIR`, leftover FIFOs from a previous run) and this reads as an unexplained CI timeout. Please make the budget a named constant or env knob, log progress, and fail rather than continue unwired.

device/cluster.cpp
```diff
@@ -650,6 +685,61 @@ Cluster::Cluster(ClusterOptions options) {
...
+                write_fd = ::open(link.write_path.c_str(), O_WRONLY | O_NONBLOCK | O_CLOEXEC);
```
pjanevskiTT: `O_NONBLOCK` is needed on the `open()` only so a missing reader returns ENXIO instead of blocking, but it stays set on the fd handed to the simulator. Unless ttsim's eth model loops on `EAGAIN`/short writes, frames get silently dropped once the peer falls behind and the 64 KiB pipe buffer fills -- which will surface as a nondeterministic fabric hang rather than an error. Was keeping it set deliberate? If not, clearing it after the handshake (`fcntl(write_fd, F_SETFL, fcntl(write_fd, F_GETFL) & ~O_NONBLOCK)`) makes the link back-pressure instead.

device/cluster.cpp
```diff
@@ -650,6 +685,61 @@ Cluster::Cluster(ClusterOptions options) {
...
+            if (write_fd < 0) {
+                log_warning(
+                    tt::LogEmulationDriver,
+                    "TTSim eth cross-rank write open failed for {}: {}",
+                    link.write_path,
+                    std::strerror(errno));
+                ::close(link.read_fd);
+                continue;
+            }
+            link.comm->configure_eth_link_fd(link.channel, write_fd, link.read_fd);
```
pjanevskiTT: This is fire-and-forget, but `configure_eth_link_fd` returns `void` and silently no-ops when `!multichip_mode_ || !dev_handle_ || !pfn_libttsim_configure_eth_link_fd_`. That pfn is resolved with a bare `dlsym` in the optional-extension block, so any libttsim that doesn't export `libttsim_configure_eth_link_fd` makes every call here a no-op -- and in the self-describing multi-MMIO/BDF path `multichip_mode_` is never set at all. Either way you get two leaked fds per remote channel per `Cluster` construction and a cluster that looks wired but has no links, with no diagnostic. Returning `bool` and closing both fds + warning on `false` would cover it.

Nothing owns these fds otherwise: if `ensure_fifo` throws partway through the first loop, every `read_fd` already pushed into `pending_fd_links` leaks, and `Cluster` teardown never closes the ones that were handed over.

device/chip/chip.cpp
```diff
@@ -88,11 +88,19 @@ RiscType Chip::get_risc_reset_state(CoreCoord core) {
 }
 
 void Chip::assert_risc_reset(CoreCoord core, const RiscType selected_riscs) {
-    get_tt_device()->assert_risc_reset(core, selected_riscs);
+    TTDevice* tt_device = get_tt_device();
```
pjanevskiTT, line 93: I think this guard is at the wrong layer, and incomplete.

Wrong layer: `MockChip` already overrides all five reset methods as no-ops (mock_chip.cpp:78-86), so this is dead code for mock. The chip it actually rescues is `SWEmuleChip` -- `get_tt_device()` returns nullptr (sw_emule_chip.cpp:254) and it overrides only `deassert_risc_resets()`. Adding the same overrides there matches the existing pattern and keeps the base strict; as written, a `LocalChip`/`RemoteChip` whose `tt_device_` is null after a failed init now silently skips the reset instead of failing, turning a clear error into a later device hang.

Incomplete: `Chip::get_risc_reset_state(core)` two lines above still dereferences `get_tt_device()` unconditionally, so the same caller set still crashes on exactly the chip type this is meant to make safe.

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -178,6 +179,12 @@ void TTSimTTDevice::initialize_backend() {
             HugepageMapping m = sysmem_manager_->get_hugepage_mapping(ch);
             configure_iatu_region(ch, m.physical_address, m.mapping_size);
         }
+        auto* sim_mgr = static_cast<SimulationSysmemManager*>(sysmem_manager_.get());
+        const uint64_t arena_offset = sim_mgr->get_mapped_arena_offset();
+        const uint64_t arena_size = sim_mgr->get_mapped_arena_size();
+        if (arena_size > 0) {
+            configure_iatu_region_at(nch, arena_offset, sim_mgr->get_host_base() + arena_offset, arena_size);
```
pjanevskiTT: Using `nch` as the arena's region index runs into the silicon layout this file deliberately mirrors: `WormholeTTDevice::configure_iatu_region` remaps channel 3 to region **4** ("Hack use region 4 for channel 3", wormhole_tt_device.cpp:71-75), so with `num_host_mem_ch_per_mmio_device == 4` on WH the arena takes the slot real WH uses for channel 3. Nothing collides in the sim model today because this class doesn't do that remap, but it's a trap if the model ever adopts it. There's also no bound on the region index, and this file already documents that unimplemented iATU register offsets throw `UnimplementedFunctionality` -- a named constant above the channel range plus an assert against the model's region count would be safer than `nch`.

device/api/umd/device/chip_helpers/simulation_sysmem_manager.hpp
```diff
@@ -31,11 +31,19 @@ class SimulationSysmemManager : public SysmemManager {
...
+    // aperture ends at offset 0xFFFE0000, so use that conservative limit for every architecture.
+    static constexpr uint64_t DEVICE_IO_WINDOW_SIZE = 0xFFFE0000ULL;
 
     uint64_t get_host_base() const { return host_base_; }
 
     uint64_t get_host_region_size() const { return PER_CHIP_HOST_STRIDE; }
 
+    uint64_t get_mapped_arena_offset() const { return system_memory_size_; }
+
+    uint64_t get_mapped_arena_size() const { return DEVICE_IO_WINDOW_SIZE - system_memory_size_; }
```
pjanevskiTT: Worth noting how tight this gets at the top end: `init_sysmem` gives 4 channels `4 GiB - 256 MiB` (`0xF0000000`), so with 4 host channels the arena is only `0xFE0000` (~16 MiB) and anything larger trips the new `UMD_ASSERT` in `map_sysmem_buffer`. The subtraction is also unsigned and silently depends on `init_sysmem` keeping `system_memory_size_ <= DEVICE_IO_WINDOW_SIZE`; an assert here would keep the two in sync if that cap ever moves.

tests/api/test_simulation_sysmem_manager.cpp
```diff
@@ -54,6 +54,25 @@ TEST(ApiSimulationSysmemManager, BasicIOSingleChannel) {
     }
 }
 
+TEST(ApiSimulationSysmemManager, GalaxyMappedArenasAreBoundedAndPerChip) {
```
pjanevskiTT, line 66: Every assertion in this loop is chip-independent: `get_mapped_arena_offset()` is `system_memory_size_`, `get_mapped_arena_size()` is a constant minus it, and `get_pcie_base()` comes from the arch, not `chip_id`. The test would pass unchanged if `host_base_` were hard-coded to 0 for every chip -- so it doesn't check the one per-chip property in its name, `get_host_base() == chip_id * PER_CHIP_HOST_STRIDE`. It also never allocates past the window, so the new exhaustion `UMD_ASSERT` (the actual bound this PR adds) has no coverage, and never checks that a second buffer gets a distinct page-aligned IOVA. Incidentally it mmaps 1 GiB per iteration, 8 GiB in total, for arithmetic that needs none of it.

device/api/umd/device/cluster.hpp
```diff
@@ -173,6 +173,21 @@ class Cluster {
     // Explicit to prevent implicit conversion from ClusterOptions at call sites.
     explicit Cluster(ClusterOptions options);
 
+    // Build ClusterOptions inside libtt-umd to avoid passing STL-bearing options across a DSO boundary.
+    static std::unique_ptr<Cluster> create_silicon_cluster(
```
pjanevskiTT: The stated rationale doesn't hold for this one: `std::optional<uint32_t>` is instantiated in the caller's TU and passed by value across the same DSO boundary -- the identical hazard to `ClusterOptions::num_host_mem_ch_per_mmio_device`, which is also a `std::optional<uint32_t>`. A plain `int32_t` (negative = auto-detect) or two overloads would actually avoid it.

Separately, `create_mock_cluster`/`create_swemule_cluster` look unrelated to CRAQ, and with the default `nullptr` descriptor they set no `target_devices` -- so `ClusterDescriptor::create_mock_cluster({}, ...)` builds a descriptor with zero chips and the caller gets a valid-looking `Cluster` whose `chips_` is empty, failing far from the cause. Could those two be split out, or at least take the device list?

PR #3387, opened 2026-09-04
[Base API] Drop the raw APB and iATU surfaces from TTDevice and the model

nbuncicTT:
### Problem description
The chain below this PR moved ARC APB access and iATU programming behind transitional `TTDeviceModel` virtuals — a deviation from the spec, which has neither on `TTDeviceModel`, `TTDevice`, nor `DeviceFirmware`. Raw register-window access shouldn't be surfaced to clients at all.

### What's changed
- **`TTDeviceModel` returns to the spec shape**: the transitional `read_from_arc_apb` / `write_to_arc_apb` / `configure_iatu_region` virtuals are deleted (with their throwing defaults and `tt_device_model.cpp`).
- **`TTDevice` loses `read_from_arc_apb` / `write_to_arc_apb` / `configure_iatu_region`** — verified unused by every client: tt-metal (grep of tt_metal/, ttnn/, tools/, tests/), nanobind bindings, UMD tools and tests. CSM needed nothing: no public CSM API exists anymore.
- **SPI devices** are architecture-committed by construction, so each holds its concrete firmware (`WormholeDeviceFirmware` / `BlackholeDeviceFirmware`, resolved and checked once in the constructor) and uses its ARC APB window directly, routing on the thread-selected NOC exactly as before.
- **`get_refclk_counter`** (spec TTDevice API, kept) moves its per-arch body onto `DeviceFirmware` as `get_refclk_counter(NocId)` — the counter lives in the ARC reset unit, reached through the firmware component's windows, so this is the one polymorphic hop it needs. Body moved verbatim, including the pre-existing quirk that `high2` is never read back (flagged, not fixed — fixing is a behavior change). **Proposed spec amendment**: add `get_refclk_counter` to DeviceFirmware's interface (it currently appears only on TTDevice).
- **iATU** stays as concrete methods on the concrete models: the spec home (`SystemMemoryAllocator` / `SysmemManager`) exists, but nothing calls `configure_iatu_region` today — KMD ≥ 2.0.0 maps system memory through the driver, and the pre-2.0.0 "program the iATU from the device IO address" step has no in-tree consumer. Parked on the models until that path is wired into `SiliconSysmemManager` or dropped. TTSim's simulated-iATU method becomes a private detail of that backend (it only ever called it itself).

### Behavior notes
- Simulation `get_refclk_counter` error text: "ARC APB access is not supported…" → "The refclk counter is not available for a simulated device." (same `RuntimeError`).

### Verification
- Builds with simulation ON and OFF; baremetal tests 254/254.

Issue: #2872


nbuncicTT:
Folded into the restructured chain per review direction: no transitional TTDeviceModel virtuals — #3380 now does the APB deletion + concrete-component migration in one diff, and #3381 the same for iATU. The chain's end tree is byte-identical to this PR's tip (modulo one comment fix in wormhole_arc_window.hpp).

PR #3216, opened 2026-08-21
Hardcode SMBus telemetry address

aleksamarkovicTT:
### Issue
#3217

### Description
Hardcode SMBus telemetry address. The address has not changed in 2 years and SMBus telemetry is not under active development. Removes SmBusTelemetryReader dependency on ArcMessenger.

### Testing
CI. Tested on 18.3 n300

### API Changes
This PR has no API changes.

PR #3388, opened 2026-09-04
Wormhole: ask the firmware where its legacy telemetry block is

### LINKED ISSUE #3217: SMBus telemetry address is hardcoded
This is a temporary solution to remove a circular dependency between FirmwareTelemetryReader and DeviceFirmware. The address used to be obtained through an Arc message, but it is now hardcode, however very unlikely to change. Figure out a long term solution.

pjanevskiTT:
### Issue

#3353
Closes #3217 

### Description

#3216 replaced the Wormhole SMBus telemetry base address with a literal, on the grounds that it had not changed in two years. Management firmware does publish it, over ARC message `GET_SMBUS_TELEMETRY_ADDR`, and a device that puts the block elsewhere is now unreadable rather than merely degraded: reading that block is how the firmware version is established, and the version is what selects the telemetry reader, so a wrong address stops the device coming up. A TTSim-simulated Wormhole is such a device, which is what makes this the base of the #3353 stack.

This asks the firmware instead of assuming. #3216's other purpose is preserved -- `SmBusArcTelemetryReader` gains no `ArcMessenger` dependency, since the address arrives as a constructor argument and the asking is done by the class that already owns the ARC message exchange. On silicon the computed address is the literal it replaces, that literal being what the message returned in the first place.

### List of the changes

- `WormholeDeviceFirmware::init_firmware()` queries `GET_SMBUS_TELEMETRY_ADDR` and passes the result to the telemetry reader and the firmware info provider
- `send_device_command()` drops its firmware-up check, since this message is sent before the firmware is declared up and a Wormhole ARC message rides scratch registers readable from reset. The one caller that wants the check takes it itself, in `TTDevice::get_clock()`'s Wormhole branch
- The address becomes a constructor argument on `SmBusArcTelemetryReader`, defaulting to `DEFAULT_TELEMETRY_NOC_ADDR` (the previous literal), threaded through `create_arc_telemetry_reader()` and `create_firmware_info_provider()` as an optional

### Testing

CI

### API Changes

`SmBusArcTelemetryReader`'s constructor, `ArcTelemetryReader::create_arc_telemetry_reader()` and `FirmwareInfoProviderImplementation::create_firmware_info_provider()` each take one new defaulted parameter; `SmBusArcTelemetryReader::DEFAULT_TELEMETRY_NOC_ADDR` is new. Existing call sites are unaffected.

`DeviceFirmware::send_device_command()` no longer throws `UninitializedDeviceError` on Wormhole before `init_firmware()` has run, and its documented contract narrows to match.


device/tt_device/firmware/wormhole_device_firmware.cpp
aleksamarkovicTT: I'm thinking we could just remove the firmware_info_provider_ requirement here and there won't be a need to define send_arc_message. The comment says it's an artificial constraint and I'm okay with these loosening up as a benefit of the modularity of the new components.

aleksamarkovicTT (approved):

device/tt_device/firmware/wormhole_device_firmware.cpp
```diff
@@ -355,6 +372,23 @@ DeviceCommandResult WormholeDeviceFirmware::send_device_command(
     return DeviceCommandResult{exit_code, std::move(return_values)};
 }
 
+uint64_t WormholeDeviceFirmware::get_legacy_telemetry_noc_addr(NocId noc_id) {
```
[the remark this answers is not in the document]
pjanevskiTT: `get_legacy_telemetry_noc_addr()` is called unconditionally from `init_firmware()` (`wormhole_device_firmware.cpp:109`), so every Wormhole device that comes up runs this path against real firmware. It is exercised on every cluster bringup in hardware CI rather than being uncovered, and against firmware that actually answers the message — which a mock cannot do for the case that matters, a firmware whose answer relocates the block.

Leaving it as is.

device/tt_device/firmware/wormhole_device_firmware.cpp
pjanevskiTT: Done in 7167fd49 — guard removed, and `send_arc_message` folded back into `send_device_command`, so `get_legacy_telemetry_noc_addr()` calls the public entry point directly.

One thing worth knowing, since it is not purely a deletion. `WormholeTTDevice::get_clock()` sends GET_AICLK straight through `send_device_command()`, and that guard was what made it throw before init — `ApiTTDeviceTest.UninitializedError` asserts exactly that at `test_tt_device.cpp:356`. Removing the check outright would have had `get_clock()` put a message to firmware that has not reported ready, and failed that test on Wormhole silicon.

So the refusal moved to `get_clock()` rather than disappearing: same condition as before (the info provider it checks is the member the old guard read), same error. The primitive is unguarded as you asked; the one caller that actually wanted the constraint now carries it.

Also updated `DeviceFirmware`'s `@throws` on `send_device_command`, which promised the pre-init throw unconditionally. Blackhole still refuses — its message queue is built from a descriptor the firmware publishes during boot — so the doc now states that rule instead of promising it for every implementation.

PR #3390, opened 2026-09-04
Make the tests batching-safe and enable unity builds on the merge queue

nbuncicTT:
## Issue

Follow-up to #3321, which added `TT_UMD_ENABLE_UNITY_BUILD` but left it off. This makes the tests batching-safe and turns it on where it pays.

## Description

Batching turns what used to be per-file names into per-batch names, which needs two kinds of fix. Both are confined to tests — **libtt-umd needs no source changes at all.**

**Five colliding file-scope constants.** Worth recording for whoever hits this next: an anonymous namespace does *not* fix these, because every `namespace {}` block in a merged translation unit denotes the same namespace. `APB_OFFSET` was already inside one and collided anyway. `CHIP_ID` was defined identically in four microbenchmark sources, so it moves to the header they all already include; the other four are renamed on one side.

**Three gtest fixtures tripping gcc's `-Werror=subobject-linkage`.** Those files closed their anonymous namespace *before* the `TEST_F` blocks, so the generated `Fixture_Name_Test` classes had external linkage over an internal-linkage base. GCC applies that check to types arriving through an include, which is exactly what a unity TU does — so it is invisible in a normal build, and invisible to clang in either mode. Moving the closing brace past the tests is also the layout gtest documents. This is why the first CI run failed 3 of 11 entries, all gcc.

**Where it is turned on: `merge_group` only, because that is where builds are effectively cold.** `Build All` on the release-flags ubuntu-24.04 clang-20 entry:

| | Build All |
| --- | --- |
| merge queue today (sccache enabled, READ_ONLY off main) | 19m14s |
| no cache at all | 19m41s |
| push to main (sccache READ_WRITE) | 12m57s |

The merge queue reads main's cache read-only and gets almost nothing from it. Pushes to main write the cache and do get real value, so they stay per-file, as do the manylinux wheel jobs that package what we ship. The emulated riscv build is included since it is build-only and uses no cache at all.

**Measured effect**, cold dispatch of the full matrix, `Build All` ([baseline](https://github.com/tenstorrent/tt-umd/actions/runs/33445523123), [unity](https://github.com/tenstorrent/tt-umd/actions/runs/33448157918)):

| matrix entry | baseline | unity | |
| --- | --- | --- | --- |
| ubuntu-24.04 clang-20 (release flags) | 19m41s | 5m16s | 3.74x |
| ubuntu-24.04 gcc-11 | 22m00s | 6m33s | 3.36x |
| ubuntu-24.04-riscv gcc (emulated) | 7m57s | 2m48s | 2.84x |
| fedora-39 clang-17 (release flags) | 9m03s | 3m33s | 2.55x |
| ubuntu-24.04-arm clang-20 | 12m20s | 4m56s | 2.50x |
| ubuntu-22.04 clang-13 | 21m27s | 8m50s | 2.43x |
| ubuntu-22.04 clang-20 | 21m11s | 8m50s | 2.40x |
| ubuntu-24.04 clang-20 | 18m21s | 7m58s | 2.30x |
| fedora-39 clang-17 | 16m57s | 8m02s | 2.11x |
| ubuntu-22.04 gcc-11 | 18m42s | 8m59s | 2.08x |
| fedora-39 gcc-13 | 11m47s | 8m42s | 1.35x |

Whole-job totals went from 26-33 min to 8-15 min.

## List of the changes

- `CHIP_ID` moved to `microbenchmark_utils.hpp`, next to the existing `ONE_KIB`/`ONE_MIB`; the four per-file definitions deleted.
- Renamed one side of four colliding pairs: `example_eth_harvesting_mask` -> `example_eth_harvest_mask` (test_soc_descriptor.cpp), `HUGEPAGE_REGION_SIZE` -> `SIMULATION_HUGEPAGE_REGION_SIZE` (test_simulation_sysmem_manager.cpp), `APB_OFFSET` -> `BLACKHOLE_APB_OFFSET` (test_blackhole_arc_apb.cpp), `DRAM_BARRIER_BASE` -> `CLUSTER_WH_DRAM_BARRIER_BASE` (test_cluster_wh.cpp).
- Closing brace of the anonymous namespace moved past the tests in test_simulation_client.cpp, test_simulation_server_socket.cpp and benchmarks/simulator/test_simulator.cpp.
- `build-device.yml`: `TT_UMD_UNITY` is ON for `merge_group`, passed to `Build All`, `Rebuild package for python without wheel`, and the emulated riscv build. An `enable-unity-build` dispatch input exists only so that configuration can be reproduced or measured from a manual run.

## Testing

All 11 `build-device` entries pass on the cold dispatch linked above, across gcc and clang and on x86, arm and emulated riscv. The three package-install jobs (`Test deb`/`rpm package`) also pass, so packages built from unity TUs install and their example client compiles against the shipped `libtt-umd.so`.

`ninja -k 0` over the whole tree was used to confirm the gcc linkage errors were the only remaining error class rather than the first of several.

`baremetal_tests` from a unity build: 253/254 pass. The one failure, `SigBusMechanismTest.MultiProcessMultiThreadStress`, fails identically on a non-unity build of the same tree, so it is pre-existing and not caused by batching.

## API Changes

This PR has no API changes. It does change how the merge-queue builds are compiled — static-initialisation order within a batch, and inlining across what used to be separate translation units — which is the reason it is scoped to CI rather than turned on for releases.

### DIFF
--- a/.github/workflows/build-device.yml
+++ b/.github/workflows/build-device.yml
@@ -16,6 +16,13 @@ on:
           Enable sccache compiler caching.
         type: boolean
         default: true
+      enable-unity-build:
+        required: false
+        description: >-
+          Batch sources into unity translation units. Always on in the merge queue; this
+          input only exists so a dispatch can reproduce or measure that configuration.
+        type: boolean
+        default: false
   # Runs on all standard code-review and integration events: PR, merge queue, and post-merge on main.
   pull_request:
   merge_group:
@@ -48,6 +55,12 @@ env:
   CREATE_MAP_BINARIES_DIR: ./device/bin/silicon
   INSTALL_OUTPUT_DIR: ./install
   WHEEL_OUTPUT_DIR: ./build/wheel
+  # Batch sources into unity translation units, for the merge queue only.
+  # Merge-queue runs read the cache off main read-only and get almost nothing from it
+  # (measured: 19m14s of Build All against 19m41s with no cache at all), so they are the
+  # builds worth batching. Pushes to main write the cache and do get value from it, so
+  # they stay per-file, as do the wheel jobs that package what we ship.
+  TT_UMD_UNITY: ${{ (github.event_name == 'merge_group' || inputs.enable-unity-build) && 'ON' || 'OFF' }}
 
 jobs:
   # Skips the heavy sweep on PRs (a build already runs as part of the tests) but
@@ -259,6 +272,7 @@ jobs:
               -DTT_UMD_BUILD_EXAMPLES=OFF \
               -DTT_UMD_ENABLE_CLANG_TIDY=ON \
               -DTT_UMD_ENABLE_SCCACHE=${{ env.SCCACHE_GHA_ENABLED == 'true' && 'ON' || 'OFF' }} \
+              -DTT_UMD_ENABLE_UNITY_BUILD=${{ env.TT_UMD_UNITY }} \
               -DCMAKE_BUILD_TYPE=Release
           else
             cmake -B ${{ env.BUILD_OUTPUT_DIR }} -G Ninja \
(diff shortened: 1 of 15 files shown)


aleksamarkovicTT (approved):

moklobdzijaTT:
nit: For the name collision within the anonymous namespaces in the test source files, guess we could have went with [`UNITY_BUILD_UNIQUE_ID` property](https://cmake.org/cmake/help/v3.24/prop_tgt/UNITY_BUILD_UNIQUE_ID.html)? For now naming seems fine, but this might help with having the unnecessary verbose identifiers in the test source files.

moklobdzijaTT:
Is the last bullet point in the list of changes up to date? I see that the option is `TT_UMD_UNITY_ENABLED` and that `TT_UMD_ENABLE_UNITY_BUILD` is set for building test.

moklobdzijaTT (approved):

PR #3396, opened 2026-09-04
[Base API] Remove TLBManager, on top of Strict ordering

### LINKED ISSUE #3377: TLBManager removal
**Description**
UMD has an ongoing plan to remove TLBManager, delegating its responsibilities into the Base API device/window classes. This issue tracks completing the remaining removal work.

**Scope**
 - Remove remaining TLBManager usage from tt-metal and any other dependent code
 - Validate the changes build and pass tests
 - Remove TLBManager itself once all call sites are migrated

**Timeline**
No hard deadline.

Not a bug — this is planned cleanup/removal work following the Base API delegation effort.

**Definition of Done**
 - TLBManager fully removed from all call sites
 - Build and tests pass with TLBManager removed

ajovanovicTT:
### Issue
#3377

### Description
Removes TLBManager and the static TLB configuration APIs built on it. A caller that wants a persistent mapping takes an IoWindow from create_io_window instead, which covers what configure_tlb and get_static_tlb_window were used for and does not require the chip to track mappings on the caller's behalf.

LocalChip's block path no longer chooses between a statically mapped window and a reconfigured one, and always uses the cached window. That path carries IoOrdering::Strict on main since #3399, so dropping the static TLB branch does not weaken the ordering any caller was getting: what the static TLBs provided, the remaining path now provides on its own.

Paired with the tt-metal migration onto create_io_window, which is what removes the last in tree caller of these APIs: https://github.com/tenstorrent/tt-metal/pull/55531

Stacked on #3454, which redesigns the TLB microbenchmark onto IoWindow ordering modes. That was split out so the measurement change is reviewed separately and so the benchmark workflow fires on the PR that changes it. This PR therefore does not touch tests/microbenchmark.

Landing order is tt-metal #55531, then #3454, then this PR.

### List of the changes
 - Deleted TLBManager (chip_helpers/tlb_manager.hpp/cpp) and the Cluster and Chip methods built on it: configure_tlb, get_static_tlb_window, get_tlb_configuration, get_tlb_manager.
 - Removed TLBManager ownership from LocalChip and SimulationChip.
 - LocalChip::write_to_device and read_from_device lose the static TLB branch and always use the cached window.
 - Removed the tests that only covered the removed APIs: test_tlb_manager.cpp and a disabled manual TLB configuration test.
 - Removed static TLB setup from the blackhole and wormhole cluster tests, which now exercise the reconfigured window path.
 - Replaced TLBStaticTensix with IoWindowTensixRoundTrip, which drives an owned window instead of the removed static TLB lookup.

### Testing
CI.

### API Changes
This PR has API changes. The static TLB configuration surface is removed; create_io_window replaces it.


device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -468,7 +466,7 @@ class TTDevice {
     virtual SimulationSysmemManager *get_sysmem_manager() { return nullptr; }
 
     /**
-     * Allocate a TlbWindow for use by callers (typically TLBManager).
+     * Allocate a TlbWindow for use by callers.
```
pjanevskiTT, line 469: Deleting `TLBManager` also deletes the only place that wired a hang check onto a caller-owned window: `TLBManager::allocate_tlb_window` did `tlb_window->set_io_timeout_hang_check([tt_device](NocId noc){ return tt_device->is_noc_hung(noc, HangAction::RETURN); })`, with a comment saying these windows "carry the same per-op MMIO timeout hang check as the cached windows".

`TTDevice::create_io_window`, which this PR points clients at, wires nothing, so `SiliconTlbWindow::update_io_timeout_callback()` installs `[]{ return true; }` — every overrun treated as a false alarm. A window from the new `Cluster::create_io_window` that hits a genuinely hung NOC therefore never raises `DeviceTimeoutError`; the read returns `HANG_READ_VALUE` and the caller consumes it. `set_io_timeout_hang_check` goes from 4 call sites to 3 here, leaving only LocalChip's cached wc/uc windows and `PcieProtocol`. Wiring it in `create_io_window` alongside the `configure()` call would restore it.

(Note: ideally on the deleted `chip_helpers/tlb_manager.cpp` hunk; placed here as the deleted side can't take a comment.)

tests/microbenchmark/benchmarks/tlb/test_tlb.cpp
```diff
@@ -135,18 +150,25 @@ TEST(MicrobenchmarkTLB, Ethernet) {
             cluster->read_from_device(pattern.data(), CHIP_ID, eth_core, ADDRESS, batch_size);
         });
     }
-    // Static TLB configuration.
-    cluster->configure_tlb(CHIP_ID, eth_core, 2 * ONE_MIB, ADDRESS, tlb_data::Relaxed);
+    // IoWindow: mapped once, driven directly, no per-op reconfigure.
+    std::unique_ptr<IoWindow> window =
+        cluster->create_io_window(CHIP_ID, eth_core, ADDRESS, {.size = 2 * ONE_MIB}, IoOrdering::Relaxed);
```
pjanevskiTT, line 155: `{.size = 2 * ONE_MIB}` at `ADDRESS = 0x20000` can't be served by a 2 MB class, so this doesn't replace the 2 MB TLB that `configure_tlb(..., 2 * ONE_MIB, ADDRESS, ...)` asked for. `TTDevice::create_io_window` picks the first class satisfying `size <= class.size - (addr % class.size)`: for 2 MiB at 128 KiB, WH's 1 MB class gives 896 KiB and its 2 MB class gives 1920 KiB, so it lands on the 16 MB class; on BH the 2 MB class fails the same way and the only class left is the 4 GB BAR4 one, which mmaps 4 GB and burns one of eight — and on a host without a large or resizable BAR4, `allocate_tlb` throws and `MicrobenchmarkTLB.Ethernet` dies instead of running. The largest batch here is 128 KiB, so `.size = 128 * ONE_KIB` is what's actually needed.

tests/microbenchmark/benchmarks/tlb/test_tlb.cpp
```diff
@@ -58,18 +59,25 @@ TEST(MicrobenchmarkTLB, DRAM) {
             cluster->read_from_device(pattern.data(), CHIP_ID, dram_core, ADDRESS, batch_size);
         });
     }
-    // Static TLB configuration.
-    cluster->configure_tlb(CHIP_ID, dram_core, 2 * ONE_MIB, ADDRESS, tlb_data::Relaxed);
+    // IoWindow: mapped once, driven directly, no per-op reconfigure.
+    std::unique_ptr<IoWindow> window =
+        cluster->create_io_window(CHIP_ID, dram_core, ADDRESS, {.size = 2 * ONE_MIB}, IoOrdering::Relaxed);
     for (size_t batch_size : BATCH_SIZES) {
+        if (batch_size > window->get_size()) {
```
pjanevskiTT, line 66: For DRAM (`ADDRESS = 0x0`, batches up to 32 MiB against a 2 MiB window) this guard silently drops the 4/8/16/32 MiB rows, so the exported table has `Dynamic TLB, write, 33554432 bytes` with no counterpart and nothing saying why — a reader concludes the IoWindow path regressed at large sizes. `README.md:41-68` also still publishes `Static TLB, write/read, N bytes` rows for exactly those sizes, so the checked-in baseline now documents output the suite can't produce.

Either chunk across the window the way `write_block_reconfigure` does, or encode the skip in the row name; the README needs the rename either way. Worth noting there too that the `IoWindow` rows call `window->write_block` directly while the `Dynamic TLB` rows go through `Cluster`, so the remaining delta includes coordinate translation and the lock, not just the mapping strategy.

device/chip/local_chip.cpp
```diff
@@ -257,23 +244,12 @@ void LocalChip::write_to_device(CoreCoord core, const void* src, uint64_t l1_des
...
-            translated_core,
-            l1_dest,
-            size,
-            get_selected_noc_id(),
-            IoOrdering::Strict);
-    }
+    // Strict: the write is non-posted, so it has landed on the target once this call returns. Callers
+    // that pair a bulk write with a subsequent trigger rely on that completion, and this path
+    // reconfigures the window between transfers, so nothing else orders them.
+    std::lock_guard<std::mutex> lock(wc_tlb_lock);
```
pjanevskiTT, line 250: With the mapped branch gone, every bulk PCIe access on a chip takes this one `wc_tlb_lock` and reconfigures the one `cached_wc_tlb_window`, so two threads writing to two different cores now serialize completely — reconfigure ioctl included, since it's inside the critical section. Previously the `is_tlb_mapped` branch took no lock and used a per-core window.

The tests that covered that are losing their setup in the same diff: `ClusterBH`/`ClusterWH` `MultiThreadedMemBar` had their `configure_tlb` loops deleted and the comment "All (fairly large) transactions go through a static TLB" removed, so their two threads contend on the mutex instead of driving concurrent mapped-window traffic. `set_membar_flag` in this file feels it worst — its unbounded readback poll is a 4-byte `read_from_device` per core per iteration, each now a lock acquisition plus a configure-TLB ioctl. Is a fast path out of scope for this PR, or is the plan to live with it?

device/api/umd/device/cluster.hpp
```diff
@@ -246,41 +244,6 @@ class Cluster {
...
-     */
-    void configure_tlb(
-        ChipId logical_device_id,
-        CoreCoord core,
-        size_t tlb_size,
-        uint64_t address,
-        uint64_t ordering = tlb_data::Relaxed);
-
     /**
      * Maps a core into host address space, anchored at an address on that core. Reads and writes
```
pjanevskiTT, line 248: This drops `configure_tlb` (both overloads) and, further down, `get_static_tlb_window`, `get_tlb_configuration` and `get_tlb_manager`, plus `api/umd/device/chip_helpers/tlb_manager.hpp` from the installed FILE_SET — all in one PR. The convention for public API removals is two stacked PRs: keep the old entry point as a `[[deprecated("...")]]` forwarder first, delete it once downstream has uplifted. Downstream vendors UMD and uplifts on its own schedule, so a consumer that hasn't landed the paired tt-metal change — or isn't in that pairing at all, like tt-exalens — fails to compile at uplift with no warning window. Is the paired metal PR considered enough coverage, or should the deletion be split out?

tests/wormhole/test_cluster_wh.cpp
```diff
@@ -1087,10 +1013,10 @@ TEST(ClusterWH, LargeAddressTlb) {
     // Read the scratch register via BAR0:
     value0 = cluster.get_chip(0)->get_tt_device()->bar_read32(0x1ff30060);
 
-    // Read the scratch register via the TLB:
+    // Read the scratch register via a dynamically mapped TLB:
     cluster.read_from_device(&value1, 0, ARC_CORE, addr, sizeof(uint32_t));
 
-    // Read the scratch register via a different TLB, different code path:
+    // Read again, to confirm the mapping is stable across ops:
```
pjanevskiTT, line 1019: With `configure_tlb` gone, `value1` and `value2` are two identical `cluster.read_from_device` calls, so `EXPECT_EQ(value2, value0)` can't fail unless `EXPECT_EQ(value1, value0)` already did, and the new comment describes a no-op. The test existed to agree three ways on the ARC scratch register — BAR0, a statically mapped window at the 64-bit NOC address `0x880030000`, and a dynamic one — so the >32-bit `local_offset` path through a caller-owned window is exactly what it stopped covering. `cluster.create_io_window(0, ARC_CORE, arc_reset_noc, {.size = ...})` + `window->read32(scratch_offset)` for the second read restores the two-path check.

device/chip/local_chip.cpp
```diff
@@ -257,23 +244,12 @@ void LocalChip::write_to_device(CoreCoord core, const void* src, uint64_t l1_des
...
-        write_block_reconfigure(
-            *get_cached_wc_tlb_window(),
-            src,
-            translated_core,
-            l1_dest,
-            size,
-            get_selected_noc_id(),
-            IoOrdering::Strict);
-    }
+    // Strict: the write is non-posted, so it has landed on the target once this call returns. Callers
```
pjanevskiTT, line 247: comment is not completely true, strict is not a guarantee that write has landed. It is just not going to order reads/writes on the host side

pjanevskiTT: you can examine https://github.com/tenstorrent/tt-isa-documentation/blob/main/BlackholeA0/PCIExpressTile/README.md in order to get a correct comment for this part of the code

pjanevskiTT (approved):

device/api/umd/device/cluster.hpp
ajovanovicTT: This is going to land to main after changes that remove all tlb manager references to metal land, so no problem

device/api/umd/device/tt_device/tt_device.hpp
ajovanovicTT: Good catch, restored. TTDevice::create_io_window now installs the check next
to the configure() call, so windows it hands out carry the same per-op timeout
hang check as the cached ones. Pulled the lambda wire_hang_detector already
used into a private make_io_timeout_hang_check() so the two paths share it.

tests/microbenchmark/benchmarks/tlb/test_tlb.cpp
ajovanovicTT: Those TLB microbenchmarks will be rewritten with underlying PR in stack, since maintaining those with Static TLB path does not make sense anymore.

ajovanovicTT: https://github.com/tenstorrent/tt-umd/pull/3454

device/chip/local_chip.cpp
ajovanovicTT: Measured on p100, MultiThreadedMemBar, 5 repeats: 2385 ms before the removal vs
6136 ms now, 2.6x. DynamicTLB_RW is flat, so it is the removal.

It is the bulk path, not the membar - that test moves 30 KiB per core per
iteration against 4 small barrier accesses. The 2.6x is two threads no longer
overlapping their reads. That is the shape of this path now: one mapping per
chip, Strict and ordered across calls, and serialized as a consequence. A
caller that wants concurrency takes an IoWindow from create_io_window. Both doc
comments now say so.

So no fast path. The concurrency coverage the tests lost belongs in an IoWindow
test since test_tlb.cpp is single-threaded, and I will add that.

PR #3399, opened 2026-09-07
[Base API] Add IoOrdering param and use Strict on LocalChip's fallback

ajovanovicTT:
### Issue
#3377

### Description
Adds an ordering parameter to the IoWindow creation APIs ahead of TLBManager removal, and flips LocalChip's block-path fallback from Relaxed to Strict.

LocalChip::write_to_device/read_from_device fall back to a reconfigured window whenever the target isn't covered by a static TLB, and that fallback is Relaxed, under which writes can reorder or interleave with each other. Metal's static TLBs cover Tensix and ETH on Blackhole MMIO chips today, so those writes run Strict. Deleting them (the next PR in this stack, and metal's PR) would drop them to Relaxed, so this PR flips the fallback first and the ordering holds throughout the stack.

Callers that want Relaxed should take an IoWindow from create_io_window with IoOrdering::Relaxed requested explicitly.

Paired with the Strict variant of the metal migration PR, which needs no ordering fixes as a result: https://github.com/tenstorrent/tt-metal/pull/55531

Cost, measured on silicon: 11x to 16x on the bulk write path on wormhole_b0, 28x to 30x on blackhole. Reads are unaffected, so the read_from_device flip is symmetry only.

### List of the changes
 - Added IoOrdering ordering = IoOrdering::Strict to TTDevice::create_io_window and Cluster::create_io_window, plumbed through to TlbWindow::configure. Default preserves current behavior for all existing callers.
 - LocalChip::write_to_device/read_from_device's reconfigure fallback now requests IoOrdering::Strict instead of IoOrdering::Relaxed.
 - Documented the ordering Cluster::write_to_device/read_from_device now give, and pointed callers needing Relaxed at create_io_window.
 - Added Relaxed/Posted round-trip coverage in tests/unified/test_tlb.cpp.

### Testing
CI green, including a benchmark run on all three arches: https://github.com/tenstorrent/tt-umd/actions/runs/34597565499
Regression check against the n150 baseline passes; ClusterConstructor is unaffected. The microbenchmark TLB cases now measure the Strict path rather than the fastest bulk path and need rewriting onto an IoWindow with Relaxed requested explicitly, tracked separately.

### API Changes
This PR has API changes, widening base api spec with IoOrdering param defaulted to Strict.

### DIFF
--- a/device/api/umd/device/cluster.hpp
+++ b/device/api/umd/device/cluster.hpp
@@ -297,6 +297,7 @@ public:
      * @param core_start Core to map, or upper-left corner of a multicast grid.
      * @param addr Address on the core(s) the window is anchored at.
      * @param host Host-side window properties (caching strategy and requested size).
+     * @param ordering Transaction ordering mode to apply to the mapping.
      * @param core_end Lower-right corner of a multicast grid, or nullopt for unicast.
      * @param flags Transaction attributes.
      * @param noc Routing selection, or nullopt to route over the NOC selected for this thread.
@@ -306,6 +307,7 @@ public:
         CoreCoord core_start,
         uint64_t addr,
         HostIoWindowConfig host = {},
+        IoOrdering ordering = IoOrdering::Strict,
         std::optional<CoreCoord> core_end = std::nullopt,
         WindowFlags flags = WindowFlags::None,
         std::optional<NocId> noc = std::nullopt);
@@ -416,6 +418,10 @@ public:
      * This API is used for writing to both TENSIX and DRAM cores. The internal SocDescriptor can be used to determine
      * which type of the core is being targeted.
      *
+     * This is a bulk-transfer path with no ordering guarantee: writes from successive calls may
+     * complete in any order. If the order of two writes matters, separate them with a barrier or use
+     * an @ref IoWindow from @ref create_io_window, which takes an ordering mode.
+     *
      * @param mem_ptr Source data address.
      * @param size_in_bytes Source data size.
      * @param chip Chip to target.
@@ -428,6 +434,8 @@ public:
      * Read uint32_t data from a specified device, core and address to host memory (defined for Silicon).
      * This API is used for reading from both TENSIX and DRAM cores. The internal SocDescriptor can be used to determine
      * which type of the core is being targeted.
+     *
+     * No ordering guarantee, same as @ref write_to_device, including against writes issued through it.
      *
      * @param mem_ptr Data pointer to read the data into.
      * @param chip Chip to target.
--- a/device/api/umd/device/cluster.hpp
+++ b/device/api/umd/device/cluster.hpp
@@ -418,10 +418,12 @@ public:
      * This API is used for writing to both TENSIX and DRAM cores. The internal SocDescriptor can be used to determine
      * which type of the core is being targeted.
      *
-     * Transfers use @ref IoOrdering::Strict, so the data has landed on the target before the call
-     * returns and successive calls are ordered with respect to each other. This costs write
-     * throughput; a caller that does not need the guarantee can take an @ref IoWindow from
-     * @ref create_io_window with a weaker ordering mode and drive it directly.
+     * Transfers use @ref IoOrdering::Strict, so successive calls are ordered with respect to each
+     * other. The call returns once the writes are issued, not once they are acknowledged by the
+     * target — the underlying MMIO stores are posted. This costs write throughput; a caller that
+     * does not need the ordering guarantee can take an @ref IoWindow from @ref create_io_window
+     * with a weaker ordering mode and drive it directly, using @ref IoWindow::configure to advance
+     * across chunks larger than the window.
      *
      * @param mem_ptr Source data address.
      * @param size_in_bytes Source data size.
@@ -436,8 +438,8 @@ public:
      * This API is used for reading from both TENSIX and DRAM cores. The internal SocDescriptor can be used to determine
      * which type of the core is being targeted.
      *
-     * Uses @ref IoOrdering::Strict, same as @ref write_to_device, and is ordered against writes
-     * issued through it.
+     * Uses @ref IoOrdering::Strict, so successive calls through this function are ordered with
+     * respect to each other.
      *
      * @param mem_ptr Data pointer to read the data into.
      * @param chip Chip to target.
--- a/device/api/umd/device/cluster.hpp
+++ b/device/api/umd/device/cluster.hpp
@@ -418,9 +418,10 @@ public:
      * This API is used for writing to both TENSIX and DRAM cores. The internal SocDescriptor can be used to determine
      * which type of the core is being targeted.
      *
-     * This is a bulk-transfer path with no ordering guarantee: writes from successive calls may
-     * complete in any order. If the order of two writes matters, separate them with a barrier or use
-     * an @ref IoWindow from @ref create_io_window, which takes an ordering mode.
+     * Transfers use @ref IoOrdering::Strict, so the data has landed on the target before the call
+     * returns and successive calls are ordered with respect to each other. This costs write
+     * throughput; a caller that does not need the guarantee can take an @ref IoWindow from
+     * @ref create_io_window with a weaker ordering mode and drive it directly.
      *
      * @param mem_ptr Source data address.
      * @param size_in_bytes Source data size.
@@ -435,7 +436,8 @@ public:
      * This API is used for reading from both TENSIX and DRAM cores. The internal SocDescriptor can be used to determine
      * which type of the core is being targeted.
      *
-     * No ordering guarantee, same as @ref write_to_device, including against writes issued through it.
+     * Uses @ref IoOrdering::Strict, same as @ref write_to_device, and is ordered against writes
+     * issued through it.
      *
      * @param mem_ptr Data pointer to read the data into.
      * @param chip Chip to target.
--- a/device/api/umd/device/tt_device/tt_device.hpp
+++ b/device/api/umd/device/tt_device/tt_device.hpp
@@ -498,9 +498,11 @@ public:
      * See @ref TargetIoWindowConfig.
      * @param host Host-side properties (caching strategy and requested size).
      * See @ref HostIoWindowConfig.
+     * @param ordering Transaction ordering mode to apply to the mapping.
      * @return An exclusively owned handle to the newly created @ref IoWindow.
      */
-    std::unique_ptr<IoWindow> create_io_window(TargetIoWindowConfig target, HostIoWindowConfig host);
+    std::unique_ptr<IoWindow> create_io_window(
+        TargetIoWindowConfig target, HostIoWindowConfig host, IoOrdering ordering = IoOrdering::Strict);
 
     /**
      * Export a NOC-addressable region as a dma-buf file descriptor for peer-to-peer PCIe DMA.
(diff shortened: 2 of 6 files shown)


device/api/umd/device/cluster.hpp
```diff
@@ -429,6 +436,9 @@ class Cluster {
      * This API is used for reading from both TENSIX and DRAM cores. The internal SocDescriptor can be used to determine
      * which type of the core is being targeted.
      *
+     * Uses @ref IoOrdering::Strict, same as @ref write_to_device, and is ordered against writes
```
pjanevskiTT: `Strict` can't provide this one. `read_block_reconfigure` tags the mapping `WindowFlags::UnicastRead` and `write_block_reconfigure` tags it `WindowFlags::UnicastWrite` (`io_window_reconfigure.cpp:40,57`), and that separation is deliberately what puts reads and writes on different interconnect resources: on BH/Quasar `ArchitectureTlbs::get_static_vc` returns `static_vc_buddy = 1` for reads and `0` for writes (`architecture_tlbs.cpp:50-56`), and on WH the comment on `static_vc_is_configurable = false` says the tile hardwires the same split.

`Strict` orders transactions within a channel, so two channels give no mutual ordering — a caller that writes and then reads back the same address on the strength of this sentence can legitimately observe the pre-write value. Suggest dropping the cross-ordering claim and keeping only that reads are ordered among themselves.

device/api/umd/device/cluster.hpp
```diff
@@ -306,6 +307,7 @@ class Cluster {
         CoreCoord core_start,
         uint64_t addr,
         HostIoWindowConfig host = {},
+        IoOrdering ordering = IoOrdering::Strict,
```
pjanevskiTT, line 310: Inserting `ordering` before `core_end`/`flags` is a source break for anyone passing those positionally, so "Default preserves current behavior for all existing callers" isn't quite right — the diff's own edit to `CreateMulticastIoWindow` (`test_tlb.cpp:699-707`, which had to grow `IoOrdering::Strict` before `grid_end`) is the counter-example.

Ordering is a per-mapping attribute exactly like `flags`, which already lives in `TargetIoWindowConfig`. Adding `IoOrdering ordering = IoOrdering::Strict;` next to it there instead would remove this parameter, the mid-list default, and the new `TTDevice::create_io_window` overload in one go — and would let `get_target_config()` round-trip the whole mapping, which it currently can't since it reports neither `flags` nor `ordering`.

device/api/umd/device/cluster.hpp
```diff
@@ -416,6 +418,11 @@ class Cluster {
      * This API is used for writing to both TENSIX and DRAM cores. The internal SocDescriptor can be used to determine
      * which type of the core is being targeted.
      *
+     * Transfers use @ref IoOrdering::Strict, so the data has landed on the target before the call
+     * returns and successive calls are ordered with respect to each other. This costs write
+     * throughput; a caller that does not need the guarantee can take an @ref IoWindow from
```
pjanevskiTT: The escape hatch doesn't reach the transfers it's apologising for. `IoWindow::write_block` validates `offset + size` against `get_size()` and `TlbWindow::validate` throws `std::out_of_range` past it (`tlb_window.cpp:70-74`), so a caller moving more than one size class (2 MB on BH, 1 MB on WH) has to reimplement the reconfigure-and-chunk loop — and UMD's own implementation of it, `write_block_reconfigure`, lives in `device/pcie/io_window_reconfigure.hpp`, which isn't under `device/api/`. Either bound the claim to a single window, or export the chunking helper.

device/chip/local_chip.cpp
```diff
@@ -269,7 +272,7 @@ void LocalChip::write_to_device(CoreCoord core, const void* src, uint64_t l1_des
             l1_dest,
             size,
             get_selected_noc_id(),
-            IoOrdering::Relaxed);
+            IoOrdering::Strict);
```
pjanevskiTT, line 275: The flip is unconditional, but this else branch is taken for a lot more than the Tensix/ETH targets that lose their static window: DRAM on Wormhole, ARC/PCIe/router cores, and — for every client that never calls `configure_tlb` (UMD's own tests and tools, tt-exalens) — every access on every architecture. None of those had a static-TLB guarantee to hold continuously; they just go from 14.35 to 9.50 GB/s.

Two things make the cost hard to size as stated: the quoted numbers are from tt-isa-documentation's WormholeB0 PCIe tile while the motivation is Blackhole (different tile, configurable static VCs, 2 MB/4 GB classes), and the description says this isn't built or run yet. Scoping the Strict request to the targets that actually lose the static window would avoid charging the rest for it.

tests/unified/test_tlb.cpp
```diff
@@ -638,13 +638,23 @@ TEST_F(TestTlb, CreateIoWindow) {
...
+            tt_device->create_io_window(target, {.size = requested_size}, ordering);
+        EXPECT_EQ(ordered_window->get_io_ordering(), ordering);
+        EXPECT_EQ(ordered_window->get_target_config().addr, l1_addr) << "Ordering should not disturb the target";
+    }
+
     // The same factory reached by chip and CoreCoord, which is how clients that hold neither a
     // TTDevice nor translated coordinates ask for a window.
     std::unique_ptr<IoWindow> chip_window =
-        cluster->create_io_window(chip, tensix_core, l1_addr, {.size = requested_size});
+        cluster->create_io_window(chip, tensix_core, l1_addr, {.size = requested_size}, IoOrdering::Relaxed);
```
pjanevskiTT, line 653: `chip_window` is now `Relaxed` — "No ordering guarantees between transactions" (`io_window_config.hpp:33`) — but still carries the `write32`/`read32` round-trip at 659-660. On Blackhole that write and read additionally land on different static VC buddies, so nothing requires the read to observe the write. Keeping the round-trip on a Strict window and asserting `Relaxed` on a separate one, as the loop at 643-648 already does, avoids making the test depend on the ordering it just turned off.

pjanevskiTT (approved):

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -502,6 +502,15 @@ class TTDevice {
      */
     std::unique_ptr<IoWindow> create_io_window(TargetIoWindowConfig target, HostIoWindowConfig host);
 
+    /**
+     * Same as the overload above, with the transaction ordering mode applied to the mapping made
+     * explicit instead of defaulting to @ref IoOrdering::Strict.
+     *
+     * @param ordering Transaction ordering mode to apply to the mapping.
+     */
+    std::unique_ptr<IoWindow> create_io_window(
```
pjanevskiTT, line 511: is this in base API spec? for vis @nbuncicTT

device/tt_device/tt_device.cpp
```diff
@@ -368,12 +368,17 @@ std::unique_ptr<TlbWindow> TTDevice::get_io_window(tlb_data config, TlbMapping m
     UMD_THROW(error::RuntimeError, "Failed to allocate TLB window.");
 }
 
+std::unique_ptr<IoWindow> TTDevice::create_io_window(TargetIoWindowConfig target, HostIoWindowConfig host) {
+    return create_io_window(target, host, IoOrdering::Strict);
```
pjanevskiTT, line 372: does it make sense to just have option with ordering in base API? And then we can make default option for ordering to strict?

device/api/umd/device/tt_device/tt_device.hpp
```diff
@@ -502,6 +502,15 @@ class TTDevice {
      */
```
pjanevskiTT, line 502: just curios, can you analyze how this impacts our tests that do a lot of IO? I would measure 6u constructor and IO heavy tests (L1 read/writes) when we go over cluster API

pjanevskiTT, line 502: we also probably want to update the baselines for benchmarks, is this going to break it now that we use strict instead of relaxed?

device/api/umd/device/cluster.hpp
ajovanovicTT: Done.

device/api/umd/device/cluster.hpp
ajovanovicTT: Leaving IoOrdering out of TargetIoWindowConfig for now as it was previosly agreed when discussing on BaseApi.

device/api/umd/device/tt_device/tt_device.hpp
ajovanovicTT: Agreed with @nbuncicTT offline for this additional method with ordering param.

device/tt_device/tt_device.cpp
ajovanovicTT: Agreed with @nbuncicTT to add it as a separate method in order to keep aligned with BaseApi for now.

tests/unified/test_tlb.cpp
ajovanovicTT: On Blackhole, chip_window uses the default WindowFlags::Bidirectional, and get_static_vc (architecture_tlbs.cpp:42-57) keeps bidirectional windows off the static VC entirely — see the comment at architecture_tlbs.cpp:114-116: combining read+write on one static VC crashes the host on BH, so it uses a dynamic VC instead. The buddy split only applies to UnicastRead/UnicastWrite-flagged windows, not this one, and IoOrdering is a separate field from WindowFlags (tlb_window.cpp:53-54) — neither write32_impl nor read32_impl even look at it. So there's no VC-buddy race here.

device/chip/local_chip.cpp
ajovanovicTT: Discussed and agreed offline, even those targets which didnt use static tlb's will be affected.

ajovanovicTT:
Measured the perf impact that was asked about.

Constructor is fine. Two independent trials on the Blackhole 6u galaxy g08blx07 came out flat, +0.23 and +0.00 percent, with the sign flipping between trials, so any real effect is under one percent. n150 agrees.

Bulk IO is the problem. On the LocalChip reconfigure fallback, writes of 1 KiB and above are 11 to 16 times slower on Wormhole and 28 to 30 times slower on Blackhole. Measured on n150 bgd-lab-06, n300 bgd-lab-05 and p150 bh-40. The n150 numbers are from the run-benchmarks CI artifacts, branch against main. On n300 and p150 I ran locally with the same binary in both arms, flipping only the two IoOrdering arguments in local_chip.cpp, so nothing else can be responsible.

DRAM 32 MiB writes go from 2.43 GiB/s to 158 MiB/s on Wormhole and from 6.91 GiB/s to 255 MiB/s on Blackhole. Strict sits at a flat plateau of roughly 0.14 to 0.27 GB/s regardless of transfer size. Blackhole is hit harder because its Relaxed path is faster while the Strict floor barely moves. Worst case I saw was a 1 MiB unicast write to 120 cores going from 247 ms to 7.34 s. The cause is that this path uses a write combining mapping, and Strict defeats the coalescing into posted bursts, so you pay a round trip per transaction.

Reads are flat everywhere, within 2.4 percent, as expected since a read is non posted anyway. Static TLB writes inside the window, DMA and multicast are all flat as well.

Constructor is safe but the write path is not, and metal hits the dynamic path whenever a core is not statically mapped or a transfer exceeds the window. Worth reconsidering whether Strict is the right default here.

device/api/umd/device/tt_device/tt_device.hpp
ajovanovicTT: No need to update baselines since they include only Cluster constructor benchmarks on n150 due to flakiness of CI machines, which are verified unaffected by the changes.

PR #3406, opened 2026-09-07
Route TLB window configure through the safe IO path and enable the safe IO warm reset tests

### LINKED ISSUE #3281: Safe-IO warm reset tests don't catch SIGBUS
**Description**
DISABLED_SafeApi* warm-reset tests (tests/api/test_warm_reset.cpp) are meant to prove safe-IO throws SigbusError on a reset-invalidated PCIe mapping instead of crashing the process. Currently they don't reliably catch it: the reset was triggered serially from the main thread instead of racing an in-flight read/write, and SiliconTlbWindow::configure() (the TLB reconfigure that runs before every read/write) was not wrapped by the execute_safe/sigsetjmp guard — only the data-transfer call was. Since KMD does invalidate mappings on reset, a SIGBUS during configure() was going uncaught and silently killing the whole process.

**Scope**
 - Add a warm_reset() from separate worker which will invalidate the mappings prior to the TTDevice API touching them
 - Route SiliconTlbWindow::configure() through execute_safe (guarded by safe_io_), same as read_block/write_block/etc., so a SIGBUS hit during TLB reconfigure is caught as SigbusError too.

**Timeline**
No hard deadline.

Yes — safe-IO tests were disabled and, even when run manually, the process was silently killed by an unhandled SIGBUS instead of exercising the intended recovery path, because configure() wasn't covered by the guard.

**Definition of Done**
 - Reset races the read/write in a separate worker thread in the multi-threaded/multi-process safe-IO tests.
 - configure() covered by execute_safe so SIGBUS during TLB reconfigure throws SigbusError.
 - DISABLED_SafeApi* tests pass consistently (no unhandled process termination)

ajovanovicTT:
### Issue
#3281

Stacked on #3414, which renames the hang_detection_tests target to destructive_tests. Review that one first.

### Description
Reconfiguring a TLB window was the one BAR0 touching entry point on SiliconTlbWindow's transfer path left outside the IoSafety guard, so a reconfigure after a warm reset killed the process with a raw SIGBUS instead of raising SigbusError. TlbWindow's constructor also writes outside the guard, but only to a mapping it just allocated: if the device is already gone, that allocation throws a RuntimeError before any write happens. This routes configure through the safe IO path and moves the safe API warm reset tests out of api_tests, where they were disabled, into destructive_tests, where they are enabled and actually exercise the fix.

### List of the changes
- SiliconTlbWindow::configure now routes through execute_safe when IoSafety is enabled, guarding the virtual configure(tlb_data) that every configure overload funnels into.
- Moved the three safe API warm reset tests from tests/api/test_warm_reset.cpp into a new tests/destructive/test_safe_io_warm_reset.cpp and enabled them.
- SafeApiHandlesReset keeps the racing shape and its delay parameterization. SafeApiMultiThreaded and SafeApiMultiProcess are deterministic and wait for warm_reset to return before touching the device, since they test guard isolation rather than timing.
- SetUp asserts a minimum KMD version, because without mapping invalidation on reset nothing faults and the tests would loop to timeout. TearDown restores SIG_DFL, as the sigbus handler is process wide.
- Extracted wait_for_child and terminate_processes into tests/test_utils/process_utils.hpp.

### Testing
The tests were run on hardware and pass on a host with a recent KMD. They hang on KMD 2.4.1, which does not invalidate the mappings, hence the version assert in SetUp. CI coverage is the Run destructive tests workflow, which can be dispatched manually per card.
Also tested locally on bgd-lab-06

### API Changes
This PR has no API changes.

### DIFF
--- a/.github/workflows/run-destructive-tests.yml
+++ b/.github/workflows/run-destructive-tests.yml
@@ -1,4 +1,5 @@
-# Destructive NOC hang / MMIO-timeout tests.
+# Destructive tests: NOC hang / MMIO-timeout detection, and the safe-I/O warm-reset tests
+# (which reset the board repeatedly to prove a reset-invalidated mapping surfaces as SigbusError).
 # Runs automatically every weeknight on a set of cards, and can also be
 # triggered manually from the Actions tab ("Run workflow"), picking the card to run on.
 name: Run destructive tests
--- a/device/pcie/silicon_tlb_window.cpp
+++ b/device/pcie/silicon_tlb_window.cpp
@@ -158,6 +158,8 @@ void SiliconTlbWindow::read_block_impl(uint64_t offset, void *data, size_t size)
     }
 }
 
+// A fault here updates nothing (the handle assigns tlb_config_ only after the write lands), and every
+// transfer reconfigures first, so a half-written window is always fully rewritten before it is used.
 void SiliconTlbWindow::configure_impl(const tlb_data &new_config) {
     TlbWindow::configure(new_config);
     update_io_timeout_callback();
--- a/device/pcie/silicon_tlb_window.cpp
+++ b/device/pcie/silicon_tlb_window.cpp
@@ -76,11 +76,6 @@ void SiliconTlbWindow::set_io_timeout_hang_check(const std::function<bool(NocId)
     update_io_timeout_callback();
 }
 
-void SiliconTlbWindow::configure(const tlb_data &new_config) {
-    TlbWindow::configure(new_config);
-    update_io_timeout_callback();
-}
-
 void SiliconTlbWindow::update_io_timeout_callback() {
     if (!hang_check_) {
         // No hang check wired: default to treating every overrun as a false alarm so a bare per-op
@@ -161,6 +156,11 @@ void SiliconTlbWindow::read_block_impl(uint64_t offset, void *data, size_t size)
     } else {
         umd::memcpy_from_device(data, src, size, io_timeout_callback_);
     }
+}
+
+void SiliconTlbWindow::configure_impl(const tlb_data &new_config) {
+    TlbWindow::configure(new_config);
+    update_io_timeout_callback();
 }
 
 void SiliconTlbWindow::memcpy_from_device(
@@ -347,6 +347,14 @@ void SiliconTlbWindow::read_block(uint64_t offset, void *data, size_t size) {
         return;
     }
     read_block_impl(offset, data, size);
+}
+
+void SiliconTlbWindow::configure(const tlb_data &new_config) {
+    if (io_safety_ == IoSafety::Enabled) {
+        execute_safe(&SiliconTlbWindow::configure_impl, new_config);
+        return;
+    }
+    configure_impl(new_config);
 }
 
 }  // namespace tt::umd
--- a/tests/destructive/test_safe_io_warm_reset.cpp
+++ b/tests/destructive/test_safe_io_warm_reset.cpp
@@ -38,6 +38,7 @@
 #include <array>
 #include <atomic>
 #include <chrono>
+#include <csignal>
 #include <cstddef>
 #include <cstdint>
 #include <exception>
@@ -63,6 +64,7 @@
 #include "umd/device/types/communication_protocol.hpp"
 #include "umd/device/types/core_coordinates.hpp"
 #include "umd/device/utils/error.hpp"
+#include "umd/device/utils/kmd_versions.hpp"
 #include "umd/device/utils/semver.hpp"
 #include "utils.hpp"
 
@@ -107,8 +109,7 @@ bool wait_for_condition(Predicate predicate, const std::chrono::milliseconds tim
 bool is_galaxy_board(int pci_device_id) {
     auto tt_device = TTDevice::create(pci_device_id);
     tt_device->init_tt_device();
-    const BoardType board_type = tt_device->get_board_type();
-    return board_type == tt::BoardType::UBB_WORMHOLE || board_type == tt::BoardType::UBB_BLACKHOLE;
+    return is_galaxy_board_type(tt_device->get_board_type());
 }
 
 }  // namespace
@@ -130,17 +131,6 @@ protected:
             GTEST_SKIP() << "Warm reset is disabled on ARM64 due to instability.";
         }
 
-        // Everything below rests on KMD invalidating a device's PCIe mappings when it resets the
-        // device. Older drivers leave the mappings live: the reset still happens, but the host keeps
-        // a valid vma over a device that is gone, so the TLB reconfigure's posted write is dropped
-        // and the transfer's read comes back as all-ones instead of raising SIGBUS. Nothing then
-        // faults, and a test waiting for SigbusError just runs its I/O loop to its timeout.
-        static constexpr SemVer MIN_KMD_VERSION{2, 6, 0};
-        const SemVer kmd_version = PCIDevice::read_kmd_version();
-        ASSERT_TRUE(kmd_version >= MIN_KMD_VERSION)
-            << "KMD " << kmd_version.str() << " does not invalidate PCIe mappings on device reset; the safe-I/O "
-            << "warm-reset tests need " << MIN_KMD_VERSION.str() << " or newer.";
-
         pci_device_ids_ = PCIDevice::enumerate_devices();
         if (pci_device_ids_.empty()) {
             GTEST_SKIP() << "No PCI devices found.";
@@ -151,6 +141,15 @@ protected:
         static const bool galaxy_board = is_galaxy_board(pci_device_ids_.front());
         if (galaxy_board) {
             GTEST_SKIP() << "Skipping test calling warm_reset() on Galaxy configurations.";
+        }
+
+        // Everything below rests on KMD invalidating PCIe mappings on reset; older drivers leave a
+        // stale mapping live, so nothing faults and a SigbusError-waiting test just times out. A
+        // skip, not a hard fail: read_kmd_version() also reads back {0,0,0} with no module loaded.
+        const SemVer kmd_version = PCIDevice::read_kmd_version();
+        if (!(kmd_version >= KMD_RESET_MAPPING_INVALIDATION)) {
+            GTEST_SKIP() << "KMD " << kmd_version.str() << " does not invalidate PCIe mappings on device reset; the "
+                         << "safe-I/O warm-reset tests need " << KMD_RESET_MAPPING_INVALIDATION.str() << " or newer.";
         }
     }
 
@@ -186,12 +185,11 @@ protected:
         }
     }
 
-    // Drops the handles and restores the default SIGBUS disposition. The handler is process-wide, so
-    // leaving it installed would turn an unrelated SIGBUS in a later test in this binary into a
-    // silent _exit() instead of a crash with a usable core dump.
+    // Restores SIG_DFL before dropping the handles: ~BlackholeTTDevice() does unguarded BAR2 writes
+    // that can SIGBUS on a dead mapping, so clearing first would silently _exit() instead of crashing.
     void close_safe_devices() {
-        tt_devices_.clear();
         TTDevice::set_sigbus_safe_handler(false);
+        tt_devices_.clear();
     }
 
     TTDevice* first_device() { return tt_devices_.at(pci_device_ids_.front()).get(); }
@@ -285,10 +283,22 @@ TEST_P(SafeIoWarmResetParamTest, SafeApiHandlesReset) {
     reset_issued_ = true;
     std::thread reset_worker([&, delay_us = GetParam()]() {
         std::this_thread::sleep_for(std::chrono::microseconds(delay_us));
-        reset_ok = WarmReset::warm_reset();
+        try {
+            reset_ok = WarmReset::warm_reset();
+        } catch (const std::exception&) {
+            // An uncaught exception here would call std::terminate() on this thread.
+        }
     });
 
-    const bool caught_sigbus = hammer_until_sigbus();
+    bool caught_sigbus = false;
+    try {
+        caught_sigbus = hammer_until_sigbus();
+    } catch (...) {
+        // Join before rethrowing: reset_worker is still joinable here, and unwinding past a
+        // joinable std::thread also calls std::terminate().
+        reset_worker.join();
+        throw;
+    }
 
     reset_worker.join();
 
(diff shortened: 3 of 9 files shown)


ajovanovicTT:
Link to successful workflow dispatch inlcuding reenabled tests: https://github.com/tenstorrent/tt-umd/actions/runs/34143195348/job/101809863342

tests/test_utils/process_utils.hpp
```diff
@@ -0,0 +1,53 @@
...
+#pragma once
+
+#include <sys/types.h>
+#include <sys/wait.h>
+
+#include <chrono>
+#include <csignal>
+#include <initializer_list>
+#include <thread>
+#include <vector>
```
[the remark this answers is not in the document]
ajovanovicTT: There is already < csignal >, unjustified.

tests/destructive/test_safe_io_warm_reset.cpp
```diff
@@ -0,0 +1,469 @@
...
+    // Opens one safe-API TTDevice per PCI device. use_safe_api is what installs the SIGBUS handler
+    // and selects IoSafety::Enabled for the windows these devices do I/O through.
+    void open_safe_devices() {
+        for (int pci_device_id : pci_device_ids_) {
+            auto tt_device = TTDevice::create(pci_device_id, IODeviceType::PCIe, /*use_safe_api=*/true);
+            tt_device->set_power_state(TTDevice::PowerState::BUSY);
+            tt_device->init_tt_device();
+            tensix_core_ = tt_device->get_soc_descriptor().get_cores(CoreType::TENSIX, CoordSystem::TRANSLATED)[0];
+            tt_devices_[pci_device_id] = std::move(tt_device);
+        }
```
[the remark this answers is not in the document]
ajovanovicTT: tensix_core_ converging to the same value across devices is expected

tests/destructive/test_safe_io_warm_reset.cpp
```diff
@@ -0,0 +1,469 @@
...
+
+    // The reset runs on its own worker so it races transfers that are already in flight on this
+    // thread, rather than being serialized behind them.
+    reset_issued_ = true;
+    std::thread reset_worker([&, delay_us = GetParam()]() {
+        std::this_thread::sleep_for(std::chrono::microseconds(delay_us));
+        reset_ok = WarmReset::warm_reset();
+    });
+
+    const bool caught_sigbus = hammer_until_sigbus();
```
pjanevskiTT: `hammer_until_sigbus()` deliberately lets anything other than `SigbusError` propagate, and at this point `reset_worker` (286) is still joinable — so a non-Sigbus exception unwinds past it, `~std::thread()` calls `std::terminate()`, and the whole `destructive_tests` binary aborts with no gtest report and no run for the remaining tests. `DeviceTimeoutError` is a realistic source: the cached window has the hang check wired (`PcieProtocol::set_io_timeout_callback`), so an MMIO overrun during the reset probes `is_noc_hung()`, the probe read on a down link returns `HANG_READ_VALUE`, and the transfer throws. The version in `tests/api/test_warm_reset.cpp` joined the reset thread before `FAIL()`; that guard didn't come across in the move.

Same shape at 346: `reset_worker`'s lambda calls `WarmReset::warm_reset()` with no try/catch, and that has throwing paths (the pre-reset notifier's asio sockets, `enumerate_devices_info()`, `read_command_byte()`'s `UMD_THROW`) — an exception there terminates the process too, and leaves `reset_complete` unset so both workers first burn the full 180 s `RESET_WAIT_TIMEOUT`. A join-on-scope-exit guard plus a try/catch in each reset lambda covers both.

tests/destructive/test_safe_io_warm_reset.cpp
```diff
@@ -0,0 +1,469 @@
...
+            tt_device->init_tt_device();
+            tensix_core_ = tt_device->get_soc_descriptor().get_cores(CoreType::TENSIX, CoordSystem::TRANSLATED)[0];
+            tt_devices_[pci_device_id] = std::move(tt_device);
+        }
+    }
+
+    // Drops the handles and restores the default SIGBUS disposition. The handler is process-wide, so
+    // leaving it installed would turn an unrelated SIGBUS in a later test in this binary into a
+    // silent _exit() instead of a crash with a usable core dump.
+    void close_safe_devices() {
```
pjanevskiTT, line 190: These two lines are in the wrong order for the reason the comment above them gives. `tt_devices_.clear()` runs the device destructors while the process-wide SIGBUS handler is still installed and no jump target is armed — and `BlackholeTTDevice::~BlackholeTTDevice()` writes BAR2 for every programmed iATU region (`*reinterpret_cast<volatile uint32_t*>(bar2 + iatu_base + 0x04) = 0`). Post-reset that store faults, `sigbus_handler` sees `jump_set == 0` and calls `_exit(sig)`: the binary vanishes mid-suite with exit code 7, no gtest output and no core — the precise outcome the comment says it's avoiding.

Restoring `SIG_DFL` first at least turns it into a visible crash, but the destructors are unguarded MMIO either way, so the real question is whether post-reset teardown should be doing BAR2 writes at all.

tests/destructive/test_safe_io_warm_reset.cpp
```diff
@@ -0,0 +1,469 @@
...
+        }
+
+        // Everything below rests on KMD invalidating a device's PCIe mappings when it resets the
+        // device. Older drivers leave the mappings live: the reset still happens, but the host keeps
+        // a valid vma over a device that is gone, so the TLB reconfigure's posted write is dropped
+        // and the transfer's read comes back as all-ones instead of raising SIGBUS. Nothing then
+        // faults, and a test waiting for SigbusError just runs its I/O loop to its timeout.
+        static constexpr SemVer MIN_KMD_VERSION{2, 6, 0};
+        const SemVer kmd_version = PCIDevice::read_kmd_version();
+        ASSERT_TRUE(kmd_version >= MIN_KMD_VERSION)
```
pjanevskiTT: Every other gate in this `SetUp` skips; this one hard-fails, and it runs ahead of the no-devices and Galaxy checks. `PCIDevice::read_kmd_version()` returns `SemVer{0,0,0}` when `/sys/module/tenstorrent/version` is absent (`pci_device.cpp:766-768`), so on any host without the module loaded — and on any scheduled card still below 2.6.0 — all 8 tests in the fixture go red, the `run` job fails, and `notify-slack` fires (`if: failure() && github.event_name == 'schedule'`). It also pre-empts the two skips below it, so the `6u` / `wh-glx-6u` runners in the dispatch list fail here instead of reaching the Galaxy skip meant for them. `GTEST_SKIP()` with the same message keeps the diagnostic without the page.

device/pcie/silicon_tlb_window.cpp
```diff
@@ -349,4 +349,12 @@ void SiliconTlbWindow::read_block(uint64_t offset, void *data, size_t size) {
     read_block_impl(offset, data, size);
 }
 
+void SiliconTlbWindow::configure(const tlb_data &new_config) {
```
pjanevskiTT, line 354: The description calls `configure` "the one BAR0 touching entry point on SiliconTlbWindow left outside the IoSafety guard", but the construction-time write is still outside it: `TlbWindow`'s constructor calls `tlb_handle->configure(aligned_config)` directly (`tlb_window.cpp:36`) rather than through the virtual, so it never reaches this `execute_safe`. `SiliconTlbHandle::configure` then stores the config words into the mapping, and if that mapping is already dead the fault lands with `jump_set == 0` → `_exit(sig)`, the silent death this PR removes for reconfigure.

`PcieProtocol::get_cached_tlb_window()` is what makes it reachable — it lazily builds the `IoSafety::Enabled` window on the first data/ctrl transfer (`pcie_protocol.cpp:73-81`) — though on a stale fd `allocate_tlb` may throw first. Worth either routing the constructor through the same guard or narrowing the claim in the description.

device/pcie/silicon_tlb_window.cpp
```diff
@@ -163,6 +158,11 @@ void SiliconTlbWindow::read_block_impl(uint64_t offset, void *data, size_t size)
     }
 }
 
+void SiliconTlbWindow::configure_impl(const tlb_data &new_config) {
```
pjanevskiTT, line 163: Now that a fault here throws instead of exiting, the state the window is left in starts to matter. `SiliconTlbHandle::configure` writes the config words and only then assigns `tlb_config_ = new_config`; `TlbWindow::configure` only then updates `offset_from_aligned_addr`; and `update_io_timeout_callback()` on the line below never runs. A SIGBUS partway through therefore leaves the hardware half-programmed, `tlb_config_` and `offset_from_aligned_addr` describing the previous config, and `io_timeout_callback_` still bound to the old `noc_sel`.

`count_sigbus_transfers` retries on the same window three rounds by design ("the guard is proven to re-arm after longjmp-ing out of the previous fault"), so retry-on-the-same-window is the contract under test — a scope guard that refreshes the callback on the throw path would keep the object self-consistent for it.

tests/destructive/test_safe_io_warm_reset.cpp
```diff
@@ -0,0 +1,469 @@
...
+    reset_issued_ = true;
+    const bool reset_ok = WarmReset::warm_reset();
+
+    // Released unconditionally: on a failed reset the children must still run and report what they
+    // saw rather than sitting on the handshake until their own timeout.
+    for (int i = 0; i < NUM_CHILDREN; ++i) {
+        reset_done.notify(i);
+    }
+
+    auto describe = [](int rc) -> const char* {
```
pjanevskiTT, line 451: No case for the status the regression this file exists to catch actually produces. `sigbus_handler`'s fallback is `_exit(sig)`, so a child that takes an unguarded SIGBUS exits *normally* with status 7 — `WIFEXITED` is true so the `EXPECT_TRUE` at 461 passes, and `describe(7)` falls through to "unknown failure (exit code 7)". A `case SIGBUS:` naming it as an unguarded SIGBUS makes the most diagnostic outcome self-explaining.

tests/destructive/test_safe_io_warm_reset.cpp
```diff
@@ -0,0 +1,469 @@
...
+        if (utils::is_arm_platform()) {
+            GTEST_SKIP() << "Warm reset is disabled on ARM64 due to instability.";
+        }
+
+        // Everything below rests on KMD invalidating a device's PCIe mappings when it resets the
+        // device. Older drivers leave the mappings live: the reset still happens, but the host keeps
+        // a valid vma over a device that is gone, so the TLB reconfigure's posted write is dropped
+        // and the transfer's read comes back as all-ones instead of raising SIGBUS. Nothing then
+        // faults, and a test waiting for SigbusError just runs its I/O loop to its timeout.
+        static constexpr SemVer MIN_KMD_VERSION{2, 6, 0};
```
pjanevskiTT: This is a new KMD capability gate, and `device/api/umd/device/utils/kmd_versions.hpp` is where the others live (`KMD_ARCH_AGNOSTIC_RESET`, `KMD_POWER_STATE`, `KMD_READ_ONLY_PAGE_PINNING`, `KMD_TLB_DMABUF_EXPORT`), each with a comment naming its feature. As a local literal the rationale above it is invisible to the rest of the tree, and it happens to equal `KMD_POWER_STATE` numerically, so a future bump of one reads as the other. `KMD_RESET_MAPPING_INVALIDATION = SemVer(2, 6, 0)` there, referenced here, avoids both.

tests/destructive/test_safe_io_warm_reset.cpp
```diff
@@ -0,0 +1,469 @@
...
+    static constexpr auto BUSY_POLL_WINDOW = std::chrono::microseconds(0);
+    static constexpr auto POLL_INTERVAL = std::chrono::microseconds(1'000);
+    return utils::poll_until(predicate, timeout, BUSY_POLL_WINDOW, POLL_INTERVAL);
+}
+
+// Galaxy (UBB) boards are reset over IPMI by ubb_warm_reset(), not by the PCIe-level warm_reset()
+// these tests drive, so they have to be skipped there. Probed from a bare TTDevice rather than from
+// a Cluster's descriptor: the board type needs one initialised device, whereas building a Cluster
+// runs a full topology discovery.
+bool is_galaxy_board(int pci_device_id) {
```
pjanevskiTT, line 109: `test_api_common.hpp`, already included here, has `is_galaxy_configuration(Cluster*)` at line 109 with the identical `UBB_WORMHOLE || UBB_BLACKHOLE` test, and the sibling `test_warm_reset_after_noc_hang.cpp:61` uses it. The reason for a second copy is sound — probing a bare TTDevice is much cheaper than full topology discovery — but that argues for a different *source* of the board type, not a second copy of the predicate. An `is_galaxy_board_type(BoardType)` in `test_api_common.hpp` that both call keeps the cheap probe without a second list to keep in sync when a UBB board type is added.

.github/workflows/run-destructive-tests.yml
```diff
@@ -1,12 +1,13 @@
...
 # triggered manually from the Actions tab ("Run workflow"), picking the card to run on.
-name: Run hang detection tests
+name: Run destructive tests
 
 on:
   schedule:
-    # Run nightly at 1 AM UTC.
-    - cron: "0 1 * * *"
+    # Run nightly at 1 AM UTC, Monday through Friday.
+    - cron: "0 1 * * 1-5"
```
pjanevskiTT, line 10: This is the one change here that reduces coverage — 2 of 7 nightly runs across 5 cards — and it's unrelated to the fix it's bundled with; the description lists it without a rationale. If the motivation is that the new tests reset boards repeatedly and that's not worth doing at the weekend, worth saying so in the comment above, so the next person doesn't "fix" it back. (Minor: 1 AM UTC Monday is Sunday evening in the Americas, so `1-5` isn't quite "every weeknight".)

tests/test_utils/process_utils.hpp
```diff
@@ -0,0 +1,57 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent Inc.
```
pjanevskiTT, line 1: ok will assume this code is just moved

pjanevskiTT (approved):

device/pcie/silicon_tlb_window.cpp
ajovanovicTT: Narrowed the claim in the description. Leaving the constructor as is: the handle is allocated in the same expression, so the constructor only ever writes to a mapping created a moment earlier, and on an already-dead device tt_tlb_alloc fails first, so allocate_tlb throws RuntimeError rather than faulting. What is left is a reset landing between the mmap and the config write, which I would rather not restructure the three window constructors for in this PR.

device/pcie/silicon_tlb_window.cpp
ajovanovicTT: Leaving as is. Because configure writes before assigning tlb_config_, the handle config, offset_from_aligned_addr and io_timeout_callback_ all skip together and still agree with each other, so refreshing the callback on the throw path would recompute the value it already holds. The hardware side is corrected by the next configure: configure_tlb rewrites every config word unconditionally (verify_config is read-back verification, not diff-and-skip), and every transfer reconfigures before touching the window.

ajovanovicTT: Just stated this in the comment above SiliconTlbWindow::configure_impl

Issue #3294, opened 2026-08-27
Define Agents.md


broskoTT:
As we're using AI more and more, we need a set of documents guiding our AIs to the workflow and code culture we already developed in our repo.

This will likely continue growing, but for now we should put extra effort into jumpstarting this, including thorough discussions and reviews.

Definition of done:
First iteration of adding some agent support for our repo is done, we will likely continue iterating.

PR #3426, opened 2026-09-11
[Agents] AGENTS.md: fill in the PR split section

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/3294

### Description
Fills in the `## PR split` section: what makes a PR contextually small, the categories a PR should
match exactly one of, and the warning against splitting so far that the stack gets harder to review
than a single PR would have been.

Also states the ordering rule: integrate first with an empty implementation, then implement, so no
PR in a stack lands dead code.

### List of the changes
- Add the PR split rules under the existing header in `AGENTS.md`.

### Testing
N/A, documentation only.

### API Changes
This PR has no API changes.

### Full stack diff
https://github.com/tenstorrent/tt-umd/compare/main...brosko/agents_9

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -21,6 +21,15 @@
 - Once you finish the PR stack of changes go through each of the created PRs, and add a new section in the PR description at the end "### Full stack diff" and add a compare link between last branch and main, so full diff can be easily accessed from any PR.
 
 ## PR split
+- PRs should be as small as possible to help reviewers keep the focus while reviewing scoped-down changes. Small doesn't necessarily mean low number of lines of code. Contextually small PRs can have a huge diff but with the same change (for example renaming a class). Do not combine multiple functional changes into a single PR, even if the total line count is small, as it is harder for a reviewer to take a grasp on the full combination of functional changes.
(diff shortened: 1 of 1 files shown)


ajovanovicTT (approved):

AGENTS.md
```diff
@@ -21,6 +21,15 @@
...
+- But be careful not to overdo the PR split. Making 10 PRs for 10 related function changes with 5 lines each is making things less readable rather than just keeping the context per PR small. Use your judgement, and if unsure ask the user.
+- To help with a meaningful PR split, PRs should match only one of these categories instead of mixing them: Feature, Performance, Bug fix, Cleanup, Test Only. Tests can be mixed with other categories if they are explicitly used to prove the changes in the same PR
+- For changes which have a large scope, try to separate into multiple changes where each of those has a smaller scope
+- The point of splitting changes into multiple PRs is to make them easier to review, if you estimate splitting will make overall review harder, don't split.
+- Don't implement first then integrate into existing code. Rather, try to do integration first with empty implementation, then follow with implementation. That way we don't have intermediate dead code.
```
pjanevskiTT, line 29: I think this bullet and the next one is very specific to one type of changes. It's ok maybe to have it in another file or under some other sections for these type of changes. Imo it's fine to implement the component with tests and show how one should use it and then as a follow up integrate it into code if it requires more work and it's not a simple delete 1 add 1 function call

pjanevskiTT (approved):

PR #3431, opened 2026-09-11
[Agents] AGENTS.md: fill in the External knowledge section

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/3294

### Description
Fills in the last empty section, `## External knowledge`: treat the parent of this repo as the work
folder holding the neighbouring Tenstorrent repos, and use the `TT_SUMMARY.md` index there to find
which one holds the code in question instead of searching blind.

Also points at Glean for internal docs and already reported issues, with the instruction to say so
when that connection is missing rather than silently investigating with less context.

This completes the stack started in #3423, so `AGENTS.md` now has no empty headers left.

### List of the changes
- Add the External knowledge rules under the existing header in `AGENTS.md`.

### Testing
N/A, documentation only.

### API Changes
This PR has no API changes.

### Full stack diff
https://github.com/tenstorrent/tt-umd/compare/main...brosko/agents_9

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -57,3 +57,9 @@
 - Any suggested comments should be presented through a git unstaged text file for user to review.
 
 ## External knowledge
+Use the parent of this repo as the work folder where you can git clone relevant repositories, or look at already cloned ones.
+In the same local folder, there should be TT_SUMMARY.md which you can load and use to point you to the right repository when you're looking for some code.
+If that file doesn't exist, generate it before proceeding. Scan local repos in work folder, analyze them briefly and produce the summary.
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -57,3 +57,9 @@
 - Any suggested comments should be presented through a git unstaged text file for user to review.
 
 ## External knowledge
```
pjanevskiTT, line 59: can we also maybe list some repos that are useful for us so when someones says "look at metal" agents know where to try and find this?

pjanevskiTT (approved):

ajovanovicTT (approved):

PR #3440, opened 2026-09-14
Cover the supported address ranges with real device I/O

pjanevskiTT:
### Issue

Found while enabling tenstorrent/tt-umd-internal#23; the coverage gap is not specific to it.

### Description

Read/write coverage touched one address per core. A window stride that is wrong by a power of two still round-trips there and only shows up at the ends of a range, so these probe the bottom, middle and top of worker L1 and of a DRAM bank, and give each DRAM channel its own pattern.

Touching the ends rather than streaming the range keeps this viable on an emulator, where a single 4 MB transfer already dominates a run.

### List of the changes

- Three tests on the existing `ClusterReadWriteL1Test` fixture, so they run against both silicon and simulation.
- The DRAM top-of-bank probe is limited to simulation. `dram_bank_size` is 4 GiB on Blackhole, where `bank_size / 2` round-trips and `bank_size - 256` reads back zeros, so the descriptor's bank size is not the span one DRAM core addresses and nothing states what is. Filed separately.

### Testing

On the previous CI run of this work the L1 range test and the per-channel DRAM test passed on all three Blackhole runners (p150b, P300, p150b-viommu); the DRAM range test failed there only at the top-of-bank probe, which is the probe this PR limits to simulation. The wormhole runner in that run had a dead board and proved nothing. All three pass on the qsr.s1 emulator. CI on this PR is what confirms the guard.

### API Changes

This PR has no API changes.

### Full stack diff


tests/api/test_device_io.cpp
```diff
@@ -499,6 +499,123 @@ TEST_P(ClusterReadWriteL1Test, ReadWriteL1) {
     }
 }
 
+// Touch the ends of a range rather than streaming it: on the emulator a single 4 MB transfer
+// already dominates a run, so covering 1 GB of DRAM by filling it is not viable. A wrong window
+// stride shows up at the top of a range, which is what these probe.
+namespace {
+
+// Write a distinct pattern at `offset`, read it back, and compare.
+void expect_round_trip(Cluster& cluster, ChipId chip_id, const CoreCoord& core, uint64_t offset, uint32_t seed) {
```
nbuncicTT: nit: Instead of a free function, use a test fixture class and make it a public method.

tests/api/test_device_io.cpp
```diff
@@ -499,6 +499,123 @@ TEST_P(ClusterReadWriteL1Test, ReadWriteL1) {
...
+    SCOPED_TRACE(fmt::format("core {} offset {:#x}", core.str(), offset));
+
+    constexpr size_t block_size = 256;
```
nbuncicTT: nit: Make the constant upper case and move it into an attribute of the test class.

tests/api/test_device_io.cpp
```diff
@@ -499,6 +499,123 @@ TEST_P(ClusterReadWriteL1Test, ReadWriteL1) {
...
+    std::vector<uint8_t> data(block_size);
```
nbuncicTT: If you already know the block size at compile time, and you do based on the `constexpr` `block_size`, use `std::array` for both data and readback and put them as attributes of the class.

tests/api/test_device_io.cpp
```diff
@@ -499,6 +499,123 @@ TEST_P(ClusterReadWriteL1Test, ReadWriteL1) {
...
+
+    for (auto chip_id : cluster->get_target_device_ids()) {
+        const SocDescriptor& soc_desc = cluster->get_soc_descriptor(chip_id);
+        const CoreCoord tensix_core = soc_desc.get_cores(CoreType::TENSIX)[0];
+        const uint64_t l1_size = static_cast<uint64_t>(soc_desc.worker_l1_size);
+        ASSERT_GT(l1_size, SAFE_IO_L1_ADDRESS + 512);
+
+        expect_round_trip(*cluster, chip_id, tensix_core, SAFE_IO_L1_ADDRESS, 0x11);
+        expect_round_trip(*cluster, chip_id, tensix_core, l1_size / 2, 0x22);
+        expect_round_trip(*cluster, chip_id, tensix_core, l1_size - 256, 0x33);
```
nbuncicTT: why 256? I mean I know it's because of the block size being 256, but instead of using a magic number use the same constant here and and in the function.

tests/api/test_device_io.cpp
```diff
@@ -499,6 +499,123 @@ TEST_P(ClusterReadWriteL1Test, ReadWriteL1) {
...
+
+    if (options.chip_type == ChipType::SIMULATION) {
+        cluster->start_device({.init_device = true});
+    }
+
+    for (auto chip_id : cluster->get_target_device_ids()) {
+        const SocDescriptor& soc_desc = cluster->get_soc_descriptor(chip_id);
+        const CoreCoord tensix_core = soc_desc.get_cores(CoreType::TENSIX)[0];
+        const uint64_t l1_size = static_cast<uint64_t>(soc_desc.worker_l1_size);
+        ASSERT_GT(l1_size, SAFE_IO_L1_ADDRESS + 512);
```
nbuncicTT: Why 512?

tests/api/test_device_io.cpp
```diff
@@ -499,6 +499,123 @@ TEST_P(ClusterReadWriteL1Test, ReadWriteL1) {
...
+    if (options.chip_type == ChipType::SIMULATION) {
+        cluster->start_device({.init_device = true});
+    }
+
+    for (auto chip_id : cluster->get_target_device_ids()) {
+        const SocDescriptor& soc_desc = cluster->get_soc_descriptor(chip_id);
+        const std::vector<CoreCoord>& dram_cores = soc_desc.get_cores(CoreType::DRAM);
+        ASSERT_FALSE(dram_cores.empty());
+        const uint64_t bank_size = soc_desc.dram_bank_size;
+        ASSERT_GT(bank_size, 512u);
```
nbuncicTT: Why 512u?

tests/api/test_device_io.cpp
```diff
@@ -499,6 +499,123 @@ TEST_P(ClusterReadWriteL1Test, ReadWriteL1) {
...
+        cluster->start_device({.init_device = true});
+    }
+
+    for (auto chip_id : cluster->get_target_device_ids()) {
+        const SocDescriptor& soc_desc = cluster->get_soc_descriptor(chip_id);
+        const std::vector<CoreCoord>& dram_cores = soc_desc.get_cores(CoreType::DRAM);
+        ASSERT_FALSE(dram_cores.empty());
+
+        constexpr size_t block_size = 256;
+        std::vector<std::vector<uint8_t>> written;
```
nbuncicTT: nit: I think here we have a pattern of over-using simple structures (i.e. primitive obsession) - we have two std::vectors whose data we want to print on a core - we could define a struct which bundles the `CoreCoord` and the data (`std::vector<uint8_t>`). If we'd bundle it, we could use a range-based loop when iterating over the test contents and we wouldn't use two vectors in that loop but one.

nbuncicTT (approved):

tests/api/test_device_io.cpp
pjanevskiTT: Moved onto the fixture: `ClusterReadWriteL1Test` now has `expect_round_trip` as a public method, and the pattern fill is a static `fill_block` on it so `ReadWriteEveryDramChannel` reuses it instead of repeating the loop.

Fixed in b29d7f06.

tests/api/test_device_io.cpp
pjanevskiTT: Now `static constexpr size_t BLOCK_SIZE = 256;` on `ClusterReadWriteL1Test`.

Fixed in b29d7f06.

tests/api/test_device_io.cpp
pjanevskiTT: Both are `std::array<uint8_t, BLOCK_SIZE>` attributes of the fixture now, `data_` and `readback_`.

Fixed in b29d7f06.

tests/api/test_device_io.cpp
pjanevskiTT: Uses `BLOCK_SIZE` now, here and at the DRAM top-of-bank probe.

Fixed in b29d7f06.

tests/api/test_device_io.cpp
pjanevskiTT: It was room for the block at `SAFE_IO_L1_ADDRESS` plus the one at the top, without the two overlapping. That is spelled out now: `ASSERT_GT(l1_size, SAFE_IO_L1_ADDRESS + 2 * BLOCK_SIZE);`, with a comment saying why.

Fixed in b29d7f06.

tests/api/test_device_io.cpp
pjanevskiTT: Same reason and same fix as the L1 one: `ASSERT_GT(bank_size, 2 * BLOCK_SIZE);`, room for the block at the bottom of the bank and the one at the top.

Fixed in b29d7f06.

tests/api/test_device_io.cpp
pjanevskiTT: Bundled into a local `struct ChannelProbe { CoreCoord core; std::array<uint8_t, BLOCK_SIZE> data; };`, so there is one vector and both loops are range-based.

Fixed in b29d7f06.

PR #3441, opened 2026-09-14
Model the Quasar ATT window and the map that holds one per role

pjanevskiTT:
### Issue

tenstorrent/tt-umd-internal#23

### Description

Quasar resolves a flat 64-bit address into a destination core plus a core-local address -- the inverse of Wormhole and Blackhole, where the core travels out of band in a TLB register. A driver targeting Quasar therefore folds the coordinate into the address, and the fold is defined by the window table the firmware programs.

This is the model for that table and nothing more: no product's values, no resolution, no device path.

### List of the changes

- `Window`, mirroring the ATT mask-table fields, with the address composition and decomposition over it, and the bounds checks that make a transfer spilling into the next slot an error rather than a misroute.
- `MapData`, holding one window per role together with that window's endpoint table, plus the `Table<T>` view it needs until UMD can use `std::span`.

### Testing

Five baremetal tests over a window shape kept independent of the product maps, so changing a map cannot move these expectations.

### API Changes

This PR has API changes: two new headers under `device/api/umd/device/coordinates/att/`. Nothing existing changes.

### Full stack diff


device/api/umd/device/coordinates/att/att_map.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+#pragma once
+
+#include <array>
+#include <cstddef>
+#include <cstdint>
+
+#include "umd/device/coordinates/att/att_window.hpp"
+
+namespace tt::umd::att {
```
nbuncicTT, line 13: Why does att have it's own namespace?

device/api/umd/device/coordinates/att/att_map.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+    constexpr const T& operator[](uint32_t index) const { return data[index]; }
+
+    constexpr uint32_t size() const { return count; }
+
+    constexpr bool empty() const { return count == 0; }
+};
+
+/** The role a window plays in a map. Count doubles as the window array extent. */
+enum class WindowClass : uint8_t {
```
nbuncicTT, line 37: What's a `WindowClass`? I understand the notion of Worker and Dram, but what's FullTile?

And also using an enum to know the count upfront is a neat thing, but I feel we shouldn't use it here, we should add a template for an array which can be indexed by enums and that could be a library component (not a hack used here).

also a nit here - use capital cases in enum classes.

device/api/umd/device/coordinates/att/att_map.hpp
```diff
@@ -0,0 +1,72 @@
...
+#include "umd/device/coordinates/att/att_window.hpp"
+
+namespace tt::umd::att {
+
+/**
+ * A pointer and a count over a transcribed table, deduced from the array so no length is spelled by
+ * hand. Stands in for std::span, which UMD cannot use until C++20.
+ */
+template <typename T>
+struct Table {
```
nbuncicTT, line 20: This is not a table, it's an `ArrayView`, so please rename.

device/api/umd/device/coordinates/att/att_window.hpp
```diff
@@ -0,0 +1,88 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <cstdint>
+
+namespace tt::umd::att {
+
+/** Mask covering the low @p bits bits. */
+constexpr uint64_t low_mask(uint32_t bits) { return bits == 64 ? UINT64_MAX : ((uint64_t{1} << bits) - 1); }
```
nbuncicTT, line 12: Maybe a bit more verbose name like `low_bits_mask`?

And we could move it to a utility also.

device/api/umd/device/coordinates/att/att_map.hpp
nbuncicTT: And if it's for spans, it could be used in some other places, so it could be a utility.

nbuncicTT (approved):

device/api/umd/device/coordinates/att/att_map.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+/** Marks an endpoint table row the map leaves unprogrammed. No core can carry this coordinate. */
+inline constexpr uint16_t ENDPOINT_UNPOPULATED = 0xffff;
+
+/**
+ * One product's ATT map, as pure data. Every field is transcribed from the generated map the
+ * firmware programs; the resolver is the only code that interprets it, so supporting another
+ * product is adding data rather than adding a branch.
+ */
+struct MapData {
```
nbuncicTT, line 55: Why isn't this called something like `AttMap` or `AttData`?

`MapData` is too general for me.

device/api/umd/device/coordinates/att/att_map.hpp
```diff
@@ -0,0 +1,72 @@
...
+    /** Window per role, indexed by WindowClass. */
+    std::array<Window, WINDOW_CLASS_COUNT> windows;
+
+    /**
+     * Each window's endpoint table, one packed (y << 6) | x word per selector, in the POR package
+     * frame and laid out as the hardware register file holds it. Slot order follows the POR
+     * endpoint table rather than the mesh geometry, so nothing but the table can say which slot a
+     * core occupies.
+     */
+    std::array<Table<uint16_t>, WINDOW_CLASS_COUNT> endpoint_words;
```
nbuncicTT, line 73: In hardware, do we call this endpoint word or is this made-up for this PR? If it's in hardware like that (the docs), then fine, but if it's made-up, I'd prefer `endpoint_entries` or just plainly `endpoints` cause they symbolize one end point, and word(s) is too vague.

device/api/umd/device/coordinates/att/att_window.hpp
```diff
@@ -0,0 +1,88 @@
...
+    constexpr uint64_t local_address(uint64_t address) const { return address & low_mask(local_address_bits()); }
+
+    /**
+     * Width of the local-address field. A window with no selector spends every ignored comparison
+     * bit on the local address; otherwise the local address ends where the selector begins.
+     */
+    constexpr uint32_t local_address_bits() const { return endpoint_size == 0 ? mask_bits : endpoint_shift; }
+
+    /** One past the largest core-local offset the window can carry. */
+    constexpr uint64_t local_address_limit() const { return uint64_t{1} << local_address_bits(); }
```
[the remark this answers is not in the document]
pjanevskiTT: Documented rather than handled. The hardware mask field is six bits wide, so a programmed window carries 0..63 and every shift below stays defined; `low_mask`'s 64 case cannot be reached by any window the ATT can hold. Noted on the `mask_bits` field so the apparent support isn't read as a supported configuration.

device/api/umd/device/coordinates/att/att_window.hpp
```diff
@@ -0,0 +1,88 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <cstdint>
+
+namespace tt::umd::att {
+
+/** Mask covering the low @p bits bits. */
```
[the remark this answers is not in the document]
pjanevskiTT: Reworded to `Mask with the low @p bits set.`

device/api/umd/device/coordinates/att/att_map.hpp
pjanevskiTT: `Window` is a generic name and UMD already has `TlbWindow` and `WindowFlags`, so `tt::umd::att` keeps the ATT vocabulary from colliding with the TLB one. It also keeps the subsystem together, so adding a second product is a header in one place rather than names scattered through `tt::umd`.

device/api/umd/device/coordinates/att/att_map.hpp
pjanevskiTT: Three things here:

- **FullTile** is now documented: it is a tile's config aperture, the control registers every tile exposes, as opposed to the L1 a worker exposes through the worker window. A NEO tile appears in both tables and is reached through whichever window matches the address.
- **Casing** fixed -- `WORKER` / `DRAM` / `FULL_TILE` / `COUNT`, matching `CoreType` and `CoordSystem`.
- **The enum-indexed array** I agree with as an idea, but adding a shared library component in a PR about the ATT window model mixes change kinds, so I'd rather raise it separately and keep the `COUNT`-as-extent idiom here for now. Leaving this thread open as the reminder.

device/api/umd/device/coordinates/att/att_map.hpp
pjanevskiTT: Holding this one for now. `low_mask`, `Table<T>` and `MapData` are currently identical to tt-metal's `noc_att` equivalents -- same fields, same method bodies, differing only in namespace and `std::` qualification -- and we're proposing to collapse those duplicates into one shared header that metal aliases, since metal depends on UMD and not the reverse. Renaming first would make that land as a rename on top of a rename.

No disagreement with the naming itself, and I agree it wants to be a shared utility rather than living under `att/`. Happy to take it as soon as the sharing question is settled either way.

device/api/umd/device/coordinates/att/att_window.hpp
pjanevskiTT: Holding this one for now. `low_mask`, `Table<T>` and `MapData` are currently identical to tt-metal's `noc_att` equivalents -- same fields, same method bodies, differing only in namespace and `std::` qualification -- and we're proposing to collapse those duplicates into one shared header that metal aliases, since metal depends on UMD and not the reverse. Renaming first would make that land as a rename on top of a rename.

No disagreement with the naming itself, and I agree it wants to be a shared utility rather than living under `att/`. Happy to take it as soon as the sharing question is settled either way.

device/api/umd/device/coordinates/att/att_map.hpp
pjanevskiTT: Holding this one for now. `low_mask`, `Table<T>` and `MapData` are currently identical to tt-metal's `noc_att` equivalents -- same fields, same method bodies, differing only in namespace and `std::` qualification -- and we're proposing to collapse those duplicates into one shared header that metal aliases, since metal depends on UMD and not the reverse. Renaming first would make that land as a rename on top of a rename.

No disagreement with the naming itself, and I agree it wants to be a shared utility rather than living under `att/`. Happy to take it as soon as the sharing question is settled either way.

device/api/umd/device/coordinates/att/att_map.hpp
pjanevskiTT: Good question, and you're right to push: it is **not** hardware terminology. "Endpoint word" is tt-metal's coinage for the packed `(y << 6) | x` value a row holds; grendelemulation, which generates the table, calls the row an `EndpointTableEntry` with separate `x` and `y` fields. So `endpoint_entries` is closer to the source than what we have.

Holding it with the other renames above only because this type may end up shared with metal -- and if we do rename, I'd rather match grendelemulation's vocabulary than metal's.

device/api/umd/device/coordinates/att/att_map.hpp
pjanevskiTT: Closing this one out. `FULL_TILE` is documented (the per-tile config aperture, as against the L1 a worker exposes through `WORKER`, with a NEO tile appearing in both tables), and the casing is now `WORKER` / `DRAM` / `FULL_TILE` / `COUNT` to match `CoreType` and `CoordSystem`.

The enum-indexed array utility we're not taking for now -- same call as the other refactors on this stack. It's a good idea and belongs in its own PR rather than as extra hunks in the ATT window model.

PR #3442, opened 2026-09-14
Resolve a core and an offset into a flat address through an ATT map

pjanevskiTT:
### Issue

tenstorrent/tt-umd-internal#23

### Description

The resolver interprets a map: it inverts each window's endpoint table once at construction and answers a coordinate with the flat address that reaches it.

The inversion is also where the map is validated, because both faults it catches are silent at use. A table longer than its selector field leaves rows unreachable, and a coordinate named in two rows leaves a lookup unable to say which slot was meant; either surfaces as an access landing on the wrong core. A core the map does not name is refused for the same reason.

The resolver carries no product's values -- it reads whichever map it is handed -- so a second product is a table rather than a branch.

### List of the changes

- `att::Resolver`: builds the coordinate lookup at construction, validates the map, resolves a `(core, core_type, offset, size)` into one flat address.

### Testing

Two baremetal tests, on synthetic maps built to trip each validation check.

### API Changes

This PR has API changes: one new header under `device/api/umd/device/coordinates/att/`. Nothing existing changes.

### Full stack diff


device/api/umd/device/coordinates/att/att_resolver.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include "umd/device/types/core_coordinates.hpp"
+#include "umd/device/types/xy_pair.hpp"
+
+namespace tt::umd::att {
+
+/**
+ * Resolves a target core and a core-local offset into the single flat address the Quasar NOC ATT
+ * needs to reach it.
+ *
+ * Wormhole and Blackhole carry the destination core out of band in a TLB config register, leaving
```
nbuncicTT: This comment can be simplified in the sense where it can just state something like this: Wormhole/Blackhole TLB registers have pre-defined bits for the x, y coordinates, while Quasar TLB registers have two modes, the legacy x, y pair and the new ATT mode where the bits are programmed. The Resolver helps in resolving the endpoint, i.e. (x, y) pair, to the flat address.

The rest is a detail.

device/api/umd/device/coordinates/att/att_resolver.hpp
```diff
@@ -0,0 +1,50 @@
...
+ * the address on the wire core-local. Quasar is the opposite: the ATT resolves a flat address into
+ * a destination core plus a local address, so the driver folds the coordinate into the address.
+ */
+class Resolver {
```
nbuncicTT: I honestly think the name `Resolver` is very vague.

I'd prefer something more like `EndpointDecoder` as we decode the endpoint from the ATT or we just get an x, y pair and the decoding is a pass-through.

device/api/umd/device/coordinates/att/att_resolver.hpp
```diff
@@ -0,0 +1,50 @@
...
+     * @p size bytes.
+     *
+     * @param core Target core, in the frame the soc descriptor uses.
+     * @param core_type Selects the window. Inferring it from the coordinate instead is ambiguous:
+     *                  the D2D ingress rows overlap coordinates that legitimately belong to
+     *                  another window.
+     * @throws error::RuntimeError if the core has no window in this map, or the transfer would
+     *         run past the target's slot.
+     */
+    uint64_t resolve(tt_xy_pair core, CoreType core_type, uint64_t offset, uint64_t size) const;
```
nbuncicTT, line 41: Should we have an overloaded function here with `CoreCoord`? That way you cut the verbosity of the function call if somebody uses `CoreCoord` where the type is already there.

device/coordinates/att/att_resolver.cpp
```diff
@@ -0,0 +1,94 @@
...
+        // config aperture.
+        default:
+            return WindowClass::FullTile;
+    }
+}
+
+}  // namespace
+
+Resolver::Resolver(const MapData& map) : map_(map) {
+    for (size_t window_class = 0; window_class < WINDOW_CLASS_COUNT; ++window_class) {
```
nbuncicTT: nit: Why window_class? Isn't this just an index? I'd prefer `i` or `index`

device/api/umd/device/coordinates/att/att_resolver.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+private:
+    MapData map_;
+
+    /** Each window's endpoint words inverted into a coordinate lookup, built once at construction. */
+    std::array<std::unordered_map<uint16_t, uint32_t>, WINDOW_CLASS_COUNT> selectors_;
```
nbuncicTT, line 51: You could maybe add a comment here that the `map_` index is aligned with the `selectors_` index (i.e. they have same indexing values, for example when somebody accesses 0 in both data structure, they get the same windows/endpoints and selectors). 

And the index is the enum class value derived from `WindowClass`.

tests/baremetal/test_att_resolver.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+}  // namespace
+
+TEST(AttResolver, RejectsAnEndpointTableLongerThanItsSelectorField) {
+    constexpr uint16_t words[] = {0x104, 0x105, 0x106};
+
+    EXPECT_THROW(att::Resolver{map_with_worker_endpoints(TWO_SELECTOR_WINDOW, words)}, std::runtime_error);
+}
+
+TEST(AttResolver, RejectsAnEndpointTableThatNamesOneCoreTwice) {
```
nbuncicTT, line 98: Add a test which passes so we know it works correctly (i.e. the decoding/resolving works fine).

nbuncicTT (approved):
I think this PR and the previous one would be much easier with an `EnumArray` utility.

I made a draft PR here: https://github.com/tenstorrent/tt-umd/pull/3477

I think it's good and you can use it. I'd merge it as soon as I'm back. If you'd rather not wait for the util to land, then ignore this comment and make a follow-up which does the cleanup (I think using the util upfront is better as we solve many ambiguities with it).

device/coordinates/att/att_resolver.cpp
```diff
@@ -0,0 +1,94 @@
...
+    }
+}
+
+uint64_t Resolver::resolve(tt_xy_pair core, CoreType core_type, uint64_t offset, uint64_t size) const {
+    const auto x = static_cast<uint32_t>(core.x);
+    const auto y = static_cast<uint32_t>(core.y);
+    const WindowClass window_class = window_class_for(core_type);
+
+    const auto& selectors = selectors_[static_cast<size_t>(window_class)];
+    const auto entry = selectors.find(pack_coord(x + map_.package_offset_x, y + map_.package_offset_y));
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed, and the failure is worse than narrowing: packing **ORs** the two axes, so an out-of-range x spills into y. Descriptor `(129, 1)` is package `(130, 2)`, which packs to `0x082` -- the exact word the table holds for package `(2, 2)` -- so it resolved to a real but wrong core.

`resolve()` now range-checks both axes against the six-bit field before packing. The test uses that colliding coordinate, and I confirmed it fails without the check rather than passing for the wrong reason.

device/coordinates/att/att_resolver.cpp
```diff
@@ -0,0 +1,94 @@
...
+    switch (core_type) {
+        case CoreType::TENSIX:
+        case CoreType::WORKER:
+            return WindowClass::Worker;
+        case CoreType::DRAM:
+            return WindowClass::Dram;
+        // Routers, dispatch, ARC, PCIe and security cores are all reached through their per-tile
+        // config aperture.
+        default:
+            return WindowClass::FullTile;
```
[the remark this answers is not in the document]
pjanevskiTT: Fixed. The config-aperture types are enumerated (`ROUTER_ONLY`, `ARC`, `PCIE`, `SECURITY`, `L2CPU`, `DISPATCH`, `ETH`, `ACTIVE_ETH`, `IDLE_ETH`) and `default` throws, so `HARVESTED`, `COUNT`, `UNSPECIFIED` and any future type are refused instead of landing on the full-tile window.

tests/baremetal/test_att_resolver.cpp
```diff
@@ -0,0 +1,50 @@
...
+    .endpoint_table_offset = 0,
+    .translate_address = false,
+};
+
+}  // namespace
+
+TEST(AttResolver, RejectsAnEndpointTableLongerThanItsSelectorField) {
+    constexpr uint16_t words[] = {0x104, 0x105, 0x106};
+
+    EXPECT_THROW(att::Resolver{map_with_worker_endpoints(TWO_SELECTOR_WINDOW, words)}, std::runtime_error);
```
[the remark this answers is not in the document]
pjanevskiTT: Added. The PR now carries nine tests: resolution through each of the three roles, the selector coming from the table rather than the coordinate, the package offset being applied, an unknown core, a known core asked for through the wrong window, a transfer crossing the slot boundary, the coordinate-range guard, and the core-type guard.

device/api/umd/device/coordinates/att/att_resolver.hpp
pjanevskiTT: Renamed to `EndpointResolver`.

I kept "resolver" over "decoder" because the operation runs coordinate -> flat address, which is the *encode* direction -- tt-metal calls the equivalent step `encode()`. Decoding would be the reverse (address -> core), which we don't implement at all. `EndpointResolver` answers the vagueness without naming the direction backwards. Happy to revisit if you meant decoding the ATT table to find the endpoint, which is a fair reading of the internal step.

device/api/umd/device/coordinates/att/att_resolver.hpp
pjanevskiTT: It exists one PR up: `att::resolve_core(resolver, soc_descriptor, core, offset, size)` in #3409 takes a `CoreCoord` and reads the core type from the descriptor when the coordinate arrives already translated. It lives there rather than here because it needs a `SocDescriptor`, which this PR deliberately doesn't depend on -- the resolver interprets a map and nothing else.

device/api/umd/device/coordinates/att/att_resolver.hpp
pjanevskiTT: Added. The `selectors_` comment now states it is indexed by `WindowClass` like `map_.windows` and `map_.endpoint_words`, so one index names the window, the rows it resolves through, and the inverse of those rows.

tests/baremetal/test_att_resolver.cpp
pjanevskiTT: Done -- the PR now has seven passing-path tests alongside the two validation ones, covering resolution through each role, the package offset, and the selector coming from the table. Same change as the Copilot thread above.

device/api/umd/device/coordinates/att/att_resolver.hpp
pjanevskiTT: Keeping the comment as it stands -- it's already down to the contrast that matters, and the reader of this class needs to know why the coordinate has to be folded into the address here when it doesn't have to be on Wormhole or Blackhole.

One thing I could not confirm either way while looking at this: whether Quasar's TLB registers really do carry both a legacy x/y mode and an ATT mode. Our own tree is ambiguous -- `grendel_implementation.hpp` defines a full TLB surface for Grendel with x/y bit fields, wired into `architecture_tlbs.cpp`, but three of those constants are marked "copied from blackhole/wormhole, need to verify". If you have the spec reference for the two modes, that would settle whether that block is real or placeholder, which is a more useful thing to pin down than this comment.

PR #3445, opened 2026-09-14
[Base API] Build the SPI devices from DeviceFirmware and DeviceProtocol, not TTDevice

broskoTT:
### Issue
#2872

### Description
`SPITTDevice` held a `TTDevice*` and reached back through the facade for everything: the ARC core
coordinate, the firmware version, the board id, the telemetry reader, and the dump-buffer data
accesses. Every one of those is either served by the firmware component itself or is a plain
protocol access, so the SPI classes now take `(DeviceProtocol*, <arch> DeviceFirmware*)` and hold no
facade pointer at all. `SPITTDevice::create(TTDevice*)` keeps its signature -- it is the only entry
point clients use -- and resolves both from the device, narrowing the firmware inside the arch
switch that already established the architecture. That also retires the `dynamic_cast` each SPI
constructor used to do, and with it the null checks in `read()`/`write()` that the cast had made
unreachable.

### List of the changes
- `SPITTDevice` holds `DeviceProtocol* protocol_` instead of `TTDevice* device_`; the Wormhole and
  Blackhole constructors take the protocol plus their concrete firmware type.
- `SPITTDevice::create` resolves the protocol and narrows the firmware once per arch arm; the
  per-constructor `dynamic_cast` and the four now-unreachable "firmware not available" checks go.
- Facade hops replaced by the component that owns the data: `get_arc_core()` ->
  `get_firmware_noc_coord()`, `get_firmware_version()` / `get_board_id()` -> the firmware's info
  provider, `get_firmware_telemetry_reader()` -> the firmware's reader.
- The dump-buffer accesses move onto two private `read_from_arc`/`write_to_arc` helpers per SPI
  class, routing on the thread-selected NOC exactly as before.

Routing is unchanged: `TTDevice::read_from_device` resolved the `tt_xy_pair` SPI passed as a LITERAL
coordinate, which `resolve_coordinate` returns untouched, so the direct `DeviceProtocol::read_data`
is the same access.

### Testing
Builds clean with `TT_UMD_BUILD_ALL=ON` (simulation included, clang-tidy on); pre-commit clean. No test changes -- `tests/api/test_spi_tt_device.cpp`
and the nanobind binding both go through `SPITTDevice::create`, which is untouched.

### API Changes
This PR has no API changes. `SPITTDevice::create`, `read`, `write` and `get_spi_fw_bundle_version`
are unchanged, and the Wormhole/Blackhole SPI headers are not installed, so their constructor
signatures are internal.


device/arc/blackhole_spi_tt_device.cpp
```diff
@@ -144,24 +145,24 @@ std::optional<uint32_t> BlackholeSPITTDevice::extract_protobuf_uint32_field(
...
-BlackholeSPITTDevice::BlackholeSPITTDevice(TTDevice* tt_device) :
-    SPITTDevice(tt_device), firmware_(dynamic_cast<BlackholeDeviceFirmware*>(tt_device->get_device_firmware())) {
-    UMD_ASSERT(
-        firmware_ != nullptr,
-        error::RuntimeError,
-        "BlackholeSPITTDevice requires a device backed by BlackholeDeviceFirmware.");
+BlackholeSPITTDevice::BlackholeSPITTDevice(DeviceProtocol* protocol, BlackholeDeviceFirmware* firmware) :
+    SPITTDevice(protocol), firmware_(firmware) {}
+
+void BlackholeSPITTDevice::read_from_arc(void* dst, uint64_t addr, size_t size) {
```
nbuncicTT, line 151: Make the parameter explicit for NOC selection in the API, that way it's easier to figure out which NOC was used.

nbuncicTT (approved):

broskoTT:
I'll remove SPITTDevice altogether

PR #3449, opened 2026-09-15
Add test for TLB window behavior at the edge of its mapping

### LINKED ISSUE #1121: Read more than 16MB read and alignment issue
https://github.com/tenstorrent/tt-exalens/issues/141

broskoTT:
### Issue
Closes #1121

### Description
A read spanning the end of one TLB window used to return wrong data. The chunking in
io_window_reconfigure handles this today, and the reporter asked for a regression test to
keep it that way.

The test seeds both sides of a window boundary, then issues three writes through a window
based one mapping below it: only the one scoped to the last bytes of the mapping succeeds,
while one running past the end and one starting on the boundary are both rejected, as is the
equivalent read. A single straddling read through the reconfiguring path then confirms the
scoped write landed and that neither rejected transfer spilled past the mapping.

DRAM is the target because the boundary sits a whole mapping past the start of the core's
address space, beyond the end of Tensix L1.

### List of the changes
- Added TestTlb.TestTlbAccessAcrossWindowBoundary to tests/unified/test_tlb.cpp.

### Testing
unified_tests builds and links, pre-commit clean. Not yet run on silicon, needs a WH or BH
board. Expected to pass on current main, it is a regression guard rather than a fix.

### API Changes
This PR has no API changes.


tests/unified/test_tlb.cpp
```diff
@@ -469,6 +469,79 @@ TEST_F(TestTlb, TestTlbOffsetReadWrite) {
...
+// Checks how a TLB window behaves at the far edge of its mapping. The window can only satisfy a
+// transfer that fits entirely within it, so one starting in range but running past the end has to be
+// rejected without touching the device. Crossing the boundary is instead the reconfiguring path's
+// job, which splits the transfer and walks the window across; that is verified here too.
+//
+// DRAM is the target because the boundary sits a whole mapping past the start of the core's address
+// space, beyond the end of Tensix L1.
+TEST_F(TestTlb, TestTlbAccessAcrossWindowBoundary) {
+    const ChipId chip = 0;
+    const size_t requested_window_size = 1 << 21;
```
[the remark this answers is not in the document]
pjanevskiTT: I think we deprecated KMDs older than 2.0. That being sad, maybe we can remove the check from allocate_tlb call as well

pjanevskiTT (approved):

tests/unified/test_tlb.cpp
```diff
@@ -469,6 +469,79 @@ TEST_F(TestTlb, TestTlbOffsetReadWrite) {
...
+    // is filled with a value neither seed uses, so a transfer that partially landed shows up below.
+    std::vector<uint8_t> overrun(chunk_size * 2, 0xFF);
+    EXPECT_THROW(window.write_block(mapping_size - chunk_size, overrun.data(), overrun.size()), std::out_of_range);
+    EXPECT_THROW(window.write_block(mapping_size, overrun.data(), chunk_size), std::out_of_range);
+    EXPECT_THROW(window.read_block(mapping_size - chunk_size, overrun.data(), overrun.size()), std::out_of_range);
+
+    // Read the straddling span in one call, which only the reconfiguring path can serve.
+    std::vector<uint8_t> readback(chunk_size * 2, 0);
+    read_block_reconfigure(
+        window, readback.data(), dram_core, boundary_addr - chunk_size, readback.size(), NocId::NOC0);
```
pjanevskiTT, line 536: Every offset in the test is 32-byte aligned, so the straddle never reaches the unaligned path the issue was actually about. #1121 asks to "read something from (TLB_size - 4, 32) and (TLB_size, 32)", and the guess in the thread was "possibly an alignment issue". With `chunk_size = 32` both reconfigure chunks are a whole number of 32-bit words, so the leading/trailing sub-word handling in `SiliconTlbWindow::memcpy_from_device` is never exercised *at the boundary* — which is the combination that bit.

Starting this read at `boundary_addr - 4` (4-byte first chunk, 60-byte second) would cover both halves of the report rather than just the boundary-crossing one.

tests/unified/test_tlb.cpp
```diff
@@ -469,6 +469,79 @@ TEST_F(TestTlb, TestTlbOffsetReadWrite) {
...
+    SiliconTlbWindow tlb_window(pci_device->allocate_tlb(requested_window_size, TlbMapping::WC));
+    IoWindow& window = tlb_window;
+
+    TargetIoWindowConfig target;
+    target.core_start = dram_core;
+    target.addr = dram_base_addr;
+    target.noc = NocId::NOC0;
+    window.configure(target);
+
+    // The mapping can come back larger than requested, so the boundary comes from the window itself.
```
pjanevskiTT, line 500: Both claims in these two lines are wrong, and one of them costs the test an assertion.

"can come back larger than requested": `SiliconTlbHandle`'s ctor stores the requested size verbatim (`tlb_size_ = size`) and `TlbWindow::get_size()` returns `handle size - offset_from_aligned_addr`. `dram_base_addr` is 2MB-aligned so the offset is 0, and `mapping_size` is always `requested_window_size`. Taking the boundary from `get_size()` alone means the test stays self-consistent with whatever `get_size()` reports, even if that regressed — `ASSERT_EQ(mapping_size, requested_window_size);` pins it, the way `IoWindowInterface` does at line 630.

"never reconfigured again": `read_block_reconfigure` on line 535 reconfigures this very window twice — `transfer_and_reconfigure` calls `window.configure` once per chunk, which is the behavior the test exists to guard.

PR #3456, opened 2026-09-16
Cluster: close_device on Cluster destructor

ndjekicTT:
### Issue
#2520 

### Description
This PR ensures that close_device is called on Cluster destruction.

### List of the changes
- Add `close_device()` call into Cluster destructor with safe guards.
- Add `ClusterDestructorClosesDevice` test into `TestRiscProgram`. This tests confirms that tensix cores are no longer working after cluster is closed.

### Testing
Manually builds clean, pre-commit passes.
### API Changes
This PR has (no) API changes.


device/cluster.cpp
```diff
@@ -1296,6 +1306,7 @@ void Cluster::start_device(const DeviceParams& device_params) {
         }
 
         deassert_resets_and_set_clock_state();
+        needs_close_ = true;
```
ajovanovicTT: needs_close_ is only set after the whole start_device() loop succeeds. If get_chip(chip_id)->start_device(...) throws partway (e.g. LockManager::acquire_mutex or pin_or_map_sysmem_to_device failing on a PCIe hang), chips that already started are left with mutex held / sysmem pinned, but needs_close_ stays false, so the destructor never cleans them up.

Suggest setting needs_close_ = true before the loop instead of after. close_device() already iterates all chips unconditionally and doesn't assume start_device() finished, so calling it on a not yet started chip is safe I think.

ajovanovicTT (approved):

nbuncicTT (approved):

PR #3457, opened 2026-09-16
Add Python bindings for RobustMutex and KmdMutex

broskoTT:
### Issue
https://github.com/tenstorrent/tt-umd/issues/2745

### Description
Surfaces UMD's two cross-process locking backends to Python, so tools and tests written
against the bindings can take the same locks UMD takes, and probe who holds one.

MutexInterface is bound as the base, so the methods both backends share are defined once
and a caller can hold either backend as a lock. Both are usable as context managers, which
is what keeps a lock from leaking when a Python caller raises. Every call releases the GIL,
since lock() blocks.

### List of the changes
- New nanobind/py_api_mutex.cpp binding MutexInterface, RobustMutex and KmdMutex, with
  probe_lock() taking a timeout in seconds and returning None or the owning (pid, tid).
- One test per backend, doubling as usage examples; the KmdMutex one skips without a device.

### Testing
Ran the new pytest files locally against a built wheel; the RobustMutex ones pass, the
KmdMutex ones need a card and are covered by CI.

### API Changes
This PR has no API changes. It adds Python bindings for existing C++ API.


nanobind/py_api_mutex.cpp
```diff
@@ -0,0 +1,105 @@
...
+    // The base is bound so that the methods both backends share are defined once and so that a Python
+    // caller can treat either backend as a lock without knowing which one it holds.
+    nb::class_<MutexInterface>(
+        m,
+        "MutexInterface",
+        "Common interface of UMD's cross-process locks. The lock is not recursive, and misuse (taking it twice from "
+        "one thread, unlocking from a thread that does not hold it, unlocking twice) is not diagnosed: it may throw, "
+        "be ignored, or block forever. Use the object as a context manager to avoid all three.")
+        .def(
+            "initialize",
```
ajovanovicTT: KmdMutex::initialize() guards a second call, RobustMutex::initialize() doesn't - it re-mmaps
and leaks the old mapping. LockManager made that unreachable in C++, but here initialize() is
exposed directly. Worth documenting "call exactly once" here, or guarding it in RobustMutex.

nanobind/tests/test_py_mutex.py
```diff
@@ -0,0 +1,110 @@
...
+    def probe():
+        owner = mutex.probe_lock(timeout_s)
+        if owner is None:
+            mutex.unlock()
+        result.append(owner)
+
+    thread = threading.Thread(target=probe)
+    thread.start()
+    thread.join()
+    return result[0]
```
ajovanovicTT: nit - If probe_lock() raises in the thread, result is empty and this fails with IndexError instead
of the real error. Catch and re-raise in the caller?

ajovanovicTT (approved):

Issue #3450, opened 2026-09-15
Create Reset component


nbuncicTT:
**Description**
Add explanation later.

**Scope**
What specific tasks need to happen? List the concrete steps, including any validation or edge cases to cover. If the scope can't be clearly defined, treat the issue as an investigation — it's done when a new, well-scoped issue is created from the findings.

**Timeline**
Is the feature tied to a deadline, or is there a timeline otherwise which would increase the impact of the feature?

**Is your feature request related to a problem? Please describe.**

**Definition of Done**
List the conditions that must be met for this feature to be considered done.

PR #3464, opened 2026-09-17
[Base API] Introduce the RiscReset component interface

nbuncicTT:
### Issue
#3450

### Description
`set_risc_reset_state` assumes one soft-reset register per tile serves every RISC on it, which holds on Wormhole and Blackhole but not on Quasar: a Tensix NEO tile keeps its TRISCs and its NOC-overlay DM cores in different registers, and reset there may need more than a register write. `RiscReset` is the component that owns that behavior: a portable `RiscType` selection in, an effect on the device out, with how the effect is produced left to each implementation — one register, several, another interconnect, or a firmware service.

This PR adds the interface and the plumbing the transitions need, and nothing else: `TTDevice` keeps its current reset path for models that do not serve the component yet, so Wormhole/Blackhole (#silicon) and the simulation backends (#simulation) can move independently on top of this.

### List of the changes
- `RiscReset`: `assert_risc_reset` / `deassert_risc_reset` return `void` and throw on failure; `get_risc_reset_state` returns `std::optional<RiscType>`, where `std::nullopt` means the implementation cannot answer, as distinct from an attempt that failed.
- `TTDeviceModel::get_risc_reset()`, null until a model serves one.
- `TTDevice::assert_risc_reset` / `deassert_risc_reset` delegate to the component when the model serves one and otherwise keep today's read-modify-write. The fallback and the remaining virtuals go away in a follow-up, once every model serves the component.
- Both gain a `noc_id` parameter, defaulted, so the component is given the NOC the coordinate was resolved for.

### Testing
Builds with simulation ON and OFF; `baremetal_tests` 268/268; CI.

### API Changes
This PR has API changes: `TTDevice::assert_risc_reset` and `TTDevice::deassert_risc_reset` take an additional defaulted `noc_id`; the Python bindings expose it as an optional argument.


device/tt_device/tt_device.cpp
```diff
@@ -593,14 +594,29 @@ void TTDevice::set_risc_reset_state(CoreCoord core, const uint32_t risc_flags) {
...
 
-void TTDevice::assert_risc_reset(CoreCoord core, const RiscType selected_riscs) {
+void TTDevice::assert_risc_reset(CoreCoord core, const RiscType selected_riscs, NocId noc_id) {
+    RiscReset *risc_reset = model_->get_risc_reset();
+    if (risc_reset != nullptr) {
+        risc_reset->assert_risc_reset(resolve_coordinate(core, noc_id), selected_riscs, noc_id);
+        return;
+    }
+
+    // TODO: transitional - the path for models that do not serve a RiscReset yet. Deleted once they
```
broskoTT, line 604: Yes thanks for this kind of separation in PRs

device/tt_device/tt_device.cpp
```diff
@@ -593,14 +594,29 @@ void TTDevice::set_risc_reset_state(CoreCoord core, const uint32_t risc_flags) {
     tt_driver_atomics::sfence();
 }
 
-void TTDevice::assert_risc_reset(CoreCoord core, const RiscType selected_riscs) {
+void TTDevice::assert_risc_reset(CoreCoord core, const RiscType selected_riscs, NocId noc_id) {
+    RiscReset *risc_reset = model_->get_risc_reset();
+    if (risc_reset != nullptr) {
+        risc_reset->assert_risc_reset(resolve_coordinate(core, noc_id), selected_riscs, noc_id);
```
broskoTT, line 600: I might've missed this detail about "resolve_coordinate":
This is the function that we want to use in general on the transition from CoreCoord to tt_xy_pair?

broskoTT (approved):
Nit: this PR could be split into 
1. Adding NOCid to current API
2. everything else (new component + wiring in tt_device_model + calling from tt_device).

Not needed now since I already looked at it

device/tt_device/tt_device.cpp
nbuncicTT: Yes

PR #3465, opened 2026-09-17
[Base API] Delegate Wormhole and Blackhole RISC reset to the RiscReset component

nbuncicTT:
### Issue
#3450

### Description
Wormhole and Blackhole keep every RISC of a tile in one soft-reset register, so they are served by one implementation rather than one per architecture: the register address and the `RiscType`-to-bits mapping are data the `ArchitectureImplementation` already provides, and the operation is a read-modify-write of that register over the device protocol. `TTDevice`'s own reset path stops running for these devices, since their models now serve the component (#3464 keeps it for the backends that have not moved yet).

### List of the changes
- `RiscResetImplementation`: read-modify-write over `DeviceProtocol`, routed on the NOC it is given, with the staggered-start bit applied on deassert exactly as the facade did.
- `WormholeTTDeviceModel` and `BlackholeTTDeviceModel` build it in every constructor and serve it from `get_risc_reset()`.
- New baremetal test `test_risc_reset.cpp`: asserting adds the selected bits and keeps the others, deasserting clears only the selected ones, staggered start sets its bit, the state read reports the RISCs in reset, and every access carries the NOC it was given. Run against both architectures against `MockDeviceProtocol`.

Behavior is unchanged: the same register, the same bits, the same order of accesses.

### Testing
Builds with simulation ON and OFF; `baremetal_tests` 280/280, including the 12 new cases; CI.

### API Changes
This PR has no API changes.


device/api/umd/device/tt_device/reset/risc_reset_implementation.hpp
```diff
@@ -0,0 +1,39 @@
...
+class DeviceProtocol;
+
+/**
+ * @brief RiscReset for architectures with one soft-reset register per tile: each operation is a
+ * read-modify-write of that register over the device protocol, with the register address and the
+ * RiscType-to-bits mapping served by the ArchitectureImplementation.
+ *
+ * The interfaces are non-owning and must outlive this object.
+ */
+class RiscResetImplementation : public RiscReset {
```
broskoTT, line 20: As we discussed offline, not sure of this naming

device/tt_device/reset/risc_reset_implementation.cpp
```diff
@@ -0,0 +1,51 @@
...
+        error::RuntimeError,
+        "RiscResetImplementation requires a DeviceProtocol and an ArchitectureImplementation.");
+}
+
+void RiscResetImplementation::assert_risc_reset(tt_xy_pair core, RiscType selected_riscs, NocId noc_id) {
+    const uint32_t current_state = read_reset_register(core, noc_id);
+    write_reset_register(core, current_state | architecture_impl_->get_soft_reset_reg_value(selected_riscs), noc_id);
+}
+
+void RiscResetImplementation::deassert_risc_reset(
```
broskoTT, line 28: Can you also remove it now from tt_device?

tests/baremetal/test_risc_reset.cpp
```diff
@@ -0,0 +1,122 @@
...
+    uint32_t written = 0;
+    EXPECT_CALL(protocol_, read_ctrl(_, _, _, _, _)).WillOnce(serves_register(bits_for(RiscType::BRISC)));
+    EXPECT_CALL(protocol_, write_ctrl(_, _, _, _, _)).WillOnce(captures_register(&written));
+
+    risc_reset_.assert_risc_reset(TENSIX_CORE, RiscType::ALL_TENSIX_TRISCS, NocId::NOC0);
+
+    EXPECT_EQ(written, bits_for(RiscType::BRISC) | bits_for(RiscType::ALL_TENSIX_TRISCS));
+}
+
+TEST_P(RiscResetImplementationTest, DeassertClearsOnlyTheSelectedBits) {
```
broskoTT, line 78: Looking at these tests (not just this one), I do like the last EXPECT in each of those as it actually tests the expected vs real effects. That is the essence of the component


I don't like the parts of the tests that count how many times a specific function was called underneath. I'd remove all these EXPECT_CALL.
I think that serves more fixing the internal implementation rather than testing for regressions, i.e. if you removed those checks, and SOMEHOW SOMEONE skipped calling "write" internally, the test would fail anyway. 
The only way it could guard something is to guard against somebody calling this function two times instead of once, but I don't think that's a realistic scenario for a bug. And it might not even be a bug. Imagine theoretically we decide (due to hardware intrinsics) that is the right way to do reset. In that case, that change is intentional, and you have to change the test, and I'm saying that is the proof that part of the test is meddling in internal component details rather than testing the effects, and my argument is that it should not.

Since you gave me feedback I should not be soft on my beliefs, I have to insist :) However, if you strongly believe opposite of what I said, let's have that discussion some other time and lets not block this PR.

device/tt_device/reset/risc_reset_implementation.cpp
```diff
@@ -0,0 +1,51 @@
...
+    tt_xy_pair core, RiscType selected_riscs, bool staggered_start, NocId noc_id) {
+    const uint32_t current_state = read_reset_register(core, noc_id);
+    uint32_t new_state = current_state & ~architecture_impl_->get_soft_reset_reg_value(selected_riscs);
```
broskoTT, line 31:
```suggestion
    // The argument selected_riscs tells us which bits should be pulled back to 0, hence the negation. 
    uint32_t new_state = current_state & ~architecture_impl_->get_soft_reset_reg_value(selected_riscs);
```

device/api/umd/device/tt_device/reset/risc_reset_implementation.hpp
broskoTT: Glean suggested ClassicRiscReset vs NeoRiscReset.
It also suggested using TileRiscReset instead, so ClassicTileRiscReset vs NeoRiscReset, but not sure that is necessary

broskoTT (approved):

device/api/umd/device/tt_device/reset/risc_reset_implementation.hpp
nbuncicTT: Having given it another though, I actually much more like the concept of TileRiscReset, so I'll rename the interface after everything gets in. 

Because of that, I'll name the new classes as `ClassicTileRiscReset` and `NeoTileRiscReset`.

Thanks for the suggestion.

tests/baremetal/test_risc_reset.cpp
nbuncicTT: So there are two parts to your comment which I will address, the first one is tied to interface testing and the other one is tied to implementation testing - and both should be done.

In interface testing, we testing the expected inputs against the expected outputs, not caring what the implementation internally does, and that's the correct way to do it and there you are completely right, we shouldn't ever expect a function call to some component because the interface allows this degree of freedom.

However, in implementation testing, there is no room for improvisation, the hardware or IP implementation works exactly how it was designed (if it doesn't have a bug or something else). Because of that, we can exactly expect what should happen, i.e. the test is the "recipe" how to do that particular implementation.

For `ClassicTileRiscReset` this is the case - we write the "recipe" and this ensures that somebody can't change the implementation without failing the test.

The failing test has a very big value here - If somebody needs to modify the implementation which was prescribed by hardware design, there are a few possibilities:
- there's a bug in hardware
- there's more ways to do it, which means we should update the hardware design spec and make the test more broad
- somebody wrongfully changed these lines of code and we found a genuine software bug introduction

Lastly, the implementation test allows for some degree of freedom in the implementation (i.e. this should "just" define the hardware interaction, whereas other things can be defined separately).

Let's have a short discussion on this topic today and I'll do another round of changes then.

PR #3466, opened 2026-09-17
[Base API] Delegate simulation RISC reset to the RiscReset component

nbuncicTT:
### Issue
#3450

### Description
The two simulation backends reset differently from silicon and from each other — TTSim writes the soft-reset register, inverted and 64 bits wide on Quasar, while the RTL simulator takes reset as its own commands with separate ones for Quasar's NEO, DM and uncore groups — which is why both overrode `TTDevice`. Each becomes a `RiscReset` implementation instead, so the facade no longer has simulation-specific reset to dispatch to. The behavior of both is moved as it stands.

### List of the changes
- `TTSimRiscReset`: the read-modify-write TTSim had, with Quasar's inverted polarity as a constructor flag rather than a PCI-id check inside the operation. The accesses still go through the device, which is what dispatches between a simulator in this process and one reached over a socket.
- `RtlSimRiscReset`: the communicator's reset commands, with the Quasar group handling moved verbatim. A client-mode device has no communicator and installs it with none, so reset there reports itself unwired instead of dereferencing it.
- `SimulationTTDeviceModel` serves the component the backend hands it; `SimulationTTDevice` keeps a typed handle on the model it was built with, taken before the base takes ownership.
- The four `assert_risc_reset` / `deassert_risc_reset` overrides on the simulation devices are deleted.

### Testing
Builds with simulation ON and OFF; `baremetal_tests` 268/268; CI covers the simulation paths.

### API Changes
This PR has no API changes.


device/tt_device/reset/tt_sim_risc_reset.cpp
```diff
@@ -0,0 +1,71 @@
...
+}
+
+void TTSimRiscReset::update_reset_register(tt_xy_pair core, RiscType selected_riscs, bool set_bits) {
+    std::lock_guard<std::recursive_mutex> lock(device_lock_);
+    const uint64_t soft_reset_addr = architecture_impl_->get_tensix_soft_reset_addr();
+    const uint32_t soft_reset_update = architecture_impl_->get_soft_reset_reg_value(selected_riscs);
+    // The coordinate arrives translated, so it is handed on as LITERAL.
+    const CoreCoord translated_core{core.x, core.y};
+
+    // Quasar holds its DM cores in reset with the bit cleared rather than set, and its register is
```
broskoTT, line 37: wow

device/tt_device/reset/tt_sim_risc_reset.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+#include <tt-logger/tt-logger.hpp>
+
+#include "umd/device/arch/architecture_implementation.hpp"
+#include "umd/device/tt_device/tt_device.hpp"
+#include "umd/device/utils/error.hpp"
+
+namespace tt::umd {
+
+TTSimRiscReset::TTSimRiscReset(
```
broskoTT, line 15: This is possibly just NeoRiscReset, right? I mean it probably works the same on silicon? @pjanevskiTT for input

device/tt_device/tt_sim_tt_device.cpp
```diff
@@ -295,46 +310,6 @@ bool TTSimTTDevice::special_dram_read(void* mem_ptr, tt_xy_pair core, uint64_t a
...
-    uint64_t soft_reset_addr = get_architecture_implementation()->get_tensix_soft_reset_addr();
-    uint32_t soft_reset_update = get_architecture_implementation()->get_soft_reset_reg_value(selected_riscs);
-
-    if (libttsim_pci_device_id == 0xFEED) {  // QSR
-        uint64_t reset_value;
-        read_from_device(&reset_value, core, soft_reset_addr, sizeof(reset_value));
-        reset_value |=
-            (uint64_t)soft_reset_update;  // QSR logic is reversed for DM cores, so we need to invert the update.
-        write_to_device(&reset_value, core, soft_reset_addr, sizeof(reset_value));
-    } else {
```
broskoTT, line 330: It seems like you completely remove "assert_risc_reset" , so this now fallsback to base tt_device's implementation?
Which is fine.
But in that case, I think you want to create TTSimTTRiscReset in case of Grendel, but regular ClassicRiscReset in case of wh bh, right?
To be specific, is this else branch still honoured?

Issue #3458, opened 2026-09-16
WarmReset sends POST_RESET too early when hotplug is suppressed


joelsmithTT:
**Describe the bug**
`WarmReset::warm_reset_arch_agnostic()` does not wait for the chip to finish resetting before sending `POST_RESET`. It worked by accident while hotplug re-probed the device; with tt-kmd >= 2.8.0 suppressing hotplug on Galaxy, a subset reset under `TT_VISIBLE_DEVICES` leaves the chips unusable.

**Environment**
- UMD: 0.9.5 (tt-smi 5.2.0)
- FW version: 19.11.0
- KMD version: 2.8.0
- Board/system type: 6U BH Galaxy

**Logs**
```
TT_VISIBLE_DEVICES="0000:03:00.0,0000:04:00.0,0000:07:00.0,0000:08:00.0"   tt-smi -r
 CPLD FW v1.16 or higher is required to use tt-smi -r on Galaxy systems. If tt-smi -r fails, please continue to use tt-smi -glx_reset instead and contact your system administrator to request a CPLD update.
Resetting all PCI devices: [2, 3, 6, 7]
2026-09-12 06:00:36.692 | info     |             UMD | Notifying all listeners of impending warm reset. (warm_reset.cpp:66)
2026-09-12 06:00:36.736 | info     |             UMD | Starting reset on devices: (PCI index: 2, UMD logical ID: 0, BDF: 0000:03:00.0), (PCI index: 3, UMD logical ID: 1, BDF: 0000:04:00.0), (PCI index: 6, UMD logical ID: 2, BDF: 0000:07:00.0), (PCI index: 7, UMD logical ID: 3, BDF: 0000:08:00.0) (warm_reset.cpp:205)
2026-09-12 06:00:38.780 | info     |             UMD | Notifying all listeners of completed warm reset. (warm_reset.cpp:95)
 Re-initializing boards after reset....
2026-09-12 06:00:38.802 | info     |             UMD | Creating TopologyDiscovery for architecture: blackhole (topology_discovery.cpp:75)
2026-09-12 06:00:38.802 | info     |             UMD | Starting topology discovery. (topology_discovery.cpp:92)
2026-09-12 06:00:38.824 | warning  |             UMD | Low power mode is disabled while UMD holds open file descriptors. The device will return to low power mode once all file descriptors are closed. (topology_discovery.cpp:140)
[...]

Issue #2345, opened 2026-03-20
IPMI reset on Galaxy is missing required USER_RESET and POST_RESET ioctls


joelsmithTT:
## Summary

All IPMI-based Galaxy reset paths (`-glx_reset`, `-glx_reset_auto`, `-glx_reset_tray`) do not issue `USER_RESET` or `POST_RESET` ioctls to the kernel driver. They invoke `ipmitool` and wait for device nodes to reappear. That's it.

This has appeared to work because `POST_RESET` is a no-op when the device has been hotplugged — hotplug runs the driver's remove and probe handlers, hardware is reinitialized, and there is nothing for `POST_RESET` to do. Galaxy reset typically triggers hotplug, so the missing ioctls were never noticed.

Hotplug on Galaxy causes several problems (device ordinal shuffling, PCIe Gen1 fallback from a Linux quirk, VM device escape under PCI passthrough). The next tt-kmd release will suppress hotplug on Galaxy.  With hotplug suppressed, the device persists across the IPMI reset.  `POST_RESET` is then required to restore PCI config space and reinitialize hardware. Without it the device is left broken.

## Required IPMI reset sequence

1. `USER_RESET` ioctl on every device being reset (before ipmitool).
2. `ipmitool raw 0x30 0x8B ...`
3. Poll for reset completion (reset marker clears in PCI config space).
4. `POST_RESET` ioctl on every device.

No repo does steps 1, 3, or 4 today.

Additionally, `tt-smi -glx_reset_tray N` is broken on the default (UMD) path — it hard-exits because `ubb_warm_reset()` only accepts the "reset everything" parameters. It only works with `--use_luwen`.

## What needs to change

Wrap the ipmitool call with `USER_RESET` before and `POST_RESET` after. The ioctl is `TENSTORRENT_IOCTL_RESET_DEVICE` with:

- `flags = 3` (`USER_RESET`) — before ipmitool
- `flags = 6` (`POST_RESET`) — after reset completion

## Urgency

Release blocker. The next tt-kmd release suppresses hotplug on Galaxy. Current IPMI reset paths will leave devices non-functional.

PR #3467, opened 2026-09-17
Polling for reset marker

### LINKED ISSUE #2853: Poll reset marker in WarmReset
Follow up to Joel's comment: https://github.com/tenstorrent/tt-umd/pull/2844#discussion_r3434406957

Instead of calculating heuristic post-reset wait time, poll the reset marker to determine when reset is complete.

aleksamarkovicTT:
### Issue
#3458
Closes #2853

### Description
Poll for reset marker on WarmReset.

### List of the changes
- Introduce polling for reset marker in WarmReset.
- Pipe reset ioctl result to return value of tt_device_reset() in kmdlib.

### Testing
CI, test on g08blx07.

### API Changes
This PR has (no) API changes.


device/warm_reset.cpp
```diff
@@ -463,7 +502,25 @@ bool WarmReset::ubb_warm_reset(const std::chrono::milliseconds timeout_ms) {
...
     ubb_wait_for_driver_load(timeout_ms);
-    return reset_success;
+
+    if (!reset_success) {
+        return false;
+    }
+
+    if (!std::all_of(
+            PCIDevice::enumerate_devices_info().begin(),
+            PCIDevice::enumerate_devices_info().end(),
```
joelsmithTT: Does this work?  I think it's UB.

```
#include <algorithm>
#include <iostream>
#include <map>
#include <string>
std::map<int, std::string> make() { return {{0, "a"}, {1, "b"}, {2, "c"}}; }
int main() {
  int visited = 0;
  bool r = std::all_of(make().begin(), make().end(), [&](const auto &p) {
    ++visited;
    if (visited > 10)
      std::terminate();
    return true;
  });
  std::cout << "visited=" << visited << std::endl;
}
```

device/pcie/pci_device.cpp
```diff
@@ -216,11 +216,12 @@ static void send_reset_ioctl(int device_id, uint32_t flags) {
     }
 
     err = tt_device_reset(dev, flags);
+    tt_device_close(dev);
```
joelsmithTT, line 219: tt_kmd_lib.h says, "The device handle will be invalidated after a successful reset" but it looks like that invalidation is happening here as opposed to within `tt_device_reset`.  Should the comment in the header be updated?

device/warm_reset.cpp
```diff
@@ -463,7 +502,25 @@ bool WarmReset::ubb_warm_reset(const std::chrono::milliseconds timeout_ms) {
...
+                if (!wait_for_reset_marker(pci_device_info.second.pci_bdf)) {
+                    log_error(tt::LogUMD, "Reset failed.");
+                    return false;
+                }
+                return true;
+            })) {
+        return false;
+    }
+
+    return true;
```
joelsmithTT: I ask that this hunk be either dropped from the PR or expanded into a full fix.  As written it can't work and makes the function look covered when it isn't.  Since the PR touches this path, I want to lay out my concerns with `ubb_warm_reset()` in one place.

Introduced by this PR:
* Iterator issue (see previous comment).
* The poll is polling for a marker that isn't getting set.

Pre-existing:
* No USER_RESET before ipmitool, no POST_RESET after.  https://github.com/tenstorrent/tt-umd/issues/2345
* `reset_success` does not indicate reset success; nothing in the function does.  `wormhole_ubb_ipmi_reset()` also logs "Reset successfully completed" before the 30s sleep.
* `ubb_wait_for_driver_load()` counts nodes until it sees 32.  With hotplug suppressed the nodes never go away, so it returns immediately (similar to https://github.com/tenstorrent/tt-umd/issues/3458's `wait_for_pci_bdf_to_reappear`).  With `TT_VISIBLE_DEVICES` set, `enumerate_devices()` may never reach 32.  I haven't verified it, but it looks like it'll spin for 100s. `TT_VISIBLE_DEVICES` is not compatible with IPMI-based reset (`ubb_warm_reset()` / the ipmitool command can't respect the env var), which may be surprising to users.
* sleep(30) vs unused UBB_POST_RESET_WAIT of 30,000ms in the header.  But once a marker is set (USER_RESET), the poll should replace this sleep.
* `wormhole_ubb_ipmi_reset()` is not Wormhole specific.

What I suggest for this PR is either to revert the `ubb_warm_reset` changes (the reset path UMD names "arch-agnostic" has the fix for the reported bug and can stand on its own) or replace `ubb_warm_reset` with the following sequence:

if TT_VISIBLE_DEVICES set, fail // ipmi reset will hit them all
send_reset_ioctl_to_devices(all, USER_RESET)
ipmitool raw 0x30 0x8b 0xf 0xff 0x0 0xf
for each device: wait_for_reset_marker(bdf) // replaces sleep(30)
sleep(100ms) // some time for CMFW to come up (it appears possible to issue POST_RESET too early after PCIe link up; this is a driver defect)
send_reset_ioctl_to_devices(all, POST_RESET, ignore_failures=false)

device/warm_reset.cpp
aleksamarkovicTT: It's some stupid readability check that forced this change. I will rework it to a for loop.

device/pcie/pci_device.cpp
aleksamarkovicTT: Updated message.

device/warm_reset.cpp
aleksamarkovicTT: I agree with all your remarks, best to do it in another PR.

device/warm_reset.cpp
```diff
@@ -179,6 +180,39 @@ int wait_for_pci_bdf_to_reappear(
...
 }
 
+bool wait_for_reset_marker(
+    const std::string& bdf, const std::chrono::milliseconds timeout_ms = timeout::WARM_RESET_DEVICES_REAPPEAR_TIMEOUT) {
+    log_debug(tt::LogUMD, "Waiting for reset marker to clear for device {}.", bdf);
+
+    auto deadline = std::chrono::steady_clock::now() + timeout_ms;
+    bool reset_complete = false;
+
+    while (std::chrono::steady_clock::now() < deadline) {
```
nbuncicTT: Can't we use the `poll_until` or `check_timeout` mechanism here?

device/warm_reset.cpp
```diff
@@ -179,6 +180,39 @@ int wait_for_pci_bdf_to_reappear(
...
+        // Check in-place reset marker
+        const std::string config_path = fmt::format("/sys/bus/pci/devices/{}/config", bdf);
+        int cfd = open(config_path.c_str(), O_RDONLY);
+        if (cfd >= 0) {
+            uint8_t cmd;
+            if (pread(cfd, &cmd, 1, 4) == 1 && ((cmd >> 6) & 1) == 0) {
```
nbuncicTT: The parameters here are very vague, we have magic numbers, three letter names (I'm fine with `cmd`, `cfd` is vague, I get that it's a file descriptor, I'm just wondering what are we reading).

Can you make this a bit more clear?

device/warm_reset.cpp
```diff
@@ -232,6 +266,11 @@ bool WarmReset::warm_reset_arch_agnostic(
             log_error(tt::LogUMD, "Reset failed.");
             return false;
         }
+
+        if (!wait_for_reset_marker(pci_bdf.second)) {
```
nbuncicTT, line 275: If we are introducing a polling mechanism, why do we need to both wait and poll? I'd omit the wait if the poll returns the info is the device is ready.

device/warm_reset.cpp
```diff
@@ -179,6 +180,39 @@ int wait_for_pci_bdf_to_reappear(
     return interface_id;
 }
 
+bool wait_for_reset_marker(
```
nbuncicTT, line 204: Please utilize this also in the `ubb_warm_reset`, as this reset is more problematic (we found this issue when using the reset on BH galaxies, not on the cards).

device/warm_reset.cpp
```diff
@@ -463,6 +502,7 @@ bool WarmReset::ubb_warm_reset(const std::chrono::milliseconds timeout_ms) {
     sleep(30);
     log_debug(tt::LogUMD, "30 seconds elapsed after reset execution.");
     ubb_wait_for_driver_load(timeout_ms);
+
```
nbuncicTT, line 510: nit: As to the boy-scout rule, could you rename `wormhole_ubb_ipmi_reset` to `galaxy_ubb_ipmi_reset`?

device/warm_reset.cpp
aleksamarkovicTT: This will be done in another PR.

nbuncicTT (approved):

PR #3477, opened 2026-09-20
Index a std::array by an enum class instead of a cast

nbuncicTT:
### Issue

None filed. Came out of reviewing #3442.

### Description

An array sized by an enum's `COUNT` sentinel is indexed today by casting the enumerator to `size_t` at every use. The cast accepts any integer, any other enum, and an enumerator outside the array, and the extent can drift from the enum without anything saying so.

`EnumArray<T, Enum>` takes `Enum::COUNT` as its extent and accepts nothing but `Enum` as an index. It composes `std::array` rather than inheriting from it, so `operator[](size_t)` stays unreachable.

A C++17 trait can only check that the index is an enum class with a `COUNT` enumerator. Contiguity from 0 stays a contract on the enum's author, checked with a `static_assert` in the class body rather than `enable_if`, which a class has no overload set to prune and which a caller can bypass by naming the defaulted parameter. `at()` covers the gap: `CoreType` declares `UNSPECIFIED` after `COUNT`, and `at()` throws for it.

`to_underlying` comes along for the cases that want the number, such as bit operations, ARC message ids and logging. It carries the C++23 signature so it can be dropped for `std::to_underlying` later.

No call site is converted here, so conversions can be reviewed as conversions.

### List of the changes

- `device/common/enum_array.hpp`: `EnumArray<T, Enum>`, the `is_enum_array_index` trait, and `to_underlying`. Private header, not part of the public API.
- `tests/misc/test_enum_array.cpp`: unit tests, registered in `umd_misc_tests`.

### Testing

Ten gtest cases plus compile-time checks, covering the trait's accept and reject set, extent following the enum, element-wise construction, copy and move not being hijacked by the variadic constructor, `constexpr` use, value and key iteration, `at()` throwing past `COUNT`, and a case over the real `CoreType`. Built and run with gcc 12 and clang 20 under `-Wall -Wextra -Wpedantic -Werror`, both green. `clang-tidy-20` with the repo config is clean on both files.

### API Changes

This PR has no API changes. The header is private to the library.

### Note for reviewers

Draft on purpose: a utility with no caller yet. The first conversion candidate is the ATT map in #3441 and #3442, whose `MapData` holds two `std::array<..., WINDOW_CLASS_COUNT>` and whose resolver casts at every index. That conversion would want the sentinel there spelled `COUNT` rather than `Count`, matching `CoreType`.


device/common/enum_array.hpp
```diff
@@ -0,0 +1,151 @@
...
+//
+// Composes std::array rather than inheriting from it, so std::array's operator[](size_t) is not part of
+// the interface. Storage is value-initialised, so a default-constructed EnumArray<int, E> holds zeros.
+template <typename T, typename Enum>
+class EnumArray {
+    static_assert(is_enum_array_index_v<Enum>, "EnumArray index must be an enum class with a COUNT enumerator.");
+    static_assert(to_underlying(Enum::COUNT) > 0, "EnumArray index enum has nothing before COUNT.");
+
+public:
+    using value_type = T;
```
moklobdzijaTT: I don't see `value_type` used in the class. Guess it should substitute `T` uses.

device/common/enum_array.hpp
```diff
@@ -0,0 +1,151 @@
...
+
+// C++17 stand-in for std::is_scoped_enum (C++23). An unscoped enum converts implicitly to its underlying
+// type, a scoped one does not. The bool parameter keeps underlying_type_t from being named for non-enums.
+template <typename E, bool = std::is_enum_v<E>>
+struct is_scoped_enum : std::false_type {};
+
+template <typename E>
+struct is_scoped_enum<E, true> : std::bool_constant<!std::is_convertible_v<E, std::underlying_type_t<E>>> {};
+
+}  // namespace detail
```
moklobdzijaTT: nit: Might be good to group these stand-ins in a way as they are not a part of `EnumArray` per se. Discussed it offline w. @nbuncicTT if they should be moved in a separate file, but for now I think it might be enough to hint the grouping [through Doxygen](https://www.doxygen.nl/manual/grouping.html#memgroup).

```suggestion
/** Utilities from C++23.
 */
///@{

// std::to_underlying from C++23, for the places that need an enum as a number: bit operations, message ids,
// logging, serialisation. Not for indexing, that is what EnumArray is for. Non-enums fail to substitute
// the return type, so there is no overload to call.
template <typename Enum>
constexpr std::underlying_type_t<Enum> to_underlying(Enum value) noexcept {
    return static_cast<std::underlying_type_t<Enum>>(value);
}

namespace detail {

// C++17 stand-in for std::is_scoped_enum (C++23). An unscoped enum converts implicitly to its underlying
// type, a scoped one does not. The bool parameter keeps underlying_type_t from being named for non-enums.
template <typename E, bool = std::is_enum_v<E>>
struct is_scoped_enum : std::false_type {};

template <typename E>
struct is_scoped_enum<E, true> : std::bool_constant<!std::is_convertible_v<E, std::underlying_type_t<E>>> {};

}  // namespace detail

///@}
```

device/common/enum_array.hpp
```diff
@@ -0,0 +1,151 @@
...
+
+namespace detail {
+
+// C++17 stand-in for std::is_scoped_enum (C++23). An unscoped enum converts implicitly to its underlying
+// type, a scoped one does not. The bool parameter keeps underlying_type_t from being named for non-enums.
+template <typename E, bool = std::is_enum_v<E>>
+struct is_scoped_enum : std::false_type {};
+
+template <typename E>
+struct is_scoped_enum<E, true> : std::bool_constant<!std::is_convertible_v<E, std::underlying_type_t<E>>> {};
```
moklobdzijaTT: nit: Similar to the general stand-ins improvement suggestion, think this comment could be handled as a group one to avoid confusion what is its scope -

```suggestion
/** C++17 stand-in for std::is_scoped_enum (C++23). An unscoped enum converts implicitly to its underlying
 * type, a scoped one does not. The bool parameter keeps underlying_type_t from being named for non-enums.
 **/
///@{
template <typename E, bool = std::is_enum_v<E>>
struct is_scoped_enum : std::false_type {};

template <typename E>
struct is_scoped_enum<E, true> : std::bool_constant<!std::is_convertible_v<E, std::underlying_type_t<E>>> {};
///@}
```

moklobdzijaTT (approved):

PR #3479, opened 2026-09-21
Probe KMD locks without taking them

broskoTT:
### Description
Probing a KMD lock acquired it to find out it was free, so anything inspecting a lock briefly took it
from whoever wanted it next. KMD can report the state directly through LOCK_CTL_TEST, which is safe to
call at any time, so probing costs the lock nothing now.

That makes probing mean two different things per backend, so probe_lock is a query on the whole
interface: it reports whether the lock is held and never leaves it held. RobustMutex still has to take
a pthread mutex to find out it was free, so it gives it straight back; the acquire-and-keep path lock()
needs moved to a private helper.

KMD tells us whether the lock is held by our own handle or by another one, so a lock this process holds
is now reported as {pid, 0} instead of the {0, 0} that means "held, owner unknown". There is no tid to
report either way: KMD tracks handles, not threads.

### List of the changes
- KmdMutex::probe_lock reads the lock state instead of acquiring it
- probe_lock never leaves the lock held, in the interface contract and in both implementations
- Drop the release-after-probe from LockManager and from the tests that did it by hand

### Testing
umd_misc_tests and the Python mutex tests. Both need a card for the KMD half.

### API Changes
This PR has API changes, probe_lock no longer leaves the lock held on success.


device/utils/lock_manager.cpp
ppetrovicTT: currently in `tt-telemetry` i'm also hardcoding this value (20).
is that considered a safe practice? or should we expose this value through a public getter?

device/api/umd/device/utils/robust_mutex.hpp
```diff
@@ -108,6 +108,10 @@ class RobustMutex : public MutexInterface {
     // Sets owner TID/PID to the calling thread and annotates for TSAN.
     void record_acquisition();
 
+    // Tries to take the lock, keeping it on success. Shared by lock() and probe_lock(), which differ only in what
```
nbuncicTT: Why is this private?

device/api/umd/device/utils/robust_mutex.hpp
```diff
@@ -57,11 +57,11 @@ class RobustMutex : public MutexInterface {
     // Unlocks the mutex.
     void unlock() override;
 
-    // Attempts to acquire the lock and returns immediately if timeout is zero (default), or waits
-    // up to `timeout` seconds before giving up.
-    // Returns std::nullopt if the lock was acquired successfully.
-    // Returns {owner_pid, owner_tid} if the lock is held by another thread/process (EBUSY/ETIMEDOUT).
-    // On EOWNERDEAD the dead owner's lock is recovered, the mutex is acquired, and nullopt is returned.
+    // Reports whether the lock is held, returning immediately if timeout is zero or waiting up to `timeout` seconds
+    // for it to become free. Returns {owner_pid, owner_tid} while it is held by another thread or process, and
```
nbuncicTT: use an alias (`using` keyword in C++) to emphasize that its `owner_pid` and `owner_tid` in the return value itself so the user doesn't guess when locking at the code.

device/api/umd/device/utils/robust_mutex.hpp
```diff
@@ -57,11 +57,11 @@ class RobustMutex : public MutexInterface {
...
+    // std::nullopt once it is not, having taken and released it to find that out.
```
nbuncicTT: The return value doesn't make sense - `std::nullopt` should mean that the lock isn't valid (i.e. initialize wasn't called on it, etc.).

For the lock not being taken, you could use a sentinel value of -1 on both the PID/TID.

I see you throw in the code if it's not initialized, but I think that conflates things.

nbuncicTT: And this is also a reason I want to introduce tl::expected, so we can have different error values return (not initialized, lock taken, etc.).

device/utils/kmd_mutex.cpp
```diff
@@ -145,25 +146,46 @@ void KmdMutex::lock() {
...
 std::optional<std::pair<pid_t, pid_t>> KmdMutex::probe_lock(std::chrono::seconds timeout) {
-    // KMD only offers non-blocking acquire and blocking acquire (no native timed acquire), so a
-    // bounded wait is implemented by polling the non-blocking acquire. On success we hold the lock and
-    // return nullopt.
-    if (try_lock()) {
-        return std::nullopt;
-    }
-    if (timeout.count() == 0) {
-        // Owner is unknown - KMD does not expose which handle/process holds the lock.
+    UMD_ASSERT(device_ != nullptr, error::RuntimeError, "KmdMutex::probe_lock() called before initialize()");
```
nbuncicTT: should be `nullopt`

device/utils/kmd_mutex.cpp
```diff
@@ -145,25 +146,46 @@ void KmdMutex::lock() {
...
+
+    // KMD can report a lock's state without taking it, so probing here costs the lock nothing: a free lock stays free
+    // and whoever holds one keeps it.
+    auto read_state = [this]() {
```
nbuncicTT: nit: `read_state` is very vague, `test_lock` should be more straightforward.

device/utils/kmd_mutex.cpp
```diff
@@ -145,25 +146,46 @@ void KmdMutex::lock() {
...
+        uint32_t state = 0;
+        int result = tt_lock_test(device_, lock_index_, &state);
+        UMD_ASSERT(
+            result == 0,
+            error::RuntimeError,
+            fmt::format("tt_lock_test() failed for lock {} on {} errno: {}", lock_index_, device_path_, -result));
+        return state;
+    };
+
+    // Held by this handle means held by this process, which is as precise as KMD gets: it tracks which handle holds a
```
nbuncicTT: "Held by this handle" - which handle? It tracks which process holds the lock, but not the thread, which is fine, I'm just wondering, why are you writing handle in the comment when it's clear it's a process (and it's ID is the pid).

device/utils/kmd_mutex.cpp
```diff
@@ -145,25 +146,46 @@ void KmdMutex::lock() {
...
+    if (!owner.has_value() || timeout.count() == 0) {
+        return owner;
     }
 
     const auto deadline = std::chrono::steady_clock::now() + timeout;
     while (std::chrono::steady_clock::now() < deadline) {
         std::this_thread::sleep_for(std::chrono::milliseconds(1));
-        if (try_lock()) {
-            return std::nullopt;
+        owner = owner_of(read_state());
```
nbuncicTT: We have a utility `check_until` which can be used here to clarify the code and keep it simpler, that way you would pass the lambda once and it would behave the same it behaves here.

device/utils/kmd_mutex.cpp
```diff
@@ -145,25 +146,46 @@ void KmdMutex::lock() {
...
+            error::RuntimeError,
+            fmt::format("tt_lock_test() failed for lock {} on {} errno: {}", lock_index_, device_path_, -result));
+        return state;
+    };
+
+    // Held by this handle means held by this process, which is as precise as KMD gets: it tracks which handle holds a
+    // lock, not which thread, so there is no tid to report. Any other holder is behind a handle we know nothing about.
+    auto owner_of = [](uint32_t state) -> std::optional<std::pair<pid_t, pid_t>> {
+        if ((state & TT_LOCK_STATE_HELD_BY_ANY) == 0) {
+            return std::nullopt;
```
nbuncicTT: So this means it's not held by any? If you stick to it, make a short comment on the return value in the code itself.

device/utils/kmd_mutex.cpp
```diff
@@ -145,25 +146,46 @@ void KmdMutex::lock() {
...
+        }
+        if ((state & TT_LOCK_STATE_HELD_BY_SELF) != 0) {
+            return std::make_pair(getpid(), static_cast<pid_t>(0));
```
nbuncicTT: So you never actually give back the tid? Why do we have it in the interface then? Because the `/dev/shm` lock had it?

device/api/umd/device/utils/robust_mutex.hpp
nbuncicTT: Nobody wants to use the `probe_lock` + `lock` mechanism because you have the time of check vs time of use problem there. I want to have for async APIs a `try_acquire` API, hence it should be public.

nbuncicTT (approved):
Please make the `try_acquire` mechanism public as it is much more useful than just the `probe_lock` mechanism. I'll close this PR: https://github.com/tenstorrent/tt-umd/pull/3459 in favor of this one.

Issue #2754, opened 2026-06-01
Extract private api


broskoTT:
The private header files should not be in api/ folder.
Move everything that is not part of the interface out of api.
Everything wh/bh or simulation specific should be moved out. At least.

https://tenstorrent.atlassian.net/wiki/spaces/LLK/pages/2283405852/Repo+restructuring

PR #3485, opened 2026-09-21
Extract private headers out of api/

ndjekicTT:
### Issue
#2754 

### Description

Extracting private headers from `/api` directory according to repo restructuring mentioned [here](https://tenstorrent.atlassian.net/wiki/spaces/UMD/pages/2283405852/Repo+restructuring)

### List of the changes

- Moved internal only header files to `/src` directories.

### Testing
- No functional/behavioral code changed — this is a pure header-location and include-path refactor.
- Full clean rebuild via `cmake . && ninja` (library, all unit test binaries, tools, and the nanobind Python bindings) succeeds with zero errors.
- CI/CD run

### API Changes
This PR has (no) API changes.


device/chip/mock_chip.hpp
broskoTT: This doesn't make sense as this should be part of public api, it's just that nobody uses it.
Please revert because that makes more sense.
But anyway, we'll likely going to remove this oneday

broskoTT (approved):

device/chip/mock_chip.hpp
ndjekicTT: Resolved

PR #3489, opened 2026-09-21
Simulation: stop reporting servers whose host is gone, and clear up after them

### LINKED ISSUE #3481: When process that hosted simulation server is killed, cleanup doesn't happen
It is OK to have files/directories on disk, but list servers should not return ones that are dead. Ideally, list servers can do cleanup of those dead servers...

pjanevskiTT:
### Issue

Closes #3481.

### Description

A host that is killed cannot remove its server directory, so its sockets stay on disk and
`list_servers()` went on reporting them. Nothing can attach to them, so they are not servers.
They are now left out, and clearing them up is part of enumerating the servers rather than
something anyone has to remember: the sweep lives in the one scan of every server directory, so
any operation that looks at all of them gets it. An operation aimed at a directory the caller
named is deliberately not built on that scan -- it still reports the truth about that directory
instead of deleting it underneath them.

### List of the changes

- `SimulationConnector::list_servers()` leaves out a server whose host is gone and clears up what
  it left behind.
- `SimulationConnector::prune_dead_servers()` is that same sweep reporting the other half, for
  clearing up without listing, or naming what was cleared. Exposed to Python alongside it.
- `sim_server` takes the matching per-server log whenever it enumerates, so logs cannot outlive
  the directories they belong to.
- The Python walkthrough no longer claims listing connects to nothing: it probes every socket and
  clears up what it finds dead.
- A directory with no socket yet is never swept: it cannot be told apart from a host still coming
  up.

### Testing

`SimulationConnector.*` passes (17 tests) and the nanobind suite passes (8 tests), both with
`TT_UMD_SIMULATOR` on a wormhole `libttsim.so`. New cases, in C++ and mirrored in Python: listing
leaves out and clears a dead server; prune reports what it cleared and is idempotent; a socket-less
directory survives both; a live host survives both. The Python cases need no simulator -- they use
a stale-socket helper mirroring the C++ one. By hand: SIGKILL a detached host, then a bare
`sim_server list` removes its directory and its log.

### API Changes

`SimulationConnector::list_servers()` no longer returns servers whose host is gone, and removes
their directories as it scans. New `SimulationConnector::prune_dead_servers()`, bound in Python.

### Full stack diff

https://github.com/tenstorrent/tt-umd/compare/main...pjanevski/sim-connector-cluster-descriptor


device/api/umd/device/simulation/simulation_connector.hpp
```diff
@@ -99,9 +99,17 @@ class SimulationConnector {
     static std::filesystem::path allocate_server_directory();
 
     // The simulation servers currently open on this machine, ordered by index, discovered by
-    // scanning the well-known server directories (the same directories a client attaches to). Does
-    // not connect to them. Exposed for management tooling (list / kill) without opening devices.
+    // scanning the well-known server directories (the same directories a client attaches to). A
```
nbuncicTT: nit: remove well-known, we don't need poetic adjectives to explain technical details

device/api/umd/device/simulation/simulation_connector.hpp
```diff
@@ -99,9 +99,17 @@ class SimulationConnector {
...
+    // server whose host is gone is left out, and what it left on disk is cleared up: every
+    // enumeration of all servers clears up, so no caller has to ask. Opens no devices. Exposed for
+    // management tooling (list / kill).
     static std::vector<SimulationServerInfo> list_servers();
+
+    // That same sweep, reporting the other half: the servers whose host was gone, now removed. For
```
nbuncicTT: nit: I dislike that this comment isn't a standalone comment and it appears as a continuation of the first previous one - so make it independent and shorter (i.e. concise).

device/simulation/simulation_connector.cpp
```diff
@@ -95,6 +96,56 @@ Classification classify(const std::filesystem::path& simulator_path) {
     return {SimulationConnector::Role::Client, std::nullopt, std::move(live_sockets)};
 }
 
+// A server is gone when it published sockets and not one of them still has a listener: the host
+// that bound them died without tearing them down. A directory with no socket at all is not gone --
+// it reads exactly like a host that has claimed its directory and not bound yet -- so neither the
+// listing nor the sweep may act on it.
+bool server_is_gone(const SimulationServerInfo& server) {
```
nbuncicTT: nit: `is_server_gone` would be more in the form of a question, so I'd prefer that name

device/simulation/simulation_connector.cpp
```diff
@@ -95,6 +96,56 @@ Classification classify(const std::filesystem::path& simulator_path) {
...
+    return !server.sockets.empty() &&
+           std::none_of(server.sockets.begin(), server.sockets.end(), [](const auto& socket) {
+               return SimulationServerSocket::is_live(socket.second);
+           });
+}
+
+// Clears up after a server that is gone. Best-effort, and quiet about failing: the temp directory
+// is sticky and simulation sockets are deliberately cross-user, so another user's leftovers are not
```
nbuncicTT: nit: again, "sticky" isn't a technical term, so for this and other comments, try to use technical terms in fewer words

nbuncicTT (approved):

